PR #743, opened 2014-04-15
Added basic account manager

edumntg:
Added an account manager with: Create account and character feature. 

This is a basic Account Manager for acc-pass 1/1.


tiagomelojuca:
> No, not only is it a code maintainability problem; but this kind of logic doesn't belong to the core and the code in this pull request is also of too poor quality.

Indeed, this is a nice point. But "too poor quality" is strong, isn't it? Despite the styles/coding standards issues, what are so wrong with this code? To be honest, seens just like he followed in a similar way what was already implemented. Just would like to understand, if possible; thinking about start contributting, but seens I'm not ready yet

Issue #775, opened 2014-05-02
onMove event?


Nu77:
Now us no have more mayNotMove function for use... It's good if someone can add this event, is necessary to do some things and it's very usefull.
##

PR #839, opened 2014-05-31
Added onMove Event

PrinterLUA:
https://github.com/otland/forgottenserver/issues/775


whitevo:
about that, has any changes come to this yet?

How about making it so if onStepOut or onStepIn returns false then, step will not take place?

PR #1061, opened 2014-11-16
onDeath moved to event and loot message moved to lua.

conde2:
cropse in onDeath is now a metatable and contains the loot (before it was
empty.)


comedinha:
no news?

marksamman:
Since I followed up here: https://github.com/otland/forgottenserver/pull/1066#issuecomment-63397342, only item userdata has been added so far. No decision has been made about creaturescripts vs. events for events that are applied to creatures. It's unlikely to be merged because it breaks backwards compatibility.

PR #1067, opened 2014-11-18
Monster event.

ReyAleman:
Added event:

```lua
Monster:onSpawn(position, isStartup)
```


gugahoa:
This is quite useful to register creature events for every monster. For example I moved drop loot from C++ to Lua, and I had to register the event on every single .xml of the monsters. onSpawn would make this a lot easier and not as hack'ish.

Issue #566, opened 2014-02-15
Walk through non-pvp players


Turox:
Im facing a problem on my server, in my config  protectionLevel = 50. A player under lvl 50 cant be attacked.
But its impossible to walk through this under lvl 50 players.
Sadly, my players are creating alot of underlevel'd accs to trap another players in combat, its possible to make walk through the underlevel'd (non-pvp) players ?
##

PR #1083, opened 2014-12-05
Walkthrough players lower than protection level

PrinterLUA:
Now it's possible to walkthrough those who are under protection level.

Also closing this issue:
https://github.com/otland/forgottenserver/issues/566

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -957,7 +957,7 @@ bool Player::canWalkthrough(const Creature* creature) const
 	}
 
 	const Tile* playerTile = player->getTile();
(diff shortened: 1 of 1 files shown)


gugahoa:
How does walkthrough work on global? It's not possible to walkthrough players that's keeping a door open?

If anything, make a walkthrough like in real Tibia. I'm up to help you if needed.

WibbenZ:
You can walk through doors, if they are "simple" doors, but if its an exp door or special door as Mark said, you can't.

marksamman:
Closing this as there hasn't been any progress on finding a solution to the problem I raised with this pull request.

PR #1084, opened 2014-12-05
Login protection

PrinterLUA:
Now players will not be able to get attacked or damaged when they login after certain seconds.


WibbenZ:
What was the reason for closing this? Seems like a cool thing to have if the player(s) is/are kicked from the server.

dalkon:
Probably because nothing has happened since December 9th.

WibbenZ:
Yes, but nothing else has been pointed out as wrong or bugged :p

Codinablack:
@WibbenZ maybe you didn't see it pointed out on here, but it was pointed out elsewhere that multiple commits like that must be squashed, and he told printer specifically that some of his needed to be squashed before he would add them...

PR #1129, opened 2015-01-02
Added loot correctly

Vichoko:
The monster: Spidris Elite wasn't giving any loot because the monster file didn't have any loot.


data/monster/Insects/spidris elite.xml
```diff
@@ -26,4 +26,21 @@
 	<elements>
 		<element holyPercent="-10"/>
 	</elements>
+		<loot>
```
marksamman: Indentation.

PR #1130, opened 2015-01-03
Fixed loot bug: thornback tortoise.xml

Vichoko:
Instead of a thorn, this turtle was dropping war hammers with 16% of chance because of an error in the id.


marksamman:
Thanks!

PR #1153, opened 2015-01-18
Fixes #1135

### LINKED ISSUE #1135: Reports directory
Nothing major, but the 'reports' folder is missing in the data directory.
If someone sends a report and you don't have a 'reports' folder, you will not receive it.

https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L5040

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7629368-reports-directory?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dominique120:
We can fix it like this or write code to create the folder if it does not exist. This way is the easiest and least problematic.

### DIFF
--- a/data/reports/.gitignore
+++ b/data/reports/.gitignore
@@ -1,0 +1,6 @@
+# We dont want people to upload reports so we will ignore them.
+*
+
+# We still want to keep note.txt and the .gitignore file though.
+!.gitignore
+!note.txt
--- a/data/reports/note.txt
+++ b/data/reports/note.txt
@@ -1,0 +1,1 @@
+In this directory you will find player reports.


diath:
This would be more appropriate: http://stackoverflow.com/a/932982

dominique120:
The note is there on purpose to tell people about this new folder not because I could not commit an empty directory :P

However this could be used to avoid accidental uploading of reports to this main repo. Lets see what mark has to say and thanks for the idea!

diath:
> However this could be used to avoid accidental uploading of reports to this main repo. Lets see what mark has to say and thanks for the idea!

Exactly my concerns.

dominique120:
Everything will be ignored except our .gitignore file and the note.txt file.

kornholi:
Thanks!

Issue #1156, opened 2015-01-19
Crash client after the recent changes to the code


RestlesPL:
Please see
http://otland.net/threads/10-41-crash-recent-amendments.227462/
and
https://www.youtube.com/watch?v=PIrMMt1_YWE

RestlesPL:
When this will be fixed?

TheSumm:
If you know the exact commit after which this started happening, it would be good to post it and would probably speed the process up.

jollejkpg:
Otclient does not work either.
Client 10.41

jollejkpg:
OTClient error:


ERROR: got a network message with invalid checksum
at:
    [C++]: ?internalRecvData@Protocol@@AAEXPAEG@Z

PR #1158, opened 2015-01-19
Fix networkmessage issues

TheSumm:
Fixes issue #1156 and bugs/typos introduced in baa9263.


jollejkpg:
Yey :) it works

marksamman:
Thanks!

PR #1166, opened 2015-01-26
onMove creature event.

gugahoa:
I'm afraid this need lots of improvement, but it's a start (right?). How can we make this better? An onMove event would help #1083, or at least help make a workaround.

I noticed that calling the creatureEvent "moveEvent" is not good (ambiguity) but I could not think of another name for it. onMoveEvent?


marksamman:
The code is mostly fine, I would suggest the following changes:
* stop at the first event that returns false
* use RETURNVALUE's instead of true/false, and return that from internalMoveCreature

This pull request will have to wait until we have changed onStepIn/onStepOut to use the return value to stop movements though.

gugahoa:
I'm new to github, I'm working on resolving the merge conflict.

gugahoa:
I'll squash the commits when you ask.

Codinablack:
I'll ask, please squash the commits, there are many of people who are waiting and would love to see this implemented... Would love to see this merged

gugahoa:
I'm completely lost about how to squash, but I'm trying.

PR #1169, opened 2015-01-29
Use table in changeGold.lua

PrinterLUA:


marksamman:
Address the following and squash your commits:
* x - x is always 0, also why transform to 0?
* item is already a userdata, so Item(item.uid) is unnecessary
* restore the new line at the end of the file

PrinterLUA:
Done ^^

marksamman:
Thank you, now just squash your commits and I'll merge this.

PrinterLUA:
Done:)

marksamman:
I'm not sure why you wrote squash in the commit message, the point of squashing your commits is to have one commit represent your pull request which describes what has been changed in the commit. Explaining that you have squashed your previous commits is not very relevant in the commit log. Even something as simple as "Update changegold.lua" is good enough in this case. Update the commit message and comment again and I'll merge.

PrinterLUA:
Oh did not know, that would be the title. Alright here you go ^^

PR #1172, opened 2015-02-01
creature getLastPosition() function

PrinterLUA:


marksamman:
I would rather get rid of lastPosition from Creature than making us more dependent on it.

PrinterLUA:
Oh okay, i found it useful to custom scripts. Such as send effect behind player or teleport to last death place ^^ But if you are going to make a another way, i'll wait then.

PR #1182, opened 2015-02-06
Moved regenerateStamina, useStamina and exp multiplier (staminaChange as well) to Lua

ninjalulz:


data/events/scripts/player.lua
```diff
@@ -110,6 +110,10 @@ local soulCondition = Condition(CONDITION_SOUL, CONDITIONID_DEFAULT)
 soulCondition:setTicks(4 * 60 * 1000)
 soulCondition:setParameter(CONDITION_PARAM_SOULGAIN, 1)
 
+local applyMultiplier = true
+local applyStaminaChange = true
+local nextUseStaminaTime = 0
```
diath: Shouldn't this be unique per player?

ninjalulz: It should, thanks for pointing it out. :P

marksamman:
This changes the behavior of Player.addExperience, which makes it apply stamina and experience stage multiplier. I'm not sure what the best way to fix it is, perhaps checking source. You should also factor out the stamina code to its own local function in that file and decrease the indentation level on the code by immediatelly returning if configManager.getBoolean(configKeys.STAMINA_SYSTEM) is false.

ninjalulz:
* Moved the stamina code to a local function,
* Player.addExperience is no longer affected by the stamina and experience stage multiplier.

marksamman:
Thanks!

PR #1183, opened 2015-02-06
Fix potions

Mitthryl:


marksamman:
Rebase off master and rename itemEx to target.

PR #1196, opened 2015-02-13
Added immunity to invis. to Jaul.

slayingworld:


EPuncker:
why not make a single pull request for all files?

slayingworld:
cause i have no idea how to use this lmao!  Still getting used to it :[ sorry!

marksamman:
I'll let this go since it's your first contribution and I don't want to discourage you from contributing again, but in the future please create a single pull request for related changes. The changes could also have been made in a single commit instead of three separate commits.

You would simply use the same branch for the commits; and to change multiple files in the same commit you may have to use the Git software instead of GitHub's web interface.

slayingworld:
Sorry about that, this is my first attempt to make a contribution, but i already learned about it. 
Wont happen again, thanks again in advance -slayingworld

PR #1224, opened 2015-03-11
Fix field runes PZ block

### LINKED ISSUE #986: Throwing field runes give players PZ block regardless of PVP contact
I apologize if this has been reported already, I had assumed it was, but I couldn't find anything related.

**Description of bug**
Basically, using any field rune will cause the player to have a bloodied-sword (red) PZ block condition, even though a monster/player is nowhere near point of impact.

**Runes tested**
Fire Field: Gives PZ block
Poison Field: Gives PZ block
Energy Field: Gives PZ block

**Test on TFS 1.0**
You can see on the right, I have the PZ block condition (bloodied swords).  No creature or player has stepped on the field.  Entrance to a PVP zone is forbidden.
ImageOfTFSTest

**Test on Tibia**
I figured (because I may be a bit behind on Tibia stuff) that this was the case because it's the same in Tibia, so I tested it on one of my characters.  You can see, throwing a fire field gives me a non-blooded sword condition that would allow me to enter PZ.
ImageOfTibiaTest

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4401676-throwing-field-runes-give-players-pz-block-regardless-of-pvp-contact?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ninjalulz:
Fixes #986

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -629,7 +629,7 @@ void Combat::combatTileEffects(const SpectatorVec& list, Creature* caster, Tile*
 						itemId = ITEM_ENERGYFIELD_NOPVP;
 					}
 				} else if (itemId == ITEM_FIREFIELD_PVP_FULL || itemId == ITEM_POISONFIELD_PVP || itemId == ITEM_ENERGYFIELD_PVP) {
-					casterPlayer->addInFightTicks(true);
+					casterPlayer->addInFightTicks();
 				}
 			}
 		}
--- a/src/tile.cpp
+++ b/src/tile.cpp
@@ -590,10 +590,11 @@ ReturnValue Tile::queryAdd(int32_t, const Thing& thing, uint32_t, uint32_t flags
 				} else if (!hasFlag(TILESTATE_PVPZONE)) { //player is trying to leave a pvp zone while being pz-locked
 					return RETURNVALUE_PLAYERISPZLOCKEDLEAVEPVPZONE;
 				}
-			}
 
-			if ((hasFlag(TILESTATE_NOPVPZONE) || hasFlag(TILESTATE_PROTECTIONZONE)) && player->isPzLocked()) {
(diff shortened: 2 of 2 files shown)


marksamman:
Will this give PZ lock to the caster if a player walks into the field afterwards? What happens if the caster has walked into PZ, will the caster be locked in that position until the PZ lock wears off? Is this the intended behavior?

TheSumm:
It should be intended that the caster can gain PZ lock in a protection zone.

But [tile.cpp](https://github.com/otland/forgottenserver/blob/2550261c2001fa98c64b72c1e0c7c6c13b5d45a8/src/tile.cpp#L595-L597) should probably be adjusted to only disallow walking if you try to change the zone rather then just checking what the next tile's zone is and not considering your current zone.

```
if (player->getTile() && player->isPzLocked()) {
	if (!player->getTile()->hasFlag(TILESTATE_NOPVPZONE) && hasFlag(TILESTATE_NOPVPZONE) || !player->getTile()->hasFlag(TILESTATE_PROTECTIONZONE)
			&& hasFlag(TILESTATE_PROTECTIONZONE)) {
		return RETURNVALUE_PLAYERISPZLOCKED;
	}
}
```

ninjalulz:
Thanks @TheSumm, seems to work properly now. Apparently it didn't go that well to add the code inside L584-593 x)

TheSumm:
Just realized that it should probably be merged with the previous if block.

EPuncker:
great

marksamman:
Thanks!

PR #1228, opened 2015-03-12
Last client version

comedinha:
I did some things wrong in that other pull request, created another for those who want to update better.

@marksamman If you find it necessary can lock again


Tazer89:
Well done @comedinha :+1:

WibbenZ:
Must say great work, but what is left to make it 100%?
Would try and help you out if I get some info of what is missing :p

comedinha:
Lack new functions... Blessings, Guild Management, Party System and others
I believe that the system issue is ready. I just changed some things not very important...

WibbenZ:
Well I got pvp modes working, but that won't be accepted I think till all of the other parts in the war system is done.

comedinha:
I can see the code and implement? send me pm on otland

WibbenZ:
Pretty much just a set function in the parser the package gets handled from.
Won't be added 99% sure of that unless everything the new war system has to offer is done aswell.
Been trying to do that :p

rottyrubas:
why does it look like this for me? XD
http://1.ii.gl/ljbcr5_Je.png

WibbenZ:
Are you talking about the map or the outfit?
If you are using OTC and you mean the outfit then the thing is that they haven't fixed it yet.
If its the map, maybe your otbm file?

walissonsantos:
@rottyrubas 

Use item.otb of comedinha

https://mega.co.nz/#!W1sgEb6D!G_bVsGK75vOu9Uay2RBTitwCnI10IJknPCvoSOMPBBc

src/const.h
```diff
@@ -102,7 +102,13 @@ enum MagicEffectClasses : uint8_t {
 	CONST_ME_CONFETTI_VERTICAL = 76,
 	// 77-157 are empty
 	CONST_ME_BLACKSMOKE = 158,
-	CONST_ME_LAST = CONST_ME_BLACKSMOKE,
+	// 158-166 are empty
```
EPuncker: // 157-166 are empty

src/const.h
```diff
@@ -180,31 +190,42 @@ enum SpeakClasses : uint8_t {
 	TALKTYPE_CHANNEL_R1 = 14, //red - #c text
 	TALKTYPE_PRIVATE_RED_FROM = 15, //@name@text
 	TALKTYPE_PRIVATE_RED_TO = 16, //@name@text
-	TALKTYPE_MONSTER_SAY = 35,
-	TALKTYPE_MONSTER_YELL = 36,
+	TALKTYPE_MONSTER_SAY = 36,
```
EPuncker: that will debug so many scripts that don't use constants instead of variables :/

WibbenZ: Well if there is something ive learned by these things is that cipsoft does keep track of the code we use.
Just by changing stuff like that makes it harder to for us keep it up to date....

But at the same time, about time people start using the constants..

comedinha: This reminds me of an update to the tibia changed the colors and all scripts used the numbering instead of constants

EPuncker:
@comedinha hey make a new one just adding what mark missed, like the new effects :+1:

WibbenZ:
He already has :p #1263

EPuncker:
oh sorry missed it xD thanks

PR #1234, opened 2015-03-14
Added Tile methods (isCreature, isItem, isTile), Creature.isTile, ..

TheSumm:
Item.isTile and organized global.lua by Class names


marksamman:
Thanks!

PR #1241, opened 2015-03-20
Fix torch behavior

gugahoa:
There was a loophole in decayto.lua that caused torch 2053 to be endless (just use it and it'll go back to full) and torch 2055 to not be able to unlit.


marksamman:
Thanks!

PR #1243, opened 2015-03-21
Connection.cpp code & efficiency improvements

djarek:
Applied the RAII idiom to mutex locking in the Connection and ConnectionManager classes' methods. Use make_shared instead of initialization with a raw, owning pointer. This should improve performance by a small amount (make_shared groups memory allocation of the object and ref count structure together which means less allocation and better memory locality).

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -35,7 +35,7 @@ Connection_ptr ConnectionManager::createConnection(boost::asio::ip::tcp::socket*
 {
 	std::lock_guard<std::recursive_mutex> lockClass(m_connectionManagerLock);
 
-	Connection_ptr connection = Connection_ptr(new Connection(socket, io_service, servicer));
+	auto connection = std::make_shared<Connection>(socket, io_service, servicer);
 	m_connections.insert(connection);
 	return connection;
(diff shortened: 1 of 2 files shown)


marksamman:
Fix the build error and I'll merge this.

djarek:
Had to write a custom allocator for Connection, however, I've looked at the assembly with -Ofast and it's all properly inlined (even if it weren't we would still gain some performance from better memory locality when accessing instances of Connection through a shared_ptr).

marksamman:
I'd rather just make the Connection constructor public than having the extra code for a custom allocator as a workaround, because if we care about restricting access to the Connection constructor, we've already failed at that when any class can use ConnectionAllocator to create a Connection object now.

djarek:
There's also the possibility of moving the ConnectionAllocator struct  into the ConnectionManager and making it private there. Should I do that, or just move the Connection constructor to public?

marksamman:
That may preserve the restriction to the Connection constructor, but I'm still in favor of just moving the Connection constructor to public.

marksamman:
Thanks!

PR #1246, opened 2015-03-22
Player.cpp code improvements

djarek:
Use std::any_of() to make code more concise in Player::hasShopItemForSale() and Player::updateSaleShopList(). Remove unnecessary check in Player::internalAddThing(). As a result of these changes Player::updateSaleShopList() will be a little more efficient as it previously checked the validity of the index on every iteration when going over all items in the container.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -3162,12 +3162,8 @@ bool Player::updateSaleShopList(const Item* item)
 				return false;
 			}
 
-			for (uint32_t index = 0; index < container->size(); ++index) {
-				if (updateSaleShopList(container->getItemByIndex(index))) {
-					return true;
-				}
-			}
-			return false;
+			const auto& items = container->getItemList();
+			return std::any_of(items.begin(), items.end(), [this](const Item* item){return updateSaleShopList(item);});
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/player.cpp
```diff
@@ -3180,22 +3176,8 @@ bool Player::updateSaleShopList(const Item* item)
...
-		}
-
-		if (itemType.isFluidContainer() && shopInfo.subType != subType) {
-			continue;
-		}
-
-		return true;
-	}
-	return false;
+	return std::any_of(shopItemList.begin(), shopItemList.end(), [&](const ShopInfo& shopInfo) {
```
marksamman: Capture by value instead of reference.

src/player.cpp
```diff
@@ -3162,12 +3162,8 @@ bool Player::updateSaleShopList(const Item* item)
...
 			}
 
-			for (uint32_t index = 0; index < container->size(); ++index) {
-				if (updateSaleShopList(container->getItemByIndex(index))) {
-					return true;
-				}
-			}
-			return false;
+			const auto& items = container->getItemList();
+			return std::any_of(items.begin(), items.end(), [this](const Item* item){return updateSaleShopList(item);});
```
marksamman: Just a style nitpick, this is more readable to me (also renamed item to containerItem to not shadow the item parameter if we ever would use it in the lambda in the future):
```cpp
return std::any_of(items.begin(), items.end(), [this](const Item* containerItem) {
	return updateSaleShopList(containerItem);
});
```

src/player.cpp
```diff
@@ -3180,22 +3176,8 @@ bool Player::updateSaleShopList(const Item* item)
...
-
-		if (itemType.isFluidContainer() && shopInfo.subType != subType) {
-			continue;
-		}
-
-		return true;
-	}
-	return false;
+	return std::any_of(shopItemList.begin(), shopItemList.end(), [&](const ShopInfo& shopInfo) {
+		return shopInfo.itemId == itemId && shopInfo.buyPrice != 0 && (!itemType.isFluidContainer() || shopInfo.subType == subType); });
```
marksamman: Move "});" to a new line.

djarek:
Done.

marksamman:
Thanks!

Codinablack:
:+1:

PR #1251, opened 2015-03-25
Monsters memory management improvements

djarek:
Applied the RAII idiom to the MonsterType and Monsters classes. Changed lists in MonsterType to vector for slightly better performance (these containers are only iterated over in a linear fashion and rarely modified so a vector makes more sense than a list). Moved ownership of MonsterType objects from an owning pointer to the monsters map.  I made each logical change in a separate commit, however if you want I can squash them together.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -11682,14 +11682,14 @@ int LuaScriptInterface::luaConditionAddDamage(lua_State* L)
 // MonsterType
 int LuaScriptInterface::luaMonsterTypeCreate(lua_State* L)
 {
-	// MonsterType(id or name)
-	MonsterType* monsterType;
-	if (isNumber(L, 2)) {
-		monsterType = g_monsters.getMonsterType(getNumber<uint32_t>(L, 2));
-	} else {
-		monsterType = g_monsters.getMonsterType(getString(L, 2));
+	// MonsterType(name)
+	
+	if (!isString(L, 2)) {
+		lua_pushnil(L);
+		return 1;
 	}
-
+	MonsterType* monsterType = g_monsters.getMonsterType(getString(L, 2));
+	
 	if (monsterType) {
 		pushUserdata<MonsterType>(L, monsterType);
 		setMetatable(L, -1, "MonsterType");
@@ -11912,10 +11912,10 @@ int LuaScriptInterface::luaMonsterTypeGetAttackList(lua_State* L)
 		return 1;
 	}
 
-	lua_createtable(L, monsterType->spellAttackList.size(), 0);
+	lua_createtable(L, monsterType->attackSpells.size(), 0);
 
 	int index = 0;
-	for (const auto& spellBlock : monsterType->spellAttackList) {
+	for (const auto& spellBlock : monsterType->attackSpells) {
 		lua_createtable(L, 0, 8);
 
 		setField(L, "chance", spellBlock.chance);
@@ -11942,10 +11942,11 @@ int LuaScriptInterface::luaMonsterTypeGetDefenseList(lua_State* L)
 		return 1;
 	}
 
-	lua_createtable(L, monsterType->spellDefenseList.size(), 0);
+	lua_createtable(L, monsterType->defenseSpells.size(), 0);
+
 
 	int index = 0;
-	for (const auto& spellBlock : monsterType->spellDefenseList) {
+	for (const auto& spellBlock : monsterType->attackSpells) {
 		lua_createtable(L, 0, 8);
 
 		setField(L, "chance", spellBlock.chance);
@@ -12009,7 +12010,7 @@ int LuaScriptInterface::luaMonsterTypeGetLoot(lua_State* L)
 		return 1;
 	}
 
-	static const std::function<void(const std::list<LootBlock>&)> parseLoot = [&](const std::list<LootBlock>& lootList) {
+	static const std::function<void(const std::vector<LootBlock>&)> parseLoot = [&](const std::vector<LootBlock>& lootList) {
 		lua_createtable(L, lootList.size(), 0);
 
 		int index = 0;
@@ -12043,8 +12044,8 @@ int LuaScriptInterface::luaMonsterTypeGetCreatureEvents(lua_State* L)
 	}
 
 	int index = 0;
-	lua_createtable(L, monsterType->scriptList.size(), 0);
-	for (const std::string& creatureEvent : monsterType->scriptList) {
+	lua_createtable(L, monsterType->scripts.size(), 0);
+	for (const std::string& creatureEvent : monsterType->scripts) {
 		pushString(L, creatureEvent);
 		lua_rawseti(L, -2, ++index);
 	}
@@ -12061,8 +12062,8 @@ int LuaScriptInterface::luaMonsterTypeGetSummonList(lua_State* L)
 	}
 
 	int index = 0;
-	lua_createtable(L, monsterType->summonList.size(), 0);
-	for (const auto& summonBlock : monsterType->summonList) {
+	lua_createtable(L, monsterType->summons.size(), 0);
+	for (const auto& summonBlock : monsterType->summons) {
 		lua_createtable(L, 0, 3);
 		setField(L, "name", summonBlock.name);
 		setField(L, "speed", summonBlock.speed);
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -37,6 +37,13 @@ extern Spells* g_spells;
 extern Monsters g_monsters;
 extern ConfigManager g_config;
 
+spellBlock_t::~spellBlock_t()
+{
+	if (combatSpell) {
+		delete spell;
+	}
+}
+
 MonsterType::MonsterType()
 {
 	reset();
@@ -62,14 +69,7 @@ void MonsterType::reset()
 	health = 100;
 	healthMax = 100;
 
-	outfit.lookHead = 0;
-	outfit.lookBody = 0;
-	outfit.lookLegs = 0;
-	outfit.lookFeet = 0;
-	outfit.lookType = 0;
-	outfit.lookTypeEx = 0;
-	outfit.lookAddons = 0;
-	outfit.lookMount = 0;
+	outfit.reset();
 	lookcorpse = 0;
 
 	skull = SKULL_NONE;
@@ -87,23 +87,12 @@ void MonsterType::reset()
 	lightColor = 0;
 
 	manaCost = 0;
-	summonList.clear();
+	summons.clear();
 	lootItems.clear();
 	elementMap.clear();
 
-	for (const spellBlock_t& spellBlock : spellAttackList) {
-		if (spellBlock.combatSpell) {
-			delete spellBlock.spell;
-		}
-	}
-	spellAttackList.clear();
-
-	for (const spellBlock_t& spellBlock : spellDefenseList) {
-		if (spellBlock.combatSpell) {
-			delete spellBlock.spell;
-		}
-	}
-	spellDefenseList.clear();
+	attackSpells.clear();
+	defenseSpells.clear();
 
 	yellSpeedTicks = 0;
 	yellChance = 0;
@@ -119,12 +108,7 @@ void MonsterType::reset()
 	creatureSayEvent = -1;
 	thinkEvent = -1;
 
-	scriptList.clear();
-}
-
-MonsterType::~MonsterType()
-{
-	reset();
+	scripts.clear();
 }
 
 uint32_t Monsters::getLootRandom()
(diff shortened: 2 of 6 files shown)


src/monsters.cpp
```diff
@@ -62,14 +61,7 @@ void MonsterType::reset()
...
 
-	outfit.lookHead = 0;
-	outfit.lookBody = 0;
-	outfit.lookLegs = 0;
-	outfit.lookFeet = 0;
-	outfit.lookType = 0;
-	outfit.lookTypeEx = 0;
-	outfit.lookAddons = 0;
-	outfit.lookMount = 0;
+	//No need to initialize outfit
```
marksamman: You can remove this comment.

src/monsters.cpp
```diff
@@ -220,9 +196,9 @@ std::list<Item*> MonsterType::createLootItem(const LootBlock& lootBlock)
 			tmpItem->setText(lootBlock.text);
 		}
 
-		itemList.push_back(tmpItem);
+		itemList.emplace_back(tmpItem);
```
marksamman: push_back was fine here, I'm not sure exactly what emplace_back does under the hood when using pointers (for objects it does in-place construction), I guess it differs depending on implementations, but in the best case it will do exactly what push_back does, in a worse case it will have performance overhead.

src/monsters.cpp
```diff
@@ -767,9 +715,25 @@ bool Monsters::loadMonster(const std::string& file, const std::string& monster_n
 		return false;
 	}
 
-	if (mType == nullptr) {
-		mType = new MonsterType();
+	if (reloading) {
+		uint32_t id = getIdByName(monster_name);
```
marksamman: Try not to overshadow other variables, this overshadows the static id. You could just move the static id down a bit and keep the names.

src/monsters.cpp
```diff
@@ -767,9 +715,25 @@ bool Monsters::loadMonster(const std::string& file, const std::string& monster_n
...
+			if (mType != nullptr) {
+				new_mType = false;
+				mType->reset();
+			}
+		}
+	}
+
+	if (new_mType) {
+		std::string lowername = monster_name;
+		toLowerCaseString(lowername);
```
marksamman: Use asLowerCaseString here instead.

src/tools.h
```diff
@@ -94,4 +94,89 @@ inline int64_t OTSYS_TIME()
 	return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
 }
 
+template<typename T>
+class ToggledPtr final
```
marksamman: This class should be moved to its own file, call it toggled_ptr.h

src/monsters.cpp
djarek: For built-in types (ints, pointers etc.) emplace_back works exactly the same as push_back. For complex types it behaves differently - it calls the move constructor if provided a copy or reference of an object, otherwise (eg. when given a const or const reference) it resorts to the copy constructor, if provided a list of arguments it actually in-place constructs the object and forwards them to the constructor.

marksamman:
There's no need to squash the commits.

As for the script interface question, I'm not sure. It looks to me like there's only a single allocation in Monsters::loadFromXml which is used by all monster types through a pointer, so storing an object would mean significantly higher memory use.
/cc @dalkon

djarek:
Oh, I made a typo in the message - I meant in the Monsters class.

src/monsters.cpp
```diff
@@ -1187,21 +1151,14 @@ bool Monsters::loadMonster(const std::string& file, const std::string& monster_n
...
 
-	static uint32_t id = 0;
-	if (new_mType) {
-		std::string lowername = monster_name;
-		toLowerCaseString(lowername);
-
-		monsterNames[lowername] = ++id;
-		monsters[id] = mType;
-	}
+	mType->shrinkToFit();
```
djarek: Removed shrinkVectors, however, I still think it's better to keep the MonsterType::shrinkToFit() method - Monsters::loadMonster() is long enough already ;).

djarek:
Rebased against master and moved shrink_to_fit calls to loadMonster. Is there anything else to fix in here?

marksamman:
You don't need to change anything else at the moment. There was just some discussion about this PR internally which is why it hasn't been merged yet.

dalkon:
I don't see the point of "ToggledPtr" at all, it's best to wait for another time and change both spells and combat spells to smart pointers at the same time rather than having one or the other.

djarek:
Just a nitpick, ToggledPtr is a smart pointer... :D. This was mostly intended to actually fix the issue (which is that a reload of spells or monsters causes undefined behavior). A proper fix requires a redesign and "assignment" of ownership (or using ref counts), however, once you use a smart pointer, swapping it for a different kind of smart pointer should be fairly easy. The question is: do we really need ref counts for spells? There certainly are benefits (you could i.e. safely store references to spell objects in lua). Even if we decide to use ref counting for Spell objects, we would still need to create our own version of a shared_ptr - in a single threaded environment shared_ptr is certainly too heavy.

dalkon:
I know that ToggledPtr is a smart pointer but as I said, if you want to change combat spells to smart pointers then normal spells should also be smart pointers and this should be done at the same time, having one or the other makes no sense.

djarek:
Removed Toggled_ptr. I previously didn't spot in all the deserializer "noise" what was the actual purpose of spellBlock_t::combatSpell. Is there any other issue with this PR?

src/monsters.cpp
```diff
@@ -37,6 +36,13 @@ extern Spells* g_spells;
 extern Monsters g_monsters;
 extern ConfigManager g_config;
 
+spellBlock_t::~spellBlock_t()
+{
+	if (spell && combatSpell) {
```
marksamman: You don't need to check if spell isn't null. It's safe to call delete on a null pointer.

src/luascript.cpp
```diff
@@ -11912,10 +11912,10 @@ int LuaScriptInterface::luaMonsterTypeGetAttackList(lua_State* L)
 		return 1;
 	}
 
-	lua_createtable(L, monsterType->spellAttackList.size(), 0);
+	lua_createtable(L, monsterType->attackSpells.size(), 0);
 
-	int index = 0;
-	for (const auto& spellBlock : monsterType->spellAttackList) {
+	int32_t index = 0;
```
marksamman: Revert this to int. It's only used to pass the index to lua_rawseti, which takes int as argument.

src/luascript.cpp
```diff
@@ -11942,10 +11942,11 @@ int LuaScriptInterface::luaMonsterTypeGetDefenseList(lua_State* L)
 		return 1;
 	}
 
-	lua_createtable(L, monsterType->spellDefenseList.size(), 0);
+	lua_createtable(L, monsterType->defenseSpells.size(), 0);
 
-	int index = 0;
-	for (const auto& spellBlock : monsterType->spellDefenseList) {
+
+	int32_t index = 0;
```
marksamman: Revert this to int. It's only used to pass the index to lua_rawseti, which takes int as argument.

src/luascript.cpp
```diff
@@ -12042,9 +12043,9 @@ int LuaScriptInterface::luaMonsterTypeGetCreatureEvents(lua_State* L)
 		return 1;
 	}
 
-	int index = 0;
-	lua_createtable(L, monsterType->scriptList.size(), 0);
-	for (const std::string& creatureEvent : monsterType->scriptList) {
+	int32_t index = 0;
```
marksamman: Revert this to int. It's only used to pass the index to lua_rawseti, which takes int as argument.

src/luascript.cpp
```diff
@@ -12060,9 +12061,9 @@ int LuaScriptInterface::luaMonsterTypeGetSummonList(lua_State* L)
 		return 1;
 	}
 
-	int index = 0;
-	lua_createtable(L, monsterType->summonList.size(), 0);
-	for (const auto& summonBlock : monsterType->summonList) {
+	int32_t index = 0;
```
marksamman: Revert this to int. It's only used to pass the index to lua_rawseti, which takes int as argument.

src/monsters.h
```diff
@@ -181,8 +196,8 @@ class Monsters
 
 		std::map<std::string, uint32_t> monsterNames;
 		std::map<MonsterType*, std::string> monsterScriptList;
-		std::map<uint32_t, MonsterType*> monsters;
-		LuaScriptInterface* scriptInterface;
+		std::map<uint32_t, MonsterType> monsters;
```
marksamman: Change the key to string. I don't see anything taking advantage of a uint32_t key here, we're always getting the id by looking up the name in monsterNames. You can also remove monsterNames while you're at it.

src/luascript.cpp
djarek: BTW this occurs in other places in luascript.cpp as well (e.g. in setMetatable()).

marksamman: It's not a big deal since int32_t is the same as int on most systems today, but I rather match the Lua API types where reasonable (i.e. in this case we're only using index to pass it to lua_rawseti and not in any context where it would make sense to have it as int32_t).

If you would like to update setMetatable() and other occurrences, please open a separate PR for that so that this one doesn't get cluttered with unrelated changes.

src/monsters.h
djarek: This will be a breaking change - we will be no longer able to create a lua MonsterType metatable by using the uid.

dalkon: You cannot get the id of a monster type in lua anyways so its fine.

marksamman: That's not a problem. I don't think MonsterType(id) was practical anyway, where'd you get the id from?

djarek: I just thought you should put that in the future changelog for 1.2. Something tells me someone actually did use that even though it makes no sense...

djarek:
Removed monster IDs. We could also store monster scripts names as pairs (Monster*, std::string) in a vector to save memory as they're actually used only on load/reload (we could in fact even destroy it after a load/reload to save memory). MonsterType has 9 bools in it, we could use a std::bitset to compact it and save 7 bytes (not much, but still something ;), plus saves some performance by not polluting the CPU cache).

src/monsters.cpp
```diff
@@ -1259,34 +1220,17 @@ void Monsters::loadLootContainer(const pugi::xml_node& node, LootBlock& lBlock)
...
-	if (mId == 0) {
-		return nullptr;
-	}
-	return getMonsterType(mId);
-}
-
-MonsterType* Monsters::getMonsterType(uint32_t mid)
-{
-	auto it = monsters.find(mid);
+	auto it = monsters.find(name);
```
marksamman: asLowerCaseString(name)

src/monsters.cpp
```diff
@@ -769,9 +723,18 @@ bool Monsters::loadMonster(const std::string& file, const std::string& monster_n
...
+	if (reloading) {
+		mType = getMonsterType(monster_name);
+		if (mType != nullptr) {
+			new_mType = false;
+			mType->reset();
+		}
 	}
+
+	if (new_mType) {
+		mType = &monsters[monster_name];
```
marksamman: asLowerCaseString(monster_name)

src/luascript.cpp
```diff
@@ -11683,13 +11683,13 @@ int LuaScriptInterface::luaConditionAddDamage(lua_State* L)
 int LuaScriptInterface::luaMonsterTypeCreate(lua_State* L)
 {
 	// MonsterType(id or name)
```
marksamman: Change the comment to MonsterType(name)

src/luascript.cpp
```diff
@@ -11683,13 +11683,13 @@ int LuaScriptInterface::luaConditionAddDamage(lua_State* L)
...
-		monsterType = g_monsters.getMonsterType(getNumber<uint32_t>(L, 2));
-	} else {
-		monsterType = g_monsters.getMonsterType(getString(L, 2));
+	
+	if (!isString(L, 2)) {
+		lua_pushnil(L);
+		return 1;
 	}
-
+	auto monsterType = g_monsters.getMonsterType(getString(L, 2));
```
marksamman: I prefer to know the type instead of auto when it isn't too long, please don't overuse auto.

marksamman:
That kills maintainability for a class that is unlikely to have more than 1000 instances, and I think your calculations are wrong.

9 bools: 9 bytes
bitset with <= 64 bits: 8 bytes on a 64-bit system with g++'s bitset implementation

You're saving 1 byte per monster on a 64-bit system, which is barely a kilobyte in total when you have 1000 monster type instances.

src/luascript.cpp
djarek: Well, the type is already "encoded" in the name of the variable and the return value of the function (and the function name itself), I though it would be needless to say it even once more.

djarek:
I guess I shouldn't have assumed anything about the way std::bitset is implemented...

marksamman:
Looks good to me. Please squash your commits.

djarek:
Done.

marksamman:
Thanks!

Issue #470, opened 2014-01-20
Frag/Skull/Death system


Luckey1729:
Hi,

This is a simple enhancement that I think a lot of server owners would find beneficial.
I think the frag system, skull system, and death system should all be written in Lua. As they all somewhat work together, 

I think it would be beneficial to have them written in Lua. This enhancement would allow server owners to make their own frag rules, power abuse rules, and death rules. Plus, a lot of the functions that would make this work already exist.
##

PR #1261, opened 2015-03-29
Moved Red Skull and Black Skull Handler to Lua. Part of #470.

Luckey1729:
This moves Red Skull and Black Skull handling from the source to Lua. It doesn't include White Skull handling or PZ lock handling but it's definitely a start.

### DIFF
--- a/data/creaturescripts/creaturescripts.xml
+++ b/data/creaturescripts/creaturescripts.xml
@@ -6,5 +6,7 @@
 	<event type="login" name="RegenerateStamina" script="regeneratestamina.lua"/>
 	<event type="death" name="PlayerDeath" script="playerdeath.lua"/>
 	<event type="death" name="DropLoot" script="droploot.lua"/>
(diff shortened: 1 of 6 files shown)


marksamman:
You won't be able to decide whether the last hit was unjustified or not in onDeath because that must already be decided before onDeath as lasthitunjustified is passed to onDeath. You could remove that code from the Lua frag handler and just check lasthitunjustified (meaning it'll still be hardcoded), then at least the code that sends the unjustified kill message, updates skull time and skull will be more customizable.

Having to adapt to the current system adds a lot of complexity (i.e. most chunks of the death flow are spread out over different scripts), and I would prefer moving the entire death flow (creating corpse, blood splash, dropping loot, etc.) to Lua, probably a Creature:onDeath() event (data/events/), however that requires so many changes that it won't make it into 1.1.

ranisalt:
Closing as the author did not continue to work on this and it now contains conflicts. If @Luckey1729 wants to keep working I'll reopen.

PR #1263, opened 2015-03-29
Small changes and fixes

comedinha:
Food Special by @WibbenZ


EPuncker:
conflicts needs to be solved and I suggest that you do a separated pull for the food script so we can discuss it better :D

comedinha:
removed food special :v

PR #1277, opened 2015-04-07
onShutdown execute properly when closing the server.

EvilHero90:
I've noticed an issue on windows (not sure if linux has that aswell)
that if you closed the console and didn't use the ingame command to set the game state to shutdown, that it wouldn't execute onShutdown() globalevent at all.
This fix worked for me.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -158,8 +158,6 @@ void Game::setGameState(GameState_t newState)
 		}
 
 		case GAME_STATE_SHUTDOWN: {
-			g_globalEvents->execute(GLOBALEVENT_SHUTDOWN);
-
 			//kick all players that are still online
 			auto it = players.begin();
 			while (it != players.end()) {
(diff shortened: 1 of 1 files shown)


marksamman:
While it does make sense to move onShutdown to Game::shutdown(), I prefer to leave it as it is for now because shutdown() isn't being called when you close the server the same way on a non-Windows OS. Once we've added signal handling, I'll consider merging this.

EvilHero90:
I've backtraced what you meant and you are absolutely right, do you either intend to move the entire thing to game::shutdown() and make use of #ifdef __linux__ and #ifdef _WIN32 or what would be your solution in this case?

marksamman:
The solution right now is to leave it as-is to have consistent behavior across different operating systems; the solution once we have implemented signal handling is just to do what you did in this PR.

ranisalt:
@marksamman as we have signal handling now maybe this PR could be rebased and merged?

wgrriffel:
@ranisalt there is only POSIX signal handlers, still left Windows [console handler](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686016(v=vs.85).aspx), reported at #1853, and maybe #2088, which is not confirmed, maybe we should cover these cases also before merging this

PR #1284, opened 2015-04-10
Refactor network stack.

djarek:
Refactor OutputMessagePool. Remove ref counter from Connection as it's really unnecessary since, we're already doing the same with the shared_ptr created by inheriting from enable_shared_from_this. Move socket cleanup code calling to the Connection destructor. Remove getAvailableMessageCount() and getAutoMessageCount() from OutputMessagePool as these were not thread-safe and unused anyway. Protocol objects are now managed by shared_ptrs. Rename member variables. Made some small code improvements for better readability. I did some performance testing (with a Release build) and time spent in getOutputMessage() dropped by 40% on average (bear in mind that it's under light load, my version of that function scales better as contention on new/delete increases (the old version would always allocate memory for the shared_ptr control block.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -83,7 +83,7 @@ void Connection::close()
 	if (!m_pendingWrite) {
 		closeSocket();
 	} else {
-		//will be closed by onWriteOperation
+		//will be closed by the destructor
 	}
 }
 
@@ -109,7 +109,6 @@ void Connection::closeSocket()
 
 Connection::~Connection()
 {
-	assert(m_connectionState != CONNECTION_STATE_OPEN); //If this fires, no one closed this Connection and we're being deleted - something went wrong
 	closeSocket();
 }
 
@@ -323,10 +322,9 @@ void Connection::onWriteOperation(OutputMessage_ptr msg, const boost::system::er
 		close();
 	}
 
-	//If someone requested a connection close, we allow all pending outputMessages to be sent,
+	//If someone requested a connection close, we allow all pending OutputMessages to be sent,
 	//unless an error occured during the send operation
 	if ((m_connectionState != CONNECTION_STATE_OPEN && messageQueue.empty()) || error) {
-		closeSocket();
 		return;
 	}
 
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -52,7 +52,6 @@ void ConnectionManager::closeAll()
 	std::lock_guard<std::mutex> lockClass(m_connectionManagerLock);
 
 	for (const auto& connection : m_connections) {
-		assert(connection.use_count() == 1);
 		try {
 			boost::system::error_code error;
 			connection->m_socket->shutdown(boost::asio::ip::tcp::socket::shutdown_both, error);
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -264,7 +264,7 @@ void Connection::send(OutputMessage_ptr msg)
 	if (!m_pendingWrite) {
 		internalSend(msg);
 	} else {
-		messageQueue.push_back(msg);
+		messageQueue.emplace_back(msg);
 	}
 }
 
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -264,7 +264,7 @@ void Connection::send(OutputMessage_ptr msg)
 	if (!m_pendingWrite) {
 		internalSend(msg);
 	} else {
-		messageQueue.push(msg);
+		messageQueue.push_back(msg);
 	}
 }
 
@@ -302,15 +302,6 @@ uint32_t Connection::getIP()
 	return htonl(endpoint.address().to_v4().to_ulong());
 }
 
-void Connection::clearMessageQueue()
-{
-	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
-	while (!messageQueue.empty()) {
-		messageQueue.pop();
-	}
-}
-
-
 void Connection::onWriteOperation(OutputMessage_ptr msg, const boost::system::error_code& error)
 {
 	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
@@ -318,7 +309,7 @@ void Connection::onWriteOperation(OutputMessage_ptr msg, const boost::system::er
 	m_pendingWrite = false;
 
 	if (error) {
-		clearMessageQueue();
+		messageQueue.clear();
 		close();
 	}
 
@@ -330,7 +321,7 @@ void Connection::onWriteOperation(OutputMessage_ptr msg, const boost::system::er
 
 	if (!messageQueue.empty()) {
 		internalSend(std::move(messageQueue.front()));
-		messageQueue.pop();
+		messageQueue.pop_front();
 	}
 }
 
@@ -354,7 +345,7 @@ void Connection::handleWriteTimeout(ConnectionWeak_ptr connectionWeak, const boo
 	}
 
 	if (auto connection = connectionWeak.lock()) {
-		connection->clearMessageQueue();
+		connection->messageQueue.clear();
 		connection->close();
 	}
 }
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -52,6 +52,7 @@ void ConnectionManager::closeAll()
 	std::lock_guard<std::mutex> lockClass(m_connectionManagerLock);
 
 	for (const auto& connection : m_connections) {
+		assert(connection.use_count() == 1);
 		try {
 			boost::system::error_code error;
 			connection->m_socket->shutdown(boost::asio::ip::tcp::socket::shutdown_both, error);
@@ -67,29 +68,19 @@ void ConnectionManager::closeAll()
 void Connection::close()
 {
 	//any thread
-	auto this_shared = shared_from_this();
-	ConnectionManager::getInstance()->releaseConnection(this_shared);
-	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
+	ConnectionManager::getInstance()->releaseConnection(shared_from_this());
 
+	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
 	if (m_connectionState != CONNECTION_STATE_OPEN) {
 		return;
 	}
-
 	m_connectionState = CONNECTION_STATE_CLOSED;
-
-	g_dispatcher.addTask(
-	    createTask(std::bind(&Connection::closeConnectionTask, std::move(this_shared))));
-}
-
-void Connection::closeConnectionTask()
-{
-	//dispatcher thread
-	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
-
+	
 	if (m_protocol) {
-		m_protocol->release();
+		g_dispatcher.addTask(
+			createTask(std::bind(&Protocol::release, m_protocol)));
 	}
-
+	
 	if (!m_pendingWrite || m_writeError) {
 		closeSocket();
 	} else {
@@ -120,6 +111,7 @@ void Connection::closeSocket()
 
 Connection::~Connection()
 {
+	assert(m_connectionState != CONNECTION_STATE_OPEN); //If this fires, no one closed this Connection and we're being deleted - something went wrong
 	closeSocket();
 }
 
@@ -272,19 +264,16 @@ bool Connection::send(const OutputMessage_ptr& msg)
 {
 	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
 	if (m_connectionState != CONNECTION_STATE_OPEN || m_writeError) {
-		if (m_protocol) {
-			m_protocol->clearOutputBuffer(msg);
-		}
-		return false;
+		m_protocol->clearOutputBuffer(msg);
+		return true;
 	}
 
 	if (!m_pendingWrite) {
 		m_protocol->onSendMessage(msg);
 		internalSend(msg);
 	} else {
-		// FIXME: This results in packets being sent in wrong order, all queued
-		// packets to the same connection should also be moved to auto send.
-		OutputMessagePool::getInstance()->addToAutoSend(msg);
+		//There's a pending write operation - let the caller now
+		return false;
 	}
 
 	return true;
@@ -333,7 +322,6 @@ void Connection::onWriteOperation(OutputMessage_ptr msg, const boost::system::er
 	}
 
 	if (m_connectionState != CONNECTION_STATE_OPEN || m_writeError) {
-		closeSocket();
 		close();
 		return;
 	}
@@ -363,7 +351,6 @@ void Connection::onReadTimeout()
 	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
 
 	if (m_pendingRead || m_readError) {
-		closeSocket();
 		close();
 	}
 }
@@ -373,7 +360,6 @@ void Connection::onWriteTimeout()
 	std::lock_guard<std::recursive_mutex> lockClass(m_connectionLock);
 
 	if (m_pendingWrite || m_writeError) {
-		closeSocket();
 		close();
 	}
 }
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -80,17 +80,16 @@ void Connection::close()
 			createTask(std::bind(&Protocol::release, m_protocol)));
 	}
 	
-	if (!m_pendingWrite || m_writeError) {
+	if (!m_pendingWrite) {
 		closeSocket();
 	} else {
-		//will be closed by onWriteOperation/handleWriteTimeout/handleReadTimeout instead
+		//will be closed by onWriteOperation
 	}
 }
 
 void Connection::closeSocket()
 {
 	if (m_socket->is_open()) {
-		m_pendingRead = false;
 		m_pendingWrite = false;
 
 		try {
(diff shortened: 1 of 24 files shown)


HeavenIsLost:
Bro it is related to " (though bear in mind that some of these are actually caused by deficiencies of code outside the casting system)."?

40% is good :)

djarek:
@HeavenIsLost 
Yes. Unfortunately I haven't found a "clean" solution to the Travis CI build issue (boost::lockfree is available in boost v1.53+ and Travis only supports Ubuntu 12.04 ATM. I'll try to fix it soon.

HeavenIsLost:
Did you merge this with your casting system to see if its solve the problems?

djarek:
@HeavenIsLost 
I didn't do it yet, but I believe that a lot of the casting system's bugs should be resolved with these changes (though I don't claim that all of them should).

I've finally removed the dependancy on boost::lockfree by implementing a simple lock-free and bounded stack. I'll move it (and possible the OutputMessageAllocator) to a separate file.

djarek:
Fixed packet reordering issue.
@marksamman 
Any thoughts on this PR?

marksamman:
I'll have a look at this after the 1.1 release, but as for packet reordering, it doesn't look like it has been properly fixed. If message A is in the queue and was added before message B, message B could still be sent before message A if internalSend fails for message A but succeeds for message B.

djarek:
You're right. I think the best solution would be to have each Connection object use a FIFO queue and perform a send if a write operation is successful and there are messages still in the queue. So Connection::send() would either perform the send if there are no pending write operations or just push the message onto the internal queue.

djarek:
@marksamman

Just to be clear - we only care about keeping at all times the relative ordering of autosend OutputMessages (which in fact are exclusively Protocol output buffers), right?

marksamman:
For the same connection, yes. What else were you thinking of?

djarek:
For a moment I thought we would have to guarantee the ordering of both autosend and explicit-send messages, but then I realized it really made no sense because they're only used for ProtocolGame and explicit-send ones are mostly used for disconnecting with an error, so it doesn't matter.

djarek:
The lock-free stack adds considerable complexity to the codebase. I'm trying to decide whether the complexity is actually worth it. I did a test to get a rough estimate (Release build, gcc 4.9.2, Phenom II X4 3,2 GHz), the results are (average time spent in the getOutputMessage() call, i'm providing min and max value for a set of a few runs):
* lock-free stack: 915-1000 ns
* a custom spinlock + vector: 960-1100 ns
* std::mutex + vector: 1000-1147 ns
The timer delay was approximately 100 ns.
Obviously the numbers themselves are quite meaningless as they're totallly dependant on the hardware. They're only supposed to give us a rough idea about the gains/losses from each synchronization primitive. One thing that has to be considered is that the stack scales better as contention on the message pool increases. The custom spinlock is a bad idea in the case of this project - on a system on which we oversubscribe threads it might actually perform A LOT worse than a std::mutex (as the thread might spin till the end of its time quantum because the thread owning the lock is sleeping). One might think "why should we care about ~100 ns difference?", but when we consider this function might potentially be called thousands or even tens of thousands of times per second it starts adding up and potentially allowing a few more dispatcher cycles to run in that time (i.e. improving dispatcher throughput).

Another thing: would it be OK to add a new scheduler function (perhaps available only in the task dispatcher code) that would allow scheduling a task with time lower than 50 ms. Currently sendAll() is called on every dispatcher cycle which is quite inefficient (we get current time on every cycle). Instead, we could just schedule a sendAll() task every 10 ms and achieve the same effect with less work done by the dispatcher thread.

marksamman:
You should use boost lock-free instead of a custom implementation. boost lock-free requires boost 1.53, Debian 8 has 1.55, and we can probably just use some PPA to get it to work in Travis.

Yes, it would be OK to add a new scheduler function.

sendAll() is indeed excessively called and was one of the obstacles in scaling ShadowCores. Breaking when reaching messages beyond frame time helped a lot, but I'm sure we don't need to call it as often as we do right now. Every 10ms sounds reasonable.

djarek:
So I presume that we're actually not going to try and drop the boost dependancies?

marksamman:
We'll have to deal with that dilemma if we make that decision, until then, using boost lock-free is the obvious choice here.

gugahoa:
While we're at that topic, why are we moving away from boost dependence?

marksamman:
We're not; we're just reconsidering whether it's worth keeping it because we're only using a small subset of boost. asio has a non-boost standalone version, and without asio we're only depending on lexical_cast and reverse iterator adaptor, both which are easily replaceable. If we're going to use boost lock-free, that's a reason to keep boost, at least for now.

djarek:
I believe that this set of changes is complete.
@marksamman 
When you have the time, could you review this PR?

src/player.cpp
```diff
@@ -771,7 +767,7 @@ void Player::addStorageValue(const uint32_t key, const int32_t value, const bool
 		storageMap[key] = value;
 
 		if (!isLogin) {
-			int64_t currentFrameTime = OutputMessagePool::getInstance()->getFrameTime();
+			int64_t currentFrameTime = OTSYS_TIME();
```
marksamman: This is intended to make sure "Your questlog has been updated." is only sent once per dispatcher task, checking OTSYS_TIME() is not good enough as it's very likely to have a different value next time it's called in the same dispatcher task.

src/connection.h
```diff
@@ -142,22 +120,20 @@ class Connection : public std::enable_shared_from_this<Connection>
 
 		std::recursive_mutex m_connectionLock;
 
+		std::queue<OutputMessage_ptr> messageQueue;
```
marksamman: I would prefer std::list here, because queue defaults to a deque as its underlying container and we don't need random access in the message queue.

src/connection.cpp
```diff
@@ -375,112 +302,59 @@ uint32_t Connection::getIP() const
...
-		// Do NOT call close() from here
-	} else {
-		/**
-		 * error == boost::asio::error::eof:
-		 *  No more to read
-		 * error == boost::asio::error::connection_reset || error == boost::asio::error::connection_aborted:
-		 *  Connection closed remotely
-		 */
-		close();
+	while (!messageQueue.empty()) {
```
marksamman: If you change messageQueue to std::list you can use clear() here.

src/player.cpp
djarek: So I will need to add a dispatcher cycle counter, right? Using current time for the counter seems like an unnecessary overhead anyway.

src/connection.h
djarek: I believe that a deque has slightly better performance (didn't test that though) than a list and causes less memory overhead.

dalkon:
Why in the world would you change something that works with raw pointers to smart pointers, you abuse the smart pointers way too much in such a way where it decreases performance (It would be worse anyways since you move away from raw pointers).

You probably have not even tested your code on Windows as it would not work at all due to constexpr (but muh C++11).

djarek:
The previous Protocol reference counting scheme contained a race condition that could lead to a leak or use-after-delete of Protocol objects. I admit, shared_ptr has a considerable overhead compared to a raw pointer, however, unless we slightly redesign the relationship between Connection and Protocol objects, there is no better alternative than using a shared_ptr (except, maybe writing our own, thread-safe, ref. counted pointer, but would we really gain that much?).

src/connection.h
marksamman: It does in the cases where it doesn't have to grow the internal vectors, so it's amortized constant time insertions, compared to constant time with list. If you prefer a deque, you may switch to that, it has the clear() method as well.

src/player.cpp
marksamman: Sure, that sounds like a good option.

djarek:
So, should I leave the management of Protocol objects to shared_ptrs or try and come up with an another solution?

djarek:
@marksamman 
Should I try finding an alternate solution to managing the Protocol object lifetimes in a thread-safe manner or should I leave it as it is now?

marksamman:
Why? Are there any issues with the current code, or are you thinking of a better solution?

djarek:
To be clear - I meant the solution currently in the PR, that is using shared_ptr to manage the life time of Protocol objects. The only issue with that currently is that (as @dalkon has mentioned) there's a minor (really minor) performance hit compared to some other, more optimized solutions.

djarek:
@marksamman 
I believe this PR is complete, could you review it when you have the time?

src/connection.h
```diff
@@ -142,22 +120,20 @@ class Connection : public std::enable_shared_from_this<Connection>
...
 		boost::asio::io_service& m_io_service;
 
 		time_t m_timeConnected;
 		uint32_t m_packetsSent;
-		uint32_t m_refCount;
-		int32_t m_pendingWrite;
-		int32_t m_pendingRead;
-		ConnectionState_t m_connectionState;
 
+		bool m_pendingWrite;
```
marksamman: Are you sure that this can be a boolean? Can there not be more than one pending write at a time?

djarek: Perhaps the name is a little misleading - this is meant to indicate that the ASIO service thread has been request to perform a write for this Connection's socket. According to boost docs, only one write and one read operation can be requested for any given socket at any time.

src/outputmessage.cpp
```diff
@@ -21,174 +21,59 @@
...
-			msg->getProtocol()->onSendMessage(msg);
-		}
-	}
+	public:
+		typedef OutputMessage value_type;
+		template<typename U>
+		struct rebind {typedef LockfreePoolingAllocator<U, OUTPUTMESSAGE_FREE_LIST_CAPACITY> other;};
+};
+
+void OutputMessagePool::scheduleSendAll() {
```
marksamman:
```
void OutputMessagePool::scheduleSendAll()
{
```

src/connection.cpp
```diff
@@ -64,57 +64,37 @@ void ConnectionManager::closeAll()
 
 // Connection
 
-void Connection::close()
+void Connection::close(bool isForced)
```
marksamman: Rename the parameter to force.

src/protocollogin.cpp
```diff
@@ -145,40 +140,43 @@ void ProtocolLogin::onRecvFirstMessage(NetworkMessage& msg)
...
 		return;
 	}
 
 	BanInfo banInfo;
-	if (IOBan::isIpBanned(getConnection()->getIP(), banInfo)) {
+	auto connection = getConnection();
+	if (!connection) {
+		return;
+	}
+	if (IOBan::isIpBanned(connection->getIP(), banInfo)) {
```
marksamman: Empty line between the if statements.

marksamman:
LGTM

/cc @kornholi

src/tasks.h
```diff
@@ -81,6 +81,9 @@ class Dispatcher
 		void shutdown();
 		void join();
 
+		uint64_t getDispatcherCycle() {
```
marksamman: This function can be const.

marksamman:
I was testing this and got a segfault when relogging a couple of times.
```
#0  0x0000000000890e04 in Game::checkCreatures(unsigned long) ()
(gdb) where
#0  0x0000000000890e04 in Game::checkCreatures(unsigned long) ()
#1  0x00000000008c026a in void std::_Mem_fn<void (Game::*)(unsigned long)>::operator()<unsigned long&, void>(Game*, unsigned long&) const ()
#2  0x00000000008bb489 in void std::_Bind<std::_Mem_fn<void (Game::*)(unsigned long)> (Game*, unsigned long)>::__call<void, , 0ul, 1ul>(std::tuple<>&&, std::_Index_tuple<0ul, 1ul>) ()
#3  0x00000000008b39ea in void std::_Bind<std::_Mem_fn<void (Game::*)(unsigned long)> (Game*, unsigned long)>::operator()<, void>() ()
#4  0x00000000008ac817 in std::_Function_handler<void (), std::_Bind<std::_Mem_fn<void (Game::*)(unsigned long)> (Game*, unsigned long)> >::_M_invoke(std::_Any_data const&) ()
#5  0x0000000000a916da in std::function<void ()>::operator()() const ()
#6  0x0000000000a9157c in Task::operator()() ()
#7  0x0000000000a90e26 in Dispatcher::dispatcherThread() ()
#8  0x0000000000a92ce9 in void std::_Mem_fn<void (Dispatcher::*)()>::operator()<, void>(Dispatcher*) const ()
#9  0x0000000000a92c39 in void std::_Bind_simple<std::_Mem_fn<void (Dispatcher::*)()> (Dispatcher*)>::_M_invoke<0ul>(std::_Index_tuple<0ul>) ()
#10 0x0000000000a92b2f in std::_Bind_simple<std::_Mem_fn<void (Dispatcher::*)()> (Dispatcher*)>::operator()() ()
#11 0x0000000000a92aac in std::thread::_Impl<std::_Bind_simple<std::_Mem_fn<void (Dispatcher::*)()> (Dispatcher*)> >::_M_run() ()
#12 0x00007f7750fea970 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#13 0x00007f77512470a4 in start_thread (arg=0x7f774f9e3700) at pthread_create.c:309
#14 0x00007f775075a04d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
(gdb)
```

djarek:
Seems like an issue with ProtocolGame::release(). I'll try to investigate later.

djarek:
Fixed.

marksamman:
Replace kick isn't working. Steps to reproduce:
1) Open client 1, login on character 1
2) Open client 2, login on character 1
3) Client 1 gets disconnected (correct), but client 2 is stuck connecting to the game world.

djarek:
Found the issue. If another client is kicked during the login process, the ProtocolGame object associated with the connecting client would not be added to the auto send list becuase of a return statement in that branch, causing ProtocolGame::login() to exit prematurely.

marksamman:
There seems to be another issue with replace kick where it will say "You are already logged in." even though there is no connected client. This is harder to reproduce, but I think these steps are fairly accurate:
1) Open client 1, login on character 1
2) Open client 2, login on character 1
3) On client 1, you should have been kicked back to character list, connect and cancel while it's connecting to the game world, client 2 should have been kicked
4) If you try to login on character 1 from either client 1 or 2 now you will get "You are already logged in.".

I think _player->client remains set to client 1 even though client 1 has already disconnected, so _player->client isn't null in ProtocolGame::connect when you try to login again.

Luckey1729:
Would this be an appropriate place to suggest that replace kick should count as a true "login" (executing onLogin creature event / showing login animation) - I believe this is how Global works now.

marksamman:
I don't like the idea of executing onLogin on replace kick. If there should be an event that's called on login and on replace kick it should be called onConnect, and it should not be implemented in this PR.

djarek:
I've found the problem with Replace kick. It occurs when ProtocolGame::connect is called after the Connection object has been destroyed.

djarek:
Fixed.

marksamman:
Now it's not working to login at all.
```
src/outputmessage.h:72: void OutputMessage::add_header(T) [with T = short unsigned int]:
Assertion `outputBufferStart >= sizeof(T)' failed.
```

djarek:
Reverted last commit, should work now.

src/protocollogin.cpp
```diff
@@ -55,60 +54,58 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
...
+	if (!motd.empty()) {
+		//Add MOTD
+		output->addByte(0x14);
 
-		const std::string& motd = g_config.getString(ConfigManager::MOTD);
-		if (!motd.empty()) {
-			//Add MOTD
-			output->addByte(0x14);
+		std::ostringstream ss;
+		ss << g_game.getMotdNum() << "\n" << g_config.getString(ConfigManager::MOTD);
```
marksamman: Use the motd variable.

src/scheduler.h
```diff
@@ -50,8 +50,13 @@ class SchedulerTask : public Task
 		uint32_t eventId;
 
 		friend SchedulerTask* createSchedulerTask(uint32_t, const std::function<void (void)>&);
+		friend SchedulerTask* createLowDelayTask(uint16_t delay, std::function<void (void)> f);
 };
 
+inline SchedulerTask* createLowDelayTask(uint16_t delay, std::function<void (void)> f) {
```
marksamman: Move the opening curly bracket down to a new line.

src/protocolgame.cpp
```diff
@@ -228,6 +198,11 @@ void ProtocolGame::connect(uint32_t playerId, OperatingSystem_t operatingSystem)
 		disconnectClient("You are already logged in.");
 		return;
 	}
+	if (isConnectionExpired()) {
```
marksamman: Insert an empty line above the if-statement.

src/outputmessage.cpp
```diff
@@ -21,174 +21,61 @@
...
-	    createTask(std::bind(&OutputMessagePool::internalReleaseMessage, this, msg)));
-}
-
-void OutputMessagePool::internalReleaseMessage(OutputMessage* msg)
-{
-	if (msg->getProtocol()) {
-		msg->getProtocol()->unRef();
-	} else {
-		std::cout << "No protocol found." << std::endl;
+	if (!bufferedProtocols.empty()) {
```
marksamman: Insert an empty line above the if-statement.

kornholi:
:+1: good enough for now

marksamman:
Thanks!

WibbenZ:
Now I get why ppl want things squashed xD

PR #1293, opened 2015-04-16
Update to game protocol 10.77

TechrlzGit:
Simple edit in definitions to update game protocol.


WibbenZ:
I suppose this has not been merged due to the otb file? @marksamman

Luckey1729:
1.1 is set to release on May 2, 2015 according to the Milestone page. I believe Mark plans to upgrade the protocol to the newest before the release of 1.1 and then begin on 1.2. That would be my assumption as to why this has not yet been merged.

PR #1296, opened 2015-04-17
[Beta] Monster Types handled in lua

EvilHero90:
With this I have the intention to drop the use of xml for monster files.
It currently supports all functions except for attacks / defenses / loot
This would be a quick example on how to create a new Monster Type in lua
<code>local test = NewMonsterType("Test")
test:setNameDescription("a test")
test:setExperience(5000)
test:setOutfit({lookType = 698, lookMount = 368, lookAddons = 3})
test:setMaxHealth(2)
test:setHealth(2)
test:setRunHealth(1)
test:setMaxSummons(3)
test:setRace("fire")
test:setManaCost(200)
test:setBaseSpeed(250)
test:setCorpseId(1987)
test:setIsAttackable(true)
test:setIsHealthHidden(false)
test:isConvinceable(true)
test:isIllusionable(false)
test:isHostile(false)
test:isPushable(true)
test:setCanPushItems(true)
test:setCanPushCreatures(true)
test:addVoice("works", 5000, 10000, false)
test:addVoice("I can talk", 5000, 10000, false)
test:addVoice("testzor", 5000, 10000, false)
test:setSkull("black")
test:addSummon("Demon", 5000, 10000)</code>

I've also made a clone function along with it which can only be used once the entire server is load (or it'll crash the server)
it'll mimic another monster types behaviour for the monster type we set it.
ex:
<code>local test = NewMonsterType("Test")
test:clone("Demon", "all") -- all means it'll clone everything except name and name description of the monster </code> ("attacks","defenses","voices","loot","summons","all" do exist)
now our monster "Test" will behave and look like a "Demon"

The auto lua file parser in global.lua will automaticly load the monsters in the monster/ folder (and subfolders) without that we need to register them in an extra file first.
So we can drag and drop monsters into the folder and they'll be automaticly load on server start or with a command (needs to be created / edited)

I'm not sure if I'm still missing something except for the stuff which is already stated to be in todo phase.
if someone feels like there is missing something, then just let me know.


Crypton33:
Very nice, but I don't like how the monster files are looking.

Here's my proposition how it could look:
**Monster file:**
```lua
local monster = NewMonsterType('Test')
monster.description = 'a Test'
monster.experience = 5000
monster.outfit = {
	lookType = 698,
	lookMount = 368,
	lookAddons = 3
}

monster.health = 5000
monster.maxHealth = monster.health
monster.race = 'fire'

monster.flags = {
	isAttackable = true,
	isHealthHidden = false,
	isConvinceable = true
}

monster.voices = {
	{'Works', 500, 1000, false},
	{'I can talk', 500, 1000, false},
	{'Test', 500, 1000, false}
}

monster:register()
```

**Lua function for registering the monster:**
```lua
function MonsterType.register(self)
	if self.description then
		self:setNameDescription(self.description)
	end
	
	if self.outfit then
		self:setOutfit(self.outfit)
	end
	
	-- etc...
end
```

It'll be similar to [otserv revscriptsys spell files](https://raw.githubusercontent.com/opentibia/server/master/data/scripts/spells/conjure/conjure_bolt.lua) and will be much easier is use.

What do you think about this?

EvilHero90:
I had a similiar idea in mind with tables aswell but your idea is far better then the one I had, I'll add it the way you suggested it.

src/monsters.cpp
```diff
@@ -186,6 +186,121 @@ void MonsterType::createLoot(Container* corpse)
 	corpse->startDecaying();
 }
 
+void MonsterType::clone(MonsterType* root, MonsterType* child, const std::string type) {
+	if (type == "attacks") {
+		if (!child->spellAttackList.empty()) {
+			for (spellBlock_t attacks : child->spellAttackList) {
+				root->spellAttackList.push_back(attacks);
```
djarek: This is bad. You're copying owning raw pointers into another container which could result in the spell objects being deleted multiple times (e.g. on a reload of monsters or spells).

src/monsters.cpp
```diff
@@ -186,6 +186,121 @@ void MonsterType::createLoot(Container* corpse)
...
+			}
+		}
+	} else if (type == "defenses") {
+		if (!child->spellDefenseList.empty()) {
+			for (spellBlock_t defenses : child->spellDefenseList) {
+				root->spellDefenseList.push_back(defenses);
```
djarek: This has the same problem as above.

EvilHero90: wasn't aware of that fact, mind telling me how to do it properly or should I try to avoid cloning them overall?

djarek: You would need to create an exact copy of the spell object that's pointed to by the spell ptr in the spell block (or introduce a ref counting scheme to Spell objects and use a custom smart pointer). Also I think you should look at this PR #1251. I've made a bunch of improvements there you might also look at. If you decide to introduce ref counting to spells, please let me know, I'll help you do it properly.

EvilHero90: would you mind having a chat on skype with me I'm having some questions regarding that.
skype: evilhero19

esfomeado:
It would be better the way the 33 @ Crypton suggested

marksamman:
We're not going to merge something that requires LuaJIT or uses platform specific code. We're also not going to remove XML from monsters anytime soon. If this was your goal, I suggest starting a fork.

Diogo-Barreiros:
Just saying, as just another OT enthusiast, I also see no good reason to remove XML.

totolol123:
Add this load directory function in C++, something like this:
https://github.com/opentibia/server/blob/d012c7297613346d4a22ed87d404a25d6f10970e/src/lua_manager.cpp#L898
Also this could be added as a option for creating monster in lua with a few custom functions.

djarek:
@Diogo-Barreiros
IMHO JSON is a much better data-exchange format than XML because it's more efficient, terse and more esthetically pleasing :D. Anyway, dropping XML can't be done any time soon for compatibility reasons and for quite some time we would actually have to support both systems.

EvilHero90:
Well my final intention was to drop xml completly, however keeping it for now to support both, as I would have to make a parser to convert xml files into lua files first.
Well I don't really see any reason to stick with something old which hinders us more then it provides.
See it like this, when servers got changed from xml to sql, all complained but it was obviously better and people had to adept to it.

I don't really see any reason to advance to an oop style and then keep xml which limits us in that case so much that we basicly cannot take full advantage of all the things which we actually could do.

About LuaJit do you have any intentions of removing it or why don't you want to make use of it?
I could write that function in a different way in c++ aswell but as jit.os was working just fine I thought it would be a waste of space.

Diogo-Barreiros:
@djarek 
I agree with JSON, or even YAML. But still think Lua is not a markup language, it's a programming one.

ranisalt:
It looks a lot like the way PyOT handles stuff

marksamman:
XML vs. SQL as a database and XML vs. Lua for configuration files are completely different matters, so I don't see why you're comparing that. XML as a database was also removed by me, so I think I would have seen what you're saying if it made any sense. We've had config.lua as a Lua file for ages, yet I don't see anyone do anything beyond what they already could have done if it was config.xml instead.

I'm not saying allowing monsters to be registered through Lua is a bad thing, I'm saying that we're not going to remove the XML files.

We have no intention of removing LuaJIT either, we're just not going to force people to use it over Lua, which you're essentially doing by using jit.os. My problem isn't solely with how you're identifying the OS, it's also that you're adding platform specific code which is bad for portability. Another issue is that you're using commands for directory listing, which has significant overhead. std::filesystem (or boost::filesystem v3 if we can't use std::filesystem yet) should be preferred.

EvilHero90:
ok my example with the comparison between xml and sql was probably not the best I agree but my point was that we should move on and finaly do a modular system, so we can do long quests easily in 1 file without registering it in a weird way in 10 different xml files instead of just doing it in the lua file, where it is supposed to be imo.

I'm currently removing the dependencies for luaJit and the os specific code, I'll make use of boost::filesystem as you suggested.

gugahoa:
I think it would be great it there was a support for both. So it would allow for runtime creation of custom monsters.

EvilHero90:
seems like travis is not up to date, got someone testing it on debian mint 17.1 and ubuntu 14.1 both confirmed it compiled just fine.

data/monster/lua/test.lua
```diff
@@ -0,0 +1,50 @@
+-- a testing layout
+local test = NewMonsterType('Test')
```
Crypton33: Keep it simple, insetad NewMonsterType you can do something like this:
```lua
local test = MonsterType('Test')
```

Nu77: @Crypton33 are right, it should create a new MonsterType if "Test" doesn't exist

gugahoa: Evil, to compile here I had to change the cmake to compile. Travis is
alright.

EvilHero90:
the problem with MonsterType("Test") creating a new monster type is that you could not check if a certain monstertype does not exist or not, if we create with a not existing.
so I thought it would be smarter to seperate it.

Nu77:
@EvilHero90 I'm thinking here and don't is better creating a separate commit for Loot class and uses this for implement Loot handling on Lua ?

EvilHero90:
loot can be handled through lua that's already implemented.
registerClass("Loot", "", LuaScriptInterface::luaCreateLoot);
registerMethod("Loot", "setId", LuaScriptInterface::luaLootSetId);
registerMethod("Loot", "setMaxCount", LuaScriptInterface::luaLootSetMaxCount);
registerMethod("Loot", "setSubType", LuaScriptInterface::luaLootSetSubType);
registerMethod("Loot", "setChance", LuaScriptInterface::luaLootSetChance);
registerMethod("Loot", "setActionId", LuaScriptInterface::luaLootSetActionId);
registerMethod("Loot", "setDescription", LuaScriptInterface::luaLootSetDescription);
registerMethod("Loot", "addChildLoot", LuaScriptInterface::luaLootAddChildLoot);

ex:
local test = Loot()
test:setId("crystal coin")
test:setChance(10000)
local monster = MonsterType("Demon")
monster:addLoot(test)

In test.lua it's already handled by the table api that's why you basicly cannot see the functions directly.

WibbenZ:
IMO this would be better then having the XML files.
But before this gets added we either need a tool that translates the XML code to lua or allow both types(atleast at the beginning.

EvilHero90:
I havn't touched xml handling yet so it's still there and I have no intention on removing it unless everything else is done and mark would accept, aslong as that's not the case, xml handling will still be there.
once I have written all the necessary functions for MonsterType handling I'll start writing the xml -> lua converter.

Nu77:
I mean to move the loot creation from core to Lua, also should be possible to get a loot like this:
```lua
local test = Loot("Demon")
```

Crypton33:
XML -> LUA converter won't be enough. We will need to add some new code to RME to be able to import monsters from lua files.

EvilHero90:
well for rme I could make a code which will add all the new monsters into monsters.xml so it works for now.

EvilHero90:
Basics are all done now, I can fully re create all monsters with the current standing.
I'd be glad if someone could tell me what I have to change in order for travis to pass the build, I'm not familiar with cmake and that stuff.
I would like to have a review @marksamman if the way I approached all the stuff is done well or if there needs to be some adjustment or overall changes.

gugahoa:
find_package(Boost 1.42.0 COMPONENTS system REQUIRED)

To

find_package(Boost 1.42.0 COMPONENTS system filesystem REQUIRED)

src/luascript.cpp
```diff
@@ -4256,6 +4338,33 @@ int LuaScriptInterface::luaSendGuildChannelMessage(lua_State* L)
 	return 1;
 }
 
+
+int LuaScriptInterface::luaLoadDirectory(lua_State* L)
+{
+	const boost::filesystem::path& dir = boost::filesystem::current_path().string() + getString(L, 1);
```
djarek: Reference to a temporary object (remove the & to fix). Use append() on the path object returned from the current_path() function.
```
const auto dir = boost::filesystem::current_path().append(getString(L, 1));
```

src/luascript.cpp
```diff
@@ -4256,6 +4338,33 @@ int LuaScriptInterface::luaSendGuildChannelMessage(lua_State* L)
...
+	if (!boost::filesystem::exists(dir) || !boost::filesystem::is_directory(dir)) {
+		pushBoolean(L, false);
+		return 1;
+	}
+
+	boost::filesystem::recursive_directory_iterator it(dir);
+	boost::filesystem::recursive_directory_iterator endit;
+
+	for (; it != endit; ++it) {
+		if (is_regular_file(*it) && it->path().extension() == ".lua") {
```
djarek: Either specify namespaces or rely on ADL. Don't mix them both because it's confusing.

src/monsters.cpp
```diff
@@ -186,6 +239,175 @@ void MonsterType::createLoot(Container* corpse)
...
+				loadChildLoot(l, lootblock);
+			}
+		}
+		monsterType->lootItems.push_back(lootblock);
+	} else {
+		monsterType->lootItems.push_back(lootblock);
+	}
+}
+
+void MonsterType::loadChildLoot(Loot* loot, LootBlock& parent)
```
djarek: This function has a lot of code common with loadLoot(). Consider using recursion to do loading with just one function.

src/luascript.cpp
```diff
@@ -11840,398 +11974,1016 @@ int LuaScriptInterface::luaMonsterTypeCreate(lua_State* L)
...
+
+	if (clone) {
+		pushUserdata<MonsterType>(L, clone);
+		setMetatable(L, -1, "MonsterType");
+	} else {
+		lua_pushnil(L);
+	}
+
+	if (monsterType) {
+		monsterType->clone(monsterType, clone, type);
```
djarek: This causes undefined behavior if clone is nullptr (e.g. if the "clone" is not found).

src/monsters.cpp
```diff
@@ -186,6 +239,175 @@ void MonsterType::createLoot(Container* corpse)
 	corpse->startDecaying();
 }
 
+void MonsterType::clone(MonsterType* root, MonsterType* child, const std::string& type) {
+	if (type == "loot") {
+		if (!child->lootItems.empty()) {
```
djarek: Use std::list::assign(Iterator begin, Iterator end) to explicitly copy the containers.

src/luascript.cpp
EvilHero90: glad you told me, havn't taken a second look on that, will fix it.

src/monsters.cpp
```diff
@@ -186,6 +239,175 @@ void MonsterType::createLoot(Container* corpse)
 	corpse->startDecaying();
 }
 
+void MonsterType::clone(MonsterType* root, MonsterType* child, const std::string& type) {
```
djarek: Consider changing the argument names to source and target. It's not clear with root and child which one is which.

src/monsters.cpp
EvilHero90: I already had that in mind, I just wanted to do that once everything was functional but thanks for reminding me.

src/luascript.cpp
EvilHero90: I'll add namespace, had issues at first because I was using it wrong but I'll re work it.

src/monsters.cpp
```diff
@@ -186,6 +239,175 @@ void MonsterType::createLoot(Container* corpse)
...
+		root->staticAttackChance = child->staticAttackChance;
+		root->maxSummons = child->maxSummons;
+		root->targetDistance = child->targetDistance;
+		root->runAwayHealth = child->runAwayHealth;
+		root->pushable = child->pushable;
+		root->baseSpeed = child->baseSpeed;
+		root->health = child->health;
+		root->healthMax = child->healthMax;
+
+		root->outfit.lookHead = child->outfit.lookHead;
```
djarek:
```
root->outfit = child->outfit;
```

src/luascript.cpp
```diff
@@ -12608,6 +13360,383 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
 	return 1;
 }
 
+// Loot
+
+int LuaScriptInterface::luaCreateLoot(lua_State* L)
+{
+	// Loot() will create a new loot item
+	Loot* loot = new Loot();
```
djarek: Memory leak (it's never deleted).

src/luascript.cpp
```diff
@@ -12608,6 +13360,383 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+	}
+	return 1;
+}
+
+// MonsterSpell
+
+int LuaScriptInterface::luaCreateMonsterSpell(lua_State* L)
+{
+	// MonsterSpell() will create a new Monster Spell
+	MonsterSpell* spell = new MonsterSpell();
```
djarek: Memory leak.

src/luascript.cpp
EvilHero90: I'll remove the reference but what benefit does append give us in this case?

src/monsters.h
```diff
@@ -70,6 +70,66 @@ struct voiceBlock_t {
...
+		void reset();
+
+		uint16_t id;
+		uint32_t chance;
+		uint8_t countMax;
+		int16_t subType;
+		int32_t actionId;
+		std::string text;
+
+		std::list<Loot*> children;
```
djarek: Potential leak of child nodes (does this have to be a pointer?).

src/luascript.cpp
djarek: You can use auto (to make it more readable) and we potentially do less work by letting the implementation of path do the string operation.

src/luascript.cpp
EvilHero90: It's acutally intended so we can make use to add them to different monsters easily without re creating the same loot item again.
ex:
goldCoin = Loot()
goldCoin:setId("gold coin")
goldCoin:setMaxCount(99)

monster = MonsterType("demon")
monster:addLoot(goldCoin)

monster = MonsterType("troll")
goldCoin:setMaxCount(20)
monster:addLoot(goldCoin)


I guess you get my idea, I could probably delete it on addLoot / addChildLoot but then again my idea wouldn't work.

src/luascript.cpp
EvilHero90: same as loot.

src/luascript.cpp
djarek: You still need to make sure that memory is not leaked after a monster reload.

EvilHero90: on reload it would make sense to delete them first I agree, will take a closer look on that in a few :)

EvilHero90: tested several cases now, it seems like relaod already clears the memory as they're bound to lua it seems that the GC takes care of it already, atleast I had no unusual memory increase, memory stayed still for 5 minutes even tho I kept reloading monsters (the entire time) with tons of test files :D

djarek: The lua GC is not fairy dust, it will NOT and in fact CANNOT cleanup memory that is used by the server (unless it's specifically setup to do so, via the __gc() metamethod). The reason that monster reloads don't affect memory usage in this case is the fact that you're not reloading lua-based monsters.

EvilHero90:
Damn phone, closed by misstake sorry

src/luascript.cpp
EvilHero90: on the newest commit I tested it, I'm reloading lua monsters aswell there, it either re adresses stuff or it gets cleaned by GC cause as I said, there is no memory increase or w/e tested this now for quite a while and it seemed fine.

djarek: Cloned your repo, added a counter in the Loot and MonsterSpell constructors and destructors. Ran 10000 reloads. No cleanup occured and over 20 MiB were leaked. You didn't notice it because you only leak 2168 bytes per reload. As I've said the GC cannot reason about your intents and will not cleanup memory allocated by the C++ code, unless you explicitly provide it with facilities to do so (in this case, by registering the __gc method).

EvilHero90: alright, will think about a solution therefore then.

src/luascript.cpp
```diff
@@ -2553,10 +2557,56 @@ void LuaScriptInterface::registerFunctions()
 
 	registerMethod("Condition", "addDamage", LuaScriptInterface::luaConditionAddDamage);
 
+	// NewMonsterType
+	registerClass("NewMonsterType", "", LuaScriptInterface::luaNewMonsterTypeCreate);
+
 	// MonsterType
 	registerClass("MonsterType", "", LuaScriptInterface::luaMonsterTypeCreate);
```
Crypton33: Maybe it could be done in the way that tile related functions are done?
For example:
```lua
MonsterType(name) -- returns existing MonsterType or nil if it doesn't exist
Game.createMonsterType(name) -- returns existing MonsterType or creates and returns new if it doesn't exist
```

src/tools.cpp
```diff
@@ -1162,3 +1165,30 @@ const char* getReturnMessage(ReturnValue value)
 			return "Sorry, not possible.";
 	}
 }
+
+bool loadDirectory(const std::string directory)
```
Crypton33: Function name should tell us what does the function do.
You've moved it outside luascript.cpp, so now it could be hard to tell how the function works without checking its definition.

EvilHero90: ok that certainly makes sense, will rename it to "loadDirectoryLuaFiles" or something like that.

src/luascript.cpp
EvilHero90: let's hear some more people what they think about it, I'm not against it at all but atm rather both are not the best way, I'd really like to have more feedback about that first :)

data/gc/monster_gc.lua
```diff
@@ -0,0 +1,15 @@
+function monster_gc_memory()
```
djarek: This is useless. By storing Loot and MonsterSpell objects here you prevent the garbage collector from cleaning it up until this function is called.

gugahoa: I suggest making loot a table in Lua, and not a class in cpp accessible
through metatable. It would be a lot easier to prevent leak that way.

EvilHero90: djarek and I already taken care of that issue, it's clearing memory on /reload monsters just fine now.

src/monsters.cpp
```diff
@@ -325,6 +517,233 @@ ConditionDamage* Monsters::getDamageCondition(ConditionType_t conditionType,
...
+		} else if (tmpName == "poisonfield") {
+			combat->setParam(COMBAT_PARAM_CREATEITEM, ITEM_POISONFIELD_PVP);
+		} else if (tmpName == "energyfield") {
+			combat->setParam(COMBAT_PARAM_CREATEITEM, ITEM_ENERGYFIELD_PVP);
+		} else if (tmpName == "condition") {
+			uint32_t tickInterval = 2000;
+
+			if (spell->conditionType == CONDITION_NONE) {
+				std::cout << "[Error - Monsters::deserializeSpell] - " << description << " - Condition is not set for: " << spell->name << std::endl;
+				return false;
```
djarek: Leaks combat object.

EvilHero90:
I'm still open for suggestions, I've pretty much finished everything to make it functional.

WibbenZ:
Why close it? ;o @EvilHero90

PR #1297, opened 2015-04-17
Moved Item methods (getId, getActionId, getUniqueId, getSubType)

ninjalulz:
and Creature.getId to Lua core lib


marksamman:
Thanks!

PR #1301, opened 2015-04-19
Make player walkthrough optional

ranisalt:
I have just stumbled across this limitation, that the walkthrough could be optionally enabled at config.lua, and done the patch :)


src/player.cpp
```diff
@@ -825,7 +825,7 @@ bool Player::canWalkthrough(const Creature* creature) const
 	}
 
 	const Player* player = creature->getPlayer();
-	if (!player) {
+	if (!player || g_config.getBoolean(ConfigManager::ALLOW_WALKTHROUGH)) {
 		return false;
```
gugahoa: Shouldn't be !g_config.getBoolean(ConfigManager::ALLOW_WALKTHROUGH)?

src/player.cpp
```diff
@@ -861,7 +861,7 @@ bool Player::canWalkthroughEx(const Creature* creature) const
 	}
 
 	const Player* player = creature->getPlayer();
-	if (!player) {
+	if (!player || g_config.getBoolean(ConfigManager::ALLOW_WALKTHROUGH)) {
```
gugahoa: Same as above.

ranisalt:
True that. Fixed.

EvanMC:
I feel like this should wait until move events are implemented.

marksamman:
I agree with @EvanMC.

EPuncker:
I agree with @marksamman that agree with @EvanMC

ranisalt:
No problem in leaving this PR open? :)

config.lua
```diff
@@ -24,6 +24,7 @@ maxPlayers = 0
 motd = "Welcome to The Forgotten Server!"
 onePlayerOnlinePerAccount = true
 allowClones = false
+allowWalkthrough = true
```
Narzerus: Shouldn't this be false by default? Sefaults should resemble vanilla in my opinion

ranisalt: Nope, in Tibia players are walkthroughable by default. This is not a TFS-only feature, maybe you didn't know it has been implemented in Tibia :P

Narzerus: WHAT? Oh hell no, sorry I got stuck in 7.4 apparently. 

Sorry man

PR #1302, opened 2015-04-19
Fixed issue that kept players seem not invited to private channels!

Herhangi:
Adding "player" instead of "invitePlayer" to "m_invites" list makes invite invalid!


marksamman:
Thanks!

PR #1305, opened 2015-04-21
Fix attack scheduler

ranisalt:
I have stumbled across this problem when I was trying to use an attack speed that was not a multiple of 1000. Apparently the server only checks if there is a new action every 1000 ms, but this limits attack speed to multiples of this time frame, as the server won't check if there is a new attack to calculate in between time frames.

This fix changes it by adding a new SchedulerTask when the attack is in the middle of a time frame. I haven't had time and resources to check if this is a heavy change and if this works properly 100% of the time, but I hope @marksamman can have a look and say somethind :grinning:

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -3383,16 +3383,18 @@ void Player::doAttacking(uint32_t)
 		Item* tool = getWeapon();
 		const Weapon* weapon = g_weapons->getWeapon(tool);
 		if (weapon) {
-			if (!weapon->interruptSwing()) {
+			uint32_t delay;
+
+			if (!weapon->interruptSwing() || canDoAction()) {
 				result = weapon->useWeapon(this, tool, attackedCreature);
-			} else if (!canDoAction()) {
-				uint32_t delay = getNextActionTime();
-				SchedulerTask* task = createSchedulerTask(delay, std::bind(&Game::checkCreatureAttack,
-				                      &g_game, getID()));
-				setNextActionTask(task);
+				delay = getAttackSpeed();
 			} else {
-				result = weapon->useWeapon(this, tool, attackedCreature);
+				delay = getNextActionTime();
 			}
+
+			SchedulerTask* task = createSchedulerTask(delay, std::bind(&Game::checkCreatureAttack,
+									 &g_game, getID()));
+			setNextActionTask(task);
 		} else {
 			result = Weapon::useFist(this, attackedCreature);
 		}


brunominervino:
Hello @ranisalt,

This change don't do effect in distance weapon.

ranisalt:
@brunominervino yes, distance attacks are treated differently (you can see the `if weapon->interruptSwing` that checks if weapon is melee or not) and I haven't figured out how to work this yet.

I will try my best.

brunominervino:
@ranisalt hmm, I understand, first thank you.
I will try to help you :) if I have any progress, I post here :+1:

src/player.cpp
```diff
@@ -3383,15 +3383,22 @@ void Player::doAttacking(uint32_t)
 		Item* tool = getWeapon();
 		const Weapon* weapon = g_weapons->getWeapon(tool);
 		if (weapon) {
+			uint32_t delay;
+
```
brunominervino: if (weapon->interruptSwing() && !canDoAction()) {
				delay = getNextActionTime();
			} else {
				result = weapon->useWeapon(this, tool, attackedCreature);
				delay = getAttackSpeed();
				updateWeapon();
			}

			SchedulerTask* task = createSchedulerTask(delay, std::bind(&Game::checkCreatureAttack,
								  &g_game, getID()));
			setNextActionTask(task);

ranisalt:
Thanks for pointing out, @brunominervino. I have fixed the code and it now works with distance weapons too :D just remember the MINIMUM attack speed is 100 ms, since it is hardcoded and setting it lower may break your server because too many actions will happen in a short amount of time.

brunominervino:
@ranisalt now is working :+1: :)
Thank you bro!

ranisalt:
How's that going?

kanohn:
if paladin try to attack while running speed attack does not work

ranisalt:
@kanohn this is intended behavior. One can not swing a weapon or throw a spear while running away, not even in Tibia.

Replace L3392 `delay = getNextActionTime();` with `delay = getAttackSpeed();` and it may work as you want.

kanohn:
Yes, i know, but the code should respect the speed attack while running
away, because while you're running and after a stop, the speed attack does
not recognize the second between 1000 ~ 2000, can you could change the code
for this to work for me? anything i send to you a payment for the work,
have skype? thank you

2015-06-26 15:21 GMT-03:00 Ranieri Althoff <notifications@github.com>:

kanohn:
but the code should respect the speed attack while running away, because while you're running and after a stop, the speed attack does not recognize the second between 1000 ~ 2000. I will test, thank you! :)

kanohn:
did not work, I'm walking I never I attack now

kanohn:
I think it was all right even if I I'm crazy, really has that same delay is that once I played in some servers 8.6 and the speed attack was so fast that it seemed that even walking was no delay ;P

kanohn:
Hey, i just put: delay = getAttackSpeed() && getNextActionTime(); and should works.. you can tell me if i write correct? i saw in line 3388 you put so !weapon->interruptSwing() || canDoAction()).

|| Means the same thing &&? Thank you for attention

kanohn:
still not working the way I want :(

grisig:
ranisalt, I have added this commit to my server and after that I have noticed that sometimes when you are attacking a creature, it will be impossible to push it. I'm not 100% sure that it's caused by this change, but it might be - what do you think?

grisig:
also attack speed behaves weirdly, both with this 
"Replace L3392 delay = getNextActionTime(); with delay = getAttackSpeed();"
and your original code. I've tested both and there doesn't seem to be any difference.

when you move, it puts a delay on the attack but it only happens sometimes, my attackspeed is 300ms. i'll give you a demonstration: http://gfycat.com/RashClearcutGannet as you can see when I move around I attack much slower

ranisalt:
That's because your attack speed is so abusive it breaks. The `canDoAction()` checks if you aren't moving, as realistically you can not attack while moving, but you don't want that. You could try replacing everything inside `if (weapon)` with:

```cpp
			result = weapon->useWeapon(this, tool, attackedCreature);
			SchedulerTask* task = createSchedulerTask(getAttackSpeed(), std::bind(&Game::checkCreatureAttack, &g_game, getID()));
			setNextActionTask(task);
```

That way the attack only cares for the last attack time, not if you are moving or whatever.

marcelobcortes:
your code ignored the fist fighting

zydent:
I have the same problem as grisig https://gfycat.com/RashClearcutGannet @ranisalt

**I do not know how to add this code could help me if it is not much annoyance**
>             result = weapon->useWeapon(this, tool, attackedCreature);
>             SchedulerTask* task = createSchedulerTask(getAttackSpeed(), std::bind(&Game::checkCreatureAttack, &g_game, getID()));
>             setNextActionTask(task);

zydent:
@ranisalt plx help bro :(

brunominervino:
@zydent have a open issue about this case #1319

Mkalo:
@zydent This is not a bug in his change, distance weapons have interrupt swing (even in real tibia), it stops your attack when you move. If you want to remove this you can change this line:
https://github.com/otland/forgottenserver/blob/master/src/weapons.h#L163
to false.

The only problem with this pull request is it switched the places for getAttackSpeed and getNextActionTime and it completely ignored fist weapon.

ranisalt:
The PR is broken because I messed with the branch and I will reopen it.

zydent:
@Mkalo I made the change "**return false;**" The speed stop when I move. https://gfycat.com/ShorttermOptimisticFlamingo

Mkalo:
@zydent Then remove those lines:
https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L1212-L1213

This may affect something else.

zydent:
@Mkalo Thank you very much, it works 👍

zydent:
@Mkalo Any action makes the fast attack stop, Of me emotion I hadn't noticed this big mistake :( https://gfycat.com/SpicyQuarrelsomeIcelandichorse

Mkalo:
@zydent Alright just delete `setNextActionTask(task);` in the doAttacking, that way nothing can interfere with the scheduled check. (That's not how real tibia works but thats just for your custom case.)

Roniox12:
Hi @Mkalo can you help me with issue that ranged weapons stop attack speed during movement and using runes on this distro https://github.com/TwistedScorpio/OTHire ? Thank you !

PR #1307, opened 2015-04-22
fromPosition now behaves properly

gugahoa:
fromPosition was not behaving properly in some cases.
Example:
When using a ladder, fromPosition should be where the player is and toPosition where the ladder is.
When using anything that was on the floor, it would behave like this, fromPosition and toPosition would be the position of the thing you used.


marksamman:
This is incorrect. fromPosition should be the items position. It may have been incorrect before, but it is certainly working as intended now. The arguments should probably be renamed to itemPosition and targetPosition though.

gugahoa:
Makes sense. If I want the player position I can do a creature:getPosition() anyway. Thanks for clarifying :)

PR #1311, opened 2015-04-25
Fix race condition in DatabaseTasks

djarek:
Fix race condition (runTask() would race with task flushing on the front queue object during server shutdown). Remove unnecessary thread state check.


ranisalt:
pretty good use of modern standards :)

marksamman:
Thanks!

PR #1317, opened 2015-04-29
Remove unused imports

ranisalt:
Any reason why those files are still being imported despite not being used?

The generated binary is exactly the same after removing those imports.


marksamman:
No reason. Thanks!

Issue #1273, opened 2015-04-05
Automatic Remount


WibbenZ:
Tried to serach for an issue but could not find one.
Would be nice to add this, if you enter PZ mounted it will automatically mount you when you leave the PZ zone.

EPuncker:
yeah indeed :+1: 
one more thing:
when you logout (doesn't matter if pz or not) and login again, you should get dismounted

WibbenZ:
@marksamman if this is something you plan to add yourself, will this make 1.1 or will it be in a later version?
I could give it a try, but in what way would you store if the player is mounted or not?
The easiest way I can see it is using the player class, here: https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L63
And a boolean value if it should remount or not.

marksamman:
No new features will make it into 1.1. Yes, you would store a boolean in class Player.

PR #1323, opened 2015-05-03
Automatic Remount https://github.com/otland/forgottenserver/issues/1273

PrinterLUA:


PrinterLUA:
It works, but i know the boolean name is bad and also should i make it option in the config?

@marksamman

djarek:
For future reference: you don't have to provide links to issues, instead, just write #number_of_the_issue in the title/description of the PR.

ranisalt:
My suggestion: name it `lastMountStatus`

ranisalt:
My suggestion: name it `lastMountStatus` and create an enum with `NOT_MOUNTED` and `MOUNTED`.

WibbenZ:
```
bool shouldMount = false;
```
Sounds better to me, then just use a boolean insted of enums.

marksamman:
Use wasMounted as the variable name, and squash your commits when you're
done.

PrinterLUA:
Done

marksamman:
Thanks!

Issue #770, opened 2014-04-30
Linux signals.


dominique120:
@marksamman I know I mentioned this before and I said I would figure it out or whatever but could you please add support for them? Just basic ones like reload all and save and quit.

I tried doing it myself but I have not been able  to do it successfully(I'm not a programmer).

Just the "Save & Quit" signal would be a huge blessing.

Thanks!
##

ranisalt:
What signals could be implemented? I'm thinking on INT/TERM to clean and stop the server, and STOP/CONT to close the server and open, respectively. This would make the server behave easily as a daemon.

djarek:
I believe that HUP is used by some init systems (at least by systemd) to signal a request for a reload of config files.

ranisalt:
Nice, a good replacement for issuing /reload ingame too.

PR #1328, opened 2015-05-03
[WIP] Proper signal handling

ranisalt:
I have started working in issue #770, which addresses the capability of handling signals. It is easy to implement some of the signals. In the issue, we have thought of handling SIGINT to shutdown the server properly e.g. on Ctrl+C, SIGSTOP/SIGCONT to close and open game server and SIGHUP to reload configuration files, as some system daemons (namely systemd and OpenRC) tend to do.

This would make it easy to run the server in background as a daemon without the need of tools such as nohup and screen, and maybe ship the server in package managers for even easier installation.

Preview with SIGINT:
Preview

QUESTION: how could one reload the entirety of server configuration? Something like a `/reload all` command would do. I have checked the `/reload` command and it needs a keyword to reload only a single part of configuration, though it would be possible to just reload one by one, there may be a simpler way to do it.

### DIFF
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -76,8 +76,8 @@ void badAllocationHandler()
 	exit(-1);
 }
 
-void sigtermHandler(int) {
-	std::cout << "SIGTERM received, shutting game server down..." << std::endl;
+void sigintHandler(int) {
+	std::cout << "SIGINT received, shutting game server down..." << std::endl;
 	g_game.setGameState(GAME_STATE_SHUTDOWN);
 }
 
@@ -97,7 +97,7 @@ int main(int argc, char* argv[])
 
 	// Setup interrupt signal handler
 	struct sigaction sigint;
-	sigint.sa_handler = sigtermHandler;
+	sigint.sa_handler = sigintHandler;
 	sigint.sa_flags = 0;
 	sigemptyset(&sigint.sa_mask);
 	sigaction(SIGINT, &sigint, nullptr);
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -76,6 +76,11 @@ void badAllocationHandler()
 	exit(-1);
 }
 
+void sigtermHandler(int) {
+	std::cout << "SIGTERM received, shutting game server down..." << std::endl;
+	g_game.setGameState(GAME_STATE_SHUTDOWN);
+}
+
 int main(int argc, char* argv[])
 {
 	// Setup bad allocation handler
@@ -89,6 +94,13 @@ int main(int argc, char* argv[])
 	sigemptyset(&sigh.sa_mask);
 	sigaction(SIGPIPE, &sigh, nullptr);
 #endif
+
+	// Setup interrupt signal handler
+	struct sigaction sigint;
+	sigint.sa_handler = sigtermHandler;
+	sigint.sa_flags = 0;
+	sigemptyset(&sigint.sa_mask);
+	sigaction(SIGINT, &sigint, nullptr);
 
 	ServiceManager serviceManager;
 


gugahoa:
+1
I agree with your suggestion of SIGSTOP, SIGCONT and SIGHUP. I would suggest one signal for server saving, without the need to close the game servers.

ranisalt:
I vote on USR1 for saving, as this signal has no specific meaning and there are no signals that mean to save the process.

djarek:
Perhaps you should move the handler definitions and registration to a separate file for readability?

diath:
How about a Lua function to pass the signals to? This way we could keep sane defaults while allowing people to customize them the way they want.

djarek:
```
A signal may be generated (and thus pending) for a process as
   a whole (e.g., when sent using kill(2)) or for a specific
   thread (e.g., certain signals, such as SIGSEGV and SIGFPE,
   generated as a consequence of executing a specific machine-
   language instruction are thread directed, as are signals
   targeted at a specific thread using pthread_kill(3)).  A
   process-directed signal may be delivered to any one of the
   threads that does not currently have the signal blocked.  If
   more than one of the threads has the signal unblocked, then
   the kernel chooses an arbitrary thread to which to deliver
   the signal.
```
The kernel is theoretically free to choose any thread to run the signal handler. Which could cause a race condition on the lua script environment/game state. We could theoretically just add the a dispatcher task from the handler, but taking locks from within a signal handler is a really bad idea.

marksamman:
I'll just leave this link with recommendations here for now:
https://www.securecoding.cert.org/confluence/pages/viewpage.action?pageId=1420

ranisalt:
Yes, it should not mess with Lua. This is a core functionality and thus should be handled in the C++ part.

@djarek I think using already made functions, such as setting game state to shutdown, is thread safe and does not cause race. I should investigate it further later.

djarek:
If we are to be pedantic, changing the "game state" flag outside the dispatcher thread will cause a race on the flag, and we risk torn write etc. In most cases this isn't going to screw anything up, however, the behavior of the application is undefined (can be fixed easily by using the atomic<> template for the flag and use relaxed memory ordering so that we don't change the performance).

HeavenIsLost:
Its kinda of the off topic, but the dispatcher thread is the one which room game world? Which one thread run the services/protocols? How much threads there is in tfs?

djarek:
Currently there are 4 threads:
*Dispatcher (Game logic)
*Scheduler
*Asynchronous I/O service (networking)
*Asynchronous database task dispatcher.

ranisalt:
Well, what if we create a new task in the dispatcher to shutdown the game?

We should be pedantic, we can not risk losing server data because of a bad implementation :P

marksamman:
It may be safe to add a dispatcher task from a signal handler **if** we make the taskLock mutex recursive. The safest way in Linux is using signalfd, but it's not portable.

djarek:
@ranisalt
Adding a task is not a problem in itself. However, a signal handler is allowed to interrupt a thread in any arbitrary location, which means that the handler could be called just after the dispatcher task mutex has been acquired. The handler would attempt to acquire the mutex and the dispatcher thread would deadlock with itself. 

Recursive mutexes impose additional overhead. I see little benefit in providing a way to define signal handlers in lua. If someone really has the need for a custom handler then we should assume they have the knowledge to modify the handlers/add custom ones. We shouldn't overgeneralize with features like these and we should optimize for the most common use-case.

I forgot to mention that this behavior (namely, that any thread is allowed to be the executor of the handler) is the default and I believe that it is possible to block and enable signal handlers for specific threads. Still, you need to be careful and should limit access to variables that are either volatile or atomic (the latter is preferable) to avoid latent bugs.

ranisalt:
According to [this part](https://github.com/torvalds/linux/blob/54e514b91b95d6441c12a7955addfb9f9d2afc65/kernel/signal.c#L921-L956) of Linux, only the threads that register the signals can be signaled. So I guess it would be good to register the signals in the dispatcher or the main thread only. The problem is this behavior is Linux-only and it can't be guaranteed to work on Windows.

djarek:
BTW, something that might not be obvious at first: only functions that are reentrant (being thread safe is not enough) can be called within a signal handler. That means that even memory allocation within a signal handler would cause a race condition and potentially corrupt heap metadata. This means that both creating a dispatcher task and adding it to the task list are unsafe and would cause undefined behavior.

marksamman:
Closing this as I don't see any progress, and signal handling is a hard problem to solve with portability in mind. Once we create a new admin protocol, there should be a command line tool to send "signals".

ranisalt:
Yes, that proved to be a challenge by now. It's better to wait for changes :D

djarek:
@marksamman 
So are we giving up on signal handling and just replacing it with the admin protocol or are both going to be implemented?

marksamman:
@djarek I'm still OK with catching SIGTERM to save and shutdown gracefully if we can get it right; but the admin protocol is going to be implemented regardless.

PR #1331, opened 2015-05-06
Spectator cache allocation efficiency improvement

djarek:
Use std::make_shared<> to avoid an unnecessary, additional heap allocation.


marksamman:
Thanks!

PR #1335, opened 2015-05-07
Fix potential deadlock and race condition in the dispatcher.

djarek:
Fix potential deadlock on dispatcher shutdown (it could potentially deadlock with itself if there was a task pending that added a new task during flushing). Use an atomic (with relaxed ordering) for the thread state flag, this fixes a race condition on that flag in the main dispatcher loop. Remove unncessary thread state check.

### DIFF
--- a/src/tasks.cpp
+++ b/src/tasks.cpp
@@ -25,14 +25,9 @@
 
 extern Game g_game;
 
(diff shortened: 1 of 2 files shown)


marksamman:
Thanks!

PR #1337, opened 2015-05-08
Move enchant spear from conjure to instant

### LINKED ISSUE #1336: Enchanted Staff does not start decaying
The spell "Enchant Staff" that uses a reagent with id 2401 to conjure an item with id 2433 (enchanted staff) does not start the decay time of the item.

gugahoa:
Fixes #1336 
As enchant staff is the only conjure spell that needs to start decay on the item, it would be easier to just make an specific script for it.

I moved from conjure to instant because with conjure, if I used the tag script, it would not automatically apply the exhaust, group exhaust, mana consumption, level requirement check and so on for some reason, if anyone could explain to me why as I didn't look into it too much I would appreciate. :)


marksamman:
I disagree. Any item that can

PR #1340, opened 2015-05-09
Reduce static tile memory usage

djarek:
The item vector will no longer be created if the item being added is a ground tile. This will considerably drop memory usage for static tiles (approximately 13 MiBs on an x64 build, with the default data pack).


marksamman:
Thanks!

Issue #1343, opened 2015-05-09
[Suggestion] Shared Experience


edumntg:
Move shared-experience mechanism to lua (Player events?)
##

Kamenuvol:
If I had time, I would have started working on it. My idea is to create a function to set the criteria of dividing xp or set a Party Rate on a new Party:onShareExperience event.
And on Player:onGainExperience multiply the rate returned by the function, or an actual party rate.

Kamenuvol:
Also the exp sharing thing is completely buggy. The experience values showed on the server log and on the game screen are different and the formula is wrong.

ranisalt:
@Kamenuvol could you explain further on why the formula is wrong? We had some discussion here some time ago and concluded the formula is correct and only misses by a few points, related to flooring or ceiling the result.

Kamenuvol:
```c++
uint32_t shareExperience = static_cast<uint64_t>(std::ceil(((static_cast<double>(experience) / (memberList.size() + 1)) + (static_cast<double>(experience) * extraExpRate))));
```
As we see, we are doing (exp)*(1/size)  + (exp)*(rate) or (exp)*(rate + 1/size)
The correct formula is: (exp)*(rate)/size

e.g: A party with 4 players with 3 different vocations kills a monster that gives 800xp.
Current formula: 800*(0.6 + 0.25) = 800*0.85 = 680
Correct: 800*1.6*0.25 = 320

Source: http://tibia.wikia.com/wiki/Party

edumntg:
Correct formula is ((exp*party_bonus)/size)*player_bonus

party_bonus = shared experience bonus (+20% for sharing +x% depending of vocations in the party)
size = size of party
player_bonus = green stamina?

e.g: A knight inside a party with 2 members (paladin and druid), and the knight has green stamina. The kill a demon:

exp = 6000
party_bonus = +20% +40%(3 different vocations) = +60% = 1.6
size = 3
player_bonus = +50% = 1.5

Total experience = ((6000*1.8)/3)*1.5 = 4800

Kamenuvol:
I've omitted the player bonus because it is already in Player events. Party shared exp is indepedent.

ranisalt:
Please have a look at the PR.

sircurse:
Hi folks, I managed to add the chagnes that Renisalt proposed, and it worked great for me, it must be a fix not a enhancement.

vankk:
[...]

PR #1344, opened 2015-05-10
Fix shared experience formula

ranisalt:
@Kamenuvol has pointed that the formula to reward shared experience is wrong in #1343. According to Tibia Wiki it should be `experience * bonus / party_size`, but it was `experience / party_size + experience * bonus` incorrectly.

Also, the container variable is of incorrent type, being uint32_t instead of uint64_t, but this is not the purpose of this pull.

### DIFF
--- a/src/party.cpp
+++ b/src/party.cpp
@@ -418,11 +418,11 @@ bool Party::setSharedExperience(Player* player, bool _sharedExpActive)
 
 void Party::shareExperience(uint64_t experience, Creature* source/* = nullptr*/)
 {
-	uint32_t shareExperience = static_cast<uint64_t>(std::ceil(((static_cast<double>(experience) / (memberList.size() + 1)) + (static_cast<double>(experience) * extraExpRate))));
+	uint64_t sharedExperience = static_cast<uint64_t>(std::ceil(experience + experience * extraExpRate) / (memberList.size() + 1));
 	for (Player* member : memberList) {
-		member->onGainSharedExperience(shareExperience, source);
+		member->onGainSharedExperience(sharedExperience, source);
 	}
-	leader->onGainSharedExperience(shareExperience, source);
+	leader->onGainSharedExperience(sharedExperience, source);
 }
 
 bool Party::canUseSharedExperience(const Player* player) const


Kamenuvol:
The formula is OK now

sircurse:
Hi Ranisalt, I managed to add this fix to my source and compiled, for me it seems to be working perfect, I dont know if that fix are alraedy in the next version but should be.

Curse.

marksamman:
I think this is wrong, which is confusing me because two users are claiming that it's working. extraExpRate is between 0 and 1, so when it's 20% it's 0.2, resulting in exp * 0.2, which means 20% of the experience being shared with the party members instead of the experience + 20% extra experience.

It should be:
(experience + (experience * extraExpRate)) / party_size

... or we could change the value of extraExpRate to be 1.2 instead of 0.2 and just do:
(experience * extraExpRate) / party_size

kanohn:
Hello friends,

I trying to change this function bellow, to share exp accept only number of members in the party instead of ids vocations [tfs 1.1]

In Party.cpp:
[code]void Party::updateVocationsList()
{
    std::set<uint32_t> vocationIds;

    uint32_t vocationId = leader->getVocation()->getFromVocation();
    if (vocationId != VOCATION_NONE) {
        vocationIds.insert(vocationId);
    }

    for (const Player* member : memberList) {
        vocationId = member->getVocation()->getFromVocation();
        if (vocationId != VOCATION_NONE) {
            vocationIds.insert(vocationId);
        }
    }

    size_t size = vocationIds.size();
    if (size > 1) {
        extraExpRate = static_cast<float>(size * (10 + (size - 1) * 5)) / 100.f;
    } else {
        extraExpRate = 0.20f;
    }
}[/code]

I tried change [code]size_t size = vocationIds.size();[/code] to [code]size_t size = memberList.size();[/code] but not works..  Can someone help me? thanks very much.

WibbenZ:
https://otland.net/forums/support.16/

kanohn:
I tested exp share and everything seems to be ok, but if you are giving a bonus of 20%, and I kill a monster that experience is "Pair", it returns me a different experience than it should be, if I kill a monster with exp "Odd" it returns me to experience 100% as it should.

This happens vice versa, if I have a bonus 30%, and a monster has an "Odd" experience, returns true exp. This is a bug?

ranisalt:
@marksamman so you say it should be something like that?
```
uint64_t shareExperience = (experience + experience * extraExpRate) / (memberList.size() + 1);
```
(I removed the static cast and ceil to make the line shorter, but consider it)

I made it like that and I'm facing a new problem, when the players share experience, it is awarded split, like that:
Imgur
With one entry for each party member, and value according to their participation on the kill. On the client, it sums up and shows only 1 value, the expected one.

Is this the expected behavior? To tell the truth it has been a long time since I shared experience on Tibia.

ranisalt:
The problem is the experience is shared for every killer, and there might be multiple killers on the same guild.

A friend of mine told me in Tibia there comes only one message, so I thought on the following:
When a monster is killed, and one of the killers is member of a party that is sharing experience, his share of the experience is sent to the leader, which then redistributes the experience equally among party members.

That way, there will be only 1 value of experience shared, instead of one for every player. Players outside the party will still get their proportional share.

ranisalt:
Can someone interested in the solution check if the values now are correct? @Kamenuvol @sircurse @kanohn

kanohn:
@ranisalt 

I get your new formula: uint64_t sharedExperience = static_cast<uint64_t>(std::ceil(experience + experience * extraExpRate) / (memberList.size() + 1));

and changed to this:
uint64_t sharedExperience = static_cast<uint64_t>(std::ceil(experience * extraExpRate) / (memberList.size() + 1));

And works 100% now, give to me a 100% expected experience.

Im using 1.2f instead 0.2f

HeberPcL:
Maybe fixed in https://github.com/otland/forgottenserver/pull/1956

PR #1348, opened 2015-05-10
Scheduler thread fix & code improvement

djarek:
Fix race condition on the threadState flag in the scheduler thread by using an atomic with relaxed ordering. Remove unnecessary comparator, use the default container for the priority_queue adapter template (which is more efficient than a vector in that role). Remove unnecessary flag check in scheduler loop.


marksamman:
Thanks!

PR #1360, opened 2015-05-16
Remove shared_ptr from the spectator cache.

djarek:
Remove unnecessary use of shared_ptr from the spectator cache.


marksamman:
Thanks!

PR #1361, opened 2015-05-17
Fix shared experience formula

edumntg:
@ranisalt wrote the following on a previous pull request: "@Kamenuvol
has pointed that the formula to reward shared experience is wrong in
#1343."

He made a change for the formula but it wasn't returning correct values
(it was returning a 1/10 of the real value)

I fixed it and tested, and now it's working as real tibia. Also, i
changed the container variable type from uint32_t to uint64_t.


edumntg:
Please close this pull request, i'll create a new one

PR #1364, opened 2015-05-17
Remove unnecessary includes.

djarek:


ranisalt:
I think tools.h is not used in player.cpp too

marksamman:
Thanks!

Issue #1365, opened 2015-05-18
Create party by command TFS 1.1


JoseEduardo:
I saw that on github TFS has this page:
https://github.com/otland/forgottenserver/wiki/Metatable:Party

but the examples are not clear, as I do to create a party with some player, noting that this party does not exist yet.
For example, we have the desert quest (as in Tibia Global) as I do for by the fourth player in a party without them to do anything.

I created a topic in otland ( https://otland.net/threads/create-party-by-command-tfs-1-1.232438/#post-2241406 ) with the matter, but did not get good results, I tried to create that way and it was not possible:

```
local player1 = Player(getPlayerByName('Admin'))
local party = Party(player1)
party:addInvite(player1)
```

but I get it in return

```
attempt to call global 'Party' <a table value>
```
##

djarek:
As it is now, you cannot create a new party within the script system. You can only get a party object (representing an already existing party) by calling player:getParty().

JoseEduardo:
it would be possible to accomplish this in another way?

djarek:
You would need to write a lua interface function which allows a new party to be created.

JoseEduardo:
There is some chance or interest in doing that? or a kindness to tell me where to start?

WibbenZ:
A tip would be to look here, ttps://github.com/otland/forgottenserver/blob/master/src/game.cpp#L4716-4721

JoseEduardo:
I do not know how does one pull request if you can make and test it, I did two LUA functions that help a lot.

luascript.h
```
static int32_t luaDoCreateParty(lua_State* L);
static int32_t luaDoJoinParty(lua_State* L);
```

luascript.cpp
```
	//doCreateParty(cid, invitedPlayer)
	lua_register(m_luaState, "doCreateParty", LuaScriptInterface::luaDoCreateParty);

	//doJoinParty(lider, invitedPlayer)
	lua_register(m_luaState, "doJoinParty", LuaScriptInterface::luaDoJoinParty);
```

```

int32_t LuaScriptInterface::luaDoCreateParty(lua_State* L)
{
	//doCreateParty(cid, invitedPlayer)
	Player* player = getPlayer(L, 1);

	if (player){
		Party* party = player->getParty();
		if (!party) {
			party = new Party(player);

			Player* invitedPlayer = getPlayer(L, 2);
			party->invitePlayer(*invitedPlayer);
		}
		else if (party->getLeader() != player) {
			std::ostringstream ss;
[...]

PR #1371, opened 2015-05-21
player:createParty(targetPlayer[, joinTargetPlayer = false])

paulosalvatore:
#1365

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1060,12 +1060,6 @@ void LuaScriptInterface::registerFunctions()
 	//sendGuildChannelMessage(guildId, type, message)
 	lua_register(m_luaState, "sendGuildChannelMessage", LuaScriptInterface::luaSendGuildChannelMessage);
 
-	//doCreateParty(cid, invitedPlayer)
-	lua_register(m_luaState, "doCreateParty", LuaScriptInterface::luaDoCreateParty);
-
-	//doJoinParty(lider, invitedPlayer)
-	lua_register(m_luaState, "doJoinParty", LuaScriptInterface::luaDoJoinParty);
-
 #ifndef LUAJIT_VERSION
 	//bit operations for Lua, based on bitlib project release 24
 	//bit.bnot, bit.band, bit.bor, bit.bxor, bit.lshift, bit.rshift
@@ -2233,6 +2227,7 @@ void LuaScriptInterface::registerFunctions()
 
 	registerMethod("Player", "getSlotItem", LuaScriptInterface::luaPlayerGetSlotItem);
 
+	registerMethod("Player", "createParty", LuaScriptInterface::luaPlayerCreateParty);
 	registerMethod("Player", "getParty", LuaScriptInterface::luaPlayerGetParty);
 
 	registerMethod("Player", "addOutfit", LuaScriptInterface::luaPlayerAddOutfit);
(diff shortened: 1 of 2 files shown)


WibbenZ:
This won't be accepted since we want to use the meta tags functions, not the legacy system.
Serach for
```
registerMethod(
```
To see how you can do that.

src/luascript.cpp
```diff
@@ -12475,6 +12481,59 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+		lua_pushboolean(L, false);
+	}
+	return 1;
+}
+
+int32_t LuaScriptInterface::luaDoJoinParty(lua_State* L)
+{
+	//doJoinParty(lider, invitedPlayer)
+	Player* player = getPlayer(L, 2);
+	if (!player) {
```
djarek: If you're doing checking for errors/invalid inputs, please use if(somecondition) {report_errors; return 1} rather than if elses because it makes the code slightly harder to read.

paulosalvatore:
I'm trying to do something like this to use methods:

```
static int luaPartyCreate(lua_State* L);
```
```
registerClass("Party", "", LuaScriptInterface::luaPartyCreate);
```
```
int LuaScriptInterface::luaPartyCreate(lua_State* L)
{
	// Party(player1, player2)
	Player* player1 = getPlayer(L, 2);
	Player* player2 = getPlayer(L, 3);
	Party* party;
	party = new Party(player1);
	if (player1 && player2 && party) {
		party->invitePlayer(*player2);
	} else {
		lua_pushnil(L);
	}
	return 1;
}
```

WibbenZ:
Tbh I would just do something like Player:createParty(targetPlayer), then check if the player or targetPlayer already has a party.
There is also no reason to push a boolean value when you push the userdata value :p

paulosalvatore:
Seems nice. I'll try to do something

paulosalvatore:
```
static int luaPlayerCreateParty(lua_State* L);
```
```
registerMethod("Player", "createParty", LuaScriptInterface::luaPlayerCreateParty);
```
```
int LuaScriptInterface::luaPlayerCreateParty(lua_State* L)
{
	// player:createParty(targetPlayer)
	Player* targetPlayer = getUserdata<Player>(L, 2);
	// Can/May I use 'getPlayer();' instead of `getUserdata<Player>() here?
	if (!player || !targetPlayer) {
		lua_pushnil(L);
		return 1;
	}

	Party* party;
	Party* partyPlayer = player->getParty();
	Party* partyInvitedPlayer = targetPlayer->getParty();

	if (partyPlayer) {
		reportErrorFunc("Player is already on a party.");
		pushBoolean(L, false);
		return 1;
	} else if (partyInvitedPlayer) {
		reportErrorFunc("Invited Player is already on a party.");
		pushBoolean(L, false);
		return 1;
	} else {
		party = new Party(player);
		party->invitePlayer(*targetPlayer);
		pushUserdata<Party>(L, party);
		setMetatable(L, -1, "Party");
	}
	return 1;
}
```

WibbenZ:
Since you don't have any objects to check you can't use getPlayer.
There is no reason to send an error out, just return false, there is also no reason to use return 1; in the if statments since you don't have any code after the if statment.

You also don't have to create party, just overwrite partyPlayer.

paulosalvatore:
```
int LuaScriptInterface::luaPlayerCreateParty(lua_State* L)
{
	// player:createParty(targetPlayer)
	Player* targetPlayer = getUserdata<Player>(L, 2);
	if (!player || !targetPlayer) {
		lua_pushnil(L);
		return 1;
	}

	Party* partyPlayer = player->getParty();
	Party* partyInvitedPlayer = targetPlayer->getParty();

	if (!partyPlayer && !partyInvitedPlayer) {
		partyPlayer = new Party(player);
		partyPlayer->invitePlayer(*targetPlayer);
		partyPlayer->joinParty(*targetPlayer); // I've added this line so that the party invite will be automatically accepted.
		pushUserdata<Party>(L, partyPlayer);
		setMetatable(L, -1, "Party");
	}
	return 1;
}
```

WibbenZ:
Looks good, but imo you shoulden't force the join function.

PrinterLUA:
I think it should be done something like this:
http://pastebin.com/m5cCDciB

Since there is no reason to use target on createParty. We have already seprate function for that: party:addMember(player)

DSpeichert:
#1816 was merged instead which provided same functionality. Inviting or adding members is covered by separate functions already.

PR #1374, opened 2015-05-23
Variable Rename

brunominervino:
According avoid variable shadowing


marksamman:
There's no variable shadowing there, but I'll merge this anyway because we don't want names starting with underscores as they may be reserved.

brunominervino:
Hmm I understand, anyway thank you @marksamman

PR #1377, opened 2015-05-24
Use OpenSSL for SHA1 hashing

ranisalt:
Being a game server, TFS should not have to worry with the implementation of cryptographic functions. As such, I have refactored the `transformToSHA1` function from the tools file to use the well tested OpenSSL library to hash messages.

This reduces the code in 100 lines and relies on a good quality library instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ include(FindCXX11)
 
 # Find packages.
 find_package(GMP REQUIRED)
+find_package(OpenSSL REQUIRED)
 find_package(MySQL)
 find_package(LuaJIT)
 find_package(Threads)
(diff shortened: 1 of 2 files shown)


ranisalt:
I have also refactored the function to allow any hash function with the same signature as SHA1 (virtually any function that does not require a salt, such as all SHA functions, MD5, and even the strong Whirlpool), and I wish to make a pull request if this is desired.

djarek:
According to this:
https://www.openssl.org/docs/crypto/sha.html
```
Applications should use the higher level functions EVP_DigestInit etc. instead of calling the hash functions directly.
```
Also remember to consider thread safety when using external libraries (OpenSSL is known not to be thread safe by default).

marksamman:
Once we move away from SHA1 (which we hopefully do in 1.2 because SHA1 is considered weak today), we will use scrypt, bcrypt, PBKDF2 or maybe even one of the Password Hashing Competition candidates. As for library, we prefer NaCl/libsodium over OpenSSL.

The problem with using more expensive hash functions is that it would block either the dispatcher or the asio thread for a significant time with our current design, so we need to deal with that problem too.

djarek:
@marksamman 
I believe that the authentication process can be moved to a separate thread fairly easily. Though running only authentication on that thread would be a waste of resources, considering how often authentication occurs.

ranisalt:
@marksamman my worry is that people will just ignore and do not choose a good salt for hashing. 

I'd like to implement PKBDF2 and will have a look at NaCl. I don't think that it takes a very long time as the only thing hashed are the passwords (not too often) and the MotD, which could easily be cached. Am I right?

WibbenZ:
When you brought the MOTD up, @marksamman why is that encrypted? Could it be to make it smaller in size or?

marksamman:
@ranisalt if we go with libsodium (which is a fork of NaCl), we should use crypto_pwhash_scryptsalsa208sha256_str or wait for the Password Hashing Competition to be over, which is when libsodium will add a high-level API for password hashing (but then we'll also have to wait for popular Linux distributions to get the new version of libsodium in their stable package repositories...).

I would prefer using bcrypt for its library availability to be sure that people will actually use it and not stay with SHA-1, bcrypt is default in PHP's password_hash, which is available since PHP 5.5, so I don't think people would have a problem adapting to it like they would have if we choose scrypt+salsa208+sha256.

The MOTD doesn't need a strong hash function, it just needs a fairly low chance of collisions.

ranisalt:
@ranisalt good enough, bcrypt being PHP standard also makes it better to implement on current working websites as it is just a single function change. For MOTD, since we are using a specialized library we could use whatever function, even something lighter like MD5, or keep SHA1.

I guess our best bet is to wait for the competition since it would be upsetting to change the hashing function multiple times :P I don't think it would take too much time for systems to update since this should be regarded as a security update.

PR #1381, opened 2015-05-25
Move some functions to compat

WibbenZ:
Moved a total of 14 functions, as following:
```
doPlayerAddItem()
doAddContainerItem()
createCombatObject()
createConditionObject()
setCombatArea()
setCombatCondition()
setCombatParam()
setConditionParam()
addDamageCondition()
addOutfitCondition()
setCombatCallBack()
setCombatFormula()
setConditionFormula()
doCombat()
```


kornholi:
Have you tested doPlayerAddItem? I don't think the return values match the old function.

WibbenZ:
By just looking at luascript.cpp they both push boolean values(or nil ofc).
But ill try them both out to be sure.

WibbenZ:
http://3.1m.yt/ziNg33hSn.png

They both return userdata values right now and the question is if we should return uid or a userdata value.

WibbenZ:
@kornholi should I make it return the uid insted of userdata?
Is there something else to do?

marksamman:
The function should return exactly what it returned before it was moved to compat.lua, in every possible case. If it was that easy to move it to compat.lua, we would have done it already.

WibbenZ:
Well spells could be moved, as long as all of them are moved.
And in this case all of them are moved and ive tested a spell that used the old functions and had no issues at all.
Ill make it return the uid insted of the userdata value.

There are still functions, like upTime that could be moved since they don't really do anything that differs from a new function(in this case returns a time).

WibbenZ:
@marksamman  should I fix it or close it?
If you feel that there is no reason to move them we can just as easily close this.
But as I said, ive tested spells and there was no problems - so I don't really see the argument for not merging this.

marksamman:
You should fix doPlayerAddItem or revert it, or if you don't want this to be merged at all then close it. I haven't reviewed the other functions yet so I can't comment that.

WibbenZ:
I did fix doPlayerAddItem, but after I fixed the report squash this one got screwed up.
But in that case ill fix it.

PR #1388, opened 2015-05-28
New Outfits & Mounts

brunominervino:
Changes:

* Min version was changed to 1077 for support the new outfits & mounts;
* Newtypes talkaction support the new outfits & mounts and checks if params are empty;
* Fix all mounts speed;
* Fix Spirit Caller outfit male.


ranisalt:
Why have you changed all mount speeds? Every level adds 2 points to speed, that's why you see (speed +10) but in fact it increases by 20. (At least that's what I remember)

Plus I don't think changing the max version UP is as easy as changing that line. More stuff has to be done in items.otb from what I recall. Let Mark do this if it's the case.

brunominervino:
@ranisalt
Check all mounts: http://tibia.wikia.com/wiki/Mounts
All have speed + 10 now

The same to version 10.77, in the current version no have bit changes. No need to update the items.otb because they were not implemented new items.

EPuncker:
dude, just look in game at boots of haste, the bonus is speed*2 because of speed level bonus as pointed by @ranisalt and pointed many times in otland

https://otland.net/threads/items-xml-speed.81217/#post-834479
https://otland.net/threads/item-speed.145022/#post-1394823
https://otland.net/threads/mounts-outfits-on-tfs-1-0.208509/#post-1996978

PR #1391, opened 2015-05-31
Move some functions to compat

WibbenZ:
Moved doPlayerAddItem and doAddContainerItem to compat.
Accidentally added 2 files that removes trailing spaces.


WibbenZ:
@marksamman I was wrong with the spells, tested with this datapack insted of ORTS (that apparently only has a weapon file left to port and for some reason it worked with that script, but spells does not seem to like it).

PR #1396, opened 2015-06-01
Fix offline traning ignoring skill/magic rate

ranisalt:
Offline training is not taking into account the rate multiplier for every skill, [check this report](https://otland.net/threads/tfs-1-1-offline-training-is-really-slow-shouldnt-skill-magic-rate-change-this.228442/).

Thanks to @WibbenZ for pointing out what lines should be modified.


WibbenZ:
Have you checked that they return the same tries?
Problem could be that we now give a slower / higher rate :p 

What I was talking about on OTLand was more or less to add new rates, so they can be exaclly the same but can still be changed with ease.

ranisalt:
Was it at least improved or fixed?

marksamman:
I haven't confirmed that it's fixed, but it looks to me like it calls the Lua event which applies the multiplier. Regardless, the system has been moved to Lua and this PR would have to be rewritten if it's the case that this is still an issue.

PR #1406, opened 2015-06-05
Fix offline training formula

WibbenZ:
It now uses three config integers to make it posible to up the tries gain.
Better then using the default skills since the formula will differ ALOT.
Closes #1396


ranisalt:
I'm not sure anyone will *ever* set those values different from the online training rate.

WibbenZ:
Well that is kinda the point, to have it as default.
I bet most "high exp" servers will tho.

The problem with the way you wanted to implement it is this:

"Normal" formula = 40% gain
Your formula = skillRate (lets say 10?) * 40% = 400%

The only way to make it work with your fix is to have the rates set to 1 and that would make it imposible to train up skills on an OT xD

This is IMO the best way to do it as of now.

ranisalt:
Not actually 400% total, but 4 times the needed tries for the current skill. That translates to about 3 skill levels.

The solution I purpose is the natural one. If you would train 100 tries online on Cipsoft Tibia, offline training should train 50 tries, since it's 1/2 of the speed (correct me if I'm wrong). If the skill rate is 20, then in the same time you would train 2000 tries on your server, and you should train 1000 tries on offline training.

You are proposing different skill rates for online and offline training, right?

WibbenZ:
Yes I am, because they differ in stages.
Have you even tested your fix?
As I explained in your PR is that you will give alot more tries then before, thats nothing to worry about in my PR since you can configure it yourself how much you want to give.

ranisalt:
Yes, I did, of course I didn't test it deeply on a running server but I slept for some minutes and increased the expected amount of tries.

I just argued that I don't think anyone would want to configure them separately instead of set up skill rate and work everywhere (equally). That's where both PRs differ and we ought to wait some more feedback since nobody besides us has commented on those requests :)

WibbenZ:
Ye well you get used to it xD
Seems to be rare that a PR is merged / closed even commented right away.

I do understand your point of it, but the way you did it will make it imposible to have a mid stage offline training and a high online training.
It will be simple to get ALOT of skills with the amount of tries that do get sent to the player when you just muliply it with the "normal" stages.
But I do agree, we can wait for Mark or Kornholi to respond.

PR #1409, opened 2015-06-07
Code style

WibbenZ:
Replaced all
````
"/>
```

with
````
" />
```


Crypton33:
It looks bad now :/ :-1:

WibbenZ:
Mark has approved prev. commits that ive posted that have contained the exact same thing.
It is the way the code style should be, you can look in (I think) globalevents.xml and you will see that both ways are used.
We should stick to one.

ranisalt:
This style is much better and almost every other XML/HTML file uses it.

WibbenZ:
Yeah I go crazy when I see tags that are missing spaces, same with if statments that are missing "{ }"...

marksamman:
Thanks!

PR #1410, opened 2015-06-08
Update outfits gender and boolean values

WibbenZ:
Swap "type" for "sex".
Use yes/no insted of 1/0 for boolean values.


ranisalt:
Much clearer to work on it now. Maybe a good addition is to use `female` and `male` instead of 0 and 1 throughout the code.

WibbenZ:
That is a valid point, ill do that aswell.

Issue #1394, opened 2015-06-01
XML tag for each flag


ranisalt:
I have read this proposal on [1.2 roadmap](https://github.com/otland/forgottenserver/wiki/Roadmap-1.2) but I couldn't quite understand why.

A numeric flag number can be arbitrary, and by itself has no meaning, and this is the single reason I thought on changing to more XML tags.

On the other hand, by using a flag tag for each possible flag, we would bloat the groups file a lot. We have currently 38 flags, and using the gamemaster group flag of `137438953471`, which has 37 flags set, would mean we need to add 37 new tags to the group, while the flag value is just an attribute.

This would hardly be pleasing to do by hand, and probably someone would create a tool for that. But what's the point, since [we already](http://comedinha.github.io/flags-calculator/) [got a ton of](http://opentibia.net/page/flagcalculator.html) [flag calculators](https://ranisalt.github.io/flags-calculator/)?

It also adds complexity since it would be necessary to parse the XML instead of just reading the int directly.
##

marksamman:
The point is that you shouldn't need to use a flag calculator to update group flags.

ranisalt:
I doubt anyone would do it by hand. It's 37 flags. Maybe if we keep both options?

Codinablack:
We could make predefined groups with variables to make it smaller and simpler for everyone if we did do it both ways...

ranisalt:
No... that's a very bad option actually IMHO. Groups are already predefined flags, creating predefined flag groups would create predefineception.

Kamenuvol:
Make a list of flags and put the group id which the flag is active
Example:
<flag id = "3" name = "canKillMonsters" groups = "1,2,3"/>

joseluis2g:
also something like this
```xml
<flag id="17" name="ignoredByMonsters">
    <group name="gamemaster"/>
    <group name="player"/>
    <monster name="demon"/>
</flag>
```
idk

WibbenZ:
Why not just make it simple insted of first listing all the groups then list what flag "array" it should have.

````
<group id="2" name="gamemaster" access="1" maxdepotitems="0" maxvipentries="200">
	<flag name="canKillMonsters" value="1" />
	<flag name="ignoredByMonsters" value="1" />
</group>
```

This way we can easily see what flag we have assigned (no flag counter is needed) aswell as with ease disable and enable the flags.

kornholi:
That's what we were planning to do.

ranisalt:
[...]

PR #1411, opened 2015-06-08
Update outfits gender and boolean values

WibbenZ:
Swap "type" for "sex".
Use yes/no insted of 1/0 for boolean values.
Use male/female insted of 1/0 for sex attributes.

### DIFF
--- a/data/XML/outfits.xml
+++ b/data/XML/outfits.xml
@@ -1,94 +1,94 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <outfits>
 	<!-- Female outfits -->
-	<outfit type="0" looktype="136" name="Citizen" premium="0" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="137" name="Hunter" premium="0" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="138" name="Mage" premium="0" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="139" name="Knight" premium="0" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="140" name="Noblewoman" premium="1" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="141" name="Summoner" premium="1" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="142" name="Warrior" premium="1" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="147" name="Barbarian" premium="1" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="148" name="Druid" premium="1" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="149" name="Wizard" premium="1" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="150" name="Oriental" premium="1" unlocked="1" enabled="1" />
-	<outfit type="0" looktype="155" name="Pirate" premium="1" unlocked="0" enabled="1" />
-	<outfit type="0" looktype="156" name="Assassin" premium="1" unlocked="0" enabled="1" />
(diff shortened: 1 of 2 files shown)


EvulMastah:
Why not true/false for boolean values? You changed them anyway.
Why not keeping backward compatibility leaving type?

WibbenZ:
Does not really matter, we used yes/no in the config file and it seems abit more understandable when it comes to a string value(as in XML).

Type can be confusing to understand, while sex tells you right away what it means.
This is not a major change for others if they decide to upgrade, what could it take 5-15 sec?

type="0" => sex="male"
type="1" => sex="female"

If a person can't use the find and replace function with the code above then they will have alot more problems.
IMO the removeItem etc (move from items to weapons) was a bigger thing then this.

ranisalt:
`sed -i -e 's/type="0"/sex="female"/g' -i -e 's/type="1"/sex="male"/g' -i -e 's/"0"/"no"/g' -i -e 's/"1"/"yes"/g' path/to/outfits.xml` and you're done.

marksamman:
This change is not worth breaking backwards compatibility.

WibbenZ:
So should I change it back to type, use 1/0 for male/female?

Still if you ask me we should really stop with keeping compatibility, those who want to move to 1.x have already done it and most of them have adapted the new script interface.
All this does is limit us to what we can do really.

Same with the weapons / items change you did, was that a thing worth doing?
Did it really optimize it to the point that users will wanna change the custom items they might have etc.
As I said, if users want to move to 1.x they have already done it and the most used server right now is probbly ORTS and they keep the datapack updated with TFS.

kornholi:
We break compatibility in cases where it can be justified. I don’t see how this “limits us to what we can do"

WibbenZ:
How does it not limit us to what we can do?
Like in this case, how do you justify calling type when you are returning a sex id?
It would be alot simpler to just call it sex or gender and then return male / female.

ranisalt:
How come this change is not worth it, but modifying the XML flags is?

djarek:
Perhaps we could try doing a gradual change, for example we could print out a big scary deprecation warning if the old syntax is used.

WibbenZ:
Implement what I did and start it with type and 1/0, you will get alot of errors haha, so ye.
But I guess the problem is that they are afraid people won't continue with 1.x while I think it's the other way around.
I still today don't get why we keep things like doPlayerAddItem when we can force users to use Player:addItem etc.
There are alot of functions like that, just go to luascript.cpp and serach for "lua_register"

ranisalt:
I think a major change in XML structure should be scheduled for a 2.0 version, with a big red NO BACKWARDS COMPATIBILITY warning.

WibbenZ:
Will probbly take a long while till we are at 2.0 xD
But maybe 1.3 / 1.4.
Atleast change out names like this that makes no sence at all without looking in the source code.

WibbenZ:
So nothing more on this Mark?
What is the point of even keeping backwards compat?
We can insted try to make it better, ex. With the code style.

ranisalt:
I don't think it was closed with compatibility in mind, else it wouldn't be a milestone to [change the flag representation](https://github.com/otland/forgottenserver/issues/1394).

marksamman:
I disagree about it being better, it's a step towards restricting the gender system to be binary and hard coding gender names. It's slower. It doesn't maintain backwards compatibility (and even if it did, it would just make the code messy).

djarek:
To be honest the name should probably be changed from type to sex(a little more intuitive ;)).

WibbenZ:
Same with using integers as bools if you ask me.
The only changes except sex was femake/male since I hate having to check if 0 or 1 is the male or female.

PR #1413, opened 2015-06-09
Moved offline training snippet (from Player::onCreatureAppear) to Lua

ninjalulz:
Moved offline training snippet (from Player::onCreatureAppear) to Lua (Makes it easier to customizable the code).
Moved doPlayerSetOfflineTrainingSkill to compat.lua.
New functions:
* Player.addOfflineTrainingTime(addTime)
* Player.getOfflineTrainingTime(self)
* Player.removeOfflineTrainingTime(removeTime)
* Player.addOfflineTrainingTries(skillType, tries)
* Player.getOfflineTrainingSkill(self)
* Player.setOfflineTrainingSkill(skillId)

Edit:
* Fixed an issue with the absence text (there was an extra space infront of 1 hour/minute), and changed to string.format.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1012,9 +1012,6 @@ void LuaScriptInterface::registerFunctions()
 	//isInWar(cid, target)
 	lua_register(m_luaState, "isInWar", LuaScriptInterface::luaIsInWar);
 
-	//doPlayerSetOfflineTrainingSkill(cid, skill)
-	lua_register(m_luaState, "doPlayerSetOfflineTrainingSkill", LuaScriptInterface::luaDoPlayerSetOfflineTrainingSkill);
-
 	//getWaypointPosition(name)
 	lua_register(m_luaState, "getWaypointPositionByName", LuaScriptInterface::luaGetWaypointPositionByName);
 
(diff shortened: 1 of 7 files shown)


WibbenZ:
Why not use events insted of creaturescripts?

ninjalulz:
@WibbenZ I didn't have that in mind, but it can easily be arranged if it's preferable.

WibbenZ:
Well im not 100% but didn't Mark or Dalkon say events are only executed when they need to, while creaturescripts are always executed or something like that?

ranisalt:
Great addition!

Luckey1729:
:+1: :+1: 

Thanks for the addition!

Kamenuvol:
@WibbenZ
I think it's the opposite, events are always executed and creaturescript only when is needed.

WibbenZ:
Now that I think that sounds wrong, events got added in the 1.x series while creaturescripts have been used for how long?
Not saying that creaturescripts are bad but why add a new system that isen't better then the old one?

Maybe Mark or Dalkon(who created it) can give us an input :p

Nu77:
I think events was added for move core code to scripting code, anyway i think creatureevents is a deprecated way which didn't should be used.

WibbenZ:
@ninjalulz 
https://github.com/ninjalulz/forgottenserver/blob/OfflineTraining/data/actions/scripts/other/skilltrainer.lua

src/luascript.cpp
```diff
@@ -8306,6 +8297,89 @@ int LuaScriptInterface::luaPlayerAddSkillTries(lua_State* L)
...
+}
+
+int LuaScriptInterface::luaPlayerAddOfflineTrainingTries(lua_State* L)
+{
+	// player:addOfflineTrainingTries(skillType, tries)
+	Player* player = getUserdata<Player>(L, 1);
+	if (player) {
+		skills_t skillType = getNumber<skills_t>(L, 2);
+		uint64_t tries = getNumber<uint64_t>(L, 3);
+		player->addOfflineTrainingTries(skillType, tries);
```
marksamman: This branch must push something to the Lua stack or return 0. It causes undefined behavior to not return the same number as objects you're pushing to the Lua stack (usually causes LuaJIT to crash).

marksamman: Looks like it should be:
```cpp
pushBoolean(L, player->addOfflineTrainingTries(skillType, tries));
```

ninjalulz: Thanks, I totally missed that! Fixed.

marksamman:
Thanks!

PR #1422, opened 2015-06-15
Fix code style

WibbenZ:
Replaces spaces with tabs
Remove trailing
Ref: a222c21


ranisalt:
Could be checked before the merge x.x

WibbenZ:
Really should, must be some addon like Travis that checks indentation etc.

PR #1424, opened 2015-06-16
Set NPCs to idle state when no player is in range

ranisalt:
NPCs now idle when no player is near (using `player->canSee(NPC position)`). Idle status callbak is only fired once per update. Updates spectators even if player logs in/out near the NPC.

### DIFF
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -97,6 +97,7 @@ bool Npc::load()
 
 void Npc::reset()
 {
+	isIdle = true;
 	loaded = false;
 	walkTicks = 1500;
 	floorChange = false;
@@ -110,6 +111,7 @@ void Npc::reset()
 
 	parameters.clear();
 	shopPlayerSet.clear();
+	spectators.clear();
 }
 
 void Npc::reload()
@@ -251,10 +253,13 @@ void Npc::onCreatureAppear(Creature* creature, bool isLogin)
 		if (npcEventHandler) {
 			npcEventHandler->onCreatureAppear(creature);
 		}
-	} else if (creature->getPlayer()) {
+	} else if (Player* player = creature->getPlayer()) {
 		if (npcEventHandler) {
 			npcEventHandler->onCreatureAppear(creature);
 		}
+
+		spectators.insert(player);
+		updateIdleStatus();
 	}
 }
 
@@ -267,10 +272,13 @@ void Npc::onRemoveCreature(Creature* creature, bool isLogout)
 		if (npcEventHandler) {
 			npcEventHandler->onCreatureDisappear(creature);
 		}
-	} else if (creature->getPlayer()) {
+	} else if (Player* player = creature->getPlayer()) {
 		if (npcEventHandler) {
 			npcEventHandler->onCreatureDisappear(creature);
 		}
+
+		spectators.erase(player);
+		updateIdleStatus();
 	}
 }
 
@@ -282,6 +290,19 @@ void Npc::onCreatureMove(Creature* creature, const Tile* newTile, const Position
 	if (creature == this || creature->getPlayer()) {
 		if (npcEventHandler) {
 			npcEventHandler->onCreatureMove(creature, oldPos, newPos);
+		}
+
+		if (creature != this) {
+			Player* player = creature->getPlayer();
+
+			// if player is now in range, add to spectators list, otherwise erase
+			if (player->canSee(position)) {
+				spectators.insert(player);
+			} else {
+				spectators.erase(player);
+			}
+
+			updateIdleStatus();
 		}
 	}
 }
@@ -316,7 +337,7 @@ void Npc::onThink(uint32_t interval)
 		npcEventHandler->onThink();
 	}
 
-	if (getTimeSinceLastMove() >= walkTicks) {
+	if (!isIdle && getTimeSinceLastMove() >= walkTicks) {
 		addEventWalk();
 	}
 }
@@ -381,6 +402,30 @@ bool Npc::getNextStep(Direction& dir, uint32_t& flags)
 	}
 
 	return getRandomStep(dir);
+}
+
+void Npc::setIdle(bool idle)
+{
+	if (isRemoved() || getHealth() <= 0) {
+		return;
+	}
+
+	isIdle = idle;
+
+	if (!isIdle) {
+		g_game.addCreatureCheck(this);
+	} else {
+		onIdleStatus();
+		Game::removeCreatureCheck(this);
+	}
+}
+
+void Npc::updateIdleStatus()
+{
+	bool status = spectators.empty();
+	if (status != isIdle) {
+		setIdle(status);
+	}
 }
 
 bool Npc::canWalkTo(const Position& fromPos, Direction dir) const
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -412,11 +412,8 @@ void Npc::setIdle(bool idle)
 
 	isIdle = idle;
 
-	if (!isIdle) {
-		g_game.addCreatureCheck(this);
-	} else {
+	if (isIdle) {
 		onIdleStatus();
-		Game::removeCreatureCheck(this);
 	}
 }
 
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -97,7 +97,6 @@ bool Npc::load()
 
 void Npc::reset()
 {
-	isIdle = true;
 	loaded = false;
 	walkTicks = 1500;
 	floorChange = false;
@@ -111,7 +110,6 @@ void Npc::reset()
 
 	parameters.clear();
 	shopPlayerSet.clear();
-	spectators.clear();
 }
 
 void Npc::reload()
--- a/src/npc.h
+++ b/src/npc.h
@@ -24,7 +24,6 @@
 #include "luascript.h"
 
 #include <set>
-#include <unordered_set>
 
 class Npc;
 class Player;
@@ -225,7 +224,7 @@ class Npc final : public Creature
 		std::map<std::string, std::string> parameters;
 
 		std::set<Player*> shopPlayerSet;
-		std::unordered_set<Player*> spectators;
+		std::set<Player*> spectators;
 
 		std::string name;
 		std::string filename;
--- a/src/npc.h
+++ b/src/npc.h
@@ -24,6 +24,7 @@
 #include "luascript.h"
 
 #include <set>
+#include <unordered_set>
 
 class Npc;
 class Player;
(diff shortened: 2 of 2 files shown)


src/npc.cpp
```diff
@@ -394,6 +417,30 @@ bool Npc::getNextStep(Direction& dir, uint32_t& flags)
 	return getRandomStep(dir);
 }
 
+void Npc::setIdle(bool _idle)
```
WibbenZ: https://github.com/otland/forgottenserver/wiki/Roadmap-1.2
```
Rename functions and variables that begin with underscore as they are reserved
```

ranisalt:
Could easily rename it :P I will do it when I get home.

src/npc.cpp
ranisalt: Done!

WibbenZ: (Y)

src/npc.h
```diff
@@ -221,6 +227,7 @@ class Npc final : public Creature
 		std::map<std::string, std::string> m_parameters;
 
 		std::set<Player*> shopPlayerSet;
+		std::set<Player*> spectators;
```
djarek: What is the reason for keeping a set of spectators? You might as well use a simple counter.

WibbenZ:
BTW have you tested the code?
Just curious but this must decrease the resources the server uses drastically(ex check an rl map).

src/npc.h
ranisalt: Because I couldn't find a reliable way to handle players both appearing, disappearing, logging in and out near the NPC.

If I replace all inserts with an increment on a counter and all erases with a decrement, it sometimes, but not always, gets double effect and the NPC gets to zero spectators when someone is nearby or -1 when someone goes away.

A set of pointers does not add so much overhead while keeping a natural O(1) check for duplicate entries.

ranisalt:
@WibbenZ I took my time to measure using ORTS, just to make sure how would be the behavior on a huge map with lots (786) of NPCs.

Using ORTS, with 786 NPCs

*upstream*:
CPU:
- 15 to 20% with 1 logged in player (god) far away from NPCs
MEM: constantly around 2800MB

*npc-idling*:
CPU:
- 7.5% to 10% with 1 logged in player (god) far
- up to 13.5% near 25 quiet (Dove) and 15 yelling (Dixi) NPCs
MEM: constantly around 2800MB

I guess that's a pretty good reduction in CPU usage for a giant server where most NPCs are sparse, far away from cities and just a few NPCs get attention all the time.

WibbenZ:
Id say those are some good results :p

src/npc.h
djarek: Fair enough, it would actually be hard to do that properly otherwise. However, the complexity of a find in std::set is O(logn). In this case an unordered_set __might__ be a better choice.

HeavenIsLost:
As far i can tell it is implemented good :) As you said servers with huge number of npcs will get a better performance with this :)

src/npc.cpp
```diff
@@ -394,6 +417,30 @@ bool Npc::getNextStep(Direction& dir, uint32_t& flags)
...
+		g_game.addCreatureCheck(this);
+	} else {
+		onIdleStatus();
+		Game::removeCreatureCheck(this);
+	}
+}
+
+void Npc::updateIdleStatus()
+{
+	bool status = spectators.size() == 0;
```
djarek: Use spectators.empty() for better readability.

WibbenZ: IMO checking size is better, if you read it you will expect something like "did it clean the list or not" since it's a boolean we are taking.

djarek: spectators.empty() is equivalent to spectators.size() == 0. The difference is that the former is more idiomatic when it comes to testing the emptiness of a container.

src/npc.h
ranisalt: It actually depends on implementation, but I think generally set is implemented with a RB tree and you are right. I will change it to an unordered_set as soon as I get home.

ranisalt:
Switched from set to unordered_set for lower insertion and removal complexity, and replaced `size() == 0` with `empty()` which calls the former ipsis litteris but with better readability.
Thanks @djarek for the suggestions.

src/npc.cpp
```diff
@@ -327,8 +348,10 @@ void Npc::onThink(uint32_t interval)
 		m_npcEventHandler->onThink();
 	}
 
-	if (getTimeSinceLastMove() >= walkTicks) {
-		addEventWalk();
+	if (!isIdle) {
+		if (getTimeSinceLastMove() >= walkTicks) {
```
djarek: is there a reason here for using nested ifs? If not, then move the second conditions into the first if statement.

ranisalt: No, not really. I just surrounded the previous code with an if. I might as well concatenate the condition checks.

HeavenIsLost:
@marksamman  what is your opinion on this?

src/npc.cpp
```diff
@@ -294,6 +302,19 @@ void Npc::onCreatureMove(Creature* creature, const Tile* newTile, const Position
 		if (m_npcEventHandler) {
 			m_npcEventHandler->onCreatureMove(creature, oldPos, newPos);
 		}
+
+		if (creature != this) {
+			Player* player = creature->getPlayer();
+
+			// if player is now in range, add to spectators list, otherwise erase
+			if (player->canSee(_position)) {
```
gugahoa: Isn't this a possible seg fault, as player can be nullptr is creature is not player?

HeavenIsLost: Yeah, clearly a segment fault.

ranisalt: Seriously, guys? The diff only shows part of the code. The line just above the diff is:
Imgur

So, if `creature != this`, then creature IS player.

gugahoa: It can be a monster, or onCreatureMove will never trigger if a monster pass by a NPC?

ranisalt: I think any creature will trigger the function. I'm sure at least other NPCs trigger the onCreatureMove callback.

gugahoa: Then it can give seg fault, as a monster can pass by a NPC and creature->getPlayer() will return nullptr

ranisalt: `creature` is either `this` or a player pointer != nullptr. I only use the return of `getPlayer()` when `creature` is not `this`, so it is a player pointer. I can't see where it will ever be nullptr.

gugahoa: Nvm, my bad, I didn't read the if before creature != this

ranisalt:
So how's it going?

marksamman:
@ranisalt waiting for you to respond to #1344

ranisalt:
Rebased against master

HeavenIsLost:
@marksamman read to be merged?

ranisalt:
@HeavenIsLost apparently not

ninjalulz:
Seems nice!

However, there is one <i>minor</i> issue. This would cause NPCs to not release focus on the last player in range. Like I said, it's a minor issue but it's an issue nonetheless.

gugahoa:
@marksamman is there anything that need more work here?
This feature would be really appreciated in a lot of servers.

ranisalt:
I think this creates another issue, actually. Tibia does not have "idle" NPCs (per se) anymore, they rather still do stuff when nobody is near (such as random yell). This PR would prevent NPCs from yelling randomly, for example, and not match Tibia behavior.

gugahoa:
You can make it go idle only if there's no player in yell range, if the NPC does yell.

ranisalt:
Then there's no advantage, since for them to yell means the server has to check every one, generate a random number, yell and propagate to every player on range... you get the point.

marksamman:
The NPC should still yell and call onThink (so it should still be processed in checkCreatures loop). It just shouldn't process selfSay calls or walk if it's "idle" (meaning there's no spectator to see it, so we can avoid calling getSpectators). The point of this is primarily to decrease CPU usage for idle servers, but it'll also help for other servers if they have lots of NPCs where players are rarely around. It's unrelated to replicating the behavior of another game.

HeavenIsLost:
I have implemented this PR into my own distro but for now the idle status only prevent the npc from walking. There is a issue when reloading the npcs: reset will clear the spectators set and isIdle will be set true, if there is any player in the npc screen the npc will just become idle(player need to move to make the npc not idle).

ranisalt:
When is a NPC reset?

HeavenIsLost:
On Npc::load() and Npc::reload() - which is called by Npcs::reload().

HeavenIsLost:
I think a fix can be just moving "spectators.clear()" and "isIdle = true" from npc::reset() to npc constructor "Npc::Npc(const std::string& _name)".

ranisalt:
Both aren't needed on the constructor as they are default initialized.

vankk:
@ranisalt 
There is a problem with this PR is that after you leave the NPC it doesn't close the shop window, seems that where onThink no longer runs once the NPC becomes idle, hence why it doesn't close the shop window when the player is no longer in screen.

vankk:
Any progress on this PR @ranisalt?

ranisalt:
Well I'm going to try something now, but I halted development because it was rather hard to set NPC to idle and keep it yelling.

ranisalt:
@vankk new commits should fix the issue you reported and allow NPCs to keep yelling while idle, can you check?

vankk:
Not sure how I can test it @ranisalt, also it is a quite hard to receive the trade window after the farewall, it does not happens everytime, yesterday for example, I maded several tests and I didn't receive. I also added the new commits on Aura to the players test it too.

src/npc.h
```diff
@@ -24,6 +24,7 @@
 #include "luascript.h"
 
 #include <set>
+#include <unordered_set>
```
ninjalulz: No longer needed :p

ninjalulz:
I ran a quick test, and the issues mentioned above didn't occur.

ranisalt:
@ninjalulz thanks!

I'll still look for more stuff that can be disabled while NPCs are idling.

ranisalt:
Well, I couldn't. The NPCs do not try to walk while idle anymore, but they still fire `onThink` events. The roadmap states "there is no reason for NPCs to walk or talk if no one is there to see it" but there is reason for them to talk, as they randomly yell.

Usage reduction is not as strong as I stated before after the changes, but there is indeed improvement.

marksamman:
onThink should still be fired, and there is reason for the NPC to still yell since there's a chance that someone can see it. This should still make a significant difference for idle/baseline CPU usage since it would reduce the getSpectators calls, but it's definitely not a high priority issue.

ranisalt:
I guess that's the best that can be done by now, then.

ranisalt:
@vankk confirmed the issue he reported is fixed now.

PR #1428, opened 2015-06-23
Fix a bug in droploot.lua

ninjalulz:
Fixes a bug where the server would print out an error message if a player died by a non-existent killer (e.g magic field with no caster).


marksamman:
Thanks!

PR #1442, opened 2015-07-08
Reward Chest & Boss Reward

MaxyTibia:
All credits to cbrm (otland), I have create this pull request to others
can help fix it.

Features:
* Custom depot as reward chest to store the unclaimed loot of bosses.
* Individual reward for each player that hits the boss.
* Loot chance determined by % of damage dealt to boss.
* Player may "pick up" his reward container directly from the corpse.
* Unclaimed reward is sent to online and offline participant players when
corpse rots.
* Unclaimed loot containers get erased after seven days of being stored.
* Only items may be taken out from reward chest, cannot put items in
there.
* Players' reward containers are stored as encoded, serialized tables on
database.
* Exhaustion limit to open reward chest, in order to avoid the spam of
sql queries.

To-Do:
* Damage taken by the boss leads to a % of loot drop.
* Healers' amount of health given to boss slayers leads to a % of loot
drop.
* Open reward chest as an inner depot.
* Set the corpses of the bosses as virtual and temporary depots to make
it more realistic.


WibbenZ:
This will still not be approved as I explained in the last pull, read the last comments I did and fix them.
Mainly indentation and stuff like that :p

MaxyTibia:
@WibbenZ,
As you can see, I applied many of his suggestions, but not all of I understood.
Could you help me, creating a pr on my page?

My intention is to leave this functional system for all users.

WibbenZ:
Not home atm, but check tabs etc to start with, skip the rest for now.

ranisalt:
Why don't you use an editor that helps with indentation?

I suggest you use at least Sublime with the [editorconfig plugin](https://github.com/sindresorhus/editorconfig-sublime), or vim with [editorconfig plugin](https://github.com/editorconfig/editorconfig-vim).

Since we have editorconfig support, indentation should be easy.

MaxyTibia:
Thanks @ranisalt, I will try

marksamman:
Please address these points before I can give this a serious review.
1) remove the base64 encoding of items and use proper item serialization which is implemented in the source code
2) don't use functions from our old Lua API, this includes both compat.lua and functions that don't belong to any metatable in luascript.cpp

schema.sql
```diff
@@ -301,6 +301,15 @@ CREATE TABLE IF NOT EXISTS `player_items` (
   KEY `sid` (`sid`)
 ) ENGINE=InnoDB;
 
+CREATE TABLE IF NOT EXISTS `player_rewardchest` (
```
comedinha: migrations?

ranisalt:
How does this work? Can I see the same loot from the dead corpse and the reward chest?

MaxyTibia:
@ranisalt look at the details:
http://www.tibia.com/news/?subtopic=newsarchive&id=2486

The loot is split according to the participation of players, and is available both in creation or in the chest at adventure's room.

Can you help with the upgrades requested by mark?

edumntg:
I've made a system like this one, but works as an inbox. The good thing from my system is that i added the 'getHealMap' function so you can give rewards to healers

MaxyTibia:
@edumntg you could share something with us?

edumntg:
@MaxyTibia Check out my repository, there's a branch named rewardchest

ranisalt:
I've been trying that too, while using a different class to handle both containers and chests. I stopped completely when I started the Lua part, as I do now know anything about Lua.

PR #1443, opened 2015-07-10
Minor changes to spells

raymondtfr:
Removed some unnecessary parenthesis, empty lines; added missing combat params that I forgot, such as "BLOCK_ARMOR, USE_CHARGES" on some physical based spells; changed "blood rage" spell effect to the right one, which is green.

I removed some unused parameters too, such as "factor" in some physical based spells, cause in some cases userdata weren't needed, or to be used. In some cases, like on spell "apprentice's strike" I removed all parameters (including cid), though I don't know if it is a proper change (is it correct to remove unused parameters from a function in Lua?)

I have also checked if Gorak has made any new updates to tibia-stats spell damage calculator, though it hasn't change (I have thoroughly checked), so the spells are still up-to-date according to ts.


data/spells/scripts/attack/avalanche.lua
```diff
@@ -7,9 +7,9 @@ local area = createCombatArea(AREA_CIRCLE3X3)
 setCombatArea(combat, area)
 
 function onGetFormulaValues(cid, level, maglevel)
-	min = -((level / 5) + (maglevel * 1.2) + 7)
-	max = -((level / 5) + (maglevel * 2.85) + 16)
-	return min, max
+	min = (level / 5) + ((maglevel * 1.2) + 7)
+	max = (level / 5) + ((maglevel * 2.85) + 16)
+	return -min, -max
 end
```
ranisalt: `(level / 5) + (maglevel * 2.85) + 16`
You can do it in multiple places, too

marksamman:
Add back factor, even if it's unused, it's good to know that it's a variable you can use in the formula. Please also rename cid to player while you're at it.

data/spells/scripts/attack/avalanche.lua
raymondtfr: Oh, yeah, I'm aware of that, I just thought of a clearer code instead of a shorter one (especially to emphasize the level quotient (or power "(level/5)") of the spell damage.

ranisalt: Oh yeah, I do prefer the more parenthesized way, it makes much clearer to see where each factor influences. I think the first form I suggested is good because it splits the level and the magic level part out of the minimum damage value :P

raymondtfr:
Hey Mark, in some spells such as "apprentices strike, practice healing, etc" the expression (level/5) won't never be used. So it is okay if, in this case I left the parameter list empty?

raymondtfr:
Should I make another pull and close this one?

marksamman:
You should include all parameters anyway, even if they're not used. No, just commit to the same branch and it will appear in this PR, and when it's good enough to be merged you should squash your commits.

raymondtfr:
jeez! I have finally made it haha xD.

WibbenZ:
Gj updating to meta functions :) (y)

marksamman:
Great, thanks!

PR #1447, opened 2015-07-11
Fix #1436 - guild rank/level bug

### LINKED ISSUE #1436: Guild Rank bug when have more than 3 different ranks
I noticed that I have a bug on my guild rank, what happens:

If I have more than 1 guild rank (names) for the same level access, it will show on your player the first created rank, for instance:

ID ---- Guild ID ---------- RANK ---------- LEVEL
1 ----- 7 ------------------- 3 -------------- Owner
2 ----- 7 ------------------- 2 -------------- Sub-owner
3 ----- 7 ------------------- 1 -------------- Sorcerer rank
4 ----- 7 ------------------- 1 -------------- Paladin rank
5 ----- 7 ------------------- 1 -------------- Druid rank
6 ----- 7 ------------------- 1 -------------- Knight rank

What happens is that all players in ranks after "Sorcerer rank" will appear ingame as "Sorcerer Rank".
Someone have noticed it? There is any fix for that?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23421493-guild-rank-bug-when-have-more-than-3-different-ranks?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ranisalt:
Currently, a guild is limited to only 3 ranks, one for leaders, one for vice-leaders and another for members. If you try to create more member levels, you end up with the description on look showing you the first rank created, not the actual one you are using (please see #1436 for further details).

This fixes the problem by the player holding its guild rank id instead of level. The only drawback is that the `Player:setGuildLevel` function is not reliable anymore, but it is rather useless when using an external guild management such as an AAC.

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -373,7 +373,7 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 				g_game.addGuild(guild);
 
 				query.str(std::string());
-				query << "SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `guild_id` = " << guildId << " LIMIT 3";
+				query << "SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `guild_id` = " << guildId;
 
 				if ((result = db->storeQuery(query.str()))) {
 					do {
@@ -385,12 +385,22 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 
 		if (guild) {
 			player->guild = guild;
-			GuildRank* rank = guild->getRankById(playerRankId);
-			if (rank) {
-				player->guildLevel = rank->level;
-			} else {
-				player->guildLevel = 1;
+			const GuildRank* rank = guild->getRankById(playerRankId);
+			if (!rank) {
+				query.str(std::string());
+				query << "SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `id` = " << playerRankId;
+
+				if ((result = db->storeQuery(query.str()))) {
+					guild->addRank(result->getNumber<uint32_t>("id"), result->getString("name"), result->getNumber<uint16_t>("level"));
+				}
+
+				rank = guild->getRankById(playerRankId);
+				if (!rank) {
+					player->guild = nullptr;
+				}
 			}
+
+			player->guildRank = rank;
 
 			IOGuild::getWarList(guildId, player->guildWarList);
 
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -8130,16 +8130,19 @@ int LuaScriptInterface::luaPlayerSetGuildLevel(lua_State* L)
 	// player:setGuildLevel(level)
 	uint8_t level = getNumber<uint8_t>(L, 2);
 	Player* player = getUserdata<Player>(L, 1);
-	if (player) {
-		if (const Guild* guild = player->getGuild()) {
-			if (const GuildRank* rank = guild->getRankByLevel(level)) {
-				player->setGuildRank(rank);
-				pushBoolean(L, true);
-				return 1;
-			}
-		}
+	if (!player || !player->getGuild()) {
+		lua_pushnil(L);
+		return 1;
 	}
-	pushBoolean(L, false);
+
+	const GuildRank* rank = player->getGuild()->getRankByLevel(level);
+	if (!rank) {
+		pushBoolean(L, false);
+	} else {
+		player->setGuildRank(rank);
+		pushBoolean(L, true);
+	}
+
 	return 1;
 }
 
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -8117,8 +8117,8 @@ int LuaScriptInterface::luaPlayerGetGuildLevel(lua_State* L)
 {
 	// player:getGuildLevel()
 	Player* player = getUserdata<Player>(L, 1);
-	if (player) {
-		lua_pushnumber(L, player->getGuildLevel());
+	if (player && player->getGuild()) {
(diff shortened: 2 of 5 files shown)


src/player.h
```diff
@@ -237,11 +237,11 @@ class Player final : public Creature, public Cylinder
 			this->guild = guild;
 		}
 
-		uint8_t getGuildLevel() const {
-			return guildLevel;
+		GuildRank* getGuildRank() const {
+			return guild->getRankById(guildRank);
```
marksamman: Null pointer dereference if player has no guild.

src/luascript.cpp
```diff
@@ -8098,7 +8098,7 @@ int LuaScriptInterface::luaPlayerSetGuildLevel(lua_State* L)
 	uint8_t level = getNumber<uint8_t>(L, 2);
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
-		player->setGuildLevel(level);
+		player->setGuildRank(player->getGuild()->getRankByLevel(level));
```
marksamman: Null pointer dereference if player has no guild.

src/luascript.cpp
```diff
@@ -8085,7 +8085,7 @@ int LuaScriptInterface::luaPlayerGetGuildLevel(lua_State* L)
 	// player:getGuildLevel()
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
-		lua_pushnumber(L, player->getGuildLevel());
+		lua_pushnumber(L, player->getGuildRank()->level);
```
marksamman: Null pointer dereference if player has no guild.

ranisalt:
Fixed null pointer dereferences

marksamman:
Change guildRank in Player to a pointer to GuildRank.
Spell out GuildRank* instead of using auto.

src/luascript.cpp
```diff
@@ -8097,8 +8097,8 @@ int LuaScriptInterface::luaPlayerSetGuildLevel(lua_State* L)
 	// player:setGuildLevel(level)
 	uint8_t level = getNumber<uint8_t>(L, 2);
 	Player* player = getUserdata<Player>(L, 1);
-	if (player) {
-		player->setGuildLevel(level);
+	if (player && player->getGuild()) {
+		player->setGuildRank(player->getGuild()->getRankByLevel(level));
```
marksamman: You should push false to Lua if getRankByLevel returns null, and not set the guild rank in that case. It could cause a null pointer dereference in Player.getGuildLevel() if the player is in a guild but the rank is null.

src/iologindata.cpp
```diff
@@ -389,12 +389,7 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 
 		if (guild) {
 			player->guild = guild;
-			GuildRank* rank = guild->getRankById(playerRankId);
-			if (rank) {
-				player->guildLevel = rank->level;
-			} else {
-				player->guildLevel = 1;
-			}
+			player->guildRank = guild->getRankById(playerRankId);
```
marksamman: guildRank can be null if the rank id of this player was inserted to the database after the first time someone in the guild logged in. The rank should be added to the guild if it does not currently exist.

marksamman:
Please rebase off master and also update Player::setGuild()

ranisalt:
I think it's ok now. I'm not sure if I should check on `setGuildRank` if the rank exists in the player's guild or just let it be.

src/luascript.cpp
```diff
@@ -8090,11 +8090,15 @@ int LuaScriptInterface::luaPlayerSetGuildLevel(lua_State* L)
...
-		lua_pushnil(L);
+		if (const Guild* guild = player->getGuild()) {
+			if (const GuildRank* rank = guild->getRankByLevel(level)) {
+				player->setGuildRank(rank);
+				pushBoolean(L, true);
+				return 1;
+			}
+		}
 	}
+	pushBoolean(L, false);
```
marksamman: Please invert the if-statements to avoid deep indentation, and this should push nil if player is null.

src/iologindata.cpp
```diff
@@ -389,13 +389,20 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
...
+			const GuildRank* rank = guild->getRankById(playerRankId);
+			if (!rank) {
+				query.str(std::string());
+				query << "SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `id` = " << playerRankId;
+
+				if ((result = db->storeQuery(query.str()))) {
+					guild->addRank(result->getNumber<uint32_t>("id"), result->getString("name"), result->getNumber<uint16_t>("level"));
+				}
+
+				rank = guild->getRankById(playerRankId);
```
marksamman: This can be null if the row is deleted from the database between the time of selecting players rank id and querying rank information. In that case, the player should be loaded without a guild (guild as nullptr, no guild nick, no guild war list).

brunominervino:
@ranisalt, this changes are working 100%?

ranisalt:
@brunominervino yes, it works

brunominervino:
@ranisalt thank you so much, I will test in my project.

src/luascript.cpp
ranisalt: Fixed, I guess. Now pushes nil if either player is null or player guild is null, and pushes false if there's no rank available.

ranisalt:
Rebased and fixed.

marksamman:
Thanks!

PR #1452, opened 2015-07-15
Move bug report to Lua

ranisalt:
This was requested by **beastn** from OTLand.

Moving the bug report system to Lua allows for better control over bug reports. One can write to a file (as previously and defaulted by this PR), or maybe save to the database, and change who can report and add filters easily using Lua.

### DIFF
--- a/data/creaturescripts/scripts/report.lua
+++ b/data/creaturescripts/scripts/report.lua
@@ -1,0 +1,26 @@
+function onReport(player, message, position, category)
+	if player:getAccountType() == ACCOUNT_TYPE_NORMAL then
+		return false
+	end
+
+	local name, playerPos = player:getName(), player:getPosition()
+	local file = io.open("data/reports/" .. name .. " report.txt", "a")
+
+	if not file then
+		player:sendTextMessage(MESSAGE_EVENT_DEFAULT, "There was an error when processing your report, please contact a gamemaster.")
+		return true
+	end
+
+	io.output(file)
+	io.write("------------------------------\n")
+	io.write("Name: " .. name)
+	if category == BUG_CATEGORY_MAP then
+		io.write(" [Map position: " .. position.x .. ", " .. position.y .. ", " .. position.z .. "]")
+	end
+	io.write(" [Player position: " .. playerPos.x .. ", " .. playerPos.y .. ", " .. playerPos.z .. "]\n")
+	io.write("Comment: " .. message .. "\n")
+	io.close(file)
+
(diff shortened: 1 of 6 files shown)


WibbenZ:
Events (AFAIK) was created for things we move from C++, check if it's enabled if it's not then I woulf handle it in C++.
I don't see the point of having it in Lua tbh, a config value to save it as a file, query or something else sounds better.

Atleast till we have the new Lua system implemented to make it more clean. Otherwise we will have 1000 events and 1000 creaturscripts xD

ranisalt:
Well, that means you would need to recompile everytime you want to change the log format. Suppose you might want to check the player's last login, or its level, you just change the script and reload.

Don't want to handle repors? `return false`.

WibbenZ:
Yes well atleast in my position I don't see the point of loading a lua script for something like this.
But you should still use mods insted if you ask me.

yrpen:
I think it's a great idea. For example in this case I can forward player report to any program in my shell using os.execute() (email me, send it to Jabber, send to any CRM if you want).

ranisalt:
Maybe open issues directly on Github? I'm trying it now :)

wgrriffel:
@ranisalt im interested in how you did the json things from github in lua, please =)

vankk:
I think it should be saved in the database and not in a file and also should have a delay to send another report.

Tarliton:
@vankk if it's in lua you can do whatever you want

ranisalt:
@wgrriffel unfortunately I lost the code, but it wasn't so hard as the HTTP library I used had native support for JSON as Lua tables.

@vankk piece of cake :) I can make a setting on config.lua to toggle logging to database.

vankk:
@Tarliton I know that, I just said my opinion to make this PR better.
@ranisalt Seems good to me! 😄

ranisalt:
I'm merging as AFAIK there is no drawbacks and it strips away some legacy C code from the source, which is good.

WibbenZ:
Rani it should have been done in events imo.

ranisalt:
@WibbenZ ooops I'll change it ASAP

data/creaturescripts/scripts/report.lua
```diff
@@ -0,0 +1,26 @@
+function onReport(player, message, position, category)
+	if player:getAccountType() == ACCOUNT_TYPE_NORMAL then
+		return false
+	end
+
+	local name, playerPos = player:getName(), player:getPosition()
```
WibbenZ: Also in the update, split this variable, since if the file does not exist there is no reason to fetch the player position.
Also please use playerPosition or just position insted of playerPos :)

ranisalt: Thanks for the suggestions

PR #1464, opened 2015-07-19
Fixes #1462 Changing outfit while mounted to unregistered looktypes.

### LINKED ISSUE #1462: Changing outfits while mounted into a looktype that has no mount sprite crashes the client
How to reproduce:
Mount in any of the outfit except the Gamemaster one then change to gamemaster outfit and it crashes the client for you and people around.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24608366-changing-outfits-while-mounted-into-a-looktype-that-has-no-mount-sprite-crashes-the-client?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Xileck:
Webm of issue: 
http://i.imgur.com/rgPNvvG.webm

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3160,6 +3160,12 @@ void Game::playerChangeOutfit(uint32_t playerId, Outfit_t outfit)
 	}
 
 	Player* player = getPlayerByID(playerId);
+	
+	const Outfit* playerOutfit = Outfits::getInstance()->getOutfitByLookType(player->getSex(), outfit.lookType);
+	if (!playerOutfit) {
+		outfit.lookMount = 0;
+	}
+	
 	if (!player) {
 		return;
 	}
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3161,13 +3161,13 @@ void Game::playerChangeOutfit(uint32_t playerId, Outfit_t outfit)
 
 	Player* player = getPlayerByID(playerId);
 	
+	if (!player) {
+		return;
+	}
+	
 	const Outfit* playerOutfit = Outfits::getInstance()->getOutfitByLookType(player->getSex(), outfit.lookType);
 	if (!playerOutfit) {
 		outfit.lookMount = 0;
-	}
-	
-	if (!player) {
-		return;
 	}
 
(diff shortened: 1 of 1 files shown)


Xileck:
My bad then, maybe someone else can get into this.

WibbenZ:
This is not even a fix, this is a hack.
If you want it to work use one of these:
If the lookType is the GM outfit = remove mount (it will still be posible to crash the client aswell as players in the area by mounting = not a good ide).

So then we end up with the best way to do it, if the lookType is the GM outfit, allow it and remove the mount.
Also if the player tries to mount, return the function if the player has the GM outfit.

ranisalt:
@WibbenZ It's actually not only for the GM outfit, but for any other outfit not choosable by common players. That includes all GM outfits, elf, dwarf, orc, that generic colorable demon used for Morgaroth and Ghazbaran, just about anything.

WibbenZ:
Yeah that is true :p I was talking about the question in hand, but it should maybe check if we have it listed in outfits.xml? Maybe add another tag (true by default - since we don't want to help people "move forward"), if it's not there we can keep / set the mount.

ranisalt:
Yes, it checks. I have just compiled the latest commit and there is no problem having mount and changing to gamemaster outfit or changing before and mounting. Works properly with no debug.

ranisalt:
Could you please try that?
https://github.com/ranisalt/forgottenserver/tree/fix-mount-debug

For me it is working as expected. Changing to an outfit that is not on outfit.xml (that is, the mountable outfits) will dismount the player.

src/game.cpp
```diff
@@ -3160,9 +3160,15 @@ void Game::playerChangeOutfit(uint32_t playerId, Outfit_t outfit)
 	}
 
 	Player* player = getPlayerByID(playerId);
+
```
ranisalt: Remove this line please.

PR #1483, opened 2015-07-29
Fix #1482 house list edit

### LINKED ISSUE #1482: Bug house invite
https://www.youtube.com/watch?v=4Mwi0yf_DBM

It is possible to invite even outside the house so seen in the video.

ranisalt:
The bug: #1482 
The player gets kicked but the house list editing window keeps open and the player still can edit and save it once. That way, he can put itself on the list and walk into the house.

The fix: just unset the player house being edited value to null when he gets kicked from the house, and the server silently ignores the update.

Video: https://youtu.be/sgNRSPQIV94

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1480,6 +1480,11 @@ void Player::onCreatureMove(Creature* creature, const Tile* newTile, const Posit
 			}
 		}
 	}
+
+	// unset editing house
+	if (editHouse && !newTile->hasFlag(TILESTATE_HOUSE)) {
+		editHouse = nullptr;
+	}
 }
 
 //container


kornholi:
hmm, maybe a better solution would be to check access in playerUpdateHouseWindow?

ranisalt:
@kornholi to accomplish what, exactly? I went straight for the simpler solution.

I thought on checking if the player had access before setting the new list, but the player is outside of the house anyway and should not be allowed to do that.

walissonsantos:
thanks you

walissonsantos:
ranisalt, your solution works only when the player is kicked, if the player walk out of the house before being kicked he can auto invite. If the player uses the Magebot he can walk with the open list

ranisalt:
@walissonsantos I reworked on it and now it should reset whenever the player leaves the house, either being kicked, moving out or anything else.

walissonsantos:
Thank you, 100%

marksamman:
I agree with @kornholi that we should check for access when saving. Consider this case:
1) Player A is a subowner of Player B's house
2) Player A opens the edit window
3) Player B puts Player A on the invited list and removes Player A from the subowner list
4) Player A saves the edit window

vankk:
There is no news to fix this issue?

marksamman:
We decided to merge #1809 instead since this wasn't a proper solution, and no attempt was made at improving it. Thanks anyway.

PR #1486, opened 2015-08-01
Fix #1470

### LINKED ISSUE #1470: Status protocol does not close connection after checking status!
Every normal server after status data sends FIN,ACK packet (I check status with "0xFF + info" packet) .
Newest TFS keeps connection open forever.
This freez acc. maker that read from socket until max. execution time (PHP).

Only change I found is this:
https://github.com/otland/forgottenserver/commit/06c179406c6fdf595f10b325af042e55c5114bae#diff-575ee7a35edffe65602c2ce54df0c677L94

And many commits about Output changes [ getConnection->close() -> disconnect() ].

Is it planned and I should rewrite acc. maker to read until it find </ xml> or you are going to fix it?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24930147-status-protocol-does-not-close-connection-after-checking-status?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

djarek:
The socket will now be properly closed once all messages  have been sent.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -83,7 +83,7 @@ void Connection::close(bool force)
 	if (messageQueue.empty() || force) {
 		closeSocket();
 	} else {
-		//will be closed by the destructor
+		//will be closed by the destructor or onWriteOperation
 	}
 }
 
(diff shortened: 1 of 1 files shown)


marksamman:
Thanks!

PR #1487, opened 2015-08-01
Remove unused imports

ranisalt:
I found over 100 includes that were not being used or were already included by other includes. Probably there are no more unused imports :laughing:


WibbenZ:
Well that was not a bad find :P

djarek:
Out of curiosity - did you use any tools for this?

src/thing.cpp
```diff
@@ -17,8 +17,6 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-#include "otpch.h"
```
marksamman: otpch should always be included first.

ranisalt:
@djarek yes, I use the CLion IDE and it has a tool to check for unused imports

src/thing.cpp
ranisalt: You mean otpch should be included on all files?

WibbenZ: Yes atleast from my findings.
If you don't include otpch you will get an error about it not being included (if you use things like uintX_t etc).

ranisalt: But if it is included on at least most of the files, it will be indirectly included on all of them since headers depend on other headers.

marksamman: It should be included from every cpp file.

marksamman:
Thanks!

PR #1493, opened 2015-08-05
Spell mods

raymondtfr:
Added more meaning words for some spell variables; updated weapon name in lua file; now we can use level param in (CALLBACK_PARAM_SKILLVALUE) spell formulas.

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -935,7 +935,7 @@ void ValueCallback::getMinMaxValues(Player* player, CombatDamage& damage, bool u
 	int parameters = 1;
 	switch (type) {
 		case COMBAT_FORMULA_LEVELMAGIC: {
-			//onGetPlayerMinMaxValues(player, level, maglevel)
+			//onGetFormulaValues(player, level, maglevel)
 			lua_pushnumber(L, player->getLevel());
 			lua_pushnumber(L, player->getMagicLevel());
 			parameters += 2;
@@ -943,7 +943,7 @@ void ValueCallback::getMinMaxValues(Player* player, CombatDamage& damage, bool u
 		}
 
 		case COMBAT_FORMULA_SKILL: {
-			//onGetPlayerMinMaxValues(player, attackSkill, attackValue, attackFactor)
+			//onGetFormulaValues(player, level, attackSkill, attackValue, attackFactor)
 			Item* tool = player->getWeapon();
 			const Weapon* weapon = g_weapons->getWeapon(tool);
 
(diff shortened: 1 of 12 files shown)


src/combat.cpp
```diff
@@ -943,7 +943,7 @@ void ValueCallback::getMinMaxValues(Player* player, CombatDamage& damage, bool u
 		}
 
 		case COMBAT_FORMULA_SKILL: {
-			//onGetPlayerMinMaxValues(player, attackSkill, attackValue, attackFactor)
+			//onGetFormulaValues(player, level, attackSkill, attackValue, attackFactor)
```
luanlucian93: Because it no longer puts the levelPower instead of level?

player-> GetLevel () / 5?

raymondtfr: For customization purposes.

WibbenZ:
Why add this?
IMO it would be better to just call player:getLevel (as we do right now) since it will otherwise push the level to each spell, that (atleast now) in most cases isen't used.

raymondtfr:
I thought it would make things clearer while coding in lua. I did this since the key callback didn't have a level param, but magic level callback does. I'm not really sure what you mean "in most cases isn't used" I guess, but at current state "all" attack spells uses level param. It would also make code in lua look clearer and short.

WibbenZ:
In most cases = if you have 10 scripts but only 2 needs the level parameter, then in most cases it isen't needed.
Like now, only 12 files edited. Either you skipped alot of files or only 10?
Compare the spells you need to change to all spells we have, IMO it is not worth pushing arguments that we don't need in most files.

raymondtfr:
I guess your right. But only nine of all spells (those I changed "paladin and knights" spells) uses CALLBACK_PARAM_SKILLVALUE which is why I did that -- all other files uses CALLBACK_PARAM_LEVELMAGICVALUE. So, is it still a good move or should I close this?

WibbenZ:
Im not the owner here, so would be best to wait for Mark or Korn to respond.
But if you ask me, it - when you call the function another argument will be pushed, an argument that might not be used in all files = waste.
Better to call it when you actually need it.

Another thing is that you are breaking backwards compatibility by adding the argument in the middle, insted of pushing it last.
If you run an old script the weapon skill will be changed to magic skill etc etc

ranisalt:
I agree with @WibbenZ

raymondtfr:
Okay, I guess I'll be waiting for Mark final words. So to speak, then what is the usefulness of factor parameter since it is also not used in any of the spells, but still it is good to have such parameter so that anyone can customize (even if it is not used).

WibbenZ:
In sence they can be useless, but it depends on the majority of the spells.
There is no need to push the "item" argument to actions unless it is used in most scripts (as it is) - or that we can use something like Action:getItem() etc

So the only "useless" argument in spells atm is the skill argument, but once again to be able to keep backwards compatability we need to keep it.
(They did not even wanna change type to sex in outfits.xml, so I doubt they would allow this that would take ALOT longer to edit).

marksamman:
This breaks backwards compatibility and is already accessible from player:getLevel().

PR #1498, opened 2015-08-07
RSA class improvements.

### LINKED ISSUE #1494: RSA mutex
Is the mutex in the RSA class really necessary? It seems that it would only be required if we offered a way to change the RSA key at runtime, because according to gmp docs, all 3rd party calls we use there are thread safe. https://gmplib.org/manual/Reentrancy.html

djarek:
Closes #1494
Added the const qualifier to RSA::decrypt(). Removed m_ prefix from
variables. Removed the mutex because it was unnecessary (decryption has
no side-effects and doesn't modify any state).

### DIFF
--- a/src/rsa.h
+++ b/src/rsa.h
@@ -32,7 +32,7 @@ class RSA
 		RSA(const RSA&) = delete;
 		RSA& operator=(const RSA&) = delete;
 
-		void setKey(const char* p_str, const char* q_str);
+		void setKey(const char* pString, const char* qString);
 		void decrypt(char* msg) const;
 
 	private:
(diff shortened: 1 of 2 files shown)


src/rsa.h
```diff
@@ -32,14 +32,12 @@ class RSA
 		RSA(const RSA&) = delete;
 		RSA& operator=(const RSA&) = delete;
 
-		void setKey(const char* p, const char* q);
-		void decrypt(char* msg);
-
-	protected:
-		std::recursive_mutex lock;
+		void setKey(const char* p_str, const char* q_str);
```
marksamman: pString and qString

djarek:
Fixed.

marksamman:
Thanks!

PR #1501, opened 2015-08-08
Add spell methods to Lua

WibbenZ:
New class:

```
Spell(id)
Spell(name)
Spell(words)
```

New methods:

```
Spell:getName()
Spell:getWords()
Spell:getLevel()
Spell:getMagicLevel()
Spell:getManaCost()
Spell:getManaPercent()
Spell:getSoulCost()
Spell:isPremium()
Spell:isLearnable()

Player:getSpellCount()
Player:canCast(spell)
```

Moved getPlayerInstantSpellInfo to compat
Moved getPlayerInstantSpellCount to compat


djarek:
I think you should document in the code somewhere that a spell reload potentially invalidates Spell userdata and could cause Undefined Behavior (most likely crashes).

WibbenZ:
Is there a way to solve that? How is otherwise ItemType, MonsterType etc working?
Did I forget something?

djarek:
The same restrictions apply to all non-reference-counted objects. So storing a reference to, for example, a MonsterType and then doing a reload is dangerous. I just wanted there to be some trace of this behavior for "future generations". The general solutions are:
-Use garbage collection
-Use reference counting
-Use an opaque handle to the resource and obtain handles to it from a "resource manager"

The 1st solution is obviously out of the question (mostly because there is no safe way to extend the lua collector to the C++ side or to somehow "synchronize" the Boehm-Demers-Weiser GC with the one in lua). The second one causes a bit of overhead every time a reference is copied, while the third one incurs an overhead whenever the object is used (though the overhead is probably negligible compared to the delays of accessing L3 cache and main memory).

WibbenZ:
But where is the notes about ex. MonsterType located?
If you ask me it would be better to add all of these one file, that you can view here(wiki page) aswell as downloaded with the datapack.

We do how ever use the "warn unsafe scripts", that you atleast help abit.
Mainly since you can't push userdata values in addEvent functions.

Kamenuvol:
What if we make something like a pointer to pointer?
One reload and changes what these pointers are pointing to, and the pointers are still the same in Lua.
Is this ever possible?

djarek:
@Kamenuvol 
Unless reference counting is introduced your solution would introduce either dangling references or resource leaks.

Kamenuvol:
What if Spell metatable works only with a copy, when it reload the variable will stay with the same value. This metatable is only for info. If one wants them, he will call the function which gets a copy of the new info.

data/lib/compat/compat.lua
```diff
@@ -1004,3 +1004,34 @@ end
 function Guild.removeMember(self, player)
 	return player:getGuild() == self and player:setGuild(nil)
 end
+
+function getPlayerInstantSpellInfo(cid, spellId)
+	local player = Player(cid)
+	if player == nil then
```
ranisalt: `if not player` always seems cleaner to me.

data/lib/compat/compat.lua
```diff
@@ -1004,3 +1004,34 @@ end
...
+		return false
+	end
+
+	local spell = Spell(spellId)
+	if spell == nil or not player:canCast(spell) then
+		return false
+	end
+
+	local spellInfo = {
```
ranisalt: Just return it instead of assigning to variable.

data/lib/compat/compat.lua
```diff
@@ -1004,3 +1004,34 @@ end
...
+	return spellInfo
+end
+
+function getPlayerInstantSpellCount(cid)
+	local player = Player(cid)
+	if player == nil then
+		return false
+	end
+	return player:getSpellCount()
+end
```
ranisalt: Make it oneliner.

```lua
function getPlayerInstantSpellCount(cid) local p = Player(cid) return p ~= nil and player:getSpellCount() end
```

src/luascript.cpp
```diff
@@ -9173,6 +9144,31 @@ int LuaScriptInterface::luaPlayerGetContainerIndex(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaPlayerGetSpellCount(lua_State* L)
```
ranisalt: Please mind it's `getInstantSpellCount`, not only `getSpellCount`.

src/luascript.cpp
```diff
@@ -11994,6 +11990,141 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
 }
 
+// Spells
+int LuaScriptInterface::luaSpellCreate(lua_State* L)
+{
+	// Spell(words, name or id)
+	InstantSpell* spell;
+	if (isNumber(L, 2)) {
+		spell = g_spells->getInstantSpellById(getNumber<uint32_t>(L, 2));
+	} else if (isString(L, 2)) {
```
ranisalt: I think you don't need this, just treat the second argument as only number or string. If nil, `isString` will return and empty string that will not match any spell and skip.

src/luascript.cpp
```diff
@@ -11994,6 +11990,141 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+int LuaScriptInterface::luaSpellGetName(lua_State* L)
+{
+	// spell:getName()
+	InstantSpell* spell = getUserdata<InstantSpell>(L, 1);
+	if (spell) {
```
ranisalt: Move the variable to inside the if (`if (InstantSpell* spell = getUserdata<InstantSpell>(L, 1))`) for better scoping.

src/luascript.cpp
```diff
@@ -11994,6 +11990,141 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+int LuaScriptInterface::luaSpellGetWords(lua_State* L)
+{
+	// spell:getWords()
+	InstantSpell* spell = getUserdata<InstantSpell>(L, 1);
+	if (spell) {
```
ranisalt: Same as above.

src/spells.cpp
```diff
@@ -235,17 +235,15 @@ uint32_t Spells::getInstantSpellCount(const Player* player) const
...
 	uint32_t count = 0;
 	for (const auto& it : instants) {
 		InstantSpell* instantSpell = it.second;
-		if (instantSpell->canCast(player)) {
-			if (count == index) {
-				return instantSpell;
-			}
-			++count;
+		if (count == spellId) {
+			return instantSpell;
```
ranisalt: `return it.second` and you don't need to get it above.

src/spells.cpp
```diff
@@ -235,17 +235,15 @@ uint32_t Spells::getInstantSpellCount(const Player* player) const
...
 		}
+		++count;
 	}
 	return nullptr;
```
ranisalt: This function is now pretty ugly. Please make it like:

```cpp
auto it = std::advance(instants.begin(), index);
if (it != instants.end()) {
  return it->second;
}
return nullptr;
```

It's much shorter and cleaner

ranisalt:
This is promising.

data/lib/compat/compat.lua
WibbenZ: Ye ofc, been a while since I wrote it haha

data/lib/compat/compat.lua
WibbenZ: Did you mean spellInfo?

WibbenZ:
Sounds good ill make the updates tomorow :)

data/lib/compat/compat.lua
ranisalt: I meant return the table directly e.g. `return { stuff = 'in here' }` instead of assigning to `spellInfo` and `return spellInfo`

WibbenZ: Oh okay np then :p

Kamenuvol:
So, there is no chance to have dangling pointers like @djarek said?

WibbenZ:
There is, but we have the same "problem" with the rest of the classes.

ranisalt:
@Kamenuvol it still has the problems, but there are more places with potentially, though extremely rare, dangling pointers throughout the code. It would incur a HUGE refactoring of the codebase to have 100% safe code and I think it's easier to teach scripters not to do dumb stuff (such as storing userdata e.g. with `addEvent`).

WibbenZ:
GG.... ill open a new one insted xD

PR #1503, opened 2015-08-09
Change getPremiumDays to isPremium

WibbenZ:
This way it will also check if the server has "free premium" activated.
I have not changed skilltrainer.lua since Ninjalulz has it already changed in the skilltrainer to Lua PR.


marksamman:
Thanks!

PR #1507, opened 2015-08-10
Fix #1506 null pointer check on Item::getTile

### LINKED ISSUE #1506: Segmentation fault on Item::getTile()
I was playing with the field creation trying to help with #1500 and coded the snippet:

```
		Item* item = Item::CreateItem(itemId);
		if (caster) {
			item->setOwner(caster->getID());

			if (MagicField* field = item->getMagicField()) {
				if (Tile* tile = field->getTile()) {
					CreatureVector* creatures = tile->getCreatures();
```
But it gives me a segmentation fault on `field->getTile()`, more precisely on that if:
```
	Cylinder* cylinder = getTopParent();
	//get root cylinder
	if (cylinder->getParent()) {
```
The `getTopParent` function returns nil and it is not checked.

ranisalt:


ranisalt:
Reopened as #1660 may be related

vankk:
@ranisalt 
Could you explain theses changes? 

Thanks.

ranisalt:
```
  	Cylinder* cylinder = getTopParent();
```
May return `nullptr`, so any use of `cylinder` methods past this point potentially renders segmentation fault. I just check if `cylinder` is not a null pointer before using it.

marksamman:
If anyone can post steps to make the server crash because of this (it may include a custom Lua script that makes use of our Lua API), I'll merge the pull request. Until then, I'm not sure that it can actually happen, and that would make the check redundant, which is why this hasn't been merged yet.

diath:
I vaguely remember a very similar issue that I fixed basically the same way as in the PR. The server would crash from an onAddItem script that teleported items that were thrown onto a teleport. What was even weirder is that the script sent 2 magic effects, one before the item was teleported (using Item:moveTo()), another after the item was teleported at the target position and merely removing the Position:sendMagicEffect() call would make the issue disappear. I just tried to reproduce it against the master branch but I can't seem to get the server to crash so I'm not sure whether it's still an issue (that's an issue I had to deal with a few months ago). Thought I'd mention it for what it's worth.

ranisalt:
@marksamman on #1506 where I first spotted this exception there is example code on how to trigger it, I can try and make a patch if needed.

marksamman:
That was with source code modifications. I'm asking for a Lua script that will trigger this crash, as that's more reasonable to apply this patch against. @diath may be on to something though as the crash report in #1660 also shows onAddItem in the trace.

conde2:
My thought abou the problem: he moves the item to the teleport, but at the same time he cleans the map causing the item to be a nullptr and maybe crashing

conde2:
Looking at #1660 he is moving the item to a Dustbin so this COULD mean that the item is being deleted before this check

doTeleportThing(item.uid, {x = 32343, y = 32218, z = 7}) -- Dustbin local

conde2:
@marksamman 

I've managed to crash the server and the problems seems to be this null reference:

image

diath:
@conde2 
That's what the PR is addressing the problem here is that there's no way to reproduce the issue and find out what causes it, unless you found a way to reproduce it, in which case you should share the method.

conde2:
Yes I found a way to reproduce, just trying to put this in a small script since my script has like 500+ lines.

conde2:
How to reproduce:

I'm using this map to test the script
image

Create a script (talkaction works) to fill the map with containers, the first container will have an actionId:

        for x = arenaFromPosition.x, arenaToPosition.x do
            for y = arenaFromPosition.y, arenaToPosition.y do
                for z = arenaFromPosition.z, arenaToPosition.z do
                    local position = Position(x, y, z)
                    for _ = 1, math.random(1, 5) do
                        local item = Game.createItem(1739, 1, position)
                        if not done then
                            done = true
                            item:setAttribute(ITEM_ATTRIBUTE_ACTIONID, 8010)
                        end
                    end
                end
            end



Now create a action script with this:

    local arenaFromPosition = Position(208, 160, 15)
    local arenaToPosition = Position(224, 174, 15)

        for x = arenaFromPosition.x, arenaToPosition.x do
            for y = arenaFromPosition.y, arenaToPosition.y do
                for z = arenaFromPosition.z, arenaToPosition.z do
                    local tile = Tile(Position(x, y, z))
                    if tile then
                        local items = tile:getItems()
                        for i = 1, #items do
                            items[i]:remove()
                        end
                    end
                end
            end
        end 


First execute the talkaction and then open the container with the actionId, the server will crash

marksamman:
Thanks! That should be fixed in b5892ef5022a1080fa21ec1314e04130ae21de34. There may be other ways getTile() can be called with a null top parent, so I'm leaving this open.

ninjalulz:
You can trigger a crash by using an onRemoveItem script. Simply register the script to a teleport, place an item on top of it, and execute cleanMap(). The server no longer crash after applying the PR changes.

Sample script
```Lua
function onRemoveItem(moveitem, tileitem, position)
	moveitem:getPosition():sendMagicEffect(CONST_ME_MAGIC_RED)
	return true
end
```

PrinterLUA:
I've tested what Ninja provided and i got crash. It pointed to Item::getTile

marksamman:
onRemoveItem should be called before the item has been removed, which would also allow us to cancel the removal with return values, but this shows that you can probably trigger this crash in any script by just calling remove() and then getPosition() on the item, so merging this PR makes more sense now.

Thanks!

PR #1509, opened 2015-08-10
Use RAII to manage CombatParam && Combat memory.

djarek:
Use the RAII idiom to reduce the amount of memory management boilerplate.


marksamman:
Thanks!

PR #1469, opened 2015-07-21
Cleanup

WibbenZ:
Removed trailing spaces
Removed brackets around if statments
Removed extra spaces in spells.xml
Updated the last spells
Updated indentation of three monsters
Updated IP ban script
Updated some legacy functions to userdata
Updated some variable names
Updated mass healing (exura gran mas res) - it now only heals players
Replaced all table.remove and table.insert
Replaced some of the for loops
Cleaned some formulas
Cleaned most? spells
Fixed surpisebag.lua
Moved the last "unknown areas" to the lib file, insted of having huge duplicated arrays.
Added ALOT of missing newlines, most in monsters

PR #1510, opened 2015-08-11
Npcsystem changes

brunominervino:


WibbenZ:
Regarding the loops, #1469 
By Mark
```
Yes, using ipairs is slower, but not as slow as the page you're linking to suggests (it's 4 years old though, so maybe it was accurate that long ago). In my tests, it was 8% slower with LuaJIT, and 20% slower with Lua, but it's a matter of nanoseconds, so it doesn't matter, we go for readability in this case. It takes millions of iterations before it's a matter of milliseconds. Whatever you're going to do inside the loop is going to dominate the total time anyway. The last paragraph under "Other Considerations" on that page is on point.
```

brunominervino:
@WibbenZ, the loops were reverted.
Thank you :+1:

data/actions/scripts/quests/annihilator.lua
```diff
@@ -15,7 +15,7 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 	if item.itemid == 1945 then
 		local players = {}
 		for _, position in ipairs(playerPosition) do
-			local topPlayer = Tile(position):getTopCreature()
+			local topPlayer = position:getTile():getTopCreature()
```
WibbenZ: Revert this, no need to go through another function when you can just use Tile(position).

data/movements/scripts/tiles.lua
```diff
@@ -25,7 +25,7 @@ function onStepIn(creature, item, position, fromPosition)
 	if Tile(position):hasFlag(TILESTATE_PROTECTIONZONE) then
 		local lookPosition = player:getPosition()
 		lookPosition:getNextPosition(player:getDirection())
-		local depotItem = Tile(lookPosition):getItemByType(ITEM_TYPE_DEPOT)
+		local depotItem = lookPosition:getTile():getItemByType(ITEM_TYPE_DEPOT)
```
WibbenZ: Revert this, no need to go through another function when you can just use Tile(position).

WibbenZ:
FYI you have to remove them anyways since Mark removed that function in 4da1360db922fd8d7adce06300c266741e4fadfc

brunominervino:
@WibbenZ, done!
Thank you again :+1:

WibbenZ:
Good, you can most likely squash your commits and rename the PR to a more accurate name.

brunominervino:
Done :)

luanlucian93:
change too getItemName(itemid) to ItemType(itemId):getName(), please

brunominervino:
Thank you @luanlucian93

luanlucian93:
change too isItemFluidContainer(itemid) to ItemType(itemid):isFluidContainer()

PR #1511, opened 2015-08-11
add support for 10.80+

Milice:


ranisalt:
Boss!

But you're missing these:
https://github.com/otland/forgottenserver/blob/master/src/definitions.h#L27-L29

Milice:
Yeah i know, i didn't change that though since there are still alot more code to actually fully support 10.80 with the premium shop and whatnot, and i don't have a clean enough code to pull with so i just kept it simple for now :3

src/protocollogin.cpp
```diff
@@ -89,10 +89,12 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
...
 	//Add premium days
-	if (g_config.getBoolean(ConfigManager::FREE_PREMIUM)) {
-		output->add<uint16_t>(0xFFFF); //client displays free premium
-	} else {
-		output->add<uint16_t>(account.premiumDays);
+	if (version >= 1080) {
+		output->addByte(g_config.getBoolean(ConfigManager::FREE_PREMIUM) || account.premiumDays > 0);
+		output->add<uint32_t>(g_config.getBoolean(ConfigManager::FREE_PREMIUM) ? 0 : (time(nullptr) + (account.premiumDays * 86400)));
+	}
+	else {
```
WibbenZ: Indentation

WibbenZ:
So you decided to release it? ;D

src/protocollogin.cpp
```diff
@@ -89,10 +89,12 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
 	}
 
 	//Add premium days
-	if (g_config.getBoolean(ConfigManager::FREE_PREMIUM)) {
-		output->add<uint16_t>(0xFFFF); //client displays free premium
-	} else {
-		output->add<uint16_t>(account.premiumDays);
+	if (version >= 1080) {
+		output->addByte(g_config.getBoolean(ConfigManager::FREE_PREMIUM) || account.premiumDays > 0);
+		output->add<uint32_t>(g_config.getBoolean(ConfigManager::FREE_PREMIUM) ? 0 : (time(nullptr) + (account.premiumDays * 86400)));
```
djarek: You should use a portable way to get UNIX time. (time() is dependant on POSIX compliance). Use the C++ chrono library (as an example, check OTSYS_TIME()).

ranisalt: Can't he just use `OTSYS_TIME()` there?

djarek: It returns time in *milliseconds* from the UNIX epoch.

Milice:
Maybe ur right about the OTSYS_TIME() but time(nullptr) is used in other files, for example bed.cpp, so it shouldn't really be a problem

ranisalt:
@Milice but `time()` is an unspecified function, and it is not reliable. It may return different values whether you are running Linux, Windows, BSD or whatever. It can be used fine where you just need to get a delta of time, but not reliably to get the current time as it is implementation dependant, which is what you want.

That is what I understand, you can read at http://en.cppreference.com/w/c/chrono/time

MaxyTibia:
@Milice, the character list works 100%, but after login, got this error:
Control.cpp 1517: unknow packet type during game (Type = 0)

Milice:
Yeah, u need to add a string and i think a uint16_t in the first byte of protocolgame

ranisalt:
Any news on this?

HyphenMaster:
When I made this change in my source code and compiled it I get tibia debug when I try to login in my character. Anyone know what's going on ?

PR #1513, opened 2015-08-11
Fix effect debug

### LINKED ISSUE #1488: Talkaction commands executed by certain group debugs all connected players
Talkaction commands "!z" and "!x" crashes all connected players if, and only if it is executed by a player of groups "2" or "3" (gamemaster or god).

WibbenZ:
Added 2 new functions to the Game class, hasEffect and hasDistanceEffect.
They will iterate through the enum to try and find the argument provided.
Also added the functions to the scripts.

Tested and works fine.
Closes #1488

### DIFF
--- a/data/talkactions/scripts/animationeffect.lua
+++ b/data/talkactions/scripts/animationeffect.lua
@@ -4,6 +4,11 @@ function onSay(player, words, param)
 	end
 
 	local effect = tonumber(param)
+	if not Game.hasDistanceEffect(effect) then
+		player:sendCancelMessage("Invalid effect id.")
+		return false
+	end
+
 	local position = player:getPosition()
 	local toPositionLow = {z = position.z}
 	local toPositionHigh = {z = position.z}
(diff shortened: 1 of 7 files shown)


brunominervino:
Nice man! :+1:

marksamman:
You can't iterate over an enum with gaps, and the server doesn't decide which effects that exist, the client does. If you use the predefined constants, you shouldn't have any problems. #1488 should be solved by doing nothing when no param is passed, value 0 or whatever it is that's crashing clients should crash the client if the effect does not exist if that's what the client chooses to do when an effect doesn't exist. !z and !x are also test commands and shouldn't be used in production.

WibbenZ:
@marksamman why is the issue still open in that case?
Tought we were looking for a solution - and this kinda is~.
I only tested a few of the effects, first, last and a couple in the middle and had no issues.
Should the issue be closed or are we looking for another solution?

marksamman:
This is not a solution, **you can't iterate over an enum with gaps**. Test your solution with magic effect 100. The issue is still open because I haven't had time to confirm that it's a bug, hence why it's labeled as unconfirmed.

PR #1516, opened 2015-08-12
Use RAII to manage Combat conditions.

djarek:
Removed remaining owning raw pointers from the Combat class.


marksamman:
Thanks!

PR #1517, opened 2015-08-12
Move doSet*Outfit functions to Lua

WibbenZ:
Moved the following functions to Lua:
doSetCreatureOutfit
doSetMonsterOutfit
doSetItemOutfit

Also added these two functions:
Creature:setMonsterOutfit
Creature:setItemOutfit

The functions return the same thing, tested and works fine.


djarek:
If you keep on going at this rate there's not going to be much left in the server engine :D (not saying it's a bad thing, TBH game logic should be mostly in lua). Keep up the good work!

WibbenZ:
Well thats the plan :)

I totally agree, right now im just trying to get rid of the old lua interface, trying to find a good way to implement the new group flags aswell as moving the get flag function to compat :p

data/lib/core/creature.lua
```diff
@@ -50,3 +50,48 @@ end
 function Creature.isTile(self)
 	return false
 end
+
+function Creature:setMonsterOutfit(monster, time)
+	if not self then
+		return false
+	end
```
ranisalt: Is this even possible? Reaching here with null `self`? The same for the other functions.

data/lib/core/creature.lua
```diff
@@ -50,3 +50,48 @@ end
...
+
+	local monsterType = MonsterType(monster)
+	if not monsterType then
+		return false
+	end
+
+	if self:getPlayer() and not getPlayerFlagValue(self, PlayerFlag_CanIllusionAll) then
+		if not monsterType:isIllusionable() then
```
ranisalt: You can merge those checks: `if self:getPlayer() and not getPlayerFlagValue(self, PlayerFlag_CanIllusionAll) and not monsterType:isIllusionable() then` or even `if self:getPlayer() and not (getPlayerFlagValue(self, PlayerFlag_CanIllusionAll) or monsterType:isIllusionable()) then`

data/lib/core/creature.lua
WibbenZ: Atleast in the past we did it like that, or it's just one of my weird things

data/lib/core/creature.lua
WibbenZ: Been a while since I wrote this but my guess is that it would be to long, this pr has been open for ages so I doubt it will even be merged during 2017 ;)

Once mark has time to check my prs I can update them, no reason to waste time when ill have to rebase in x months, but thanks for the tip and now I have a reminder of it :p

data/lib/core/creature.lua
```diff
@@ -52,16 +52,12 @@ function Creature.isTile(self)
...
-	end
-
 	local monsterType = MonsterType(monster)
 	if not monsterType then
 		return false
 	end
 
-	if self:getPlayer() and not getPlayerFlagValue(self, PlayerFlag_CanIllusionAll) then
+	if self:getPlayer() and not (getPlayerFlagValue(self, PlayerFlag_CanIllusionAll) or monsterType:isIllusionable()) then
 		if not monsterType:isIllusionable() then
```
ranisalt: Double check for illusionable.

data/lib/core/creature.lua
```diff
@@ -50,3 +50,38 @@ end
...
 	return false
 end
+
+function Creature:setMonsterOutfit(monster, time)
+	local monsterType = MonsterType(monster)
+	if not monsterType then
+		return false
+	end
+
+	if self:getPlayer() and not (getPlayerFlagValue(self, PlayerFlag_CanIllusionAll) or monsterType:isIllusionable()) then
```
ranisalt: You can use `self:isPlayer()` here as it's lighter and works the same.

Also wouldn't you want to move `getPlayerFlagValue` to the `Player` metatable before? :P just a suggestion

data/lib/core/creature.lua
```diff
@@ -50,3 +50,38 @@ end
...
+	local itemType = ItemType(item)
+	if not itemType then
+		return false
+	end
+
+	local outfit = {}
+	outfit.lookTypeEx = itemType:getId()
+
+	local condition = Condition(CONDITION_OUTFIT)
+	condition:setOutfit(outfit)
```
ranisalt: I think you don't need to assign outfit to a variable, just use
```lua
condition:setOutfit({
    lookTypeEx = itemType:getId()
})
```

data/lib/compat/compat.lua
```diff
@@ -1005,3 +1005,33 @@ end
...
+	return creature:setMonsterOutfit(name, time)
+end
+
+function doSetItemOutfit(cid, item, time)
+	local creature = Creature(cid)
+	if not creature then
+		return false
+	end
+	return creature:setItemOutfit(item, time)
+end
```
ranisalt: Most short functions are written like this in compat:
```lua
function doSetItemOutfit(cid, item, time) local c = Creature(cid) return c ~= nil and creature:setItemOutfit(item, time) end
function doSetMonsterOutfit(cid) local c = Creature(cid) return c ~= nil and creature:setMonsterOutfit(name, time) end
```
And you could do the same to save a few extra lines for the two above.

ranisalt:
Sorry it'll need another rebase 🤣 🤣 🤣 so I have some feedback you can address while at it.

data/lib/core/creature.lua
WibbenZ: Ye I wanted to move it but couldn't find a "clean" way to do it while keeping backwards compat, but might give it another try soon :P

WibbenZ:
Should be OK now @ranisalt

data/lib/compat/compat.lua
```diff
@@ -1006,3 +1006,21 @@ end
 function Guild.removeMember(self, player)
 	return player:getGuild() == self and player:setGuild(nil)
 end
+
+function doSetItemOutfit(cid, item, time) local c = Creature(cid) return c ~= nil and creature:setItemOutfit(item, time) end
+function doSetMonsterOutfit(cid) local c = Creature(cid) return c ~= nil and creature:setMonsterOutfit(name, time) end
```
ranisalt: Oooops its `c:` not `creature:`

ranisalt:
Just one minor change I accidentally made you make.

WibbenZ:
There we go haha

data/lib/compat/compat.lua
```diff
@@ -1006,3 +1006,21 @@ end
 function Guild.removeMember(self, player)
 	return player:getGuild() == self and player:setGuild(nil)
 end
+
+function doSetItemOutfit(cid, item, time) local c = Creature(cid) return c ~= nil and c:setItemOutfit(item, time) end
+function doSetMonsterOutfit(cid) local c = Creature(cid) return c ~= nil and c:setMonsterOutfit(name, time) end
```
ranisalt: Where should `name` and `time` come from? I think they are missing here

WibbenZ: Correct haha, just copied your comment and only tested the doSetCreatureOutfit compat function :P

ranisalt:
Thanks! You are very helpful 😄

WibbenZ:
Just fun to see new things getting added :)
Hopefully we get to see some more devs picking up :p

PR #1519, opened 2015-08-13
Network stack cleanup.

djarek:
Socket objects are now embedded directly into Connection objects.
Removed m prefix from member variable names. Removed logError from
Connection - errors should always be reported. Connection will now
be always closed if an exception is thrown by socket reads or writes.
Added the final specifier to the Service class.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -26,16 +26,13 @@
 #include "scheduler.h"
 #include "server.h"
 
-bool Connection::logError = true;
-
 extern ConfigManager g_config;
 
-Connection_ptr ConnectionManager::createConnection(boost::asio::ip::tcp::socket* socket,
-        boost::asio::io_service& io_service, ServicePort_ptr servicer)
+Connection_ptr ConnectionManager::createConnection(boost::asio::io_service& io_service, ConstServicePort_ptr servicePort)
 {
 	std::lock_guard<std::mutex> lockClass(connectionManagerLock);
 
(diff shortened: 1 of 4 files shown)


src/connection.cpp
```diff
@@ -26,16 +26,13 @@
...
 extern ConfigManager g_config;
 
-Connection_ptr ConnectionManager::createConnection(boost::asio::ip::tcp::socket* socket,
-        boost::asio::io_service& io_service, ServicePort_ptr servicer)
+Connection_ptr ConnectionManager::createConnection(boost::asio::io_service& io_service, ConstServicePort_ptr servicePort)
 {
 	std::lock_guard<std::mutex> lockClass(connectionManagerLock);
-
-	auto connection = std::make_shared<Connection>(socket, io_service, servicer);
+
```
WibbenZ: Delete, send me a pm on otland and ill send you a regex code to search for trailing spaces tomorow :p

djarek: That's the only thing I hate in my current IDE... it leaves trailing tabs (spaces are removed correctly most of the time).

WibbenZ: I use this to find traling spaces:
```
[ \t]+$
```

I use notepad++, if you do the same - ctrl+f - find in files:
Find what: [ \t]+$
Replace with: (leave empty, otherwise it will replace things in items.otb aswell as *.otbm.
Directory: path to the folder
Search mode: regular expression

marksamman:
Thanks!

PR #1520, opened 2015-08-14
Move reload to talkactions

WibbenZ:
Moved /reload (from commands) to talkactions.
Added Game.reload(reloadType)

Removed an extra import from game.cpp
Removed trailing from rsa.cpp


Luckey1729:
Impressive! Great initiative! :+1:

src/game.cpp
```diff
@@ -5544,3 +5550,47 @@ void Game::removeUniqueItem(uint16_t uniqueId)
...
+		return g_monsters.reload();
+	} else if (reloadType == RELOAD_TYPE_MOVEMENTS) {
+		return g_moveEvents->reload();
+	} else if (reloadType == RELOAD_TYPE_NPCS) {
+		Npcs::reload();
+		return true;
+	} else if (reloadType == RELOAD_TYPE_RAIDS) {
+		return g_game.raids.reload() && g_game.raids.startup();
+	} else if (reloadType == RELOAD_TYPE_SPELLS) {
+		return g_spells->reload() && g_monsters.reload();
```
djarek: If a monster reload fails after a spells reload, you should print an error and terminate the server.

WibbenZ: I agree about not reloading monsters, but terminating the server?

djarek:
@WibbenZ 
You didn't understand my comment. If ANY failure occurs during a spell reload you HAVE TO terminate the server (std::terminate()), because monsters might have dangling spell pointers.

WibbenZ:
Ok ive added termination now, but how come that wasn't added before? :o

djarek:
Probably because I've only realized it 30 minutes ago :D. BTW. if monsters fail to reload after spells you have to terminate as well.

WibbenZ:
Oh okay :P 
Added a monster check aswell.

Kamenuvol:
Why is it >>forgotten<<? =P

ranisalt:
> Why is it >>forgotten<<? =P

I don't remember

WibbenZ:
There are alot of open PRs that aren't getting any attention, if we skip Djareks cast PR from 2014 there are alot from 2015 aswell.

src/const.h
```diff
@@ -517,6 +517,27 @@ enum PlayerFlags : uint64_t {
...
+	RELOAD_TYPE_TALKCTIONS = 9,
+	RELOAD_TYPE_ITEMS = 10,
+	RELOAD_TYPE_WEAPONS = 11,
+	RELOAD_TYPE_QUESTS = 12,
+	RELOAD_TYPE_MOUNTS = 13,
+	RELOAD_TYPE_GLOBALEVENTS = 14,
+	RELOAD_TYPE_EVENTS = 15,
+	RELOAD_TYPE_CHAT = 16,
+	RELOAD_TYPE_GLOBAL = 17,
+	RELOAD_TYPE_COMMANDS = 18, // Planned to be removed, so let's put it last
```
ranisalt: You probably don't need to define values for each one, only for commands where you can `= <last enum value> + 1`. It would also be much better to have them alphabetically ordered.

Add a `RELOAD_TYPE_ALL` at the top so that one can reload everything when not using a parameter.

src/luascript.cpp
```diff
@@ -1702,6 +1702,25 @@ void LuaScriptInterface::registerFunctions()
...
+	registerEnum(RELOAD_TYPE_TALKCTIONS)
+	registerEnum(RELOAD_TYPE_ITEMS)
+	registerEnum(RELOAD_TYPE_WEAPONS)
+	registerEnum(RELOAD_TYPE_QUESTS)
+	registerEnum(RELOAD_TYPE_MOUNTS)
+	registerEnum(RELOAD_TYPE_GLOBALEVENTS)
+	registerEnum(RELOAD_TYPE_EVENTS)
+	registerEnum(RELOAD_TYPE_CHAT)
+	registerEnum(RELOAD_TYPE_GLOBAL)
+	registerEnum(RELOAD_TYPE_COMMANDS)
```
ranisalt: If you change the order on the enum please make it alphabetical too.

src/luascript.cpp
```diff
@@ -4396,6 +4417,24 @@ int LuaScriptInterface::luaGameStartRaid(lua_State* L)
...
+	if (!reloadType) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	if (reloadType == RELOAD_TYPE_GLOBAL) {
+		pushBoolean(L, g_luaEnvironment.loadFile("data/global.lua") == 0);
+	} else {
+		pushBoolean(L, g_game.reload(reloadType));
+	}
```
ranisalt: You could move this condition to the reload method directly, instead of treating on the Lua interface.

src/game.cpp
```diff
@@ -5544,3 +5550,54 @@ void Game::removeUniqueItem(uint16_t uniqueId)
...
+	} else if (reloadType == RELOAD_TYPE_MOUNTS) {
+		return mounts.reload();
+	} else if (reloadType == RELOAD_TYPE_GLOBALEVENTS) {
+		return g_globalEvents->reload();
+	} else if (reloadType == RELOAD_TYPE_EVENTS) {
+		return g_events->load();
+	} else if (reloadType == RELOAD_TYPE_CHAT) {
+		return g_chat->load();
+	} else if (reloadType == RELOAD_TYPE_COMMANDS) {
+		return commands.reload();
```
ranisalt: For the love of god please use a `switch`.

data/talkactions/scripts/reload.lua
```diff
@@ -0,0 +1,75 @@
...
+	["events"] = { targetType = RELOAD_TYPE_EVENTS, name = "events"},
+
+	["chat"] = { targetType = RELOAD_TYPE_CHAT, name = "chatchannels"},
+	["channel"] = { targetType = RELOAD_TYPE_CHAT, name = "chatchannels"},
+	["chatchannels"] = { targetType = RELOAD_TYPE_CHAT, name = "chatchannels"},
+
+	["global"] = { targetType = RELOAD_TYPE_GLOBAL, name = "global.lua"},
+
+	["command"] = { targetType = RELOAD_TYPE_COMMANDS, name = "commands"},
+	["commands"] = { targetType = RELOAD_TYPE_COMMANDS, name = "commands"}
```
ranisalt: You don't add spaces to the end but add to the start of the inner tables, please make it consistent.

src/luascript.cpp
WibbenZ: Sure would look cleaner but we would have to include LuaEnvironment and it would only be used for that.
To me it seems like a waste, but if you want ill do it.

ranisalt: Hmm looks good to me then, fine.

ranisalt:
Just waiting for rebase.

WibbenZ:
@ranisalt Btw its rebased :smile:

src/game.cpp
```diff
@@ -47,6 +50,10 @@ extern Spells* g_spells;
 extern Vocations g_vocations;
 extern GlobalEvents* g_globalEvents;
 extern Events* g_events;
+extern CreatureEvents* g_creatureEvents;
+extern Monsters g_monsters;
+extern MoveEvents* g_moveEvents;
+extern Weapons* g_weapons;
```
ranisalt: While at it, can you maintain these in alphabetical order?

src/game.cpp
```diff
@@ -5530,3 +5537,121 @@ void Game::removeUniqueItem(uint16_t uniqueId)
...
+			quests.reload();
+			mounts.reload();
+			g_globalEvents->reload();
+			g_events->load();
+			g_chat->load();
+			commands.reload();
+			return true;
+		}
+
+		case RELOAD_TYPE_ACTIONS: {
```
ranisalt: On those cases that are a single line, keep them short e.g. `case RELOAD_TYPE_ACTIONS: return g_actions->reload();`

src/game.cpp
```diff
@@ -5530,3 +5537,121 @@ void Game::removeUniqueItem(uint16_t uniqueId)
...
+			return g_talkActions->reload();
+		}
+
+		case RELOAD_TYPE_WEAPONS: {
+			bool results = g_weapons->reload();
+			g_weapons->loadDefaults();
+			return results;
+		}
+
+		default: {
```
ranisalt: The default could be reload all.

src/luascript.cpp
```diff
@@ -1709,6 +1709,26 @@ void LuaScriptInterface::registerFunctions()
...
+	registerEnum(RELOAD_TYPE_GLOBALEVENTS)
+	registerEnum(RELOAD_TYPE_ITEMS)
+	registerEnum(RELOAD_TYPE_MONSTERS)
+	registerEnum(RELOAD_TYPE_MOUNTS)
+	registerEnum(RELOAD_TYPE_MOVEMENTS)
+	registerEnum(RELOAD_TYPE_NPCS)
+	registerEnum(RELOAD_TYPE_QUESTS)
+	registerEnum(RELOAD_TYPE_RAIDS)
+	registerEnum(RELOAD_TYPE_SPELLS)
+	registerEnum(RELOAD_TYPE_TALKCTIONS)
```
ranisalt: Typo

ranisalt:
There's still a typo somewhere related to the `TALKCTIONS` typo.

ranisalt:
Thanks!

PR #1521, opened 2015-08-15
Replace reinterpret_casts.

djarek:
Replaced reinterpret_casts where the semantic meaning of the type
cast is satisfied by a static_cast (in other word where we cast
pointers/references to objects that are related).

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -86,7 +86,7 @@ Event* Actions::getEvent(const std::string& nodeName)
 
 bool Actions::registerEvent(Event* event, const pugi::xml_node& node)
 {
-	Action* action = reinterpret_cast<Action*>(event);
+	Action* action = static_cast<Action*>(event); //event is guaranteed to be an Action
 
 	pugi::xml_attribute attr;
 	if ((attr = node.attribute("itemid"))) {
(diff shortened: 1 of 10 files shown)


marksamman:
reinterpret_cast should be preferred over static_cast when casting pointers because it doesn't impose any runtime overhead, it just interprets the same bits of data as another type. static_cast will re-construct the new type with the old type's value (semantically no difference when casting pointers). Correct me if I'm wrong.

djarek:
AFAIK the only C++-style cast that imposes an overhead is dynamic_cast (because it does checks using RTTI).
```
dynamic_cast
 is primarily used to perf
orm “safe downcasting,” i.e.,
to determine whether an object is 
of a particular type in an inher-
itance hierarchy. It is the only cast that cannot be performed us-
ing the old-style syntax. It is also the only cast that may have a
significant runtime cost
. (I’ll provide details on this a bit later.)
```
source: http://www.aristeia.com/EC3E/3E_item27.pdf
(it quotes Meyer's book, so I guess it's a reliable source)

marksamman:
So assuming they generate the same code when casting pointers, what's the benefit of replacing reinterpret_cast with static_cast?

djarek:
The benefit is that it's clear at first glance that we're not doing any fancy pointer magic and the compiler checks whether types are related when using a static_cast (basically reinterpret_cast throws away type safety).

marksamman:
Fair enough. Thanks!

PR #1526, opened 2015-08-17
Apply the RAII idiom in Chat classes.

djarek:
Also reformatted the ChatChannel constructor for better readability.

### DIFF
--- a/src/chat.h
+++ b/src/chat.h
@@ -112,15 +112,15 @@ class PrivateChatChannel final : public ChatChannel
 		void closeChannel() const;
 
 		const InvitedMap& getInvitedUsers() const {
-			return m_invites;
+			return invites;
 		}
 
 		const InvitedMap* getInvitedUsersPtr() const final {
-			return &m_invites;
+			return &invites;
 		}
 
 	protected:
-		InvitedMap m_invites;
+		InvitedMap invites;
 		uint32_t owner;
 };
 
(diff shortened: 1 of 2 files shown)


HeavenIsLost:
You pretend to do that on the whole tfs? danm, it will break a lot of my code i guess(tfs 1.1).

src/chat.h
```diff
@@ -152,13 +159,13 @@ class Chat
...
-		std::map<Party*, ChatChannel*> partyChannels;
-		std::map<uint32_t, ChatChannel*> guildChannels;
+		std::map<uint16_t, PrivateChatChannel> privateChannels;
+		std::map<Party*, ChatChannel> partyChannels;
+		std::map<uint32_t, ChatChannel> guildChannels;
 
 		LuaScriptInterface m_scriptInterface;
 
-		ChatChannel* dummyPrivate;
+		std::unique_ptr<PrivateChatChannel> dummyPrivate;
```
kornholi: does this actually need to be a pointer?

djarek: Actually, it doesn't have to. I'll embed that variable directly into the Chat class.

djarek:
@HeavenIsLost 
These really are low-hanging fruit. Applying RAII to the map code or core game logic is a little out of my reach at the moment - there's way too much spaghetti in there for my taste. Unless you had some customized chat features this shouldn't break any lua scripts or code outside the chat files.

marksamman:
Thanks!

Issue #1497, opened 2015-08-07
Documentation and code style guidelines


djarek:
What is the project's policy on documentation? If there is one I'd like to suggest creating a wiki page that outlines the format and content of docs. We should also have a page on code style to make it easier for inexperienced people to contribute.

ranisalt:
:+1: for the code style page, I see lots of people trying to contribute but they completely miss on code style, even people with several contributions already.

marksamman:
If you'd like to write documentation/code style guidelines, I can set up a repository for the wiki and give you write access to it.

djarek:
I guess I could put some work into it from time to time (no guarantees, though).

WibbenZ:
I could help out aswell when I have nothing else do to :p 
But what are we actually talking about? How our code style is to how Lua works?
Or is it just how the code we use works?
Ill let someone else start before I help out I think haha

djarek:
At first we should focus on the things that might be obvious to more experienced contributors, like say, where to put braces, indentation (both for C++ and lua). As for documentation - I think we should include information like (on the function level):
-pre, post conditions (what program states are required or allowed, what to expect when the function returns)
-modification of global and object state
-thread safety
-other information(for example, warning the users about possible UB occuring when you schedule an event in lua and store userdata in it)

As for the format of the documentation itself - I used doxygen in a few projects so I'm most familliar with that, however I'm open to suggestions.

WibbenZ:
Sounds good, however I have never heard of doxygen, ill serach it up abit.

PR #1531, opened 2015-08-20
Removed needless spaces around operators

bakerj76:
Generally, spaces around operators imply a higher precedence.

### DIFF
--- a/data/global.lua
+++ b/data/global.lua
@@ -13,18 +13,18 @@ levelDoors = {1227, 1229, 1245, 1247, 1259, 1261, 3540, 3549, 5103, 5112, 5121, 
 keys = {2086, 2087, 2088, 2089, 2090, 2091, 2092, 10032}
 
 function getDistanceBetween(firstPosition, secondPosition)
-	local xDif = math.abs(firstPosition.x - secondPosition.x)
-	local yDif = math.abs(firstPosition.y - secondPosition.y)
+	local xDif = math.abs(firstPosition.x-secondPosition.x)
+	local yDif = math.abs(firstPosition.y-secondPosition.y)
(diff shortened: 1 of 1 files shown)


WibbenZ:
It's done that way to make the code cleaner.

djarek:
@WibbenZ 
Seems that #1497 would be useful in a situation like this.

WibbenZ:
Yes truly would :p

ranisalt:
Disagreed. I have never ever stumbled across anyone that said that. Higher precedence is implied (and enforced) with parenthesis, not subtle coding style.

bakerj76:
Please look at PEP8 coding style guidelines, and you'll see what I mean.
It's not syntactical. It's to make the precedence explicit rather than implicit.

ranisalt:
PEP8 is good for Python syntax, I'm a Python programmer myself. Mathematics is explicit by itself, everyone above the basic ed. 4th grade can read those expressions and understand the order.

Anyway, you contradict yourself when you say this is good for explicit precedence and you only use it where there is absolutely no ambiguity.

WibbenZ:
@ranisalt dear god no...
Using ' for strings insted of ".
Using 2 spaces insted of a tab.

ORTS is using some of that and I hate it :p 
We would also have to rewrite alot of our code as of right now (changing tabs to spaces can be done quickly) but quotes can be abit harder.

We could write something like it, but just how we do it as of right now.
That way we don't have to change anything we have as of right now and have something to link in cases like this.

Luckey1729:
We could also adopt a syntax like lisp.

![Lisp]
(https://www.abc.se/~m9071/highlight/lisp.png)

ranisalt:
\*\*amazing\*\*

PR #1534, opened 2015-08-22
Tile cleanup

djarek:
Use implicit dynamic dispatch instead of explicit (which wasless efficient). Use private inheritance  instead of composition to reduce the amount of boilerplate in TileItemVector.

### DIFF
--- a/src/tile.cpp
+++ b/src/tile.cpp
@@ -933,7 +933,7 @@ void Tile::addThing(int32_t, Thing* thing)
 
 			items = makeItemList();
 			items->insert(items->getBeginDownItem(), item);
-			++items->downItemCount;
+			items->addDownItemCount(1);
 			onAddTileItem(item);
 		}
 	}
@@ -1115,7 +1115,7 @@ void Tile::removeThing(Thing* thing, uint32_t count)
 
 			item->setParent(nullptr);
 			items->erase(it);
-			--items->downItemCount;
+			items->addDownItemCount(-1);
 			onRemoveTileItem(list, oldStackPosVector, item);
 		}
 	}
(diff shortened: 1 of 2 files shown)


ranisalt:
Good, any improvements on Tile are always great :+1:

src/tile.cpp
```diff
@@ -933,7 +933,7 @@ void Tile::addThing(int32_t, Thing* thing)
 
 			items = makeItemList();
 			items->insert(items->getBeginDownItem(), item);
-			++items->downItemCount;
+			items->addDownItemCount(1);
```
ranisalt: Why that?

djarek: Writing separate increment and decrement functions seems a little redundant.

ranisalt: Yes, but what's the advantage of having a new function when ++ and -- already exist?

djarek: downItemCount is a private member variable. It was accessible before because Tile was a friend of TileItemVector(which breaks Encapsulation principle), but that's not the biggest problem. The biggest issue is safety - a friend class can improperly delete the TileItemVector (by deleting a pointer to a vector), so to prevent such mistakes I've decided to remove the friend relation.

ranisalt: Good one, didn't notice that.

djarek:
Yeah tile cleanup is necessary but the biggest problem is that we're just scratching the itch without trying to solve the underlying problem - large portions of the core game logic are in dire need of refactoring (have you ever tried tracing the callgraph for moving a creature? or an item? the code for these operations is spilled over 4 or 5 source files with functions seemingly performing the same tasks calling each other).

djarek:
Rebased to master.

@marksamman 
What is your opinion about this PR?

marksamman:
I'm a bit worried about the performance implications of dynamic dispatch where performance may be critical, not sure how much it differs from our current implementation which is basically just a branch prediction. It would be interesting to see if anyone wants to benchmark it.

The code using a tile flag was originally from https://github.com/opentibia/server, so I'm not even sure of the real reason why dynamic dispatch was avoided. The code looks good enough that I'll merge it though, hopefully someone will open an issue if this makes tile operations significantly slower.

PR #1536, opened 2015-08-22
Threads cleanup

djarek:
Simplified shutdown sequence when an error occurs in the main loader because there is
no point in letting threads go into the `CLOSING` state as there is most likely no meaningful change in game state and no work items in their queues. Moved common thread management member variables into a common base class for better code reuse. This also fixes a data race on the thread state flag in the async db task runner.


marksamman:
Thanks!

PR #1541, opened 2015-08-25
Small mounts optimization.

djarek:
Replace list with vector for better look-up performance. Also use the std::find_if algorithm for mounts search.


marksamman:
Thanks!

PR #1542, opened 2015-08-25
Move sellHouse to Lua

WibbenZ:
Added the following functions:
Player:startTrade(tradePartner, transferItem)
Player:hasBiddedOnHouse()

House:getTransferItem()
House:resetTransferItem()
House:moveTransferItemTo(item, player)

### DIFF
--- a/data/XML/commands.xml
+++ b/data/XML/commands.xml
@@ -2,5 +2,4 @@
 <commands>
 	<command cmd="/reload" group="2" acctype="5" log="yes" />
 	<command cmd="/raid" group="2" acctype="4" log="yes" />
-	<command cmd="!sellhouse" group="1" acctype="1" log="no" />
 </commands>
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -97,7 +97,7 @@ function Player.addManaSpent(...)
 	return ret
 end
 
-function Player:hasBiddedOnHouse()
+function Player:hasBidOnHouse()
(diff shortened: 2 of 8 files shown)


Quintinon:
Sorry, but I'm going to be that grammar guy ...
Player:hasBiddedOnHouse should be Player:hasBidOnHouse. The past tense form of the verb "bid" is "bid".

raymondtfr:
Lol, I had to comment on this one, uahdshua. He's right though!

WibbenZ:
There fixed, only copied the name from IOLoginData :p

marksamman:
The Lua functions are too specific for this functionality that we should just leave this in the source code until it can be solved in a more generic way through Lua functions in the future. An example would be having item:setParent instead of house:moveTransferItemTo, but we need to think about it carefully so that it is exposed in a safe way to the end-users of the Lua API.

WibbenZ:
Well I tought about that, but as you said it has to be a safe function and it felt like a time waste to even think about it.

This way you can still use that exact function in other things, eg. events to give or remove houses.
If you ask me this is the best way to do it, Item:setParent should not be moved to Lua IMO.

yoiker:
There is a possibility that it works with premium_points? A quick example would be : name points , count but has premium_points function and not home because I tried but does not open the window but does not execute the function but have points in the account.

WibbenZ:
Yes you could easily make it use premium points insted of gps.
But as Mark said, we need a better function if we want this in Lua.

yoiker:
God the problem is I do not know something I ask but thank you very much for your answer!

Kamenuvol:
!sellhouse as it is now looks like a mod. It should not be in core.
We already have a function to change house owner.

PR #1547, opened 2015-08-28
Fix Undefined Behavior in Dispatcher.

djarek:
Scheduler tasks are deleted with the base pointer which
causes Undefined Behavior as the base class Task does not have
a virtual destructor.


marksamman:
Thanks!

PR #1548, opened 2015-08-28
Fix strict aliasing violations

djarek:
Fixed strict aliasing violations (and potential violations) in the codebase, unfortunately, pugixml still contains potentially strict aliasing unsafe code.

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -490,9 +490,10 @@ CallBack* Combat::getCallback(CallBackParam_t key)
 	return nullptr;
 }
 
-void Combat::CombatHealthFunc(Creature* caster, Creature* target, const CombatParams& params, void* data)
+void Combat::CombatHealthFunc(Creature* caster, Creature* target, const CombatParams& params, CombatDamage* data)
 {
-	CombatDamage damage = *reinterpret_cast<CombatDamage*>(data);
+	assert(data);
+	auto& damage = *data;
 	if (g_game.combatBlockHit(damage, caster, target, params.blockedByShield, params.blockedByArmor, params.itemId != 0)) {
 		return;
 	}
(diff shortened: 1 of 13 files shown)


ranisalt:
If we are to assert the pointer is not null and then derreference it, why not use reference directly?

djarek:
I'm not sure how much code would need to be changed. Experience has taught me to try to keep PRs as small as possible, but you're absolutely right, if we want to refer to an object and not deal with a null we idiomatically use a reference (or not_null<> from the GSL, however, that's still being worked on and the prototype won't be shipped until 2016).

marksamman:
This hasn't been merged yet because I'm worried about the performance implications of using memcpy instead of aliasing, I don't care so much about strict aliasing that I'm willing to sacrifice performance in functions such as XTEA. I was going to check if GCC, Clang and VS generate the same assembly with optimizations enabled, but haven't had time to do so.

gugahoa:
Isn't memcpy the fastest way for memory-to-memory copy? How can aliasing be faster? Honest question

djarek:
G++ 4.9.2 on x86-64 generated this asm:
```
movl	%esi, %esi	# readPos, D.32402
movl	4(%rdi,%rsi), %eax	# MEM[(char * {ref-all})_7], ret$4
movl	(%rdi,%rsi), %edx	# MEM[(char * {ref-all})_4], D.32290
```
For the 2x memcpy calls and this:
```
leal	1(%rsi), %eax	#, D.32395
movl	%esi, %esi	# readPos, D.32395
movl	(%rdi,%rsi,4), %edx	# *_5, D.32283
movl	(%rdi,%rax,4), %eax	# *_12, D.32397
```
for the 2 aliasing assignments.

marksamman:
Looks good to me. Could you also remove -fno-strict-aliasing from CMake?

djarek:
Seems to have generated the same code.

marksamman:
I meant that you should remove it from the repository (these 3 lines: https://github.com/otland/forgottenserver/blob/master/CMakeLists.txt#L19-L21), so that Travis shows that it builds fine with strict aliasing (and so that it will complain if we violate the strict aliasing rules in the future).

djarek:
The problem is that pugixml has some potentially unsafe code.

marksamman:
Thanks!

PR #1550, opened 2015-08-29
Fix #1533

### LINKED ISSUE #1533: db.asyncStoreQuery is crashing the server
the function asyncStoreQuery work just fine, the problem happens when there is any lua error inside the callback.

Code to crash the server(very simple):

http://hastebin.com/obiseweroc.lua

Visual Studio 2013 debug print(with callstack):

http://prntscr.com/86zmya

Hope @marksamman can fix it.

@djarek check this too :+1:

djarek:
Async queries in lua will now properly setup a ScriptEnvironment for their lua callbacks. Closes #1533.


marksamman:
Thanks!

PR #1551, opened 2015-08-29
Replaced Optional with unique_ptr.

djarek:
Optional was just a poorly implemented unique_ptr that could be easily mistaken for "real" optional types.


Nu77:
Nice :+1:

marksamman:
The reason the Optional class exists is so that we don't need to explicitly define the ItemType copy constructor, assign everything manually and manually copy the values of Abilities and ConditionDamage. Optional is not meant to be a unique_ptr, unique_ptr is not copyable, Optional is.

We can use unique_ptr and remove Optional if we can ensure that ItemType is not copyable. Please check if Visual Studio 2013 and/or 2015 allows non-copyable value types in vectors.

djarek:
@marksamman 
No need to check it, we're already using vectors of non-copyable(but moveable) types in the codebase (see spell vectors in MonsterTypes, spellblocks are movable, but not copyable). AFAIK this functionality has been (at least partially) available in VS 2010, definitely available in VS 2012.

marksamman:
In that case we should make ItemType non-copyable just to be sure that it isn't copied anywhere, and also remove optional.h from vc12/theforgottenserver.vcxproj

djarek:
Placing a member with deleted functions in a class/struct automatically prevents the compiler from generating the default ones (so the explicit deletion of ItemType copy constructors is redundant), but I've added explicit declarations for clarity anyways.

marksamman:
Thanks!

Issue #1236, opened 2015-03-16
Experience Share giving less experience with different stages


TwistedScorpio:
Basically I have two players:

Level 305 and Level 260

Experience rate for level 305 is 5x, experience rate for 260 is 7x.

NOTE: I modified the Amazon experience to be 2100.
So if Level 305 kills the Amazon and is not in a party he will get 10500 experience *  1.5 (stamina bonus) = 15750

But when shared experience with only Level 260 in the party, he will get as low as 11032.

This happened in the latest commit of TFS 1.1-development.

Pictures:
![image](https://cloud.githubusercontent.com/assets/6238582/6677084/11292400-cbea-11e4-894a-2f7a68518f27.png)

![image](https://cloud.githubusercontent.com/assets/6238582/6677096/27e8f710-cbea-11e4-9d75-d6f9ca8ed3c4.png)

Kamenuvol:
Exp Stage and Rate should be the last thing to multiply:
total: stageRate * happyHour * ceil(exp * (1/members + partyBonus))
In this case:
total: 5 * 1,5 * 2100 * (0,5 + 0,2)
level 305: 11025
level 260: 15435

It is not the same but is near and correct according to the sources.

TwistedScorpio:
So the fact that after experience share the lvl 305 gets less experience with shared exp is normal ?

Kamenuvol:
Yes, because of the expStage. Both characters should gain the same without it, whenever shareExp is activated.

However, we need to see why the xp numbers are not exactly correct.

Kamenuvol:
I am doing reverse mathematics =P and I see that:
15445,5 / 7 = 11032,5 /5
There is something multiplying 1,00068... to the result. I wonder what.

Kamenuvol:
Yes, the problem is using ceil to round 1470.xxxx to 1471. And it messes up everything, may we need to check Tibia Global to see what happens.
I have tested with this code:
http://www.heypasteit.com/clip/1V5E
And it returns the same as your xp.

marksamman:
This may have been fixed by #1579, please open a new issue if the issue persists.

Issue #1484, opened 2015-07-29
Issue with Shared Exp


kanohn:
Hello, im here to report a new issue with shared exp.

I Tested sometimes, when you enabled share experience, sometime give the correct exp and sometimes gives incorrect exp.
I'm already considering the stamina exp bonus, is not that what you're doing incorrectly give exp
Example:

If a monster gives 2000 exp and we have a 20% share bonus and sharing with 2 peoples:

 2000 + 20% / 2 = 1200 (It was to give this exp every time, but sometimes the exp increases about 400 + - with no reason)

ranisalt:
What is your testing ground? Players vocations, stamina bonus, and what monster did you test hunting?

kanohn:
I Tested with different vocations and same vocations, the exp bug even though.
I tested in others servers using tfs 1.1+ and the same thing happens..

I summoned a rotworm and others monsters, tested with different vocations and same vocations, with different bonus, does not matter. sometimes the shared exp returns the correct experience and sometimes an experience somewhat higher than expected.

has nothing to do with stamina bonus, I also tested everything, without the party exp active the experience returns with 100% all working bonus.

ranisalt:
Can you tell me exactly what happens when you hunt with different vocations (and how many) and the same (and how many players)?

It happens to track the bug.

kanohn:
As I said, and I repeat, no matter how many people are in the party, whether it is the same vocation or not, the bug is on exp when shared is active.

There are times that the experiment returns as expected, but sometimes returns a exp a little more or less, we can call it something out of logic, I have no idea of what can be, since the formula seems to be correct, but this bug happens with all servers that use tfs 1.1+

You can even test, open a test server and enable exp shared, you will realize that the exp will not return 100% every time you kill a monster

TwistedScorpio:
Could we relate this issue to this?

https://github.com/otland/forgottenserver/issues/1236

ranisalt:
Follow #1344

marksamman:
This may have been fixed by #1579, please open a new issue if the issue persists.

PR #1579, opened 2015-10-04
Fix an issue with Shared Experience

ninjalulz:
The experience gained is now being distributed once through the party leader, hence fixing the issue where you would receive multiple experience gained message depending on how many party members participated in the kill.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -666,6 +666,7 @@ void Creature::onDeath()
 	const int64_t timeNow = OTSYS_TIME();
 	const uint32_t inFightTicks = g_config.getNumber(ConfigManager::PZ_LOCKED);
 	int32_t mostDamage = 0;
+	std::map<Creature*, uint64_t> experienceMap;
 	for (const auto& it : damageMap) {
 		if (Creature* attacker = g_game.getCreatureByID(it.first)) {
 			CountBlock_t cb = it.second;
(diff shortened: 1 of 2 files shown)


Luckey1729:
Does this happen to fix the issue where, if you have 3 party members, for example, you see the experience gained message 3 times in the server log?

ninjalulz:
@Luckey1729 I forgot to mention that, but yes, it does fix that issue.

Luckey1729:
Awesome! I thought it might! :)

CristoferMartins:
As far i can tell, its good :)

marksamman:
Is there any reason why experienceMap can't be a local variable in Creature::onDeath?

ninjalulz:
No reason at all, I'll make an update.

marksamman:
Use the attacker pointer (Creature*) as key in the map so that you don't have to do the hash lookup again when distributing the experience.

src/creature.cpp
```diff
@@ -673,10 +674,32 @@ void Creature::onDeath()
...
 				mostDamageCreature = attacker;
 			}
-			attacker->onAttackedCreatureKilled(this);
+
+			if (attacker != this) {
+				uint64_t gainExp = getGainedExperience(attacker);
+				if (Player* player = attacker->getPlayer()) {
+					Party* party = player->getParty();
+					if (party && party->isSharedExperienceActive() && party->isSharedExperienceEnabled()) {
+						if (Player* leader = party->getLeader()) {
```
djarek: Move the condition to the if above.

ninjalulz: Alright xD

```C++
if (party && party->getLeader() && party->isSharedExperienceActive() && party->isSharedExperienceEnabled()) {
    attacker = party->getLeader();
}
```
Should I change it to look like this, or is it supposed to be done in another way?

marksamman: That's good enough, and I prefer it over declaring variables in if-statements, especially since party->getLeader() doesn't do anything heavy and most likely gets inlined by the compiler.

ninjalulz: Okay, good to know. :)

marksamman:
Thanks! Does this fix #1236 and/or #1484?

ninjalulz:
I believe so!

I made a quick test with two characters, both sorcerers, which means the shared exp bonus is set to 20% (1.2).

Experience Formula: ((monsterExp * sharedExpBonus) / members) * characterBonuses

Killing a rotworm (40 exp):
Level 20, 175x exp: ((40 * 1.2) / 2) * 1.5 * 175 = 6300 exp
Level 30, 125x exp: ((40 * 1.2) / 2) * 1.5 * 125 = 4500 exp

The experience gained was never changed

kanohn:
I found an bug, before and after your fix.

Try to reproduce:
1 - Invite to party and enable exp shared.
2 - In first test, give other char exura sio "name and with other char attack the monster ( you will gain different exp than second test)
3 - in second test, attack the monster with the two players ( you will gain other value of exp)

Theses values was tested with: Exp Stage (250x), + Bonus Green (50%) + Bonus Exp Shared (60%).

ninjalulz:
Open a new issue.

PR #1589, opened 2015-10-10
Fix #1587

### LINKED ISSUE #1587: Attack spell/rune causes negative (and positive) damage in creatures
If there are two creatures (or player) around you, if you reduce these creature's HP twice (to ease the observation) and use any attack spell/rune, you'll see that in one of the creatures the spell will deal "negative" damage, and will also deal damage on the other creature (as expected), but it'll deal a "positive" damage (the same damage value as the other creature that got a negative damage), that is, one of the creatures gets healed.
You can only reproduce this if there a minimum of 2 creatures around you.
screen shot 10-10-15 at 02 05 am
My guess is that this came with #1548 .

marksamman:


ranisalt:
Why replacing reference with copy?

djarek:
The original, type-unsafe code that used void* copied the CombatDamage object, however, I replaced the copy with a reference because I didn't expect side-effects.

ranisalt:
I suggest using a copy parameter instead of passing pointers, if possible.

PR #1600, opened 2015-10-26
Added / Improved items (maxHitChance)

raymondtfr:
These newer **maxHitChance** values were gathered from Tibia-Stats.com calculators.

* Updated some existing **`maxHitChance`** tags with new values.
* Added missing **`maxHitChance`** tags in a few one-handed distance weapons -- *this one is a critical update since the maxHitChance from one-handed distance weapons and two-handed distance ammo varies with the spear/ammo type, so that this prevents* **`WeaponDistance::useWeapon()`** *from loading the same maxHitChance to same weapon type (not taking into account that they may differ).*

### DIFF
--- a/data/items/items.xml
+++ b/data/items/items.xml
@@ -425,6 +425,7 @@
 		<attribute key="attack" value="5" />
 		<attribute key="weaponType" value="distance" />
 		<attribute key="shootType" value="smallstone" />
+		<attribute key="maxHitChance" value="76" />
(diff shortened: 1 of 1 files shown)


kornholi:
where did you get these numbers from?

raymondtfr:
Oh, I just coincidentally updated my post with the info you asked me xD.

kornholi:
Thanks!

PR #1603, opened 2015-10-27
Add time-based one-time password support

ranisalt:
This adds support for the one-time password support introduced in a recent Tibia version. I opted for the time-based since it is much more widely used.

It is possible to configure the life span and the length of tokens, but the user won't be able to use Google Authenticator, only other solutions such as Authy that support different time and length parameters.

You can use a third party tool to generate QR codes to the users to automatically provide the (encrypted) secret, or just hand them the secret to enter in token generation apps. I will provide a reference implementation soon.

If one wants to use counter-based tokens, it would be necessary an extra column at the account table for the use count, and replacing the "time" with this new column value. Don't forget to increase this counter on every use!

Last but not least: the token is *optional*, just null the column and the token will not be checked. If there is something other than null, then the token will be *required* for that account.

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -52,12 +52,40 @@ bool IOLoginData::saveAccount(const Account& acc)
 	return Database::getInstance()->executeQuery(query.str());
 }
 
+std::string decodeSecret(const std::string& secret) {
+	// simple base32 decoding
+	std::string key;
+	key.reserve(10);
+
+	uint32_t buffer = 0, left = 0;
+	for (const auto& ch : secret) {
+		buffer <<= 5;
+		if (ch >= 'A' && ch <= 'Z') {
+			buffer |= (ch & 0x1F) - 1;
+		} else if (ch >= '2' && ch <= '7') {
+			buffer |= ch - 24;
+		} else {
+			// if a key is broken, return empty and the comparison
+			// will always be false since the token must not be empty
+			return {};
+		}
+
+		left += 5;
+		if (left >= 8) {
+			left -= 8;
+			key.push_back(static_cast<char>(buffer >> left));
+		}
+	}
+
+	return key;
+}
+
 bool IOLoginData::loginserverAuthentication(const std::string& name, const std::string& password, Account& account)
 {
 	Database* db = Database::getInstance();
 
 	std::ostringstream query;
-	query << "SELECT `id`, `name`, `password`, `type`, `premdays`, `lastday` FROM `accounts` WHERE `name` = " << db->escapeString(name);
+	query << "SELECT `id`, `name`, `password`, `secret`, `type`, `premdays`, `lastday` FROM `accounts` WHERE `name` = " << db->escapeString(name);
 	DBResult_ptr result = db->storeQuery(query.str());
 	if (!result) {
 		return false;
@@ -69,6 +97,7 @@ bool IOLoginData::loginserverAuthentication(const std::string& name, const std::
 
 	account.id = result->getNumber<uint32_t>("id");
 	account.name = result->getString("name");
+	account.key = decodeSecret(result->getString("secret"));
 	account.accountType = static_cast<AccountType_t>(result->getNumber<int32_t>("type"));
 	account.premiumDays = result->getNumber<uint16_t>("premdays");
 	account.lastDay = result->getNumber<time_t>("lastday");
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -52,7 +52,8 @@ bool IOLoginData::saveAccount(const Account& acc)
 	return Database::getInstance()->executeQuery(query.str());
 }
 
-std::string decodeSecret(const std::string& secret) {
+std::string decodeSecret(const std::string& secret)
+{
 	// simple base32 decoding
 	std::string key;
 	key.reserve(10);
--- a/src/protocollogin.cpp
+++ b/src/protocollogin.cpp
@@ -43,7 +43,7 @@ void ProtocolLogin::disconnectClient(const std::string& message, uint16_t versio
 	disconnect();
 }
 
-void ProtocolLogin::getCharacterList(const std::string& accountName, const std::string& password, uint16_t version)
+void ProtocolLogin::getCharacterList(const std::string& accountName, const std::string& password, const std::string& token, uint16_t version)
 {
 	Account account;
 	if (!IOLoginData::loginserverAuthentication(accountName, password, account)) {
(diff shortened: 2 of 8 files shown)


ranisalt:
Here's a video:
[Token authentication](http://www.youtube.com/watch?v=gBGaf7FRWWo)

PrinterLUA:
Wow cool, great job :+1:

Nu77:
:+1:

mattyx14:
it is ... nice

src/protocollogin.cpp
```diff
@@ -179,6 +192,20 @@ void ProtocolLogin::onRecvFirstMessage(NetworkMessage& msg)
 	}
 
 	std::string password = msg.getString();
+	if (password.empty()) {
+		disconnectClient("Invalid password.", version);
+		return;
+	}
+
+	// read authenticator token and stay logged in flag from last 128 bytes
+	msg.setBufferPosition(msg.getLength() - 128);
```
HeavenIsLost: Client always send it? i have get the impression that you can't login without token anymore.

TheSumm: There is always a token, if you do not enter one it is a empty string.
https://github.com/edubart/otclient/blob/master/modules/gamelib/protocollogin.lua#L116

src/iologindata.cpp
```diff
@@ -52,12 +52,40 @@ bool IOLoginData::saveAccount(const Account& acc)
...
+		} else {
+			// if a key is broken, return empty and the comparison
+			// will always be false since the token must not be empty
+			return {};
+		}
+
+		left += 5;
+		if (left >= 8) {
+			left -= 8;
+			key.push_back(static_cast<char>(buffer >> left));
```
djarek: I know I'm being annoyingly pedantic, but right shift on a signed integer has implementation-defined behavior. You're also making assumptions about bit-length of int which is implementation defined as well.

src/tools.cpp
```diff
@@ -186,6 +186,50 @@ std::string transformToSHA1(const std::string& input)
...
+
+	// calculate hmac offset
+	uint32_t offset = static_cast<uint32_t>(std::stol(message.substr(39, 1), nullptr, 16) & 0xF);
+
+	// get truncated hash
+	uint32_t truncHash = std::stol(message.substr(2 * offset, 8), nullptr, 16) & 0x7FFFFFFF;
+	message.assign(std::to_string(truncHash));
+
+	// return only last AUTHENTICATOR_DIGITS (default 6) digits.
+	message.assign(message.substr(std::max(static_cast<std::string::size_type>(0), message.length() - AUTHENTICATOR_DIGITS), AUTHENTICATOR_DIGITS));
```
djarek: You should move the first argument of std::max to a separate line for better readability.

src/iologindata.cpp
ranisalt: Good eye, I really missed it. Changed to uint32_t for both.

ColorsCL:
Everytime I put a invalid (random) verification token the server just crashes... Is this just happening to me? @ranisalt

ranisalt:
@ColorsCL yes, I guess. As you can see in the video I posted, I tried an invalid token and I just got a message on the client, but the server is ok.

Could you run on gdb to check where exactly the server crashes? Don't forget to compile with debug flags.

ColorsCL:
I used the Visual Studio debugger and this is where the server crashes:
> uint32_t truncHash = std::stol(message.substr(2 * offset, 8), nullptr, 16) & 0x7FFFFFFF;

ranisalt:
Rebased against master.

@ColorsCL could you try again, possibly with an updated version of VS?

src/iologindata.cpp
```diff
@@ -52,12 +52,40 @@ bool IOLoginData::saveAccount(const Account& acc)
 	return Database::getInstance()->executeQuery(query.str());
 }
 
+std::string decodeSecret(const std::string& secret) {
```
marksamman: Move the opening curly bracket to a new line.

src/iologindata.cpp
```diff
@@ -52,12 +52,40 @@ bool IOLoginData::saveAccount(const Account& acc)
...
+	for (const auto& ch : secret) {
+		buffer <<= 5;
+		if (ch >= 'A' && ch <= 'Z') {
+			buffer |= (ch & 0x1F) - 1;
+		} else if (ch >= '2' && ch <= '7') {
+			buffer |= ch - 24;
+		} else {
+			// if a key is broken, return empty and the comparison
+			// will always be false since the token must not be empty
+			return {};
```
marksamman: Is this equivalent to calling std::string default constructor? I'm not familiar with this syntax.

diath: @marksamman Yes it deduces the type from function prototype (std::string) and calls the constructor via initializer list (https://ideone.com/SckVus).

ranisalt: It's C++11 syntax, instead of `return ns::thing(arg1, arg2, arg3...)`  you can just `return {arg1, arg2, arg3...}` provided that the return type is `ns::thing`.

ranisalt:
Rebased and fixed.

@marksamman is there any intent to support TOTP or does this PR serve as proof of concept only?

src/iologindata.cpp
marksamman: This still needs to be moved to a new line.

marksamman:
I'd actually like to merge this, there's only a small code style issue left.

ranisalt:
My bad, didn't see it. Fixed!

marksamman:
Thanks!

ranisalt:
Here's a guide on how to use it, may you want to include in the wiki:

-

Two factor authentication adds another layer of security to an account. A
[pseudo-random token](https://tools.ietf.org/html/rfc6238) based on time and a secret is generated and must be input
for a login to be valid, as an additional ephemeral password.

A secret is a 10 character, Base32 ([RFC 3548](https://tools.ietf.org/html/rfc3548)) encoded string, thus resulting in
16 characters. This secret is shared only between the server and the account
owner, generally hidden in a QR code that is read by a tool such as Google
Authenticator or Authy, which will generate the tokens.

This secret is **NOT** created by the server, but rather should be generated by
whatever you use to create accounts (i.e. your AAC). To generate the secret,
simply generate 16 characters from the Base32 alphabet from the RFC mentioned
above. Example given, in Lua:

```lua
-- prologue
local alphabet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ234567'

-- generating secret
function generateSecret()
    local output = {}
    for i = 1, 16 do
        local digit = math.random(1, alphabet:len())
        output[i] = alphabet:sub(digit, digit)
    end
    return table.concat(output)
end
```

You should store this secret in the `secret` column of the account table, and
this column will be created automatically on server startup if you have older
database version.

That said, you need to generate an `otpauth://` URL that will be converted to
a QR code or handled straight to the user (consider giving this option for
power users). Generate the URL as following:

```lua
local url = 'otpauth://totp/%s:%s?secret=%s&issuer=%s'
function generateURL(accountName, serverName)
    local secret = generateSecret()
    return url:format(serverName, accountName, secret, serverName)
end
```

Provide account name and server name accordingly. You can also pass the owner
name or email as the last parameter (`issuer`). It will generate an URL like
this:

```
otpauth://totp/Forgotten:ranisalt?secret=FQRBV6IAOR4NQECI&issuer=Forgotten
```

And it can be converted into a QR code using whatever tools your language or
framework provides. Show it to the user so he can add to a token generator and
use the token.

https://imgur.com/vatNyuz

**DON'T FORGET** that users might lose their token generator. Provide a way to
log in without using a token (such as one-time use tokens that disable secret)
so people don't lose accounts for being clumsy.

PR #1613, opened 2015-11-09
Use correct block type for ethereal spells

raymondtfr:
The *correct* **block type** for these two specific paladin**¹** spells:
*Ethereal Spear* and *Strong Ethereal Spear* -- is ``BLOCKSHIELD``

This applies to damage dealt by distance weapons too. Do you think it is appropriate to make such change now?

**¹** *The damage reduction for those spells only take into account the
target's defense.*

research source: ts

### DIFF
--- a/data/spells/scripts/attack/ethereal spear.lua
+++ b/data/spells/scripts/attack/ethereal spear.lua
@@ -2,7 +2,7 @@ local combat = Combat()
 combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_PHYSICALDAMAGE)
 combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_HITAREA)
 combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ANI_ETHEREALSPEAR)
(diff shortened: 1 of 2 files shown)


marksamman:
Yes, it would be appropriate to make that change to distance weapons too, but this PR is simple enough to be merged without waiting for that additional change. Thanks!

raymondtfr:
@marksamman I took sometime now (right after pushing new PR, aha) to revisit/review (again) the research page, distance damage calculator, spell damage calculator, and I found out there are some inconsistencies between the info provided at the research page, and the one provided by the calculators.
**The research page says:** *"As I mentioned in my previous article at [...], distance weapons ignore the shield and shielding skill of an opponent. Only the targets armor affects the damage you do to it."*
But I just rechecked the spell/distance damage calculator and I found out that the damage is reduced only by the target's defense.
So I guess it slipped through their minds all those years to update those calculators to reduce the damage by the target's armor (according to the research page) -- but instead both calculators reduces damage by the target's defense.

Due to those inconsistencies I think that you could be *revert* this PR.

[distance fighting research page:] (http://www.tibia-stats.com/index.php?akcja=6003&research=distance)

Issue #1625, opened 2015-11-30
Using math.random() in addDamage always generates the same value after first spell execution


raymondtfr:
If I do:

``` lua
condition:addDamage(1, 5000, -math.random(1, 10))
```

it gives a random number at the first time I execute the spell, but afterwards it keeps giving the same random value, it really doesn't randomize.
If I do `/reload spells` it does give another value.

I have tested this with latest nightly build.

diath:
That's expected behavior since you're basically assigning the result of math.random.

LuigiLazzari:
You need to run math.randomseed() once before using math.random(), like this:
```lua
math.randomseed(os.time())
```
One possible problem is that the first number may not be so "randomized" in some platforms. So a better solution is to pop some random number before using them for real:
```lua
math.randomseed(os.time())
math.random(); math.random(); math.random()
```

ranisalt:
@LuigiLazzari your statement is false. "Popping some random number" does not increase randomness as the sequence is always the same, you just shift it forward by a constant number of times.

And IIRC the engine already seeds the (pseudo) random number generator, so it is not needed to seed it again on the script, it would just reduce the randomness as the number of numbers popped would be reset, instead of increasing from every single use of math random on the entire server.

raymondtfr:
@diath I think I got a grasp of what you said. But then I also made some functions calls to generate random  values but it won't randomize.
@LuigiLazzari That's the second thing I tried, but it seems to me that diath is right, I guess I can't think of anything else to fix it lol. Could this be ``enhanced`` in C++? (reset addDamage values each different spell execution)?

gugahoa:
You're calling :addDamage only one time, that's when you're loading spells. To make it truly random, you need to assign :addDamage every time the spell is cast (onCast),

It's not an issue with the source, but with how you're doing the script.

kornholi:
Right, can't you just do something like this?

```lua
local combat = Combat()
combat:setParameter(...)

function onCastSpell(creature, var)
        local condition = Condition(...)
        condition:addDamage(1, 5000, -math.random(1, 10))

        combat:setCondition(condition)

	return combat:execute(creature, var)
end
```

I'm not sure if creating new conditions on cast would cause any performance issues.

raymondtfr:
Weird, it did work a few times but then it gives the same value over and over. (even setting randomseed);

ranisalt:
[...]

PR #1635, opened 2015-12-09
Enhanced utori-like spells & soul fire rune

raymondtfr:
All the spells have been researched in the most recent test server.
The formulas for **rounds**, **time** _(if not constant)_, and the **damage** _(if not constant)_  are very close to vanilla ones.

- [x] **Soul Fire rune**
- [x] **Holy Flash** *"utori san"*
- [x] **Electrify** *"utori vis"*
- [x] **Ignite** *"utori flam"*
- [x] **Envenom** *"utori pox"*
- [x] **Inflict Wound** *"utori kor"*
- [x] **Curse** *"utori mort"*

All codes have now been moved from ``onCastSpell`` to ``onTargetCreature`` callback function (which fixed the #1625 and is also perfomance-wise.)
Some remaining issues have been solved in #1825 with the help of other contributors.

### DIFF
--- a/data/spells/scripts/attack/envenom.lua
+++ b/data/spells/scripts/attack/envenom.lua
@@ -3,10 +3,15 @@ combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_EARTHDAMAGE)
 combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_CARNIPHILA)
 combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ANI_EARTH)
 
-local condition = Condition(CONDITION_POISON)
-condition:setParameter(CONDITION_PARAM_DELAYED, true)
-condition:addDamage(25, 4000, -45)
-combat:setCondition(condition)
+function onTargetCreature(creature, target)
+	local min = (creature:getLevel() * 0.01) + (creature:getMagicLevel() * 0.55) + 6
+	local max = (creature:getLevel() * 0.01) + (creature:getMagicLevel() * 0.75) + 7
+	local damage = math.random(math.floor(min), math.floor(max))
+	creature:addDamageCondition(target, CONDITION_POISON, 1, target:isPlayer() and math.floor(damage / 2 + 0.5) or damage)
+	return true
+end
+
+combat:setCallback(CALLBACK_PARAM_TARGETCREATURE, "onTargetCreature")
 
 function onCastSpell(creature, variant)
 	return combat:execute(creature, variant)
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -248,3 +248,45 @@ CORPSES = {
 
 -- This array contains all destroyable field items
 FIELDS = {1487,1488,1489,1490,1491,1492,1493,1494,1495,1496,1500,1501,1502,1503,1504}
+
+-- The numbered-keys represents the damage values, and their table
+-- contains the minimum and maximum number of rounds of those damage values.
+RANGE = {
+[1]  = {19, 20}, [2]  = {10, 10}, [3]  = {6, 7}, [4]  = {4, 5}, [5]  = {3, 4},
+[6]  = {3, 4},   [7]  = {2, 3},   [8]  = {2, 3}, [9]  = {2, 3}, [10] = {1, 2},
+[11] = {1, 2},   [12] = {1, 2},   [13] = {1, 2}, [14] = {1, 2}, [15] = {1, 2},
+[16] = {1, 2},   [17] = {1, 2},   [18] = {1, 2}, [19] = {1, 2}
+}
+
+function Creature:addDamageCondition(target, conditionType, listType, damage, time, rounds)
+	local condition = Condition(conditionType)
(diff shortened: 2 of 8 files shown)


marksamman:
The random value is evaluated when the script is loaded, which I don't think is intended. I'm not sure if it's fixable without leaking memory with the current way conditions work though.

raymondtfr:
Oh, thanks for the heads-up. I guess you meant to explain this in #1625 .

marksamman:
Yes, we should address that first.

djarek:
#1645 provides facilities for creating spells with "dynamic" conditions.

data/spells/scripts/attack/envenom.lua
```diff
@@ -3,10 +3,69 @@ combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_EARTHDAMAGE)
...
+	[10] = {1, 2},
+	[9] = {2, 3},
+	[8] = {2, 3},
+	[7] = {2, 3},
+	[6] = {3, 4},
+	[5] = {3, 4},
+	[4] = {4, 5},
+	[3] = {6, 7},
+	[2] = {10, 10},
+	[1] = {19, 20}
```
ranisalt: Why don't just write backwards? Also it looks very much like a function, I'll try to find it.

raymondtfr:
Because I chose to follow the same "descendant order" pattern as with the damage values. 

Yes, those rounds values are constant, and they are used in conditions (such as Bleeding and Poison) in vanilla, so they might have achieved these values through a function (or not), although IMO it is better written as a table and with the respective values for each damage, as users can easily modify them.

I initially thought about writing a simple function for it though, so we could load them from **(global.lua ?)**.

data/spells/scripts/attack/envenom.lua
raymondtfr: Because I chose to follow the same "descendant order" pattern as with the damage values. 

Yes, those rounds values are constant, and they are used in conditions (such as Bleeding and Poison) in vanilla, so they might have achieved these values through a function (or not), although IMO it is better written as a table and with the respective values for each damage, as users can easily modify them.

I initially thought about writing a simple function for it though, so we could load them from **(global.lua ?)**.

ranisalt: Writing a function wouldn't prevent one to modify the output. You can easily modify the function to just `return 7` if you want. I think it gets cleaner, especially if other functions use the same table.

raymondtfr: In this case, don't you think it would be better if I'd add this table to spells.lua?

ranisalt: Surely. What I meant initially is that the values look very much something like `dmg(x) = x/20` (check the outputs), not that it should be its own function :stuck_out_tongue_closed_eyes:

raymondtfr: I thinking about doing it like this (in **lib/spells.lua**):
```lua
CONDITION = {
	poison = {
		[20] = {1, 1}, [19] = {1, 2}, [18] = {1, 2}, [17] = {1, 2},   [16] = {1, 2},
		[15] = {1, 2}, [14] = {1, 2}, [13] = {1, 2}, [12] = {1, 2},   [11] = {1, 2},
		[10] = {1, 2}, [9]  = {2, 3}, [8]  = {2, 3}, [7]  = {2, 3},   [6]  = {3, 4},
		[5]  = {3, 4}, [4]  = {4, 5}, [3]  = {6, 7}, [2]  = {10, 10}, [1]  = {19, 20}
	},

	bleeding = {
		[20] = {1, 1}, [19] = {1, 2}, [18] = {1, 2}, [17] = {1, 2},   [16] = {1, 2},
		[15] = {1, 2}, [14] = {1, 2}, [13] = {1, 2}, [12] = {1, 2},   [11] = {1, 2},
		[10] = {1, 2}, [9]  = {2, 3}, [8]  = {2, 3}, [7]  = {2, 3},   [6]  = {3, 4},
		[5]  = {3, 4}, [4]  = {4, 5}, [3]  = {6, 7}, [2]  = {10, 10}, [1]  = {19, 20}
	},

	cursed = {
		{rounds = 3, 2},
		{rounds = 2, 3},
		{rounds = 1, 4, 5, 6, 7, 8, 10, 12, 14, 17, 21, 25, 30, 36, 43, 52, 62, 74, 89, 107, 128, 154, 185, 222, 266, 319, 383, 460, 552, 662, 795, 954}
	}
}

In addition to this, I would add more values in Curse table.
```

raymondtfr:
@marksamman After so many changes seeking for improvements, I finally believe this PR is ready (as stated in my first post). Can you review this PR?

marksamman:
Looks good to me, but it will have to wait until after the 1.2 release.

raymondtfr:
I have now removed curse condition table from ``lib/spells.lua`` and instead I added a function (with shorten and improved code) which generates a damage list for the condition and adds them. The type of generated damage list can be specified in the first argument of the function. I have decided to use a number type (not a condition type) for the user to choose which damage list to generate, since it would be limiting if I'd chosen to check the condition type or similar.

marksamman:
Thanks!

PR #1636, opened 2015-12-09
Update Weapons::getMaxWeaponDamage() formula

raymondtfr:
This formula is based on *tibia-stats.com* Melee / Distance Damage Calculator. Check the links below.
[distance damage calculator] (http://www.tibia-stats.com/index.php?akcja=distanceCalc)
[melee damage calculator] (http://www.tibia-stats.com/index.php?akcja=meleDmgCalc)

**Level** = 100
**Weapon skill** = 100
**Weapon attack** = 50

| **Attack Factor** | **New Formula** | **Current Formula** | **Diff. range** |
|:------------:|:---------------:|:-----:|:-----:|
| **Offensive** | 466 | 443 | 23 |
| **Balanced**  | 392 | 370 | 22 |
| **Defensive** | 243 | 222 | 21 |
Uses [t-s] (http://www.tibia-stats.com/) rounding method.

It is written like this:
* round( (level / 5) + (damage / factor) )

``update``
I've been doing some rigorous researches and, unfortunately, ts' calculators are not up to date. Therefore, I have updated the formula constant and multiplier to address this.
This is the reason:
* I dealt a damage of 469 hitpoints (so far it was 463, and in tibia this is a huge amount of luck) to a Dragon, which has a armor value of 25 (or 24), and a defense value of 38, and it has neutral resistance to physical damage. Considering that (it is indeed) the minimum armor reduction for a armor value of 25 is 12 (floor(25 / 2) or 24 / 2), and the minimum defense reduction for a defense value of 38 is 19 (38 / 2), then the maximum damage my character could've dealt is: 469 + minimum armor reduction + minimum defense reduction = 500 (no reduction), whereas ts gives me 496 (no reduction), and if applied the reductions, it equals 465.
* This considers the damage dealt was indeed reduced with the min. arm./def. reduction, if this was not true, then the hit could even be higher than 469, hence higher than 500.

**Character skills**
* level = 160, skill = 103, attack = 51

### DIFF
--- a/src/weapons.cpp
+++ b/src/weapons.cpp
@@ -142,7 +142,7 @@ int32_t Weapons::getMaxMeleeDamage(int32_t attackSkill, int32_t attackValue)
 //players
 int32_t Weapons::getMaxWeaponDamage(uint32_t level, int32_t attackSkill, int32_t attackValue, float attackFactor)
 {
-	return static_cast<int32_t>(std::ceil((2 * (attackValue * (attackSkill + 5.8) / 25 + (level - 1) / 10.)) / attackFactor));
+	return static_cast<int32_t>(std::round((level / 5) + (((((attackSkill / 4.) + 1) * (attackValue / 3.)) * 1.03) / attackFactor)));
 }
 
 Weapon::Weapon(LuaScriptInterface* _interface) :


HeavenIsLost:
But does it make it more like global?

ranisalt:
Those magic numbers make this formula stink. Do you have the source for it?

raymondtfr:
@HeavenIsLost Yes, it does makes it feel more like global. This formula is based on tibia-stats Melee / Distance Damage Calculators. I also think that I might have discovered their **exact** formula. The only left thing to do to make it **really** look like tibia is to implement a  critical damage chance formula similar to theirs (the more skill/attack you have, the more hard it is to do critical damages.). Right now, in TFS, it is very easy to do such critical hits (knights/pallies precisely).

@ranisalt Oh sorry, I mentioned the source as T-S, an abbreviation of tibia-stats.com.
Those *magic numbers* have their own meaning. Mathematically speaking, it means that **each 4 skill levels** your damage will **accumulate by approximately 1** ``("(attackSkill / 4.")`` plus 0.981 ``("(attackSkill / 4.) + 0.981")``. Also **each 3 attack values** your damage will **accumulate by approximately 1** ``("(attackValue / 3.")``. After this, the whole damage (excluding the ``(level / 5)`` part) will be divided by the attack factor, meaning that ``(level / 5)`` does not get modified by the attack factor. And the last thing the formula does is ``(level / 5)``, which means that **each 5 levels the player advances, it adds *exactly* (truncates) 1 to the whole damage**, i.e. ``(level / 5) + (damage / factor)``.
This expression ``(level / 5)`` is well known in Tibia since the PvP balance after 7.x~ updates.

ranisalt:
I meant this 0.981 and the 1.02 are very strange and don't seem to fit in a formula. Dividing by 4, 5 and such integers is probably fine, but where those two floats come from?

raymondtfr:
That's the whole idea of *discovering the formula*, you can call that reverse mathematics as there is reverse engineering or whatsoever lol. 1.02 is a generic multiplier -- 0.981 is a generic number that is added altogether to the previous arithmetic expression.

Code this formula to use in lua interactive mode or to a simple c++ console program, and compare the generated values with this:
Take a look at here: [melee damage calculator] (http://www.tibia-stats.com/index.php?akcja=meleDmgCalc)

Test all bunch, range of values (from level, to skill, and attack) and you'll understand.

ranisalt:
I'm not sure. You divide by the attack factor, but that means if a player is under defensive mode, it divides the major part of the attack by 0.5, effectively doubling damage. Shouldn't it be multiplied instead of divided?

raymondtfr:
You confused yourself. The current attack factors in the source code are:
```C++
float Player::getAttackFactor() const
{
	switch (fightMode) {
		case FIGHTMODE_ATTACK: return 1.0f;
		case FIGHTMODE_BALANCED: return 1.2f;
		case FIGHTMODE_DEFENSE: return 2.0f;
		default: return 1.0f;
	}
}
```
In this case, that part of the formula has to be divided, hence the damage **will not** *get doubled*. By the way, the current code also happens to use the same attack factors as tibia-stats's one, that's why I didn't change them in this PR.

We can make any random formula by placing a bunch of divisors, multipliers, additions (but with no meaning regarding the aspects in the game) that may generate values closer to the ones provided by t-s' calculators, though at some point, it won't be a match for input values (level, skill, attack).
Just to clarify, I didn't *reversed engineered* their formulas, I just discovered them *empirically*, then with enough *tests* I *mathematically reversed* all the operations into a formula (I guess this is a better way to put it, haha).

The formula tibia uses for melee/distance weapon damages is not complex, their formula of *critical hit chance* might be.

Try those values and see if they match tibia-stats' formula (I'm using *level = 0 and factor = 1* so we can verify with more accuracy).
Level = 0, Skill = 999, Attack = 999
Level = 0, Skill = 555, Attack = 555
Level = 0, Skill = 300, Attack = 600
** The formula you provided.
```lua
function g(level, skill, atk, factor)
	return math.floor((level/5) + ((skill + 4) * atk * factor * 0.085))
end
```
** The formula I made based on t-s.
```lua
function f(level, skill, attack, factor)
	local max = math.floor(level / 5) + ((((skill / 4) + 0.981) * (attack / 3)) * 1.02) / factor
	return math.floor(max + 0.5)
end
```
``g(0, 999, 999, 1); 85169`` does not match ``t-s = f() = 85163``
``g(0, 555, 555, 1); 26370`` does not match ``t-s = f() = 26367``
``g(0, 300, 600, 1); 15504`` does not match ``t-s = f() = 15500``

With the formula I made it will always match t-s' values, ranging from 0 to 999 (their calculator's max input) to the level, skill or attack.

ranisalt:
You are right, I completely missed this factor part.

raymondtfr:
Actually it **does not** curve off by at most 1 point (considering multiplying by factor = 1, which is the same if divided by factor = 1). With your formula it can go up to 6 points off.

Not everything in that formula is considered empirical, the ``level / 5`` can be easily proven in the game.
My intention here was creating a formula from a **reliable** source. Unfortunately, the range of values you wrote above for level, skill, and attack is a bit out of sense, since players can reach levels above 700, skills above 120 (you forgot considering strengthening spells), and the same goes for the weapon attack.
Since this is not tibia, the players can surpass any level, skill, or attack (taking into account the variables type sizes).

But most importantly, this formula is based on of the most reliable source of statistics (and believe me, it is accurate, because I've tested them [if you know to apply the armor/defense reduction correctly]).

By the way, this is part of the job I've been making to improve the PvP balance in TFS (so far I managed to update potions healing, spell damage). The only left thing I have to do to finish our PvP balance completely based on t-s is the armor/defense formula (done), this PR, maxHitChance formula for distance weapons (done), weapon's element damage (done).

kornholi:
What about `level / 5 + (0.333 * attackValue + 0.085 * attackValue * attackSkill) / attackFactor`?

This should give the same numbers as @raymondtfr but looks a lot simpler

raymondtfr:
@kornholi The formula you provided generate values with more accuracy than ranisalt 's one, but I think it is a little redundant to use the same variable ``attackValue`` twice. Also using ``attackValue / 3.`` instead of ``0.333 * attackValue`` gives more matchable values with t-s' one and mine (both of which rounded down).
I have made some tests, and ranisalt's one has a damage off of [0, 8] -- on the other hand, yours and mine is [0, 1] damage off.
When rounding down, in most of the cases, the formula I made ([0, 1] damage off) provides the same values as t-s' calculators.

HeavenIsLost:
@raymondtfr @marksamman @ranisalt  any update on this?

raymondtfr:
There is not a thing I can do (mathematically speaking) to improve this any further.

I guess the thing is that TFS (ots) has been using the current formula for a long time and people never complained about it, but since the changes I have made to spells, potions to make it more like vanilla (which doesn't means it is balanced), I guess this could be a plus for those changes.
Also the discussion stuck at the expression readability.

I also plan in the future to make a major update in monster files regarding their strength (armor, defense, resistances) to make them more balanced, because right now it is weird to see some monsters such as ***Abyssador*** with (armor = 15, defense = 20).

So I think **marksamman** and **kornolio** are deciding if it is a worth change.

marksamman:
Thanks!

PR #1628, opened 2015-12-04
Reward Boss implementation faithful to Tibia's one.

socket2810:
## Reasoning
After seeing a few reward systems on [OTLand](https://otland.net/) try and fail to replicate the original one, I decided to give it a try myself. Some images of the result:
Boss Loot Reward Chest       Reward Chest 2

### Boss Corpses
The difference between a normal monster corpse and a corpse that should yield individual rewards to players that participated in the fight is its attributes. **ITEM_ATTRIBUTE_CORPSEOWNER** is set to 2^31-1 and **ITEM_ATTRIBUTE_DATE** stores the time it was killed. 
### New class: Reward
It's the *Reward* class responsibility to store the different rewards a player may receive and, just like a depot chest, its contents are stored into the database when the player logs out and again loaded when they log back.
### New class: Reward Chest
Analogous to the Depot Locker, the *Reward Chest* contains the *Reward* (that inside the game is a [reward container](http://tibia.wikia.com/wiki/Reward_Container). However, it contains all the *Rewards*. They are inserted into the chest during the player load phase and every time another *Reward* is created via:
````
Player::getReward(uint32_t timestamp, bool autoCreate);
// or
Player.getReward(self, timestamp, autoCreate)
````

### Timestamps
Timestamps are set both on the boss corpse and on the in-game *reward container* / *Reward* object. They're used to identify the *Reward* and to limit how long the reward is stored in the database (Tibia chose a maximum of 7 days).
### How is each player attributed a different ~~corpse~~reward container?
~~The *Player* class now holds a *rewardCorpses* map that stores each corpse, which are instantiated the first time that the player opens the body and are collected in the player destructor (together with the player's *Rewards*).~~ (Not anymore: 7db6344)
The boss now drops a *proxy* reward container, meaning it is **not** a *Reward* object, but simply an item with the same id. Whenever a player tries to open this container, he will actually open his own *Reward*, this way there is no need to store unique corpses objects for each player.
### How is the loot handled?
Each player's contribution is measured by an *onHealthChange* event. The player will be considered **active** in the fight if he is currently in the boss' **targetList** - or whenever he heals someone that is currently active. The contribution is kept stored even if the player disconnects during the fight. If during the loot distribution phase the player is offline, he'll still receive his reward (Lua implementation: [insertRewardBag(guid, timestamp, itemList)](https://github.com/socket2810/forgottenserver/blob/master/data/creaturescripts/scripts/boss.lua#L188)).
### How to enable reward distribution to a boss
[...]

PR #1641, opened 2015-12-16
Reward System [remake of #1628]

socket2810:
Remake of #1628 
## Reasoning

After seeing a few reward systems on [OTLand](https://otland.net/) try and fail to replicate the original one, I decided to give it a try myself. Some images of the result:
Boss Loot Reward Chest       Reward Chest 2
### Boss Corpses

The difference between a normal monster corpse and a corpse that should yield individual rewards to players that participated in the fight is its attributes. **ITEM_ATTRIBUTE_CORPSEOWNER** is set to 2^31-1 and **ITEM_ATTRIBUTE_DATE** stores the time it was killed. 
### New class: Reward

It's the _Reward_ class responsibility to store the different rewards a player may receive and, just like a depot chest, its contents are stored into the database when the player logs out and again loaded when they log back.
### New class: Reward Chest

Analogous to the Depot Locker, the _Reward Chest_ contains the _Reward_ (that inside the game is a [reward container](http://tibia.wikia.com/wiki/Reward_Container). However, it contains all the _Rewards_. They are inserted into the chest during the player load phase and every time another _Reward_ is created via:

```
Player::getReward(uint32_t timestamp, bool autoCreate);
// or
Player.getReward(self, timestamp, autoCreate)
```
### Timestamps

Timestamps are set both on the boss corpse and on the in-game _reward container_ / _Reward_ object. They're used to identify the _Reward_ and to limit how long the reward is stored in the database (Tibia chose a maximum of 7 days).
### How is each player attributed a different ~~corpse~~reward container?

~~The _Player_ class now holds a _rewardCorpses_ map that stores each corpse, which are instantiated the first time that the player opens the body and are collected in the player destructor (together with the player's _Rewards_).~~ (Not anymore: 7db6344)
The boss now drops a _proxy_ reward container, meaning it is **not** a _Reward_ object, but simply an item with the same id. Whenever a player tries to open this container, he will actually open his own _Reward_, this way there is no need to store unique corpses objects for each player.
### How is the loot handled?

Each player's contribution is measured by an _onHealthChange_ event. The player will be considered **active** in the fight if he is currently in the boss' **targetList** - or whenever he heals someone that is currently active. The contribution is kept stored even if the player disconnects during the fight. If during the loot distribution phase the player is offline, he'll still receive his reward (Lua implementation: [insertRewardBag(guid, timestamp, itemList)](https://github.com/socket2810/forgottenserver/blob/master/data/creaturescripts/scripts/boss.lua#L188)).
### How to enable reward distribution to a boss

Set the *_rewardboss_ flag to true on the monster .xml file:

``` xml
<flag rewardboss="1" />
```

To make a loot unique and awarded to the highest scoring player:

``` xml
<item id="5903" chance="100000" unique="1"/><!-- ferumbras' hat -->
```

And register the corresponding events by calling _Monster.setReward(self, enable)_ whenever a boss is summoned (e.g [place_monster.lua](https://github.com/socket2810/forgottenserver/blob/master/data/talkactions/scripts/place_monster.lua#L14)).
### Player contribution

Player contribution is currently calculated like this:

``` lua
local score = (playerDamageOut / totalDamageOut) + (playerDamageIn / totalDamageIn) + (player.Healing / totalHealing)
score = score / 3 -- normalize to 0-1
local expectedScore = 1 / participants
local lootFactor = 1
lootFactor = lootFactor / participants ^ (1 / 3) -- tone down the loot a notch if there are many participants
lootFactor = lootFactor * (1 + lootFactor) ^ (score / expectedScore) -- increase the loot multiplicatively by how many times the player surpassed the expected score
```

Each _lootBlock_ is therefore **lootFactor** times more probable to be awarded to a specific player. However it is far from being optimal if we want to achieve a result similar to Tibia's and _only_ with the community help we will manage that.


vankk:
Both corpose keep saying I'm not the owner :p 

reward

socket2810:
If you can't be targeted by monsters (i.e god) you will never be detected as active in the fight and won't be attributed a score, so no loot for you.

src/player.cpp
```diff
@@ -925,6 +931,49 @@ DepotLocker* Player::getDepotLocker(uint32_t depotId)
...
+	return reward;
+}
+
+void Player::removeReward(uint32_t rewardId) {
+	rewardMap.erase(rewardId);
+}
+
+std::vector<uint32_t> Player::getRewardList() {
+	std::vector<uint32_t> list;
+	for (auto& it : rewardMap) {
```
djarek: Use this: 
```
list.reserve(rewardMap.size());
```
More of a personal thing, not sure if others share my opinion: I dislike using container names like ```list```. The name should reflect what is stored in it so in this case ```rewards``` is the most straightforward name.
to reduce memory allocation overhead.

src/player.h
```diff
@@ -511,6 +513,11 @@ class Player final : public Creature, public Cylinder
 		void addConditionSuppressions(uint32_t conditions);
 		void removeConditionSuppressions(uint32_t conditions);
 
+		Reward* getReward(uint32_t rewardId, bool autoCreate);
```
djarek: Remove ```autoCreate``` from the arguments list - there's really no point in having it there as it basically just causes the function to return nullptr if false.

src/item.h
```diff
@@ -495,6 +495,13 @@ class Item : virtual public Thing
 			return getIntAttr(ITEM_ATTRIBUTE_CORPSEOWNER);
 		}
 
+		void setRewardCorpse() {
+			setCorpseOwner(static_cast<uint32_t>(std::numeric_limits<int32_t>::max()));
```
djarek: It's not clear here what owner uid value is supposed to mean. You should preferable put it into a typed contexpr(or const) with an appropriate name.

src/iologindata.cpp
```diff
@@ -470,6 +470,52 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
...
+				if (reward) {
+					it->second = std::pair<Item*, int32_t>(reward->getItem(), pid); //update the map with the special reward container
+				}
+			} else {
+				break;
+			}
+		}
+
+		//second loop (this time a reverse one) to insert the items in the correct order
+		for (ItemMap::const_reverse_iterator it = itemMap.rbegin(), end = itemMap.rend(); it != end; ++it) {
```
djarek: I'd rather see boost::reverse used here with a range-based for loop, i.e. 
```
for (const auto& itemPair : boost::adaptors::reverse(itemMap))
```

src/iologindata.cpp
```diff
@@ -470,6 +470,52 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
...
+	//load reward chest items
+	itemMap.clear();
+
+	query.str(std::string());
+	query << "SELECT `pid`, `sid`, `itemtype`, `count`, `attributes` FROM `player_rewards` WHERE `player_id` = " << player->getGUID() << " ORDER BY `sid` DESC";
+	if ((result = db->storeQuery(query.str()))) {
+		loadItems(itemMap, result);
+
+		//first loop handles the reward containers to retrieve its date attribute
+		for (ItemMap::iterator it = itemMap.begin(), end = itemMap.end(); it != end; ++it) {
```
djarek: Use range-based for loop:
```
for (const auto& itemPair : itemMap)
```

src/item.h
```diff
@@ -495,6 +495,13 @@ class Item : virtual public Thing
...
+		}
+		bool isRewardCorpse() {
+			return getCorpseOwner() == static_cast<uint32_t>(std::numeric_limits<int32_t>::max());
```
djarek: The same as above.

src/iologindata.cpp
ranisalt: We could have it in tools code instead of depending on boost for that. That's really matter of forwarding begin to rbegin and end to rend.

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,226 @@
...
+		local timestamp = os.time()
+
+		local totalDamageOut, totalDamageIn, totalHealing = 0.1, 0.1, 0.1 -- avoid dividing by zero
+
+		local scores = {}
+		local info = globalBosses[bossId]
+		local damageMap = creature:getDamageMap()
+
+		for guid, stats in pairs(info) do
+			local player = Player(stats.playerId)
```
Cirocc: If the player gives damage to the boss and die before the boss dies, the server crashes because the player does not think the solution I found was to add in ondeath function:
	                if not player then
		        return true
	                end

gunzino:
In game.cpp this causes segfault:
if (ret != RETURNVALUE_NOERROR && toCylinder->getItem()->getID() != ITEM_REWARD_CONTAINER) {

Should be changed to:
if (ret != RETURNVALUE_NOERROR && toCylinder->getItem() && toCylinder->getItem()->getID() != ITEM_REWARD_CONTAINER) {

src/player.cpp
ranisalt: +1 for a better name and no hungarian notation bs

src/player.h
socket2810: There is a point. Calling ````getReward(rewardId, false)```` won't give the player a reward if he doesn't have one already, i.e the player tries to open the boss' corpse but didn't help in the fight, if there was no ````autoCreate```` parameter an empty reward container would be presented to the player in the boss corpse and in the reward chest.

src/reward.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <iostream>
+#include "reward.h"
+
+
+Reward::Reward() :
+	Container(ITEM_REWARD_CONTAINER)
+{
+	maxSize = 32;
+	unlocked = true;
```
ranisalt: Set to false to have a different animation when a player tries to put something inside the bag.

src/reward.cpp
```diff
@@ -0,0 +1,60 @@
...
+		parent->postAddNotification(thing, oldParent, index, LINK_PARENT);
+	}
+}
+
+void Reward::postRemoveNotification(Thing* thing, const Cylinder* newParent, int32_t index, cylinderlink_t)
+{
+	Cylinder* parent = getParent();
+	if (parent != nullptr) {
+		parent->postRemoveNotification(thing, newParent, index, LINK_PARENT);
+	}
```
ranisalt: Shouldn't the bag be destroyed around here when it gets emptied?

src/rewardchest.cpp
```diff
@@ -0,0 +1,35 @@
+#include "otpch.h"
+
+#include "rewardchest.h"
+
+RewardChest::RewardChest(uint16_t _type) :
+	Container(_type)
+{
+	maxSize = 32;
+	unlocked = true;
```
ranisalt: `false`

src/item.h
socket2810: What it is supposed to mean is in the member name. Don't you think this looks redundant ?

````C++
void setRewardCorpse() {
    setCorpseOwner(REWARD_CORPSE_OWNER);
}
````

File: src/reward.cpp (hunk 8 above)
Comment (socket2810) [inline src/reward.cpp]: What do you mean?

File: src/reward.cpp (hunk 9 above)
Comment (socket2810) [inline src/reward.cpp]: If we destroy the bag here, ````rewardMap```` in the Player object will hold a dangling pointer to it.

**Comment by socket2810** [comment]
@Cirocc 
Pode conversar comigo em Português. Eu não consegui reproduzir esse bug aqui, consegue detalhar mais como ele ocorre? Um video ajudaria bastante.

@gunzino 
Thank you for catching that.

File: src/reward.cpp (hunk 8 above)
Comment (ranisalt) [inline src/reward.cpp]: When unlocked is set to false, the client renders a red line bordering the container slot when a player tries to move an item to that container. This is used for containers that are only for "ephemeral" storage, such as depot inbox and reward chests. It has nothing to do with `queryAdd`, though.

File: src/reward.cpp (hunk 9 above)
Comment (ranisalt) [inline src/reward.cpp]: Good eye, but when you remove the last item from the reward bag and close it, it is destroyed. I'm not sure if there is a callback for closing a container, if not, this would be pretty onerous to implement.

Comment (socket2810) [inline src/reward.cpp]: I know it is destroyed but as you just realized it is not a trivial "feature" to implement, hence I believe we should just leave it as it is - that is, empty reward containers aren't saved when the player logs off.

**Comment by Cirocc** [comment]
Se o player der dano no boss e morrer quando o boss morrer da core dump falha de memoria pois não encontra o player para entregar o reward,então simplesmente adicionei no creaturescripts a função de se não reconhecer o player,retornar normal no caso fico assim:
local player = Player(stats.playerId)
if not player then
return true
end

**Comment by socket2810** [comment]
@Cirocc 
Eu não aconselharia você a fazer isso, pois nesse caso ninguém receberia recompensa alguma. Infelizmente eu não consegui reproduzir esse crash de nenhuma maneira e, portanto, não posso ajudá-lo. Eu lhe aconselharia a utilizar a versão mais nova desse sistema que pode ser encontrado em minha [branch](https://github.com/socket2810/forgottenserver/tree/reward_system) e ver se o problema persiste.


##### For those who can't read Portuguese:
Cirocc claims the server is crashing whenever a boss dies and a participant had previously died/is dead (I can't specify from what he told me), however that makes no sense to me because the player object used in boss.lua comes from Player(stats.playerId) - stats.playerId being the last known id which is updated whenever the player becomes active in the fight or logs in - and if that playerId isn't valid, that function should return nil. I also couldn't reproduce this crash, so without further information I can't do anything.

**Comment by Cirocc** [comment]
Cara eu instalei o sistema não faz nem 2 dias  TFS 1.2 tudo perfeito sem erros somente foi isso mesmo que eu falei e resolveu não tem nada haver oque você falou,somente não vai receber se o player tiver morrido antes do boss você não entendeu mas tudo bem,com certeza esse problema é geral. :D

**Comment by socket2810** [comment]
@Cirocc 
If you are going to take the objective observations I made as a personal insult you shouldn't be discussing anything in here. This system isn't supposed to be bug-free, or work as you expect, that is why it is a pull request. People come here, like you did, to report bugs, discuss the pull or propose solutions to standing problems. If all you have to add to the discussion is salty comments, please, refrain from doing so.

**Comment by Cirocc** [comment]
Se fode ai então otário continue com seu sistema bugado,quem colocar essa merda esteja ciente que o servidor vai cair,estou tentando ajuda e você quer tudo na mão já disse o problema e a solução e você fica ai pedindo vídeo e sei la oque mais,quem tem que testar é você o sistema é seu eu só vim aqui avisa mais nada.

**Comment by ranisalt** [comment]
stop it kid

**Comment by vankk** [comment]
I know 118 people will be annoyed but I want make a comment about it.. 

That's why no one likes Brazil, by attitudes that neither that, asshole.

**Comment by BenDol** [comment]
Great contribution :+1

**Comment by degolaot** [comment]
Hello friends, this system is correct or crashing? I dont have test server for check ... Thanks

**Comment by degolaot** [comment]
Hello, thanks for system, but this is only working if We use talkactions for summon, Raids don't working the system, any fix?

File: data/creaturescripts/creaturescripts.xml [hunk 11]
```diff
@@ -8,4 +8,8 @@
 	<event type="death" name="PlayerDeath" script="playerdeath.lua" />
 	<event type="death" name="DropLoot" script="droploot.lua" />
 	<event type="extendedopcode" name="ExtendedOpcode" script="extendedopcode.lua" />
+
```
WibbenZ: Trailing spaces

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,226 @@
...
+	end
+end
+
+local function insertItems(buffer, info, parent, items)
+	local start = info.running
+	for _, item in ipairs(items) do
+		if _ ~= 1 or parent > 100 then
+			table.insert(buffer, ",")
+		end
+		info.running = info.running + 1
```
WibbenZ: New line here

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,226 @@
...
+	end
+	return info.running - start
+end
+
+local function insertRewardItems(playerGuid, timestamp, itemList)
+	db.asyncStoreQuery('SELECT `pid`, `sid` FROM `player_rewards` WHERE player_id = ' .. playerGuid .. ' ORDER BY `sid` ASC;', 
+		function(query)
+			local lastReward = 0
+			local lastStoreId   
+			if(query) then
```
WibbenZ: Remove the parenthesis

data/creaturescripts/scripts/login.lua
```diff
@@ -29,8 +29,21 @@ function onLogin(player)
 		player:setVocation(vocation:getDemotion())
 	end
 
+	-- Rewards notice
+	local rewards = #player:getRewardList()
+	if(rewards > 0) then
```
WibbenZ: Remove the parenthesis

data/lib/rewardboss.lua
```diff
@@ -0,0 +1,130 @@
...
+end
+
+function Item.getNameDescription(self)
+	local subType = self:getSubType()
+	local itemType = self:getType()
+
+	local buffer = {}
+
+	local name = self:getName() or ''
+	if(#name ~= 0) then
```
WibbenZ: Remove the parenthesis
Aswell on the line below

data/lib/rewardboss.lua
```diff
@@ -0,0 +1,130 @@
...
+	return table.concat(buffer)
+end
+
+function Container.getContentDescription(self, outputBuffer)
+	local firstItem = true
+	local buffer = outputBuffer or {}
+	for i = 1, self:getSize() do
+		local item = self:getItem(i - 1)
+
+		if(firstItem) then
```
WibbenZ: Remove the parenthesis

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,226 @@
...
+
+			--reward bag
+			local info = {
+				playerGuid = playerGuid,
+				running = lastStoreId or 100
+			}
+
+			local bag = Game.createItem(ITEM_REWARD_CONTAINER)
+			bag:setAttribute(ITEM_ATTRIBUTE_DATE, timestamp)
+			if itemList then
```
WibbenZ: New line here

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,226 @@
...
+				for _, p in ipairs(itemList) do
```
WibbenZ: Use a better name than "p"

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,226 @@
...
+end
+
+function onThink(creature, interval)
+	local bossId = creature:getId()
+	local info = globalBosses[bossId]
+	-- Reset all players' status
+	for _, player in pairs(info) do
+		player.active = false
+	end
+	-- Set all players in boss' target list as active in the fight
```
WibbenZ: New line

data/creaturescripts/scripts/logout.lua
```diff
@@ -3,5 +3,19 @@ function onLogout(player)
 	if nextUseStaminaTime[playerId] ~= nil then
 		nextUseStaminaTime[playerId] = nil
 	end
+
+	local stats = player:inBossFight()
+	if stats then
+		local boss = Monster(stats.bossId)
+		-- Player logged out (or died) in the middle of a boss fight, store his damageOut and stamina
```
WibbenZ: Move the comment up a line, so local boss if above if boss then

data/events/scripts/player.lua
```diff
@@ -180,3 +176,37 @@ function Player:onGainSkillTries(skill, tries)
 	end
 	return tries * configManager.getNumber(configKeys.RATE_SKILL)
 end
+
+function Player:onMoveItem(item, count, fromPosition, toPosition)
+	if toPosition.x == CONTAINER_POSITION then
+		local cid = toPosition.y - 64 -- container id
```
WibbenZ: Use containerId insted of cid, cid could also be creatureId

data/lib/rewardboss.lua
```diff
@@ -0,0 +1,130 @@
...
+
+function Player.getRewardChest(self, autocreate)
+	return self:getDepotChest(99, autocreate)
+end
+
+function Player.inBossFight(self)
+	if not next(globalBosses) then
+		return false
+	end
+	local playerGuid = self:getGuid()
```
WibbenZ: New line above

data/lib/rewardboss.lua
```diff
@@ -0,0 +1,130 @@
...
+			return stats
+		end
+	end
+	return false
+end
+
+-- by https://otland.net/members/cbrm.25752/ with some modifications
+function MonsterType.createLootItem(self, lootBlock, chance, lootTable)
+	local lootTable, itemCount = lootTable or {}, 0
+	local randvalue = math.random(0, 100000) / (getConfigInfo("rateLoot") * chance)
```
WibbenZ: Update both getConfigInfo calls to use meta insted.

data/talkactions/scripts/place_monster.lua
```diff
@@ -10,6 +10,9 @@ function onSay(player, words, param)
 	local position = player:getPosition()
 	local monster = Game.createMonster(param, position)
 	if monster ~= nil then
+		if monster:getType():isRewardBoss() then
+			monster:setReward(true)
+		end
 		monster:getPosition():sendMagicEffect(CONST_ME_TELEPORT)
```
WibbenZ: New line

src/iologindata.cpp
```diff
@@ -470,6 +472,54 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
...
+			}
+		}
+
+		//second loop (this time a reverse one) to insert the items in the correct order
+		//for (ItemMap::const_reverse_iterator it = itemMap.rbegin(), end = itemMap.rend(); it != end; ++it) {
+		for (const auto& it : boost::adaptors::reverse(itemMap)) {
+			const std::pair<Item*, int32_t>& pair = it.second;
+			Item* item = pair.first;
+
+			int32_t pid = pair.second;
```
WibbenZ: Change the variable name

src/iologindata.cpp
```diff
@@ -470,6 +472,54 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
...
+			if (pid >= 0 && pid < 100) {
+				break;
+			}
+			
+			ItemMap::const_iterator it2 = itemMap.find(pid);
```
WibbenZ: Change the variable name

src/iologindata.cpp
```diff
@@ -776,6 +826,33 @@ bool IOLoginData::savePlayer(Player* player)
 		}
 	}
 
+	auto rewardList = player->getRewardList();
```
WibbenZ: Move this to line 834 !db->executeQuery(query.str()
Waste to load it if the query statment returns false.

src/reward.h
```diff
@@ -0,0 +1,37 @@
+#ifndef FS_REWARD_H
+#define FS_REWARD_H
+
+#include "container.h"
+
+class Reward :
+	public Container
+{
+public:
```
WibbenZ: Indentation, move everything one tab in

src/rewardchest.h
```diff
@@ -0,0 +1,30 @@
+#ifndef FS_REWARDCHEST_H
+#define FS_REWARDCHEST_H
+
+#include "container.h"
+
+class RewardChest final : public Container
+{
+public:
```
WibbenZ: Indentation, move everything one tab in

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,229 @@
...
+
+				insertItems(buffer, info, info.running, subItems)
+			end
+		end
+	end
+	return info.running - start
+end
+
+local function insertRewardItems(playerGuid, timestamp, itemList)
+	db.asyncStoreQuery('SELECT `pid`, `sid` FROM `player_rewards` WHERE player_id = ' .. playerGuid .. ' ORDER BY `sid` ASC;',
```
WibbenZ: Variable names here aswell, they should mean something.
Is pid playerId and sid storageId?
No matter, change them to what they actually mean.

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,229 @@
...
+function onDeath(creature, corpse, killer, mostDamageKiller, lastHitUnjustified, mostDamageUnjustified)
+	local monsterType = creature:getType()
+	if monsterType:isRewardBoss() then -- Make sure it is a boss
+		local bossId = creature:getId()
+		local timestamp = os.time()
+
+		local totalDamageOut, totalDamageIn, totalHealing = 0.1, 0.1, 0.1 -- avoid dividing by zero
+
+		local scores = {}
+		local info = globalBosses[bossId]
```
WibbenZ: Place this above the loop, this way it's easier to find the "info".

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,229 @@
...
+				player = player, 
+				guid = guid,
+				damageOut = damageOut, 
+				damageIn = damageIn,
+				healing = healing,
+			})            
+		end
+
+		local participants = 0
+		for _, con in ipairs(scores) do
```
WibbenZ: con => score?

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,229 @@
...
+
+			local playerLoot
+			if --[[stamina > 840 and]] con.score ~= 0 then
+				local lootFactor = 1
+				lootFactor = lootFactor / participants ^ (1 / 3) -- tone down the loot a notch if there are many participants
+				lootFactor = lootFactor * (1 + lootFactor) ^ (con.score / expectedScore) -- increase the loot multiplicatively by how many times the player surpassed the expected score
+				playerLoot = monsterType:getBossReward(lootFactor, _ == 1)
+
+				if con.player then
+					for _, p in ipairs(playerLoot) do
```
WibbenZ: p => loot or item

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,229 @@
...
+			con.score = score / 3 -- normalize to 0-1
+			if score ~= 0 then
+				participants = participants + 1
+			end
+		end
+		table.sort(scores, function(a, b) return a.score > b.score end)
+
+		local expectedScore = 1 / participants
+
+		for _, con in ipairs(scores) do
```
WibbenZ: con => score?

data/creaturescripts/scripts/boss.lua
```diff
@@ -0,0 +1,229 @@
...
+					for _, p in ipairs(playerLoot) do
+						reward:addItem(p[1], p[2])
+					end
+				end
+			end
+
+			if con.player then
+				local lootMessage = {"The following items are available in your reward chest: "}
+
+				if --[[stamina > 840]]true then
```
WibbenZ: ???

data/creaturescripts/scripts/logout.lua
```diff
@@ -3,5 +3,19 @@ function onLogout(player)
 	if nextUseStaminaTime[playerId] ~= nil then
 		nextUseStaminaTime[playerId] = nil
 	end
+
+	local stats = player:inBossFight()
+	if stats then
+		-- Player logged out (or died) in the middle of a boss fight, store his damageOut and stamina
+		local boss = Monster(stats.bossId)		
+		if boss then
+			local dmgOut = boss:getDamageMap()[playerId]
```
WibbenZ: damage

data/lib/rewardboss.lua
```diff
@@ -0,0 +1,130 @@
...
+	globalBosses = {}
+end
+
+function Monster.setReward(self, enable)
+	if enable then
+		if not self:getType():isRewardBoss() then
+			error("Rewards can only be enabled to rewards bosses.")
+			return false
+		end
+		globalBosses[self:getId()] = {}
```
WibbenZ: New line

data/talkactions/scripts/place_monster.lua
```diff
@@ -10,6 +10,10 @@ function onSay(player, words, param)
 	local position = player:getPosition()
 	local monster = Game.createMonster(param, position)
 	if monster ~= nil then
+		if monster:getType():isRewardBoss() then
+			monster:setReward(true)
+		end
+
```
WibbenZ: Trailing spaces

src/iologindata.cpp
```diff
@@ -470,6 +472,54 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 		}
 	}
 
+	//load reward chest items
+	itemMap.clear();
+
+	query.str(std::string());
+	query << "SELECT `pid`, `sid`, `itemtype`, `count`, `attributes` FROM `player_rewards` WHERE `player_id` = " << player->getGUID() << " ORDER BY `sid` DESC";
+	if ((result = db->storeQuery(query.str()))) {
```
WibbenZ: if (result = db->storeQuery(query.str())) {

src/iologindata.cpp
```diff
@@ -470,6 +472,54 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
...
+		loadItems(itemMap, result);
+
+		//first loop handles the reward containers to retrieve its date attribute
+		//for (ItemMap::iterator it = itemMap.begin(), end = itemMap.end(); it != end; ++it) {
+		for (auto& it : itemMap) {
+			const std::pair<Item*, int32_t>& pair = it.second;
+			Item* item = pair.first;
+
+			int32_t pid = pair.second;
```
WibbenZ: Pid = ??

src/reward.cpp
```diff
@@ -0,0 +1,60 @@
+#include "otpch.h"
+
+#include <iostream>
+#include "reward.h"
+
+
```
WibbenZ: Delete the extra line

src/reward.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+	if (actor) {
+		return RETURNVALUE_NOTPOSSIBLE;
+	}
+
+	const Item* item = thing.getItem();
+	if (!item) {
+		return RETURNVALUE_NOTPOSSIBLE;		
+	}
+
```
WibbenZ: Trailing

src/reward.cpp
```diff
@@ -0,0 +1,60 @@
...
+	const Item* item = thing.getItem();
+	if (!item) {
+		return RETURNVALUE_NOTPOSSIBLE;		
+	}
+	
+	if (item == this) {
+		return RETURNVALUE_THISISIMPOSSIBLE;		
+	}
+
```
WibbenZ: Trailing

src/rewardchest.cpp
```diff
@@ -0,0 +1,35 @@
...
+	pagination = true;
+}
+
+ReturnValue RewardChest::queryAdd(int32_t, const Thing&, uint32_t,
+	uint32_t, Creature* actor/* = nullptr*/) const
+{
+	if (actor) {
+		return RETURNVALUE_NOTPOSSIBLE;
+	}
+
```
WibbenZ: Trailing

src/iologindata.cpp
socket2810: The same thing it means in every item storage table, [parent id](https://github.com/otland/forgottenserver/blob/master/src/iologindata.cpp#L424).

src/iologindata.cpp
socket2810: [link](https://github.com/otland/forgottenserver/blob/master/src/iologindata.cpp#L446)

data/creaturescripts/scripts/logout.lua
socket2810: damage made or damage taken?

data/creaturescripts/scripts/boss.lua
socket2810: Normally you can't loot items in Tibia without that much stamina, however I was informed this limit does not apply to bosses, I'm leaving it like that until someone confirms or denies this information.

data/creaturescripts/scripts/boss.lua
socket2810: pid and sid are already used in [player_depotitems](https://github.com/otland/forgottenserver/blob/master/schema.sql#L271), player_inboxitems and player_items. I'm just following the current standard.

data/creaturescripts/scripts/boss.lua
socket2810: I do not find the idiom score.score appealing.

src/iologindata.cpp
ranisalt: It should be named as `parent_id` then, both in the code and in the database.

src/iologindata.cpp
ranisalt: @WibbenZ compilers issue a warning when such attribution is not enclosed in parenthesis as it may solve some ambiguity problems.

data/creaturescripts/scripts/boss.lua
ranisalt: The inner `score` should be named `value` then since if you have a list of `scores` , every item is a `score` by itself.

src/iologindata.cpp
socket2810: Perhaps it should, but I'd rather sacrifice clarity in this case, instead of having: A) to change every other ocurrence of ````pid```` both in the code and in the database; B) ````player_rewards```` use ````parent_id```` and ````player_depotitems, player_inboxitems and player_items```` use ````pid````.

ranisalt: Yes, agreed, that's not an issue this PR should care for.

data/creaturescripts/scripts/boss.lua
WibbenZ: Oh okay, well then it should wait till we update the SQL schema.

data/creaturescripts/scripts/boss.lua
WibbenZ: Singular an plural.
Plural = the table
Singular = the row of the table

Items => item
Positions => position
etc

data/creaturescripts/scripts/boss.lua
WibbenZ: In that case add a comment above, and id comment out the if statment so when we know about it we can remove the comments.
One less thing to load for now.

data/creaturescripts/scripts/logout.lua
WibbenZ: damageTaken I suppose, just avoid writing names like p insted of player, it's fine if it's your own server since you coded it - you know what everything is.
But just try to convert 0.x scripts to 1.x, where all variables are named b_2 etc
As long as the name describes what is stored in the variable it's fine.

src/iologindata.cpp
WibbenZ: Oh nvm it then, can't remove the comment tho ..

src/iologindata.cpp
WibbenZ: Ye I didn't read everything, but as I wrote above you can nvm.
We can update the schema.

data/creaturescripts/scripts/boss.lua
vankk: Can someone confirm what @socket2810 said?

WibbenZ: Not a super easy thing to check xD
Best to try and talk to experienced players and or using wikia etc

Pretty rare (that atleast I) go below 30~hs, botters might but then they won't kill raid bosses xD

Mojmistrz:
@socket2810 How you can register raid monster in reward chest system? When I set rewardboss flag to 1 in monster file and then it is spawned by a raid, it won't yield any loot and corpse will say - 'You are not the owner.' . Bug or I have missed something?

socket2810:
@Mojmistrz 
You missed this:

> And register the corresponding events by calling Monster.setReward(self, enable) whenever a boss is summoned (e.g place_monster.lua).

You need to call `monster:setReward(true)` whenever a monster you want to yield rewards is spawned. You can use a raid script for that.

vankk:
@socket2810 
Why won't make a enhancement on _raids.cpp_ for it?

Mojmistrz:
@socket2810 
But, wouldn't it require to make this raid in lua? I have this in xml and I don't think there is possibility to put `monster:setReward(true)` there.

darkjav:
/rewardboss.lua:1: unexpected symbol near '�'

Hello, that appear in console, any idea how fix? Thanks.

darkjav:
@Mojmistrz 
`script="boss.lua"` in first line or later of:
`<?xml version="1.0" encoding="UTF-8"?>`

u can explian me?? Thanks.

WibbenZ:
```
<monster name="Ferumbras" nameDescription="Ferumbras" race="venom" experience="12000" speed="320" manacost="0" script="boss.lua">
```

lipix:
@socket2810 estou com um erro quando mato alguns outros monstros, tipo o rato ele diz que nao posso abrir pq nao sou dono e tambem nem mostra msg de loot.

data/lib/rewardboss.lua
```diff
@@ -0,0 +1,130 @@
...
+
+	return lootTable
+end
+
+function MonsterType.getBossReward(self, lootFactor, topScore)
+	local result = {}
+	if configManager.getNumber("rateLoot") > 0 then
+		for _, lootBlock in pairs(self:getLoot()) do
+			if lootBlock.unique and not topScore then
+				--continue
```
vankk: unique is not working?

darkjav: If you do all fine, not should have problems there, the code is very good!

Schiffers:
sorry the bump, but i'm getting two errors with functions:
isRewardCorpse() and setRewardCorpse()
the errors is:
the class container has not member isRewardCorpse() or setRewardCorpse()
how solve it, sorry for my bad knowledge, I have no time to study it by the base of the programming

darkjav:
@socket2810, @WibbenZ, @Mojmistrz Hello, i found a bug, when the flags and script are in a monster that respawn each "x" time, and you kill him for get the loot, this say: "You are not the owner." --I think that is a big problem with bosses that respawn like Jaul, Tanjis, etc...

I wait for you help, thanks!

ziomek280:
Hello, i have problem when i kill monster falls all loot.
I wait for you help, thanks.
images

ranisalt:
> Hello, i have problem when i kill monster falls all loot.

What is your loot rate and how many people killed the monster?

ziomek280:
Monster killing two people and two receive full loot.

ranisalt:
And the loot rate? How many x is the loot?

ziomek280:
2x full
images

Schiffers:
@ziomek280 aff bro, why you don't see the chances in the "monsters.xml" ?
¬¬
exemple, ferumbras, you need to set a small chance to rare items, exemple:
<itemid="xxx" chance"500"/>  go test..

ziomek280:
I have chance= 500 but all the loot falls :/

darkjav:
Nobody will answer to my question of 21 days ago? Plx!

kodzy15:
Why when I kill one boss with my character he creates the bag everything right but has not loot gives the message:
16:07 The Following items are available in your chest reward: nothing.
NOTE: I'm not killing with char GM.

thexamx:
Something is bugged with this system. For me it always drop full loot even if i change chances to 100-200 etc. Its always full for every player, expect when I put unique then only one player get the loot, but still all of these items i make unique...

Wasterl4nd:
I can't install, someone can send me all archives please ?

Wasterl4nd:
The system are bugged, are dropping all items

Schiffers:
for who get bugs with the chance, try:
replace
local lootFactor = 1
to 
local lootFactor = 0.01
and decrease the chance in the monsters.xml files, then you need to edit to fill your needs, I use this factor, chances < 200 is very rare items > 200 rare < 1000 semi-rare > 1000 normal > 4000 almost all times > 10000 always something like this, and the unique works fine for me, the player that get more points in the fight will get it.

joseluis2g:
@Schiffers do you know how do I send 2 unique items to 2 different players
I mean, ex. Gaz'haragoth loot 2 umbrals, give them to 2 players (most dmg, and the killer)

Schiffers:
@joseluis2g in this case I don't use the unique, the unique is used to the "unique" item, as the hat, this unique item always will go to the player that GET MORE POINTS, not necessary the mostdamagekiller
to the case of the gaz'haragoth I use unique only to the nightmare horn, and for umbral items I use lower chances ;)

kito2:
This system has the same bug as CipSoft.

Players can bug the points in order to achieve the best items, how? Easy... Just go to the boss, throw some field runes and the boss will start receveing dmg. When the entire team comes to dmg the boss (after hours, lol) the player who has bugged it will receive the unique items.

CipSoft has not resolved this yet and the solution is quite easy.

Go to boss.lua and onThink just add an event which reset all players score after 15~30 minutes (you choose).

ranisalt:
@kito2 if the reset happens just before the last hit only one player will get everything.

kito2:
The reset should be done after 15-30 minutes of receiving the first hit.

Kamenuvol:
I think a decremental counter would be better, the player lose points with time passing.

kito2:
What would u do against people bugging with field runes? That wouldnt help.

Kamenuvol:
Why do you think it is a bug? If you hit a monk with field to red hp and it heals you win most xp.
Boss Points should count with fields too.

kito2:
Have you ever played CipSoft? Do you know that people doesn't go to kill
bosses like Gaz'haragoth or Omrafir unless they haven't been unbugged?



 Sent with Mailtrack
<https://mailtrack.io/install?source=signature&lang=en&referral=fjoliva@miuandes.cl&idSignature=22>

Francisco Oliva
*Estudiante Ingeniería Civil Industrial | Universidad de los Andes*
(+56 9 4925 4225)

2016-10-13 23:40 GMT-03:00 Kamenuvol <notifications@github.com>:

Kamenuvol:
I think it is a weakness not a bug.  I see no problem using fields. The chance to win items is not best to the best. The player with more points has more chance, not 100%, to win the unique item/a better item. I've never killed a boss with this system in Cipsoft, but I think they not go kill because they don't want to, they would still have chance to receive good items.

kito2:
They dont. You cant compete against a player who put energy field and fire
field for around 2-4 hours. He would be on top of the score and he would
receive the best items always. Thats why people dont go on CipSoft, cause
of this weakness as u say.

Mojmistrz:
I've got one issue with charges attribute. Everytime boss drops an item which has charges, it's always 1 charge. How to fix that?

kito2:
Since the items are added to the container via lua you should check if it
is an item with charges and then give the charges that it is ment to be...
A solution was to dont set uniquedrop and just reduce the chance.

ranisalt:
@socket2810 can you update it following the feedback and rebase?

brunominervino:
@ranisalt, he is away from Tibia. Maybe he never back :s

ranisalt:
I have some contact with this guy, gonna check it out, but anyway keep the PR closed.

lamonato29:
Socket pelo amor de deus me ajude..
Mano, baixei um ot e arrumei todos os bugs, só faltou um que é o do reward system feito por vc..
Quando o monstro morre, recebo a msg : you are not the owner..
e quando vou no reward chest não tem nada..
Vc poderia ensinar como implementar o sistema desde o zero? pra eu poder conferir todas as etapas e tal.. 
Realmente não sei o que pode ta causando isso, mas ACHO que pode ser alguma coisa faltando nos arquivos source..

Dyablots:
For those having issue with lootFactor, simply replace in your rewardboss.lua:
configManager.getNumber("rateLoot")
with
getConfigInfo("rateLoot")

thexamx:
@dyablots
`configManager.getNumber(configKeys.RATE_LOOT`

Deux88:
any update on this system or similar for 1.3? Many things changed on source. Any way to point me into the right direction? Thanks.

EPuncker:
@ranisalt shouldn't we left it open just in case? 😛 I've stumbled on it by accident and found it quite interesting just like the store PR

sppeect:
should be reopened

EPuncker:
> should be reopened

can't because the repository that submitted this PR has been deleted

cezarguimaraes:
I have no idea what happened with this fork of mine. Well... it has been 9 years since the PR was opened.

Someone willing enough could recreate it (a second time) from the changes and fix the inevitable conflicts that would pop up.

On a last note, while checking my repositories to see if I had perchance privated it, I found a private fork of `tfs` which contains some other systems like `store`, `character_store` and `imbue`. I have no idea if any of it could be useful nowadays but in any case I made it [public](https://github.com/cezarguimaraes/tfs/branches/all)

Issue #605, opened 2014-03-05
Multiplier float support


Luckey1729:
config.lua and stages.xml should support float values.
fe. < stage minlevel="200" multiplier="0.5" />
##

Luckey1729:
#975 was closed, but #605 was not.
Yes, `Player:onGainExperience` can be used to emulate the desired effect; it seems like it's more of a "workaround" rather than a proper implementation. If it's intended to be this way, this issue should be closed as well.

andersonfaaria:
#975 should be merged not closed, this isn't about exp but rather using floats as possible options in config.lua. Also this issue should be closed as well...

EPuncker:
> #975 should be merged not closed, this isn't about exp but rather using floats as possible options in config.lua. Also this issue should be closed as well...

#2750 isn't this one the same?

ranisalt:
You can now use stages with decimal multipliers in config.lua :)

PR #1642, opened 2015-12-16
Move experience stages to Lua

ranisalt:
Removed experience stages logic from C++, too.

The logic is a bit simpler in Lua. The complexity is O(quantity of stages + range of the largest stage), both are runtime constants. Moving to Lua code enables the possibility of more complex multiplier functions, such as the low level bonus introduced on recent Tibia updates as I have done here: ranisalt/forgottenserver@8552826 

If the lower constant complexity is really wanted, I can refactor it.

Closes #3049 #605

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -286,6 +286,7 @@ bool ConfigManager::load()
 	} else {
 		std::cout << "[Warning - ConfigManager::load] XML stages are deprecated, consider moving to config.lua." << std::endl;
 	}
+	expStages.shrink_to_fit();
 
 	loaded = true;
 	lua_close(L);
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -133,20 +133,17 @@ ExperienceStages loadXMLStages()
 		} else {
 			uint32_t minLevel, maxLevel, multiplier;
 
-			pugi::xml_attribute minLevelAttribute = stageNode.attribute("minlevel");
-			if (minLevelAttribute) {
+			if (auto minLevelAttribute = stageNode.attribute("minlevel")) {
 				minLevel = pugi::cast<uint32_t>(minLevelAttribute.value());
 			} else {
 				minLevel = 1;
 			}
 
-			pugi::xml_attribute maxLevelAttribute = stageNode.attribute("maxlevel");
-			if (maxLevelAttribute) {
+			if (auto maxLevelAttribute = stageNode.attribute("maxlevel")) {
 				maxLevel = pugi::cast<uint32_t>(maxLevelAttribute.value());
 			}
 
-			pugi::xml_attribute multiplierAttribute = stageNode.attribute("multiplier");
-			if (multiplierAttribute) {
+			if (auto multiplierAttribute = stageNode.attribute("multiplier")) {
 				multiplier = pugi::cast<uint32_t>(multiplierAttribute.value());
 			} else {
 				multiplier = 1;
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -333,10 +333,6 @@ bool ConfigManager::getBoolean(boolean_config_t what) const
 
 float ConfigManager::getExperienceStage(uint32_t level) const
 {
-	if (expStages.empty()) {
-		return getNumber(ConfigManager::RATE_EXPERIENCE);
-	}
-
 	auto it = std::find_if(expStages.begin(), expStages.end(), [level](ExperienceStages::value_type stage) {
 		return level >= std::get<0>(stage) && level <= std::get<1>(stage);
 	});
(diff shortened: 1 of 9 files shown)


djarek:
Nice. As for the complexity considerations - you shouldn't really worry. It's _incredibly unlinkely_ that your code has any significant impact on the overall performance. Last time I checked, around 60% of the total run time of the server is spent executing code related to spectator checking logic, so linear complexity is fine.

mrhultin:
This is really neat. Should allow for both databasedriven stages aswell as 1.7x exp for example. Definatly going to use this.

ranisalt:
@mrhultin it's perfectly possible as `player.lua` is executed at startup, just load from the database using `db` and you should be fine.

mrhultin:
@ranisalt I have already done this, and it works like a charm I would say. I have not quite decided which way I like it more however, right now I use it compressed as Json in the row in the DB using server_config table as my table. This allows me to decompress the json quickly on the fly with PHP aswell.

ranisalt:
Care to share the code? :D

Raggaer:
How would external resources load the stages then? I think XML is better for this, how will AACs load your stages if they are on lua?

ranisalt:
> How would external resources load the stages then? I think XML is better for this, how will AACs load your stages if they are on lua?

Using the status protocol, just like otserv lists do. There's even [a PHP library](https://github.com/renatorib/otinfo) to abstract that.

ranisalt:
Is anyone interested? I'm not sure how people will prefer Lua over XML.

WibbenZ:
Seems like a great thing to change, good job!
But you didnt think about the retards that can't read a change log to update servers from 0.x

I mean we should use both ways so ppl aren't afraid of using 1.x ofc

ranisalt:
I'm not sure how can I maintain both behaviors. Maybe if the XML file does not exist, use Lua? But that wouldn't be really clean.

WibbenZ:
It was more of a joke that I tought you would understand xD
It's abt actually trying new things and making noobs spend more than 10min to upgrade a server.

Tvh imo you should not keep the xml crap, if you are chaning smth and its for the better, why would you keep old crappy stuff for no good reason?

ranisalt:
As a follower of http://semver.org/ that's actually something I'd hold for 2.0

WibbenZ:
You never know, my guess is that 2.0(if we still have devs by then) will probbly be the same as now except some minor changes.

Sad to say thay since we could update alot of things if we stop all the "oh no the noobs cant do that"
But thats just my opinion tho :p

gugahoa:
Use a flag on config.lua to choose which stages to use, Lua or XML, which defaults for XML
When we move to 2.0, remove the flag. (Add this to the roadmap so we don't forget, if merged?)

marksamman:
You can already override Game.getExperienceStage if you rather implement it in Lua, or make changes to the Player.onGainExperience Lua event. Backwards compatibility is important and XML isn't bad for this kind of thing. The only thing that could use an improvement in the XML file is support for decimal precision.

ranisalt:
Not only decimal precision, but you can really have a function instead of a table of values. I used the Lua system to have a very short and functional implementation of the low level bonus Tibia has: https://github.com/ranisalt/forgottenserver/commit/85528268409aedbd0cc7140996bf3f8773addeef

Xaekai:
I don't like this with the configuration being a local to the tucked away player.lua.
```diff
- Game.getExperienceStage <-- this is unaddressed lost functionality -
```
If you're gonna move it to Lua, the configuration block should be a global setup, available to query by any script that may require it. If you're gonna make it difficult to use and begging for the server owner to rip it out and replace it with something else, why have it at all? Just my 2 cents. 🪙🪙

ranisalt:
Now it has backwards compatibility with XML stages.

nekiro (dismissed):

DSpeichert (dismissed):

EPuncker (approved):

src/configmanager.cpp
```diff
@@ -248,6 +328,19 @@ bool ConfigManager::getBoolean(boolean_config_t what) const
 	return boolean[what];
 }
 
+
+float ConfigManager::getExperienceStage(uint32_t level) const
+{
+	auto it = std::find_if(expStages.begin(), expStages.end(), [level](ExperienceStages::value_type stage) {
+		return level >= std::get<0>(stage) && level <= std::get<1>(stage);
+	});
```
nekiro, line 336: actually, no newline after this? :(

nekiro (approved):

EPuncker (approved):

Underewarrr:
Thats what i'm talking about, simple code for stages.
Goood JOB!

DSpeichert (approved):

PR #1645, opened 2015-12-18
Added condition clearing functionality.

djarek:
This enables the creation of spells with dynamically changing combat conditions which differ between spell casts. The script needs to clear all current conditions and add them again.

### DIFF
--- a/src/combat.h
+++ b/src/combat.h
@@ -316,6 +316,11 @@ class Combat
 		void setCondition(const Condition* condition) {
 			params.conditionList.emplace_front(condition);
 		}
+
+		void clearConditions() {
+			params.conditionList.clear();
+		}
+
 		void setPlayerCombatValues(formulaType_t _type, double _mina, double _minb, double _maxa, double _maxb);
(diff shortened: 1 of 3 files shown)


ranisalt:
Great, we've been seeing a lot of spells needing this (or complex logic) :P

raymondtfr:
Great :+1: !!!
I'll be testing this in the spells pr you referenced.

PrinterLUA:
Great job, but while you are on it. Why not make also possible to clear all players conditions?

gugahoa:
Why wasn't this merged too? It would be cool and it would make it possibly to merge #1635

HeavenIsLost:
@marksamman why it was not been merge?

marksamman:
Because I'm not convinced that this is the right way to approach the problem. I think temporary Combat objects would be better. I'll merge this if @djarek rebases it though, since we don't have any better solution yet.

DSpeichert:
@djarek rebase, please?

PR #1648, opened 2015-12-19
Update Travis to new infrastructure

ranisalt:
This changes the Travis build to the modern infrastructure, which provides faster speed and less complex configuration. Mainly there's no need to manually configure PPAs, apt installs and has two dedicated cores and extra memory instead of the previous 1.5 cores. It usually takes 3 to 4 minutes to build compared to currently 7 minutes most of the builds.

What's the opinion on adding clang to the build matrix?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ addons:
       - boost-latest
       - ubuntu-toolchain-r-test
     packages:
-      - gcc-4.8
       - g++-4.8
       - libboost1.55-dev
       - libboost-system1.55-dev
@@ -15,6 +14,6 @@ addons:
       - liblua5.2-dev
       - libmysqlclient-dev
 install:
-  - if [ "$CXX" = "g++" ]; then export CXX="g++-4.8" CC="gcc-4.8"; fi
+  - if [ "$CXX" = "g++" ]; then export CXX="g++-4.8"; fi
 before_script: mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release ..
 script: make -j2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 sudo: required
 dist: trusty
 language: cpp
-compiler: gcc
+compiler:
+  - clang
+  - gcc
 cache: apt
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ addons:
 install:
   - if [ "$CXX" = "g++" ]; then export CXX="g++-4.8" CC="gcc-4.8"; fi
 before_script: mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release ..
-script: make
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ else()
     find_package(Lua)
 endif()
 
-find_package(Boost 1.53.0 COMPONENTS system REQUIRED)
+find_package(Boost 1.55.0 COMPONENTS system REQUIRED)
```
marksamman: Why?

.travis.yml
```diff
@@ -1,12 +1,20 @@
...
-  - sudo add-apt-repository ppa:apokluda/boost1.53 -y
-  - sudo apt-get update -qq
-  - if [ "$CXX" = "g++" ]; then sudo apt-get install -qq g++-4.8; fi
+addons:
+  apt:
+    sources:
+      - boost-latest
+      - ubuntu-toolchain-r-test
+    packages:
+      - gcc-4.8
```
marksamman: Isn't g++-4.8 enough, if gcc is also needed g++ probably has it as an install dependency?

marksamman:
If it builds successfully with Clang, you should add it.

CMakeLists.txt
ranisalt: Travis only whitelists 1.55 and the latest from Ubuntu Precise, which is too old: https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise#L3854-L3934

I forgot to link this on the PR message :P

marksamman: I'm not questioning the use of 1.55 in Travis, but this is the version requested by CMake. Using 1.55 should still work even if CMake asks for 1.53.

.travis.yml
ranisalt: Yes, you are right. I removed the gcc entry but it is a dependency for g++. There's also no need to set the `CC` variable as it won't be used.

ranisalt:
Unfortunately it's not possible, since it would be needed to rebuild boost with Clang, else it fails to link because of namespace errors between stdlib and libcxx.

marksamman:
Have you tried using the Trusty beta? We shouldn't have to use any PPA's and manually set CXX to g++ 4.8 with it, and it's also possible that Clang would build without errors in Trusty. It requires enabling sudo again, but if it means faster builds or that Clang works, then it's worth it.

See https://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta

ranisalt:
I'm going to take a loot at that when possible. It is great news from Travis :smile:

ranisalt:
Done, building under both GCC 4.8 and Clang 3.5 on Travis Trusty. Mind that Trusty is horribly slow to boot, but is faster to build.

I'd like to add compilation with LuaJIT to the matrix too, if wanted.

marksamman:
Sure, you can do so in a separate PR. Thanks!

PR #1649, opened 2015-12-19
Fix #1631

### LINKED ISSUE #1631: Splash/Blood on Hole
If any type of splash/blood is on hole you cant open it.
On picture we clearly see that blood is above hole.

ninjalulz:
Players should now be able open a hole even though there are items/splash on top of it.

### DIFF
--- a/data/actions/scripts/tools/shovel.lua
+++ b/data/actions/scripts/tools/shovel.lua
@@ -1,9 +1,27 @@
 local holes = {468, 481, 483}
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
-	if isInArray(holes, target.itemid) then
-		target:transform(target.itemid + 1)
-		target:decay()
-	elseif target.itemid == 231 then
+	if toPosition.x == CONTAINER_POSITION then
+		return false
+	end
+
+	local tile = Tile(toPosition)
+	if not tile then
+		return false
+	end
+
+	local ground = tile:getGround()
+	if not ground then
+		return false
+	end
+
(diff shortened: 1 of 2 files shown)


data/actions/scripts/tools/shovel.lua
```diff
@@ -1,9 +1,27 @@
...
+	end
+
+	local groundId = ground:getId()
+	if isInArray(holes, groundId) then
+		ground:transform(groundId + 1)
+		ground:decay()
+
+		toPosition.z = toPosition.z + 1
+		tile:relocateTo(toPosition)
+	elseif groundId == 231 then
```
ranisalt: What is 231?

ranisalt:
> and for who use the tools actions at the lib ?

IIRC ninja is the maintainer of ORTS, he probably will update that too as soon as this PR is merged

marksamman:
Are all hole items ground tiles? If not, then this breaks them, but if they are, then this looks good to me.

data/actions/scripts/tools/shovel.lua
ninjalulz: It's sand, probably an old id.

ninjalulz:
@marksamman The item ids whom are stored in the holes array are, but there is an id that's not being used that isn't a ground tile.

I could however, update the pull request to use this: https://gist.github.com/ninjalulz/1cc7a6f3106b7ef683bb

marksamman:
That's good enough for now then, getItemById should be avoided if possible since it iterates over all items on the tile.

PR #1654, opened 2015-12-22
Potions like RL tibia.

l04d:
instead get target level/voc, get player level/voc, also flasks are added from potions source, e.g. if i use a potion from the ground, empty flask goes to ground instead BP like actual.

### DIFF
--- a/data/actions/scripts/other/potions.lua
+++ b/data/actions/scripts/other/potions.lua
@@ -34,6 +34,8 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 	end
 
 	local itemId = item:getId()
+	local itemPos = item:getPosition()
+	
 	if itemId == antidotePot then
 		if not antidote:execute(target, numberToVariant(target:getId())) then
 			return false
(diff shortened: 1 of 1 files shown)


data/actions/scripts/other/potions.lua
```diff
@@ -42,7 +44,11 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 		player:addCondition(exhaust)
 		target:say("Aaaah...", TALKTYPE_MONSTER_SAY)
 		item:remove(1)
-		player:addItem(emptyPot, 1)
+			if (fromPosition.x == 65535) then
```
WibbenZ: All of these needs the indentation fixed.

data/actions/scripts/other/potions.lua
```diff
@@ -34,6 +34,8 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 	end
 
 	local itemId = item:getId()
+	local itemPos = item:getPosition()
+
```
WibbenZ: Trailing tab

PR #1655, opened 2015-12-26
Use correct block type effect

raymondtfr:
The correct (resembling vanilla) block type effect when (damage <= 0) due to element resistances is ``BLOCK_ARMOR``.


marksamman:
Can you update your commit message to better describe what you have changed?

raymondtfr:
Yes, I can.

marksamman:
Thanks!

PR #1659, opened 2015-12-31
Update Spells

raymondtfr:
**Some changes I made in #1635 doesn't need to make the use of #1645 nor does it requires the enhancement suggested in #1625 , that's why I'm making this new PR.**

Updated rounds, time, and damage (based on vanilla) for the spell: "Holy Flash".

Added ``casterTargetOrDirection="1"`` to some spells that can both be cast
with or without target. These spells are listed below.
* Curse
* Envenom
* Electrify
* Holy Flash
* Ignite

Removed the following incorrect (hence unused) parameters from the spell
"Inflict Wound".:
* ``combat:setParameter(COMBAT_PARAM_TARGETCASTERORTOPMOST, 1)``
* ``combat:setParameter(COMBATPARAM_USECHARGES, 1)``

Updated some spell effects resembling vanilla -- this change was important for the spells "Ignite", "Holy Flash", since by adding ``casterTargetOrDirection="1"`` it would need a "default effect" besides the "distance effect".

### DIFF
--- a/data/spells/scripts/attack/curse.lua
+++ b/data/spells/scripts/attack/curse.lua
@@ -1,6 +1,6 @@
 local combat = Combat()
 combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_DEATHDAMAGE)
-combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_SMALLCLOUDS)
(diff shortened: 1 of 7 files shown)


marksamman:
Thanks!

PR #1661, opened 2016-01-03
Update spells 2

raymondtfr:
Renamed **var** to **variant** in ``onCastSpell()`` to avoid code ambiguities..
Renamed **1/0** to **true/false** in some combat/condition parameters.
Renamed some conditions starting with the word **exhaust** to **cooldown** -- cause
it isn't related to the exhaustion, but to the cooldown.
Added missing **isHotkey** parameter in ``onCastSpell()`` to Paralyze Rune.
Enhanced the spell **mass healing** (added a *bool* ``healMonsters = false``to choose whether or not monsters should be healed.)

Updated spell formulas of the following spells:
* Ethereal Spear
* Strong Ethereal Spear

I've removed the arithmetic around **attack**, since it only considers the player's **distance skill** -- also simplified their formulas.

Rearranged some **boolean** parameters:
*	``COMBAT_PARAM_TARGETCASTERORTOPMOST``
*	``COMBAT_PARAM_AGGRESSIVE``
*	``CONDITION_PARAM_BUFF_SPELL``

in some healing/support spells to the last line of parameters.

Updated knight attack spells -- defines local min, max -- uses two less
variables.

Updated spell **heal friend** (based on vanilla) to send **green shimmer** effect on target and
**blue shimmer** effect on the caster simultaneously (or a blue shimmer
on the caster when healing himself) -- *making it easy to identify the healer*.

Updated some support **haste based** spell formulas.
Since they launched the flash client, it is now possible to effectively check your character's speed by looking at your character profile in-game. Now the formulas for the following spells: Haste, Charge, Swift Foot -- now gives **exactly** the same speed changes as vanilla. I'll update the remaining ones in the future.

Updated the spell **food**. It is now very similar to vanilla, even though
it might look redundant to use ``addItem()`` twice.

Removed some unnecessary checks, most of them were there because of *monster spells*, but they aren't used in monsters -- only *soul fire* does (which can later be changed for another soul fire spell within scripts/monster).

Fixed the spells **Blood Rage** and **Sharpshooter**, cause they didn't have the proper value "-100" to **null** the *Shielding Skill*.

Removed unused condition from the spell **Protector**:
*	``Condition(CONDITION_EXHAUST_COMBAT)``

it should be:
*	``Condition(CONDITION_PACIFIED)``

Also removed incorrect condition:

*	``Condition(CONDITION_EXHAUST_HEAL)``

The correct conditions should only prevent the player from *attacking* (pacified) or casting *attack spells* (cooldownAttackGroup) at the target.

Small code cleanups/re-arrangements.


ranisalt:
> Renamed true to 1 in some combat parameters.

I'm against that, since true means... true while 1 does not have the same semantic, even though it works. I'm also for changing all zeroes to false too.

WibbenZ:
And if you ask me we should use variable, minimum and maximum insted of var, min and max.

raymondtfr:
@ranisalt I thought using false/true instead of 0/1 wasn't supported in tfs?
@WibbenZ I did not understand the sarcasm :question:

WibbenZ:
Sarcasm no.
Im saying we should use correct names, position insted of pos etc

The src supports both integers and booleans, https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L11226-L11230

But id go with a boolean value, since once again use the correct names.
1 could be seen as a boolean value but mainly as an integer.

marksamman:
I agree with using boolean instead of 0 and 1 where the parameter is used as a boolean, as it makes it clearer a boolean is expected for the specific parameter, and not any numerical value. However, that is currently not supported for Combat.setParameter. If you would like to add support for it in this pull request, I'm for it.

I also agree with using the name **variant** instead of var, as var is quite ambiguous (a good example here is that WibbenZ thought it meant variable), however I disagree about min/max because they're obvious.

raymondtfr:
> The src supports both integers and booleans, https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L11226-L11230

Oh, I thought it wasn't supported for both combat and condition.
I also thought about renaming var to variant, cause as you said @marksamman it is quite ambiguous, though it was really around in most of the files. so I chose to keep it. I'll be renaming *var* to **variant**.
Ok, I'll look through the src and I'll try to add support for it.

I have also noticed something in the source, regarding the current speed formula. It seems that it has changed (I have confirmed), instead of ``vocation->getBaseSpeed() + (2 * (level - 1))`` it is now ``vocation->getBaseSpeed() + (level - 1)``. It is said in [here] (http://tibia.wikia.com/wiki/Talk:Speed) they changed the formula since update 9.8.

WibbenZ:
@marksamman Just now noticed that I wrote variable insted of variant haha :facepunch: 

It's pretty simple actually, just add the same if statment to this function, https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L10907-L10919

https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L11225-L11231

raymondtfr:
@WibbenZ Thanks for pointing that out.

marksamman:
Thanks!

PR #1665, opened 2016-01-05
Fix #1634: show vocations on runes look message

### LINKED ISSUE #1634: mission vocation info on runes
global:
You see a paralyse rune.
It can only be used by druids with level 54 and magic level 18 or higher.
It weighs 2.10 oz.

tfs:
You see a paralyze rune ("adana ani"). It can only be used with level 54 and magic level 18 or higher.
It weighs 2.10 oz.

ranisalt:
As @Zbizu reported, vocation names were missing from runes look message.

Things changed a bit: vocations are now required to be listed on `spells.xml` even if the unpromoted version already is. Also, it is assumed that if you specify no vocation, it is available to all of them (except, of course, `none`) and it will show `by players` instead of the vocation list.

The vocation list is, when used, shows as `(...) used by as, bs, cs... and zs with level (...)`. Examples:
```
It can only be used by players with level 54 and magic level 18 or higher.
It can only be used by druids with level 54 and magic level 18 or higher.
It can only be used by druids and sorcerers with level 54 and magic level 18 or higher.
It can only be used by druids, paladins and sorcerers with level 54 and magic level 18 or higher.
```

Runes do not show magic words anymore, but I feel like it could be toggled by a configuration flag (not the matter of this PR).

Please report any errors. Thanks to @slavidodo who was also developing a solution.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -31,8 +31,12 @@
 
 #include "actions.h"
 #include "combat.h"
+#include "spells.h"
+#include "vocation.h"
 
 extern Game g_game;
+extern Spells* g_spells;
+extern Vocations g_vocations;
 
 Items Item::items;
 
@@ -814,10 +818,6 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
 	}
 
 	if (it.isRune()) {
-		if (!it.runeSpellName.empty()) {
-			s << " (\"" << it.runeSpellName << "\")";
-		}
-
 		if (it.runeLevel > 0 || it.runeMagLevel > 0) {
 			int32_t tmpSubType = subType;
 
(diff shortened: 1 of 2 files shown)


src/item.cpp
```diff
@@ -825,7 +825,41 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
 				tmpSubType = item->getSubType();
 			}
 
-			s << ". " << (it.stackable && tmpSubType > 1 ? "They" : "It") << " can only be used with";
+			s << ". " << (it.stackable && tmpSubType > 1 ? "They" : "It") << " can only be used by ";
+
+			VocSpellMap vocMap = g_spells->getRuneSpell(it.id)->getVocMap();
```
marksamman: 1) getRuneSpell() can return null
2) is there any reason you're copying the map instead of holding it by const reference?

src/item.cpp
```diff
@@ -825,7 +825,41 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
...
+			if (vocMap.empty()) {
+				s << "players";
+			} else {
+				std::vector<Vocation*> showVocMap;
+
+				// vocations listed are mostly unpromoted ones and the promoted version, with the latter hidden from
+				// description, so probably always `total / 2` is the amount of vocations to be shown.
+				showVocMap.reserve(vocMap.size() / 2);
+
+				for (const auto& voc: vocMap) {
```
marksamman: Add a space before colon.

src/item.cpp
```diff
@@ -825,7 +825,41 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
...
+					if (voc.second) {
+						showVocMap.push_back(g_vocations.getVocation(voc.first));
+					}
+				}
+
+				auto vocIt = showVocMap.begin(), vocLast = (showVocMap.end() - 1);
+
```
marksamman: Remove this empty line.

src/item.cpp
```diff
@@ -825,7 +825,41 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
...
+				while (vocIt != vocLast) {
+					auto vocName = asLowerCaseString((*vocIt)->getVocName());
```
marksamman: Append the result from asLowerCaseString directly into the stringstream instead of holding it in a variable first, as it may result in copying it more than we need to (make sure the iterator is incremented after appending it and not before).

src/spells.cpp
```diff
@@ -572,11 +572,8 @@ bool Spell::configureSpell(const pugi::xml_node& node)
 
 		int32_t vocationId = g_vocations.getVocationId(attr.as_string());
 		if (vocationId != -1) {
-			vocSpellMap[vocationId] = true;
-			int32_t promotedVocation = g_vocations.getPromotedVocation(vocationId);
-			if (promotedVocation != VOCATION_NONE) {
-				vocSpellMap[promotedVocation] = true;
-			}
+			attr = vocationNode.attribute("showInDescription");
+			vocSpellMap[vocationId] = not attr or attr.as_bool();
```
marksamman: Use !, || instead of not, or.

slavidodo: I notified him with that already!

marksamman: Yes, but he didn't change it. It's not about whether or not they're defined, but about being consistent. We're using ! and || everywhere else, it's just confusing to suddenly have a line where we use different syntax.

src/item.cpp
ranisalt: 1) Doesn't that means someone has not configured a rune properly? By setting it as a rune but not assigning any callback to it. Maybe it can lead to other problems too.
2) No, my bad, will fix.

ranisalt:
Styling fixes done and also protected against null pointers and unneeded copies.

src/item.cpp
marksamman: Removing a rune from spells.xml, reloading and looking at the rune would have been enough to trigger a null pointer dereference unless I'm missing something.

marksamman:
Thanks!

PR #1668, opened 2016-01-06
Adding ability to equip hotkeys.

slavidodo:
Thanks to @ninjalulz .

### DIFF
--- a/src/creature.h
+++ b/src/creature.h
@@ -48,7 +48,7 @@ enum slots_t : uint8_t {
 	CONST_SLOT_LAST = CONST_SLOT_AMMO,
 };
 
-+enum clientSlots_t : uint16_t {
+enum clientSlots_t : uint16_t {
 	CLIENT_SLOT_HAND = 48,
 	CLIENT_SLOT_HEAD = 49,
 	CLIENT_SLOT_NECKLACE = 50,
--- a/src/creature.h
+++ b/src/creature.h
@@ -48,6 +48,16 @@ enum slots_t : uint8_t {
 	CONST_SLOT_LAST = CONST_SLOT_AMMO,
 };
 
++enum clientSlots_t : uint16_t {
+	CLIENT_SLOT_HAND = 48,
+	CLIENT_SLOT_HEAD = 49,
+	CLIENT_SLOT_NECKLACE = 50,
+	CLIENT_SLOT_ARMOR = 56,
+	CLIENT_SLOT_LEGS = 112,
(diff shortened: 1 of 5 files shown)


src/creature.h
```diff
@@ -48,6 +48,16 @@ enum slots_t : uint8_t {
 	CONST_SLOT_LAST = CONST_SLOT_AMMO,
 };
 
++enum clientSlots_t : uint16_t {
```
esfomeado: Extra '+' here

slavidodo: Edited, Thanks.

src/protocolgame.cpp
```diff
@@ -818,6 +819,34 @@ void ProtocolGame::parseUpdateContainer(NetworkMessage& msg)
 	addGameTask(&Game::playerUpdateContainer, player->getID(), cid);
 }
 
+void ProtocolGame::parseEquipObject(NetworkMessage& msg) {
```
esfomeado: Should be parseEquipHotkey

src/protocolgame.cpp
```diff
@@ -432,6 +432,7 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
 		case 0x70: addGameTaskTimed(DISPATCHER_TASK_EXPIRATION, &Game::playerTurn, player->getID(), DIRECTION_EAST); break;
 		case 0x71: addGameTaskTimed(DISPATCHER_TASK_EXPIRATION, &Game::playerTurn, player->getID(), DIRECTION_SOUTH); break;
 		case 0x72: addGameTaskTimed(DISPATCHER_TASK_EXPIRATION, &Game::playerTurn, player->getID(), DIRECTION_WEST); break;
+		case 0x77: parseEquipObject(msg); break;
```
esfomeado: Should be parseEquipHotkey

marksamman:
There are a couple of problems with this:
1) you can't just copy someone elses code and submit it as a pull request, you need their permission as the code becomes licensed under GPL when we merge it
2) your commit messages do not describe what you have changed
3) this only works for items you had equipped when you logged in, and sending every equipped item every time something in the inventory changes is too costly

slavidodo:
Understood, Close it please.

ranisalt:
Whose code is this? It's actually pretty neat, but the interval between sending inventory changes could be a debounced interval of, say, 5 seconds, meaning 5 seconds after the last change, if no changes occur, we send an update.

PR #1669, opened 2016-01-06
missing books from spellbook.lua

l04d:


data/actions/actions.xml
```diff
@@ -132,6 +132,11 @@
 	<action itemid="6120" script="other/spellbook.lua" />
 	<action fromid="8900" toid="8904" script="other/spellbook.lua" />
 	<action itemid="8918" script="other/spellbook.lua" />
+	<action fromid="22422" toid="22424" script="other/spellbook.lua" />
+	<action itemid="18401" script="other/spellbook.lua" />
+	<action itemid="16112" script="other/spellbook.lua" />
+	<action itemid="15411" script="other/spellbook.lua" />
```
marksamman: Depth scutum should not work as a spellbook. Could you also re-order them in ascending order of the item IDs? Please squash your commits when you're done.

l04d:
@marksamman Oh sorry for the mess, about the depth scutum, I'd never notice if you don't tell me, it's funny because I used to use it so much and never noticed =O

marksamman:
It's OK, just squash your commits and write a commit message that better describes what you have changed. "Update filename" is vague.

data/actions/actions.xml
```diff
@@ -132,7 +132,11 @@
 	<action itemid="6120" script="other/spellbook.lua" />
 	<action fromid="8900" toid="8904" script="other/spellbook.lua" />
 	<action itemid="8918" script="other/spellbook.lua" />
-
+	<action itemid="12647" script="other/spellbook.lua" />
+	<action itemid="16112" script="other/spellbook.lua" />
+	<action itemid="18401" script="other/spellbook.lua" />
+	<action fromid="22422" toid="22424" script="other/spellbook.lua" />
+
```
ranisalt: Trailing whitespace

ranisalt:
Do you have a git client installed locally? I don't think it's possible to squash commits via GitHub tools but you seem to be editing the files directly in GitHub editor.

l04d:
@ranisalt actually I'm waiting my interval to reboot and squash via terminal, on windows I only can via github tool, and actually I was trying and i closed to do it after as i noticed i just can't do it through tools.

PR #1670, opened 2016-01-06
Optional LuaJIT + Travis matrix

ranisalt:
Currently LuaJIT is not optional, it is used whenever available and not used otherwise.

I changed it to be more smart: the user can specify if he wants LuaJIT or not but, by default, it uses LuaJIT when available. This way I could set environment variables on Travis to toggle LuaJIT (it would be a pain to conditionally install) and  I found a nice way to compile under both conditions without having to uninstall LuaJIT or mess with CMakeLists.

LuaJIT is togglable with the `USE_LUAJIT` flag, meaning you can force LuaJIT with `cmake -DUSE_LUAJIT=ON` or `cmake -DUSE_LUAJIT=OFF`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ language: cpp
 compiler:
   - clang
   - gcc
+env:
+  - LUAJIT=OFF
+  - LUAJIT=ON
 cache: apt
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


marksamman:
Thanks!

PR #1675, opened 2016-01-07
New spells -- weapons enhancement

raymondtfr:
Added new spells from dawnport island -- 10.55 update.
Added a few items required for some spells.
Enhanced weapon files (effects, block type, formula, etc).


ranisalt:
For curiosity: why all your PRs have a non-FF merge as first commit? You could rebase instead of pulling/merging to prevent useless, non-wanted commits.

raymondtfr:
@ranisalt How do I do that?:D
Every time I update my fork with new changes from tfs I get this merge message, and I really don't know how to take it off.

ranisalt:
I expect you have two remotes configured locally, `origin` which is your fork and, say, `upstream`, which is the original repo. Whenever you want to resync, checkout to your branch and `git rebase upstream/master`, and git will do the dirty job of rewriting history.

Since this is history rewriting, you will need to force push with `git push -f`, but this is OK. Anyway you don't need to always keep in sync, as long as your edits do not conflict with the upstream they can be merged seamlessly.

raymondtfr:
Thanks for the tip. When I use ``git rebase -i HEAD~n`` I get the latest commits from that merge message, so I removed all of them for the first merge commit message, then I did the same for the second merge commit message, then i ```git push -f``.

ranisalt:
Precisely :wink:

data/spells/scripts/attack/buzz.lua
```diff
@@ -0,0 +1,23 @@
+local combat = Combat()
+combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_ENERGYDAMAGE)
+combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_ENERGYAREA)
+combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ANI_ENERGY)
+
+function onGetFormulaValues(player, level, maglevel)
+	if level > 20 then
```
marksamman: Use math.min instead of all these if-statements, e.g.
```lua
level = math.min(level, 20)
```

raymondtfr: Omg, I entirely forgot the magic level part xD.

raymondtfr:
Ok, added math.min(x, 20) for both level and magic level.

marksamman:
Thanks!

PR #1677, opened 2016-01-08
Move getMaster down 1 line - mass healing.lua

WibbenZ:
There is no reason to get the master, since it might not be needed.

### DIFF
--- a/data/spells/scripts/healing/mass healing.lua
+++ b/data/spells/scripts/healing/mass healing.lua
@@ -8,11 +8,11 @@ local healMonsters = false
 
 function onTargetCreature(creature, target)
 	local player = creature:getPlayer()
-	local min = ((player:getLevel() / 5) + (player:getMagicLevel() * 4.6) + 100)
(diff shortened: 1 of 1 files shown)


raymondtfr:
While your at it, you could also remove that pair of redundant parenthesis in min/max -- so that ``((a / 5) + (b * n) + c)`` becomes ```(a / 5) + (b * n) + c``.

WibbenZ:
But c is still used in the formula xD
The point of using parentheses is so you can click on one to see where that part begins and ends.
But if Mark prefers it ill change it.

raymondtfr:
Well, **``c``** will still be part of the arithmetic. By the way, all formulas of spell files now uses this style -- it was great a suggestion of @ranisalt . This style makes formulas look even more clearer when it is a attack spell -- because we can use ``-min`` instead of enclosing the whole formula with **``()``** and appending a **``-``** sign.

WibbenZ:
Yes both ways works, but I updated it.

marksamman:
Thanks!

PR #1683, opened 2016-01-12
Spells minor update

raymondtfr:
Added *missing* **secondary group** ``special`` (and their respective cooldowns) for the spells:
 * **Strong Ice Strike**
 * **Strong Terra Strike**
 * **Strong Energy Strike**
 * **Strong Flame Strike**
 * **Lightning**

**Fixed** the ``range`` for the spell **Lightning** - the correct range is ``5``.
**Modified** ``Condition(CONDITION_HASTE)`` to ``Condition(CONDITION_PARALYZE)`` in spell **Sharpshooter** - so the *"paralyze icon"* shows up and the player *can easily acknowledge he's been paralyzed*.
**Renamed** ``AREA_SHORTWAVE3`` to ``AREA_WAVE3`` in spell **Strong Ice Wave**.
**Renamed** ``exhaust`` to ``cooldown`` [compatibility is already there] (https://github.com/otland/forgottenserver/blob/master/src/spells.cpp#L513) for clarity and because the value in exhaust stands for the *spell/rune* cooldown.

### DIFF
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -1,7 +1,7 @@
 --Pre-made areas
 
 --Waves
-AREA_SHORTWAVE3 = {
+AREA_WAVE3 = {
 {1, 1, 1},
 {1, 1, 1},
 {0, 3, 0}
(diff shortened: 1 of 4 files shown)


ranisalt:
Wow that's a lot of commits

raymondtfr:
>Wow that's a lot of commits

But in the end I always manage to get rid of them! :sleeping:

marksamman:
Thanks!

PR #1686, opened 2016-01-16
Added toContainer on event "onMoveItem". Fixes #1446

### LINKED ISSUE #1446: How to get container location or owner on Player:onMoveItem() ?
I wanted to change the way items work with containers in LUA, but I'm missing 1 important parameter and I can't get that. (toPosition owner or uid if its container)

When I move item to container the toPosition.x is always 65535
But all the other parameters are VERY USELESS. (cyclinder and bag window id.)

There seems no way to found out what is the real position of the container or does it have owner.

I can get the owner when the fromPosition is bag, but not with the toPosition.
Also I cant create temporary item with these Position values, to check if the item reaches to owner.

And I don't think its good idea to add an addEvent() before true, to check afterwards where the item ended up. 

Is there some kind of workaround I'm missing or where in sourcecode I have push more parameters to get my desired values?

That would be real epic if i could get the container unique Id or the container userdata if its not a tile where item is moved.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24144287-how-to-get-container-location-or-owner-on-player-onmoveitem?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ricardosohn:
This is my first little contribution here. Hope you like it.
If there is something wrong with it, tell me.

### DIFF
--- a/data/events/scripts/player.lua
+++ b/data/events/scripts/player.lua
@@ -85,7 +85,7 @@ function Player:onLookInShop(itemType, count)
 	return true
 end
 
-function Player:onMoveItem(item, count, fromPosition, toPosition, toContainer)
+function Player:onMoveItem(item, count, fromPosition, toPosition, toCylinder)
 	return true
 end
 
--- a/data/events/scripts/player.lua
+++ b/data/events/scripts/player.lua
@@ -85,7 +85,7 @@ function Player:onLookInShop(itemType, count)
 	return true
 end
 
-function Player:onMoveItem(item, count, fromPosition, toPosition)
+function Player:onMoveItem(item, count, fromPosition, toPosition, toContainer)
 	return true
 end
 
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -519,16 +519,18 @@ bool Events::eventPlayerOnMoveItem(Player* player, Item* item, uint16_t count, c
 	LuaScriptInterface::pushPosition(L, fromPosition);
 	LuaScriptInterface::pushPosition(L, toPosition);
 
-	if (toCylinder->getCreature()) {
-		LuaScriptInterface::pushUserdata<Creature>(L, toCylinder->getCreature);
-		LuaScriptInterface::setMetatable(L, -1, "Creature");
-	} else if (toCylinder->getItem()){
-		LuaScriptInterface::pushUserdata<Item>(L, toCylinder->getItem());
-		LuaScriptInterface::setItemMetatable(L, -1, toCylinder->getItem());
+	if (toCylinder->getCreature()->getPlayer()) {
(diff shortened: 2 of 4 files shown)


ranisalt:
You should push fromCylinder too

ricardosohn:
@ranisalt
At the time I did that, I didn't knew for sure if I could push nil to that.
And why do we need fromCylinder?

ranisalt:
We might not need now, but it could help e.g. to control a reward system or whatever.

I favor implementing it since if we might want in the future, 1) we break backwards compatibility since we would push `(...), fromCylinder, toCylinder` and servers using `toCylinder` would start to get `fromCylinder` instead or 2) we would make the signature be `fromPos, toPos, toCylinder, fromCylinder` which would be dumb.

So either implement it right now both of it may become a pain.

ricardosohn:
But you don't actually need fromCylinder, because you can just use the getParent() function on Item, which would result on the same

ranisalt:
Fair enough, didn't remember it.

socket2810:
This is an useful feature and I'd like to see it merged. Anyway, this is how you can get the destiny container using toPosition:
```lua
local function getDestContainer(player, position)
	if position.x ~= CONTAINER_POSITION then
		return nil
	end 

	if bit.band(position.y, 0x40) == 0 then
		local slotItem = player:getSlotItem(position.y)
		if slotItem and slotItem:isContainer() then
			return slotItem
		end
	else
		local cid = bit.band(position.y, bit.bnot(0x40))
		local container = player:getContainerById(cid)
		if container then
			local target = container:getItem(position.z)
			if target and target:isContainer() then
				return target
			end

			return container
		end
	end
end

function Player:onMoveItem(item, count, fromPosition, toPosition)
	local toContainer = getDestContainer(self, toPosition)
	if toContainer then
		toContainer:addItem(2160, 5)
	end
	return true
end
```

marksamman:
It should be called toCylinder in Lua too, and should push the metatable of any supported Lua cylinders (e.g. Player and Tile), not just containers.

ricardosohn:
Hmm, so I should make it return one of these cylinders: creature, item, container(?), right ?

marksamman:
No, Item and Creature aren't cylinders. Cylinders are objects that you can add items to.

ricardosohn:
I need to check every possibility then ? Container, Player, Tile and what else?

ranisalt:
A tile is a cylinder. You can add items to a tile, meaning a tile has an internal list of items.

marksamman:
I've added a pushCylinder function in my last commit, please use it as it should help you get this right.

PR #1688, opened 2016-01-17
Check two-handed weapons on equip

socket2810:
As the title says, shields are moved from their slot whenever a two-handed weapon is equipped, and vice versa.


nekiro:
IMO this should be done in sources.

Kamenuvol:
Events were created to move code from the sources to Lua. So, it is right.

vankk:
I think that the toPosition condition should be change to `==` and not `~=` to return true - and use the toPosition condition scope to the rest of the script. It will avoid possibles people mistakes that add a code above the last return true when they need the toPosition be equal CONTAINER_POSITION * [Like this](https://pastebin.com/Lig4LDXZ).

PR #1693, opened 2016-01-20
Weapons damage code enhancement

raymondtfr:
Improves weapons code by creating a new function ``Weapons::getMinWeaponDamage()`` as a complement of ``Weapons::getMaxWeaponDamage()`` -- instead of using ``if``s to accomplish that task.

### DIFF
--- a/src/weapons.cpp
+++ b/src/weapons.cpp
@@ -140,6 +140,12 @@ int32_t Weapons::getMaxMeleeDamage(int32_t attackSkill, int32_t attackValue)
 }
 
 //players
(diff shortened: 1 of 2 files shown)


raymondtfr:
Oh, I misunderstood it. I will close this and may be re-open it when I take more time to review the code.

Well, melee weapons indeed don't have minimum damage itself, but the player's damage grows with his level (each 5 levels). So the minimum damage dealt by a player -- either by using a melee or distance weapon -- is (level / 5). *source: t-s*

PR #1694, opened 2016-01-23
Added some missing functionality

BenDol:
- Added some missing lua methods.
- Added creature move locking (setMoveLocked/isMoveLocked).
- Added less loss attribute to vocations to customize death loss per vocation.
- Added definitions.h to cmake source definitions.
- Always reload global.lua libraries.

Will squish the commits when its ready.

### DIFF
--- a/src/creature.h
+++ b/src/creature.h
@@ -182,6 +182,13 @@ class Creature : virtual public Thing
 			hiddenHealth = b;
 		}
 
+		bool isMoveLocked() const {
+			return moveLocked;
+		}
+		void setMoveLocked(bool ml) {
+			moveLocked = ml;
+		}
+
 		int32_t getThrowRange() const final {
 			return 1;
 		}
@@ -534,6 +541,7 @@ class Creature : virtual public Thing
 		bool hasFollowPath;
 		bool forceUpdateFollowPath;
 		bool hiddenHealth;
+		bool moveLocked;
 
 		//creature script events
 		bool hasEventRegistered(CreatureEventType_t event) const {
--- a/src/creature.h
+++ b/src/creature.h
@@ -185,8 +185,8 @@ class Creature : virtual public Thing
 		bool isMoveLocked() const {
 			return moveLocked;
 		}
-		void setMoveLocked(bool ml) {
-			moveLocked = ml;
+		void setMoveLocked(bool locked) {
+			moveLocked = locked;
 		}
 
 		int32_t getThrowRange() const final {
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3115,6 +3115,16 @@ void Game::playerTurn(uint32_t playerId, Direction dir)
 
 	player->resetIdleTime();
 	internalCreatureTurn(player, dir);
+
+	if(player->getDirection() != dir || (player->getAccountType() < ACCOUNT_TYPE_GAMEMASTER))
+		return;
+
+	Position pos = getNextPosition(dir, player->getPosition());
+
+	Tile* tile = g_game.map.getTile(pos);
+	if(tile && tile->getGround()) {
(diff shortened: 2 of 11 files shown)


src/creature.h
```diff
@@ -182,6 +182,13 @@ class Creature : virtual public Thing
 			hiddenHealth = b;
 		}
 
+		bool isMoveLocked() const {
+			return moveLocked;
+		}
+		void setMoveLocked(bool ml) {
```
WibbenZ: ml should be named something else

src/game.cpp
```diff
@@ -3115,6 +3115,16 @@ void Game::playerTurn(uint32_t playerId, Direction dir)
 
 	player->resetIdleTime();
 	internalCreatureTurn(player, dir);
+
+	if(player->getDirection() != dir || (player->getAccountType() < ACCOUNT_TYPE_GAMEMASTER))
```
WibbenZ: Curly brackets, "if (" not "if(" note the space.

src/vocation.cpp
```diff
@@ -109,6 +109,10 @@ bool Vocations::loadFromXml()
 			voc.fromVocation = pugi::cast<uint32_t>(attr.value());
 		}
 
+		if ((attr = vocationNode.attribute("lessloss"))) {
```
WibbenZ: Well you added the xml loading part, but nothing happens with the code.
Did you miss uploading some code?

PrinterLUA:
GM teleport feature can be achived through lua:
https://otland.net/threads/event-walkthrough-everything-ctrl-arrow-keys-tfs-1-x.226973/

vankk:
I agree with @PrinterLUA .

src/vocation.cpp
BenDol: Appears I forgot to copy something, will do it now.

src/creature.h
BenDol: I disagree, unless you have a better name this is perfectly fine.

BenDol:
It should be added to the tfs project none the less.

ranisalt:
> Added definitions.h to cmake source definitions.

Added **header** to **source**... are you sure? It's not needed :P

BenDol:
When run your project using cmake IDE's like qtcreator only load the files from the cmake definition.

ranisalt:
I use CLion, I have no problems and didn't understand what you mean. Anyway it's your IDE's fault, not TFS'.

BenDol:
It compiles fine, it just doesn't include definitions.h in the file list for some IDE's. Its not a big deal.

ranisalt:
Do the same happens with enums and const headers?

BenDol:
If its not in the cmake resource files then yes they won't be imported to the project. Since the IDE uses cmake to build the source file index.

src/luascript.cpp
```diff
@@ -7059,6 +7058,18 @@ int LuaScriptInterface::luaCreatureSetMaxHealth(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaCreatureIsHealthHidden(lua_State* L)
```
djarek: Is there any particular reason you moved this function?

conde2: He just grouped it with luaCreatureSetHiddenHealth

src/vocation.cpp
ranisalt: `deathpenaltyreduction` instead of `lessloss` sounds better

BenDol: I thought about that too but its very long winded

BenDol:
Don't have time to finish this PR, will leave it open for the ideas at least.

nekiro:
Closing this as it's not maintained and deprecated.

Issue #1591, opened 2015-10-12
Walkthrough Quest Door


brunominervino:
If one player made a quest and stand in quest door, anyone can enter:

![sem titulo](https://cloud.githubusercontent.com/assets/10432401/10440431/732f18c8-7115-11e5-9096-c241d21026d6.png)
It's a example.

marksamman:
This is a known issue, although it has only been brought up in a pull request which is now closed (#1083), so we could use this issue to track it.

PrinterLUA:
@marksamman this i thougt maybe do something like this:
	const Item* topItem = playerTile->getTopTopItem();
	if (topItem && topItem->getDoor() && !topItem->hasWalkStack()) {
		std::cout << "Door: " << topItem->getID() << std::endl;
		return false;
	}

HeavenIsLost:
printer idea is good i believe, but we will need to edit every door with the walkstack flag.

vankk:
No fix for this problem?

mrhultin:
Vankk, what I did to fix was add a action id to the Tile below and add this movement, it's obviously not a fix but a temporary solution (ie, atleast it wont be exploitable for now) 
pastebin: http://pastebin.com/kLFCHc2k
function onStepIn(creature, item, position, fromPosition)
	local player = creature:getPlayer()
	if not player then
		return true
	end

	if(player:getStorageValue(Storage.TheInquisition.Questline) >= 24) then
		return true
	end

	player:teleportTo(fromPosition)
	player:sendTextMessage(MESSAGE_EVENT_ADVANCE, "The door seems to be sealed against unwanted intruders.")
	return true
end

degolaot:
if not have a fix, remove the protection zone in sqm of door. I did it on my server ^^

TheSumm:
@mrhultin that's a pretty good fix but you should add that script to all quest doors and make the storage depends on the items unique id.

brunominervino:
@TheSumm yes, I'm using a temporary fix like this. But how I have much doors, it is a hard work,

conde2:
Great ideia !

mazen-mardini:
This should probably not be an item-issue but a character one. There should be a flag for characters (players only?) to prevent anyone from walking over them. This will of course override the global configuration.

mrhultin:
[...]

PR #1695, opened 2016-01-24
Level Door require level to walkthrough

PrinterLUA:
This should fix the issue with this:
https://github.com/otland/forgottenserver/issues/1591

### DIFF
--- a/data/movements/movements.xml
+++ b/data/movements/movements.xml
@@ -35,95 +35,185 @@
 
 	<!-- (Level & quest) doors -->
 	<movevent event="StepOut" itemid="1224" script="closingdoor.lua" />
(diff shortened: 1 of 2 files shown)


marksamman:
It won't. The proper solution is setting walk stack to 0 for the open doors, and making walk stack work for other items than ground. There should be a tile flag for walk stack.

PrinterLUA:
Ok, ill take a look at it then.

TheSumm:
This is the proper solution as multiple players should be able to pass through quest doors and gates of expertise.

marksamman:
I still think using walk stack is the better solution, but in that case I would be OK with merging a solution that solves it this way. This PR is more broken than just missing a player variable though, I suggest you test that your solution actually works before submitting a pull request.

PR #1698, opened 2016-01-27
Level Door require level to walkthrough

PrinterLUA:
I find this way more promising, since players who are worhty to pass. Then they should just walk through and enter the door, instead of making walkstack of only max 1 player.

This should fix this issue:
https://github.com/otland/forgottenserver/issues/1591


PrinterLUA:
Before i merge the commits, what do you think @marksamman

PrinterLUA:
Before i merge the commits, what do you think @marksamman

brunominervino:
Good idea @PrinterLUA, I will test.

data/movements/movements.xml
```diff
@@ -35,95 +35,185 @@
 
 	<!-- (Level & quest) doors -->
 	<movevent event="StepOut" itemid="1224" script="closingdoor.lua" />
+	<movevent event="StepIn" itemid="1224" script="opendoor.lua" />
```
marksamman: What's opendoor.lua?

PrinterLUA: my fault, it was named opendoor, but then i changed to enterdoor.

Ill fix it tomorrow. Anything else you want me to change?

marksamman: Yes. Split level doors and quest doors into two different script files, then you don't need the isInArray check.

conde2:
Thanks Printer, great ideia !!!

PrinterLUA:
@marksamman, done :)

PrinterLUA:
@marksamman will it get merged, or shall we wait for the onStepIn/onStepOut enchantments?

marksamman:
It doesn't need to wait for onStepIn/onStepOut changes, but you need to squash your commits. Have you also tested the changes?

PrinterLUA:
@marksamman yes and works perfect.
preview: http://2.1m.yt/THWGfKD.gif

wgrriffel:
@PrinterLUA It work as expected, only one thing i'd like to suggest to this PR, isn't good make a statement to gms and adms can walk through? like
`if player:getGroup():getAccess() == true then return false end`
imo it's should be onUse also, but it's just a thought, what you think?

PR #1699, opened 2016-01-27
[WIP] Load monsters transaction-like

### LINKED ISSUE #1611: Setting interval (in monsters *.xml files) to 0 (zero) crashes the server
In **monster files**, this piece of code:
```xml
	<summons maxSummons="2">
		<summon name="War Wolf" interval="0" chance="50" />
	</summons>
```
will *crash* the server **(when it tries to summon up a creature)** if, and **only if** the ``interval`` is set to **0** *(zero)*. I don't know if that's what it is intended to happen, because 0 is an undefined value to set an interval *(I guess those variables should be initialized to a set of default values)*, but shouldn't we handle it properly *by not letting the server crash* *(additionaly a notifying message in console)*?

ranisalt:
Fixes #1611 and possible more errors that would be caused by invalid parameters set on a monster configuration. I consider it a WIP since there may be more stuff to be checked or something that may not be checked too.

Monsters are built and checked before being added to the server (pretty much like a builder or factory) and invalid parameters are reported to the console. Non-breaking values (such as health < 0) do not prevent a monster loading, only issue a warning. Breaking errors (such as looktype < 0) prevent the monster from loading and, if it is being reloaded, do not change the previous one.

I intend to make it possible to either show on the console if the server is loading or show in the game chat if a gamemaster has issued `/reload` and also extend this pattern to other stuff like NPCs and spells.

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -700,9 +700,6 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
 
 bool Monsters::loadMonster(const std::string& file, const std::string& monsterName, std::list<std::pair<MonsterType*, std::string>>& monsterScriptList, bool reloading /*= false*/)
 {
-	MonsterType* mType = nullptr;
-	bool new_mType = true;
-
 	pugi::xml_document doc;
 	pugi::xml_parse_result result = doc.load_file(file.c_str());
 	if (!result) {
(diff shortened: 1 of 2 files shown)


marksamman:
IIRC using pugi::cast with uint8_t doesn't do what you think it does, it interprets a single character as its ASCII code. Have you checked if monsters with look head/body/legs/feet are loaded properly?

src/monsters.cpp
```diff
@@ -722,77 +719,71 @@ bool Monsters::loadMonster(const std::string& file, const std::string& monsterNa
...
+	mType.name = attr.as_string();
 
 	if ((attr = monsterNode.attribute("nameDescription"))) {
-		mType->nameDescription = attr.as_string();
+		mType.nameDescription = attr.as_string();
 	} else {
-		mType->nameDescription = "a " + mType->name;
-		toLowerCaseString(mType->nameDescription);
+		mType.nameDescription = "a " + mType.name;
+		toLowerCaseString(mType.nameDescription);
```
marksamman: You can change this to:
```
mType.nameDescription = "a " + asLowerCaseString(mType.name);
```

ranisalt:
Yes, you are right, lexical casting to (u)int8_t indeed reads a single character. I reverted the commit.

src/monsters.cpp
```diff
@@ -1130,10 +1137,20 @@ bool Monsters::loadMonster(const std::string& file, const std::string& monsterNa
...
+				if (speed <= 0) {
+					std::cout << "[Error - Monsters::loadMonster] Invalid summon speed " << mType.speed << ". " << file << std::endl;
+					return false;
+				}
 			}
 
 			if ((attr = summonNode.attribute("chance"))) {
 				chance = pugi::cast<int32_t>(attr.value());
+
+				if (speed <= 0) {
```
brunominervino: This is `chance` instead of `speed` right?

ranisalt: Fixed

src/monsters.cpp
ranisalt: True, fixed

ranisalt:
Actually there is a problem with `pugi::cast` as it silently ignores bad lexical cast exceptions. I'm going to skip it entirely since I'm interested on invalid values but not on failing and just assuming 0.

src/monsters.h
```diff
@@ -193,7 +196,7 @@ class Monsters
 		void loadLootContainer(const pugi::xml_node& node, LootBlock&);
 		bool loadLootItem(const pugi::xml_node& node, LootBlock&);
 
-		std::map<std::string, MonsterType> monsters;
+		std::unordered_map<std::string, MonsterType> monsters;
```
djarek: Is there a particular reason you changed map to unordered map? Can you make sure that the resulting iterator instability on insert doesn't cause silent bugs?

ranisalt: No, that was accidental and wasn't supposed to be commited.

src/monsters.cpp
```diff
@@ -722,77 +719,76 @@ bool Monsters::loadMonster(const std::string& file, const std::string& monsterNa
...
 	}
 
 	if ((attr = monsterNode.attribute("skull"))) {
-		mType->skull = getSkullType(attr.as_string());
+		mType.skull = getSkullType(attr.as_string());
 	}
 
 	if ((attr = monsterNode.attribute("script"))) {
-		monsterScriptList.emplace_back(mType, attr.as_string());
+		monsterScriptList.emplace_back(&mType, attr.as_string());
```
marksamman: Is there any guarantee that this will point to the object once it is moved into the monsters map, and that the monster object will retain the same position in memory, given that monsters is a std::map, which is usually implemented as a red-black tree, which can be re-balanced?

ranisalt: No. `monsterScriptList` could be modified to hold the monster name (as in the monster map) instead, or the emplacement could happen after updating the monster list.

marksamman: Even if you address monsterScriptList, you'll still have dangling pointers in monsters as they hold a MonsterType pointer.

ranisalt: Hmm that's true. It could be a shared pointer instead, but only new monsters will have the updated MonsterType values after a reload.

Or it could store a reference to the monster type list entry instead, since IIRC no monster type is ever removed, and the entry be swapped with the new MonsterType.

PR #1703, opened 2016-02-02
Use correct value for TEXTCOLOR_PURPLE

raymondtfr:
The correct value value for ``TEXTCOLOR_PURPLE`` is **155**.
The value 154 stands for dark magenta -- the value 155 is more like lilac.

textcolor_vanilla_ex


marksamman:
Thanks!

ranisalt:
Whoa that's a good eye you have.

raymondtfr:
@ranisalt Yeah, haha. I always noticed that something was off about this text color when I used to watch tibiacast recordings and playing with a sorcerer (using energy based attack spells) and compare it with tfs, but just now I took the time to investigate it.

PR #1705, opened 2016-02-04
Renamed data members using "m_"-style names

raymondtfr:
**Renamed** data members that have the prefix ``m_``.
**Use** ``this->`` whenever necessary.

Besides removing the **"m_"** prefix from some some data members in **npc** I also had to remove *(because member-functions and data members having the same name would conflict)* the **on** prefix -- i.e. ``m_onCreatureAppear`` to ``creatureAppear`` -- otherwise, VS 2015 would give the following errors:
* *declaration is incompatible with "int32_t NpcEventsHandler::onCreatureAppear".*
* *expression preceding parentheses of apparent call must have (pointer-to-) function type.*
* *member "NpcEventsHandler::onCreatureAppear" is inaccessible.*

Additionally, I had to rename data members ``isLeaf`` to ``leaf`` and ``isLoaded`` to ``loaded`` *(I'm not sure if in this case, the clarity of the code is compromised)* for the same reasons.

**This is something that is on the [Roadmap for 1.2] (https://github.com/otland/forgottenserver/wiki/Roadmap-1.2).**

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -33,9 +33,9 @@ extern Actions* g_actions;
 extern ConfigManager g_config;
 
 Actions::Actions() :
-	m_scriptInterface("Action Interface")
+	scriptInterface("Action Interface")
 {
-	m_scriptInterface.initState();
+	scriptInterface.initState();
 }
 
 Actions::~Actions()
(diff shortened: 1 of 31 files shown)


marksamman:
Thanks!

PR #1709, opened 2016-02-07
Minor fix to spells

raymondtfr:
**Electrify** now has the correct *effect*.
**Great Energy Beam** has now the correct *area*.


marksamman:
Thanks!

PR #1710, opened 2016-02-07
New default value for the spread attribute

raymondtfr:
Changed the default value of the spread attribute to 0 (zero).
Therefore, this will avoid inumerous unnecessary occurences of (spread="0") across monster files.

Currently, there are 126 occurences of(spread="0") across 107 files.

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -383,7 +383,7 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
 		if ((attr = node.attribute("length"))) {
 			int32_t length = pugi::cast<int32_t>(attr.value());
 			if (length > 0) {
-				int32_t spread = 3;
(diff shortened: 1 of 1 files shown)


marksamman:
What about monsters that were written relying on the default behavior? My opinion is that this change is unnecessary. Yes, 0 might be a more sane default value, but what's better is that spread is explicitly included in all monster files.

raymondtfr:
> What about monsters that were written relying on the default behavior?
You mean creating a wave attack with ``length="x"`` and and omit writing ``spread="3"`` because that's just exactly the *spread* value the user would set?
I concur about writing them explicitly, I thought about changing this value because its possible to do that with the ``target="x"`` (it is ``false`` by defaut) attribute.

I guess this should be closed then.

raymondtfr:
Oh, I get it. This will break monster spell configurations that relies on the default behavior you mentioned.

PR #1713, opened 2016-02-10
Fix #1712

### LINKED ISSUE #1712: desintegrate rune like tibia rl
Actually the [desintegrate rune ](https://github.com/otland/forgottenserver/blob/master/data/spells/scripts/support/desintegrate%20rune.lua)is not working as suppost to be [like tibia rl](http://tibia.wikia.com/wiki/Disintegrate_Rune)

ninjalulz:


data/spells/scripts/support/disintegrate rune.lua
```diff
@@ -0,0 +1,24 @@
...
+local removalLimit = 500
+
+function onCastSpell(creature, variant, isHotkey)
+	local position = variant:getPosition()
+	local tile = Tile(position)
+	if tile then
+		local items = tile:getItems()
+		if items then
+			for i, item in ipairs(items) do
+				if not isInArray(corpseIds, item:getId()) and item:getType():isMovable() and item:getUniqueId() > 65535 and item:getActionId() == 0 then
```
ranisalt: I think `isInArray` is the most expensive check and could be moved to the end of the comparison.

ninjalulz: Yeah, you're probably right. I'll move it to the end, thanks!

marksamman:
Thanks!

PR #1714, opened 2016-02-11
fix #1622

### LINKED ISSUE #1622: [FIXED] Bug with pvp zone
Seems PvP Zone is bugged at last TFS. When you attack a player at pvp zone you get pz locked (http://2.1m.yt/rO0BHKBl.png). Normally pvp zone doesn't get pz locked, only battle.

Any fix for this? Thanks.

vankk:
#1622

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -3501,6 +3501,10 @@ void Player::onAttackedCreature(Creature* target)
 {
 	Creature::onAttackedCreature(target);
 
+	if (target && target->getZone() == ZONE_PVP) {
+		return;
+	}
+	
 	if (target == this) {
 		addInFightTicks();
 		return;


sgioia9:
I don't like the way it's being fixed, but it's just my personal opinion. It looks forced.

vankk:
We have another way to fix it? Also why looks forced to you, did not understand your view.

sgioia9:
I misread, LGTM

marksamman:
Thanks!

src/player.cpp
```diff
@@ -3501,6 +3501,10 @@ void Player::onAttackedCreature(Creature* target)
 {
 	Creature::onAttackedCreature(target);
 
+	if (target && target->getZone() == ZONE_PVP) {
+		return;
+	}
+
```
WibbenZ: Trailing tab, remove it in another PR.

PR #1716, opened 2016-02-13
Added speech bubble

Tulioh:


WibbenZ:
I really don't get the point of this, we could do it in Lua (NPC system) with like 1 line of code.
I propsed this change a while ago and Mark did not want it, since we are not trying to copy Cipsoft.
The code might be on the ORTS project, not sure tho.

But if Mark agrees with merging an update like this, just change the Lua files insted.
* Tip; Serach for setSpeechBubble and just set a default one if no other speech bubble is set.

marksamman:
It is intentional that NPC's don't have any speech bubble by default.

PR #1726, opened 2016-02-23
Missing creatures & rebuild

vomar:


data/monster/Arena/greenhorn/achad.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<monster name="Achad" nameDescription="achad" race="blood" experience="70" speed="185" manacost="0">
+<monster name="Achad" nameDescription="Achad" race="blood" experience="70" speed="185" manacost="0">
```
raymondtfr: Strings inside ``nameDescription=""`` are no longer case-sensitive -- all strings are converted to lowercase.

ranisalt: @raymondtfr wrong, if there is a `nameDescription` attribute then it is used, else it is `"a " + mType->name;`. Check monsters.cpp L740

raymondtfr: Oh, I thought about the name description it shows when you click in a creature, which really is all lower case.

DSpeichert:
@WibbenZ is this something we want/can review? It seems like most of the monsters were updated since then.

WibbenZ:
@DSpeichert IMO it's way to long to review and we probably have most of the changes by now.
Checked Amazon and the intervals are not changed but the grammar is updated

DSpeichert:
@vomar Please resubmit the PR with changes rebased from latest master due to the huge number of conflicts in here.

PR #1728, opened 2016-02-25
use TCP_NODELAY

divinity76:
this can significantly reduce lag when walking fast or when there's many players around, a much smoother walking experience in debian 6 servers has been observed in older tfs versions, but i don't see newest TFS doing this, unless boost is doing it behind my back..

just to be clear, i have not tested this code in a recent tfs, i simply back-ported it from a proprietary server (actually 3 separate ones, where i got paid to do it :p )

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -25,6 +25,15 @@
 #include "protocol.h"
 #include "scheduler.h"
 #include "server.h"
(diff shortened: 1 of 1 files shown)


marksamman:
https://github.com/otland/forgottenserver/blob/master/src/server.cpp#L193

divinity76:
oh great, nevermind ^^

PR #1730, opened 2016-02-25
Add an issue template

ranisalt:
Now that Github has added templates for issues and pull requests, it would be nice to add one and prevent people creating useless reports just for us to ask "what system?" "what compiler?" "did you checkout to the latest version?" and remember people with prehistoric compilers do not work.

Please check https://github.com/ranisalt/node-argon2/issues/new to see how it works

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,34 +1,22 @@
 Welcome to the issues section if it's your first time!
 
-### Before creating an issue, please be sure to:
-- [ ] Checkout to the latest version and recompile if you haven't
-- [ ] Try to find an *isolated* way to reproduce the behavior
-- [ ] Use the support board first: https://otland.net/forums/support.16/
-- [ ] Search if there's already a similar issue, and bump it instead
-- [ ] Fill in all the blanks in the most specific way you can
+### Before creating an issue, please ensure:
+- [ ] This is a bug in the software that resides in this repository, and not a
+      support matter (use https://otland.net/forums/support.16/ for support)
+- [ ] You have not made changes to the source code related to the problem
+- [ ] Search for a similar issue, add your reaction there instead :tada:
 
-### Steps to reproduce
+### Steps to reproduce (include any configuration/script required to reproduce)
 1.
 2.
 3.
 
 ### Expected behaviour
-Tell us what should happen.
+Tell us what should happen
 
 ### Actual behaviour
-Tell us what happens instead. Post your faulty script, if there is any.
+Tell us what happens instead
 
 ### Environment
-**Operating system**:
-(in Linux, you can get with `cat /etc/os-release`)
-
-**Compiler version**:
-(you can get with `clang --version` or `gcc --version`)
-pro-tip: clang < 3.3 and gcc < 4.8 **do not work**
-
-**Dependencies**:
-- Boost version:
-- GMP version:
-- PugiXML version:
-- Lua/LuaJIT version: (you can get with `lua -v` or `luajit -v `)
-- MySQL version: (you can get with `mysql -v`)
+If the issue is environment specific (e.g. compiling errors), please include the
+name and version of the operating system and compiler you are using.
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -3,8 +3,7 @@ Welcome to the issues section if it's your first time!
 ### Before creating an issue, please ensure:
 - [ ] This is a bug in the software that resides in this repository, and not a
       support matter (use https://otland.net/forums/support.16/ for support)
-- [ ] You have not made changes to the source code related to the problem
-- [ ] Search for a similar issue, add your reaction there instead :tada:
+- [ ] This issue is reproducible without changes to the code in this repository
 
 ### Steps to reproduce (include any configuration/script required to reproduce)
 1.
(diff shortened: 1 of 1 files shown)


WibbenZ:
This... would be perfect.
Would make even more sence on the support board haha :+1:

ranisalt:
> Would make even more sence on the support board haha

Good one, forgot to warn people to check the support board before creating an issue

Tarliton:
Is it good to add this?
Before creating an issue, please be sure to:
-try to install all the dependencies and compile again
I remember 2 or 3 issues open and the reason was that pugixml became a dependency

ranisalt:
> Is it good to add this?

It is, but I think it's enough to ask for the dependency version. If one looks at "pugixml version" and does not has it installed, it will ring the bell.

ranisalt:
Anything holding this back?

marksamman:
I think it's asking too much which may discourage people from reporting issues. The biggest problem with issues is people using them for support matters, and sometimes writing comments that don't contribute to solving the issue. I don't recall any issue where MySQL version, PugiXML version or GMP version has mattered. There aren't many issues where the environment matters. It's also encouraging spam when it says to bump issues (as in just posting "bump", or "+1" or anything else that doesn't contribute to the issue).

ranisalt:
> It's also encouraging spam when it says to bump issues (as in just posting "bump", or "+1" or anything else that doesn't contribute to the issue)

Hmmm I thought the opposite, it could prevent some duplicate issues which happens pretty a lot for older bugs (or sometimes not even a bug). I agree with the rest.

ranisalt:
> It's also encouraging spam when it says to bump issues

Oh yeah, Github has just released an official feature to "+1" comments (and more): https://github.com/blog/2119-pull-request-and-issue-reactions I think on including it instead.

marksamman:
That's a step in the right direction for GitHub, should make issues less noisy. As for this issue, I think something simpler, like this, would be better:
```
By creating this issue, I confirm that:
- [ ] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [ ] I have not made any changes to the source code

### Steps to reproduce (include any configuration/script that is required to reproduce the issue)
1.
2.
3.

### Expected behavior
Tell us what should happen

### Actual behavior
Tell us what happened instead

### Environment
If the issue is environment specific (e.g. compiling errors), please include the name and version of the operating system and compiler you are using
```

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
+Welcome to the issues section if it's your first time!
+
+### Before creating an issue, please ensure:
+- [ ] This is a bug in the software that resides in this repository, and not a
+      support matter (use https://otland.net/forums/support.16/ for support)
+- [ ] You have not made changes to the source code related to the problem
```
marksamman: I'd rather be stricter about this, or at least reword it to say something like:
"The issue is reproducible without changes to the source code"

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
...
+- [ ] Search for a similar issue, add your reaction there instead :tada:
```
marksamman: This depends on the situation, and I prefer to not have it in the template. Sometimes the person who would open a duplicate issue may complement the original issue with additional information, and I'd rather read their comment than just see their reaction even though the issue will be closed and marked as duplicate.

ranisalt: Better not to have or to rewrite to `add your reaction and complement information`?

marksamman: It's better to remove it altogether, I'd like to keep the template simple.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+Welcome to the issues section if it's your first time!
```
marksamman: This should be commented out.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+      support matter (use https://otland.net/forums/support.16/ for support)
+- [ ] This issue is reproducible without changes to the code in this repository
+
+### Steps to reproduce (include any configuration/script required to reproduce)
+1.
+2.
+3.
+
+### Expected behaviour
+Tell us what should happen
```
marksamman: This should also be commented out.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Actual behaviour
+Tell us what happens instead
```
marksamman: This should also be commented out.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Environment
+If the issue is environment specific (e.g. compiling errors), please include the
```
marksamman: This should also be commented out.

ISSUE_TEMPLATE.md
ranisalt: You mean, comment so that one creating an issue do not post this text accidentally?

marksamman: Yes. By wrapping it in ""

ranisalt: Good. HTML comments are prettier than Markdown comments.

marksamman:
Thanks!

PR #1731, opened 2016-02-29
Function updated

vankk:
Now we can use _player:doPlayerAddExp(exp, multi)_ instead of _doPlayerAddExp(player:getId(),exp, multi)_


diath:
That function's in compat for a reason - to keep backwards compatibility with old scripts.

vankk:
Oh, I'm sorry, I did not know that.

ranisalt:
Also there's `player:addExp` which is what that function uses.

vankk:
@ranisalt 
You didn't mean `player:addExperience` [1692](https://github.com/otland/forgottenserver/blob/c4b10a905b11728522e715bd41f2493858ab0e46/src/player.cpp#L1692)?

vankk:
maybe `player:addExperience` is missing experience multiply value?

ranisalt:
Yes, I meant, and no, you should calculate multipliers prior to applying addExperience.

vankk:
@ranisalt 
I know this is a support board but could you give me a example of that?

PR #1734, opened 2016-03-03
Minor Pick Changes, Fixes #1733

### LINKED ISSUE #1733: Splash/Blood on Hole
If any type of splash/blood is on hole you can't use the pick, it is the same problem of #1631 but now with pick.

ninjalulz:


marksamman:
Was the magic effect intentionally removed?

ninjalulz:
Nope, there are no longer any poff effect for picking holes.

marksamman:
Thanks!

Issue #1729, opened 2016-02-25
Wrong ID for Demonic Essence in Hellspawns


Marahin:
https://github.com/otland/forgottenserver/blob/master/data/monster/Demons/hellspawn.xml#L62 
I'm pretty sure it shouldn't be 10000 (carrot cake), but 6500 (demonic essence).

PR #1735, opened 2016-03-03
Fix demonic essence id

vankk:
#1729


Tarliton:
I'm pretty sure you need to squash those commits

marksamman:
As @Tarliton already said, please squash your commits.

PR #1737, opened 2016-03-07
Just update the item id.

vankk:
patch 10.22


marksamman:
Thanks!

ranisalt:
"Update die" was an excellent commit message :laughing: :laughing: :laughing:

PR #1741, opened 2016-03-10
Fix effect for not attackable creatures

raymondtfr:
This ensures the correct block type effect is used - when a  creature is
not attackable -- i.e. Lava Hole - otherwise we see two different block type effects
"CONST_ME_POFF" and "CONST_ME_BLOCKHIT" at the same time.

This is something I didn't notice in #1655 .

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3872,7 +3872,7 @@ bool Game::combatChangeHealth(Creature* attacker, Creature* target, CombatDamage
 	} else {
 		if (!target->isAttackable()) {
 			if (!target->isInGhostMode()) {
-				addMagicEffect(targetPos, CONST_ME_POFF);
+				addMagicEffect(targetPos, CONST_ME_BLOCKHIT);
 			}
 			return true;
 		}


marksamman:
What do you mean with that we see two different effects at the same time? If there's already an effect happening somewhere else with BLOCKHIT, wouldn't it be better to just remove this line instead of sending the same effect twice?

raymondtfr:
> What do you mean with that we see two different effects at the same time?

Taking for example, the monster *Lavahole*, considering it has flag ``attackable`` set to ``1`` and it is immune to all kinds of damage, in this case, I'll be using an energy-based attack spell.

First ``BlockType_t Creature::blockHit`` will set ``blockType`` type to ``BLOCK_IMMUNITY``, then ``bool Game::combatBlockHit`` will set the ``hitEffect`` to ``CONST_ME_BLOCKHIT`` then ``addMagicEffect(targetPos, hitEffect);`` happens.
Next call comes ``bool Game::combatChangeHealth`` in which it'll come to the if statement containing the condition ``!target->isAttackable()`` which will add another magic effect, that is, ``CONST_ME_POFF``.

So we see the **yellow spark** effect and the **poff**, happening simultaneously.

mwsnap017


> If there's already an effect happening somewhere else with BLOCKHIT, wouldn't it be better to just remove this line instead of sending the same effect twice?

If we do that, there won't be a default effect in case a monster has flag ``attackable`` set to ``true``.
I fixed it with a lambda:

```C++
		if (!target->isAttackable()) {
			if (!target->isInGhostMode()) {
				[this](BlockType_t blockType, const Position& targetPos) {
					if (blockType == BLOCK_NONE) {
						addMagicEffect(targetPos, CONST_ME_POFF);
					}
				};
			}
			return true;
		}
```

instead of:

```C++
		if (!target->isAttackable()) {
			if (!target->isInGhostMode()) {
				addMagicEffect(targetPos, CONST_ME_POFF);
			}
			return true;
		}
```

By the way, I've tested in tibia and all kind of ``BLOCK_IMMUNITY`` block type effects should be (CONST_ME_BLOCKHIT), even for ``COMBAT_EARTHDAMAGE`` and ``COMBAT_HOLYDAMAGE``. (in ``Game::combatBlockHit``).

ranisalt:
You don't need the lambda, just add the code. Actually, your lambda isn't running at all, because you define it but do not store it neither run it (like a function), but as you are aiming to remove the effect it works... for your specific try.

```cpp
        if (!target->isAttackable()) {
            if (!target->isInGhostMode() && blockType == BLOCK_NONE) {
                addMagicEffect(targetPos, CONST_ME_POFF);
            }
            return true;
        }
```

raymondtfr:
@ranisalt But ``blockType`` isn't defined.

ranisalt:
Well, if the lambda isn't ever called then it doesn't even need to exist. Where does the block type would come from then?

marksamman:
So does removing that line achieve the same thing, or should this be merged?

ninjalulz:
You achieve the same thing by removing that line, and shouldn't the same line be removed from combatChangeMana as well?

raymondtfr:
> So does removing that line achieve the same thing, or should this be merged?

Yes, it does, although there will be no effect in case a creature has attackable flag set to true, which isn't a very important detail.

I will update this with removing that line.

> You achieve the same thing by removing that line, and shouldn't the same line be removed from combatChangeMana as well?

Yeah, I didn't look at this, should I make this change too?

ninjalulz:
> although there will be no effect in case a creature has attackable flag set to true
Why wouldn't there be any effect?

> Yeah, I didn't look at this, should I make this change too?
No, that isn't necessary as there are some differences between health and mana functions.

raymondtfr:
@ninjalulz There won't be any effect if the monster flag is set to ``attackable=0`` and it is not immune to  the damage type. The default effect in this case should be the yellow sparkle.

PR #1752, opened 2016-04-03
Fix cooldown of spell Curse

raymondtfr:
Vanilla's site has incorrect information about the cooldown for this spell.

compare_spell_cooldown


marksamman:
Thanks!

PR #1756, opened 2016-04-06
Remove most of underscore names

ranisalt:
Almost all underscore-named variables were removed from the code. There's only one missing in `internalMoveItem` where both an argument and an internal variable are called `moveItem` and I couldn't figure a suitable name for any.

Also I refactored the constructors I stumbled upon with RAII.

Where a function would have both an argument and a variable with the same name (e.g. `_dir` and `dir` in `Monster::getRandomStep`, I renamed the argument to `direction` and the internal is still `dir`.

This PR is very open to suggestions.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -39,7 +39,7 @@ Creature::Creature() :
 	referenceCounter = 0;
 
 	id = 0;
-	_tile = nullptr;
+	tile = nullptr;
 	direction = DIRECTION_SOUTH;
 	master = nullptr;
 	lootDrop = true;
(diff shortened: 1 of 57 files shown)


src/creature.cpp
```diff
@@ -652,11 +652,11 @@ void Creature::onDeath()
 {
 	bool lastHitUnjustified = false;
 	bool mostDamageUnjustified = false;
-	Creature* _lastHitCreature = g_game.getCreatureByID(lastHitCreature);
+	Creature* lastHitCreature = g_game.getCreatureByID(this->lastHitCreature);
```
marksamman: The data member lastHitCreature should be renamed to lastHitCreatureID

PrinterLUA:
Great work!

src/creature.cpp
ranisalt: Done. Also renamed a few occurrences of `houseid` and `townid` to `houseId` and `townId` for consistency.

marksamman:
Thanks!

PR #1759, opened 2016-04-09
Replace Boost with standalone Asio

ranisalt:
This drops dependency on Boost (as stated on [roadmap](https://github.com/otland/forgottenserver/wiki/Roadmap)) by replacing all usages of `boost::asio` with `asio`, which also packs other related Boost stuff such as error codes. 

There is one less package for debian-based users and one less library to link against, since asio is header-only. I added a CMake script to find the library, too.

`boost::asio::deadline_timer` is replaced by `asio::basic_waitable_timer<std::chrono::steady_clock>`, which are basically the same except for how they get the `now()` timestamp, the former requiring boost time libraries and the latter capable of using STL. [(source)](http://stackoverflow.com/questions/26501936/difference-between-deadline-timerand-waitable-timer-in-boost-asio)

**Please note** on most distros asio will still pull Boost. I'm almost certain it's because some functions use Boost stuff (like the aforementioned deadline timer), but tfs itself **won't** link against boost. You can just compile, then purge Boost. If one wants to be sure to not depend on it, it's possible to `#define ASIO_STANDALONE`, but compilers will do this job.


marksamman:
We still depends on Boost headers for lockfree, lexical_cast and reverse iterator. We can't rely on asio to bring them in. Either you'll have to find suitable replacements for them or Boost headers must still be required in CMake and installed by Travis.

ranisalt:
Good, I already have a solution for reverse iterator and will have a look on lexical cast and lockfree.

src/pugicast.h
```diff
@@ -20,18 +20,13 @@
...
 	T cast(const pugi::char_t* str)
 	{
 		T value;
-		try {
-			value = boost::lexical_cast<T>(str);
-		} catch (boost::bad_lexical_cast&) {
-			value = T();
-		}
+		std::stringstream ss(str);
+		ss >> value;
```
ranisalt: @marksamman is this acceptable as a lexical cast replacement? If extraction fails, `value` will be left unchanged (in this case, will be a default-constructed value) so it should perform the same conversion. [(source)](http://en.cppreference.com/w/cpp/io/basic_istream/operator_gtgt)

The same happens on DBResult, and `operator>>` has a lot of overloads.

marksamman: It's probably significantly slower than lexical_cast, but it's only used to load config files so it's OK.

src/database.h
```diff
@@ -161,11 +159,8 @@ class DBResult
 			}
 
 			T data;
-			try {
-				data = boost::lexical_cast<T>(row[it->second]);
-			} catch (boost::bad_lexical_cast&) {
-				data = 0;
-			}
+			std::stringstream ss(row[it->second]);
+			ss >> data;
```
marksamman: For DBResult the performance should probably be measured to see if it's a significant difference (which I think it is), and if it is, we should add specialized templates for the most common types.

ranisalt: You are right, according to my tests string stream is generally 3x slower than `boost::lexical_cast` except for float.

Specialized templates works best and is 2x faster than `boost::lexical_cast` on my machine, but a whopping 10x faster for floats.

Sample code: https://gist.github.com/ranisalt/7bb2af8b7fa21dd7eeb2b4e8f7aef156
I will create a cast function that both `pugi::cast` and `DBResult` will use based on specialized templates.

marksamman: LGTM. The static istringstream object should be thread local to prevent race conditions though.

ranisalt: That fallback should rarely be called, if ever. Going to fix it.

marksamman:
The commits and discussion in #1284 should be worth looking at when considering a replacement for lockfree.

ranisalt:
Mighty wizard @djarek I summon you.

I see you did try to implement a lockfree stack but after the discussion you decided by boost::lockfree which was the best option at that time, but what was missing from your lockfree stack?

kornholi:
I don't see the point in getting rid of Boost. All we're doing here is replacing tested, working code with own implementations that might not work for all edge cases. Boost has a lot of useful features, and I think we haven't been using them since getting rid of boost was on the roadmap (why?).

If having boost as a dependency is too difficult for some, we can always do something like [ycmd](https://github.com/Valloric/ycmd/tree/09f216441f99c02d51f9460e006683c5fe83669e/cpp/BoostParts) and copy needed parts to the repo with [bcp](http://www.boost.org/doc/libs/1_60_0/tools/bcp/doc/html/index.html). However, @marksamman recently removed pugixml from the tree and added it as a dependency, so I'm not sure what our strategy is.

marksamman:
The reason pugixml was in the repository was because it wasn't included in Debian's repository (and probably many other Linux distributions at the time) before Debian Jessie, so installing it wasn't straightforward. The same can be said for Boost on Windows because it's at least one additional step when using Visual Studio (and probably an additional hour of reading documentation and compiling if you're not using the precompiled binary). It's too heavy to be bundled in our SDK, but perhaps we can bundle a minimal version using bcp.

The roadmap entry about dropping Boost also says that we should consider using more of it, which we were leaning towards when #1284 was accepted using lockfree from Boost, so we're not avoiding Boost if Boost is our best option, but there really hasn't been much else to use Boost for, especially since mutex, thread and shared_ptr were added in C++11. The only thing I can think of is SHA1, which is in the detail namespace of Boost, so I think that should be avoided. Speaking of SHA1, we need something stronger to protect passwords with... but I don't think Boost has anything to offer in that area.

With C++17 we may have networking in the standard library, so then we won't even need asio. But for now, we should still allow using asio through Boost (I think it's easily achieved with a "using namespace" wrapped in an #ifdef) even if this PR is merged.

ranisalt:
Given the situation, I think the best is to still depend on Boost for lockfree (and maybe revert Asio) while pushing it out wherever possible. It will be easier once C++17 is available to don't depend on Boost for networking, and FWIW Boost's aim is to complement STL, not replace it, so we can only use what's really needed.

> Speaking of SHA1, we need something stronger to protect passwords with... but I don't think Boost has anything to offer in that area.

Once you said we could wait for PHC to end and use the chosen algorithm, and I implemented it based on libsodium (https://github.com/ranisalt/forgottenserver/tree/libsodium), but since Debian/Ubuntu takes ages to update software I guess we'll have to wait a lot.

djarek:
@ranisalt 
The custom lockfree stack implementation was complete, however, I've decided to use boost because I didn't want such a piece of complex logic (let's be honest, there are very few people in the OT community who have a clue about lockfree stuff) in this codebase. I'd strongly recommend against having the custom implementation in here.

HeavenIsLost:
I'm against removing boost. Cmake makes it really easy to add and remove boost dependency and downloading the boost library is pretty easy on windows and on linux it just one more command inside the whole download command.
Boost got a lot of good things that can be used in the future, in the oficial distro and in ours custom distros.

ranisalt:
> Boost got a lot of good things that can be used in the future, in the oficial distro and in ours custom distros.

I disagree. C++ tends to incorporate major parts of Boost in the future, as it has already done with C++11 and C++14, futurely with C++17 and proposals. That said, if we have today the tools to easily depend less on Boost (e.g. in this PR removing 2 usages, lexical cast and reverse iterator), we should since that's the direction the language is pointing to.

Now for personal matters, Boost here occupies 150 MB (while asio is only 2.5 MB) on a reduced SSD just for a few utilities, for me it's unbearable.

As it's pretty hard to implement a pure C++ lockfree stack (in C++11 at least) and that's the only feature missing, we could use something like [boost-bcp](http://www.boost.org/doc/libs/1_60_0/tools/bcp/doc/html/index.html) to have a small portion of the Boost library in the repository without depending on the entirety of Boost, and also get the benefits of static libraries.

ranisalt:
I could not make any progress on a strong lockfree stack and thus will close this PR by now, but it was the only missing feature I could not find suitable replacement. Other stuff (asio, reverse adaptor) are easily replaceable, meaning we do not really rely that much on Boost.

I'll see if I can take care of that in the future.

Issue #971, opened 2014-08-31
Monsters do not create summons if there are field items present


Luckey1729:
Is it intended that a monster won't make summons on fields unless they're explicitly immune to the element? It makes fields very powerful if, for example, a boss won't spawn its summons simply because they aren't immune to an element type.

Issue #928, opened 2014-07-29
Monster summon bug


Nu77:
I don't know if this is a bug but if we only run away with master and summon lost target on return to summon radius no one move again.

PS: Tested with slime

marksamman:
Can anyone confirm that the summon should be teleported to the master?

ninjalulz:
It shouldn't be teleported to the master but it should disappear if you are:
* two floors above or beneath it
* 30+(?) SQM's away from it

Also the summon shouldn't become idle if the master is out of range.

marksamman:
When you say that it shouldn't become idle, what do you mean that the summon should do? Take random steps? Only use defense abilities (e.g. heal)?

ninjalulz:
Sorry, forgot to mention that. They take random steps, and use their defense abilities (e.g heal).

Issue #1660, opened 2016-01-01
Crash


degolaot:
(gdb) thread 1
[Switching to thread 1 (Thread 0x788eca8ff700 (LWP 12614))]
#0  0x00000000008eaedc in Item::getTile (this=0x788eac1f02f0)
    at /home/Viking/src/item.cpp:341
341             if (cylinder->getParent()) {
(gdb) bt
#0  0x00000000008eaedc in Item::getTile (this=0x788eac1f02f0)
    at /home/Viking/src/item.cpp:341
#1  0x0000000000a60e8b in Thing::getPosition (this=0x788eac1f02f0)
    at /home/Viking/src/thing.cpp:27
#2  0x000000000094e562 in LuaScriptInterface::luaItemGetPosition (L=0x29d9210)
    at /home/Viking/src/luascript.cpp:6063
#3  0x0000788ecc6bf61d in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#4  0x0000788ecc6ca9b4 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#5  0x0000788ecc6bf989 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#6  0x0000788ecc6befac in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#7  0x0000788ecc6bfbc1 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#8  0x0000788ecc6bbc9d in lua_pcallk ()
   from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#9  0x00000000008fbc78 in LuaScriptInterface::protectedCall (L=0x29d9210,
    nargs=3, nresults=1) at /home/Viking/src/luascript.cpp:297
#10 0x00000000008fca71 in LuaScriptInterface::callFunction (
    this=0x788ec4278950, params=3) at /home/Viking/src/luascript.cpp:526
#11 0x00000000009a0e18 in MoveEvent::executeAddRemItem (this=0x788ec4e808a0,
    item=0x788eac1f02f0, tileItem=0x0, pos=...)
    at /home/Viking/src/movement.cpp:907
#12 0x00000000009a0cc0 in MoveEvent::fireAddRemItem (this=0x788ec4e808a0,
    item=0x788eac1f02f0, tileItem=0x0, pos=...)
    at /home/Viking/src/movement.cpp:882
[...]

Issue #1621, opened 2015-11-24
[Error - MoveEvent::executeStep] Call stack overflor


brunominervino:
[![Youtube Movie](http://img.youtube.com/vi/ube4b-q4Ncg/0.jpg)](http://www.youtube.com/watch?v=ube4b-q4Ncg)

If anyone interested, I was trying this:
```c++
bool MoveEvent::executeStep(Creature* creature, Item* item, const Position& pos, const Position& fromPos)
{
	//onStepIn(creature, item, pos, fromPosition)
	//onStepOut(creature, item, pos, fromPosition)
	if (!m_scriptInterface->reserveScriptEnv()) {
		std::cout << "[Error - MoveEvent::executeStep] Call stack overflow" << std::endl;
		return false;
	}

	ScriptEnvironment* env = m_scriptInterface->getScriptEnv();
	env->setScriptId(m_scriptId, m_scriptInterface);

	lua_State* L = m_scriptInterface->getLuaState();

	m_scriptInterface->pushFunction(m_scriptId);
	LuaScriptInterface::pushUserdata<Creature>(L, creature);
	LuaScriptInterface::setCreatureMetatable(L, -1, creature);
	LuaScriptInterface::pushThing(L, item);
	LuaScriptInterface::pushPosition(L, pos);
	Tile* tile = g_game.map.getTile(fromPos);
	Player* player = creature->getPlayer();
	LuaScriptInterface::pushPosition(L, !tile->getGround()->isBlocking ? fromPos : player->getTemplePosition());
	return m_scriptInterface->callFunction(4);
}
```
It is not working, but it can be a start ^^

vankk:
In chest quests can have it too, right? Doing the same thing.

brunominervino:
Anywhere that has fromPos, it happens.

Weboa:
well it was noticed : P

vankk:
hey @marksamman
You can't tell us how did you fix this problem in ShadowCores?

Thanks.

darkjav:
Yes, i had the same error, and solved with "not-logout" zone there...

ranisalt:
I think we could executeOnStep on login and, if it returns false, teleport to temple. But it should happen before the player is connected so he does not see the teleport happening.

PrinterLUA:
I dont understand how he end up there, did you login near that position that has a fromPosition code or what?

ranisalt:
[...]

PR #1761, opened 2016-04-11
Summon Fixes #665 #928 #971

### LINKED ISSUE #665: Monster summon not working properly
When a player is around, monsters from the same floor and some below/above start to move. The problem is: they start to summon too. So imagine places like the Alchemist Quarter or Drefia, when many creatures are capable to summon, this can get very annoying.

The monster should only summon when a player is targeted -in the same way attacks are triggered-, it works that way in Tibia, tested right now.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1446529-monster-summon-not-working-properly?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

PrinterLUA:
Should close these issues:
https://github.com/otland/forgottenserver/issues/665
https://github.com/otland/forgottenserver/issues/928
https://github.com/otland/forgottenserver/issues/971

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -246,7 +246,7 @@ void Monster::onCreatureMove(Creature* creature, const Tile* newTile, const Posi
 		}
 
 		if (canSeeNewPos && isSummon() && getMaster() == creature) {
-			isMasterInRange = true;    //Turn the summon on again
+			isMasterInRange = true;    //Follow master again
 		}
 
 		updateIdleStatus();
@@ -417,9 +417,8 @@ void Monster::onCreatureEnter(Creature* creature)
 	// std::cout << "onCreatureEnter - " << creature->getName() << std::endl;
 
 	if (getMaster() == creature) {
-		//Turn the summon on again
+		//Follow master again
 		isMasterInRange = true;
-		updateIdleStatus();
 	}
 
 	onCreatureFound(creature, true);
@@ -472,9 +471,8 @@ void Monster::onCreatureLeave(Creature* creature)
 	// std::cout << "onCreatureLeave - " << creature->getName() << std::endl;
 
 	if (getMaster() == creature) {
-		//Turn the monster off until its master comes back
+		//Take random steps and only use defense abilities (e.g. heal) until its master comes back
 		isMasterInRange = false;
-		updateIdleStatus();
 	}
 
 	//update friendList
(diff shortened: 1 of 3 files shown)


src/monster.cpp
```diff
@@ -932,7 +926,7 @@ void Monster::onThinkDefense(uint32_t interval)
 		}
 	}
 
-	if (!isSummon() && summons.size() < mType->maxSummons) {
+	if (!isSummon() && summons.size() < mType->maxSummons && hasFollowPath) {
```
EvulMastah: Wouldnt it be better to check`attackedCreature` instead?

PrinterLUA: Maybe, i thougt hasFollowPath check if the creature is reachable. But it does not. I have no idéa how it should behave, either summon when it has a target or summon if target is reachable.

PrinterLUA: @marksamman 
I would love some feedback, so i can make the necessary changes, if there is needed.

marksamman: There isn't much to comment code-wise. I'm mostly concerned about the side effects of these changes. Have they been thoroughly tested? Otherwise I rather keep this PR open until after the 1.2 release.

PrinterLUA: I have not tested them throughly to be honest. But i have tested and everything run smooth.

Any tests you want me to do, since i tried the changes and they behaved like they were coded todo.

marksamman: I'm not concerned that they're not working as intended, I'm concerned for unintended side effects that some of the changes may have. Since I don't know how long it'll take before we can release 1.2, I'll just merge this and worst case revert it if we solve the release blocking 1.2 issues (onStepIn/Out call stack overflows and crashes) and this turns out to cause issues that nobody wants to try to solve.

marksamman:
Thanks!

src/monster.cpp
vankk: @marksamman Could you help us what do you want test? Also, I can test it.
I want see if there is any side effects of theses changes.

PrinterLUA: Can you link me to the issues, so we can speed up the release of 1.2. I would love to take a look at it.

marksamman: #1621, #1660

It's mostly the first one that's blocking the release.

PrinterLUA: Ok Cheers, i will take a look on those and try reproduce.

PR #1764, opened 2016-04-12
Replace const_cast with mutable

ranisalt:
A `const_cast` inside player was not needed since it was possible to have a `mutable` variable instead. This cast was invoking undefined behavior and there's another two `const_cast` further in the file that may cause players [to travel time](https://blogs.msdn.microsoft.com/oldnewthing/20140627-00/?p=633).

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -836,18 +836,17 @@ bool Player::canWalkthrough(const Creature* creature) const
 		return false;
 	}
 
-	Player* thisPlayer = const_cast<Player*>(this);
 	if ((OTSYS_TIME() - lastWalkthroughAttempt) > 2000) {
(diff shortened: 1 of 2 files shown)


gugahoa:
Any reason why this isn't merged yet?

ranisalt:
It's not possible to remove that last cast without a huge refactor. A querying function `canWalkthrough` should never be able to change player state, yet it does.

marksamman:
It hasn't been merged because mutable is also a workaround, albeit safer than const_cast. I'd rather solve this properly or just remove lastWalkthroughAttempt/Position.

ranisalt:
Lots of "querySomething" and "canSomething" functions are not well written, IMHO they should not have side effects, and it would fix this issue for all. I prefer to use exception handling for those cases instead, what is your opinion?

PR #1769, opened 2016-04-14
Fix #1766

### LINKED ISSUE #1766: Teleport use
Use ladder with pz tile is possible to use for skull players

ranisalt:


vankk:
I think that @marksamman will ask for you to splash your commits.

wgrriffel:
@ranisalt If that is really an issue, you should add `player:sendCancelMessage("You can not enter a protection zone after attacking another player.")`

vankk:
Also I don't think that is an issue, due in tibia rl you can use ladders with pz locked :yum:

data/actions/scripts/other/teleport.lua
```diff
@@ -1,5 +1,10 @@
 local upFloorIds = {1386, 3678, 5543}
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
+	local tile = item:getTile()
+	if tile and tile:hasFlag(TILESTATE_PROTECTIONZONE) and player:isPzLocked() then
```
EvulMastah: You should also check destination tile.
There should also be a check on the rope script side to prevent players from roping others that are PZ locked to PZ zone.

PrinterLUA:
I agree with Evual, if we going to adress it. Check for destination position instead.

ranisalt:
> I think that @marksamman will ask for you to splash your commits.

Github now allows commits to be squashed on PR.

> @ranisalt If that is really an issue, you should add player:sendCancelMessage("You can not enter a protection zone after attacking another player.")

Good, I'm not a Lua person myself hehe

data/actions/scripts/other/teleport.lua
ranisalt: > You should also or actually only check destination tile.
> There should also be a check on the rope script side to prevent players from roping others that are PZ locked to PZ zone.
> 
> Anyway, I don't think this should be addressed at all, since this is not the way the map should be designed.

Well, I'm not sure how this works on Tibia. I don't think there are rope spots on protected areas and if there are, both the hole and the spot have a wide protection zone.

wgrriffel:
@EvulMastah agreed and as i said at #1766, maps aren't suppose to be like this.

PrinterLUA:
@ranisalt
You may consider change it up, to something like this:
http://pastebin.com/qS8grZY9

Great work between :+1:

PR #1773, opened 2016-04-20
Add missing DB migration

ranisalt:
Not sure if needed, but on #1603 I changed `18.lua` and forgot to add `19.lua` returning false. It didn't seem to cause any issue, but just in case...


marksamman:
Thanks!

Issue #1745, opened 2016-03-21
Pathfinding


jo3bingham:
http://i.imgur.com/fREMiVv.png

If you left-click on the terra boots you will get a "There is no way." message like you should because the client checks for a path and sees that there isn't one because of the crate. However, if you try to move, use, browse field, or perform any other action on the terra boots/the tile it's on that forces the server to check for a path, your character will walk right over the crate and it should not. It seems the server's pathfinding algorithm doesn't take in to account tiles with items that have the "blocks pathfinding" flag set.

marcelobcortes:
looks like a client bug as server just follow a direction list

jo3bingham:
@marcelobcortes This is definitely not a client bug. Study how the server<->client works.

vankk:
I don't know why we should care about this. We currently has too many others problems to be solved.

marcelobcortes:
ok Ive made a mistake
the listDir in these specifics acts are made by server. this already sounds like a bug for me as it'd be better in client side

so.. just a small fix in getPathMatching

jo3bingham:
That, or, more specifically, Tile::queryAdd(). I've yet to find a solution that doesn't effect something else.

wgrriffel:
@jo3bingham there's anyway to get path finding through lua? if so events/player.lua is an easy way to solve.

jo3bingham:
@wgrriffel this isn't something that should be fixed in Lua anyway.

wgrriffel:
@jo3bingham i just thought in ease, but you're right, i'll start look for some solution also.

marksamman:
This should probably be checked in Map::canWalkTo

wgrriffel:
Just to add, this also happen when moving creatures and players.

marcelobcortes:
try this:

tile.cpp
line 526
http://hastebin.com/lamexefupe.coffee

slavidodo:
Is this related to `Creature::getPathTo` ?

PR #1774, opened 2016-04-21
Pathfinding blockPath check, fix issue #1745

PrinterLUA:
This is how it's currently:
http://imgland.net/?img=6WFLCfD.gif

And after:
http://imgland.net/?img=IccUuMe.gif

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -564,7 +564,7 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
 	//used for non-cached tiles
 	Tile* tile = getTile(pos.x, pos.y, pos.z);
 	if (creature.getTile() != tile) {
-		if (!tile || tile->queryAdd(0, creature, 1, FLAG_PATHFINDING | FLAG_IGNOREFIELDDAMAGE) != RETURNVALUE_NOERROR) {
+		if (!tile || tile->queryAdd(0, creature, 1, FLAG_PATHFINDING | FLAG_IGNOREFIELDDAMAGE) != RETURNVALUE_NOERROR || tile->hasFlag(TILESTATE_BLOCKPATH)) {
 			return nullptr;
 		}
 	}


vankk:
Good job, mate.

marksamman:
This affects monsters and NPC's too (this is what I mean with side effects of your changes, as I said in the other PR). Aren't some monsters able to walk on those items?

PrinterLUA:
But npcs should not walk over boxes? As for monster e.g orc should not be able, while Demon should push the box.

Since i remember on Tibia rook, people blocked orcs with boxes.

wgrriffel:
@PrinterLUA you right, on tibia is just like that, but i remember trying something like this and screw up monsters canpushitems, i suggest you try it before, if it's okay so npc and monsters is even with tibia.

PrinterLUA:
@marksamman 
I have made some tests and it seems that npcs has their own canWalkTo
https://github.com/otland/forgottenserver/blob/master/src/npc.cpp#L391
Here is a gif, to prove it:
http://3.1m.yt/aEyMTmm.gif

While monsters works like i expected:
Here is a Orc, which cannot push items e.g boxes:
http://1.1m.yt/0LEyTEA.gif

Here is a Demon, which can push items e.g boxes:
http://1.1m.yt/i7Fg2-C.gif

wgrriffel:
@PrinterLUA lgtm, but still an issue imo npcs walking through blockpath items.

PrinterLUA:
But that can easily be fixed. Or it's maybe intended that they walk over boxes.

wgrriffel:
@PrinterLUA sure it's easy, if @marksamman like to have it, you could commit in this same PR.

PrinterLUA:
I have no idea, if npcs should be able to walk over boxes or not. Maybe a option? I'll wait for the feedback by @marksamman

HeavenIsLost:
"ignoreheight" option in the npc.xml make the npc able to walk over boxes/chairs.
check: https://github.com/otland/forgottenserver/blob/master/src/npc.cpp#L411

wgrriffel:
I didn't ever noticed that, so right then i get home i'll compile with this PR looking for any unexpected behavior.

jo3bingham:
Platinum coin marks the crate trying to be moved/used. Gold coins mark the tiles where neither are possible ("There is no way.").
How it is currently: http://i.imgur.com/MYUSSGB.jpg
With your PR: http://i.imgur.com/x65l4BV.jpg
On RL-Tibia it's possible from any of those tiles.

wgrriffel:
agreed @jo3bingham i completely forgot about it, and this shouldnt happen.

HeavenIsLost:
Why you did close it? did you ever found the fix?

wgrriffel:
When git set as unknown repository can't make commits anymore, i've tried many ways to github link my branch to PR again, all failed so i think happened same to him.

PrinterLUA:
I will take a look at it later. I going to make other pull requets that is on the roadmap.

PR #1778, opened 2016-04-25
Add `explicit` keyword to single-argument ctors

ranisalt:
Two constructors (for `Vocation` and `WildcardTreeNode`) receive a single `uint16_t` argument and are not marked as `explicit` and could accidentally be (and were) used as conversion constructors ([source](https://www.securecoding.cert.org/confluence/pages/viewpage.action?pageId=3416)) and thus should be marked as `explicit`, else code like `if (someVoc == 2)` would be allowed instead of `if (someVoc.getID() == 2)`.


marksamman:
Thanks!

PR #1781, opened 2016-04-27
Updating Spawns.

slavidodo:
This update was implemented in 10.76 update.
Image

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -999,6 +999,53 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
 
 					s << "protection all " << std::showpos << show << std::noshowpos << '%';
 				}
+				
+				// reflect
+				show = it.abilities->reflectPercent[0];
+				if (show != 0) {
+					for (size_t i = 1; i < COMBAT_COUNT; ++i) {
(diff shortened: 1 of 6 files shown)


WibbenZ:
Sad thing about is that some monsters still can be blocked, 
Found this on otland, might be diffrent now, but just a week ago I could block a monster can't remember what it was tho..
http://1.ii.gl/GzdzEGBbt.png

slavidodo:
Well, this is good point of this spawn system which will work overall monsters.

mattyx14:
:D nice ^^

ranisalt:
Please don't implement multiple things on a single PR, instead do each one on a different branch and PR individually.

slavidodo:
Okay.

slavidodo:
@ranisalt I am not so familer with git, how to revert commit?

ranisalt:
I would use `git reset HEAD~2` to revert 2 commits

WibbenZ:
Just saying that this isen't the best way.
Ex the spawns in the poi tp room are blockable, to make it easier..
This would make it alot harder to do a poi service or a poi boss check.

slavidodo:
I will close the pull request and rework it with monster flag.

PR #1785, opened 2016-05-10
Add underscores in spells file names.

wgrriffel:


marksamman:
Why? Underscores are better because they don't have to be escaped.

wgrriffel:
@marksamman I noticed all "newer" (like spells) files haven't underscores anymore, so i assumed it needed an indentation, can close if you think isn't what i thought

ranisalt:
I think it should be the other way around, all should have underscores. Spaces are a pain to navigate through the shell and other places where spaces need to be escaped.

esfomeado:
Why not camelCase?

marksamman:
@wgrriffel If you would like to revert your changes and update the spells to use underscores instead I'll accept this PR. Otherwise you may close it.

ranisalt:
> Why not camelCase?

Camel case is horrible for file names, let it be for some code only.

wgrriffel:
@marksamman i'll do it soon, should I change for xml files? (monsters and npcs) as well?

marksamman:
Thanks! Yes, you can open a new PR for that though.

PR #1792, opened 2016-05-12
Replace spaces to underscores in monsters.

wgrriffel:


marksamman:
You must also update monsters.xml.

marksamman:
Nevermind. I see you've done that, it just wasn't visible in GitHub's diff view. Thanks!

PR #1794, opened 2016-05-15
Implement equipping item via hotkey feature

gugahoa:
This is a implementation based on Jo3Bingham thread on otland (https://otland.net/threads/tfs-1-1-and-tibia-10-76.232100/)

I used this in a server with about 50 players online, for two months.
This is marked as WIP because I would like suggestions on how to improve the code, as I don't think it's worth to merge it yet. There was an issue when it was being used that made me not post it until today, the issue has to do mostly with a better way to send items information to the client as with the way I'm doing now it iterate over every item on player every time a Player::postAddNotification or Player::postRemoveNotification, and this caused some freezes if I spammed "/i parcel, 100" for example.

Now that I have time to work on it, fire every suggestion you have at me. I'll gladly do my best to make it better. :)


kornholi:
1. Why are you recalculating the hashmap everytime in sendItems? Store it in the player and keep it in sync with postAddNotification/postRemoveNotification

2. Use `unordered_map` instead of `map`

3. Use a stack (two vectors) instead of recursion in parseItems, etc. Something like this:
```
vector containers;
for (item : slots) { 
    ..
    if (container) add to containers
}

vector new_containers;
while (!containers.empty()) {
    for (thing : containers) {
        ..
        if (thing is container) add to new_containers

        containers.swap(new_containers);
    }
    new_containers.clear();
}
````

**Comment by gugahoa** [comment]
Thanks for the guidance, @kornholi 
Is it better now? I was thinking that my code readability is lacking, any suggestion on that front too?

I couldn't use unordered_map, it seems you have to send it ordered for the client, when using unordered_map the client didn't send the packet for equipping item.

**Comment by gugahoa** [comment]
Can someone test one thing on real tibia? Items like life ring, if you select the ring before transformation for the hotkey can you equip and de equip it using the same hotkey or not?
And for arrows and the like, if you have 100, or 50, (with 0 arrows on the backpack) equipped and use the hotkey does it de equip it or does nothing?

@marksamman @kornholi  I'm having trouble identifying edge cases for subtraction of -1 from 0 on cached_items, I already identified one. I think there maybe some that I'll look for in future testing. If possibly, could you explain a little of the inner working of item transformation and its interaction with Player::postAddNotification and Player::postRemoveNotification?

**Comment by wgrriffel** [comment]
@gugahoa 

> Can someone test one thing on real tibia? Items like life ring, if you select the ring before transformation for the hotkey can you equip and de equip it using the same hotkey or not?

no, it need two hotkeys to do that, one to deequip another to equip

> And for arrows and the like, if you have 100, or 50, (with 0 arrows on the backpack) equipped and use the hotkey does it de equip it or does nothing?

it deequip all arrows you have either haven't at last one at backpack.

**Comment by slavidodo** [comment]
Actually for lifering it's a bug, if you want to solve it, just search as well if player has item's transformed.

**Comment by wgrriffel** [comment]
@slavidodo actually cipsoft does think it's not, i saw some time ago at forums it is on purpose.

**Comment by gugahoa** [comment]
Adjusted the behavior to cover what @wgrriffel said about the arrows.
Performance wise for equipping and de equipping the items I think it's already good enough.
I have one backpack with 361968.00 oz and can't feel any kind of freeze when equipping and de equipping items

But I can still freeze the server a little spamming /i parcel, 100

**Comment by marksamman** [comment]
I don't think this is good enough performance wise. What if you go to a NPC and buy 500 of a non stackable item? The protocol for hotkey equip is ridiculous, we should just bypass it by sending a list of all equipable items on login.

File: src/game.cpp [hunk 1]
```diff
@@ -1701,6 +1701,117 @@ ReturnValue Game::internalTeleport(Thing* thing, const Position& newPos, bool pu
...
+	if (Item* item = player->getInventoryItem(CONST_SLOT_BACKPACK)) {
+		if ((backpack = item->getContainer())) {
+			if (backpack->getLastIndex() != backpack->capacity()) {
+				toContainer = backpack;
+			} else {
+				for (ContainerIterator it = backpack->iterator(); it.hasNext(); it.advance()) {
+					if (Container* container = (*it)->getContainer()) {
+						if (container->getLastIndex() != container->capacity()) {
+							toContainer = container;
+							break;
```
ranisalt: god damn it, feels like javascript

gugahoa:
I mean performance-wise for equipping and de equipping alone, for the freezes I still don't know what we can do about. Sending a list of every equipable seems like a really good idea. I thought about it but didn't try because it was not obvious how to make such a list.
I'll try to look into it.

gugahoa:
Just tested the scenario of buying lots of non stackable item from a NPC. It's just like /i parcel, 100. If you spam, sooner or later there will be freezes. :/

HeavenIsLost:
@marksamman i got annother approach. Add to a std::vector<Item*> the items the player got on postAddNotification(adding a ref) and remove the items from that list on postRemoveNotification(removing the ref) then on each onThink send the updated list to the client(creating a std::map from the std::vector<Item*>  and then send it to the client).

What you think?

marksamman:
@HeavenIsLost I think it's useless unless you are going to hold pointers to the items so that they can be quickly located through hash/tree lookups, but that opens up the risk for dangling pointers if it's not done properly. It's not only a performance issue, but also a bandwidth issue to send everything the player is carrying every time something changes in the inventory. The packets also contribute to worse performance for the network thread because we have to XTEA encrypt the packets, which we otherwise wouldn't have to send at all.

diath:
I'm using this snippet to generate a list of equippable inventory and send it only on login, it's greatly reduced (i.e. doesn't contain items that are unlikely to be equipped via a hotkey like say a piece of meat) but covers most items that people would want to equip via hotkeys.

```
void Items::buildInventoryList()
{
	inventory.reserve(30000);
	for(const auto &type: items)
	{
		if(type.weaponType != WEAPON_NONE || type.ammoType != AMMO_NONE ||
			type.attack != 0 || type.defense != 0 ||
			type.extraDefense != 0 || type.armor != 0 ||
			(type.slotPosition & SLOTP_NECKLACE) == SLOTP_NECKLACE ||
			(type.slotPosition & SLOTP_RING) == SLOTP_RING)
		{
			inventory.push_back(type.clientId);
		}
	}
	inventory.shrink_to_fit();
	std::sort(inventory.begin(), inventory.end());
}
```

You could go with something similar in the PR should you switch to this approach.

slavidodo:
I was exactly going to say that, It's actually very better.
And this will fix as well the problem with rings.

gugahoa:
Thanks @diath for the function.  (Sorry for misspelling your username on the commit) 
Now there's no freeze if I spam /i parcel, 100. There's only freezes in the same situation as there would be on TFS master branch.

@marksamman what other aspects could I improve the code?

src/protocolgame.cpp
```diff
@@ -2533,6 +2542,28 @@ void ProtocolGame::sendInventoryItem(slots_t slot, const Item* item)
 	writeToOutputBuffer(msg);
 }
 
+void ProtocolGame::sendItems() {
+	NetworkMessage msg;
+	msg.addByte(0xF5);
+
+	std::vector<uint16_t> inventory = Item::items.getInventory();
```
HeavenIsLost: Why you are not using & here? may be better performance wise

marksamman:
playerEquipItem could use a lot of improvements, it's too messy. I think @ninjalulz had a prototype of this feature where he handled it cleanly.

gugahoa:
@marksamman I reduced a lot of playerEquipItem, now I was thinking about the part to determine the slot. Should it be a method of Item?

I don't know if it should because slotPosition can hold more than one SLOTP_*

ninjalulz:
@gugahoa You can take a look at the prototype here: https://gist.github.com/ninjalulz/ec6a8ef25072482ea688.

gugahoa:
Didn't know it wasn't need to remove the item from the slot before moving a new item, cool!
Thanks @ninjalulz !

I tried a lot of variations that I could think of to get the internalMoveItem to be just one call or two, but it seems to replicate Tibia's feature just the way it is, the minimum necessary is what I made... :(

I thought about using player and CONST_SLOT_WHEREVER so that I could delete the snippet to find toContainer, but CONST_SLOT_WHEREVER sometime sent the item to the left hand slot, and the feature wasn't supposed to do that. Thought about using CONST_SLOT_BACKPACK, instead of CONST_SLOT_WHEREVER but it seems it will only see if it's possible to put the item in the first backpack, instead of looking for the first free spot on any container inside the backpack.

I'll go ahead and remove the WIP from the title, as I feel like it's only one or so commit from being good enough.

src/game.cpp
```diff
@@ -1701,6 +1701,84 @@ ReturnValue Game::internalTeleport(Thing* thing, const Position& newPos, bool pu
...
+			return;
+		}
+	}
+
+	if (item && toContainer) {
+		if (item->getID() == it.id) {
+			internalMoveItem(player, toContainer, toContainer->getLastIndex(), item, item->getItemCount(), nullptr);
+		} else if (Item* equipItem = searchForItem(backpack, it.id)) {
+			internalMoveItem(player, toContainer, toContainer->getLastIndex(), item, item->getItemCount(), nullptr);
+			internalMoveItem(equipItem->getParent(), player, slot, equipItem, equipItem->getItemCount(), nullptr);
```
gugahoa: Here I first send the item to the first available spot because if I only send the equipItem to the slot, and the parent of equipItem does not have a free spot, it'll return not enough room.

Healek:
@gugahoa this system is stable? you recommend or is it still a prototype. Hugs!!

gugahoa:
As far as I tested, it's stable.

darkjav:
Anybody know if this work fine with the @djarek cast system?
@gugahoa You test it with a macro that changed rings very fast?, because if you do it, make that server get crash!

wgrriffel:
@gugahoa what @darkjav said is a good point, i dont tested with bot because i dont trust on them nowadays so i cant test, what i can say is on tibia any event (such as walking and item moving) player receive a global "exhaustion" which i would say it's about 250~500 ms, i don't know exactly im saying based on what i played between 10.70 and 10.80, currently i bet its the same, saying just in case =)

gugahoa:
@wgrriffel, just tested on rl tibia, there's no exhaust between hotkey equip action
I didn't quite understand @darkjav, the server crashed when you equipped and de equipped with a macro?

wgrriffel:
@gugahoa how you did it? i make some test and i can notice a big difference between your code and tibia one, ofc i am at 100~120ms with tibia servers, but im sure the diff between is not only 120ms, my tests were made just pressing down hotkey with wedding ring, try it.

gugahoa:
@wgrriffel if you press a button on CipSoft's client, it won't spam. Just press it multiple times.

wgrriffel:
@gugahoa it spam like for 2 sec after i receive a kind of exhaustion, i can press down the hotkey only for some seconds after server take 500~1000ms to send back the action.

gugahoa:
@wgrriffel I just tested again to make sure, and spent sometime spamming click + hotkey button. It doesn't have a exhaust, I tested on Flash Client equipping and deequipping.

wgrriffel:
@gugahoa sorry, maybe my connection was not good with tibia servers, tested right now there is no meaningful delay to be added, about another global exhaustion that i mentioned is nothing to do with this pr, try yourself to run and spam any action you will see, but it has nothing to do with your pr, delay that i was talking about probably i had some bad connection time, can't reproduce again, nevermind.

Healek:
@gugahoa 
Gustavo, its possible to add one delay on equip with hotkeys? 1-2sec maybe. Thank you!

joseluis2g:
@gugahoa have you solved the transformable items like lifering or energyring?

gugahoa:
@joseluis2g No, it works like global.

joseluis2g:
@gugahoa didnt knew it, thanks bro

gugahoa:
I received some private messages saying this has been in use for quite some time at various servers. Have anyone found anything worth reporting?

It would be nice to have a little feedback, as it can improve the feature for everyone.

gugahoa:
@marksamman any comment about this pr?

src/game.cpp
```diff
@@ -1701,6 +1701,84 @@ ReturnValue Game::internalTeleport(Thing* thing, const Position& newPos, bool pu
 }
 
 //Implementation of player invoked events
+void Game::playerEquipItem(uint32_t playerId, uint16_t spriteId)
```
marksamman: As mentioned in a comment, I'd rather see a simpler implementation for this, like in https://gist.github.com/ninjalulz/ec6a8ef25072482ea688. Just find an item with the client id and move it to the appropriate slot using internalMoveItem.

src/protocolgame.cpp
```diff
@@ -916,6 +917,14 @@ void ProtocolGame::parseFollow(NetworkMessage& msg)
 	addGameTask(&Game::playerFollowCreature, player->getID(), creatureId);
 }
 
+void ProtocolGame::parseEquipObject(NetworkMessage& msg)
+{
+	uint16_t spriteId = msg.get<uint16_t>();
+	msg.get<uint8_t>();
```
marksamman: You can comment this line since it doesn't really do anything other than move the read position, which makes no difference here, but it's good to know there's one more byte being sent in this packet.

src/game.cpp
gugahoa: I think this is close to the simplest it can be, it's a little big because of sanity checks
They guarantee:
- Only move item to backpack if there's space (Can see if I can remove this part if internalMoveItem already does it)
- Only de-equip the item equipped in the proper place (If the sword is in the ammo slot, it won't try to de equip it for example)
- If there's nothing in the slot or if there is and if equipped item is stackable, stack what we have in the backpack with what we have equipped, instead of de-equipping (and return)
- If slot item is equal to equip item, de-equip it, else find item in the backpack, de-equip what's in the slot and equip what's in the backpack

There's two things I can look into this week to see if it's possible to make simpler: Remove backpack capacity sanity check, remove de-equip call before equip. Both in case internalMoveItem already does it.

Sorry if I'm missing something here and it can indeed be made much simpler

gugahoa:
Sorry for taking too long to refactor, was in vacation. Created a new PR with a clean work tree this time. #2053

PR #1802, opened 2016-05-27
Fix #1690

### LINKED ISSUE #1690: Monsters cast beams and waves through wall
When monster is in front to wall (anything, any BLOCKSOLID wall, included magic wall) and he is blocked by player..
Monster will cast beam and wave through wall!
http://i.imgur.com/63nsVoh.png

kamil-karkus:


kamil-karkus:
Real tibia allows to cast spells infront of walls(blockprojecttiles) even, 
so it's kind of workaround until cast spells infront of wall will be implemented.

PR #1806, opened 2016-05-29
Fix #1805

### LINKED ISSUE #1805: Level Door
Here the example

bez nazwy

wgrriffel:
More than one, I found about 5~7 doors with same conflict, also just to convenience i changed file names.


marksamman:
Thanks!

PR #1810, opened 2016-06-02
onStep callbacks accept return value to stop action

### LINKED ISSUE #1621: [Error - MoveEvent::executeStep] Call stack overflor

gugahoa:
Should fix #1621 
Only side effect from this PR: creature:getPosition() will return the position before updating. Example: If I step in a tile at 100, 100, 7 from 99, 100, 7 creature:getPosition() will return 99, 100, 7
position and fromPosition is still the same as 'master' branch


HeavenIsLost:
I think there is a great mistake here. Calling the event on Tile::postAddNotification make sure that each time a creature is added by any mean to the tile it will call the event. With your method it seems to work for all ways of "Creature Move".

gugahoa:
That's true. Thanks for the heads up, I'll work on it.

gugahoa:
@HeavenIsLost Anything else I may be missing?
I put the call for onStepIn on queryAdd, as for onStepOut I put before every Map::moveCreature and on Game::removeCreature
The only place I didn't was for Bed::sleep, since it doesn't make much sense to call onStepIn for sleep action, and onStepOut is already called when the player is kicked

gugahoa:
@marksamman @kornholi any comments? I'm willing to maintain this PR up to date, and it should fix #1621 

There's a side effect already reported on the description, if it's blocking the PR I would gladly look into other solutions to the problem.

marksamman:
This will have to wait until after the 1.2 release, #1621 will be fixed in 1.2 by reverting the code that caused it, after that we'll look for a better solution (possibly this PR).

gugahoa:
@marksamman should I rebase against master now? Since it's the start of 1.3

marksamman:
@gugahoa Go ahead. Do scripts that teleport to fromPosition and return false still work fine after these changes? The teleportTo call should not do anything since fromPosition should be the creatures current position.

gugahoa:
Yes, last time I tested they were working. I'll test again this sunday if something unexpected was merged and didn't conflict.

gugahoa:
I'll close this PR and open a new one with a clean work tree

PR #1812, opened 2016-06-02
OnMoveItem accept cylinder, close #1446

### LINKED ISSUE #1446: How to get container location or owner on Player:onMoveItem() ?

PrinterLUA:


marksamman:
Please add fromCylinder too, and put it before toCylinder in the parameter list.

PrinterLUA:
done

marksamman:
Thanks!

PR #1813, opened 2016-06-02
Move raid system to globalevents

ranisalt:
This effectively nukes the raid system in favor of global events, which permit better management for raids. It's easier to define raids on interval, on exact times, force execution, force scheduling, add monsters on different delays and randomized amounts and positions... and also gets rid of a command.

~I added a lib with helper functions to add monsters to a raid. The event is called `onRaid` and there is an example bundled pretty equivalent to the previous XML raid example.~ keep reading


PrinterLUA:
Great job, but is it even necessary to have a seprate event for raid? We could just use onThink or onTime. If the wish and just create a raid library in lua.

It should be possible?

ranisalt:
`onRaid` falls between `onTime` and `onInterval` here, it works as both. Yes it is possible to completely ditch `onRaid` and use just those two.

ranisalt:
True that, I will fix it to use the Position constructor later.

I will also patch the addSpawn function to allow another parameter, but I see it as removing clutter from the raid (explicit) to move to a lib function (implicit).

EvilHero90:
actually this entire thing should be handled completly in lua by coroutines and a simple addEvent timer at startup, could get rid of all c code then

ranisalt:
@EvilHero90 I agree with handling completely on Lua, but I couldn't figure how to use coroutines. Could you give me sample code for how you see raids?

EvilHero90:
I can do that on monday, I'm busy with work the entire weekend

data/globalevents/scripts/testraid.lua
```diff
@@ -0,0 +1,27 @@
...
+	-- You can generate a random amount of creatures
+	local amount = math.random(4, 10)
+	for _ = 1, amount do
+		local x, y = math.random(89, 99), math.random(122, 130)
+		addEvent(Game.createMonster, 30000, "Rat", Position(x, y, 7))
+	end
+
+	-- Single spawns
+	addEvent(Game.createMonster, 15000, Position(93, 123, 7))
+	addEvent(Game.createMonster, 30000, Position(98, 125, 7))
```
conde2: Missing monster name ?

ranisalt: Yes, my eyes failed me. Fixed

ranisalt:
This PR is (almost) on it's final form. I added a generic interface to search and force execute global events. I couldn't find an use to force execute startup, shutdown or record events, so they are ignored.

Also, it means it's possible to force trigger a server save on the Lua side.

wgrriffel:
@ranisalt since seems he doesn't got time yet i made an example with coroutines.
```lua
addEvent(checkRaids, 15000)
local inProgress = false
function ratPlague()
	local pos = Position(146, 388, 7)
 	Game.createMonster("Rat", pos)
	broadcastMessage("Rat plague in city!", MESSAGE_STATUS_WARNING)
 	coroutine.yield()
 	inProgress = false
end

function checkRaids()
	print(coroutine.status(start))
	if not inProgress then
		if math.random(1, 1) == 1 then
			inProgress = true
			coroutine.resume(start)
		end
	end
	addEvent(checkRaids, 15000)
end

start = coroutine.create(ratPlague)
```
to make this a real raid system with coroutines would needed a lib as big as npcs is, i personally don't know performace impact with this, imo coroutines are a complicated way to call functions.

ranisalt:
If that is, I don't know how much better this is than `addEvent`.

wgrriffel:
i don't see the point also, if is needed to call a addEvent it becomes a complicated way to call functions, tho im curious to see how he would do that.

ranisalt:
Rebased against master.

ranisalt:
I still intend to keep it up to date and merge when possible. This alone shaves 20 seconds off the build time 😄

WibbenZ:
Only thing I can think you should "add" is a converter XML -> Lua
Maybe this should wait for the next stable release, but looks good to me :+1:

ranisalt:
@WibbenZ working on it! Check the new stuff under tools/ 😄

src/luascript.cpp
```diff
@@ -4248,25 +4250,18 @@ int LuaScriptInterface::luaGameCreateTile(lua_State* L)
 	return 1;
 }
 
-int LuaScriptInterface::luaGameStartRaid(lua_State* L)
```
djarek: API breaking change. Should we provide a compat function?

ranisalt: Maybe. There is `Game.startEvent` now, it could be an alias and expect people won't try `Game.startRaid("Server Save")`.

DSpeichert:
Rebased this into https://github.com/otland/forgottenserver/tree/bye-bye-raids because I couldn't force push to fork repo.

ranisalt:
Ok, gonna rebase it and fix compilation issues.

ranisalt:
@djarek I added backwards compatibility for `Game.startRaid`.

data/globalevents/scripts/testraid.lua
```diff
@@ -0,0 +1,27 @@
+function onTime(interval)
+	-- Message broadcast
+	addEvent(Game.broadcastMessage, 1000, "Rats are attacking neart Trekolt Temple!", MESSAGE_STATUS_WARNING)
```
djarek: neart -> near

src/position.h
```diff
@@ -39,6 +39,7 @@ enum Direction : uint8_t {
 struct Position
 {
 	constexpr Position() = default;
+	constexpr Position(Position&&) = default;
```
djarek: I don't this declaration is necessary.

ranisalt: It is not, I forget to push a commit :P

ranisalt:
Still didn't finish conversion tool, few stuff missing.

WibbenZ:
@ranisalt Would be good if that tool is not related to the TFS source code, so people who use the appveyor artifacts can still use the tool.
Aka come pre-compiled or as a migration.

ranisalt:
@WibbenZ I am not sure how to do that. Is it possible to add a second artifact to appveyor?

soul4soul:
Would it make sense to move the conversion tool into it's own repo under the otland organization?

WibbenZ:
Well we don't need to use appveyor but maybe compile it, zip it and add a link in the readme or something like that. But if we want to we should do what @soul4soul said.

The point was so people who don't want to / can't compile can still use this tool.

ranisalt:
Since the tool uses much of TFS code, how would you do that?

WibbenZ:
Good question, nvm it.
Someone will most likely compile and upload for those who can't or they can convert them manually.

tools/raid_xmltolua.cpp
```diff
@@ -0,0 +1,479 @@
...
+	int16_t z;
+};
+
+template<class T>
+T require_attr(const pugi::xml_node& node, const std::string& tag)
+{
+	if (pugi::xml_attribute attr = node.attribute(tag.c_str())) {
+		return pugi::cast<T>(attr.value());
+	}
+	throw std::runtime_error(tag + " tag missing.");
```
djarek: Minor issue: I'd introduce a separate type of exception (`MissingTagError`) that inherits from a `std::runtime_error` to make it clear at catch point that other `std::runtime_errors` are unexpected and fatal.

tools/raid_xmltolua.cpp
```diff
@@ -0,0 +1,479 @@
...
+	}
+
+	Position from_, to;
+	uint32_t radius;
+	std::vector<Spawn> spawnList;
+};
+
+struct ScriptEvent : public RaidEvent
+{
+	ScriptEvent(uint32_t delay) : RaidEvent{delay} {}
```
djarek: Not a big issue, but I kinda dislike storing quantities that have "units" (in this case, time) in uint32_t.

tools/raid_xmltolua.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+	std::vector<std::unique_ptr<RaidEvent>> eventList;
+	RaidInfo info;
+};
+
+std::unique_ptr<RaidEvent> parseEvent(const std::string& eventType, const pugi::xml_node& eventNode)
+{
+	pugi::xml_attribute delayAttr = eventNode.attribute("delay");
+	if (not delayAttr) {
+		throw std::runtime_error("delay tag missing.");
```
djarek: Introduce new exception type: `DelayTagMissing`

tools/raid_xmltolua.cpp
```diff
@@ -0,0 +1,479 @@
...
+	} else if (eventType == "areaspawn") {
+		uint16_t radius = 0;
+		if (auto radiusAttr = eventNode.attribute("radius")) {
+			radius = pugi::cast<uint16_t>(radiusAttr.value());
+		}
+		return std::unique_ptr<RaidEvent>{AreaSpawnEvent::from(delay, eventNode, radius)};
+	} else if (eventType == "script") {
+		return std::unique_ptr<RaidEvent>{new ScriptEvent{delay}};
+	}
+	throw std::runtime_error("invalid event type " + eventType);
```
djarek: new exception type: `InvalidEventType`

tools/raid_xmltolua.cpp
```diff
@@ -0,0 +1,479 @@
...
+		output << raid.to_script() << std::endl;
+	}
+}
+
+int main(int argc, char** argv)
+{
+	fs::path globalevents_path, raids_path;
+
+	int opt;
+	while ((opt = getopt(argc, argv, "hg:r:")) != -1) {
```
djarek: Since we're already depending on boost, might as well use `boost::program_options` ?

tools/raid_xmltolua.cpp
```diff
@@ -0,0 +1,479 @@
...
+			try {
+				return messageTypes.at(type.as_string());
+			} catch (const std::out_of_range&) {
+				std::cout << "[Notice] Raid: Unknown type tag for announce"
+							 "event. Using default: MESSAGE_TYPE_ADVANCE\n";
+				return defaultMessageType;
+			}
+		}(eventNode.attribute("type"));
+
+		return new AnnounceEvent{delay, message.as_string(), messageType};
```
djarek: Raw, owning pointers and exceptions don't mix well...

ranisalt: Just noticed I return those pointers and encapsulate in unique_ptrs. Might as well return unique_ptrs instead.

tools/raid_xmltolua.cpp
ranisalt: What do you suggest?

src/globalevent.cpp
```diff
@@ -263,21 +275,10 @@ bool GlobalEvent::configureEvent(const pugi::xml_node& node)
...
-		} else if (strcasecmp(value, "record") == 0) {
-			eventType = GLOBALEVENT_RECORD;
-		} else {
-			std::cout << "[Error - GlobalEvent::configureEvent] No valid type \"" << attr.as_string() << "\" for globalevent with name " << name << std::endl;
-			return false;
-		}
 	} else if ((attr = node.attribute("interval"))) {
 		interval = std::max<int32_t>(SCHEDULER_MINTICKS, pugi::cast<int32_t>(attr.value()));
 		nextExecution = OTSYS_TIME() + interval;
+		eventType = GLOBALEVENT_TIMER;
```
ninjalulz: Are you sure about this change? 😄

tools/raid_xmltolua.cpp
djarek: `uint32_t` => `std::chrono::milliseconds` (or w/e unit it uses ATM).

slawkens:
So you basically removed raid system. But how to write a raid being a beginner? :P Your provided Lua example is complicated for newbie IMO. The old raid system was way easier.

You mentioned something about helper lib, but I can't seem to find it in this PR?

ranisalt:
> Your provided Lua example is complicated for newbie IMO

Why? It is simple Lua code that a lot of actions, events and movements already do (e.g. annihilation quest summons monsters, server save broadcasts message). I thought this was enough as example.

I didn't implement the helper lib since all functions already existed in TFS.

slawkens:
In old raid system, you needed to know some basic XML syntax, and copy-paste to add more monsters/spawns. Here you need to know some basic programming functions if you want to modify it. It is good as an example, but to make something more advanced from this, as a beginner it wouldn't be easy.

Especially I mean that "for" loops for creating more spawns. Even for me it took some time to understand it.

I would suggest either make a parser that parses old raids, so you can still use them, or make some helper library in Lua, with some functions to simplicify the whole process of creating a raid.

Like instead of this in old raid system:
```
<areaspawn delay="30000" fromx="89" fromy="122" fromz="7" tox="99" toy="130" toz="7">	
	<monster name="Rat" minamount="4" maxamount="10" /><!-- Random amount in specified range -->	
</areaspawn>
```
you could write some helper function like this in Lua:
```
Raids.spawn(30000, {from = {x = 89, y = 122,  z = 7}, to = {x = 99, y = 130, z = 7}}, 
	{
		{"Rat", 1, 4}, -- name, minamount, maxamount
		{"Cave Rat", 1, 4}, -- name, minamount, maxamount
	}
)
```

So you can easily add more monsters to this by copying lines.

ranisalt:
There's a parser that converts old raids to new Lua code. I didn't wish to introduce new functions since I didn't want to repeat code. I could add a more extensive documentation though.

slawkens:
Ok then except that it looks good. Didn't saw that converter, cause it was hidden by default.

Good job!

DSpeichert:
@ranisalt it seems we need a rebase again.

joseluis2g:
Can someone compile the tool for me. Great pr btw.

kito2:
Tried to compile on Ubuntu, an some errors: https://github.com/ranisalt/forgottenserver/issues/6

ranisalt:
@DSpeichert will do sir

gpedro:
raids should be killed in TFS 2.0. PLEASE!

EPuncker:
@ranisalt can we have a new rebase please, I'm gonna test it

ranisalt:
@EPuncker currently I don't have access to a computer, I will in a couple days.

EPuncker:
so far so good, the only thing I miss is a chance setting or I suppose raids will be executing every now and then (or have I missed something? 😄 )

joseluis2g:
thats something you can do with the Lua itself thats why this PR is awesome, a chance for your raid event its just a random check at the beginning of the event

ranisalt:
I rebased it but what in the royal fuckery is happening to the builds, reporting error on success to find Boost?

ranisalt:
@EPuncker

src/globalevent.cpp
ranisalt: Pretty much! Otherwise, it is registered as `onThink` which makes no sense for interval events...

(somewhat late, this response...)

ranisalt:
While this is backwards incompatible, it does not add much new functionality - only the converter -, it only removes functionality as the support already exists.

data/scripts/globalevents/#testraid.lua
```diff
@@ -0,0 +1,53 @@
+local raid = GlobalEvent("Testraid")
+raid:type("timer")
+raid:interval(1800)
+
+local function event0()
+    Game.broadcastMessage("Rats are attacking neart Trekolt Temple!", MESSAGE_STATUS_WARNING)
```
EPuncker, line 6:
```suggestion
    Game.broadcastMessage("Rats are attacking near Trekolt Temple!", MESSAGE_STATUS_WARNING)
```

nekiro (approved):
This is super nice, gj

ranisalt:
> we got some conflicts though

oh yeah it's that time of the year that I rebase it again

nekiro (approved):

EPuncker:
here it still shows a conflict

ranisalt:
@EPuncker not anymore

EPuncker (dismissed):

DSpeichert:
This breaks backwards compatibility so we should hold off until 2.0.

EPuncker:
what if we kept both modes and removed xml on 2.0 release?

DSpeichert:
> what if we kept both modes and removed xml on 2.0 release?

If there's some extra functionality that this PR provides then yes, it makes sense. But if we end up having two ways to do the same thing, it probably isn't worth it.

ranisalt:
Ideas on how to get at least parts of this merged?

DSpeichert:
I am all for 1 and 2.

I think removing support for XML unnecessary as there's huge value in the ability to pick up any of thousands of data packs from the forum and run it, ideally with minimal changes. Early success for users will motivate them to learn more, I'd rather not have us make that more difficult by requiring extra tools to convert.

Now, on the actual tools - can we integrate it into the main binary as a "subcommand" and just ship one binary? I think that would be more powerful as it would never get "lost" in distributions and I can't imagine many drawbacks to it.

omarcopires:
I like these changes and the new system, but I have a question about tokens.
The question is, is there already a way to add silver/gold tokens to monsters that are spawned by raids?

EPuncker:
> I like these changes and the new system, but I have a question about tokens. The question is, is there already a way to add silver/gold tokens to monsters that are spawned by raids?

it would indeed be cool if monsters created by raid had a way of identification, like `isRaid()`, so we could rum additional events on them

@ranisalt I agree with 1 and 2 too

ranisalt:
Well, raids in global events are fully scripted, you can do whatever you want with the monsters you spawn.

@andremiles1 can you take care of 1.? Just change your PR to add the rad script to scripts/globalevents instead, like in this PR.

EPuncker (approved):

ranisalt:
The time for the yearly rebase cometh.

ranisalt:
I think you'll like it more now. There are no external tools, and the new globalevent-based system is drop in.

omarcopires:
TFS 2.0 is so far away, why not now in 1.6? 🥹

ranisalt:
> TFS 2.0 is so far away, why not now in 1.6? 🥹

It's even backwards-compatible :sweat_smile:

EPuncker (approved):

src/globalevent.cpp
MillhioreBT: If interval can no longer be onThink then how do we make an onThink event from xml?

src/luascript.cpp
```diff
@@ -4952,25 +4952,18 @@ int LuaScriptInterface::luaGameCreateMonsterType(lua_State* L)
...
-		lua_pushnumber(L, RETURNVALUE_NOSUCHRAIDEXISTS);
-		return 1;
-	}
+	// Game.startEvent(event)
+	const std::string& eventName = getString(L, 1);
 
-	if (g_game.raids.getRunning()) {
-		lua_pushnumber(L, RETURNVALUE_ANOTHERRAIDISALREADYEXECUTING);
-		return 1;
+	const auto& eventMap = g_globalEvents->getEventMap(GLOBALEVENT_TIMER);
```
MillhioreBT, line 4960: maybe use contains and also pushBoolean:
```cpp
const auto& eventMap = g_globalEvents->getEventMap(GLOBALEVENT_TIMER);
    if (eventMap.contains(eventName)) {
		const auto& event = eventMap.at(eventName);
		pushBoolean(L, event.executeEvent());
	} else {
		lua_pushnil(L);
	}
```

MillhioreBT (changes requested):
You also need to add `self:type("timer")` in the index metamethod of globalevent for the callback
[HERE](https://github.com/otland/forgottenserver/blob/edf61f99a57fb66be4025be76c47cd7b7309f270/data/lib/compat/compat.lua#L239)

[Other Question](https://github.com/otland/forgottenserver/pull/1813#discussion_r1174465390)

data/talkactions/scripts/force_event.lua
MillhioreBT, line 1: `Game.startEvent` returns boolean not RETURNVALUE

MillhioreBT (changes requested):

src/luascript.cpp
```diff
@@ -4952,25 +4952,17 @@ int LuaScriptInterface::luaGameCreateMonsterType(lua_State* L)
...
-		lua_pushnumber(L, RETURNVALUE_NOSUCHRAIDEXISTS);
-		return 1;
-	}
-
-	if (g_game.raids.getRunning()) {
-		lua_pushnumber(L, RETURNVALUE_ANOTHERRAIDISALREADYEXECUTING);
-		return 1;
+	const auto& eventMap = g_globalEvents->getEventMap(GLOBALEVENT_TIMER);
+	if (auto it = eventMap.find(eventName); it != eventMap.end()) {
+		pushBoolean(L, it->execute());
```
EPuncker, line 4962: github action check failure

EPuncker (dismissed):

EPuncker:
let's go?!

omarcopires:
> let's go?!

Yes please! <3

ranisalt:
@MillhioreBT your change request is blocking this PR

EPuncker (approved):

ramon-bernardo:
Ready to merge, right?

omarcopires:
> Ready to merge, right?

This should have been added years ago, we really need this! 🥹

EPuncker (dismissed):

MillhioreBT (dismissed):

EPuncker (approved):

ranisalt:
I can't believe what my eyes are looking at

ramon-bernardo:
2023/12/12, a milestone!

PR #1815, opened 2016-06-02
Minor optimizations

PrinterLUA:


data/lib/core/game.lua
```diff
@@ -20,39 +20,26 @@ function Game.convertIpToString(ip)
...
-	return SKILL_FIST
+	local config = {
+		[WEAPON_CLUB] = SKILL_CLUB
+		[WEAPON_SWORD] = SKILL_SWORD
+		[WEAPON_AXE] = SKILL_AXE
+		[WEAPON_DISTANCE] = SKILL_DISTANCE
+		[WEAPON_SHIELD] = SKILL_SHIELD
+	}
+
+	return config[weaponType] or SKILL_FIST
```
marksamman: Have you benchmarked this? I think the cost of constructing a table is more than some if-statements.

src/creature.cpp
```diff
@@ -121,10 +121,7 @@ bool Creature::canSee(const Position& pos) const
 
 bool Creature::canSeeCreature(const Creature* creature) const
 {
-	if (!canSeeInvisibility() && creature->isInvisible()) {
-		return false;
-	}
-	return true;
+	return canSeeInvisibility() && creature->isInvisible();
```
marksamman: You introduced a bug here.

src/protocolgame.cpp
```diff
@@ -2400,16 +2400,10 @@ void ProtocolGame::sendAddCreature(const Creature* creature, const Position& pos
...
-	sendInventoryItem(CONST_SLOT_NECKLACE, player->getInventoryItem(CONST_SLOT_NECKLACE));
-	sendInventoryItem(CONST_SLOT_BACKPACK, player->getInventoryItem(CONST_SLOT_BACKPACK));
-	sendInventoryItem(CONST_SLOT_ARMOR, player->getInventoryItem(CONST_SLOT_ARMOR));
-	sendInventoryItem(CONST_SLOT_RIGHT, player->getInventoryItem(CONST_SLOT_RIGHT));
-	sendInventoryItem(CONST_SLOT_LEFT, player->getInventoryItem(CONST_SLOT_LEFT));
-	sendInventoryItem(CONST_SLOT_LEGS, player->getInventoryItem(CONST_SLOT_LEGS));
-	sendInventoryItem(CONST_SLOT_FEET, player->getInventoryItem(CONST_SLOT_FEET));
-	sendInventoryItem(CONST_SLOT_RING, player->getInventoryItem(CONST_SLOT_RING));
-	sendInventoryItem(CONST_SLOT_AMMO, player->getInventoryItem(CONST_SLOT_AMMO));
+	for (uint32_t i = CONST_SLOT_FIRST; i <= CONST_SLOT_LAST; i++) {
```
marksamman: This will make it slower if it's not unrolled, so it's not an optimization, but I'm not opposed to the change.

data/lib/core/game.lua
marksamman: I'm guessing that's a no since the functions cause errors with your changes.

PrinterLUA: Well it's true that if statments are faster, than construct a table everytime it gets executed. Should i move it outside?

Also i consider to change the PR title, to clean up. No idea why i called it optimizations.

src/creature.cpp
PrinterLUA: Ok, i will revert it.

src/protocolgame.cpp
ranisalt: The default optimization when not in debug build is `-O2`, which unrolls short, safe to unroll loops, so my bet is it will be unrolled.

data/lib/core/game.lua
```diff
@@ -20,39 +20,26 @@ function Game.convertIpToString(ip)
...
-	end
-	return NORTH
+	local config = {
+		[DIRECTION_NORTH] = DIRECTION_SOUTH,
+		[DIRECTION_SOUTH] = DIRECTION_NORTH,
+		[DIRECTION_EAST] = DIRECTION_WEST,
+		[DIRECTION_WEST] = DIRECTION_EAST
+	}
+
+	return config[direction] or DIRECTION_NORTH
```
ranisalt: You can omit `DIRECTION_SOUTH` from the table too since it's default.

src/player.cpp
```diff
@@ -734,11 +734,7 @@ bool Player::canOpenCorpse(uint32_t ownerId) const
 
 uint16_t Player::getLookCorpse() const
 {
-	if (sex == PLAYERSEX_FEMALE) {
-		return ITEM_FEMALE_CORPSE;
-	} else {
-		return ITEM_MALE_CORPSE;
-	}
+	return sex == PLAYERSEX_FEMALE ? ITEM_FEMALE_CORPSE : ITEM_MALE_CORPSE;
```
ranisalt: Personally, I don't like the ternary operator. A switch case block would be prettier code here and already be more efficient since it (most certainly) will become a jump table (exact identical assembly to ternary operator)

src/player.cpp
```diff
@@ -790,10 +786,7 @@ bool Player::getStorageValue(const uint32_t key, int32_t& value) const
 
 bool Player::canSee(const Position& pos) const
 {
-	if (!client) {
-		return false;
-	}
-	return client->canSee(pos);
+	return client && client->canSee(pos);
```
ranisalt: Suggestion: `client != nullptr` for explicit meaning

PrinterLUA: Alright, thanks

data/lib/core/game.lua
PrinterLUA: Ok

data/lib/core/game.lua
```diff
@@ -20,39 +20,26 @@ function Game.convertIpToString(ip)
...
-	end
-	return NORTH
+	local config = {
+		[DIRECTION_NORTH] = DIRECTION_SOUTH,
+		[DIRECTION_SOUTH] = DIRECTION_NORTH,
+		[DIRECTION_EAST] = DIRECTION_WEST,
+		[DIRECTION_WEST] = DIRECTION_EAST
+	}
+
+	return config[direction] or DIRECTION_SOUTH
```
ranisalt: You completely ignored diagonal directions and changed the default from north to south.

ranisalt:
Closing by the lack of feedback from the author.

PR #1816, opened 2016-06-03
Create party in lua, close #1365

### LINKED ISSUE #1365: Create party by command TFS 1.1

PrinterLUA:
`
Party(userData)
`

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2509,7 +2509,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("MonsterType", "getChangeTargetSpeed", LuaScriptInterface::luaMonsterTypeGetChangeTargetSpeed);
 
 	// Party
-	registerClass("Party", "", nullptr);
+	registerClass("Party", "", LuaScriptInterface::luaPartyCreate);
 	registerMetaMethod("Party", "__eq", LuaScriptInterface::luaUserdataCompare);
 
 	registerMethod("Party", "disband", LuaScriptInterface::luaPartyDisband);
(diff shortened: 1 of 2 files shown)


src/luascript.cpp
```diff
@@ -11849,6 +11849,28 @@ int LuaScriptInterface::luaMonsterTypeGetChangeTargetSpeed(lua_State* L)
 }
 
 // Party
+int32_t LuaScriptInterface::luaPartyCreate(lua_State* L)
+{
+	// Party(userdata)
+	Player* player = getUserdata<Player>(L, 2);
```
HeavenIsLost: I think it should be 1 istead of 2 here.

PrinterLUA: Nope this is the correct way, Party is the first and player is second.

Znote:
Please update this:
https://github.com/otland/forgottenserver/wiki/Metatable:Player

PR #1819, opened 2016-06-04
Fix #1753

### LINKED ISSUE #1753: Decay Bug when put decaying item inside Depot Chest and logout.
Hi,

1 - First Decay Bug:
Edit an item with duration less than 5 minutes, decayto attribute and try to reproduce:
Put this item inside the house, then !leavehouse.

Duration will stop when the item reach the inbox, and never count again, and item become infinite.

2 - Second Decay Bug:

Edit an item with duration less than 5 minutes, decayto attribute and try to reproduce:
Put this item inside depot chest, then logout your character.
Login in again, and look at the item, stop the duration forever and become infinite.

Why need to fix this? For the item "claw of the nexus spawn" when lose item power, transform to new claw of the nexus spawn that have 86400 seconds duration and decay attribute, and is not the same thing like rings or anything that count the time only when equipped, always this item stop decaying forever when the player put in depot chest and leave the game, same with houses.

gugahoa:
Now you're able to stop a item from decaying simple setting ITEM_ATTRIBUTE_DECAYSTATE to 0.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2190,11 +2190,27 @@ void Player::death(Creature* lastHitCreature)
 
 bool Player::dropCorpse(Creature* lastHitCreature, Creature* mostDamageCreature, bool lastHitUnjustified, bool mostDamageUnjustified)
 {
-	if (getZone() == ZONE_PVP) {
-		setDropLoot(true);
-		return false;
+	if (getZone() != ZONE_PVP) {
+		return Creature::dropCorpse(lastHitCreature, mostDamageCreature, lastHitUnjustified, mostDamageUnjustified);
 	}
-	return Creature::dropCorpse(lastHitCreature, mostDamageCreature, lastHitUnjustified, mostDamageUnjustified);
+
+	Player* lastHitPlayer = nullptr;
+	if (lastHitCreature) {
+		lastHitPlayer = lastHitCreature->getPlayer();
+		if (!lastHitPlayer) {
+			Creature* lastHitMaster = lastHitCreature->getMaster();
+			if (lastHitMaster) {
+				lastHitPlayer = lastHitMaster->getPlayer();
+			}
+		}
+	}
+
+	if (!lastHitPlayer) {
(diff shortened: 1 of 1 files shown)


gugahoa:
Do not merge yet. There's a strange behavior I'm still trying to solve.
The very first parcel sent will not start decaying onStepIn, but if you send two parcels they'll start decaying properly. It only happens if you only send ONE parcel. Thoughts?

gugahoa:
There's still a way to stop decay:
Put a item that's decaying inside depot/inbox. Logout. If you take the item out the first time you open the depot locker, it'll not start decaying.

That's because the items inside inbox and depot chest returns true for Item::isRemoved, since they do not have a parent yet. The parent is only set when you open the depot for the first time.

Any suggestions on how to start decaying before opening the depot locker for the first time?

HeavenIsLost:
Another fix can be to set all items to decay when the player login:
After this: https://github.com/otland/forgottenserver/blob/master/src/iologindata.cpp#L542
add this:
`	for (const auto& it : player->depotLockerMap) {
		it.second->setParent(VirtualCylinder::virtualCylinder);
		it.second->startDecaying();
	}`

marksamman:
That seems like a better solution. Do you really need to set the parent to virtual cylinder though?

gugahoa:
You need to set a parent to the depot chest/inbox, or else the items will not decay because item->isRemoved() will return true

If we go with @HeavenIsLost solution, I'll leave this PR open for ppl who want to mimic how it works exactly on global.

vankk:
@gugahoa 
Could you tell me where I can reach you to talk about this code? Skype, otland pm, something?

gugahoa:
My skype is the same as my github username, you can add me there.

PrinterLUA:
Would it not problem, such as with rings which people have. DeEquip and moved to depot, then when you login they will start decay.

gugahoa:
It would, be it would solve the problem.
I made a fix that's not really like this one (that mimics global), but it's smaller.
Stop decay when you put item inside of depot (Player.onMoveItem) and stop decay when sending item to inbox.
Start decay when take any item from inside depot chest (Player.onMoveItem)

Issue #1834, opened 2016-06-09
Monsters's IDLE bug.


slavidodo:
Monsters don't move(dance) when the player is in screen on PZ.

mattyx14:
On PZ

whitevo:
if monster doesn't have targetList of friendList, the onThink event wont work. Maybe that is the case

HeavenIsLost:
On this line:
https://github.com/otland/forgottenserver/blob/master/src/monster.cpp#L1129
I have added:
`	} else if (getTimeSinceLastMove() > 1000) {
		//choose a random direction
		result = getRandomStep(getPosition(), direction);
	}`

it will make the monster do a random walk to any direction while its not idle.

slavidodo:
@HeavenIsLost I will check it as soon as i get to a pc.

Well can someone confirm this for multi-floor as well?

Schiffers:
@HeavenIsLost this change solve the problem with the protection zone, but if the player is trapped by monsters or walls, the monster will stay idle just attack from far

PrinterLUA:
This issue has something releated to my PR: https://github.com/otland/forgottenserver/commit/064630baad244f986de7c53f8d751f8c4f3c655e

slavidodo:
@PrinterLUA If you agree that, please open an issue to revert it back until you make a solution.

PR #1835, opened 2016-06-10
 monsters will just random step if not idle

HeavenIsLost:
I'm just trying to fix this issue #1834 
@slavidodo 

Need testing.


slavidodo:
Yeah i can say it's working, we just need to check if this have side-effects.

PR #1837, opened 2016-06-15
LoadMap in Game

urban119:
This allow load map as specified point (x,y). 
Example:
Game.loadMap("data/world/castle.otbm", 662, 371);
Map Castle.otbm is position (x=1,y=1)
https://youtu.be/gNEJnYDhcT8?t=46s

Old:
Game.loadMap("data/world/castle.otbm");

Regards
Urban

### DIFF
--- a/src/iomap.cpp
+++ b/src/iomap.cpp
@@ -214,7 +214,7 @@ bool IOMap::loadMap(Map* map, const std::string& identifier, int place_x, int pl
 				base_x = area_coord.x;
 				base_y = area_coord.y;
 			}
-			uint16_t base_z = area_coord.z;
+			uint16_t z = area_coord.z;
 
 			NODE nodeTile = f.getChildNode(nodeMapData, type);
 			while (nodeTile != NO_NODE) {
--- a/src/iomap.cpp
+++ b/src/iomap.cpp
@@ -203,7 +203,6 @@ bool IOMap::loadMap(Map* map, const std::string& identifier, int place_x, int pl
 			uint16_t base_y;
 			uint16_t offset_x;
 			uint16_t offset_y;
-			uint16_t base_z = area_coord->z;
 			
 			if (place_x != 0 && place_y != 0) {
 				offset_x = base_x - place_x;
@@ -215,6 +214,7 @@ bool IOMap::loadMap(Map* map, const std::string& identifier, int place_x, int pl
 				base_x = area_coord->x;
 				base_y = area_coord->y;
 			}
+			uint16_t base_z = area_coord->z;
 
 			NODE nodeTile = f.getChildNode(nodeMapData, type);
 			while (nodeTile != NO_NODE) {
--- a/src/iomap.cpp
+++ b/src/iomap.cpp
@@ -201,8 +201,6 @@ bool IOMap::loadMap(Map* map, const std::string& identifier, int place_x, int pl
 			
 			uint16_t base_x = 0;
 			uint16_t base_y = 0;
-			uint16_t offset_x = 0;
-			uint16_t offset_y = 0;
 			
 			if (place_x != 0 && place_y != 0) {
 				base_x = place_x;
--- a/src/iomap.cpp
+++ b/src/iomap.cpp
@@ -199,10 +199,10 @@ bool IOMap::loadMap(Map* map, const std::string& identifier, int place_x, int pl
 				return false;
 			}
 			
-			uint16_t base_x;
-			uint16_t base_y;
-			uint16_t offset_x;
-			uint16_t offset_y;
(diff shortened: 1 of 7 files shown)


src/iomap.cpp
```diff
@@ -198,9 +198,22 @@ bool IOMap::loadMap(Map* map, const std::string& identifier)
...
+			uint16_t base_y = 0;
+			uint16_t offset_x = 0;
+			uint16_t offset_y = 0;
+			
+			if (place_x != 0 && place_y != 0) {
+				offset_x = base_x - place_x;
+				offset_y = base_y - place_y;
+				base_x = base_x - offset_x;
+				base_y = base_y - offset_y;
+			}
```
gugahoa: This is the same as:
base_x = -place_x;
base_y = -place_y;

src/map.h
```diff
@@ -185,8 +185,8 @@ class Map
 		  * Load a map.
 		  * \returns true if the map was loaded successfully
 		  */
-		bool loadMap(const std::string& identifier, bool loadHouses);
-
+		bool loadMap(const std::string& identifier, bool loadHouses, int place_x = 0, int place_y = 0);
+
```
gugahoa: Trailing whitespace

src/iomap.cpp
```diff
@@ -198,9 +198,22 @@ bool IOMap::loadMap(Map* map, const std::string& identifier)
 				setLastErrorString("Invalid map node.");
 				return false;
 			}
-
-			uint16_t base_x = area_coord.x;
-			uint16_t base_y = area_coord.y;
+			
+			uint16_t base_x = 0;
```
djarek: This breaks if you load a map at a position that has multiple areas (all areas are loaded into the same coordinates).

djarek:
Unfortunately this functionality isn't as simple as it might seem. This lua function is inherently unsafe - ***Bad Things™*** will hapen if a new map fragment overlays already existing tiles.

src/iomap.cpp
urban119: no it no breaks. It is just need, becouse declarations must be assigned to anything. Default have 0.
In down code I am have "else", and that assignet to base_x = area_coord.x; (it be standard code).

urban119:
It for me work. TFS use standard loadMap. If you in lua code write loadMap with attribute x,y only then You start use my modyfications.

src/iomap.cpp
djarek: if we take this branch:
```
if (place_x != 0 && place_y != 0) {
```
then:
```
base_x = place_x; 
base_y=place_y;
```
which means that all areas within a map file will be loaded into the same coordinates(overlayed on top of each other).

djarek:
Yeah it will work as long as the loaded map file has one area and the positions of the new tiles are empty. If not, the server might blow up.

src/iomap.cpp
urban119: Yea you have right. I forgot remove that, becouse first my version been create offset.

src/iomap.cpp
urban119: so, that is why is important to chose empty X,Y positions on Your map?

urban119:
If you create script to load x5 map without calculate offset your width partMap... Yes you load x5 partMap on this same position and destroy u server.
You have two way... You clean all tiles x,y or just calculate offset -.^

ranisalt:
Man, your phrases are completely unreadable

djarek:
@urban119 
Okay, let's assume that the caller checks if the area that we load into is empty. This code still fails if you load in a map bigger than 255x255 tiles or with disjoint tile areas.

urban119:
You are the better programmer in TFS, so if you have propositions, just write.
I create yourself OTS, where I am changing standard Quests for My "dynamics team style" raids, dungeons, wars etc. like a Word Of Warcaft.
I changed static loadMap to dynamic loadMap and published that changes for peoples. If someone use that.. its okey, if no so no.
That's all.

Regards
Urban

ranisalt:
I'd like to point that not a single teleport will work if the map is offset.

DSpeichert:
It looks like this PR would introduce a number of bugs as pointed out in review.

The TFS engine was not quite designed for dynamic map editing and this feature, while may be useful to you, does not seem to be something that is popular. If it worked, it could be reviewed but until then, I'm closing this PR.

TheSumm:
Fixing the function to load the map properly would be fairly simple. It just requires to add the offsets to `area_coord`, instead of loading each map area at `place_x, place_y`.

Teleports are an issue though, since the destination might be relative to the loaded map (in this case no teleport would work properly) or the destination might be on the main map (in this case the teleport would work). However, a map might contain both kinds of teleports.

PR #1838, opened 2016-06-17
Small fix to rope

wgrriffel:
in 29be5163dbcdd206a13527a9d437f05af58a2f39 i forget that toPos were defined in tile not in if statement, it need to be checked before isPzLocked.


ranisalt:
Why? I can't see a change, it doesn't seem to matter the order of the check.

wgrriffel:
@ranisalt it matters because if player isn't pz locked it will not check Tile(toPos) then toPosition is not indexed with moveUpstairs, so teleportTo(toPosition) will not work while player isnt pzlocked.

marksamman:
Thanks!

PR #1839, opened 2016-06-18
Added condition type immunity check

raymondtfr:
We can now check if a creature is immune to a condition type.

This I something I had to add to address a latter issue at #1635 when using callbacks, but it is more like a optimization really, since the condition object is in onTargetCreature callback function, and is not bound to a combat object, so it would be added to the target creature everytime (no damage is added, but the block effect keeps happening) regardless of its immunity to that condition.
Example: http://imgur.com/64u6snS

When using combat:setCondition(condition) if the target creature is immune to that condition type then **only** the impact block type effects occurs, but the condition is not added to the creature.

With this, we could prevent a lot of code to be run if we don't need to.


HeavenIsLost:
good :+1:

HeavenIsLost:
@raymondtfr can you send to me your contact info? i want to start a opensource project related to spells... my user on otland is Yamaken, you can send me a private message :)

ranisalt:
It should be named `isImmune`, just like the C++ function.

ranisalt:
Thanks!

PR #1842, opened 2016-06-27
Fix logic expression

ranisalt:
There's no need to double check if monster is summon, lazy evaluation is our best friend.


marksamman:
Thanks!

PR #1844, opened 2016-06-28
FIX Autostack bugs.

Mkalo:
There are 2 bugs in the current autostack code:

1) When moving from a stackable item to the same item but in a different
cylinder it should stack them but currently it stacks with the first
stack in the container.

2) When moving a stackable item to a different cylinder it should
autostack the remain count if it exceeds 100.

Video: https://www.youtube.com/watch?v=jJHsi9MoWNE

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -443,9 +443,15 @@ Cylinder* Container::queryDestination(int32_t& index, const Thing &thing, Item**
 		return this;
 	}
 
+	if (index != INDEX_WHEREEVER) {
+		Item* itemFromIndex = getItemByIndex(index);
+		if (itemFromIndex) {
+			*destItem = itemFromIndex;
+		}
+	}
+
 	bool autoStack = !hasBitSet(FLAG_IGNOREAUTOSTACK, flags);
-	if (autoStack && item->isStackable() && item->getParent() != this) {
-		//try find a suitable item to stack with
+	if (autoStack && item->isStackable() && item->getParent() != this && !(*destItem && (*destItem)->isStackable() && (*destItem)->equals(item))) {		//try find a suitable item to stack with
 		uint32_t n = 0;
 		for (Item* listItem : itemlist) {
 			if (listItem != item && listItem->equals(item) && listItem->getItemCount() < 100) {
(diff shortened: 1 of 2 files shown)


src/container.cpp
```diff
@@ -443,9 +443,16 @@ Cylinder* Container::queryDestination(int32_t& index, const Thing &thing, Item**
...
+		if (itemFromIndex) {
+			*destItem = itemFromIndex;
+		}
+	}
+
 	bool autoStack = !hasBitSet(FLAG_IGNOREAUTOSTACK, flags);
-	if (autoStack && item->isStackable() && item->getParent() != this) {
-		//try find a suitable item to stack with
+	//try find a suitable item to stack with
+	if (autoStack && item->isStackable() && item->getParent() != this && !(*destItem && (*destItem)->isStackable() && (*destItem)->equals(item))) {
```
gugahoa: !(*destItem && (*destItem)->isStackable() && (*destItem)->equals(item))
Can't this be simplified to:
!(*destItem && (*destItem)->equals(item))

I can't verify this right now, but afaik, ->equals should be enough for this check

ranisalt: Keep in mind it's `*destItem` both places.

Seliro:
i tried this update and this is something i noticed:
with (classicEquipmentSlots = true) in config. if you equip a sword for example in the left hand without anything else equipped, it automatic switches to the right hand side instead

src/game.cpp
```diff
@@ -1183,7 +1183,12 @@ ReturnValue Game::internalMoveItem(Cylinder* fromCylinder, Cylinder* toCylinder,
 
 	//add item
 	if (moveItem /*m - n > 0*/) {
-		toCylinder->addThing(index, moveItem);
+		if (fromCylinder == toCylinder) {
+			toCylinder->addThing(index, moveItem);
+			
+		} else {
+			internalAddItem(toCylinder, moveItem, INDEX_WHEREEVER, flags);
```
thexamx: Just FYI this is somehow bugged. When you move item from bp to eq slot and your cap is not enough to carry the item you are moving then item will dissappear.

DSpeichert:
Fixed in 452cf1d / #2400.

PR #1852, opened 2016-07-05
Fix #1656

### LINKED ISSUE #1656: Bug with VIP in account/character
Just a little bug i noticed. You can add yourself on vip list in any other character on the same account and after log in in that character you will see yourself on the vip and also you can talk with yourself.
Shouldn't be possible add or talk with yourself right? I think that should be impossible add any character off the same account on vip list.

wgrriffel:
I've tested on cipsoft servers, you can actually add yourself and all your others characters from account, just cannot open a private chat with yourself.


marksamman:
Thanks!

PR #1854, opened 2016-07-05
NetworkMessage methods

Mkalo:
I think this is very useful.

networkMessage:seek(position) -- sets the position
networkMessage:tell() -- return the current position
networkMessage:len() -- return the length of the buffer

Example of usage:
http://pastebin.com/zGyexExr

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -5473,8 +5473,7 @@ int LuaScriptInterface::luaNetworkMessageSeek(lua_State* L)
 	if (message && isNumber(L, 2)) {
 		message->setBufferPosition(getNumber<uint16_t>(L, 2));
 		pushBoolean(L, true);
-	}
-	else {
+	} else {
 		lua_pushnil(L);
 	}
 	return 1;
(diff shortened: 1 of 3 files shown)


ranisalt:
You should use `len()` instead of `length()` since Lua uses `len()` for everything.

src/luascript.cpp
```diff
@@ -1910,6 +1910,9 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("NetworkMessage", "addItemId", LuaScriptInterface::luaNetworkMessageAddItemId);
 
 	registerMethod("NetworkMessage", "reset", LuaScriptInterface::luaNetworkMessageReset);
+	registerMethod("NetworkMessage", "seek", LuaScriptInterface::luaNetworkMessageSeek);
+	registerMethod("NetworkMessage", "tell", LuaScriptInterface::luaNetworkMessageTell);
+	registerMethod("NetworkMessage", "length", LuaScriptInterface::luaNetworkMessageLength);
```
djarek: Change the length to len

DSpeichert:
@djarek Why is this scheduled for 2.0? Anything in this we can't have now?

DSpeichert:
@Mkalo can you rebase this and maybe fix whatever errors are then thrown by CI?

Issue #1278, opened 2015-04-07
getBaseMaxHealth, getBaseMaxMana


Alscara:
I would love to see a function to get the base, unmodified max health and mana of a creature/player. Certain spells or items give bonuses and buffs to hp and mana and if I try to add 10 to max hp while wearing something that gives a bonus of 100, I gain 110 on top of my 100 bonus, resulting in a permanent increase of 110 rather than 10.

Luckey1729:
Are you asking for functions that return the Tibia max base HP/mana for a player's level and vocation or are you asking for functions that return the unmodified max base HP/mana for a player's level and vocation per rules defined vocations.xml?

Alscara:
I think the second one. I just want to be able to easily grab the maximum hp/mana they would have without equipment or buffs active that give max hp/mana bonuses

EvilHero90:
this can be done in lua with ease, just evaluate the base hp + lvls the character has lvl'd up with player:getVocation():getHealthGainAmount()
ex:
return 185 + (player:getVocation():getHealthGainAmount() * (player:getLevel() - 8))

the -8 is set because 185 is the base health for a lvl 8 character, so we don't want to add the 8 lvls in again.

yrpen:
If someone had "rookie" longer than rest players?

EvilHero90:
I don't see anything wrong with evaluating as lvl 8, barely anyone is using the rook system nowdays.
if someone makes use of it, then he could adjust the formula for itself in his needs.
Storing such stuff in database would equal garbage storing, as you barely make use of it and it'll just make things messier.

PR #1855, opened 2016-07-08
Added Player methods (getBaseMaxHealth & getBaseMaxMana) #1278

ninjalulz:
Simply returns a player's unmodified max health / mana.


marksamman:
Thanks!

Issue #1651, opened 2015-12-21
Game.createItem count max 100?


PrinterLUA:
Hello,

Game.createItem(ITEM_GOLD_COIN, 200, position) only create 100 gold coin and not 200. Bug or intended?

ranisalt:
Intended, as of https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L4244-L4247

What do you expect to happen when >100 is set? Create more than 1 stack of items? Issue a warning at the console?

sgioia9:
I think creating creating more than 1 stack of times would be reasonable

ranisalt:
@PrinterLUA what do you think?

Kamenuvol:
I agree with @sgioia9. But it could be configurable too in config lua.

PrinterLUA:
Yes, it should create then 2x stacks of 100x golf coins.

Mkalo:
Then what should the function return? A table of userdatas?

Milice:
i believe the best and smoothest way to actually perform this action is to just use the function twice instead.

```
Game.createItem(ITEM_GOLD_COIN, 100, position)
Game.createItem(ITEM_GOLD_COIN, 100, position)
```

or even a loop if you want more stacks

```
for i = 1, 5, 1 do
  Game.createItem(ITEM_GOLD_COIN, 100, position)
end
```

Zbizu:
I think that it should stay this way. CreateItem and addItem returns a single object that could be referenced by the scripts (especially items in chests from mapeditor questsystem). It's also really easy to overflow the call stack if you allow any amount of item to be added.

Through looping you can control every stack you're adding.

DSpeichert:
I agree it would be hard to handle errors, what if it "partially fails"?
I can't think of a situation like that (it should be possible to pre-calculate cap and slots for all stacks) but it's possible and if there are ever unique IDs for items, it won't be possible to return one properly.
I'll close this as rejected for now.

Zbizu:
technically we could increase max stack size (eg. 100 -> 999), but we're bound by client limitations so far.

PR #1856, opened 2016-07-08
Added stack support for Game.createItem #1651

ninjalulz:
Returns an array of Item Userdata objects if the specified count exceeds 1 for non-stackable items or 100 for stackable items.


Schiffers:
Then I try test it:
[code]
attempt to index a number value
stack traceback:
        [C]: in ?
        [C]: in function 'createItem'
[code]
is my bad ? my during the test I use it:
Game.createItem(2160, 300, toPosition) and it get this error

PR #1857, opened 2016-07-10
Fix invalid SQL

ranisalt:
Oh, the glorious power of regex engines. That one came unnoticed for 3 years, though it would never fail under normal circumstances.


kornholi:
Thanks!

PR #1858, opened 2016-07-12
fix stack overflow in walkback.lua

Zbizu:
probably related to #1621 also

### DIFF
--- a/data/movements/scripts/walkback.lua
+++ b/data/movements/scripts/walkback.lua
@@ -1,8 +1,12 @@
 function onStepIn(creature, item, position, fromPosition)
 	if item.uid > 0 and item.uid <= 65535 then
 		if position == fromPosition then
-			local temple = creature:getTown():getTemplePosition()
-			creature:teleportTo(temple, false)
+			if creature:isPlayer() then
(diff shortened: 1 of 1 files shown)


marksamman:
This shouldn't be fixed by editing the script, since it worked fine in previous versions. We should pass fromPosition as templePosition instead if position equals fromPosition, that might actually be an elegant fix for #1621.

HeavenIsLost:
@marksamman 

```
Position fromPos = getPosition();
if (oldParent) {
    fromPos = oldParent->getPosition();
} else if (creature && creature->getPlayer()) {
    fromPos = creature->getPlayer()->getTemplePosition();
}

//calling movement scripts
if (creature) {
    g_moveEvents->onCreatureMove(creature, this, fromPos, MOVE_EVENT_STEP_IN);
} else if (item) {
    g_moveEvents->onItemMove(item, this, true);
}
```

PrinterLUA:
I have to agree with Mark at this one, it should be taken care inside the engine and the fix sounds reasonable.

ninjalulz:
This is no longer relevant.

PR #1860, opened 2016-07-14
Update config.lua

urban119:
Many persons have problem with installatin www, becouse no have "passwordType" in standard config


ranisalt:
This configuration key does not exist.

PR #1861, opened 2016-07-14
Update Party Spells

raymondtfr:
I have updated indentation, fixed duplicated ``subId``, simplified some expressions, replaced name of local variable ``tmp`` with ``count`` (a bit better name, as it refers to the affected players list), replaced ``FALSE`` with ``false``, added a missing condition parameter and fixed values of existing ones based on wikia.

### DIFF
--- a/data/spells/scripts/party/enchant.lua
+++ b/data/spells/scripts/party/enchant.lua
@@ -10,9 +10,9 @@ condition:setParameter(CONDITION_PARAM_STAT_MAGICPOINTS, 1)
 condition:setParameter(CONDITION_PARAM_BUFF_SPELL, true)
 
 local baseMana = 120
+
 function onCastSpell(creature, variant, isHotkey)
 	local position = creature:getPosition()
-
 	local party = creature:getParty()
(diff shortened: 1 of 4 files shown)


data/spells/scripts/party/enchant.lua
```diff
@@ -10,9 +10,9 @@ condition:setParameter(CONDITION_PARAM_STAT_MAGICPOINTS, 1)
 condition:setParameter(CONDITION_PARAM_BUFF_SPELL, true)
 
 local baseMana = 120
```
ranisalt: You can actually move this var inside the function, it's not used anywhere else.

raymondtfr: But if I move it into ``onCastSpell()`` won't it be created every time the spell is cast?

ranisalt: Hmm yeah it will, though I don't see if it is a performance hit at all. Keep it that way actually.

marksamman:
Thanks!

PR #1862, opened 2016-07-15
Use round instead of ceil

raymondtfr:
Use round method instead of ceil in means of damage reduction caused by element resistances.

This is something I proved when I was researching monsters for wikia in this last update.

Also removed a trailling tab.


marksamman:
Thanks!

Issue #1618, opened 2015-11-21
Being able to eat more food than intended


peonso:
https://github.com/otland/forgottenserver/blob/26f3e3731915fe5e2984d4b997f82d10769f205c/data/actions/scripts/other/food.lua#L108

should be

    if condition and math.floor(condition:getTicks() / 1000 + food[1] * 12) >= 1200 then

marksamman:
Fixed in 215ddf4dc9be0b9c29c7276a5aecbba110e93d0f.

PR #1863, opened 2016-07-16
update of food.lua

t0kenz:
fixes food bug where you consume too much than supposed to (obviously as it was just checking the initial value in the table)


marksamman:
Thanks!

wgrriffel:
@marksamman #1618

t0kenz:
I see someone did it before I did. I simply fixed it on the project I'm working on and noticed the same issue on a lot of other servers, so I just committed the change ^^

PR #1864, opened 2016-07-18
Remove not necessary statements

wgrriffel:
Items used to a container position will return false in the statment below, about change in spells i see no way to a spell with a range of 1 can pass through walls.

I'd like to ask a question here, why tfs uses 0/1 for booleans in all xml files but outfits.xml? sometimes it confuses and need to check in sources to be sure, like in this thread in [OTLand] (https://otland.net/threads/tfs-1-2-lottery-system.244420/#post-2376102), personally in my distro i changed all booleans to true/false, i think it is more readable.

### DIFF
--- a/data/actions/scripts/tools/pick.lua
+++ b/data/actions/scripts/tools/pick.lua
@@ -1,10 +1,6 @@
 local groundIds = {354, 355}
 
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
-	if toPosition.x == CONTAINER_POSITION then
-		return false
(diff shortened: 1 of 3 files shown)


ranisalt:
For that last statement, probably someone did it first and it stayed that way. All booleans can be true/false, yes/no or 1/0 IIRC.

wgrriffel:
@ranisalt i know, what i mean was should i push this commits to branch or keep as is?

marksamman:
Thanks!

PR #1870, opened 2016-08-01
Fix code style

raymondtfr:


ranisalt:
We could push a stricter style guide by using a tool, but I only know clang-format and it's unsuitable (too much hardcoded defaults)

PR #1875, opened 2016-08-16
Improve NodeStruct, PropStream and PropWriteStream for simplicity

ranisalt:
- `NodeStruct` and `PropStream` were refactored for default attributes instead of initializing constructor, and now use compiler-generated constructor.

- `PropWriteStream` was doing very much the same `std::vector` already does, so it was refactored to use a vector internally as a buffer.

Additionally they are pure RAII now.

Also, spec says "a function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function", so the keyword is not needed.

It is probably unrelated, but the server now loads about 7% faster for me, might just be a lighter CPU load.

### DIFF
--- a/src/fileloader.h
+++ b/src/fileloader.h
@@ -20,18 +20,19 @@
 #ifndef FS_FILELOADER_H_9B663D19E58D42E6BFACFE5B09D7A05E
 #define FS_FILELOADER_H_9B663D19E58D42E6BFACFE5B09D7A05E
 
+#include <limits>
+#include <vector>
+
 struct NodeStruct;
 
 typedef NodeStruct* NODE;
 
 struct NodeStruct {
-	NodeStruct() : start(0), propsSize(0), type(0), next(nullptr), child(nullptr) {}
-
-	uint32_t start;
-	uint32_t propsSize;
-	uint32_t type;
-	NodeStruct* next;
-	NodeStruct* child;
+	uint32_t start = 0;
+	uint32_t propsSize = 0;
+	uint32_t type = 0;
+	NodeStruct* next = nullptr;
+	NodeStruct* child = nullptr;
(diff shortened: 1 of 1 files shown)


kornholi:
Thanks! I don't think this code has been touched for 10 years.

ranisalt:
:) I found some old code that can be modernized and I'm going to submit some patches

djarek:
@ranisalt
I've done some work on OTBI/OTBM parsing but did not make a PR yet because the code hasn't been verified in 100% yet:
https://github.com/djarek/forgottenserver/commits/fileloader_cleanup

3a32a77df41d9f9be8cf3866cc9d46533eccca69 should be fairly stable and working

I've also done some work to create a single-pass otbm parser (without a parsing tree) that loads the file into memory using portable mmap-like interfaces. But the changes are sitting on my disk still and I haven't yet ported map loading to this new parser.

ranisalt:
@djarek if you could just push as WIP we could contribute and help you figure it out 😄

kornholi:
A streaming OTBM parser is relatively easy to implement if you save the area origin and current tile position. My toy map renderer loads the TFS map in 100ms and a ~90MB map in 2500ms. Not sure how this compares to TFS, but maybe it can help you.

https://github.com/kornholi/mapeditor/blob/master/src/opentibia/binaryfile.rs#L52
https://github.com/kornholi/mapeditor/blob/master/src/opentibia/map.rs#L181

djarek:
@kornholi 
A streaming OTBM is indeed relatively easy. The hard part is hooking it up to the tfs map loader which requires inverting the control flow of the whole loading process. ;)

kornholi:
...?

PR #1876, opened 2016-08-17
Fix const correctness and references

ranisalt:
Some methods unnecessarily copy data instead of passing plain references, hopefully this will improve (minor) performance and makes code safer.

### DIFF
--- a/src/chat.h
+++ b/src/chat.h
@@ -33,7 +33,7 @@ class ChatChannel
 {
 	public:
 		ChatChannel() = default;
-		ChatChannel(uint16_t channelId, std::string channelName):
+		ChatChannel(uint16_t channelId, const std::string& channelName):
 			name(channelName),
 			canJoinEvent(-1),
 			onJoinEvent(-1),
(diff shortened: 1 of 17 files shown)


marksamman:
The missing references are intentional because they will be copied anyway (see http://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html#pass-by-value-in-constructors, although they use std::move there, but I think the compiler is smart enough that we don't need that). I don't think this results in the objects being copied twice, but correct me if I'm wrong.

ranisalt:
> I don't think this results in the objects being copied twice, but correct me if I'm wrong.

Well, if they do, copying shared ptrs is very cheap as it's simple as copying a pointer and incrementing a counter. The copies are missing the move, though.

Anyway, that's a great article, was the entire code checked with clang-tidy? It would do good for code quality.

djarek:
@ranisalt 
> Well, if they do, copying shared ptrs is very cheap as it's simple as copying a pointer and incrementing a counter.
A shared_ptr copy can be an order of magnitude more expensive than a simple double word(2 pointers) copy, but worrying about it is probably pointless, because we don't copy shared_ptrs anywhere near the hot code paths. 
Source: http://stackoverflow.com/questions/2538070/atomic-operation-cost

PR #1877, opened 2016-08-17
Optimize build flags

ranisalt:
Those options reduce the binary size by 30% (for me, 3.55MB to 2.56MB), though compilation times take longer with LTO. Visibility change is responsible for around 400kB, while link time optimization shrinks binary by >600kB and improves performance.

On the other hand, `-Ofast` is not supported by clang until 3.8 which is very recent and did little extra optimization over `-O3`, besides doing optimizations that are not compliant to the spec, so `-O3` is a safer option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,19 +5,14 @@ set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
 
 project(tfs)
 
-set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
+list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 include(cotire)
 
-# Initialize CXXFLAGS.
-set(CMAKE_CXX_FLAGS                "-Wall -Werror -pipe")
-set(CMAKE_CXX_FLAGS_DEBUG          "-O0 -g")
-set(CMAKE_CXX_FLAGS_MINSIZEREL     "-Os -DNDEBUG")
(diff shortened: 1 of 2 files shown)


marksamman:
The clang build is failing in Travis.

ranisalt:
Yeah, just saw that, and it's strange since clang supports both LTO and visibility since a long time. I'll check what I can do.

ranisalt:
@marksamman I just checked and all our targets (debug, relwithdebinfo etc) are just setting the defaults, so we don't need to set them explicitly. Can I remove them? We could shrink to just setting extra `CMAKE_CXX_FLAGS` for warnings and those new options.

marksamman:
Sure.

I've had problems with -flto in the past where it was causing linker errors related to Boost if the gold linker wasn't installed, but this seems to be something different.

djarek:
@ranisalt
Before this is merged please make sure that there are no strict aliasing issues anywhere in the code base (check pointer casts - reinterpret_cast, C-style casts and static_cast from void*). IIRC I fixed some of them a while ago, but I'm not certain if I got all of them. In my experience LTO has a bad habit of making dragons fly out of your nose if you have some wonky pointer casts.

ranisalt:
I added some more commits to improve cmake stuff in general, such as using cmake commands instead of control strings and dropping deprecated stuff (e.g. testing for -std=c++0x flag as -std=c++11 is ubiquitous on supported systems)

marksamman:
Cool, but it seems like Travis is still failing with the clang builds, if you can fix that I'll merge this.

ranisalt:
Yeah, clang is being a pain, I can't reproduce it locally even with the same clang version (it fails with another error >.< while 4.0 works)

ranisalt:
It's done now, yet I still couldn't figure how to use LTO with Clang.

marksamman:
Thanks!

PR #1879, opened 2016-08-19
Modernize code to improve quality

ranisalt:
After #1876 I searched more about modernizing code and applied a few fixes to improve code quality, readability and maintainability:

- Modernize constructors with pass-by-value + move (the purpose of #1876)
- Modernize loops to use new range-based loops
- Use auto where it improves code readability and maintainability

Using range-based loops reduces indirection overhead.

The latter is mainly using `auto` for iterators, but could be extended to use auto for obvious return types (e.g. on `Player* player = something->getPlayer()` or `MatrixArea* area = new MatrixArea()`), though I saw it being rejected once.

It adds maintainability because if we ever change the underlying container we do not need to reflect the change elsewhere (e.g. there's a `std::map<Item*, uint32_t>::iterator` that could be changed to an unordered map).

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -853,9 +853,9 @@ int_fast32_t AStarNodes::getTileWalkCost(const Creature& creature, const Tile* t
 // Floor
 Floor::~Floor()
 {
-	for (uint32_t i = 0; i < FLOOR_SIZE; ++i) {
-		for (uint32_t j = 0; j < FLOOR_SIZE; ++j) {
-			delete tiles[i][j];
+	for (auto& row : tiles) {
+		for (auto tile : row) {
+			delete tile;
 		}
 	}
 }
@@ -927,7 +927,7 @@ void QTreeLeafNode::addCreature(Creature* c)
 
 void QTreeLeafNode::removeCreature(Creature* c)
 {
-	CreatureVector::iterator iter = std::find(creature_list.begin(), creature_list.end(), c);
+	auto iter = std::find(creature_list.begin(), creature_list.end(), c);
 	assert(iter != creature_list.end());
 	*iter = creature_list.back();
 	creature_list.pop_back();
@@ -964,9 +964,8 @@ uint32_t Map::clean() const
 					continue;
 				}
 
-				for (size_t x = 0; x < FLOOR_SIZE; ++x) {
-					for (size_t y = 0; y < FLOOR_SIZE; ++y) {
-						Tile* tile = floor->tiles[x][y];
+				for (auto& row : floor->tiles) {
+					for (auto tile : row) {
 						if (!tile || tile->hasFlag(TILESTATE_PROTECTIONZONE)) {
 							continue;
 						}
@@ -992,8 +991,7 @@ uint32_t Map::clean() const
 				}
 			}
 		} else {
-			for (size_t i = 0; i < 4; ++i) {
-				QTreeNode* childNode = node->child[i];
+			for (auto childNode : node->child) {
 				if (childNode) {
 					nodes.push_back(childNode);
 				}
--- a/src/player.h
+++ b/src/player.h
@@ -1231,7 +1231,7 @@ class Player final : public Creature, public Cylinder
 		Group* group = nullptr;
 		Inbox* inbox;
 		Item* tradeItem = nullptr;
- 		Item* inventory[CONST_SLOT_LAST + 1] = { nullptr };
+ 		Item* inventory[CONST_SLOT_LAST + 1] = {};
 		Item* writeItem = nullptr;
 		House* editHouse = nullptr;
 		Npc* shopOwner = nullptr;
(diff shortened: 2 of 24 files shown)


ranisalt:
There are some variables I couldn't figure a better name for, please review them prior to merging 👍

Kamenuvol:
Why use std::move only with strings? How about the other object types, classes? I really want to understand this move thing in order to improve my project.

ranisalt:
> Why use std::move only with strings? How about the other object types, classes? I really want to understand this move thing in order to improve my project.

Because moving a string is simple, not a significant cost (only primitive type swaps). Other types might not have that costless move constructors, but if you find any, please comment :)

src/map.cpp
```diff
@@ -994,10 +994,9 @@ uint32_t Map::clean() const
 					continue;
 				}
 
-				for (size_t x = 0; x < FLOOR_SIZE; ++x) {
-					for (size_t y = 0; y < FLOOR_SIZE; ++y) {
-						Tile* tile = floor->tiles[x][y];
-						if (!tile || tile->hasFlag(TILESTATE_PROTECTIONZONE)) {
+				for (auto& row : floor->tiles) {
+					for (auto tile : row) {
+							if (!tile || tile->hasFlag(TILESTATE_PROTECTIONZONE)) {
```
marksamman: Fix the indentation.

src/player.h
```diff
@@ -90,8 +90,8 @@ enum tradestate_t : uint8_t {
 };
 
 struct VIPEntry {
-	VIPEntry(uint32_t guid, std::string name, const std::string& description, uint32_t icon, bool notify)
-		: guid(guid), name(name), description(description), icon(icon), notify(notify) {}
+	VIPEntry(uint32_t guid, std::string name, std::string  description, uint32_t icon, bool notify)
```
marksamman: Remove the extra space between std::string and description.

src/tasks.h
```diff
@@ -31,11 +31,11 @@ class Task
 {
 	public:
 		// DO NOT allocate this class on the stack
-		Task(uint32_t ms, const std::function<void (void)>& f) : func(f) {
+		Task(uint32_t ms, std::function<void (void)>  f) : func(std::move(f)) {
```
marksamman: Remove the extra space.

src/spawn.h
```diff
@@ -38,7 +38,7 @@ struct spawnBlock_t {
 class Spawn
 {
 	public:
-		Spawn(const Position& pos, int32_t radius) : centerPos(pos), radius(radius), interval(60000), checkSpawnEvent() {}
+		Spawn(Position  pos, int32_t radius) : centerPos(std::move(pos)), radius(radius), interval(60000), checkSpawnEvent() {}
```
marksamman: Remove the extra space.

src/tasks.h
```diff
@@ -31,11 +31,11 @@ class Task
...
 			expiration = std::chrono::system_clock::now() + std::chrono::milliseconds(ms);
 		}
-		explicit Task(const std::function<void (void)>& f)
-			: expiration(SYSTEM_TIME_ZERO), func(f) {}
+		explicit Task(std::function<void (void)>  f)
```
marksamman: Remove the extra space.

src/map.cpp
```diff
@@ -863,9 +863,9 @@ int_fast32_t AStarNodes::getTileWalkCost(const Creature& creature, const Tile* t
 // Floor
 Floor::~Floor()
 {
-	for (uint32_t i = 0; i < FLOOR_SIZE; ++i) {
-		for (uint32_t j = 0; j < FLOOR_SIZE; ++j) {
-			delete tiles[i][j];
+	for (auto& tile : tiles) {
+		for (auto& j : tile) {
```
marksamman: j is already a pointer to a tile here, so we don't need a reference to it since copying pointers is cheaper than accessing it by another indirection. You could also change it to explicitly say Tile* instead of auto& so that it's easier to understand that when looking at this code.

ranisalt: Well, assuming the two-dimensional array is contiguous, I find it nicer to be:
```
for (Tile* ptr = tiles[0][0]; ptr < tiles[FLOOR_SIZE][FLOOR_SIZE] + 1; ++ptr) {
  delete ptr;
}
```
Since it has only 1 level.

marksamman: That won't work since ptr is the pointer to the tile in the first index, the tile stored at the second index is not necessarily allocated at the the memory address of the tile allocated at the first index + sizeof(Tile), in fact it might be NULL but your code will try to delete at some other address anyway, also if the tile at the first index is NULL then it'll most likely segfault at the second index anyway because then you'll be deleting at 0+sizeof(Tile). This destructor is rarely called, if it's called at all, that it's not worth trying to do something fancy although I understand what you're trying to do.

ranisalt: Hmm yeah, I just realized each value is a Tile**, not a Tile*, and it would be more like this:

```
for (Tile** ptr = &tiles[0][0]; ptr < (&tiles[FLOOR_SIZE][FLOOR_SIZE]) + 1; ++ptr) {
  delete *ptr;
}
```

But the range for is much cleaner. I'll adjust it accordingly.

ranisalt:
Rebased against master

marksamman:
All Travis builds failed.

marksamman:
Thanks!

PR #1880, opened 2016-08-22
Summons to ignore 'runonhealth' & attack anyway

### LINKED ISSUE #1867: Summon runonhealth
Summoned monster does not attack while "running on low health".

Steps to reproduce:
1. summon monster > bring its health below "runonhealth" value.
2. attack monster
3. summon will run up to the target and then randomly "dance".

What is suppose to happen:
Summon ignores "runonhealth" and attacks anyway. (tested and confirmed on live servers)

ninjalulz:
Fixes #1867


marksamman:
Thanks!

Issue #1437, opened 2015-07-02
monsterType:getCreatureEvents() does not list registered events.


whitevo:
I am using TFS 1.0

When I do this:
target:registerEvent(eventName)

monsterType = target:getType()
allMonsterEvents = monsterType:getCreatureEvents()

allMonsterEvents only has the creaturescripts I have added on the monster.xml
It doesn't show registered events trough LUA.

Supposed to be like that? Is it different in higher TFS versions?
##

whitevo:
I currently making dispel spell for game, and it should remove creaturescripts too.
I want to count how many creaturescripts it is removing or does monster have creaturescrips activated.

Problem is that: that function does not show registered events.. unregisterEvent() always returns true does not matter did monster have that creaturescript activated or not.

whitevo:
what do you mean by unconfirmed xD?

Alscara:
you would have to understand what the class means. monsterType pulls info about a monster from the xml. It can pull things like name, resistances, etc. monsterType cannot access actual monsters in-game. It only pulls what can be found in the xml file. This is not a bug, it's exactly as the system was designed. monsterType:getCreatureEvents() will return a creatureEvent if there is only one, or a table containing all creatureEvents within that monster's file.

Alscara:
In addition, the "unconfirmed" label means Mark has seen your issue but has not actually looked into it, or has not looked into it enough to offer a verdict on whether it's a problem or not.

whitevo:
Ah ok I understand.
Seems I have generate new table what keeps track of monster creaturescripts then xD

Although currently I just register the same event on monster each time something happens.
Doesn't seem its doing anything bad, but i can be wrong.

ninjalulz:
@ranisalt Please close this issue.

PR #1881, opened 2016-08-22
Added a new Creature method (getEvents)

ninjalulz:
I just saw an old issue (#1437) where the user tried to fetch all creature events registered to a monster by using MonsterType.getCreatureEvents, but that would only fetch creature events registered through its XML file.

So I thought a method where you can fetch creature events by a specific event type would come in handy.

Example of usage:
```Lua
local monster = Game.createMonster("Rat", Position(1000, 1000, 7))
if not monster then
    return
end

local killEvents = monster:getEvents(CREATURE_EVENT_KILL)
for index, name in ipairs(killEvents) do
    print(index, name)
end
```

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1347,6 +1347,20 @@ void LuaScriptInterface::registerFunctions()
 	registerEnum(CONST_SLOT_RING)
 	registerEnum(CONST_SLOT_AMMO)
 
+	registerEnum(CREATURE_EVENT_NONE)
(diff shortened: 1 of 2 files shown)


marksamman:
Thanks!

ranisalt:
@marksamman in the wiki you wrote it's been added on version 1.0, did you mean 1.2?

marksamman:
@ranisalt yup, I just copied the function above and forgot to edit the version, I've fixed it. Thanks for spotting it!

Issue #1491, opened 2015-08-04
Death to monster in Pvp Zone Tile


walissonsantos:
If a monster kill the player in a pvp zone tile, the player dies and relogin infinitely.

Sorry my bad english.

luanlucian93:
Confirmed.

ranisalt:
Can you provide a video or gif with demonstration?

vankk:
I can confirm this bug!

@ranisalt 
![animation](https://cloud.githubusercontent.com/assets/14372849/13309918/f4e92066-db5f-11e5-8a0e-3b398cfd6abc.gif)

ranisalt:
I believe it's because you are not supposed to be killed on a PvP zone tile, but that is only checked if you are hit by a player. PvP zone tiles are not supposed to host monsters, only players and summons if applicable.

vankk:
> PvP zone tiles are not supposed to host monsters

I agree with that but there is no way to fix this? Avoid this type of error.

ninjalulz:
@ranisalt Please close this issue.

PR #1883, opened 2016-08-22
Fix an issue with deaths by monsters in PVP Zones

ninjalulz:
This should fix the following issue: #1491

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2190,11 +2190,27 @@ void Player::death(Creature* lastHitCreature)
 
 bool Player::dropCorpse(Creature* lastHitCreature, Creature* mostDamageCreature, bool lastHitUnjustified, bool mostDamageUnjustified)
 {
-	if (getZone() == ZONE_PVP) {
-		setDropLoot(true);
-		return false;
+	if (getZone() != ZONE_PVP) {
+		return Creature::dropCorpse(lastHitCreature, mostDamageCreature, lastHitUnjustified, mostDamageUnjustified);
 	}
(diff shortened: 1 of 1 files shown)


marksamman:
Thanks!

src/player.cpp
```diff
@@ -2190,11 +2190,27 @@ void Player::death(Creature* lastHitCreature)
...
+		}
 	}
-	return Creature::dropCorpse(lastHitCreature, mostDamageCreature, lastHitUnjustified, mostDamageUnjustified);
+
+	if (!lastHitPlayer) {
+		return Creature::dropCorpse(lastHitCreature, mostDamageCreature, lastHitUnjustified, mostDamageUnjustified);
+	}
+
+	setDropLoot(true);
+	return false;
```
ranisalt: This function to check if the last hit was by a player could be externalized:

```
bool lastHitIsPlayer(Creature* lastHitCreature)
	if (!lastHitCreature) {
		return false;
	}

	if (lastHitCreature->getPlayer()) {
		return true;
	}

	return lastHitCreature->getMaster() != nullptr;
}
```

And this dropLoot function would be reduced to:

```
if (getZone() != ZONE_PVP || !lastHitIsPlayer(lastHitCreature)) {
	return Creature::dropCorpse(lastHitCreature, mostDamageCreature, lastHitUnjustified, mostDamageUnjustified);
}

setDropLoot(true);
return false;
```

Which in my opinion has greater clarity of what are the conditions to drop loot.

PR #1581, opened 2015-10-04
Update weapons - New Melee/Distance dmg, Armor and Defense formulas

raymondtfr:
**Weapons folder**
* Renamed explosion_arrow to ```burst_arrow```
* Updated weapons formulas, conditions, effects, and added missing
parameters ```block_armor``` and ```block_shield```.
* ```setFormula```parameter values has been modified in a way that it won't modify the values
generated by ```getMaxWeaponDamage``` formula in *weapons.cpp*.

**Now (Melee / Distance Damage, Armor and Defense) formulas are based on Tibia-stats.**

**New Defense formula**
* Defense formula now uses ```Fist Fighting``` skill if neither weapon nor shield is
equipped. No matter if your weapon's defense ```weaponDefense +
weaponExtraDefense``` is higher than your shield's, defense from shield
```shieldDefense``` or ```shieldDefense + weaponExtraDefense``` will always be used instead.
* No matter which defense mode you activated *(Defensive, Balanced or
Offensive)* it will all act as *Defensive* unless you are actively attacking a creature, etc...

A few comparisons in **PvM**

**Armor Reduction formula**
* Armor = 50

| **Reduction**  | **New Formula**  | **Current Formula** | **Diff. range** |
|:------------:|:---------------:|:-----:|:-----:|
| **Maximum**     | 49  | 47 | +2 |
| **Minimum**      | 25  | 24 | +1 |
| **Variation**      | 24 | 23 | +1|

**Defense formula**
* Shielding skill = 100
* Shield defense = 37
* Weapon defense = 31 + 3 (extra defense)

| **Defense Factor**  | **New Formula**  | **Current Formula** | **Diff. range** |
|:------------:|:---------------:|:-----:|:-----:|
| **Defensive**  | 163 | 144 | +19 |
| **Balanced**  | 122        |   87 | +35
| **Offensive** | 81        |    72 | +9

**Melee / Distance Damage formula**
Level = 100
Weapon attack = 50
Weapon skill = 100

| **Attack Factor**  | **New Formula**  | **Current Formula** | **Diff. range** |
|:------------:|:---------------:|:-----:|:-----:|
| **Offensive**      | 462 | 443 | +19 |
| **Balanced**      | 388        |   370 | +18
| **Defensive** | 241        |    222 | +19

PR #1887, opened 2016-08-23
New Armor Reduction Formula

raymondtfr:
This is a re-re-remake of #1581, 

This new formula for damage reduction by armor value is based of off tibia-stats research (which I have also researched and confirmed on my own).

There's just one mistake about t-s research which is that their calculator shows that the damage is reduced by 1 when the armor value is 0.

This also brings a small refactor to this block of code.

I have researched and confirmed that there's no damage reduction for an armor value of 0.

A comparison table

* Armor = 50

| **Reduction**  | **New Formula**  | **Current Formula** | **Diff. range** |
|:------------:|:---------------:|:-----:|:-----:|
| **Maximum**     | 49  | 47 | +2 |
| **Minimum**      | 25  | 24 | +1 |
| **Variation**      | 24 | 23 | +1|

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -865,15 +865,10 @@ BlockType_t Creature::blockHit(Creature* attacker, CombatType_t combatType, int3
 		}
 
 		if (checkArmor) {
-			int32_t armorValue = getArmor();
-			if (armorValue > 1) {
-				double armorFormula = armorValue * 0.475;
-				int32_t armorReduction = static_cast<int32_t>(std::ceil(armorFormula));
-				damage -= uniform_random(
-					armorReduction,
-					armorReduction + static_cast<int32_t>(std::floor(armorFormula))
-				);
-			} else if (armorValue == 1) {
+			int32_t armor = getArmor();
+			if (armor > 3) {
+				damage -= uniform_random(armor / 2, armor - (armor % 2 + 1));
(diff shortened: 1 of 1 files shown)


ranisalt:
`(armor % 2) == 0 ? armor - 1 : armor - 2` is equivalent to `armor - (armor % 2 + 1)` but the latter does no branching.

raymondtfr:
I find the ternary operator more clear for this operation, although I can change it to your example if readability is not so important in this case.

marksamman:
Thanks!

src/creature.cpp
```diff
@@ -865,15 +865,10 @@ BlockType_t Creature::blockHit(Creature* attacker, CombatType_t combatType, int3
...
-				int32_t armorReduction = static_cast<int32_t>(std::ceil(armorFormula));
-				damage -= uniform_random(
-					armorReduction,
-					armorReduction + static_cast<int32_t>(std::floor(armorFormula))
-				);
-			} else if (armorValue == 1) {
+			int32_t armor = getArmor();
+			if (armor > 3) {
+				damage -= uniform_random(armor / 2, armor - (armor % 2 + 1));
+			} else if (armor > 0) {
```
ranisalt: This is so much more concise and beautiful than the previous formula 👍

PR #1888, opened 2016-08-23
Rename config.lua to prevent accidental push, fixes #1772

### LINKED ISSUE #1772: config.lua should be added to gitignore
### Expected behaviour
In my opinion config.lua should be added to .gitignore. Someone may accidentally push his config to repository...

### Actual behaviour
It's not

ranisalt:
As proposed by #1772 config.lua is moved to config.lua.sample to force new users to read and change it prior to starting server. Also, config.lua is added to ignored files so it will not be pushed to a fork accidentally.


HeavenIsLost:
Good :)

marksamman:
I'm not sure I want to add an additional step that people need to understand to run the server. If we do this, we should at least make the config loader try to load config.lua.sample if config.lua is not found.

ranisalt:
That's a point, but I don't think the defaults should be allowed to run. People should, at least, change the password database, and with this additional step either you look the configuration and change to use or you explicitly want the defaults.

That's a production feature, though. I run the default configuration while developing anyways. Also, since in source build is disabled, people will need to copy config.lua anyways, it's just a name change.

WibbenZ:
What about if we can't find config.lua we inform that the file (might) need to be renamed?
That way they have to change it but we won't get a billion questions that config.lua won't load, the code will insted tell them what the issue might be.

PR #1889, opened 2016-08-25
Fix script environment alignment padding

ranisalt:
Script environment instances were taking 11 bytes of padding, taking a total of 96 bytes, which is suboptimal. It now takes a total of 88 bytes, an optimal padding for those members.

This is a very small optimization though, as only 16 script environments are ever constructed, saving 64 bytes in total.


kornholi:
You probably wasted more time than it's worth, but thanks!

PR #1809, opened 2016-06-01
Check for access when editing house access list, fixes #1482

ninjalulz:

PR #1890, opened 2016-08-25
Minor code cleanup

ninjalulz:
Related to #1809 & #1883 

Special thanks to @ranisalt

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1990,7 +1990,7 @@ void Player::death(Creature* lastHitCreature)
 
 	if (skillLoss) {
 		uint8_t unfairFightReduction = 100;
-		bool lastHitPlayer = lastHitIsPlayer(lastHitCreature);
+		bool lastHitPlayer = Player::lastHitIsPlayer(lastHitCreature);
 
 		if (lastHitPlayer) {
 			uint32_t sumLevels = 0;
@@ -2167,7 +2167,7 @@ void Player::death(Creature* lastHitCreature)
 
 bool Player::dropCorpse(Creature* lastHitCreature, Creature* mostDamageCreature, bool lastHitUnjustified, bool mostDamageUnjustified)
 {
-	if (getZone() != ZONE_PVP || !lastHitIsPlayer(lastHitCreature)) {
+	if (getZone() != ZONE_PVP || !Player::lastHitIsPlayer(lastHitCreature)) {
 		return Creature::dropCorpse(lastHitCreature, mostDamageCreature, lastHitUnjustified, mostDamageUnjustified);
 	}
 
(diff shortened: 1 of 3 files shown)


src/player.cpp
```diff
@@ -1990,33 +1990,24 @@ void Player::death(Creature* lastHitCreature)
...
-				for (const auto& it : damageMap) {
-					CountBlock_t cb = it.second;
-					if ((OTSYS_TIME() - cb.ticks) <= inFightTicks) {
-						Player* damageDealer = g_game.getPlayerByID(it.first);
-						if (damageDealer) {
-							sumLevels += damageDealer->getLevel();
-						}
+		bool lastHitPlayer = lastHitIsPlayer(lastHitCreature);
+
+		if (lastHitPlayer) {
```
ranisalt: Can shrink to `if (lastHitIsPlayer(lastHitCreature))`.

src/player.h
```diff
@@ -594,6 +594,7 @@ class Player final : public Creature, public Cylinder
 		bool isImmune(ConditionType_t type) const final;
 		bool hasShield() const;
 		bool isAttackable() const final;
+		bool lastHitIsPlayer(Creature* lastHitCreature) const;
```
ranisalt: Use a free function, not a method for faster compiling as it's only useful in player.cpp compilation unit (will never be used outside)

src/player.cpp
gugahoa: I think he did that way because of what's at line 2107

ranisalt: Yes, I'm short sighted and wasn't wearing my internet glasses.

src/player.h
ninjalulz: Alright. I have never heard of the expression 'free function' before, are you simply referring to a non-member function that you only declare inside the cpp file? If so, what is the preferred method to do so? Declare the function before the member function where it's being utilized, or at the bottom (and provide a function prototype for the function at the top).

raymondtfr: I guess  @ranisalt meant a "lambda"? or a ``friend`` function. :dancers:

marksamman: I think he means moving the function into player.cpp and not declaring it in the header file. You could also just make it static.

ninjalulz: I see, well I decided to make it static instead to avoid any code inconsistency. :smile:

marksamman:
Thanks!

src/player.h
ranisalt: Yes, you understood correctly. A free function is a non-member function, and if you make it static and only on the cpp it's not visible outside that file compilation unit, so you don't need to recompile if the function signature ever changes, etc.

Declare it just before the function it is used. A static lambda is also useful if it is used only by a single method.

PR #1891, opened 2016-08-25
Restart item decay in Lua after taking it out of depot

### LINKED ISSUE #1753: Decay Bug when put decaying item inside Depot Chest and logout.

gugahoa:
Fixes #1753 
The item that stopped decaying on logout (bug), will return to decay after being taken out of depot or inbox

### DIFF
--- a/data/events/events.xml
+++ b/data/events/events.xml
@@ -16,7 +16,7 @@
 	<event class="Player" method="onLookInBattleList" enabled="1" />
 	<event class="Player" method="onLookInTrade" enabled="1" />
 	<event class="Player" method="onLookInShop" enabled="0" />
-	<event class="Player" method="onMoveItem" enabled="0" />
+	<event class="Player" method="onMoveItem" enabled="1" />
 	<event class="Player" method="onMoveCreature" enabled="0" />
 	<event class="Player" method="onTurn" enabled="0" />
 	<event class="Player" method="onTradeRequest" enabled="0" />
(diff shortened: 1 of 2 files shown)


gugahoa:
There shouldn't be anything blocking this from fixing #1753 
@marksamman @ranisalt

ranisalt:
I'd rather fix in the sources as I don't like the idea of a script to fix a bug, it's too fragile for a clumsy scripter to unhide a bug inadvertently.

wgrriffel:
@ranisalt i agree, either i think #1591 should be reopen to the track

DSpeichert:
@ranisalt I don't see how #1591 relates to this issue. Are you sure you meant to reopen that?

ranisalt:
I don't remember, probably a mistake.

wgrriffel:
@DSpeichert @ranisalt it does not relate directly with #1591, the problem is that it was fixed through a lua script, which is not proper to this kind of issue, the same with this one

PR #1894, opened 2016-08-29
use round method instead of ceil

raymondtfr:
I missed this one in #1862.


marksamman:
Thanks!

ranisalt:
I think it does not matter if you round since it will be rounded by the static cast anyway.

kornholi:
casting doesn't round, it truncates the fraction part

ranisalt:
Yes, I got confused with the parenthetizing

PR #1895, opened 2016-08-29
Fix field damage to players

raymondtfr:
Damage dealt by magical fields to players should be half damage and rounded.


marksamman:
Thanks!

Issue #1896, opened 2016-08-29
Segmentation Fault when using LuaJIT


gugahoa:
1. Clone the repo with the latest commit, compile it (having LuaJIT 5.1)
2. Execute tfs

### Expected behaviour
Should load everything

### Actual behaviour
Segmentation fault when loading script systems

### Environment

> The Forgotten Server - Version 1.2
> Compiled with GNU C++ version 6.1.1 20160802
> Compiled on Aug 29 2016 19:41:56 for platform x64
> 
> A server developed by Mark Samman
> Visit our forum for updates, support, and resources: http://otland.net/.
> 
> \>\> Loading config
> \>\> Establishing database connection... MySQL 10.1.16-MariaDB
> \>\> Running database manager
> \>\> Loading vocations
> \>\> Loading items
> \>\> Loading script systems
> [1]    4182 segmentation fault (core dumped)  ./tfs
> 

`(gdb) bt
#0  0x00007ffff760cd6a in ?? () from /usr/lib/libluajit-5.1.so.2
#1  0x00007ffff7625b1c in lua_next () from /usr/lib/libluajit-5.1.so.2
#2  0x0000000000922185 in LuaScriptInterface::getArea (L=0x40000378, list=..., rows=@0x7ffff49560fc: 0)
    at /home/gugah/Development/Tibia/forgottenserver/src/luascript.cpp:2991
#3  0x000000000092248f in LuaScriptInterface::luaCreateCombatArea (L=0x40000378)
    at /home/gugah/Development/Tibia/forgottenserver/src/luascript.cpp:3039
#4  0x00007ffff75e23c6 in ?? () from /usr/lib/libluajit-5.1.so.2
#5  0x00007ffff7627ba0 in lua_pcall () from /usr/lib/libluajit-5.1.so.2
#6  0x00000000008cb498 in LuaScriptInterface::protectedCall (L=0x40000378, nargs=0, nresults=0)
    at /home/gugah/Development/Tibia/forgottenserver/src/luascript.cpp:295
#7  0x00000000008cb5d4 in LuaScriptInterface::loadFile (this=0x7fffec058ab0, file="data/spells/scripts/party/enchant.lua", npc=0x0)
    at /home/gugah/Development/Tibia/forgottenserver/src/luascript.cpp:325
#8  0x00000000007b378b in Event::checkScript (this=0x7fffec080ac0, basePath="data/spells/", scriptsName="spells", 
[...]

PR #1897, opened 2016-08-29
Enforce backward compatibility (#1896)

gugahoa:


marksamman:
This should be in compat.

gugahoa:
It doesn't load compat before loading spells, so it'll crash anyway if it's in compat. :/

Kamenuvol:
The problem is the opposite.
compat.lua loads before spells.lua, so there is no AREA_CIRCLEnXn yet to assign to AREA_CROSSnXn
We should transfer all subfolder libs to main lib folder and "dofile" them.

ranisalt:
dofile sucks, we instead should reorder which files are loaded if there are no other implications

Kamenuvol:
Compat should be the last lib to be loaded, because all previously libs are supposed to be code updated and not need it. Custom scripts, of course, are loaded after because may still use old code.

marksamman:
This should be fixed by explicitly defining the areas in the compat file.

gugahoa:
I won't be able to make the changes mentioned here until next week, but there's a better solution opened by @kamenuvol, so I'll close the PR

PR #1898, opened 2016-08-30
Simplify code and fix ambiguity

ranisalt:
My tools kept complaining `cpos.y < (min_y + offsetZ) || cpos.y > (max_y + offsetZ)` is not a valid template function call, and indeed it's ambiguous. Also there were two ifs where only one is sufficient

The if/do-while pair is doing basically what a range for does.

We should probably rename that list tho, it's confusing as f*ck: it's named `list`, being an instance of `SpectatorVec`, a specialization of `std::unordered_set`.


marksamman:
Thanks! I agree about the naming, it used to be a vector IIRC, but I changed it to unordered_set as an optimization so that it would filter out duplicates more efficiently, without renaming variables/typedefs.

Issue #1893, opened 2016-08-29
MoveEvent::onRemoveItem: Stackoverflow


TwistedScorpio:
The following code snippet crashed my server when attempting to add a _candelabrum_ item for the map (that does not decay), so that when a player moves or uses this _candelabrum_ it transforms to a decaying one.

```
<movevent event="RemoveItem" itemid="2042" script="candelabrum.lua" />
function onRemoveItem(item, tile, position)
	item:transform(2057)
	return true
end
```

The only way for it to not crash, is to use the **AddItem** function, but it will not work if the player moves the item around his containers, or from a tile to the inventory/container, rather only when it is added to a tile.

### Steps to reproduce (include any configuration/script required to reproduce)
1. Add the given moveevent script code
2. Move the item from a tile to anywhere

### Expected behaviour
The item should simply transform to the given target

### Actual behaviour
Stackoverflow Crash, more precisely: 
**[Error - MoveEvent::executeAddRemItem] Call stack overflow**

### Environment
Windows 7 x64 Ultimate
Visual Studio 2015

ranisalt:
So you're trying to transform an item that's supposed to be deleted?

TwistedScorpio:
So this script transforms a deleted trap?

https://github.com/otland/forgottenserver/blob/master/data/movements/scripts/trap.lua#L32

ranisalt:
Hmm that script checks if the item is really being moved prior to transforming it, but I doubt it's the problem.

You only get stack overflows when you reserve too much script environments (https://github.com/otland/forgottenserver/blob/4fa8f3a225d70304991555d560c56f8bd246e209/src/movement.cpp#L892), are you using TFS without any source code edits? I hit a similar problem once with wrong code.

TwistedScorpio:
TFS 1.1 branch, the function onItemMove is being called over 15 times and all with "isAdd" set to false.

TwistedScorpio:
I doubt this is normal:

![callstack](https://cloud.githubusercontent.com/assets/6238582/18042620/75ad556a-6d90-11e6-9a97-669933c2e9c5.jpg)

gugahoa:
I think I found what's happening:
1. 2042 fires onRemoveItem
2. 2042 is transformed into 2057
3. Inside Game::transformItem Tile::postRemoveNotification is caleed when 2042 is still 2042 (not transformed yet)
4. Goes to 1, because of 3

gugahoa:
[...]

PR #1900, opened 2016-08-30
Check if item is removed before transforming (#1893)

gugahoa:


marksamman:
I'm not sure this is the right approach to fix that problem, aren't there any other functions that would also cause it to crash? Or do they already check for isRemoved (in which case this PR is fine)?

gugahoa:
I checked other functions that could cause this crash, when they don't check if the item is removed in the Lua Interface, it checks on the Game method called

gugahoa:
This does nothing as Game::transformItem already check it, didn't notice until now

PR #1902, opened 2016-09-01
Monsters cleanup

raymondtfr:
A small cleanup in monster files.

I'm having some troubles with renaming some files with uppercase letters to lowercase, so I'll leave this for another PR.

### DIFF
--- a/data/monster/Bosses/abyssador.xml
+++ b/data/monster/Bosses/abyssador.xml
@@ -69,13 +69,13 @@
 		<voice sentence="*** DEATH ***" />
 	</voices>
 	<loot>
-		<item id="18496" chance="100000" /> <!-- Abyssador's Lash -->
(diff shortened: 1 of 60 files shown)


ranisalt:
I personally prefer `<immunity drunk="yes">` if available.

raymondtfr:
There's this way ``<immunity name="drunk"/>`` besides that.

But all those changes I made reflect the way most monster files are configured.

There are some other names too that don't fit, such as ``drunk`` where it should be ``drunkenness``, etc.

I accidentally pushed two different kind of changes into the same commit (about removing spaces), I think I'm gonna revert them and redo each one in a different commit to make things clearer to analyze.

ranisalt:
Whatever, commits will be squashed together anyway.

marksamman:
Thanks!

PR #1903, opened 2016-09-03
Comment out unimplemented functions

ranisalt:
Those functions are not implemented and it can be confusing if someone tries to use them just because they're available on the headers. I commented out the signatures so they are hidden for autocomplete tools and if one wants to use them, it's clear they need to be implemented.

### DIFF
--- a/src/actions.h
+++ b/src/actions.h
@@ -108,10 +108,6 @@ class Actions final : public BaseEvents
 		Event* getEvent(const std::string& nodeName) final;
 		bool registerEvent(Event* event, const pugi::xml_node& node) final;
 
-		void registerItemID(int32_t itemId, Event* event);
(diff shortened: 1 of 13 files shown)


kornholi:
Why not just delete them?

ranisalt:
@kornholi well, because I don't know for sure why they were declared.

marksamman:
I don't know why they were declared either, but they should be deleted. I'm curious how you spotted them though, I've used multiple linters/static analyzers (PVS Studio, Coverity, cppcheck) and none of them were capable of finding functions declared without a body.

ranisalt:
That was CLion bionic eyes job hahahah it has a great (but sloooow) analyzer.

Removed the functions now.

marksamman:
Thanks!

PR #1905, opened 2016-09-04
More Monsters Cleanup

raymondtfr:
More cleanup to monster files.


data/monster/Bosses/the_horned_fox.xml
```diff
@@ -39,7 +39,7 @@
 	<summons maxSummons="6">
 		<summon name="minotaur archer" interval="1000" chance="13" max="2" />
 		<summon name="minotaur guard" interval="1000" chance="13" max="2" />
```
marksamman: Looks like there are more instances of "max" in summon tags, any reason why you didn't remove all of them?

data/monster/Bosses/man_in_the_cave.xml
```diff
@@ -40,7 +40,7 @@
 		<voice sentence="A MIC to rule them all!" />
 	</voices>
 	<summons maxSummons="2">
-		<summon name="Monk" interval="2000" chance="20" max="1" />
+		<summon name="Monk" interval="2000" chance="20" />
```
ranisalt: Man in the Cave summons only 1 monk, doesn't?

marksamman:
I would prefer parsing the max value and not allowing the monster to summon more than max instances of the specific monster instead of removing it.

data/monster/Bosses/man_in_the_cave.xml
raymondtfr: Right now I'm only doing small cleanups so I'm not paying attention to these details as of now.

data/monster/Bosses/the_horned_fox.xml
raymondtfr: Not really, I just missed them. I am using basic search functions in sublime text as I don't understand much of regular expressions, so I miss some things. I'll try to find more occurrences of it.

By the way, I will be replacing melee attributes "attack" and "skill" with "min" and "max" in a few minutes.

raymondtfr:
I find that better too. But would we still need of ``maxSummons``?

marksamman:
Yes, maxSummons should continue to function as it currently does.

raymondtfr:
So should I remove the ``removed nonexistent summons attribute "max"`` commit?

marksamman:
Only if we actually implement the max feature, otherwise we might just as well remove it.

raymondtfr:
Hmm, it could take a while if I were to implement it since I am not much experienced in C++.

By the way, I have just replaced melee attributes with min, max as it is more clear.
I have found more things to fix, I'll be adding more commits soon.

marksamman:
Let me know when you feel that this is ready to be merged.

raymondtfr:
I'll do.

It seems to me that it is going to take a while though, it so seems way every time I go through the files.

ranisalt:
What was this chance1 for?

raymondtfr:
Just another alias for "chance".

marksamman:
Thanks!

PR #1908, opened 2016-09-05
Use standard variable name

ranisalt:
All but those 2 instances of lock guards are called lockClass, I refactored them to be named equally for better searchability.


marksamman:
Thanks!

PR #1910, opened 2016-09-06
Added support for summons tag "max"

raymondtfr:
This was discussed in #1905.

This code was originally from a different distribution of TFS.

I just updated it use a range-based for loop and also renamed a variable to be more clear.

I have tested it and it works fine.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -942,6 +942,17 @@ void Monster::onThinkDefense(uint32_t interval)
 				continue;
 			}
 
+			uint32_t summonCount = 0;
+			for (Creature* summon : summons) {
+				if (summon->getName() == summonBlock.name) {
+					++summonCount;
+				}
+			}
+
+			if (summonCount >= summonBlock.max) {
+				continue;
+			}
+
 			if (summonBlock.chance < static_cast<uint32_t>(uniform_random(1, 100))) {
 				continue;
 			}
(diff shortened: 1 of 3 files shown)


ranisalt:
I wouldn't use "amount" as it refers to a current value rather than a maximum value. It's even possible to use "maximum amount". "max" is fine.

marksamman:
It should also be uint32_t when parsing it, and default to the max value (or 0 which would mean no limit) so that the current behavior of not specifying max is intact.

raymondtfr:
You mean renaming both the summonBlock_t data member and the other "amount" to "max"?

ranisalt:
> You mean renaming both the summonBlock_t data member and the other "amount" to "max"?

Yep, "amount" does not have the meaning you want. "max" is the correct word here.

src/monster.cpp
```diff
@@ -942,6 +942,17 @@ void Monster::onThinkDefense(uint32_t interval)
 				continue;
 			}
 
+			uint32_t summonCount = 0;
+			for (Creature* summon : summons) {
+				if (!summon->isRemoved() && summon->getName() == summonBlock.name) {
```
marksamman: You can remove the isRemoved() check since we use summons.size() to check for maxSummons above.

marksamman:
Thanks!

PR #1913, opened 2016-09-09
Fix compiling errors

raymondtfr:
A few compiler errors/warning about narrowing conversions.

@kornholi @marksamman I think the lastest commits requires a new release of nightlies, otherwise server won't load with newer datapack changes.


marksamman:
Thanks!

PR #1915, opened 2016-09-10
Remove unique items from Browse Field

Kamenuvol:
Items with unique id should not be browseable.


mattyx14:
good

peonso:
Why?

Kamenuvol:
It is supposed to show only moveable items and containers.

ranisalt:
Thanks!

PR #1917, opened 2016-09-11
Use default constructors where possible, else simplify them

ranisalt:
This PR aims to use default constructors as much as possible, as they hint the compiler to optimize as much as possible and lead to cleaner and shorter code, as is seen on line delta. Also, some constructors can be `constexpr`d which can lead to more optimizations, notably `QTreeNode` is probably the most used one that was marked as `constexpr`.

Three major changes in classes were splitting information attributes for `Events`, `MonsterType` and `NetworkMessage` in a pimpl-like manner for cleaner `clear`/`reset` methods. In fact, now only `NetworkMessage` keeps a `reset` method and the others are not needed anymore.

Last, for some classes where the destructor was only used to delete a pointer, I changed it to unique_ptr which does the job automatically and implies no overhead.

There might be some non-consistent default attributes and unrelated changes because I've been working on this commit for a few weeks now, please point out anything strange.

### DIFF
--- a/src/combat.h
+++ b/src/combat.h
@@ -73,7 +73,7 @@ struct CombatParams {
 	CombatType_t combatType = COMBAT_NONE;
 	CombatOrigin origin = ORIGIN_SPELL;
 
-	uint8_t impactEffect = CONST_ANI_NONE;
+	uint8_t impactEffect = CONST_ME_NONE;
 	uint8_t distanceEffect = CONST_ANI_NONE;
 
 	bool blockedByArmor = false;
@@ -88,13 +88,7 @@ typedef void (*COMBATFUNC)(Creature*, Creature*, const CombatParams&, CombatDama
 class MatrixArea
 {
 	public:
-		MatrixArea(uint32_t rows, uint32_t cols) {
-			centerX = 0;
-			centerY = 0;
-
-			this->rows = rows;
-			this->cols = cols;
-
+		MatrixArea(uint32_t rows, uint32_t cols): centerX(0), centerY(0), rows(rows), cols(cols) {
 			data_ = new bool*[rows];
 
 			for (uint32_t row = 0; row < rows; ++row) {
@@ -335,9 +329,7 @@ class Combat
 class MagicField final : public Item
 {
 	public:
-		explicit MagicField(uint16_t type) : Item(type) {
-			createTime = OTSYS_TIME();
-		}
+		explicit MagicField(uint16_t type) : Item(type), createTime(OTSYS_TIME()) {}
 
 		MagicField* getMagicField() final {
 			return this;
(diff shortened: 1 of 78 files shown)


src/networkmessage.h
```diff
@@ -149,28 +146,31 @@ class NetworkMessage
...
-		bool overrun;
+		struct NetworkMessageInfo {
+			MsgSize_t length = 0;
+			MsgSize_t position = INITIAL_BUFFER_POSITION;
+			bool overrun = false;
+		};
 
-		uint8_t buffer[NETWORKMESSAGE_MAXSIZE];
+		NetworkMessageInfo info;
+		uint8_t buffer[NETWORKMESSAGE_MAXSIZE] = {};
```
djarek: I don't think we need to zero-fill the buffer before use.

src/networkmessage.h
```diff
@@ -43,14 +43,11 @@ class NetworkMessage
...
-			reset();
-		}
+		constexpr NetworkMessage() = default;
 
 		void reset() {
-			overrun = false;
-			length = 0;
-			position = INITIAL_BUFFER_POSITION;
+			NetworkMessageInfo other;
+			std::swap(info, other);
```
djarek: I think it would be clearer what's you're doing here if you unstead used this:
```
info = {};
```

src/combat.h
```diff
@@ -329,15 +323,13 @@ class Combat
 		double maxa = 0.0;
 		double maxb = 0.0;
 
-		std::unique_ptr<AreaCombat> area;
+		std::unique_ptr<AreaCombat> area = nullptr;
```
djarek: Redundant, unique-ptr default constructor already does this.

src/networkmessage.h
ranisalt: Yeah, I actually did it on another places, thanks

src/networkmessage.h
ranisalt: Probably not because it keeps track of what was already written

ranisalt: Actually, for the constexpr specifier to work it needs to initialize all attributes

kornholi:
I don't see how putting constexpr everywhere gives us anything

ranisalt:
Currently not much. Once moving forward to C++14 and beyond, we can benefit from [constant initialization](http://en.cppreference.com/w/cpp/language/constant_initialization).

src/networkmessage.h
marksamman: I'm not sure what kind of optimizations constexpr constructors bring, but I can imagine zero filling the buffer every time we construct a NetworkMessage object is more costly.

ranisalt: Hmm, I thought they were static initialized, but it seems lots are instanced on runtime, nulling any advantages. I'll revert those on *Message.

marksamman:
Thanks!

PR #1920, opened 2016-09-15
Monster update (4)

raymondtfr:
This one is going to take a while...


ranisalt:
I think monster folders should be lowercase too.

marksamman:
Thanks!

raymondtfr:
@marksamman Something went wrong with this. Now there is a duplicate file for each monster in each folder. Could you please revert this?

marksamman:
Reverted.

raymondtfr:
I guess this was a problem I caused by changing the folder names to lowercase (through the Git Shell) and somehow the files inside were duplicated.

So I'll just stick with updating the files and I'll let someone do that if this is still a welcome change.
Thanks!

gugahoa:
When renaming files use `git mv`

raymondtfr:
I think I did that, or not, can't quite remember now, anyway there's that thing about Linux and path names mentioned in #1936 so I don't think this is a good change anymore?! (unfortunately, I'm only a Windows user)

PR #1921, opened 2016-09-16
fix group cooldowns of spell sharpshooter

raymondtfr:
Currently, it is:
current

But it should be:
fixed *(both support/healing group cooldowns (10 s) ends at the same time)*

### DIFF
--- a/data/spells/scripts/support/sharpshooter.lua
+++ b/data/spells/scripts/support/sharpshooter.lua
@@ -14,16 +14,6 @@ speed:setParameter(CONDITION_PARAM_TICKS, 10000)
 speed:setFormula(-0.7, 56, -0.7, 56)
 combat:setCondition(speed)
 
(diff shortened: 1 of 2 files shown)


marksamman:
Thanks!

Kamenuvol:
The problem is not fixed. What happens is that spells.xml cooldown overwrote the scripted cooldown.
In my opinion, the script should overwrite anything from xml. I will see that.

raymondtfr:
@Kamenuvol 

> problem is not fixed

It is fixed actually, it doesn't really need to be written in a script.

> only workarounded. What happens is that spells.xml cooldown was overwriting the scripted cooldown.
In my opinion, the script should overwrite anything from xml. I will see that.

Yes, I have noticed the conditions for **Support** cooldown in the script didn't override that of spells.xml... @marksamman This might be a bug?

PR #1922, opened 2016-09-19
fix umbral weapons

raymondtfr:


kornholi:
Thanks a lot!

wgrriffel:
@raymondtfr would be good have an update to spellsbooks, only old ones are declared in actions.

PR #1795, opened 2016-05-16
Better modifiable potions.lua

slavidodo:
Explanation of all features are included.

PR #1927, opened 2016-09-25
Update potions script

raymondtfr:
I have updated potions script to look more cleaner, added buff potions, etc.

Aside from that, I removed the ``CONDITION_EXAUST_HEAL" because as it is stated in the source code (enum.h) it not used anymore (also not used in vanilla, though there is still a delay, around 1-2s), atleast it doesn't work anymore (getCondition always return nil/false).

I have also added 3 buff potions.
Also using a potion on another player is now permitted (as in vanilla) if his requirements (the player's) are met (vocation, level).

I tried to do a simpler and maybe cleaner version of #1795.

### DIFF
--- a/data/actions/scripts/other/potions.lua
+++ b/data/actions/scripts/other/potions.lua
@@ -1,158 +1,113 @@
-local ultimateHealthPot = 8473
-local greatHealthPot = 7591
-local greatManaPot = 7590
-local greatSpiritPot = 8472
-local strongHealthPot = 7588
-local strongManaPot = 7589
-local healthPot = 7618
-local manaPot = 7620
-local smallHealthPot = 8704
-local antidotePot = 8474
-local greatEmptyPot = 7635
-local strongEmptyPot = 7634
-local emptyPot = 7636
+local berserk = Condition(CONDITION_ATTRIBUTES)
+berserk:setParameter(CONDITION_PARAM_SUBID, 7)
+berserk:setParameter(CONDITION_PARAM_TICKS, 10 * 60 * 1000)
+berserk:setParameter(CONDITION_PARAM_SKILL_MELEE, 5)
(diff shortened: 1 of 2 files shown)


data/actions/scripts/other/potions.lua
```diff
@@ -1,158 +1,110 @@
...
+			description = "Only knights may drink this potion."},
+
+	[7440] = {condition = mastermind, vocations = {1, 2, 5, 6}, effect = CONST_ME_MAGIC_BLUE,
+			text = "Only sorcerers and druids may drink this potion.",
+			description = "You feel smarter."},
+
+	[7443] = {condition = bullseye, vocations = {3, 7}, effect = CONST_ME_MAGIC_GREEN,
+			text = "Only paladins may drink this potion.",
+			description = "You feel more accurate."},
+
```
Kamenuvol: I think text and description are inverted in 7440 and 7443

raymondtfr: Actually not. The item id 7443 is for bullseye potion (paladins) and 7440 is for mastermind potion (sorcerers and druids).

Kamenuvol: I mean the text and description are inverted in each one, not the items.

raymondtfr: Yeah, I misunderstood you, it is fixed now.

Kamenuvol:
You should add new potions from 10.92 (also in items.xml):
Ultimate Mana Potion
Ultimate Spirit Potion
Supreme Health Potion

raymondtfr:
I just updated ``potions.lua`` and ``actions.xml`` and I'll leave ``items.xml`` as it is because it might be updated by someone.

marksamman:
Thanks!

PR #1928, opened 2016-09-26
Add Redis Support

elderapo:
At the beginning I would like to mention that I am C++ newbie so if it's possible, please point out all mistakes I made with possible explanations.

One of Roadmap goals is to make **'Built-in HTTP, JSON API using Lua scripts to get real-time data from the server'**.

I decided to go one step further and added Redis support for TFS which will allow communication TFS <-> nodes/PHP/basically anything through Redis in real time. It will give us opportunity to make new modern acc makers and other cool things.

At this moment it uses https://github.com/nekipelov/redisclient but it might be good to drop that dependency in the future.

### DIFF
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -1239,6 +1239,12 @@ class LuaScriptInterface
 		static int luaPartyShareExperience(lua_State* L);
 		static int luaPartySetSharedExperience(lua_State* L);
 
+		// Redis
+		static int luaRedisGet(lua_State* L);
+		static int luaRedisSet(lua_State* L);
+		static int luaRedisPublish(lua_State* L);
+		static int luaRedisSubscribe(lua_State * L);
+
 		//
 		lua_State* luaState = nullptr;
 		std::string lastLuaError;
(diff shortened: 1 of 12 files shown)


src/luascript.h
```diff
@@ -1258,46 +1264,46 @@ class LuaScriptInterface
...
-		uint32_t lastCombatId = 0;
-		uint32_t lastAreaId = 0;
+	uint32_t lastEventTimerId = 1;
+	uint32_t lastCombatId = 0;
+	uint32_t lastAreaId = 0;
 
-		friend class LuaScriptInterface;
-		friend class CombatSpell;
+	friend class LuaScriptInterface;
+	friend class CombatSpell;
```
ranisalt: You messed with the indentation of this entire class without the need.

ranisalt:
You didn't add redis dev package to the build script, both Travis and CMake.

I'm not sure this is what the roadmap suggests. This is overkill and adds an external dependency to the server.

elderapo:
@ranisalt As I said I am new to C++ and I don't know how to add things to Travis and CMake. Mind helping me a little bit with that?

src/luascript.h
elderapo: Fixed it.

kornholi:
I don't think this is the right approach. How do you want to use this with TFS? We were thinking something more like gRPC (especially if we want to support 11.x in the future).

yrpen:
I don't care about windows, but:
```
The Redis project does not officially support Windows.
```

HeavenIsLost:
So i can use this to query any data from my server? nice :+1:

elderapo:
@kornholi what do you mean because I am already using it with TFS. I just don't know if it's 100% thread safe since it's using io_service in a "hacky" way I would say.

@yrpen https://github.com/nekipelov/redisclient it does support windows (I only used it on windows).

@HeavenIsLost it allows you to send events to Redis and subsribe to Redis events. If you are familiar with nodejs it will allow you to make real time accmakers etc...

src/otserv.cpp
```diff
@@ -38,10 +38,12 @@
 #include "databasemanager.h"
 #include "scheduler.h"
 #include "databasetasks.h"
+#include "redis.h"
 
 DatabaseTasks g_databaseTasks;
 Dispatcher g_dispatcher;
 Scheduler g_scheduler;
+Redis *g_redis = nullptr;
```
djarek: Naked owning pointers are bad, even if they're globals.

src/redis.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+void Redis::connectPublisher()
+{
+	publisher.connect(address, port, [=](bool status, std::string err) {
+		if (status) {
+			publisherConnected = true;
+		}
+	});
+
+	ioService.run_one();
```
djarek: You completely messed up the concurrency. run_one() will cause a completion handler to be run in the context of this thread, which I believe is the dispatcher thread.

src/otserv.cpp
elderapo: I just looked that up from https://github.com/otland/forgottenserver/blob/master/src/scriptmanager.cpp#L33-L41

src/redis.cpp
elderapo: These two run_one calls are required. Otherwise subscriber and publisher connections won't get estabilished before loading lua scripts that use publish/subscribe method from Redis class and that will cause a crash.

marksamman:
Thanks for your contribution. There doesn't seem to be any progress on addressing the issues that have been raised here, and using Redis is not how we were planning on doing this. #2010 is on the right track with this, so I'm closing this pull request in favor of that.

PR #1930, opened 2016-10-01
Update music script

raymondtfr:


data/actions/scripts/other/music.lua
```diff
@@ -1,5 +1,73 @@
...
+	[6573] = {effect = CONST_ME_SOUND_GREEN, text = "TOOOOOOT", transformId = 13578}, -- party trumpet
+	[13759] = {effect = CONST_ME_SOUND_BLUE} -- small whistle (actual effect is unknown)
+}
+
+local effect, chance = nil, nil
+
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
-	-- TODO: Different music effect for different instruments.
-	item:getPosition():sendMagicEffect(CONST_ME_SOUND_BLUE)
+	local instrument = instruments[item:getId()]
```
vankk: Isn't better use? 

`local instrument = instruments[item:getId()]
if not instrument then
	return true
end`

raymondtfr: No, I don't think. As far as you don't add the wrong itemid in actions.xml, which would still be better anyway as it would only help you debugging.

vankk: I saw in a few scripts on _data/actions/other_ that use the param that I said.

kornholi:
Thanks!

PR #1933, opened 2016-10-01
avoid debug assertions from specific param ids

raymondtfr:
Executing the command `/i` with any of these number ids will cause debug assertion to all connected clients, and also to any new login try, forcing the user to restart the server.

### DIFF
--- a/data/talkactions/scripts/create_item.lua
+++ b/data/talkactions/scripts/create_item.lua
@@ -1,3 +1,5 @@
+local debugAssertionIds = {1, 2, 3, 4, 5, 6, 7, 10, 11, 13, 14, 15, 19, 21, 26, 27, 28, 35, 43}
+
 function onSay(player, words, param)
 	if not player:getGroup():getAccess() then
 		return true
@@ -16,6 +18,10 @@ function onSay(player, words, param)
 			player:sendCancelMessage("There is no item with that id or name.")
 			return false
 		end
+	end
+
+	if isInArray(debugAssertionIds, itemType:getId()) then
+		return false
 	end
 
 	local count = tonumber(split[2])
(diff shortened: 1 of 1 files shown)


data/talkactions/scripts/create_item.lua
```diff
@@ -48,5 +54,5 @@ function onSay(player, words, param)
 		end
 		player:getPosition():sendMagicEffect(CONST_ME_MAGIC_GREEN)
 	end
-	return false
+	return true
```
ranisalt: Why?

raymondtfr:
I think its a logical/aesthetics change, I mean we should return the boolean true by default if to signal we had found no obstacles in the code, otherwise return false to signal we faced some breaking (or not) obstacle ending the function call prematurely.

But as I have forgotten to change one more instance in line 4, one of the maintainers or I code revert this change. This also happens in more files.

data/talkactions/scripts/create_item.lua
raymondtfr: I think its a logical/aesthetics change, I mean we should return the boolean true by default if to signal we had found no obstacles in the code, otherwise return false to signal we faced some breaking (or not) obstacle ending the function call prematurely.

But as I have forgotten to change one more instance in line 4, one of the maintainers or I code revert this change. This also happens in more files.

ranisalt: Talkactions return `false` when the engine is not supposed to look further (e.g. try to find spells). They return `true` when the talkaction is unavailable, rather than failed, so that the game next tries to find a spell with the same words. That's why you can't do this change right now.

ranisalt:
Please rename the table `invalidIds` and break lines to keep consistent with other talkactions (see `looktype.lua`)

ranisalt:
Thanks!

PR #1935, opened 2016-10-02
Update game protocol to 10.98

marksamman:


PrinterLUA:
👍

HeavenIsLost:
I don't belive the game store belongs to tfs. @marksamman can you create a small tutorial of items.otb update?

joseluis2g:
ingame store should be handled by sources imo

gugahoa:
First of all, how does the in-game store works?
It would be for the best, if it's possible, to make it in Lua since entry-level to Lua is pretty low and the shop is a feature most people would like to customize.

raymondtfr:
Guys, the title refers to an update to the game protocol, it does not mention ingame store 🎉 , and since it is a WIP...

If it were to be implemented in C++ (safer?), it could have funcions in the Lua API, hence providing customization possibilities...

marksamman:
I don't know enough about the in-game store to tell if it's a good idea to add Lua callbacks to it, but either way this pull request is just meant to add basic protocol compatibility so we can push out the 1.2 release and move on to 1.3.

@HeavenIsLost You need to create a plugin compatible with the latest dat file for OTItemEditor, and then use the Update feature. It's not straight-forward as it depends on how the dat file has changed. That doesn't necessarily mean that it will match the official one though, e.g. sometimes old items are changed in an undesired way and we choose to keep the old attributes on the old item and create a new one with the new attributes. Sometimes we choose to persist new attributes in the OTB. Creating your own OTB is discouraged because it would split the community if you release it and it's used in the wild. You should only mess with the OTB if you have a custom client.

Schiffers:
and what about the critical hit system ?

HeavenIsLost:
@Schiffers almost every 10.9+ server got this system, its not good enough?

joseluis2g:
@HeavenIsLost I got a version made by gesior and almost working, I think is the best one, maybe can share with you guys.

HeavenIsLost:
@joseluis2g plenty of servers already are advertising that they have mana leech/life leech/critical hit items.

Schiffers:
well, this servers that have it don't share with the community the codes '-'

joseluis2g:
@HeavenIsLost 
I suppose that most of them just leech things and have a very bugged version so idm to me, I always wanted a os community but its really hard

brunominervino:
@marksamman 
In ProtocolGame::sendBasicData()

Change:
```cpp
void ProtocolGame::sendBasicData()
{
	NetworkMessage msg;
	msg.addByte(0x9F);
	msg.addByte(player->isPremium() ? 0x01 : 0x00);
	msg.add<uint32_t>(std::numeric_limits<uint32_t>::max());
	msg.addByte(player->getVocation()->getClientId());
	msg.add<uint16_t>(0x00);
	writeToOutputBuffer(msg);
}
``` 

To:
```cpp
void ProtocolGame::sendBasicData()
{
	NetworkMessage msg;
	msg.addByte(0x9F);
	msg.addByte(player->isPremium() ? 0x01 : 0x00);
	msg.add<uint32_t>(player->isPremium() ? time(nullptr) + (player->premiumDays * 86400) : 0);
	msg.addByte(player->getVocation()->getClientId());
	msg.add<uint16_t>(0x00);
	writeToOutputBuffer(msg);
}
``` 

This will fix premium time when player pressing ctrl+g.

src/protocolgame.cpp
```diff
@@ -2841,6 +2849,9 @@ void ProtocolGame::AddPlayerStats(NetworkMessage& msg)
 	msg.add<uint16_t>(condition ? condition->getTicks() / 1000 : 0x00);
 
 	msg.add<uint16_t>(player->getOfflineTrainingTime() / 60 / 1000);
+
+	msg.add<uint16_t>(0); // xp boost time (seconds)
+	msg.addByte(0); // ?
```
TheSumm: This byte indicates whether the player may buy further experience boosts in the store.
It will enable/disable the "Exp Boost" button in the skill window.

joseluis2g: @TheSumm can you explain me a lil bit? iam a rookie on this
thanks Samuel

comedinha: CanBuyMoreStoreXpBoosts

https://github.com/comedinha/Flash-News/blob/7eb2f71c6f26580ea0625c89a802efd8f319928b/scripts/tibia/creatures/ExperienceGainInfo.as#L57

PR #1938, opened 2016-10-03
Monster updates

raymondtfr:
This PR contains all fixes (and a few more) that I made to monster experiences & hitpoints (from #1920) and also the deletion of duplicate entries in monster.xml/file.

Definitely not on my days, haah, going to fix this.
-- Edit
Everything's good now.


marksamman:
Thanks!

PR #1940, opened 2016-10-05
add damage decrement range to generateDamageList

raymondtfr:
Now condition damages won't take too long to disappear.

By the way, the decrement values generated by this code is probably the same as vanilla (for all `i` values).

### DIFF
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -1202,7 +1202,7 @@ void ConditionDamage::generateDamageList(int32_t amount, int32_t start, std::lis
 	int32_t sum = 0;
 	double x1, x2;
 
-	for (int32_t i = start; i > 0; --i) {
+	for (int32_t i = start; i > 0; i = i - (i > 21 ? (i / 20) + boolean_random() : 1)) {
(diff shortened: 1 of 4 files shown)


ranisalt:
A while loop would better fit. I find complex for conditions/statements ugly and misleading.

raymondtfr:
Initially I thought about using a while loop too. So is it better if I replace it with a while?

Oh, I think I'm also going to update the time interval for such conditions (it should be 4 s, I guess it is 4 s atm).

I could also add support to generate damage lists for condition CURSED, so that people could use it on monster files...

raymondtfr:
Closing this since it would take a lot more changes for it to work as expected (using the same code when ``amount != 0`` in ``ConditionDamage::Init``, *condition params: min, max and start*, etc.) and it would/could modify some scripts that depended on current behavior.

I may update this if compatibility with this kind of change should not be a concern.

I'll open another PR just to update tickIntervals of some damage conditions. In the future, I'll update spells lib function ``Creature:addDamageCondition`` for ``listType == 1`` with a formula.

PR #1943, opened 2016-10-08
PropStream and PropWriteStream to Lua.

Mkalo:
This will help to do serialization and deserialization in Lua.

I implemented 2 functions:
Game.serializeItem and Game.unserializeItem
(Basically load and save item from IOMapSerialize with a couple of changes)

And the Item.serializeAttr and Item.unserializeAttr.

Here is a small example of how you could use it to store items in the database:
http://pastebin.com/Hk4iqwaK
Another example:
http://pastebin.com/Yk9CJYqh

Still a work in progress, advices are welcome.


djarek:
This would become obsolete with #2139.

PR #1944, opened 2016-10-08
Tile:addItem and Tile:addItemEx

Mkalo:
Removed doTileAddItemEx from the source and placed in compat.lua


data/lib/compat/compat.lua
```diff
@@ -1006,3 +1006,13 @@ end
 function Guild.removeMember(self, player)
 	return player:getGuild() == self and player:setGuild(nil)
 end
+
+function doTileAddItemEx(pos, uid, flags)
+	flags = flags or 0
```
ranisalt: You don't need to set default to `0` because you already do it on `getNumber`.

data/lib/compat/compat.lua
```diff
@@ -1006,3 +1006,13 @@ end
...
+	local tile = Tile(pos)
+	local item = Item(uid)
+	if tile and item then
+		tile:addItemEx(item, flags)
```
ranisalt: Isn't `return true` missing here?

data/lib/compat/compat.lua
```diff
@@ -1006,3 +1006,13 @@ end
...
+	end
+	return false
```
ranisalt: It's far more common to invert the conditions and fail fast. You'd usually do:

```lua
local tile = Tile(pos)
if not tile then
    return false
end

local item = Item(uid)
if not item then
    return false
end

-- do what happens when everything is ok
return true
```

And you don't need to check if `item` is not null because you already check on C++ (`getUserdata` then checking if it's not null).

src/luascript.cpp
```diff
@@ -5208,6 +5175,77 @@ int LuaScriptInterface::luaTileQueryAdd(lua_State* L)
...
+	uint16_t itemId;
+	if (isNumber(L, 2)) {
+		itemId = getNumber<uint16_t>(L, 2);
+	} else {
+		itemId = Item::items.getItemIdByName(getString(L, 2));
+		if (itemId == 0) {
+			lua_pushnil(L);
+			return 1;
+		}
+	}
```
ranisalt: It should definitely take Item userdata as parameter instead of item id.

src/luascript.cpp
```diff
@@ -5208,6 +5175,77 @@ int LuaScriptInterface::luaTileQueryAdd(lua_State* L)
...
+		delete item;
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+int LuaScriptInterface::luaTileAddItemEx(lua_State* L)
+{
+	// tile:addItemEx(item[, flags = 0])
+	Item* item = getUserdata<Item>(L, 2);
```
ranisalt: Why did you invert? First get userdata at index 1, then 2, etc.

src/luascript.cpp
Mkalo: Why? It would entirely miss the purpose of having addItem and addItemEx function. Look at all other addItem and addItemEx functions.

The "ex" function should take an item userdata of a virtual item, the normal function takes the id and creates the item.

src/luascript.cpp
Mkalo: Basically following the pattern:
 https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L6485
https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L8498

data/lib/compat/compat.lua
Mkalo: I need to check if item is not nil, calling the API just to check if something is waste.

data/lib/compat/compat.lua
Mkalo: Yep

data/lib/compat/compat.lua
Mkalo: Sure

data/lib/compat/compat.lua
ranisalt: But it will be checked anyway by the target function, it's not like you do it in Lua so it doesn't happen in C++, that's what I mean.

src/luascript.cpp
ranisalt: That's a very odd pattern actually hahahah

src/luascript.cpp
ranisalt: I don't think it makes sense to have a function to add by ID and another that adds by userdata. If you have the item ID just create an instance in Lua and have a single `addItem` function that receives userdata instead. Or am I understanding wrong?

PR #1945, opened 2016-10-10
Use constexpr for compile time constants and defines

ranisalt:
`#define` macros for constant values usually mask errors, e.g. if you accidentally type a macro name in the middle of the code (I once run into this problem with NodeJS that has a `#define THIS`...). Also, it's a preprocessor directive and not exactly a language feature.

`constexpr` instead of just `const` is better explained [here](https://stackoverflow.com/questions/13346879/const-vs-constexpr-on-variables).

(crossing fingers hoping clang behaves properly)


marksamman:
Thanks!

PR #1950, opened 2016-10-12
fixed debug when more then 7 players are on tile

gunzino:
Problem explantation:
If player tries to login and the tile has more than 7-8 players on it, player got debug. The logging player should be always part of the tile after logging in game. 

I've experienced this issue since TFS 1.0. After some investigation I managed to fix.

Please correct my typo.

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -516,20 +516,53 @@ void ProtocolGame::GetTileDescription(const Tile* tile, NetworkMessage& msg)
 		}
 	}
 
-	const CreatureVector* creatures = tile->getCreatures();
-	if (creatures) {
-		for (const Creature* creature : boost::adaptors::reverse(*creatures)) {
-			if (!player->canSeeCreature(creature)) {
-				continue;
+	if (tileLogin && tile->getPosition() == player->getPosition()) {
+		bool playerSpawned = false;
+		const CreatureVector *creatures = tile->getCreatures();
+		if (creatures) {
+			for (const Creature *creature : boost::adaptors::reverse(*creatures)) {
+				if (!player->canSeeCreature(creature)) {
+					continue;
+				}
+
+				if (creature == player) {
+					playerSpawned = true;
+				}
+
+				bool known;
+				uint32_t removedKnown;
+				checkCreatureAsKnown(creature->getID(), known, removedKnown);
+				AddCreature(msg, creature, known, removedKnown);
+
+				if (count == 8 && playerSpawned == false) {
+					bool known;
+					uint32_t removedKnown;
+					checkCreatureAsKnown(player->getID(), known, removedKnown);
+					AddCreature(msg, player, known, removedKnown);
+					++count;
+				}
+
+				if (++count == 10) {
+					return;
+				}
 			}
+		}
+	} else {
+		const CreatureVector *creatures = tile->getCreatures();
+		if (creatures) {
+			for (const Creature *creature : boost::adaptors::reverse(*creatures)) {
+				if (!player->canSeeCreature(creature)) {
+					continue;
+				}
 
-			bool known;
-			uint32_t removedKnown;
-			checkCreatureAsKnown(creature->getID(), known, removedKnown);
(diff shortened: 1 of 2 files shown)


src/protocolgame.cpp
```diff
@@ -516,20 +516,51 @@ void ProtocolGame::GetTileDescription(const Tile* tile, NetworkMessage& msg)
...
+				if (creature == player) {
+					playerSpawned = true;
+				}
+
+				bool known;
+				uint32_t removedKnown;
+				checkCreatureAsKnown(creature->getID(), known, removedKnown);
+				AddCreature(msg, creature, known, removedKnown);
+
+				if (count == 8 && playerSpawned == false) { // player still not spawned and we need to send him too
```
djarek: Can count be equal to 9 at the first iteration?

src/protocolgame.h
```diff
@@ -320,6 +320,7 @@ class ProtocolGame final : public Protocol
 
 		bool debugAssertSent = false;
 		bool acceptPackets = false;
+		bool loggedIn = false;
```
djarek: Is this unused?

nekiro: @djarek 
`if (!loggedIn && tile->getPosition() == player->getPosition()) {`

src/protocolgame.cpp
nekiro: !playerSpawned instead of playerSpawned == false?

DSpeichert: Technically it could if there are 8 tile items on the ground and the ground. However, the tile "items" that I checked don't quite stack that much in any practical way - you can probably only stack a few types of items on each other (ground, second ground type for half-tile, grass, different grass come up to 4 which is max what I found out).

Creating 8 "counters" (like in the shop) makes your character disappear if you `/a 1` onto it in game and the client does not crash. However, when you try to log in to such position, the server would put you around it as the tile is not walkable. Therefore, I haven't found a way to stack that many walkable tile items to go over 8.

Now, if we do go over 8, then the character won't be sent to the client and it will MOST LIKELY crash due to the same issue this PR is supposed to fix. If we change this line to `count >= 8`, it will probably still crash due to too many things on a tile.

This PR really addresses the issue of multiple creatures on one tile, which seems to be the most probably reason for the current issue of client crash.

Thoughts @djarek ?

DSpeichert:
How about:
```diff
diff --git a/src/protocolgame.cpp b/src/protocolgame.cpp
index 7edfc519..2455df1f 100644
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -517,7 +517,10 @@ void ProtocolGame::GetTileDescription(const Tile* tile, NetworkMessage& msg)
                for (auto it = items->getBeginTopItem(), end = items->getEndTopItem(); it != end; ++it) {
                        msg.addItem(*it);

-                       if (++count == 10) {
+                       count++;
+                       if (tile->getPosition() == player->getPosition() && count == 8) {
+                               break; // ensure we have a chance to add the player
+                       } else if (count == 9) {
                                return;
                        }
                }
@@ -525,13 +528,23 @@ void ProtocolGame::GetTileDescription(const Tile* tile, NetworkMessage& msg)

        const CreatureVector* creatures = tile->getCreatures();
        if (creatures) {
+               bool known;
+               uint32_t removedKnown;
+
+               if (tile->getPosition() == player->getPosition()) {
+                       checkCreatureAsKnown(player->getID(), known, removedKnown);
+                       AddCreature(msg, player, known, removedKnown);
+
+                       if (++count == 10) {
+                               return;
+                       }
+               }
+
                for (const Creature* creature : boost::adaptors::reverse(*creatures)) {
-                       if (!player->canSeeCreature(creature)) {
+                       if (!player->canSeeCreature(creature) || creature->getID() == player->getID()) {
                                continue;
                        }

-                       bool known;
-                       uint32_t removedKnown;
                        checkCreatureAsKnown(creature->getID(), known, removedKnown);
                        AddCreature(msg, creature, known, removedKnown);
```

This will always add the player on top of creatures. Not sure if that matters.

PR #1953, opened 2016-10-15
Replace arrays with maps for search

ranisalt:
Replaces some arrays that map names to enum values (and one vice-versa) to drop search complexity from O(n) to O(1) and reduce a lot of code since structs are not needed.

### DIFF
--- a/src/items.cpp
+++ b/src/items.cpp
@@ -564,19 +564,19 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 				std::cout << "[Warning - Items::parseItemNode] Unknown slotType: " << valueAttribute.as_string() << std::endl;
 			}
 		} else if (tmpStrValue == "ammotype") {
-			it.ammoType = getAmmoType(valueAttribute.as_string());
+			it.ammoType = getAmmoType(asLowerCaseString(valueAttribute.as_string()));
 			if (it.ammoType == AMMO_NONE) {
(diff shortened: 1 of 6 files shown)


gugahoa:
Tab size is mismatched with the code base tab size

ranisalt:
Oops, fixed.

marksamman:
Your changes make the functions case sensitive, and I'm not sure how much faster they'll actually be in practice (may be slower in many cases), especially after you have to transform the strings into lowercase. strcasecmp is incredibly fast on the platforms I've tested it on, so much that we've kept it despite its portability issues.

I'm not opposed to merging this though if you address the case sensitivity, but it'll have to wait until after the 1.2 release.

gugahoa:
As @marksamman said, I don't know if this will be faster simple because of hashing. From my experience, array works a lot faster even thought it's O(n), but this is coming from my experience using array vs map in a backtrack algorithm so the experience may not be applicable here

ranisalt:
I see how the array might not be slower here, as the best structure would be a trie, but STL does not provide one.

I can make it for the case insensitiveness, but is this really necessary?

marksamman:
If you can be certain that the input is always in lowercase (e.g. "user input" from monster files etc. is already transformed to lowercase before these functions are called), then no, it's not necessary. Otherwise it is, because we rather not break things that were working before.

As for the performance, it's whatever, the functions aren't called that frequently anyway, which is why arrays were already fine. A trie is overkill IMO, but if we were to use a trie, then we should rework WildcardTree so that it can be used for that too.

marksamman:
Thanks!

PR #1957, opened 2016-10-20
fix seek in container to be able to move back after changing page

gunzino:
If you have more pages in Inbox or you use global depot and change panges than it should be possible to go back to previous container too.


marksamman:
Thanks!

PR #1961, opened 2016-10-24
use new lua lib function to add curse condition

raymondtfr:
Use new lua function from spells lib in monster spells to add Curse conditions.


marksamman:
Thanks!

PR #1963, opened 2016-10-24
Update schema.sql : secret column to account table

Znote:
Right now the secret column is generated on server bootup, however this can be bothersome if you set up the server and the AAC before starting the server.


marksamman:
You should change the db_version to 19 in the schema as well.

marksamman:
Thanks!

PR #1970, opened 2016-10-31
Fix #1851

### LINKED ISSUE #1851: Open loot
Have two corpses dead.

Open loot.. and open loot first death corpse.

kornholi:
Can someone test this?

Fixes #1851

### DIFF
--- a/src/tile.cpp
+++ b/src/tile.cpp
@@ -1606,7 +1606,7 @@ Item* Tile::getUseItem() const
 		return ground;
 	}
 
-	for (Item* item : boost::adaptors::reverse(*items)) {
+	for (Item* item : *items) {
 		if (Item::items[item->getID()].forceUse) {
 			return item;
 		}


wgrriffel:
Im not sure what purpose of force use flag, i assume thats for items that can be used even not being at top of a stack, without the reverse list when a corpse drop on a ladder stack it will get stuck until it decays, maybe should forceUse flag be removed from these corpses.

Kamenuvol:
In my opinion, forceUse flag in corpses is used to get fluids from them even if there is something on it, pretty useless.

HeavenIsLost:
@kornholi 
```
	for (Item* item : boost::adaptors::reverse(*items)) {
		//if the behavior is wrong remove && !Item::items[item->getID()].isContainer()
		if (Item::items[item->getID()].forceUse && !Item::items[item->getID()].isContainer()) {
			return item;
		}
	}
```

This is my fix for this problem.

brunominervino:
@HeavenIsLost fix works fine.

DSpeichert:
Wouldn't this fix now force you to open the corpse instead of walking up a ladder?
(Just thinking out loud, haven't actually tested.)

DSpeichert:
Fixed in #2393

PR #1971, opened 2016-11-01
Validate tokens when connecting to gameworld (Fixes #1965)

### LINKED ISSUE #1965: 2FA Security hole

1. You can connect directly to protocolgame using accuont name password (there are more ways to do that)
2. Editing OTClient
3. Route IP Address to loopback, login to local TFS, remove route, connect to remote server without need of 2FA

### Expected behaviour
ProtocolGame should have finally session key generatde by protocollogin that is real session key (random string), not based on acc name & password. There may be more session keys stored in database (and deleted after user changes password or smth..)

### Actual behaviour
Session key is generated by acc name & password, therefore 2FA is useless after connecting directly to protocol game.

kornholi:
Not ideal, but better than not checking at all.

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -117,15 +117,27 @@ bool IOLoginData::loginserverAuthentication(const std::string& name, const std::
 	return true;
 }
 
-uint32_t IOLoginData::gameworldAuthentication(const std::string& accountName, const std::string& password, std::string& characterName)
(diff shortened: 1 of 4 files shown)


gunzino:
This is very restrictive to players since they always have to retype account name, password after they want to change character. We simply need session keys management :/

kornholi:
did you actually test it?

ranisalt:
I think the correct approach is to generate a random string for session key, store it in an account->key map and check it. Non-random session keys are not really session keys.

gunzino:
Ohh sorry, I've just saw you pass the ticks too.

marksamman:
Thanks!

PR #1973, opened 2016-11-05
Moving Questlog to Lua

Mkalo:
Adds more flexibility to questlogs, as you can now use callbacks that take a player object and return a string. You can add your quests using 
Game.addQuest(quest) in any script, so you can have it more organized.

You can convert your quests.xml to table format using this code (It will output the table Quests to quest.lua converting xml string located in the script):
http://pastebin.com/CqfThQfV

Here is an example inside an action:
http://pastebin.com/CQD9NWD8
Note the usage of callback function there.

If you have any ideas to add to this feel free to comment.


marksamman:
Lua should be added as a complement to quests.xml, it shouldn't replace quests.xml completely. Backwards compatibility is important to make upgrading to newer versions seamless, otherwise we'll have people stick with older versions and fragment the community. Quests should still be registered in quests.xml, but should be able to reference script files with callbacks that return missions and mission descriptions.

HeavenIsLost:
@marksamman i think its easy to write a tool(i did it to my own lua questlog system) to convert quests.xml into questlog.lua, people would only need to run a tool/script and its done.

ranisalt:
We can add this new functionality while keeping the old one as deprecated and dropping support as 2.0 ships. People would have plenty of time to upgrade, and it would work just like when upgrading from 0.x to 1.x

andersonfaaria:
For some reason texts in quests for Lua are limited to 200 characters while this limit didn't existed in XML ones. Any guess why this happens?

andersonfaaria:
I've been using this system for a while now, I think it's same to assume it has no bugs despite the one I mentioned earlier. 

What is pending for this to be merged? It would be a good start for someone check why the limit exists.

DSpeichert:
It won't be merged unless backwards compatibility with XML is provided.

andersonfaaria:
> It won't be merged unless backwards compatibility with XML is provided.

he gave a conversion script, it would be way easier if we implement this (along other missing PRs) and move to tfs 1.4 already

EPuncker:
closed on behalf of https://github.com/otland/forgottenserver/commit/d50730a29a05b38cfc9f3b8adde75f6d0815d5f5

PR #1974, opened 2016-11-05
Adding house:getItems()

Mkalo:
I know it was possible to use house:getTiles() and then tile:getItems() but having a method to do that is intuitive.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2375,6 +2375,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("House", "getDoorCount", LuaScriptInterface::luaHouseGetDoorCount);
 
 	registerMethod("House", "getTiles", LuaScriptInterface::luaHouseGetTiles);
+	registerMethod("House", "getItems", LuaScriptInterface::luaHouseGetItems);
 	registerMethod("House", "getTileCount", LuaScriptInterface::luaHouseGetTileCount);
 
 	registerMethod("House", "getAccessList", LuaScriptInterface::luaHouseGetAccessList);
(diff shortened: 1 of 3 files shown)


HeberPcL:
Very useful! Thx

wgrriffel:
@Mkalo knowing that why not write it through Lua API? function House.getItems( ... ), instead of in src?

Mkalo:
@wgrriffel Because if we think about methods that could be written in Lua instead of doing it in the soruce we would move a lot more to Lua. Also C++ is just faster, and doing it in Lua could lead to a possible lag calling that function in big houses with lots of tiles and items (not saying that it couldn't happen in C++).

wgrriffel:
But as you said, lua already handle that, about lua performance, did you benchmark it? i dont think lua in this case will make any relavant change in server performance at all.

ranisalt:
It definitely won't be a performance issue. From my experience with other languages (JS, to be precise), it's worse to switch between core and script than do some scripting instead i.e. it's probably slower to call this new `getItems()` via the Lua API than just script it as @wgrriffel suggested.

Mkalo:
@ranisalt What you just said doesn't make sense at all, it would only be proving that making it in C++ would perform better than using Lua to do it.

Lua: house:getTiles() - 1 call, tile:getItems() - #tiles calls (total number of switching between core and script would be 1 + #tiles)

C++: house:getItems() - 1 call (total number of switching between core and script is 1).

ranisalt:
Yes, you are right, my thinking process is flawed, sorry hahahahah

src/luascript.cpp
```diff
@@ -10416,6 +10417,31 @@ int LuaScriptInterface::luaHouseGetTiles(lua_State* L)
...
 
+int LuaScriptInterface::luaHouseGetItems(lua_State* L)
+{
+	// house:getItems()
+	House* house = getUserdata<House>(L, 1);
+	if (!house) {
+		lua_pushnil(L);
+		return 1;
+	}
+	const auto& tiles = house->getTiles();
```
WibbenZ: New line

Znote:
Please update this:
https://github.com/otland/forgottenserver/wiki/Metatable:House

PR #1976, opened 2016-11-06
Update game protocol to 10990

jo3bingham:
Adds basic compatibility with upcoming update

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -1290,6 +1290,7 @@ void ProtocolGame::sendBasicData()
 		msg.add<uint32_t>(0);
 	}
 	msg.addByte(player->getVocation()->getClientId());
+	msg.addByte(1); // has reached Main (allow player to open Prey window)
 	msg.add<uint16_t>(0x00);
 	writeToOutputBuffer(msg);
 }
@@ -1946,6 +1947,8 @@ void ProtocolGame::sendMarketDetail(uint16_t itemId)
 	} else {
 		msg.add<uint16_t>(0x00);
 	}
+
+	msg.add<uint16_t>(0x00); // imbuement detail
 
 	MarketStatistics* statistics = IOMarket::getInstance()->getPurchaseStatistics(itemId);
 	if (statistics) {


marksamman:
Thanks. It would be nice if we could get these changes into OTClient first, it seems like TFS currently isn't compatible with OTClient and we don't want to drift further away from OTClient.

jo3bingham:
@marksamman Done. https://github.com/edubart/otclient/pull/770

marksamman:
Thanks! I'll leave this open until the same protocol changes make it into OTClient then.

brunominervino:
@jo3bingham great!!
You can make this change too:
https://github.com/otland/forgottenserver/blob/master/src/definitions.h#L28-L29

Before @marksamman merge this commit.

Just a suggestion 👍

jo3bingham:
@brunominervino These are changes in the test server client for the upcoming winter update. CipSoft is using 10990 as the protocol version, but this could change when it's officially released. When the update is released, and I'm sure of the protocol version, I will change the definition. However, there were no changes from 10.98 to the current client version of 10.99.

brunominervino:
@jo3bingham Excuse me, since I'm away from Tibia, I didn't know about this information.

comedinha:
https://github.com/comedinha/Flash-News/commit/0ff896fa8a8947533fb5a5307b853185a972f24c
10.99 launch option to show focus on flash...

comedinha:
Have update to 11.01?

Olimpotibia:
the bytes for sendmarketdetails dont work if you try to pen market and check item details wirh client 10.00

anyone have solution for this?


thnx anywayD:

Riverlance:
@marksamman I use your TFS as 10.99 with the OTClient. And there is no problem with. The OTClient supports the protocol 10.99 yet. :)

jo3bingham:
> Have update to 11.01?
@comedinha https://otland.net/threads/protocol-version-1101.247854/

> the bytes for sendmarketdetails dont work if you try to pen market and check item details wirh client 10.00
@Olimpotibia No other changes were made to the MarketDetail packet apart from what I committed. Not sure what the problem could be.

comedinha:
@jo3bingham When there are many players this fix does not work.

Marahin:
So, any update on this @comedinha @jo3bingham? Is there a way to run Tibia 11(.01) with Forgotten Server right now?

jo3bingham:
@Marahin there's been further changes to the protocol since then that I've addressed on OTLand forums. In fact, this PR should be closed in favor of a new one.

Marahin:
@jo3bingham thank you for caring to answer :-). For now, until Forgotten Server supports it, I'll stick with some unofficial changes I've found. Thanks!

PR #1977, opened 2016-11-06
Send known spells to client

jo3bingham:
Sends all possible known spell ids to the client. Only necessary so that the flash client can use rune/spell hotkeys


marksamman:
Thanks!

PR #1980, opened 2016-11-10
Update ipban.lua

vankk:
Should have a message if you ban the player to warn the Gamemaster that banned.

### DIFF
--- a/data/talkactions/scripts/ipban.lua
+++ b/data/talkactions/scripts/ipban.lua
@@ -16,7 +16,7 @@ function onSay(player, words, param)
 	local targetPlayer = Player(param)
 	if targetPlayer then
 		targetIp = targetPlayer:getIp()
-		player:sendTextMessage(MESSAGE_EVENT_ADVANCE, targetPlayer:getName() .. " has been banned in IP.")
+		player:sendTextMessage(MESSAGE_EVENT_ADVANCE, targetPlayer:getName() .. " IP has been banned.")
 		targetPlayer:remove()
 	end
 
--- a/data/talkactions/scripts/ipban.lua
+++ b/data/talkactions/scripts/ipban.lua
@@ -16,6 +16,7 @@ function onSay(player, words, param)
 	local targetPlayer = Player(param)
 	if targetPlayer then
 		targetIp = targetPlayer:getIp()
+		player:sendTextMessage(MESSAGE_EVENT_ADVANCE, targetPlayer:getName() .. " has been banned in IP.")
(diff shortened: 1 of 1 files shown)


marksamman:
The message should be sent if the player is offline too, and the message should be sent after the IP-ban has actually been applied.

vankk:
That is enough to be merged now @marksamman ?

Tarek1337:
@marksamman Shouldn't it be like this then? 
``local ipBanDays = 7

		return true
	end

	local resultId = db.storeQuery("SELECT `account_id`, `lastip` FROM `players` WHERE `name` = " .. db.escapeString(param))
	if resultId == false then
		return false
	end

	local targetIp = result.getDataLong(resultId, "lastip")
	result.free(resultId)

	local targetPlayer = Player(param)
	if targetPlayer then
		targetIp = targetPlayer:getIp()
		targetPlayer:remove()
	end

	if targetIp == 0 then
		return false
	end

	resultId = db.storeQuery("SELECT 1 FROM `ip_bans` WHERE `ip` = " .. targetIp)
	if resultId ~= false then
		result.free(resultId)
		return false
	end

	local timeNow = os.time()
	db.query("INSERT INTO `ip_bans` (`ip`, `reason`, `banned_at`, `expires_at`, `banned_by`) VALUES (" ..
			targetIp .. ", '', " .. timeNow .. ", " .. timeNow + (ipBanDays * 86400) .. ", " .. player:getGuid() .. ")")
			player:sendTextMessage(MESSAGE_EVENT_ADVANCE, targetPlayer:getName() .. " IP has been banned.")
	return false
end``

vankk:
If you use like this @Tarek1337  then you will receive a attempt to index a nil value.

Tarek1337:
I see Vankk, never been good at Lua, still learning. xd

marksamman:
targetName won't be defined unless the player is online, you should also select it from the database to make sure it's formatted properly. You can also remove account_id from the SELECT query since we don't read it.

vankk:
Mark, I can use string.format in the db.Query and also in db.storeQuery or we don't use it in The Forgotten datapack?

marksamman:
You can use it if there's a good reason to. In this case, it shouldn't be necessary for any of the changes I proposed.

vankk:
I maded a few changes Mark, what did you think? http://pastebin.com/SRVveDfk

marksamman:
That's not how pull requests work. I've committed the fixes to your branch instead. Thanks!

PR #1984, opened 2016-11-12
Fileloader refactor

djarek:
The map loading function will now use the new simplified loader. It has also been split into multiple functions to make it easier to read. Use mmap to load the OTB file. This improves load time slightly by mmaping the loaded map file into memory to reduce unnecessary copying. Moved OTB loader related code to the OTB namespace which makes it clearer that the OTB submodule is mostly indepenent from the rest of the application.

I've tested the map loading code both with the default and the ORTS maps. The load times were about the same.

Signed-off-by: Damian Jarek <damian.jarek93@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ addons:
     packages:
       - libboost-dev
       - libboost-system-dev
+      - libboost-iostreams-dev
       - liblua5.2-dev
       - libluajit-5.1-dev
       - libmysqlclient-dev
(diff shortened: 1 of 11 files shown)


marksamman:
This is great! I’ll merge it when Travis passes so that it gets tested enough before the 1.3 release.

djarek:
Fixed the build errors and squashed the commits.

marksamman:
Thanks!

Issue #1952, opened 2016-10-15
------


A-Syntax:
------------

Kamenuvol:
If you have rewritten it, make a pull request here. We will be grateful 😉

A-Syntax:
I just created a series of maps for all similar attributes, for instance.
```
std::map<std::string, SlotPositionBits> spTypes = {
	{ "head", SLOTP_HEAD },
	{ "body", SLOTP_ARMOR },
	{ "legs", SLOTP_LEGS },
	{ "feet", SLOTP_FEET },
	{ "backpack", SLOTP_BACKPACK },
	{ "two-handed", SLOTP_TWO_HAND },
	{ "necklace", SLOTP_NECKLACE },
	{ "ring", SLOTP_RING },
	{ "ammo", SLOTP_AMMO },
	{ "hand", SLOTP_HAND }
};
```
```
	else if (tmpStrValue == "slottype") {
		tmpStrValue = asLowerCaseString(valueAttribute.as_string());
		if (!spTypes[tmpStrValue]) {
			std::cout << "[Warning - Items::parseItemNode] Unknown slotType: " << tmpStrValue << std::endl;
		}
		else if (tmpStrValue == "right-hand") {
			it.slotPosition &= ~SLOTP_LEFT;
		}
		else if (tmpStrValue == "left-hand") {
			it.slotPosition &= ~SLOTP_RIGHT;
		}
		else {
			it.slotPosition |= spTypes[tmpStrValue];
		}
	}
```
By creating these maps you can add new attributes without worrying about a fatal error. I would provide a fix for you guys but at this point my source code is completely different than the official distribution.

A-Syntax:
If we look at weapontypes in the official sources.
```
	} else if (tmpStrValue == "weapontype") {
		tmpStrValue = asLowerCaseString(valueAttribute.as_string());
		if (tmpStrValue == "sword") {
			it.weaponType = WEAPON_SWORD;
		} else if (tmpStrValue == "club") {
			it.weaponType = WEAPON_CLUB;
		} else if (tmpStrValue == "axe") {
			it.weaponType = WEAPON_AXE;
		} else if (tmpStrValue == "shield") {
			it.weaponType = WEAPON_SHIELD;
		} else if (tmpStrValue == "distance") {
			it.weaponType = WEAPON_DISTANCE;
		} else if (tmpStrValue == "wand") {
			it.weaponType = WEAPON_WAND;
[...]

PR #1985, opened 2016-11-13
Use switch to load items.xml

Mkalo:
This is faster (not really that relevant as either way its fast) but I think it's easier to read and maintain. 

I want to know what you guys think about this, and also we could do something as mentioned in #1952 for some attributes. (just wanted to know if this will ever get merged before doing it.)


A-Syntax:
You are still going to get a fatal error C1061 somewhere down the road, because all you did was band-aide the code, not much of a difference between using switch and if/else. 

As far as the changes go I fail to see how you simplified anything for the better. But what do I know you people don't even want me in your community (banning me for no reason at all). Whatever...

Kamenuvol:
I think all these attributes should be in Lua side.
In C, there would be a single function to add any of these attributes in a map of attributes of that item; and other function to read its value.

src/items.h
```diff
@@ -57,6 +57,220 @@ enum ItemTypes_t {
...
+	ITEM_PARSE_ELEMENTFIRE,
+	ITEM_PARSE_ELEMENTENERGY,
+	ITEM_PARSE_WALKSTACK,
+	ITEM_PARSE_BLOCKING,
+	ITEM_PARSE_ALLOWDISTREAD
+};
+
+const std::unordered_map<std::string, ItemParseAttributes_t> ItemParseAttributesMap = {
+	{"type",				ITEM_PARSE_TYPE},
+	{"description",				ITEM_PARSE_DESCRIPTION},
```
ranisalt: Either align every line or use a single space.

src/items.cpp
```diff
@@ -372,502 +374,801 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
...
+						std::cout << "[Warning - Items::parseItemNode] Unknown field value: " << valueAttribute.as_string() << std::endl;
+					}
+
+					if (combatType != COMBAT_NONE) {
+						it.combatType = combatType;
+						it.conditionDamage.reset(conditionDamage);
+						uint32_t ticks = 0;
+						int32_t damage = 0;
+						int32_t start = 0;
+						int32_t count = 1;
```
ranisalt: All variables above should be inside for loop.

src/items.cpp
```diff
@@ -372,502 +374,801 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
...
+					} else if (tmpStrValue == "axe") {
+						it.weaponType = WEAPON_AXE;
+					} else if (tmpStrValue == "shield") {
+						it.weaponType = WEAPON_SHIELD;
+					} else if (tmpStrValue == "distance") {
+						it.weaponType = WEAPON_DISTANCE;
+					} else if (tmpStrValue == "wand") {
+						it.weaponType = WEAPON_WAND;
+					} else if (tmpStrValue == "ammunition") {
+						it.weaponType = WEAPON_AMMO;
```
ranisalt: You should use a map instead of this if/elseif chain. It also applies to a lot of other places.

src/items.cpp
```diff
@@ -372,514 +532,787 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
...
-				conditionDamage = new ConditionDamage(CONDITIONID_COMBAT, CONDITION_DROWN);
-				combatType = COMBAT_DROWNDAMAGE;
-			} else if (tmpStrValue == "physical") {
-				conditionDamage = new ConditionDamage(CONDITIONID_COMBAT, CONDITION_BLEEDING);
-				combatType = COMBAT_PHYSICALDAMAGE;
-			} else {
-				std::cout << "[Warning - Items::parseItemNode] Unknown field value: " << valueAttribute.as_string() << std::endl;
-			}
+		auto parseAttribute = ItemParseAttributesMap.find(tmpStrValue);
+		if(parseAttribute != ItemParseAttributesMap.end()) {
```
WibbenZ: if (

src/items.cpp
```diff
@@ -29,6 +29,164 @@
...
+	{"femalesleeper", ITEM_PARSE_FEMALETRANSFORMTO},
+	{"transformto", ITEM_PARSE_TRANSFORMTO},
+	{"destroyto", ITEM_PARSE_DESTROYTO},
+	{"elementice", ITEM_PARSE_ELEMENTICE},
+	{"elementearth", ITEM_PARSE_ELEMENTEARTH},
+	{"elementfire", ITEM_PARSE_ELEMENTFIRE},
+	{"elementenergy", ITEM_PARSE_ELEMENTENERGY},
+	{"walkstack", ITEM_PARSE_WALKSTACK},
+	{"blocking", ITEM_PARSE_BLOCKING},
+	{"allowdistread", ITEM_PARSE_ALLOWDISTREAD}
```
WibbenZ: Add a comma here or remove from the the other maps

src/items.cpp
```diff
@@ -372,514 +532,787 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
...
+				case ITEM_PARSE_BLOCKPROJECTILE: {
+					it.blockProjectile = valueAttribute.as_bool();
+					break;
+				}
+
+				case ITEM_PARSE_PICKUPABLE: {
+					it.allowPickupable = valueAttribute.as_bool();
+					break;
+				}
+
```
WibbenZ: Trailing

src/items.h
```diff
@@ -57,6 +57,115 @@ enum ItemTypes_t {
...
+	ITEM_PARSE_FEMALETRANSFORMTO,
+	ITEM_PARSE_TRANSFORMTO,
+	ITEM_PARSE_DESTROYTO,
+	ITEM_PARSE_ELEMENTICE,
+	ITEM_PARSE_ELEMENTEARTH,
+	ITEM_PARSE_ELEMENTFIRE,
+	ITEM_PARSE_ELEMENTENERGY,
+	ITEM_PARSE_WALKSTACK,
+	ITEM_PARSE_BLOCKING,
+	ITEM_PARSE_ALLOWDISTREAD
```
WibbenZ: Same here, if you wanna add a comma or remove the others

src/items.cpp
```diff
@@ -372,514 +532,788 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
...
+				case ITEM_PARSE_CONTAINERSIZE: {
+					it.maxItems = pugi::cast<uint16_t>(valueAttribute.value());
+					break;
+				}
+
+				case ITEM_PARSE_FLUIDSOURCE: {
+					tmpStrValue = asLowerCaseString(valueAttribute.as_string());
+					if (tmpStrValue == "water") {
```
DSpeichert: why not make this a map?

src/items.cpp
```diff
@@ -372,514 +532,788 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
...
+						it.weaponType = it2->second;
+					} else {
+						std::cout << "[Warning - Items::parseItemNode] Unknown weaponType: " << valueAttribute.as_string() << std::endl;
+					}
+					break;
+				}
+
+				case ITEM_PARSE_SLOTTYPE: {
+					tmpStrValue = asLowerCaseString(valueAttribute.as_string());
+					if (tmpStrValue == "head") {
```
DSpeichert: This could be a map as well

src/items.cpp
ranisalt: Legacy code. Truly useful maps are a feature of C++11 and the codebase is probably older :P

DSpeichert:
@Mkalo can you rebase this so we can merge?

Issue #1962, opened 2016-10-24
ItemType(item):getSlotPosition() returns the same value (48) for all slot positions


raymondtfr:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Equip at least your character equipment slots (head and feet).
2. Check the script below
```lua
	local slotItem = player:getSlotItem(CONST_SLOT_FEET)
	if slotItem then
		local slotPos = ItemType(slotItem):getSlotPosition()
		print("Slot position (feet): ", slotPos) -- prints the value 48

		slotItem = player:getSlotItem(CONST_SLOT_HEAD)
		if slotItem then
			slotPos = ItemType(slotItem):getSlotPosition()
			print("Slot position (head): ", slotPos) -- also prints 48
		end
	end
```

### Expected behaviour

It should print the appropriate value for each specific equipment slot.

### Actual behaviour

It prints the same value (48) for all equipment slots.

### Environment

the forgotten server 1.3

ninjalulz:
It's because you're constructing an ItemType object with a userdata value instead of item id or name.

You should receive the following results when it's constructed properly:
```
Slot position (feet): 	176
Slot position (head): 	49
```

andersonfaaria:
@ninjalulz anyway it should work when making an itemtype with a userdata.

raymondtfr:
I could sense I was making something wrong and you just showed me that hehe, thanks for explaining.
I thought I had searched in the sources already so I thought I could use a Item to construct an ItemType but then I checked the sources now that you explained me and found: ``// ItemType(id or name)``.

Should I close this?

@ninjalulz Is it a valid (possible) idea what @andersonfaaria mentioned?

Mistakes of mine like these makes this looks more like a support matter 💃 lol, I should be closing this?

andersonfaaria:
@ninjalulz I have a question, the item class inherit the methods for itemType? if so, @raymondtfr could use slotItem:getType():getSlotPosition(); If not, he could use ItemType(slotItem:getId());

But again, the TFS should accept userdata for ItemType and MonsterType. It's a simple if thing and makes it more consistent.

Mkalo:
[...]

PR #1986, opened 2016-11-13
Fixing ItemType constructor.

Mkalo:
It shouldn't return the ItemType object when it's id is 0 (cause it was
not loaded). This prevents people from constructing it using an Item
userdata as they are going to get an error for indexing a nil value.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -10467,16 +10467,20 @@ int LuaScriptInterface::luaHouseSetAccessList(lua_State* L)
 int LuaScriptInterface::luaItemTypeCreate(lua_State* L)
 {
 	// ItemType(id or name)
-	uint32_t id;
+	uint32_t id = 0;
 	if (isNumber(L, 2)) {
 		id = getNumber<uint32_t>(L, 2);
-	} else {
+	} else if (isString(L, 2)) {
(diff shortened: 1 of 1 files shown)


src/luascript.cpp
```diff
@@ -10467,16 +10467,20 @@ int LuaScriptInterface::luaHouseSetAccessList(lua_State* L)
 int LuaScriptInterface::luaItemTypeCreate(lua_State* L)
 {
 	// ItemType(id or name)
-	uint32_t id;
+	uint32_t id = 0;
```
ranisalt: Why? You are setting the id value on both branches on the next lines, `id` will guaranteedly have a value.

marksamman:
I think this is a bad idea, because there are scripts that expect the ItemType constructor to always return a valid userdata. What you could do to address that issue is to support passing item userdata to the ItemType constructor.

Mkalo:
In my opinion those scripts shouldn't exist, as people can get unexpected behaviours when using it, like in #1962. Yes we could fix that expecific issue supporting the item userdata as parameter but this would still happen if you try to construct the ItemType with an invalid id or name. So every time you construct an ItemType object you have to check if its id is valid before doing whatever you want, this wouldn't be necessary if it returned nil.

PR #1988, opened 2016-11-13
Removing double forward declaration.

Mkalo:
Double forward declaration of the class Npc.


Mkalo:
I have no idea why that check failed.

ranisalt:
Clang just crashed, it happens once in a while.

marksamman:
Thanks!

PR #1990, opened 2016-11-13
Removing commands.xml

Mkalo:
Moved remaining commands to talkaction. I've made some changes in Game.startRaid to work more like the command.

Some of the return values in House.startTrade could be moved to the lua but in my opinion people shouldn't be able to mess with those conditions in order to not break it.


HeberPcL:
Much better in _Lua_. Great job!

Kamenuvol:
The difference is that there is no logging in talkactions

Mkalo:
"Move commands to talkactions" is in the roadmap, maybe now that is completely removed we can add logging to talkactions. It's actually something really simple but I would rather wait for this to be merged before doing it.

Kamenuvol:
Yeah, that should be easy. Also, great job!! I was almost forgetting to congratulate 😄

data/talkactions/scripts/sellhouse.lua
```diff
@@ -0,0 +1,20 @@
+function onSay(player, words, param)
+	local tradePartner = Player(param)
+	if not tradePartner or tradePartner == player then
+		player:sendCancelMessage("Trade player not found.")
+		return false
+	end
+
+	local tile = Tile(player:getPosition())
+	local house = tile:getHouse()
```
ranisalt: Just use `Tile(player:getPosition()):getHouse()`

src/enums.h
```diff
@@ -389,6 +389,15 @@ enum ReturnValue {
...
 	RETURNVALUE_YOUARENOTTHEOWNER,
+	RETURNVALUE_NOSUCHRAID,
+	RETURNVALUE_ANOTHERRAIDEXECUTING,
+	RETURNVALUE_RAIDCONTAINSNODATA,
+	RETURNVALUE_TRADEPLAYERFARAWAY,
+	RETURNVALUE_YOUDONTOWNTHISHOUSE,
+	RETURNVALUE_TRADEPLAYERALREADYOWNSAHOUSE,
+	RETURNVALUE_TRADEPLAYERHIGHESTBIDDER,
+	RETURNVALUE_YOUCANNOTTRADETHISHOUSE,
+
```
ranisalt: Extra blank line

ranisalt:
I'd like to merge this if you could address the issues I pointed out and rebase.

Mkalo:
@wgrriffel I just rebased it, still have to fix stuff.

src/luascript.cpp
```diff
@@ -4413,12 +4438,76 @@ int LuaScriptInterface::luaGameStartRaid(lua_State* L)
...
+	g_game.raids.setRunning(raid);
+	raid->startRaid();
+	lua_pushnumber(L, RETURNVALUE_NOERROR);
+	return 1;
+}
+
+int LuaScriptInterface::luaGameReload(lua_State* L)
+{
+	// Game.reload(interfaceName)
+	const std::string& interfaceName = asLowerCaseString(getString(L, 1));
```
wgrriffel: imo, this is ugly, i'd use a enum with all reload types, also it performs better.

src/luascript.cpp
```diff
@@ -4413,12 +4438,76 @@ int LuaScriptInterface::luaGameStartRaid(lua_State* L)
...
+		|| interfaceName == "moveevents" || interfaceName == "moveevent") {
+		g_moveEvents->reload();
+	} else if (interfaceName == "npc" || interfaceName == "npcs") {
+		Npcs::reload();
+	} else if (interfaceName == "raid" || interfaceName == "raids") {
+		g_game.raids.reload();
+		g_game.raids.startup();
+	} else if (interfaceName == "spell" || interfaceName == "spells") {
+		g_spells->reload();
+		g_monsters.reload();
```
wgrriffel: what if somehow `g_spells->reload()` fail? monsters would have a dangling pointer to some spell? just wondering.

ranisalt: This issue is not related to this PR per se, but it would be good to have a separate issue to understand better.

src/scriptmanager.cpp
```diff
@@ -115,3 +120,79 @@ bool ScriptingManager::loadScriptSystems()
...
+			break;
+
+		case GLOBAL:
+			g_luaEnvironment.loadFile("data/global.lua");
+			break;
+
+		default:
+			break;
+	}
+}
```
ranisalt: Missing EOF linebreak

src/scriptmanager.cpp
```diff
@@ -115,3 +120,79 @@ bool ScriptingManager::loadScriptSystems()
...
+		case CHAT:
+			g_chat->load();
+			break;
+
+		case GLOBAL:
+			g_luaEnvironment.loadFile("data/global.lua");
+			break;
+
+		default:
+			break;
```
ranisalt: You don't need this default case, but it's okay if the compiler whines.

src/scriptmanager.h
```diff
@@ -35,7 +35,29 @@ class ScriptingManager
 			return instance;
 		}
 
+		enum ReloadInterface_t {
+			NONE,
```
ranisalt: This `NONE` makes sense? Why would anyone want to reload nothing?

src/scriptmanager.cpp
Mkalo: It would give an error compiling on linux saying that it doesn't have case for all the possible values of interfaceId, it's either this or making a case for NONE.

src/scriptmanager.h
Mkalo: Used in luascript.cpp to avoid having to write something like:
scriptingManager.reloadInterface(INTERFACEID) in every if statement

src/luascript.cpp
```diff
@@ -10463,6 +10541,52 @@ int LuaScriptInterface::luaHouseSetAccessList(lua_State* L)
...
+	House* house = getUserdata<House>(L, 1);
+	Player* player = getUserdata<Player>(L, 2);
+	Player* tradePartner = getUserdata<Player>(L, 3);
+
+	if (!player || !tradePartner || !house) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	if (!Position::areInRange<2, 2, 0>(tradePartner->getPosition(), player->getPosition())) {
```
WibbenZ: You should move as much as you can to Lua

data/talkactions/scripts/sellhouse.lua
```diff
@@ -0,0 +1,19 @@
...
+		return false
+	end
+
+	local house = Tile(player:getPosition()):getHouse()
+	if not house then
+		player:sendCancelMessage("You must stand in your house to initiate the trade.")
+		return false
+	end
+
+	local ret = house:startTrade(player, tradePartner)
```
WibbenZ: Use returnValue insted

data/talkactions/scripts/raid.lua
```diff
@@ -0,0 +1,17 @@
+function onSay(player, words, param)
+	if not player:getGroup():getAccess() then
+		return true
+	end
+
+	if player:getAccountType() < ACCOUNT_TYPE_GAMEMASTER then
+		return false
+	end
+
+	local ret = Game.startRaid(param)
```
WibbenZ: Use returnValue insted

src/luascript.cpp
```diff
@@ -10463,6 +10541,52 @@ int LuaScriptInterface::luaHouseSetAccessList(lua_State* L)
...
 }
 
+int LuaScriptInterface::luaHouseStartTrade(lua_State* L)
+{
+	// house:startTrade(player, tradePartner)
+	House* house = getUserdata<House>(L, 1);
+	Player* player = getUserdata<Player>(L, 2);
+	Player* tradePartner = getUserdata<Player>(L, 3);
+
+	if (!player || !tradePartner || !house) {
```
WibbenZ: Is this if statment even needed since you check them in Lua?

src/luascript.cpp
ranisalt: It won't work, you are binding a rvalue to a reference. The return for `asLowerCaseString(...)` is destroyed as soon as it's assigned, and as it's assigned to a reference, the content of `interfaceName` become invalid.

ranisalt:
@WibbenZ already had a PR merged removing /reload, so you can rebase and incorporate these changes.

src/luascript.cpp
Mkalo: In my opinion, that way is better to keep consistency of the method, avoiding the user to mess things up in case he tries to use it in other context.

src/luascript.cpp
Mkalo: Just because it's going to be checked in the sell house talkaction, it doesn't mean that someone could still mess this up in another place.

src/luascript.cpp
Mkalo: It is totally fine since its const.

src/luascript.cpp
ranisalt: But this distance requirement can be tweaked in Lua if the scripter wants to. It's not something to be bounded by the engine.

Mkalo:
Closed via #2165

PR #1991, opened 2016-11-13
Replace chaseMode enumeration to bool.

wgrriffel:
I think it makes more sense by this way, since chaseMode can have only 2 values.


marksamman:
Thanks!

Issue #1993, opened 2016-11-16
[Question] Custom item attributes.


Mkalo:
As this is in the roadmap, I would like to ask how it should be done.
I was thinking in a way that would be backward compatible (wouldn't need to change anything in the current structure used to serialize the existent attributes) and it would accept a number or a string for key and number, string, bool or double for value.

I would simply add a new attr byte to be used in the serialization of the structure of an unordered_map that had all the custom attributes in the database. We could use the item:setAttribute() and check the key to decide if its either a normal attribute key (a power of 2 or an attribute string) or a custom attribute (any other number or string). Or create a new function item:setCustomAttribute() that wouldn't need that check.

This is how I was planning to do the structure: http://pastebin.com/dzVdV3HZ
Still have plenty of things to do such as serialize, deserialize, copy methods... Anything can be changed there to fit TFS standards.

I would like to know if this would be a good approach to custom attributes or not and if not what do you think it should be like, specially @marksamman.

marksamman:
Most important is that it doesn't add any memory overhead for items that don't have any custom attributes. Make use of the current ItemAttributes class and add an attribute which is a pointer to a std::unordered_map which is used for custom attributes (which sounds like what you've described, in that case you're on the right track). As long as you get that part right, I don't mind how you implement the value type of the attributes or the Lua API.

Kamenuvol:
I would substitute the entire class ItemAttributes for the unordered_map. Simple items that already exist would have smaller footprint too.

Mkalo:
@Kamenuvol The problem is if we substitute the entire class with the map, we would need to change how attributes are currently serialized and then bye bye backward compatibility (or we would need some ugly code to convert outdated db entries to new format).

marksamman:
@Kamenuvol You're wrong. An unordered_map probably uses twice as much memory as the ItemAttributes class. Unless you're implying we should use a pointer to an unordered_map, in which case there's no difference in terms of memory use for items without attributes, but negatively impacts other great things about the ItemAttributes class, such as quickly checking if an item has an attribute with just a bitwise AND. Iterating over a small enough forward_list may also be faster than a hash lookup.

totolol123:
Items with custom attributes cannot be placed in market. Any thought in changing it? @marksamman @Mkalo

Mkalo:
[...]

PR #1997, opened 2016-11-17
Item custom attributes

Mkalo:
Introducing 3 new methods to Item class to Lua API:

item:setCustomAttribute(key, value)
item:getCustomAttribute(key)
item:removeCustomAttribute(key)

Key types: string or number
Value types: string, number, double or boolean.

OBS: Setting a custom attribute to nil doesn't remove the attribute, it has no effect at all.

Any suggestion is welcome.


infernus74:
I suggest you to fix this whole thing, make it be 7.4 cause this new tibia is shit.

totolol123:
How about market?
You can also provide some example code to better insight?
Looks great 👍

Mkalo:
@totolol123 True, I forgot about market. You shouldn't be able to place items that have custom attributes in the market.

src/item.h
```diff
@@ -206,6 +207,226 @@ class ItemAttributes
...
+		};
+
+		struct CustomAttribute
+		{
+			union {
+				std::string* string;
+				int64_t integer;
+				double doublen;
+				bool boolean;
+			} value;
```
djarek: This union should be a separate class (or preferably use boost:any/boost::variant instead)

src/item.h
```diff
@@ -206,6 +207,226 @@ class ItemAttributes
...
+
+			CustomAttributeKey& operator=(CustomAttributeKey&& other) {
+				if (this != &other) {
+					if (type == ATTR_STRING_TYPE) {
+						delete value.string;
+					}
+
+					value = other.value;
+					type = other.type;
+					memset(&other.value, 0, sizeof(value));
```
djarek: memset/memcpy is considered bad practice in C++11 and newer (refer [here](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#a-namepro-bounds-stdlibabounds4-dont-use-standard-library-functions-and-types-that-are-not-bounds-checked)).

src/item.h
```diff
@@ -206,6 +207,226 @@ class ItemAttributes
...
+
+			CustomAttributeKey& operator=(CustomAttributeKey other) {
+				CustomAttributeKey::swap(*this, other);
+				return *this;
+			}
+
+			CustomAttributeKey& operator=(CustomAttributeKey&& other) {
+				if (this != &other) {
+					if (type == ATTR_STRING_TYPE) {
+						delete value.string;
```
djarek: Naked delete.

src/item.h
```diff
@@ -206,6 +207,226 @@ class ItemAttributes
...
+					}
+
+					default: {
+						memset(&value, 0, sizeof(value));
+					}
+				}
+			}
+
+			CustomAttributeKey(CustomAttributeKey&& attribute) : value(attribute.value), type(attribute.type) {
+				memset(&attribute.value, 0, sizeof(value));
```
djarek: memset/memcpy is considered bad practice in C++11 and newer (refer [here](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#a-namepro-bounds-stdlibabounds4-dont-use-standard-library-functions-and-types-that-are-not-bounds-checked)).

src/item.h
```diff
@@ -206,6 +207,226 @@ class ItemAttributes
...
+					case ATTR_STRING_TYPE: {
+						value.string = new std::string(*i.value.string);
+					}
+
+					case ATTR_INTEGER_TYPE: {
+						value.integer = i.value.integer;
+					}
+
+					default: {
+						memset(&value, 0, sizeof(value));
```
djarek: memset/memcpy is considered bad practice in C++11 and newer (refer [here](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#a-namepro-bounds-stdlibabounds4-dont-use-standard-library-functions-and-types-that-are-not-bounds-checked)).

src/item.h
```diff
@@ -206,6 +207,226 @@ class ItemAttributes
...
+		{
+			union {
+				std::string* string;
+				int64_t integer;
+			} value;
+
+			customAttrTypes type = ATTR_NO_TYPE;
+
+			explicit CustomAttributeKey(customAttrTypes type) : type(type) {
+				memset(&value, 0, sizeof(value));
```
djarek: memset/memcpy is considered bad practice in C++11 and newer (refer [here](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#a-namepro-bounds-stdlibabounds4-dont-use-standard-library-functions-and-types-that-are-not-bounds-checked)).

src/item.h
```diff
@@ -206,6 +207,226 @@ class ItemAttributes
 			return static_cast<ItemDecayState_t>(getIntAttr(ITEM_ATTRIBUTE_DECAYSTATE));
 		}
 
+		struct CustomAttributeKey
+		{
+			union {
+				std::string* string;
+				int64_t integer;
+			} value;
```
djarek: This union should be a separate class (or preferably use boost::variant/boost::any instead).

Mkalo:
@djarek If you could check it again, never used boost::variant before.

src/item.h
```diff
@@ -207,76 +208,67 @@ class ItemAttributes
...
-			} value;
+			VariantKey value;
 
 			customAttrTypes type = ATTR_NO_TYPE;
 
 			explicit CustomAttributeKey(customAttrTypes type) : type(type) {
-				memset(&value, 0, sizeof(value));
 			}
 
 			CustomAttributeKey(const std::string& v) {
```
djarek: The constructor can be templated to reduce boilerplate.

src/item.h
```diff
@@ -327,95 +318,106 @@ class ItemAttributes
...
+				value = v;
 			}
 
 			CustomAttribute(bool v) {
 				type = ATTR_BOOLEAN_TYPE;
-				value.boolean = v;
+				value = v;
 			}
 
 			CustomAttribute(const CustomAttribute& i) {
```
djarek: I don't think you need to define the copy and move constructors/assignment operators. The defaults are fine.

src/luascript.cpp
```diff
@@ -6197,11 +6197,11 @@ int LuaScriptInterface::luaItemSetCustomAttribute(lua_State* L) {
 	if (isNumber(L, 2)) {
 		int64_t tmp = getNumber<int64_t>(L, 2);
 		key.type = ATTR_INTEGER_TYPE;
-		key.value.integer = tmp;
+		key.value = tmp;
 	} else if (isString(L, 2)) {
 		const std::string& tmp = getString(L, 2);
 		key.type = ATTR_STRING_TYPE;
-		key.value.string = new std::string(tmp);
+		key.value = std::string(tmp);
```
djarek: The construction of a temporary std::string is unnecessary - you can just call key.value = getString(...).

src/luascript.cpp
```diff
@@ -6212,19 +6212,19 @@ int LuaScriptInterface::luaItemSetCustomAttribute(lua_State* L) {
 		double tmp = getNumber<double>(L, 3);
 		if (std::floor(tmp) < tmp) {
 			val.type = ATTR_DOUBLE_TYPE;
-			val.value.doublen = tmp;
+			val.value = tmp;
 		} else {
 			val.type = ATTR_INTEGER_TYPE;
-			val.value.integer = (int64_t)tmp;
+			val.value = (int64_t)tmp;
```
djarek: static_cast\<int64_t\>()

src/item.h
```diff
@@ -327,95 +318,106 @@ class ItemAttributes
...
 		struct CustomAttribute
 		{
-			union {
-				std::string* string;
-				int64_t integer;
-				double doublen;
-				bool boolean;
-			} value;
+			VariantAttribute value;
 			customAttrTypes type = ATTR_NO_TYPE;
```
djarek: The type can be inferred from the Variant class, thus saving a little memory.

src/item.cpp
```diff
@@ -749,6 +839,60 @@ void Item::serializeAttr(PropWriteStream& propWriteStream) const
...
+			customAttrTypes valueType = entry.second.type;
+			propWriteStream.write<customAttrTypes>(valueType);
+			switch (valueType) {
+				case ATTR_STRING_TYPE: {
+					propWriteStream.writeString(boost::get<std::string>(entry.second.value));
+					break;
+				}
+
+				case ATTR_INTEGER_TYPE: {
+					propWriteStream.write<int64_t>(boost::get<int64_t>(entry.second.value));
```
djarek: I think that specifying the type in write is a little redundant:
```
propWriteStream.write(boost::get<int64_t>(entry.second.value));
```
is sufficient IMO.

src/item.cpp
```diff
@@ -749,6 +839,60 @@ void Item::serializeAttr(PropWriteStream& propWriteStream) const
 		propWriteStream.write<uint8_t>(ATTR_SHOOTRANGE);
 		propWriteStream.write<uint8_t>(getIntAttr(ITEM_ATTRIBUTE_SHOOTRANGE));
 	}
+
+	if (hasCustomAttributes()) {
+		propWriteStream.write<uint8_t>(ATTR_CUSTOM_ATTRIBUTES);
+		propWriteStream.write<size_t>(attributes->getCustomAttributeMap().size());
+		for (const auto &entry : attributes->getCustomAttributeMap()) {
+			// Serializing key type and value
+			customAttrTypes keyType = entry.first.type;
```
djarek: const&

src/enums.h
```diff
@@ -61,6 +61,14 @@ enum itemAttrTypes : uint32_t {
 	ITEM_ATTRIBUTE_DOORID = 1 << 22,
 };
 
+enum customAttrTypes : uint8_t {
```
djarek: I think the convention in tfs is to use CamelCase for type names and camelBack for object/variable/function/member names.

src/luascript.cpp
```diff
@@ -6182,6 +6185,138 @@ int LuaScriptInterface::luaItemRemoveAttribute(lua_State* L)
...
+			val.type = ATTR_DOUBLE_TYPE;
+			val.value = tmp;
+		} else {
+			val.type = ATTR_INTEGER_TYPE;
+			val.value = (int64_t)tmp;
+		}
+	} else if (isString(L, 3)) {
+		const std::string& tmp = getString(L, 3);
+		val.type = ATTR_STRING_TYPE;
+		val.value = std::string(tmp);
```
djarek: Constructing a temporary string is redundant here.

src/item.h
```diff
@@ -282,6 +527,60 @@ class ItemAttributes
 		const Attribute* getExistingAttr(itemAttrTypes type) const;
 		Attribute& getAttr(itemAttrTypes type);
 
+		typedef std::unordered_map<CustomAttributeKey, CustomAttribute, CustomAttributeKey::hash> CustomAttributeMap;
+		CustomAttributeMap customAttributes;
+		bool hasCustomAttributes = false;
```
djarek: Do we need this boolean? It is equivalent to customAttributes.empty().

src/item.h
```diff
@@ -206,6 +208,249 @@ class ItemAttributes
...
+					}
+
+					default: {
+						value = boost::blank();
+						break;
+					}
+				}
+			}
+
+			CustomAttribute(const std::string& v) {
```
djarek: Use template for the constructor.

src/item.h
```diff
@@ -206,6 +208,249 @@ class ItemAttributes
...
+				type = ATTR_DOUBLE_TYPE;
+				value = v;
+			}
+
+			CustomAttribute(bool v) {
+				type = ATTR_BOOLEAN_TYPE;
+				value = v;
+			}
+
+			CustomAttribute(const CustomAttribute& i) {
```
djarek: I don't think we need copy/move constructors/assignment operators here either.

src/item.h
ranisalt: And please `CustomAttribute(...): attr(value)`, not this Java-like assignment

Mkalo:
@djarek Yes it does increase memory usage quite a bit, I couldn't measure it in TFS tho (every time I started up I got a different memory usage so this wasn't helping) so I just copied the structs and everything and made a simple class that had the unordered map as in item and constructed 300000 items (I think this is a good amount for a realistic server) and only with only the map it was consuming 106MB~ish of memory. So what should I do in this situation?
I was thinking maybe use an unique_ptr and only initialize it when requested or maybe boost::flat_map but probably the unique_ptr as Mark said in #1993, this shouldn't add any memory overhead.

marksamman:
This hasn't been implemented the way I suggested in #1993. You've simply added the map to the ItemAttributes class, which will create a map for every item that has at least one attribute value set, regardless if it's a custom attribute or not. You should add the map as a pointer in the value union of the Attribute struct. I would also suggest using the most significant bit (1 << 31) for ITEM_ATTRIBUTE_CUSTOM.

Mkalo:
@marksamman But then we are back at the explicit memory management that @djarek told me to avoid.

marksamman:
We're already doing explicit memory management for the string type. If there's a way to avoid that by using boost variant or something else (and it doesn't add additional memory overhead), then please go ahead and replace the Attribute value union with that.

src/item.cpp
```diff
@@ -749,6 +772,19 @@ void Item::serializeAttr(PropWriteStream& propWriteStream) const
 		propWriteStream.write<uint8_t>(ATTR_SHOOTRANGE);
 		propWriteStream.write<uint8_t>(getIntAttr(ITEM_ATTRIBUTE_SHOOTRANGE));
 	}
+
+	if (hasAttribute(ITEM_ATTRIBUTE_CUSTOM)) {
+		const ItemAttributes::CustomAttributeMap* customAttrMap = attributes->getCustomAttributeMap();
+		propWriteStream.write<uint8_t>(ATTR_CUSTOM_ATTRIBUTES);
+		propWriteStream.write<size_t>(customAttrMap->size());
```
djarek: Don't write size_t to the prop stream - it's not portable (use uint64_t instead).

src/item.h
```diff
@@ -206,19 +210,271 @@ class ItemAttributes
...
+
+			bool operator==(const CustomAttributeKey& other) const {
+				if (value.which() != other.value.which()) {
+					return false;
+				}
+
+				return boost::apply_visitor(equalityVisitor(), value, other.value);
+			}
+
+			struct hashVisitor : public boost::static_visitor<size_t> {
```
djarek: HashVisitor

src/item.h
```diff
@@ -206,19 +210,271 @@ class ItemAttributes
...
+
+					default: {
+						value = boost::blank();
+						return false;
+					}
+				}
+				return true;
+			}
+
+			struct equalityVisitor : public boost::static_visitor<bool> {
```
djarek: EqualityVisitor

src/item.h
```diff
@@ -206,19 +210,271 @@ class ItemAttributes
...
+
+			template<typename T>
+			T get() {
+				if (value.type() == typeid(T)) {
+					return boost::get<T>(value);
+				}
+				return T();
+			}
+
+			struct serializeVisitor : public boost::static_visitor<> {
```
djarek: SerializeVisitor

src/item.h
```diff
@@ -206,19 +210,271 @@ class ItemAttributes
...
+					return 0;
+				}
+
+				template<typename T>
+				size_t operator()(const T& value) const {
+					return std::hash<T>()(value);
+				}
+			};
+
+			struct hash {
```
djarek: Hash

src/item.h
```diff
@@ -206,19 +210,271 @@ class ItemAttributes
...
+
+			template<typename T>
+			const T get() {
+				if (value.type() == typeid(T)) {
+					return boost::get<T>(value);
+				}
+				return T();
+			}
+
+			struct pushLuaVisitor : public boost::static_visitor<> {
```
djarek: PushLuaVisitor

src/item.h
```diff
@@ -206,19 +210,271 @@ class ItemAttributes
...
+
+				void operator()(const boost::blank& v) const {
+					lua_pushnil(L);
+				}
+
+				void operator()(const std::string& v) const {
+					LuaScriptInterface::pushString(L, v);
+				}
+
+				void operator()(const bool& v) const {
```
djarek: Fundamental types can be passed by value - they usually fit in registers. But I think it won't matter here, because these functions will be inlined anyway.

src/item.h
```diff
@@ -206,19 +210,271 @@ class ItemAttributes
...
+				void operator()(const double& v) const {
+					lua_pushnumber(L, v);
+				}
+			};
+
+			void pushToLua(lua_State* L) const {
+				boost::apply_visitor(pushLuaVisitor(L), value);
+			}
+
+			struct serializeVisitor : public boost::static_visitor<> {
```
djarek: SerializeVisitor

Mkalo:
If there is nothing else to change or wrong with this I believe this is ready.

src/enums.h
```diff
@@ -59,6 +59,8 @@ enum itemAttrTypes : uint32_t {
 	ITEM_ATTRIBUTE_CHARGES = 1 << 20,
 	ITEM_ATTRIBUTE_FLUIDTYPE = 1 << 21,
 	ITEM_ATTRIBUTE_DOORID = 1 << 22,
+
+	ITEM_ATTRIBUTE_CUSTOM = 1 << 30
```
marksamman: 1 << 31

Mkalo: Yes I've tried that and got this:
`c:\users\user\documents\github\fork\forgottenserver\src\enums.h(64): warning C4309: 'initializing': truncation of constant value` 
`c:\users\user\documents\github\fork\forgottenserver\src\enums.h(64): warning C4369: 'ITEM_ATTRIBUTE_CUSTOM':  enumerator value '-2147483648' cannot be represented as 'unsigned int', value is '2147483648' `

marksamman: You have to cast it.
```
static_cast<uint32_t>(1) << 31,
```

djarek: or  a little simpler: ```1u << 31```  ;)

src/item.h
```diff
@@ -206,19 +210,265 @@ class ItemAttributes
...
+				// This is hard coded so it's not general, depends on the position of the variants.
+				uint8_t pos;
+				if (!propStream.read<uint8_t>(pos)) {
+					return false;
+				}
+
+				switch (pos) {
+					case 1:  { // std::string
+						std::string tmp;
+						if (!propStream.readString(tmp)) {
```
marksamman: Isn't it possible to read directly into value instead of a temporary? Same thing for the int64_t.

Mkalo: Not that I know. Using propStream.readString(value) doesn't work.

marksamman:
Changing the key type to std::string would be good enough in my opinion. It would simplify the implementation a lot. Numbers can be converted to strings so that it still works if a number is passed as key. The keys could also be case insensitive by always transforming the input to lowercase.

Mkalo:
@marksamman Yes it does make it simplier but also removes the possibility to add custom types to be strings but that would really only be useful if you were going to. So I've changed to what you said.

src/item.cpp
```diff
@@ -1661,3 +1700,39 @@ bool Item::hasMarketAttributes() const
...
+}
+
+template<>
+const bool& ItemAttributes::CustomAttribute::get<bool>() {
+	if (value.type() == typeid(bool)) {
+		return boost::get<bool>(value);
+	}
+
+	return emptyBool;
+}
```
brunominervino: Fix EOF

brunominervino: Fix EOF

src/item.h
ranisalt: In the best outcome, it doesn't make any difference, but in the worst it adds overhead for the boolean pointer. I'd change it just in case, and throughout the code, primitive types are always passed by value.

PrinterLUA:
Is it possible to make it accept tables as attribute and also what is left to do?

src/enums.h
```diff
@@ -59,6 +59,8 @@ enum itemAttrTypes : uint32_t {
 	ITEM_ATTRIBUTE_CHARGES = 1 << 20,
 	ITEM_ATTRIBUTE_FLUIDTYPE = 1 << 21,
 	ITEM_ATTRIBUTE_DOORID = 1 << 22,
+
+	ITEM_ATTRIBUTE_CUSTOM = static_cast<uint32_t>(1) << 31
```
ranisalt: You don't need to cast to uint32_t, 1 << 31 fits into `int`. Alternatively you can use `1u << 32`

Mkalo:
Anything else to add/change @ranisalt @djarek ?

ranisalt:
I'd like to hold this while #2139 is pending, because it would require a major refactor of this same stuff there and because using protobuf would ease a LOT the implementation of custom attributes.

Mkalo:
@ranisalt But doesn't it break backward compatibility? I thought pull requests that are not backward compatible would have to wait for a major release that I don't even know when it's going to happen. You could make a migration script to convert all items attribute saved in the db to protobuf format but are you considering that?

DSpeichert:
@ranisalt In the interest of getting stuff out, I think it shouldn't be blocked by #2139. It's a different, more complicated PR that can then reimplement those Lua API methods with different backend storage.

djarek:
@DSpeichert @ranisalt 
Perhaps we could assign #2139 to tfs 2.0 and use protobuf(or flatbuffers) instead of otbm as well as for serialization.

DSpeichert:
@djarek I generally like the idea of moving big change to the next big release. Protobuf/flatbuffers would be a good addition if implemented properly but it will definitely take time.

LuigiLazzari:
Is there documentation on how this can be used?

Mkalo:
@LuigiLazzari There's nothing to make a documentation, you just need to know the functions and the parameters. If you're asking for documentation to use it inside the source code, you just have to read the new methods. If you don't know the type stored in the CustomAttribute (if your code uses various types for one attribute) you will need to code a visitor or make a switch to check what type the CustomAttribute is storing, otherwise you can just use `item->getCustomAttribute(key)->get<TYPE>()`

PrinterLUA:
This is something which has been delayed for to long. For those who want to make custom server. So this should be pushed into the tfs 1.3 release. Since it has been discussed about custom attribute since tfs 1.0

brunominervino:
I'm agree with @PrinterLUA. How @DSpeichert told, will definitely take time, to do for next big release and this implementation is taking a long time, always being left for later.

tuxico:
does not work, impossible add attributes in TFS??
```
item:setCustomAttribute(holdingid, 1515)
player:sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, "Holding Id: " .. (getCustomAttribute(holdingid)))
```

> Attempt to call global 'getCustomAttribute' <a nil value>

diath:
Perhaps merging stackable items with custom attributes should first check if both stacks have the same custom attributes.

Codinablack:
I don't know if this matters to anyone or not, but using this PR increases my map load time from 0.004 seconds to 22.** every time. Also increases ram using from 0.1 gigs to 0.6 gigs. I did not alter the data pack to utilize the PR.

Codinablack:
Just an idea to throw out there that would avoid the problems mentioned in my previous post.

How about we create a new attribute called ITEM_ATTRIBUTE_CUSTOM (enum) that just represents a string value the exact same way ITEM_ATTRIBUTE_TEXT does, except that it isn't hardcoded to be read by anything besides the luascript method item:getAttribute. Then we could use a simple lua method to serialize a table to the string type attribute, and another one to read it (unserialize a table). From there people could create their own functions for their own custom attributes, would only be saved to the item the same way someone would save text to an item, only difference is the text doesn't show up, just hidden.

Mkalo:
@Codinablack I don't think this solution would be incorporated, it's just a bodge. But if you are interested in something like this you can check here https://otland.net/threads/lib-storing-information-items-and-players-tfs-1-2.241873/ that is exactly what you're looking for.

jo3bingham:
>Also I think it's weird that it has any impact in map loading, I didn't notice anything when I was testing it and I cannot see any reason for that either, I could be wrong tho.

If items on the map were using this attribute then it would cause some impact with map loading. However, the impact would be basically unnoticeable, and definitely not at the 5500% increase that @Codinablack is suggesting. Also, the claim that it `increases ram using from 0.1 gigs to 0.6 gigs` makes no sense here.

Codinablack:
@Mkalo thanks for the link but that is not what I was talking about although it is close, right now I already utilize a method for serializing tables to strings and saving those tables to the ITEM_ATTRIBUTE_TEXT, but my function has to check if the item is writable otherwise it would be useless because a player could alter the text value. 

I do realize I may have been wrong about the above post for the benchmarks, the only comparison I used was with the custom branch I forked from master and merged this PR into, and the latest AppVeyor build that was online the day I reported this, although when testing other AppVeyor builds, I noticed it wasn't the PR that did this... It was from switching between Debug, and Release. I am sorry I hadn't noticed this before.

Kamenuvol:
Why do not use three custom containers instead?
vector<bool> customBooleanAttributes
map<size_t, string> customStringAttributes
map<size_t, int32_t> customIntegerAttributes

DSpeichert:
@Mkalo can you please rebase against master?

src/item.cpp
```diff
@@ -617,6 +617,29 @@ Attr_ReadValue Item::readAttr(AttrTypes_t attr, PropStream& propStream)
 			return ATTR_READ_ERROR;
 		}
 
+		case ATTR_CUSTOM_ATTRIBUTES: {
+			uint64_t size;
+			if (!propStream.read<uint64_t>(size)) {
+				return ATTR_READ_ERROR;
+			}
+			for (uint64_t i = 0; i < size; i++) {
```
WibbenZ: New line

src/item.cpp
```diff
@@ -617,6 +617,29 @@ Attr_ReadValue Item::readAttr(AttrTypes_t attr, PropStream& propStream)
...
+				// Unserialize key type and value
+				std::string key;
+				if (!propStream.readString(key)) {
+					return ATTR_READ_ERROR;
+				};
+
```
WibbenZ: Trailing

src/luascript.cpp
```diff
@@ -6182,6 +6185,91 @@ int LuaScriptInterface::luaItemRemoveAttribute(lua_State* L)
...
+		}
+	} else if (isString(L, 3)) {
+		val.set<std::string>(getString(L, 3));
+	} else if (isBoolean(L, 3)) {
+		val.set<bool>(getBoolean(L, 3));
+	} else {
+		lua_pushnil(L);
+		return 1;
+	}
+
```
WibbenZ: Trailing

src/luascript.cpp
```diff
@@ -6182,6 +6185,91 @@ int LuaScriptInterface::luaItemRemoveAttribute(lua_State* L)
...
+	if (isNumber(L, 2)) {
+		pushBoolean(L, item->removeCustomAttribute(getNumber<int64_t>(L, 2)));
+	} else if (isString(L, 2)) {
+		pushBoolean(L, item->removeCustomAttribute(getString(L, 2)));
+	} else {
+		lua_pushnil(L);
+	}
+
```
WibbenZ: Remove the new line

src/item.cpp
```diff
@@ -617,6 +617,29 @@ Attr_ReadValue Item::readAttr(AttrTypes_t attr, PropStream& propStream)
...
+			}
+			for (uint64_t i = 0; i < size; i++) {
+				// Unserialize key type and value
+				std::string key;
+				if (!propStream.readString(key)) {
+					return ATTR_READ_ERROR;
+				};
+				
+				// Unserialize value type and value
+				ItemAttributes::CustomAttribute val;
```
WibbenZ: val => attr

src/luascript.cpp
```diff
@@ -6182,6 +6185,91 @@ int LuaScriptInterface::luaItemRemoveAttribute(lua_State* L)
...
+	if (isNumber(L, 2)) {
+		key = boost::lexical_cast<std::string>(getNumber<int64_t>(L, 2));
+	} else if (isString(L, 2)) {
+		key = getString(L, 2);
+	} else {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	ItemAttributes::CustomAttribute val;
```
WibbenZ: val => attr

src/item.cpp
Mkalo: Why would it be named attr? That type is the value of the attribute.

src/luascript.h
```diff
@@ -703,6 +704,10 @@ class LuaScriptInterface
 		static int luaItemGetAttribute(lua_State* L);
 		static int luaItemSetAttribute(lua_State* L);
 		static int luaItemRemoveAttribute(lua_State* L);
+		static int luaItemSetCustomAttribute(lua_State* L);
```
DSpeichert: Order: get, set, remove instead of set, get, remove.

src/luascript.h
```diff
@@ -703,6 +704,10 @@ class LuaScriptInterface
...
+		static int luaItemGetCustomAttribute(lua_State* L);
+		static int luaItemRemoveCustomAttribute(lua_State* L);
+
```
DSpeichert: Unnecessary empty line

PR #1999, opened 2016-11-17
Added attack speed to Weapons

Ledan-Chen:
This will override the vocations attack speed setting and use the item's attack speed instead. If there is no attack speed set on the item then attack speed will default to the vocation attack speed. The settings only apply to weapons and they must be in either the left or right hand.


Ledan-Chen:
Closed because I made a pull from the master branch

Issue #984, opened 2014-09-10
Live cast system [$205]


Znote:
# Live Cast System

twifysoft.net has a very good cast system, but I understand if they wish to keep the sources closed. 
However, I believe a good live cast system will help the growth of this community, and motivate more people to jump on projects based on this repository. 

**So I want a live cast system, on this official repository.**
## How I want it to work (copied from twifysoft):
- You can start casting by typing the command !cast ingame. 
- You can also start casting with a password, by using the command !cast password. 
- Login to the server without account name to watch a live cast. 
- If the cast has a password, enter the password in the password login field (still empty account name).
- An in-game cast group chat channel. If the caster has the channel open, it is considered enabled, if caster is not in the channel its considered disabled. 
## Code quality guidelines:

[@marksamman](https://github.com/otland/forgottenserver/issues/294#issuecomment-30718539):
- The use of the dispatcher thread must be minimal (viewers should receive a copy of the packets sent to the caster in asio thread). 
- All caster variables must be stored in (or in another object referenced from) the ProtocolGame instance of the caster player.

I will be the first here to take initiative and use Bountysource, I hope more people will follow. :+1: 
##

rowaraujo:
If I publish my system, I get how much?

EPuncker:
if it fulfills all @marksamman requirements you may get the bounty I guess

Znote:
@marksamman @djarek Not sure whats going on with the bountysource integration, but I chipped in some extra bucks raising this from $170 to $205.

ranisalt:
@Znote it's working.
`There is a $205 open bounty on this issue. Add to the bounty at Bountysource.`

gugahoa:
Isn't @djarek system already done and working? Most of the people that shows up with errors are using it with modifications

Znote:
@ranisalt I edited that myself manually, the title still says $170 and it appears I cannot change that. 

@gugahoa There is a pull request by djarek, but it hasn't been accepted and merged yet by Marksamman. Last thing I read there are some dodgy code in the shadows of this request that need some work for the sake of this request, as well as some good testers who know how to use a debugger. 

I don't care if a pull request is working, it has to be merged to master or the bounty still stands.

marksamman:
[...]

PR #2001, opened 2016-11-18
Cast system, 10.99 support and premium display in character list fix.

Milice:
I have added a cast system fully compatible with tfs 1.3, with multiple chat commands including but not limited to /kick, /togglechat, /ban, etc.

I also have changed the definitions to support login with the 10.99 client thus also the Tibia 11 client (i have released a functioning webscript for this on OTLand), and i noticed that the premium days weren't displaying properly in the character list so i also provided a fix for this.

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -330,6 +330,15 @@ CREATE TABLE IF NOT EXISTS `tile_store` (
   FOREIGN KEY (`house_id`) REFERENCES `houses` (`id`) ON DELETE CASCADE
 ) ENGINE=InnoDB;
 
+CREATE TABLE IF NOT EXISTS `live_casts` (
+  `player_id` int(11) NOT NULL,
+  `name` varchar(32) NOT NULL,
+  `password` varchar(30) NOT NULL DEFAULT '0',
+  `spectators` int(11) NOT NULL DEFAULT '0',
+  PRIMARY KEY (`player_id`),
+  UNIQUE KEY `name` (`name`)
+) ENGINE=InnoDB;
+
 DROP TRIGGER IF EXISTS `ondelete_players`;
 DROP TRIGGER IF EXISTS `oncreate_guilds`;
 
--- a/schema.sql
+++ b/schema.sql
@@ -333,7 +333,7 @@ CREATE TABLE IF NOT EXISTS `tile_store` (
 CREATE TABLE IF NOT EXISTS `live_casts` (
   `player_id` int(11) NOT NULL,
   `name` varchar(32) NOT NULL,
-  `password` varchar(30) NOT NULL DEFAULT '0',
+  `password` varchar(30) NOT NULL DEFAULT '',
   `spectators` int(11) NOT NULL DEFAULT '0',
   PRIMARY KEY (`player_id`),
   UNIQUE KEY `name` (`name`)
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -160,6 +160,35 @@ uint32_t IOLoginData::gameworldAuthentication(const std::string& accountName, co
 	return accountId;
 }
 
+std::list<std::string> IOLoginData::liveCastAuthentication(const std::string& password)
+{
+	std::list<std::string> casts;
+
+	Database* db = Database::getInstance();
+
+	std::ostringstream query;
+	query << "SELECT `player_id`, `name`, `password`, `spectators` FROM `live_casts` WHERE `password` = '" << password << "'";
+	DBResult_ptr result = db->storeQuery(query.str());
+	if (!result) {
+		return casts;
+	}
+
+	query.str(std::string());
+	query << "SELECT `name` FROM `live_casts` WHERE `password` = '" << password << "' ORDER BY `name` ASC";
+	result = db->storeQuery(query.str());
+	if (result) {
+		do {
+			std::string name = result->getString("name");
+			if(Player* castPlayer = g_game.getPlayerByName(name)){
+				if(castPlayer->castPassword == password) {
+					casts.push_back(name);
+				}
+			}
+		} while (result->next());
+	}
+	return casts;
+}
+
 AccountType_t IOLoginData::getAccountType(uint32_t accountId)
 {
 	std::ostringstream query;
@@ -638,6 +667,10 @@ bool IOLoginData::saveItems(const Player* player, const ItemBlockList& itemList,
 
 bool IOLoginData::savePlayer(Player* player)
 {
+	if(player->isSpectator) {
+		return false;
+	}
+
 	if (player->getHealth() <= 0) {
 		player->changeHealth(1);
 	}
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -30,6 +30,7 @@
 #include "rsa.h"
 #include "protocolold.h"
 #include "protocollogin.h"
+#include "protocolspectator.h"
 #include "protocolstatus.h"
 #include "databasemanager.h"
 #include "scheduler.h"
@@ -168,6 +169,14 @@ void mainLoader(int, char*[], ServiceManager* services)
 	if (g_config.getBoolean(ConfigManager::OPTIMIZE_DATABASE) && !DatabaseManager::optimizeTables()) {
 		std::cout << "> No tables were optimized." << std::endl;
 	}
+	
+	// clear live casts
+	if (g_config.getBoolean(ConfigManager::LIVE_CAST_ENABLED)) {
+		std::cout << ">> Clearing live casts" << std::endl;
+		std::ostringstream query;
+		query << "DELETE FROM `live_casts`;";
+		Database::getInstance()->executeQuery(query.str());
+	}
 
 	//load vocations
 	std::cout << ">> Loading vocations" << std::endl;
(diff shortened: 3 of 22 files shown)


DSpeichert:
You should split this PR into separate: cast system, 10.99 and other fixes.
Don't include "reverts" and merges that are not relevant to the upstream repository (this one).
That will help your PR get reviewed faster/easier.

src/protocolgame.cpp
```diff
@@ -366,8 +371,13 @@ void ProtocolGame::disconnectClient(const std::string& message) const
 	disconnect();
 }
 
-void ProtocolGame::writeToOutputBuffer(const NetworkMessage& msg)
+void ProtocolGame::writeToOutputBuffer(const NetworkMessage& msg, bool broadcast/* = true*/)
 {
+	if(broadcast && player->isLiveCasting()) {
+		for(auto& spectator : player->spectators) {
+				spectator->writeToOutputBuffer(msg, false);
```
djarek: I don't think this solution will be accepted https://github.com/otland/forgottenserver/issues/984#issue-42451810 - the work of broadcasting is done in the dispatcher thread.

src/iologindata.cpp
```diff
@@ -160,6 +160,35 @@ uint32_t IOLoginData::gameworldAuthentication(const std::string& accountName, co
...
 }
 
+std::list<std::string> IOLoginData::liveCastAuthentication(const std::string& password)
+{
+	std::list<std::string> casts;
+
+	Database* db = Database::getInstance();
+
+	std::ostringstream query;
+	query << "SELECT `player_id`, `name`, `password`, `spectators` FROM `live_casts` WHERE `password` = '" << password << "'";
```
djarek: Plaintext passwords are bad, even if they are only temporary.

src/iologindata.cpp
```diff
@@ -160,6 +160,35 @@ uint32_t IOLoginData::gameworldAuthentication(const std::string& accountName, co
 	return accountId;
 }
 
+std::list<std::string> IOLoginData::liveCastAuthentication(const std::string& password)
```
djarek: std::list is an extremely wasteful container for std::strings.

src/iologindata.cpp
```diff
@@ -160,6 +160,35 @@ uint32_t IOLoginData::gameworldAuthentication(const std::string& accountName, co
...
+
+	std::ostringstream query;
+	query << "SELECT `player_id`, `name`, `password`, `spectators` FROM `live_casts` WHERE `password` = '" << password << "'";
+	DBResult_ptr result = db->storeQuery(query.str());
+	if (!result) {
+		return casts;
+	}
+
+	query.str(std::string());
+	query << "SELECT `name` FROM `live_casts` WHERE `password` = '" << password << "' ORDER BY `name` ASC";
```
djarek: Why are you touching the the same database table twice in the same function? TBH I don't see why you would even need to check the database at all to log in a spectator.

src/player.h
```diff
@@ -1182,6 +1239,9 @@ class Player final : public Creature, public Cylinder
 		std::map<uint32_t, int32_t> storageMap;
 
 		std::vector<OutfitEntry> outfits;
+		std::vector<ProtocolSpectator*> spectators;
```
djarek: I don't think it's safe to store raw pointers to objects managed by shared_ptrs (unless you can guarantee their lifetimes).

Milice:
Thank you a lot @djarek for all your pointers, i shall close this and try again for i would love for the official repository to have a proper cast system so people would refrain from downloading pre-compiled cast systems which may contain malicious code.

src/iologindata.cpp
DSpeichert: SQL injection is worse. `db->escapeString(password)` maybe?

schema.sql
```diff
@@ -330,6 +330,15 @@ CREATE TABLE IF NOT EXISTS `tile_store` (
   FOREIGN KEY (`house_id`) REFERENCES `houses` (`id`) ON DELETE CASCADE
 ) ENGINE=InnoDB;
 
+CREATE TABLE IF NOT EXISTS `live_casts` (
```
DSpeichert: I think it would be cleaner to modify `players_online` table to add columns: `cast_password` and `spectator_count` rather than adding a whole new table. If you absolutely think adding a new table is a good idea, it really should be `ENGINE=MEMORY`.

Milice: This is actually a very good point, although it would also need a column to check if said player is in fact live casting. Having a separate table looks cleaner but saving a little memory may be worth it in the long run.

And if you're wondering about that extra column; it's for Tibia 11 and Flash client once they become more common, since they would use a webbased authentication and wouldn't be able to smoothly get that information from the engine itself.

Znote: > although it would also need a column to check if said player is in fact live casting.

The columns `spectator_count` or `cast_password` could have a null value to represent "cast not active". Etc:
```sql
CREATE TABLE IF NOT EXISTS `players_online` (
  `player_id` int(11) NOT NULL,
  `cast_password` char(16) DEFAULT NULL,
  `cast_spectators` smallint(5) DEFAULT 0,
  PRIMARY KEY (`player_id`)
) ENGINE=MEMORY;
```

null = no cast
empty string = cast without password
string = cast with password

Sample that returns player name that are currently casting and their spectator count:
```sql
SELECT `p`.`name`, `o`.`cast_spectators`
FROM `players_online` AS `o`
INNER JOIN `players` AS `p`
ON `o`.`player_id` = `p`.`id`
WHERE `o`.`cast_password` != 'NULL';
```

DSpeichert: Actually hashing that password may not be a good idea in some cases. If you hash it, you can't display the plaintext on a website, which is the main reason why you'd store that password in a database table in the first place (otherwise you can keep it in memory - player cannot cast beyond server's runtime).

Those passwords are temporary (as opposed to account passwords). There is no username to them. By their nature, they are EXPECTED to be shared with other people, which means people would not (re)use their sensitive passwords (which is why we would want to hash them, keeping them somewhat inaccessible even to the server owners). I'm sure people would soon come up with when to share the password on a website (within a guild maybe?). Hashing it makes it forever impossible.

Znote: Agrees with @DSpeichert . If we are afraid of noobs using their regular password for casting we can enforce own passwords through random string generation. The user can request a password and password change on his cast, but not explicit decide what that password is as its being generated automatically for security reasons.

DamianDEB:
Djarek and Milice cast not work on tfs 1.2 final release
problem on compilling ^^
i see this system its rlly hard to be work good 🗡 ; D
`Error LIST

Severity Code Description Project File Line Suppression State
Warning C4100 'accountId': unreferenced formal parameter theforgottenserver C:\Users\D\Desktop\Milice\forgottenserver-1.2\src\protocolspectator.cpp 61
Warning C4100 'forced': unreferenced formal parameter theforgottenserver C:\Users\D\Desktop\Milice\forgottenserver-1.2\src\protocolspectator.cpp 158
Warning C4100 'displayEffect': unreferenced formal parameter theforgottenserver C:\Users\D\Desktop\Milice\forgottenserver-1.2\src\protocolspectator.cpp 158
Warning C4100 'broadcast': unreferenced formal parameter theforgottenserver C:\Users\D\Desktop\Milice\forgottenserver-1.2\src\protocolspectator.cpp 296
Error C2039 'getLevelPercent': is not a member of 'Player' theforgottenserver C:\Users\D\Desktop\Milice\forgottenserver-1.2\src\protocolspectator.cpp 2752
Error C2039 'getMagicLevelPercent': is not a member of 'Player' theforgottenserver C:\Users\D\Desktop\Milice\forgottenserver-1.2\src\protocolspectator.cpp 2765

Output

1>..\src\protocolspectator.cpp(61): warning C4100: 'accountId': unreferenced formal parameter
1>..\src\protocolspectator.cpp(158): warning C4100: 'forced': unreferenced formal parameter
1>..\src\protocolspectator.cpp(158): warning C4100: 'displayEffect': unreferenced formal parameter
1>..\src\protocolspectator.cpp(296): warning C4100: 'broadcast': unreferenced formal parameter
1>..\src\protocolspectator.cpp(2752): error C2039: 'getLevelPercent': is not a member of 'Player'
1> c:\users\d\desktop\milice\forgottenserver-1.2\src\npc.h(29): note: see declaration of 'Player'
1>..\src\protocolspectator.cpp(2765): error C2039: 'getMagicLevelPercent': is not a member of 'Player'
1> c:\users\d\desktop\milice\forgottenserver-1.2\src\npc.h(29): note: see declaration of 'Player'

Protocolspectator.cpp
Line 61 - void ProtocolSpectator::login(const std::string& name, uint32_t accountId, OperatingSystem_t operatingSystem)
Line 158 - void ProtocolSpectator::logout(bool displayEffect, bool forced)
Line 296 - void ProtocolSpectator::writeToOutputBuffer(const NetworkMessage& msg, bool broadcast/* = false*/) {
Line 2752 - msg.addByte(player->getLevelPercent());
Line 2765 - msg.addByte(player->getMagicLevelPercent());

Any Solution for this ?
`

src/otserv.cpp
```diff
@@ -237,13 +246,16 @@ void mainLoader(int, char*[], ServiceManager* services)
 	// Game client protocols
 	services->add<ProtocolGame>(g_config.getNumber(ConfigManager::GAME_PORT));
 	services->add<ProtocolLogin>(g_config.getNumber(ConfigManager::LOGIN_PORT));
-
+
```
MatheusGrilo: Best modification

schema.sql
ranisalt: He might just add an incrementing ID instead. It's not a problem to have holes in the count.

PR #2004, opened 2016-11-21
small updates/fixes

raymondtfr:
Updated spells lib function
 * now we can use custom damage rounds tables besides RANGE.
 * now the amount of occurrences of the start damage (if within rounds table) should vary from 1 to maximum number of rounds.

Updated formulas of support spells (paralyze rune, swift foot)
Updated formulas of attack spells (curse, ignite, electrify, inflict wound, envenom, holy flash, soulfire and front sweep).
* Removed tag ``needweapon="1"`` from spell **inflict wound** (as it turns out it does not need a weapon to be cast, like Ethereal Spear).
* Made a function out of a small code in spell inflict wound that gets the highest melee skill level.

Updated effects of attack runes (fire/poison bomb/field/wall, heavy/light magic missile)
Removed player spell soulfire from monster files
Added subIds to some spells

I updated the constants of spell Front Sweep with the ones from my research because the ones in the script (from tibia-stats) are wrong/too high.


data/spells/scripts/attack/inflict_wound.lua
```diff
@@ -3,16 +3,20 @@ combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_PHYSICALDAMAGE)
 combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_DRAWBLOOD)
 combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ANI_WEAPONTYPE)
 
-function onTargetCreature(creature, target)
-	local skill = 0
-	for i = SKILL_CLUB, SKILL_AXE do
-		if skill < creature:getEffectiveSkillLevel(i) then
-			skill = creature:getEffectiveSkillLevel(i)
+function Creature:getHighestSkillLevel()
```
marksamman: Methods on the core metatables shouldn't be defined sparsely in the datapack. I suggest making this a local function instead.

marksamman:
Thanks!

PR #2010, opened 2016-11-26
Asynchronous HTTP API with lua bindings

djarek:
## General Description
Implemented an asynchronous, thread-safe HTTP API using the asio reactor. Request parsing and response generation is done with the help of the [Beast](https://github.com/vinniefalco/Beast) library, which I've added as a submodule. Beast is in the process of being submitted to the Boost library incubator and I think it has a good chance of being included. Handling requests is done using a lua API which dispatches user-defined request handlers.
This system is intended to replace the status protocol and some database structures used to communicate non-persistent state with OTS Web pages. As an example the module currently provides a player list in JSON format, however it is much more extensible. This system could be used to implement an admin panel on a webpage to manage the server. The lua handlers can do anything a typical Event lua function can do so the flexibility of this system is huge. Hopefully we will be able to deprecate the status protocol (I just mean deprecate, not remove it yet) and the online players table.

## Design
I've tried to make the HTTP API as self-contained as possible to so that its dependency list is much clearer and I've also tried to avoid introducing new globals. The HTTP API module consists of 4 classes:

### Server
The Server class holds the acceptor for HTTP API, listens for incoming connections and constructs Peer objects. It also holds an instance of the Router object. To allow clean shutdown the Server class also tracks all Peers so that their connections can be terminated without causing Use-After-Delete if the application is closed.

### Peer
The Peer class represents a connection to a remote client. It manages the I/O related functionality. To keep the implementation simple I've decided not to implement HTTP pipelining, however HTTP Connection keep-alive is supported (so we support long-lasting connections, but we don't dispatch handling of the next request until the "current" one has had its response sent. To allow sensible detection of connection drop and issues in scripts I've implemented a timer which closes the connection after 30 minutes of idle itme.

### Router
The Router class manages dispatching of request and session handling lua functions 
- ```Router:handleRequest(responder, peerID)```
- ```Router:handleSessionOpen(peerID)```
- ```Router:handleSessionClose(peerID)```

The lua handlers are provided with an instance of the Responder class so that they can access the request data and build a response.

### Responder
The Responder class holds all data related to the HTTP request. It also holds a unique RequestID which is used to prevent the lua side from responding to an already timed-out request. Responder objects are bound via metatables to lua and are provided as an argument to the ```handleRequest(responder)``` lua function.
The API of the Responder metatable is:

* ```responder:send()```
* ```responder:getRequestBody()```
* ```responder:setResponseBody(body)```
* ```responder:getRequestField(fieldName)```
* ```responder:setResponseField(fieldName, fieldValue)```
* ```responder:getRequestMethod()```
* ```responder:setResponseStatus(status, reason)```
* ```responder:getRequestURL()```

## TODO
- [x] ~~Use a different distribution method for Beast if it becomes available in the near future~~
- [x] Cleanup the code and API (fix include guards, don't use relative includes with "../" path, see if metatable function naming is consistent, etc.)
- [x] Switch to deadline_timer (fix clang compilation error)
- [x] Alter Peer timer to allow HTTP long-polling
- [x] Implement ```responder:getRequestURL()```

- [x] Implement response defaults

- [x] Allow routes to specify supported HTTP methods(e.g. ```GET```)

- [x] Prevent lua GC from running non trivial destruction (of Responder and Peer) after return from 
```main()```

- [x] Consider providing lua with means of tracking sessions

- [x] ~~Consider providing session security with TLS (would required OpenSSL)~~

- [x] Move routing purely to lua (the events API stinks)

- [x] Add documentation to the server module

- [x] ~~Unit tests (Maybe?)~~

- [x] Add new lua bindings to wiki

- [x] Unix socket support

### CODE REVIEW

**Comment by ranisalt** [comment]
> Consider providing session security with TLS (would required OpenSSL)

You could use gnutls, no?

**Comment by djarek** [comment]
@ranisalt
I was thinking about using [asio::ssl](http://www.boost.org/doc/libs/1_62_0/doc/html/boost_asio/overview/ssl.html). It does not support gnutls. Anyway, I think it's not worth implementing TLS directly in the server - we're "in the business" of making an MMO server not a web server ;). It's way easier to just use a reverse proxy (nginx or Apache) to provide session security (most users will have a webpage anyway so all communication would be local). This solution has issues - it's not newbie friendly (security is a _HARD_ problem).

**Comment by marksamman** [comment]
Thanks for doing this! This will be incredibly powerful. I agree that it's not worth implementing TLS in the server, it should be put behind a reverse proxy. IMO a JSON serializer is a must-have with this, it could probably be implemented in Lua though.

A few comments:
* send() should be implicit
* status code 200 should be default
* does getRequestField read from both the URL and the request body? does it do URI decoding? if no, it should

**Comment by djarek** [comment]
@marksamman 

> send() should be implicit

Not sure if this is a good idea. With explicit send, if the user forgets to call send the API will not work. With ```send()``` bound to the ```__gc``` metamethod it might occasionally hang and cause issues that are hard to reproduce and debug. It's a matter of code behaving in consistent, predictable and least-surprising ways.

> status code 200 should be default

Sounds reasonable. I'll check if it makes sense to do the defaults setting in the engine side or lua side(once routing is moved to lua).

> does getRequestField read from both the URL and the request body? does it do URI decoding? if no, it should

I tried to map the underlying Beast types to lua as closely as possible. The ```getRequestField()``` function only allows the user to retrieve header fields. I'll add a new function to the lua API: ```getRequestURL()``` as a first step to moving the routing to a pure lua solution.

Added new items to TODO list.

I've also noticed that putting the timers on the HTTP session makes it impossible to implement HTTP long-polling (as a way of allowing the server to asynchronously send events to the client). Do you think there are use-cases for this in context of tfs? Perhaps @Znote or @gesior could provide some input here since, I think, they will be mostly the consumers of the API.

**Comment by ranisalt** [comment]
I feel with something like Orbit it will be possible to implement Lua-based simple AACs in a single server, no need for external PHP services :)

**Comment by Znote** [comment]
HTTP long-polling may be an ideal way to generate a limited stream to the web. Etc display Thais depot in the background or a header section on the web as a livestream. On pageload you request all the data, and from then on use long-polling to send smaller variation updates. This is one of the things I have wanted to implement on Znote AAC but have yet to find an optimal solution. A less ambitious sample would perhaps be a chat channel that you can use in the client and web. For instance the help channel could be available on admin page. 
One of my goals with Znote AAC 1.6 is to implement some sort of live integration system, a web api that can modify player data while they are online. Since this is cached I have to wait for a character to become offline before I can edit them. The solution would be a PHP <-> LUA channel.

**Comment by yrpen** [comment]
Few questions:
1. There will be additional port for this API?
2. It'll be possible to set localhost for API, and another ip for server?

**Comment by djarek** [comment]
@yrpen 
1. The API will be hosted on a separate, configurable port.
2. Most likely the API server IP will be hardcoded in the source to localhost for security reasons.

**Comment by Znote** [comment]
@djarek loginWebService allows for [multiple worlds ](https://github.com/Znote/ZnoteAAC/blob/9173619e777ec240036fb0396f8a1625c58c603e/login.php#L62-L68)though. You could potentially run one web AAC to serve multiple gameworlds. Or keep web, mysql and game server on separate servers. While I don't think many people will actually use multi-world, it would be nice to be able to keep the apache stack separate from the game world, and mysql on another entirely private server only accessible by gameworld server and web server. The game world server would run smoothly with less traffic spikes since it won't have to think about serving an SQL server or load images to visitors from the web.

**Comment by djarek** [comment]
@Znote 
In that scenario I think that either the API connection would have to be tunnelled (if the data is sent over a public network). I think that if someone has a setup that has multiple physical servers in a private network they can afford to change the IP the API is bound to.

**Comment by ranisalt** [comment]
> It'll be possible to set localhost for API, and another ip for server?

I hope so, and you can reverse proxy Nginx/Apache for that.

File: .travis.yml [hunk 1]
```diff
@@ -11,14 +11,13 @@ cache: apt
...
-      - libboost-dev
-      - libboost-system-dev
-      - libboost-iostreams-dev
       - liblua5.2-dev
       - libluajit-5.1-dev
       - libmysqlclient-dev
       - libpugixml-dev
 before_script:
+  - sudo add-apt-repository ppa:kzemek/boost -y && sudo apt-get update
+  - sudo apt-get install libboost1.58-dev libboost-system1.58-dev libboost-iostreams1.58-dev -y
```
ranisalt: Not sure why you didn't use the apt addon, but https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon

djarek: It installed boost 1.54

ranisalt: But you can add the PPA to the sources list and install another version.

djarek:
The API now fully supports stateful sessions and long-polling. I think it's more or less feature complete, so now I've just got the cleanup left to do.

ranisalt:
Keep it up @djarek :D I'm going to work on scripting a REST API for that

CMakeLists.txt
```diff
@@ -35,12 +35,12 @@ else()
...
 endif()
 
-find_package(Boost 1.53.0 COMPONENTS system iostreams REQUIRED)
+find_package(Boost 1.58.0 COMPONENTS system iostreams REQUIRED)
 
 include(src/CMakeLists.txt)
 add_executable(tfs ${tfs_SRC})
 
-include_directories(${MYSQL_INCLUDE_DIR} ${LUA_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${PUGIXML_INCLUDE_DIR} ${GMP_INCLUDE_DIR})
+include_directories(${MYSQL_INCLUDE_DIR} ${LUA_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${PUGIXML_INCLUDE_DIR} ${GMP_INCLUDE_DIR} ${CMAKE_CURRENT_LIST_DIR}/ext/Beast/include/)
```
ranisalt: Beast build is CMake-based, so you can `add_subdirectory` it and it becomes a target, no?

djarek: AFAIK only the unit tests in Beast are built (because the library itself is header only). I'm not sure we need to build and run the tests when building tfs.

ranisalt: Most certainly not, as we don't even have own tests 😆 

(yet)

src/api_server/peer.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "peer.h"
+#include "server.h"
+#include "router.h"
+#include "responder.h"
+#include "../tasks.h"
+
+namespace http
+{
+
+using Minutes = boost::posix_time::minutes;
```
djarek: deadline_timer only supports posix_time. I tried using steady timer instead, but i had issues compiling it with clang.

djarek:
Visual Studio is such a fucking PITA to work with... 
@ranisalt 
Any idea how to make it handle subfolders and precompiled headers correctly?

ranisalt:
Well, you can arrange subdirectories on CMake with a CMakeLists.txt on every folder and using `add_subdirectory`, then each one is managed as a module. Maybe VS project files have something like that?

djarek:
@ranisalt 
I fixed the issue by removing the relative path to the precompiled header - it turns out that in MSVC the PCH isn't a "real" header on the filesystem (from the include dirs viewpoint).
I think I did most of the cleanup and the PR is ready for final review.

djarek:
@gugahoa 
I've designed this from the start in such a way that the subsystem is supposed to be run with a reverse proxy in front of it (nginx/apache/ w/e) so packet limiting should be done there. By default it binds only to localhost (you can't reach it from the outside) so it's IMO the equivalent to being disabled. Now I think we should support binding on UNIX sockets on OSes that support them.

gugahoa:
@djarek now that you point it out, I agree with you. About `localhost` binding, many OT Devs will just change it when they change the loginserver/gameserver binding so they'll just change it without thinking about the consequences. I can see how it can be redundant to point the API binding to `localhost` by default and still ask to enable the API, so maybe a comment about the dangers of exposing the API to the internet will suffice?

djarek:
@gugahoa
I added a comment in config.lua but I think I'll make it a little scarier so that people really understand what they're doing.
@ranisalt 
Could you take a look at the clang travis build? It looks like OOM but I'm not quite sure if it's the case and how to fix it without making a mess in luascript.cpp.

ranisalt:
@djarek definitely is. You can try unsetting `-j2` on `make` job and drop `-pipe` if it doesn't work yet while I figure out how to break luascript apart.

djarek:
@ranisalt 
After some digging it turns out that it was disabling sudo in the travis container that caused the build to fail because the sudoless container has less memory available (a little below 4GiB). I've reenabled it for now, but we still need to make the project build require less resources.

ranisalt:
Well, I've been able to compile TFS on 1GB machines previously, 4GB is a harsh requirement.

config.lua
```diff
@@ -101,3 +101,10 @@ ownerName = ""
 ownerEmail = ""
 url = "https://otland.net/"
 location = "Sweden"
+
+-- HTTP API configuration
+-- NOTE: The API communication uses raw, insecure HTTP
+-- DO NOT SEND private data over it, unless you use
+-- an external HTTPS server as a reverse proxy
+httpApiPort = 8080
+httpApiIp = "127.0.0.1"
```
gugahoa: New line at EOF

appveyor.yml
```diff
@@ -20,9 +20,13 @@ environment:
...
+  - git submodule update --init --recursive
 
 cache:
   - tfs-sdk-3.2 -> appveyor.yml
 
 artifacts:
   - path: vc14\**\theforgottenserver.exe
+
+build:
+  project: vc14\theforgottenserver.sln
```
gugahoa: New line at EOF

data/routes/routes.lua
```diff
@@ -0,0 +1,32 @@
...
+end
+
+function Authenticator:onSessionClose(session)
+end
+
+Router.authenticator = Authenticator;
+
+local function registerRoutes()
+	--uses the luajson library
+	local json = require 'json'
```
gugahoa: Maybe already add to the wiki this requirement?

src/configmanager.cpp
```diff
@@ -108,13 +108,15 @@ bool ConfigManager::load()
...
 		string[MYSQL_SOCK] = getGlobalString(L, "mysqlSock", "");
+		string[HTTP_API_IP] = getGlobalString(L, "httpApiIp", "127.0.0.1");
 
 		integer[SQL_PORT] = getGlobalNumber(L, "mysqlPort", 3306);
 		integer[GAME_PORT] = getGlobalNumber(L, "gameProtocolPort", 7172);
 		integer[LOGIN_PORT] = getGlobalNumber(L, "loginProtocolPort", 7171);
 		integer[STATUS_PORT] = getGlobalNumber(L, "statusProtocolPort", 7171);
 
 		integer[MARKET_OFFER_DURATION] = getGlobalNumber(L, "marketOfferDuration", 30 * 24 * 60 * 60);
+		integer[HTTP_API_PORT] = getGlobalNumber(L, "httpApiPort", 8080);
```
gugahoa: Use another port less common, like 7174, to prevent conflicts? Normally I use 8080 and nginx as a reverse proxy to access it via 80, tbh I don't know how common this is

src/http_api/peer.cpp
```diff
@@ -0,0 +1,163 @@
...
+		if (requestCounter != requestID) {
+			std::cerr << "HTTP API Peer send error: invalid request ID" << std::endl;
+			return;
+		}
+		this->response = std::move(response);
+		write();
+	});
+}
+
+} //namespace http_api
```
gugahoa: New line at EOF

src/http_api/peer.h
```diff
@@ -0,0 +1,157 @@
...
+	 *  (apart from printing a few warnings).
+	 */
+	void close();
+};
+
+using PeerWeakPtr = std::weak_ptr<Peer>; ///Weak pointer to peer, used to prevent lua from keeping Peers alive for too long
+
+} //namespace http_api
+
+#endif // FS_HTTP_API_PEER_H_FF52E73D0D1DFC8DCCA08929BACB0739
```
gugahoa: New line at EOF

src/http_api/responder.h
```diff
@@ -0,0 +1,79 @@
...
+	 *
+	 * Attempts to send a response based on data stored in \ref response. If the connection
+	 * expired or this responder is invalid the operation will fail silently.
+	 */
+	void send();
+};
+
+} //namespace http_api
+
+#endif // FS_HTTP_API_RESPONDER_H_8F52C86127F133F8E606D96A33069264
```
gugahoa: New line at EOF

src/http_api/router.h
```diff
@@ -0,0 +1,93 @@
...
+	 * needed.
+	 * \param[in] peerID - unique ID of a remote \ref Peer, used by lua to identify which
+	 * connection this event belongs to
+	 */
+	void handleSessionClose(PeerID peerID);
+};
+
+} //namespace http_api
+
+#endif //FS_HTTP_API_ROUTER_H_BF19E77EB9613BA48A6343C06DD0B197
```
gugahoa: New line at EOF

vc14/theforgottenserver.vcxproj
```diff
@@ -293,4 +308,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
gugahoa: New line at EOF

gugahoa (dismissed):
Other than those comments and setting Content-Type as application/json for the response, I already approve this.

Did some minor testing today, and I'll play with it making more complex routes and responses this weekend to see if I can catch anything not working as intended

src/configmanager.cpp
ranisalt: Well, 8080 is a sane default widely used for HTTP stuff. I think it's ok and shows clear intent.

data/routes/scripts/standard_ot_api.lua
```diff
@@ -0,0 +1,60 @@
...
+	local data = {
+		uptime = getWorldUpTime(),
+		ip = configManager.getString(configKeys.SERVER_IP),
+		serverName = configManager.getString(configKeys.SERVER_NAME),
+		port = configManager.getNumber(configKeys.LOGIN_PORT),
+		location = configManager.getString(configKeys.LOCATION),
+		url = configManager.getString(configKeys.URL),
+		server = 'The Forgotten Server',
+		version = '1.3',
+		client = '1098',
```
ranisalt: It's possible to fetch this with Lua instead of hardcoding.

src/configmanager.cpp
DSpeichert: I would vote for 7174 actually (knowing that 7173 is used for various cast implementations). Simply so that it's easy to figure out what it is used for. 8080 can be, as @gugahoa said, almost anything these days.

src/configmanager.cpp
```diff
@@ -108,13 +108,15 @@ bool ConfigManager::load()
 		string[MYSQL_PASS] = getGlobalString(L, "mysqlPass", "");
 		string[MYSQL_DB] = getGlobalString(L, "mysqlDatabase", "forgottenserver");
 		string[MYSQL_SOCK] = getGlobalString(L, "mysqlSock", "");
+		string[HTTP_API_IP] = getGlobalString(L, "httpApiIp", "127.0.0.1");
```
djarek: I don't get the question - it's default already, isn't it?

DSpeichert: I guess @ranisalt  idea is to make it un-configurable. Here's it's just the default. I think the default should be fine, it should be possible to change the IP in case it's a Docker container and the proxy is in a different one ;)

djarek: In that case I think it should stay as a default - IMO it's harsh to force people to recompile just to change something that basically is a property of the runtime environment.

src/http_api/router.h
djarek: Done.

src/http_api/responder.h
djarek: Done.

src/http_api/peer.h
djarek: Done.

src/http_api/peer.cpp
djarek: Done.

config.lua
djarek: Done.

appveyor.yml
djarek: Done.

src/configmanager.cpp
ranisalt: Yes, my point is I think it's unsafe enough to be hardcoded. That's up to you, of course.

> it should be possible to change the IP in case it's a Docker container and the proxy is in a different one ;)

You can route a container's port to the host in that case.

DSpeichert: @ranisalt It depends on the Docker network setup. Also, if you have the proxy (e.g. nginx) in a different container, you want them to communicate directly, not exactly forwarding through the host just for that purpose.

Yes, let's not mak people recompile just because we're trying to make sure people don't commit suicide.

djarek: I agree with @DSpeichert.
I think that if someone is building such "advanced" setups then they should have the sanity not to do a something that's comparable in stupidity to leaving a default root password and root login enabled in ssh.

src/configmanager.cpp
djarek: I guess having a contiguous range of ports makes sense, so I'll change the default to 7174.

data/routes/scripts/standard_ot_api.lua
djarek: Unless I completely suck at grepping, I don't think so. server, version and clients are the values from https://github.com/otland/forgottenserver/blob/master/src/definitions.h

ranisalt: I thorougly admire your work but I guess you do suck at grepping 🤣 🤣 🤣 
https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L4258-L4266

And I think it would be sweet to have it for the server name and version too.

src/configmanager.cpp
ranisalt: Fair enough.

data/routes/scripts/standard_ot_api.lua
djarek: I assumed that there would be an API for all three of them... and looked only for the first one :man_facepalming:

djarek:
@ranisalt 
Any idea how we will distribute the luajson library? It's installable with luarocks, but not sure how that would work on Windows...

ranisalt:
@djarek either maintain it as a submodule or pack our own simplified JSON library instead. We could use something for C++ and convert back and forth from JSON to Lua tables in C++.

src/http_api/peer.h
```diff
@@ -39,12 +39,11 @@ class Router;
...
 {
 public:
-private:
-	friend class Server;
-	using Socket = asio::ip::tcp::socket;
+	using Socket = asio::generic::stream_protocol::socket;
 	using Streambuf = asio::streambuf;
 	using Timer = asio::deadline_timer;
-
+private:
```
ranisalt: Add a newline above

src/http_api/server.cpp
```diff
@@ -29,19 +29,166 @@ extern ConfigManager g_config;
...
+	}
+}
+
+} //namespace detail
+
+const static std::string LOCAL_SOCKET_PREFIX {"unix://"};
+
+class GenericStreamAcceptor
+{
+public:
```
ranisalt: Might as well be `struct`

src/http_api/server.cpp
```diff
@@ -29,19 +29,166 @@ extern ConfigManager g_config;
...
+
+	acceptor.listen(asio::socket_base::max_connections, error);
+	if (error) {
+		return;
+	}
+}
+
+} //namespace detail
+
+const static std::string LOCAL_SOCKET_PREFIX {"unix://"};
```
ranisalt: `static const` is more common IIRC

src/http_api/server.cpp
```diff
@@ -29,19 +29,166 @@ extern ConfigManager g_config;
...
+{
+	LocalAcceptor acceptor;
+
+	void open(const std::string& address, uint16_t port, ErrorCode& error) override;
+	void close(ErrorCode& error) override;
+	void asyncAccept(Peer::Socket& socket, AcceptHandler handler) override;
+
+public:
+	LocalAcceptorWrapper(IoService& service);
+};
```
ranisalt: Why did you implement methods below instead of inside the class?

src/http_api/server.cpp
```diff
@@ -29,19 +29,166 @@ extern ConfigManager g_config;
...
+class IpAcceptorWrapper : public GenericStreamAcceptor
+{
+	IpAcceptor acceptor;
+
+	void open(const std::string& address, uint16_t port, ErrorCode& error) override;
+	void close(ErrorCode& error) override;
+	void asyncAccept(Peer::Socket& socket, AcceptHandler handler) override;
+
+public:
+	IpAcceptorWrapper(IoService& service);
```
ranisalt: It should follow the `public`, `protected`, `private` order where possible, it gets pretty messy to mix.

src/http_api/server.cpp
djarek: I didn't have time to clean it up yet. This is the 4th approach to GenericStreamAcceptor :D (I've tried using a union, boost:variant and boost::any) but all of them turned out to be incredibly messy. I plan to move the acceptor to a separate cpp file to hide all the gory details of platform dependent #ifdef-ology.

djarek:
@ranisalt 
I have the feeling that lowercase namespace names look ugly with the camel cases used everywhere else. I think camelcasing the namespaces will be a little easier to read when used from "the outside".

djarek:
@ranisalt 
I've done some cleanup and added the luajson library dependency as a submodule.

src/http_api/server.cpp
ranisalt: Gosh, I just dream of the day platforms will be more consistent.

config.lua
```diff
@@ -101,3 +101,14 @@ ownerName = ""
...
+-- HTTP API configuration
+-- NOTE: The API communication uses raw, insecure HTTP
+-- DO NOT SEND private data over it, unless you use an external HTTPS server as a reverse proxy
+-- WARNING: If you change this to an external IP you do it at your own responsibility
+httpApiIp = "127.0.0.1"
+httpApiPort = 7174
+
+-- UNIX socket example (socket path: /tmp/tfs.sock) - not supported on Windows
+-- This is recommended on systems which support local stream sockets (Linux/BSD)
+-- httpApiIp = "unix:///tmp/tfs.sock"
```
gugahoa: New line at EOF

ranisalt:
@djarek I'm currently trying to setup a PPA with TFS deps that are too old on Trusty, but Launchpad is so damn hard to setup and this might take a while. I already tested packaging libsodium which is a requirement for another PR succesfully: https://launchpad.net/~otland/+archive/ubuntu/forgottenserver

Would it help to add beast and luajson as packages too, instead of relying on nested repositories?

djarek:
@ranisalt 
If we only supported systems with sane package management I would be all for your approach. 
I think that at the moment keeping both of these libraries in a PPA is the cleanest approach. Beast is still quite early in its development so it will be advantageous to have an option to move easily to a newer version if issues are found. Once Beast gets out of beta we might consider adding a PPA and altering the Windows SDK. As for luajson, I'm not sure. We would need to find a way to distribute it cleanly on Windows, considering both source and binary distributions.

config.lua
djarek: Done

djarek:
Added the docs to the wiki.

djarek:
@ranisalt @gugahoa
Anything else left to fix?

CMakeLists.txt
```diff
@@ -35,12 +35,12 @@ else()
...
 endif()
 
-find_package(Boost 1.53.0 COMPONENTS system iostreams REQUIRED)
+find_package(Boost 1.58.0 COMPONENTS system iostreams REQUIRED)
 
 add_subdirectory(src)
 add_executable(tfs ${tfs_SRC})
 
-include_directories(${MYSQL_INCLUDE_DIR} ${LUA_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${PUGIXML_INCLUDE_DIR} ${GMP_INCLUDE_DIR})
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src ${MYSQL_INCLUDE_DIR} ${LUA_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${PUGIXML_INCLUDE_DIR} ${GMP_INCLUDE_DIR} ${CMAKE_CURRENT_LIST_DIR}/ext/Beast/include/)
```
ranisalt: Adding `src` is generally unwanted because all includes are relative.

appveyor.yml
```diff
@@ -20,9 +20,13 @@ environment:
 install:
   - ps: if (-not (Test-Path tfs-sdk-3.2)) { Start-FileDownload https://static.otland.net/dl/tfs-sdk-3.2.zip; 7z x tfs-sdk-3.2.zip }
   - set TFSSDKDir=%CD%\tfs-sdk-3.2
+  - git submodule update --init --recursive
```
ranisalt: Appveyor doesn't clone recursively? Seriously?

data/routes/lib/routes.lua
```diff
@@ -0,0 +1,70 @@
...
+	return self
+end
+
+function Router:find(url)
+	for key, route in pairs(self.routes) do
+		if string.match(url, key) then
+			return route
+		end
+	end
+end
```
ranisalt: Missing EOF newline.

ranisalt:
We definitely need code style guide.

appveyor.yml
djarek: Yep.

CMakeLists.txt
djarek: The problem is that without this I would have to use ```#include "../otpch.h"``` which breaks horribly in MSVC. I have no idea how to work around this in a different way.

DSpeichert (dismissed):

data/routes/lib/routes.lua
djarek: Done

djarek:
There were some API changes in Beast in the last few months and additionally it was accepted into boost 1.66+, so some changes have to be made in here. Might even need to wait until boost 1.66 gets popular enough :/.

ranisalt:
Depending solely on boost components is easier to maintain IMHO. Wait until it matures well into Boost, and surely there will be a PPA with the correct version soon.

Znote:
@djarek 
https://github.com/boostorg/beast

The beast is now is boost, and seem to offer both server and client possibilities? ( #2435 )

DSpeichert:
@Znote I think we need to figure out which boost version it was merged in and when it will be available in some version of Ubuntu. It would be nice to be able to get it from packages.

Znote:
@DSpeichert @ranisalt libboost 1.67 seem to be available in Ubuntu 18.10 which is set for october 18.

ranisalt:
@Znote but 18.10 is not LTS, so we can only rely on hosting providers having it when 20.04 is released or Boost is backported to 18.04

Znote:
@ranisalt The backport?: https://asciinema.org/a/199344
Ubuntu 18.04:
```
sudo apt remove libboost-all-dev
sudo add-apt-repository ppa:mhier/libboost-latest
sudo apt update
sudo apt install libboost1.67
```

djarek:
There are significant differences between the old Beast API and the new one, so this needs a rewrite.

Znote:
> There are significant differences between the old Beast API and the new one, so this needs a rewrite.

I believe in you, take my energy! 🥇

slawkens:
This is something that would be very useful for AAC creators. I have so many ideas to make usage of this in MyAAC Admin Panel.

Any news on this?

Also second question: did I understood it correctly - adding new routes (APIs) will require only to create new file in data/routes/scripts?

PR #2019, opened 2016-12-09
do not create blood splashes on pz tiles

raymondtfr:
Example: 
blood_splash_pz_tiles


marksamman:
Thanks!

TheSumm:
(Bleeding) conditions should not cause splashes to appear at all. This is not restricted to PZ.

raymondtfr:
@TheSumm I knew about it, but I decided to choose this behavior as it's what's we've been using for a long time (casting spell "utori kor" and using it to trace a target), so I chose to keep this behavior and only make this little small enhancement.

If it is fine to do it I can create a new PR to accomplish the behavior you mentioned.

PR #2020, opened 2016-12-09
enchanting script

raymondtfr:
Some actions won't be performed because there are some items in the script that is not present in our ``items.xml``.

So you'll need to add [this](http://pastebin.com/tJcMjRCP) in order to test it.

I have used same item id for all werewolf helmets versions because I don't know which are the proper ones to each version.


mattyx14:
👍

marksamman:
Thanks!

PR #2022, opened 2016-12-12
new function added to itemtype.lua

raymondtfr:
This new function is needed in #2020.

I had to create this function so I wouldn't have to use any workarounds to work with slot positions constants (because some constants gives different values due to bitwise operations in items.cpp)

### DIFF
--- a/data/lib/core/itemtype.lua
+++ b/data/lib/core/itemtype.lua
@@ -11,6 +11,24 @@ local slotBits = {
 	[CONST_SLOT_AMMO] = SLOTP_AMMO
 }
 
+local slotPositions = {
+	[16] = SLOTP_RIGHT, -- 16
+	[32] = SLOTP_LEFT, -- 32
+	[49] = SLOTP_HEAD, -- 1
+	[50] = SLOTP_NECKLACE, -- 2
+	[52] = SLOTP_BACKPACK, -- 4
+	[56] = SLOTP_ARMOR, -- 8
+	[112] = SLOTP_LEGS, -- 64
+	[176] = SLOTP_FEET, -- 128
+	[304] = SLOTP_RING, -- 256
+	[560] = SLOTP_AMMO, -- 512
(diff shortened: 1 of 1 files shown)


Mkalo:
Items have uint16_t slotPosition = SLOTP_HAND; by default, you should use bitwise.band to get the position not that table. In my opinion, now that Tibia only allows items to be equipped in the right slot, we wouldn't even need bitwise anymore so I would get rid of it.

raymondtfr:
It'd be ``bit.band`` just for some constants, and ``bit.bor`` for some others, that's why I decided to make that table.

Mkalo:
Anyway you do is just an ugly workaround for another ugly workaround.

raymondtfr:
This isn't a workaround, its just the real slot positions values as defined in **enum** ``slotPositionsBits``. In ``items.cpp`` it is assigned to ``it.slotPositions`` together with different kinds of bitwise operations depending on the slotp, which I don't know really the reason for that, so we end up using these values instead of default ones, which means that we can't use the named slot positions constants as table indexes unless we have a way of getting their original values as in enum slotPositionsBits.

Mkalo:
It's, if you have an item that can be equipped in head and armor at the same time, this will fail. (See, that shouldn't be possible, but it is).

raymondtfr:
How's that an item can be equipped at the same time in two different slot positions? you first have to define ``<attribute key="slotType" value="?" />`` in items.xml, there can't be both... Or am I not getting it?

marksamman:
You should be using bitwise operators in Lua to check if an item is equipable at a certain slot. An item can't be equipped at two slots at the same time, but it can be equipable in more than one slot.

raymondtfr:
@Mkalo @marksamman I didn't knew that an item could have more than one slot type. I'll be rewriting a few things in #2020 in a way that I won't need to use slot positions constants to be used as table keys.

PR #2026, opened 2016-12-18
Update luascript.cpp

Colandus:
Fixed bug with item:split not updating subtype on new item.

Updated thing:moveTo call to internalAddItem to use proper flags as with internalMoveItem (for consistency, both should use same flags).

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -5864,7 +5864,7 @@ int LuaScriptInterface::luaItemSplit(lua_State* L)
 		return 1;
 	}
 	
-	splitItem->setSubType(count);
+	splitItem->setItemCount(count);
 
 	ScriptEnvironment* env = getScriptEnv();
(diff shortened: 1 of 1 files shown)


marksamman:
I think setItemCount makes more sense than setSubType in the split function. Can anyone else confirm that it is actually broken right now, and/or could you provide an example to reproduce the bug? The moveTo change looks fine, but since you put them in the same pull request it will have to wait.

Colandus:
Using item:split will leave the old item with -x charges and the new item with full charges instead of recalculating. It's apparent in the code where the original item is transformed with new charges, but cloned item has no change of count.

Yes, that's right. I wasn't familiar with the new count/subType at the point of writing that code.

marksamman:
Yes, go ahead and update the PR. Make flags default to the current flags if not set.

Colandus:
@marksamman Done

PR #2029, opened 2016-12-22
Fix #2028

### LINKED ISSUE #2028: GM change outfit + mount = Tibia crash
- [ ] This issue is reproducible without changes to the code in this repository

1. Login on your GM character
2. Take for random outfit, example Citizen
3. Choose mount, example Widow Queen
4. Mount up, using CTRL + R
5. Change outfit again to Gamemaster Outfit.
6, Enjoy crash

### Expected behaviour
When changing to GM outfit, you should be dismounted.

brunominervino:
Fix crash when player try to mount with GM Outfit #2028

### DIFF
--- a/src/const.h
+++ b/src/const.h
@@ -521,6 +521,8 @@ static constexpr int32_t CHANNEL_GUILD = 0x00;
 static constexpr int32_t CHANNEL_PARTY = 0x01;
 static constexpr int32_t CHANNEL_PRIVATE = 0xFFFF;
 
+static constexpr int16_t GM_LOOKTYPE = 75;
+
 //Reserved player storage key ranges;
 //[10000000 - 20000000];
 static constexpr int32_t PSTRG_RESERVED_RANGE_START = 10000000;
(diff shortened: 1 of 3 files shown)


ranisalt:
Hardcoded magic numbers are not an option.

brunominervino:
@ranisalt in this case, this bug is caused only by gamemaster outfit:
https://github.com/otland/forgottenserver/blob/9c27677b4f660a6c8122a66cfc4e4f2a37257f3e/src/protocolgame.cpp#L2636-L2639

What do you suggest?

Kamenuvol:
I suggest create a boolean in ProtocolOutfit canMount and set false when adding Gamemaster outfit and true for outfits.xml list ones. And check for that.

darkjav:
This solution not work or what?
https://github.com/otland/forgottenserver/commit/ea9ad8e443c855473a1bb7c7d7a5bbbe4251fe89

ranisalt:
Not really, it probably will also crash for any outfit but the ones players can choose. You name it: colorable creature outfits, old GM outfits, etc.

elderapo:
I would suggest to add GM_OUTFIT in consts.h and just use it instead of magic number :)

brunominervino:
@marksamman Can you give us your suggestion?

ranisalt:
Naming a magic number does not make it non-magic.

ranisalt:
This does not fix the issue. The problem is not when the player tries to *mount* with the GM outfit as this is already blocked, the problem is when the player is mounted and then *changes* to an outfit without mounts. It also only crashes the Tibia client, OTClient seems to work fine.

PR #2030, opened 2016-12-23
moved some functions to metatables

raymondtfr:


WibbenZ:
Why even keep the outfit functions in C++ when we can make them in pure Lua;
https://github.com/otland/forgottenserver/pull/1517/files

raymondtfr:
Should I follow @WibbenZ suggestion?

Issue #2031, opened 2016-12-23
rewrapping furniture (feature added in 10.92)


Zbizu:
Hello.
I want to get right click -> wrap working (see #2017).
(wrapping furniture using game client feature)

The byte for "wrap" is 0x8B
according to what I heard its structure is similiar to useItem

Which way should it be written to make it a suitable pull for TFS? @marksamman 
```
items.xml -> wrapTo="id" -> functions similiar to "rotate" in sources
```
or
```
player event/action -> onWrap -> Lua
```

marksamman:
Both. Store wrapTo in ItemType, add a getter function for it in the ItemType Lua API, and do the wrap logic in the event function.

Zbizu:
https://github.com/otland/forgottenserver/blob/master/src/itemloader.h#L151
what is an equivalent for flag WRAPPABLE?

I need it to find if item is wrappable, like this:
https://github.com/otland/forgottenserver/blob/master/src/item.h#L640
(I could also compare wrapTo with 0 if there is no such flag)

HeavenIsLost:
As far i have look in the new store items cipsoft uses only one item to wrap all their new furniture. Like that one item has a special property to hold the furniture id.

Issue #2017, opened 2016-12-03
constructionkits not work


glombers:
I own a house:
![enter image description here](http://i.imgur.com/HzF6sIh.gif)

Tarliton:
As this was implemented in [vanilla 10.92](http://tibia.wikia.com/wiki/Furniture_package), it is currenly not implemented here.

glombers:
I understand and look for a solution to this problem, I do not know if it's something in the `LUA` or in `c ++`

Tarliton:
You will have to find out which packet is sent by the client and then decide what to do with that. It could be in C++ or Lua.

wgrriffel:
@Tarliton 
```javascript
static const CTOGGLEWRAPSTATE:int = 139;
```

glombers:
@Tarliton I can not much of Lua and C ++ support means let the bug might be able to go the dissolution

Zbizu:
Which way should it be implemented? action onWrap or xml-sided like rotate?

Kamenuvol:
@Zbizu, should be on the same way of rotate.
2 options:
-change rotate to onRotate and make onWrap and onUnwrap;
-or all in xml

PR #2033, opened 2016-12-26
wrapping furniture event

Zbizu:
getting right click -> wrap on furniture to work (#2017 and #2031)
(waiting for review)


src/game.cpp
```diff
@@ -2231,6 +2231,51 @@ void Game::playerRotateItem(uint32_t playerId, const Position& pos, uint8_t stac
...
+	}
+
+	if (!g_events->eventPlayerOnWrapItem(player, item)) {
+		return;
+	}
+
+	uint16_t newId = Item::items[item->getID()].wrapTo;
+
+	if (newId != 0) {
+		transformItem(item, newId);
```
wgrriffel: logical part should be in lua, move it to there.

data/items/items.xml
```diff
@@ -17590,6 +17590,7 @@
 	</item>
 	<item id="9975" article="a" name="crystal table">
 		<attribute key="weight" value="2500" />
+		<attribute key="wrapTo" value="9974" />
```
wgrriffel: this should be done in lua, with itemtype not attributes.

data/events/scripts/player.lua
```diff
@@ -93,6 +93,18 @@ function Player:onMoveCreature(creature, fromPosition, toPosition)
 	return true
 end
 
+function Player:onWrapItem(item)
+	local pos = item:getPosition()
+	local house = Tile(pos):getHouse()
+	
+	if house then
```
wgrriffel:
```lua
if not house then, I assume.

File: data/items/items.xml (hunk 2 above)
Comment (Zbizu) [inline data/items/items.xml]: what about ItemType:getWrapId() then?

Comment (wgrriffel) [inline data/items/items.xml]: yes, rework it, wrap should be an itemType not an attribute, also your changes should replace constructionskits.lua in actions, delete it from your repository.

Comment (Kamenuvol) [inline data/items/items.xml]: why not attributes?

Comment (Zbizu) [inline data/items/items.xml]: unwrapping is still onUse while "wrap" is a new thing
why should I remove constructionkits.lua then?

File: data/events/scripts/player.lua (hunk 3 above)
Comment (marksamman) [inline data/events/scripts/player.lua]: It is correct, but I would rather see this inverted so that return true is at the bottom.

File: data/items/items.xml (hunk 2 above)
Comment (wgrriffel) [inline data/items/items.xml]: @Zbizu Wrap/Unwrap are the same bytes, you can just do a table with ids = {[x] = y}, kind off already is written in construction kits
@Kamenuvol because items that can be wrapped is a type not an attribute, its more easier to deal with it in lua and makes more sense, since we got more similar itemtypes such as isTeleport and isTile, why not isWrappable.

File: data/events/scripts/player.lua (hunk 3 above)
Comment (wgrriffel) [inline data/events/scripts/player.lua]: my mistaken, I rather too, makes more sense 0 as output when no errors.

File: data/items/items.xml (hunk 2 above)
Comment (wgrriffel) [inline data/items/items.xml]: @Zbizu I mean, according to link provide in #2017, in tibia.wikia.com:

> Since update 10.92 you can no longer "use" a furniture kit to open it, and it will not cause a spark effect anymore if you try to. In order to open it, you now need to right-click the furniture package and choose "unwrap". An item can be rewrapped by right-clicking it and choosing "wrap".
so actions script should be removed and handled in Wrap event.

Comment (Zbizu) [inline data/items/items.xml]: @wgrriffel I see, you may be right
some construction kits have "unwrap" feature, others don't
I don't know if it depends on ids or not

Comment (wgrriffel) [inline data/items/items.xml]: @Zbizu it's client sided , if the option doesn't pop up when you right click, its because its the wrong id, with lucky otclient is up to date with 10.92, then you could take look there what items has this flag or attribute, I don't know otclient project very well.

Comment (Zbizu) [inline data/items/items.xml]: @wgrriffel regarding your previous comment ```because items that can be wrapped is a type not an attribute, its more easier to deal with it in lua and makes more sense, since we got more similar itemtypes such as isTeleport and isTile, why not isWrappable.``` I can't check if item is wrappable. It requires either items.xml attribute or otb flag (can also be done in lua lib folder if I get @marksamman's permission).

**Comment by comedinha** [comment]
The store itens need wrap with unique/action id...

**Comment by Zbizu** [comment]
let's get wrap/unwrap to work first, actionid can be handled through lua
I did a couple of changes. What else needs to be done?

**Comment by ranisalt** [comment]
I believe it's better to do like doors, and have a map of unwrapped -> wrapped on global.lua instead of having new attributes on items.xml

**Comment by Kamenuvol** [comment]
I believe it's better to put wrap/unwrap; doors and other simple transform on use directly in attributes.

**Comment by HeavenIsLost** [comment]
@ranisalt it can be added to a global table but i believe if its added to items.xml its better organized? Cipsoft had a better items.xml than ours because it had a lot of custom and needed data in the item config while we have several hacks to add certain actions/attributes to items.

**Comment by ranisalt** [comment]
I don't like attributes on items.xml because they require source modifications to work. If there was a generic way to add attributes to items.xml, then it would be a good option.

**Comment by Kamenuvol** [comment]
We also need to add a lot of attributes because of the new updates: imbuing slots, critical, etc., so wrapping is just one more.
I'd suggest to even change items.xml structure but would be very difficult
Example:
```xml
<item id="xxx" name="xxx" description="">
....
<defense melee="xxx" fire="xxx" ice="xxx"/>
<actions rotateTo="xxx" destroyTo="xxx" wrapTo="xxx"/>
</item>
```
but also make a Lua script to read the xml, I don't know if it is even possible

ranisalt:
I wouldn't call it difficult, it would just be a pain to maintain backwards compatibility if desired.

HeavenIsLost:
We also need a special item attribute to save(to blob/database) the wrapped item inside. 10.92 store decoration items are wrapped into one item id instead of several item ids each representing the wrapped item. If we have a special item attribute we can just set it when wrapping and get+reset it when de-wrapping and transform it into the real decoration item.

Kamenuvol:
There is not difference about size of the package? There was 4 sizes in old furniture packages, and different names and capacity. How about them now?

Zbizu:
@HeavenIsLost @Kamenuvol actionid

marksamman:
@HeavenIsLost Do I understand it correctly that all wrappable items are wrapped into the same Item ID?

TheSumm:
Most of the furniture from the Tibia store wraps to the same item ID ( http://tibia.wikia.com/wiki/Tibia_Store#House_Equipment ). The other furniture still uses the old furniture kits.

GoularPink:
Hello, I tested this system here and dont work, sorry if you already knew that.

Mkalo:
@Zbizu Do you plan to rebase and fix the issues?

darkjav:
@Zbizu 
Confirmed, not work.

GoularPink:
We need wrap to furniture kit and decoration kit, two types of wrap

EPuncker:
just referencing #2407 here in case anyone wants a working one and if its even gets open and merged to close this one

PR #2036, opened 2016-12-29
Improve ConfigManager class

ranisalt:
This moves the static functions to the source file (.cpp) and scopes them inside a namespace, as they are only needed for that single translation unit. This moves a `#include` from a header that is included in a lot of files to a cpp file, potentially improving compilation times.

For optimized buils, this also allows for more aggressive optimizations, even inlining entirely all those functions if the compiler decides to.

Also it moves a string from the string list to the translation unit, this is personal.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -132,11 +132,13 @@ bool ConfigManager::reload()
 	return result;
 }
 
+static std::string dummy;
+
 const std::string& ConfigManager::getString(string_config_t what) const
 {
 	if (what >= LAST_STRING_CONFIG) {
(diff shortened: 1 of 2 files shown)


Kamenuvol:
Nice, I've learned one more thing in C++ (using anonymous namespaces to make functions in file scope)
But, why not keep them static in cpp?

marksamman:
Thanks!

ranisalt:
@Kamenuvol static keyword has different meanings depending on context. Anonymous namespaces are a superior (IMHO) alternative that, in this case, has the same effect.

PR #1906, opened 2016-09-04
Add metatable function Group:hasFlag(flag)

Kamenuvol:
Group has now hasFlag(flag) function, moved from old Lua API. Old version in compat.

PR #2037, opened 2016-12-29
Moved getPlayerFlagValue(cid, flag) to metatable

raymondtfr:


marksamman:
#1906 already does this, but unfortunately it also does other stuff which has blocked it from being merged. However, it does it by adding hasFlag to the Group metatable, which I think is better because the flags are on the group. The best solution would be using getFlags with a bit-and, but 64-bit bitwise operations aren't supported in LuaJIT yet.

raymondtfr:
Closing this due to #1906 and Mark's comment.

PR #2038, opened 2016-12-29
Moved getPlayerInstantSpellCount & getPlayerInstantSpellInfo to metatable

raymondtfr:


marksamman:
I don't think the old API was good enough that it should just be replicated as metatable functions. IMO it needs a discussion, because we don't want to be deprecating the new functions later. I think having a function that returns the list of spells is better, then you can just check the length of that list to get the count. Don't move old functions to compat just because you can reimplement them in the same way in the metatables, the reason many of the remaining functions are still in C++ is because we haven't thought of a better API for them yet.

WibbenZ:
Isent my pr better in this case since I added a spell class to Lua?
That way it woulf be easier if we wanna add new stuff related to spells in Lua.
Not to mention we get more functions to the Lua API that might be useful.

WibbenZ:
Ye I understand that haha
Just tell me if you want smth changed with it and ill do it tomorow

raymondtfr:
Closing this due to #2076 .

PR #2039, opened 2016-12-29
Moved getPlayerInstantSpellInfo to metatable

raymondtfr:


marksamman:
My comment in #2038 applies here as well. You could put the spell related functions in the same pull request, or perhaps open an issue for discussion first.

raymondtfr:
Ok, I'll move this to #2038 .

PR #2040, opened 2016-12-29
Moved getCreatureCondition to metatable

raymondtfr:
Renamed function name ``getCreatureCondition`` to ``hasCondition`` and parameter **condition** to **type** as it stands for the condition type.

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -138,6 +138,7 @@ function getCreaturePosition(cid) local c = Creature(cid) return c ~= nil and c:
 function getCreatureOutfit(cid) local c = Creature(cid) return c ~= nil and c:getOutfit() or false end
 function getCreatureSpeed(cid) local c = Creature(cid) return c ~= nil and c:getSpeed() or false end
 function getCreatureBaseSpeed(cid) local c = Creature(cid) return c ~= nil and c:getBaseSpeed() or false end
+function getCreatureCondition(cid, ...) local c = Creature(cid) return c ~= nil and c:hasCondition(...) or false end
 
 function getCreatureTarget(cid)
 	local c = Creature(cid)
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -919,9 +919,6 @@ void LuaScriptInterface::registerFunctions()
 	//doSetCreatureLight(cid, lightLevel, lightColor, time)
 	lua_register(luaState, "doSetCreatureLight", LuaScriptInterface::luaDoSetCreatureLight);
 
-	//getCreatureCondition(cid, condition[, subId])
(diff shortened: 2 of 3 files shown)


ranisalt:
You should group this function code with the others (addCondition, removeCondition)

data/lib/compat/compat.lua
```diff
@@ -181,6 +181,7 @@ function doSetCreatureDropLoot(cid, doDrop) local c = Creature(cid) return c ~=
 function doChangeSpeed(cid, delta) local c = Creature(cid) return c ~= nil and c:changeSpeed(delta) or false end
 function doAddCondition(cid, conditionId) local c = Creature(cid) return c ~= nil and c:addCondition(conditionId) or false end
 function doRemoveCondition(cid, conditionType, subId) local c = Creature(cid) return c ~= nil and (c:removeCondition(conditionType, CONDITIONID_COMBAT, subId) or c:removeCondition(conditionType, CONDITIONID_DEFAULT, subId) or true) end
+function getCreatureCondition(cid, type, subId) local c = Creature(cid) return c ~= nil and c:hasCondition(type, subId) or false end
```
ranisalt: For simplicity, I suggest you use:

```
function getCreatureCondition(cid, ...) local c = Creature(cid) return c ~= nil and c:hasCondition(...) or false end
```

src/luascript.cpp
```diff
@@ -7253,6 +7236,21 @@ int LuaScriptInterface::luaCreatureRemoveCondition(lua_State* L)
...
+int LuaScriptInterface::luaCreatureHasCondition(lua_State* L)
+{
+	//creature:hasCondition(conditionType, subId)
+	Creature* creature = getUserdata<Creature>(L, 1);
+	if (!creature) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	ConditionType_t condition = getNumber<ConditionType_t>(L, 2);
```
raymondtfr: ``conditionType``

src/luascript.cpp
```diff
@@ -7253,6 +7236,21 @@ int LuaScriptInterface::luaCreatureRemoveCondition(lua_State* L)
...
+	uint32_t subId = getNumber<uint32_t>(L, 3, 0);
+	pushBoolean(L, creature->hasCondition(condition, subId));
```
raymondtfr: ``conditionType``

src/luascript.cpp
ranisalt: It's because other functions call this parameter `condition`. You can create another PR renaming once this is merged.

ranisalt:
I merged with master because the rebase was very strange and it really doesn't matter now. Is it ok @raymondftr?

PR #2042, opened 2016-12-30
Assorted code improvements

ranisalt:
As an extension of #2036, I applied similar changes throughout the code:

- Moved private static functions to the implementation (cpp) where possible, reducing header sizes.
- Replace `typedef` with `using`, as to have a standard type aliasing method.
- Standardized type aliases and renamed variables where needed (some may went unnoticed, please point)
- Use of `std::function` as function wrapper instead of raw function pointers.

A rune spell function was renamed (`Illusion` to `RuneIllusion`) because of a name clash when converting private static to free function.

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -440,6 +440,32 @@ bool Action::configureEvent(const pugi::xml_node& node)
 	return true;
 }
 
+namespace {
+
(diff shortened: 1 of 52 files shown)


marksamman:
Thanks!

PR #2043, opened 2016-12-31
Fix targetPlayer bug in find_person.lua

WibbenZ:
Forgot to add the variable

### DIFF
--- a/data/spells/scripts/support/find_person.lua
+++ b/data/spells/scripts/support/find_person.lua
@@ -46,6 +46,7 @@ function onCastSpell(creature, variant)
 	end
 
 	local playerPosition = creature:getPosition()
+	locla targetPosition = targetPlayer:getPosition()
 	local offset = {
 		x = playerPosition.x - targetPosition.x,
 		y = playerPosition.y - targetPosition.y,
--- a/data/spells/scripts/support/find_person.lua
+++ b/data/spells/scripts/support/find_person.lua
@@ -46,7 +46,7 @@ function onCastSpell(creature, variant)
 	end
 
 	local playerPosition = creature:getPosition()
-	locla targetPosition = targetPlayer:getPosition()
+	local targetPosition = targetPlayer:getPosition()
 	local offset = {
 		x = playerPosition.x - targetPosition.x,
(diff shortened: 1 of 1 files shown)


wgrriffel:
You should fix one more thing, i'd point out in your last pr, however i didn't make before it gets merged. In this case variant.text is nill, to variant.text be filled [this](https://github.com/otland/forgottenserver/blob/master/src/spells.cpp#L1157) statement must be false, which is not because it needs a target, instead you should get variant.number which store player id, also i think player shouldn't be able to exiva himself.

data/spells/scripts/support/find_person.lua
```diff
@@ -46,6 +46,7 @@ function onCastSpell(creature, variant)
 	end
 
 	local playerPosition = creature:getPosition()
+	local targetPosition = targetPlayer:getPosition()
```
wgrriffel: You should fix one more thing, i'd point out in your last pr, however i didn't make before it gets merged. In this case variant.text is nill, to variant.text be filled this statement must be false, which is not because it needs a target, instead you should get variant.number which store player id, also i think player shouldn't be able to exiva himself.

WibbenZ: This should not matter, if the text variable of variant is nil, it will return false.

https://github.com/WibbenZ/forgottenserver/blob/062fe5f2b0403e4786ffd01a60dd29e94906ef81/data/spells/scripts/support/find_person.lua#L37-L40

wgrriffel: @WibbenZ you don't get me, it will always be nill, doesnt matter if it has or not parameters

WibbenZ: But I can still not get it to throw an error, so how is it bugged?
We still have the text variable.

I tested casting it without any name, gives me the cancel message, same a misspelled name.
As I said, the script works, so please tell me how to make it bug.

And I highly doubt it would even throw an error, since if the variant dosn't have a string variable it would return false since it can't get the userdata for targetPlayer.

marksamman:
Thanks!

Issue #1064, opened 2014-11-17
Questlog scripting.


reyaleman:
I think that it's time to discuss (in order to start to work on it) about the way we want that lua should be implemented in questlogs, I don't know if someone it's already working on it or so.

I'll make some suggestions and I'll appreciate some ideas or improvements about it.
###### data/quests/quests.xml

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<quests>
    <quest name="Quest #1" script="example/quest.lua">
        <mission name="Mission #1" description="You've killed %d rats and %d cave rats." script="example/mission1.lua" />
        <!--<mission script="example/mission2.lua" />-->
    </quest>
</quests>
```

We can set a default name and descriptions for missions and quests (optional) and let lua do the job.
###### scripts/example/quest.lua (Quest#1)

``` lua
function Quest:setName(player)
    return self:isCompleted() and (self:getName() .. " (completed)") or self:getName()
end

function Quest:isCompleted(player)
    return player:getStorageValue(9999) >= 10
end

function Quest:isStarted(player)
    return player:getStorageValue(9999) > 0
end
```
###### scripts/example/mission1.lua (Mission#1)

``` lua
function Mission:setName(player)
    return self:isCompleted() and (self:getName() .. " (completed)") or self:getName()
end

function Mission:setDescription(player)
    local description = self:getDescription()
    if not description then
        description = "You've killed %d rats and %d cave rats."
    end 
    return string.format(description, player:getStorageValue(9001), player:getStorageValue(9002))
end

function Mission:isCompleted(player)
    return player:getStorageValue(9000) >= 10
end

function Mission:isStarted(player)
    return player:getStorageValue(9000) > 0
end
```

We'll have 2 new lua metatables only accessible by this lua interface:
- Quest
- Mission

``` lua
Quest:getId()
Quest:getName() -- returns the name (if provided in xml) or nil
Quest:isStarted(player) -- check if the player have started this quest (trigger the Quest:isStarted event)
Quest:isCompleted(player) -- check if the player have finished this quest (trigger the Quest:isCompleted event)
Quest:getMissions([player]) -- returns  the quest missions (if player ~= nil it will check for the player missions that has been started...)
Quest:getMissionCount([player]) --  returns the quest missions total (if player ~= nil it will check for the player missions that has been started...)

Mission:getId()
Mission:getName()  -- returns the name (if provided in xml) or nil
Mission:getDescription() -- returns the description (if provided in xml) or nil
[...]

PR #2048, opened 2017-01-02
Questlog scripting.

Mkalo:
Still not completed, there are bugs that I want opinion from you on how
to deal with them:

Calling hasStarted/hasCompleted in isStarted/isCompleted events will
result in a stack overflow, for obvious reasons.

Calling Mission:getDescription without forceXml being true inside
Mission:setDescription will result in a stack overflow, for obvious
reasons (You're trying to get what your function was supposed to
return.)

Worse in my opinion:
Calling Mission:getName(player) inside Mission:isCompleted will result
in stack overflow, (getName uses isCompleted to concat the "
(completed)" string)

And those are the known bugs until now, there are probably more that I
still haven't found.

Edit: One solution would be removing all those methods, leaving only the methods to fetch information that doesnt change when calling events.

We wouldn't have those problems if we used #1973.

This was based in #1064.


gugahoa:
For backward compatibility quests.xml should still be inside XML, or is this too trivial to worry about?

gugahoa:
For stack overflow problem you can use mutex per player

brunominervino:
@Mkalo you will finish this? This is a awesome implementation.

flaviiojr:
@Mkalo Will this implementation be completed?
Really is a great implementation, would change the whole way of questlog!

ranisalt:
Signed. Let's move towards Lua scripting instead of XML scripting wherever we can.

ColorsCL:
Is someone -still- working on this system? 🤔

omarcopires:
Are there any reasons to keep this open? (has been out of business for over 5 years).

ranisalt:
@omarcopires this was promising, but will very likely be superseded by #4240 now

EPuncker:
closed on behalf of #4240

PR #2049, opened 2017-01-03
onStep callback return value matter

gugahoa:
Creating a new PR with a clean work tree. Previous PR #1810

### DIFF
--- a/data/movements/scripts/tiles.lua
+++ b/data/movements/scripts/tiles.lua
@@ -23,9 +23,8 @@ function onStepIn(creature, item, position, fromPosition)
 	end
 
 	if Tile(position):hasFlag(TILESTATE_PROTECTIONZONE) then
-		local lookPosition = position
-		lookPosition:getNextPosition(player:getDirection())
-		local depotItem = Tile(lookPosition):getItemByType(ITEM_TYPE_DEPOT)
+		position:getNextPosition(player:getDirection())
+		local depotItem = Tile(position):getItemByType(ITEM_TYPE_DEPOT)
 		if depotItem ~= nil then
 			local depotItems = player:getDepotChest(getDepotId(depotItem:getUniqueId()), true):getItemHoldingCount()
 			player:sendTextMessage(MESSAGE_STATUS_DEFAULT, "Your depot contains " .. depotItems .. " item" .. (depotItems > 1 and "s." or "."))
--- a/data/movements/scripts/tiles.lua
+++ b/data/movements/scripts/tiles.lua
@@ -15,15 +15,15 @@ function onStepIn(creature, item, position, fromPosition)
 
 	if item.actionid >= 1000 then
 		if player:getLevel() < item.actionid - 1000 then
-			player:teleportTo(fromPosition, false)
 			position:sendMagicEffect(CONST_ME_MAGIC_BLUE)
 			player:sendTextMessage(MESSAGE_INFO_DESCR, "The tile seems to be protected against unwanted intruders.")
+			return false
 		end
 		return true
 	end
 
 	if Tile(position):hasFlag(TILESTATE_PROTECTIONZONE) then
-		local lookPosition = player:getPosition()
+		local lookPosition = position
 		lookPosition:getNextPosition(player:getDirection())
 		local depotItem = Tile(lookPosition):getItemByType(ITEM_TYPE_DEPOT)
 		if depotItem ~= nil then
@@ -34,10 +34,9 @@ function onStepIn(creature, item, position, fromPosition)
 	end
 
 	if item.actionid ~= 0 and player:getStorageValue(item.actionid) <= 0 then
-		player:teleportTo(fromPosition, false)
 		position:sendMagicEffect(CONST_ME_MAGIC_BLUE)
 		player:sendTextMessage(MESSAGE_INFO_DESCR, "The tile seems to be protected against unwanted intruders.")
-		return true
+		return false
 	end
 	return true
 end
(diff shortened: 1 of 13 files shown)


gugahoa:
cc @ranisalt @marksamman 
Everything is tested and working

One thing I think is worth noting is that Teleport::addThing is now using Game::internalTeleportThing instead of directly moving the creature/item, so if onStepOut/onStepIn returns false, the teleport will fail.

data/movements/scripts/tiles.lua
```diff
@@ -15,15 +15,15 @@ function onStepIn(creature, item, position, fromPosition)
...
 			position:sendMagicEffect(CONST_ME_MAGIC_BLUE)
 			player:sendTextMessage(MESSAGE_INFO_DESCR, "The tile seems to be protected against unwanted intruders.")
+			return false
 		end
 		return true
 	end
 
 	if Tile(position):hasFlag(TILESTATE_PROTECTIONZONE) then
-		local lookPosition = player:getPosition()
+		local lookPosition = position
```
ranisalt: Do you really need to alias `position` to `lookPosition` or could you use `position` straight away?

data/movements/scripts/walkback.lua
```diff
@@ -1,6 +1,6 @@
 function onStepIn(creature, item, position, fromPosition)
 	if item.uid > 0 and item.uid <= 65535 then
-		creature:teleportTo(fromPosition, false)
+		return false
 	end
 	return true
```
ranisalt: Simplify to `return item.uid == 0` (uid can't be > 65535)

marksamman: uid is over 65535 for items without a unique id, this is how we used to reference items before item userdata.

ranisalt: But it is pushed as an `uint16_t`, isn't it?

Anyway it can be `return item.uid == 0 or item.uid > 65535`

marksamman: Nope, see https://github.com/otland/forgottenserver/blob/master/data/lib/compat/compat.lua#L74 and https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L5908, actually I think item.uid == 0 is the useless one.

ranisalt: Just read through the code and from what I understood, if uid is >65535 the unique id was generated, not set through the map/actions.

`addThing` can return 0 if the item is removed, though I'm not sure if this is ever possible, it's better to be on the safe side.

gugahoa:
I made so that if the teleport action is made from Lua, the onStepOut return value should not matter. But this is only for onStepOut, onStepIn (the tile the player is being sent to) is still triggered and can cancel the action.

Use case: Player stuck on a onStepOut tile that keeps returning false, /c Player should bring the player to the GM anyway. I made so the onStepOut event is still triggered, though.

Feedback about this would be appreciated, as this will have noticeable impact on the previous behavior of teleport to/from tiles with onStep event

HeavenIsLost:
I'm in favor of a preStepIn and postStepIn(like the current opentibia one). It will keep compatibility and + it will give better organization. preStepIn to prevent step in and postStepIn(when the walk did actually happened for real).

ranisalt:
@HeavenIsLost how do they work? What happens if pre or post fail?

kornholi:
If pre fails, the action is cancelled. Post is just a notification and doesn't have a return value. If a player is being teleported away by someone else, it doesn't make sense to call preStepOut since you can't cancel it.

ranisalt:
@marksamman these changes result in different behavior, so it is not backwards-compatible per se. It probably will not break anything, since previously scripts manually teleported the player back.

I think it could lead to an infinite loop (both `fromPos` and `toPos` refusing to accept the player), but not sure if any code path reaches it.

gugahoa:
@ranisalt it doesn't enter infinite loop, already tested it

marksamman:
onStepIn/onStepOut should default to return true if there's no return value, that's my only backwards compatibility concern with this PR.

gugahoa:
@marksamman how can I differentiate between return false and no return at all?

ranisalt:
@gugahoa you'll probably need to change `LuaScriptInterface::callFunction` to receive a default return value. Internally it uses `LuaScriptInterface::getBoolean` which already accepts default value as parameter.

gugahoa:
I made a little change to `LuaScriptInterface::getBoolean`
It now checks if the argument is boolean (since it can be nil), if not, return the defaultValue
This will make getBoolean method more strict, as any non-boolean value will evaluate to defaultValue instead of true

> int lua_toboolean (lua_State *L, int index);
> 
> Converts the Lua value at the given acceptable index to a C boolean value (0 or 1). Like all tests in Lua, lua_toboolean returns 1 for any Lua value different from false and nil; otherwise it returns 0. It also returns 0 when called with a non-valid index. (If you want to accept only actual boolean values, use lua_isboolean to test the value's type.)

src/luascript.cpp
```diff
@@ -504,14 +504,14 @@ int LuaScriptInterface::luaErrorHandler(lua_State* L)
 	return 1;
 }
 
-bool LuaScriptInterface::callFunction(int params)
+bool LuaScriptInterface::callFunction(int params, bool defaultReturn /* = false */)
```
ranisalt: Sorry if I'm being too pedantic here, but wouldn't it be nice to have standard parameter name? Like using `defaultValue` everywhere.

gugahoa: It's because in this one, it'll be the default return value of the function being called, in `LuaScriptInterface::getBoolean` it's the value of the variable being read.
I think it's more explicit this way what that default represents

ranisalt:
Thanks!

PR #2050, opened 2017-01-03
create_item.lua bugfix

Colandus:
Fixed bugs if wrong name, a blank item would be created. Also if param left blank. (This is because ItemType returns a blank item with id 8893 if given a nil value). Creation of such items can lead to things and creatures disappearing on that tile.

Another one of my concerns is creating items with ID's such as e.g. 1, which returns the name "water" but crashes the client when created.

I kind of consider this a tempfix though. I suppose this should be solved in the sources.

```
// ItemType(id or name)
uint32_t id;
if (isNumber(L, 2)) {
	id = getNumber<uint32_t>(L, 2);
} else {
	id = Item::items.getItemIdByName(getString(L, 2));
}
```

Why do you expect it to be a string if its not a number? Should use isString, else pushnil?


data/talkactions/scripts/create_item.lua
```diff
@@ -8,10 +8,17 @@ function onSay(player, words, param)
...
+	end
 
 	local itemType = ItemType(split[1])
 	if itemType:getId() == 0 then
-		itemType = ItemType(tonumber(split[1]))
+		if tonumber(split[1]) then
+			itemType = ItemType(tonumber(split[1]))
+		end
+
 		if itemType:getId() == 0 then
```
Colandus: Yes, it is quite unnecessary to perform the check although we know for sure it will fail if above condition passes. Although, I felt it was better than creating variables to check for success (or repaeting the cancel message).

ranisalt: It is useless, `tonumber` does not fail, so you can't check for success. If the argument is not a number, it returns nil, and it's coerced to zero anyway.

data/talkactions/scripts/create_item.lua
```diff
@@ -8,10 +8,17 @@ function onSay(player, words, param)
 	end
 
 	local split = param:split(",")
+	if not split[1] then
```
ranisalt: You should use `if #split < 1` instead, it is semantically better to check if the array has the correct amount of values rather than check an index.

data/talkactions/scripts/create_item.lua
```diff
@@ -28,11 +35,7 @@ function onSay(player, words, param)
 			count = math.max(0, count)
 		end
 	else
-		if not itemType:isFluidContainer() then
-			count = 1
-		else
-			count = 0
-		end
+		count = itemType:isFluidContainer() and 0 or 1
```
ranisalt: A suggestion: create a local function `getDefaultCount` and return the count for each situation. Ternary operations are ugly in Lua.

data/talkactions/scripts/create_item.lua
Colandus: Yes it does. tonumber("string") returns nil

Colandus: I think you misunderstood me or I misunderstood you. On a second read I can't relate your comment to the code.

data/talkactions/scripts/create_item.lua
Colandus: Yeah I considered that but I thought it felt somewhat hackish doing that. But I guess you're right.

data/talkactions/scripts/create_item.lua
Colandus: I think they are cute. If more people agree with you, I will accept. I feel if they did, we would already have such a function (do we?).

ranisalt: I meant a function would be useful for the entire `count` definition, not just this line :P e.g.

```
local function getCount(count)
	if count ~= nil then
		if itemType:isStackable() then
			return math.min(10000, math.max(1, count))
		end
		if not itemType:isFluidContainer() then
			return math.min(100, math.max(1, count))
		end
		return math.max(0, count)
	end
	return itemType:isFluidContainer() and 0 or 1
end

(...)

local result = player:addItem(itemType:getId(), getCount(tonumber(split[2])))
```

data/talkactions/scripts/create_item.lua
ranisalt: My comment is related to the lines you added, I accidentally commented one line below.

Colandus: no its the same

ranisalt: And when the argument is "0"? Will it coerce 0 to false?

Colandus: no

DSpeichert:
This was fixed in 20737e5c1b08e2bfc228e0b3c0cb1694d549d46e

PR #2053, opened 2017-01-04
Implement equip item via hotkey feature

gugahoa:
Started with a clean work tree again. Previous #1794


src/game.cpp
```diff
@@ -1681,6 +1681,67 @@ ReturnValue Game::internalTeleport(Thing* thing, const Position& newPos, bool pu
...
+
+	Item* item = nullptr;
+	if (!(item = player->getInventoryItem(CONST_SLOT_BACKPACK))) {
+		return;
+	}
+
+	Container* backpack = nullptr;
+	if (!(backpack = item->getContainer())) {
+		return;
+	}
```
ranisalt: This start is all out of consistency, it does not have a pattern.

Throughout the code, usually you will see
```
Player* player = getPlayerById(playerId);
if (player == nullptr) {
    return;
}
```
So I ask you do it accordingly.

src/game.cpp
```diff
@@ -1681,6 +1681,67 @@ ReturnValue Game::internalTeleport(Thing* thing, const Position& newPos, bool pu
...
+
+	static const std::function<Item*(Container*, uint16_t)> searchForItem = [](Container* container, uint16_t itemid) -> Item* {
+		for (ContainerIterator it = container->iterator(); it.hasNext(); it.advance()) {
+			if ((*it)->getID() == itemid) {
+				return *it;
+			}
+		}
+		return nullptr;
+	};
```
ranisalt: As this function does not depend on context, you can make it a free function instead. It's better for readability and makes code simpler.

You can use `std::find_if` instead, if this `->iterator()` returns valid C++11 range.

src/items.cpp
```diff
@@ -326,9 +326,33 @@ bool Items::loadFromXml()
...
 }
 
+void Items::buildInventoryList()
+{
+	inventory.reserve(30000);
+	for(const auto &type: items) {
+		if(type.weaponType != WEAPON_NONE || type.ammoType != AMMO_NONE ||
+			type.attack != 0 || type.defense != 0 ||
+			type.extraDefense != 0 || type.armor != 0 ||
+			(type.slotPosition & SLOTP_NECKLACE) == SLOTP_NECKLACE ||
```
ranisalt: You don't need to compare it to `SLOTP_NECKLACE`, it will coerce to `true` if not equal to zero, which, bitwisely, means it's equal to `SLOTP_NECKLACE` anyways. Same for below.

src/items.cpp
```diff
@@ -326,9 +326,33 @@ bool Items::loadFromXml()
...
 	}
+
+	buildInventoryList();
 	return true;
 }
 
+void Items::buildInventoryList()
+{
+	inventory.reserve(30000);
+	for(const auto &type: items) {
```
ranisalt: `for (const auto& type : items)`

src/items.cpp
```diff
@@ -326,9 +326,33 @@ bool Items::loadFromXml()
...
 		}
 	}
+
+	buildInventoryList();
 	return true;
 }
 
+void Items::buildInventoryList()
+{
+	inventory.reserve(30000);
```
ranisalt: You don't need that much memory, reserving `items.size()` is enough.

src/items.cpp
```diff
@@ -326,9 +326,33 @@ bool Items::loadFromXml()
...
+			(type.slotPosition & SLOTP_HEAD) == SLOTP_HEAD ||
+			(type.slotPosition & SLOTP_ARMOR) == SLOTP_ARMOR ||
+			(type.slotPosition & SLOTP_LEGS) == SLOTP_LEGS)
+		{
+			inventory.push_back(type.clientId);
+		}
+	}
+	inventory.shrink_to_fit();
+	std::sort(inventory.begin(), inventory.end());
+}
```
ranisalt: In this function, take a look at [std::copy_if](http://en.cppreference.com/w/cpp/algorithm/copy_if), it does what you need.

src/items.h
```diff
@@ -294,6 +294,9 @@ class Items
 		bool loadFromXml();
 		void parseItemNode(const pugi::xml_node& itemNode, uint16_t id);
 
+		void buildInventoryList();
+		const std::vector<uint16_t>& getInventory() const { return inventory; }
```
ranisalt: It can be wise to alias `InventoryVector` to `std::vector<uint16_t>`

src/player.h
```diff
@@ -1285,8 +1290,8 @@ class Player final : public Creature, public Cylinder
 		bool ghostMode = false;
 		bool pzLocked = false;
 		bool isConnecting = false;
-		bool addAttackSkillPoint = false;
 		bool inventoryAbilities[CONST_SLOT_LAST + 1] = {};
+		bool addAttackSkillPoint = false;
```
ranisalt: ?

src/protocolgame.cpp
```diff
@@ -902,6 +903,14 @@ void ProtocolGame::parseFollow(NetworkMessage& msg)
 	addGameTask(&Game::playerFollowCreature, player->getID(), creatureId);
 }
 
+void ProtocolGame::parseEquipObject(NetworkMessage& msg)
+{
+	uint16_t spriteId = msg.get<uint16_t>();
+	// msg.get<uint8_t>();
```
ranisalt: Remove this line if really unneeded.

src/protocolgame.cpp
```diff
@@ -2530,6 +2539,28 @@ void ProtocolGame::sendInventoryItem(slots_t slot, const Item* item)
...
+{
+	NetworkMessage msg;
+	msg.addByte(0xF5);
+
+	const std::vector<uint16_t>& inventory = Item::items.getInventory();
+	msg.add<uint16_t>(inventory.size() + 11);
+	for (uint16_t i = 1; i <= 11; i++) {
+		msg.add<uint16_t>(i);
+		msg.addByte(0); //always 0
+		msg.add<uint16_t>(1); // always 1
```
ranisalt: If you know where those values come from, please add a comment instead of "always x"

src/protocolgame.cpp
```diff
@@ -2530,6 +2539,28 @@ void ProtocolGame::sendInventoryItem(slots_t slot, const Item* item)
...
+	}
+
+	for (int clientId : inventory) {
```
ranisalt: `clientId` is not `int`, it's `uint16_t`, though it's probably better to use `auto`.

src/items.cpp
gugahoa: It seems I can't use copy_if
```	std::vector<ItemType> type_inventory;
	std::copy_if(items.begin(), items.end(), std::back_inserter(type_inventory), predicateInsertItem);```
A snippet like this will give me some errors because of this `ItemType(const ItemType& other) = delete;`
:(

File: src/player.h (hunk 8 above)
Comment (gugahoa) [inline src/player.h]: Ops

File: src/protocolgame.cpp (hunk 9 above)
Comment (gugahoa) [inline src/protocolgame.cpp]: Mark asked in the previous PR to just comment it and let it sit there so we know there's something being sent

File: src/protocolgame.cpp (hunk 10 above)
Comment (gugahoa) [inline src/protocolgame.cpp]: Sadly I don't know :(

File: src/items.cpp (hunk 6 above)
Comment (ranisalt) [inline src/items.cpp]: My bad, didn't see you add `type` instead of `type.clientId`

File: src/game.cpp (hunk 2 above)
Comment (gugahoa) [inline src/game.cpp]: Not possible, sadly ContainerIterator is not a valid C++11 range

**Comment by gugahoa** [comment]
Merge status? @ranisalt \cc @djarek

File: src/items.h (hunk 7 above)
Comment (ranisalt) [inline src/items.h]: I approved but you still might change this.

**Comment by ranisalt** [comment]
Just one more thing to look at.

**Comment by diath** [comment]
I'm not sure whether such feature would be desirable or not but with classicEquipmentSlots set to true in the config this code will not consider items that you have in backpacks that you're holding in your hands/arrow slot.

**Comment by gugahoa** [comment]
@diath Yeah, I thought about that too and for simplicity didn't try to solve those cases
@ranisalt Done

File: src/protocolgame.cpp (hunk 10 above)
Comment (Mkalo) [inline src/protocolgame.cpp]: The uint16_t is the item count. You are only sending all equippable items once this doesn't really matter.

**Comment by thexamx** [comment]
shouldn't this system have delay? You can instantly equip for ex. SSA

File: src/game.cpp [hunk 12]
```diff
@@ -1688,7 +1688,75 @@ ReturnValue Game::internalTeleport(Thing* thing, const Position& newPos, bool pu
 	return RETURNVALUE_NOTPOSSIBLE;
 }
 
+Item* searchForItem(Container* container, uint16_t itemid)
```
raymondtfr: You should rename the parameter ``itemid`` to ``itemId``.
You should also rename the function name to (since ``searchForItem`` alone implies a ``void`` function, etc...) ``getContainerItem`` (since it actually ``return``s a pointer to the ``Item`` inside the container that matches the item id you're comparing to).

src/items.cpp
```diff
@@ -326,9 +326,33 @@ bool Items::loadFromXml()
...
 	}
+
+	buildInventoryList();
 	return true;
 }
 
+void Items::buildInventoryList()
+{
+	inventory.reserve(items.size());
+	for(const auto& type: items) {
```
raymondtfr: You forgot to add space between ``for`` and ``(``. You should also rename ``type`` to ``itemType`` (weird for a ``vector`` used to hold ``ItemType``s to be named loosely as ``items``)

src/items.cpp
```diff
@@ -326,9 +326,33 @@ bool Items::loadFromXml()
...
+		if(type.weaponType != WEAPON_NONE || type.ammoType != AMMO_NONE ||
```
raymondtfr: You also forgot to add a space here, between ``if`` and ``(``.

src/items.h
```diff
@@ -294,6 +295,9 @@ class Items
 		bool loadFromXml();
 		void parseItemNode(const pugi::xml_node& itemNode, uint16_t id);
 
+		void buildInventoryList();
+		const InventoryVector& getInventory() const { return inventory; }
```
raymondtfr: You should break down this member-function signature into new lines.

src/game.cpp
```diff
@@ -1688,7 +1688,75 @@ ReturnValue Game::internalTeleport(Thing* thing, const Position& newPos, bool pu
...
+	if (it.weaponType != WeaponType_t::WEAPON_SHIELD) {
+		int32_t slotPosition = it.slotPosition;
+
+		if (slotPosition & SLOTP_HEAD) {
+			slot = CONST_SLOT_HEAD;
+		} else if (slotPosition & SLOTP_NECKLACE) {
+			slot = CONST_SLOT_NECKLACE;
+		} else if (slotPosition & SLOTP_ARMOR) {
+			slot = CONST_SLOT_ARMOR;
+		} else if(slotPosition & SLOTP_LEGS) {
```
raymondtfr: Missing space between ``if`` and ``(``.

PR #2054, opened 2017-01-04
Added optional showEffect to player:setGhostMode

Colandus:
Added optional showEffect to player:setGhostMode with default value true (as it was) for "isLogin" value of Player::sendCreatureAppear

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -9187,7 +9187,7 @@ int LuaScriptInterface::luaPlayerGetHouse(lua_State* L)
 
 int LuaScriptInterface::luaPlayerSetGhostMode(lua_State* L)
 {
-	// player:setGhostMode(enabled)
+	// player:setGhostMode(enabled[, showEffect=true])
 	Player* player = getUserdata<Player>(L, 1);
 	if (!player) {
 		lua_pushnil(L);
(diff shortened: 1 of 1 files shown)


raymondtfr:
You could even make this better by getting which ``MagicEffectClasses`` the user would like to send, instead of a boolean.

Colandus:
@raymondtfr Frankly, I would have rather renamed it to "isLogin" as the source code has it and also go deeper and not update online status if isLogin == false. Just thought it might not be appreciated by @marksamman or whoever, so I stuck to this simple change.

When it comes to effects though, it can be easily handled in Lua if desired. Thus, I decided to just enable/disable it.

ranisalt:
You are using `showEffect` as a hack when you pass it to the `isLogin` parameter. The effect of setting `isLogin` to true are much bigger than just showing a blue effect around the player.

If you want to optionally show the effect, do it in the action, not in the source code.

Colandus:
@ranisalt That's not true actually. It's only used to send effect. I obviously looked that up before assuming it's just showing an effect. But yes, it could in future change meaning and then break the code. Either way, as I said, I would much rather prefer allowing the entire isLogin as optional and not just the effect and with this include all other login effects that aren't considered by the variable such as the VIP notifications and offline status in general.

ranisalt:
@Colandus sorry, I looked after `Player::creatureAppear` instead of just `ProtocolGame::sendCreatureAppear`, that's why my answer doesn't make sense.

Anyway, I still think it would be better to have the effect on the action, not the source. `isLogin` is more an internal flag.

src/luascript.cpp
```diff
@@ -9199,6 +9199,8 @@ int LuaScriptInterface::luaPlayerSetGhostMode(lua_State* L)
 		pushBoolean(L, true);
 		return 1;
 	}
+
```
WibbenZ: Trailing
LGTM

PR #2055, opened 2017-01-05
update action scripts

raymondtfr:


data/actions/scripts/other/music.lua
```diff
@@ -28,42 +28,47 @@ local instruments = {
 	[3953] = {effect = CONST_ME_SOUND_RED}, -- war drum
 	[3957] = {effect = CONST_ME_SOUND_YELLOW, item = {id = 2681, count = 10}, chance = 80, remove = true}, -- cornucopia
 	[5786] = {effects = {failure = CONST_ME_SOUND_RED, success = CONST_ME_SOUND_YELLOW}, monster = "war wolf", chance = 60, remove = true}, -- wooden whistle
-	[6572] = {effect = CONST_ME_SOUND_GREEN, text = "TOOOOOOT", transformId = 13578}, -- party trumpet
-	[6573] = {effect = CONST_ME_SOUND_GREEN, text = "TOOOOOOT", transformId = 13578}, -- party trumpet
+	[6572] = {effect = CONST_ME_SOUND_GREEN, text = "TOOOOOOT", transform = {id = 13578}}, -- party trumpet
+	[6573] = {effect = CONST_ME_SOUND_GREEN, text = "TOOOOOOT", transform = {id = 13578}}, -- party trumpet
```
WibbenZ: Why do you keep making the item id in another table when all the parent table has is that value?
Just name it transformId or w/e but you don't have to make transform a table, unless you add support to actually transform into more than one item.

raymondtfr: So that it can be easily customizated and also adds more maintainability to the code.

ranisalt: Make the simplest code that works. If new features are needed, they are to be added later, not now.

data/actions/scripts/other/constructionkits.lua
```diff
@@ -1,4 +1,4 @@
-local constructionKits = {
+local kits = {
```
ranisalt: No need to rename the variable...

data/actions/scripts/other/constructionkits.lua
```diff
@@ -8,24 +8,23 @@ local constructionKits = {
...
+	if not Tile(fromPosition):getHouse() then
 		player:sendTextMessage(MESSAGE_STATUS_SMALL, "You may construct this only inside a house.")
-	else
-		item:transform(kit)
-		fromPosition:sendMagicEffect(CONST_ME_POFF)
+		return true
 	end
+
+	item:transform(kits[item:getId()])
+	fromPosition:sendMagicEffect(CONST_ME_POFF)
```
ranisalt: I didn't understand why you changed the `if/elseif/else` structure.

data/actions/scripts/other/decayto.lua
```diff
@@ -1,16 +1,11 @@
-local decayItems = {
+local items = {
```
ranisalt: No need to rename this one too.

data/actions/scripts/other/constructionkits.lua
raymondtfr: I find it more clear not to use them, but I can revert it if there is no need.

raymondtfr:
Later I'll send a new PR with these changes (much better than dealing with conflicting files 😄 . I'll be closing this for now.

PR #2057, opened 2017-01-05
Minor update to Support spells scripts

raymondtfr:
* Updated ``find_person.lua`` to use Lua ternary operation in one case and also replaced whitespaces with tabs.
* Updated ``food.lua``.
* Replaced ``Variant.getPosition(variant)`` with ``variant:getPosition()``

### DIFF
--- a/data/spells/scripts/support/animate_dead_rune.lua
+++ b/data/spells/scripts/support/animate_dead_rune.lua
@@ -1,5 +1,5 @@
 function onCastSpell(creature, variant, isHotkey)
-	local position = Variant.getPosition(variant)
+	local position = variant:getPosition()
 	local tile = Tile(position)
 	if tile and creature:getSkull() ~= SKULL_BLACK then
 		local corpse = tile:getTopDownItem()
--- a/data/spells/scripts/support/destroy_field_rune.lua
+++ b/data/spells/scripts/support/destroy_field_rune.lua
@@ -1,8 +1,7 @@
 function onCastSpell(creature, variant, isHotkey)
-	local position = Variant.getPosition(variant)
+	local position = variant:getPosition()
 	local tile = Tile(position)
 	local field = tile and tile:getItemByType(ITEM_TYPE_MAGICFIELD)
-
 	if field and isInArray(FIELDS, field:getId()) then
 		field:remove()
 		position:sendMagicEffect(CONST_ME_POFF)
--- a/data/spells/scripts/support/find_person.lua
+++ b/data/spells/scripts/support/find_person.lua
@@ -19,18 +19,18 @@ local directions = {
 }
 
 local messages = {
-    [DISTANCE_BESIDE] = {
-        [LEVEL_LOWER] = " is below you.",
-        [LEVEL_SAME] = " is standing next to you.",
-        [LEVEL_HIGHER] = " is above you."
-    },
-    [DISTANCE_CLOSE] = {
-        [LEVEL_LOWER] = " is on a lower level to the ",
-        [LEVEL_SAME] = " is to the ",
-        [LEVEL_HIGHER] = " is on a higher level to the "
-    },
-    [DISTANCE_FAR] = " is far to the ",
-    [DISTANCE_VERYFAR] =  " is very far to the "
+	[DISTANCE_BESIDE] = {
+		[LEVEL_LOWER] = " is below you.",
+		[LEVEL_SAME] = " is standing next to you.",
+		[LEVEL_HIGHER] = " is above you."
+	},
+	[DISTANCE_CLOSE] = {
+		[LEVEL_LOWER] = " is on a lower level to the ",
+		[LEVEL_SAME] = " is to the ",
+		[LEVEL_HIGHER] = " is on a higher level to the "
+	},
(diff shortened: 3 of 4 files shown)


data/spells/scripts/support/animate_dead_rune.lua
```diff
@@ -18,6 +18,6 @@ function onCastSpell(creature, variant, isHotkey)
 	end
 
 	creature:getPosition():sendMagicEffect(CONST_ME_POFF)
-	creature:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
+	creature:sendTextMessage(MESSAGE_STATUS_SMALL, Game.getReturnMessage(RETURNVALUE_NOTPOSSIBLE))
```
ninjalulz: Is there any specific reason to this change?

raymondtfr: To not use any old function from compat, and instead use the original function. (sendCancelMessage uses the sendTextMessage inside it, so its better to use the original function as the old one may be deprecated sometime?)

ninjalulz: [Player.sendCancelMessage](https://github.com/otland/forgottenserver/blob/master/data/lib/core/player.lua#L58-L63) is not a compatibility function though, and it's widely used in the data pack & source code.

ranisalt: It's not from compat, `sendCancelMessage` is different because it gets the message from a predefined list.

data/spells/scripts/support/food.lua
```diff
@@ -1,19 +1,19 @@
-local food = {
+local foods = {
 	2666, -- meat
 	2671, -- ham
 	2681, -- grape
 	2674, -- apple
 	2689, -- bread
 	2690, -- roll
-	2696  -- cheese
+	2696 -- cheese
```
ranisalt: You should add a trailing comma instead to keep it consistent and aligned.

data/spells/scripts/support/food.lua
```diff
@@ -1,19 +1,19 @@
...
 }
 
 function onCastSpell(creature, variant)
-	creature:addItem(food[math.random(#food)])
-	if math.random(1, 100) > 50 then
-		creature:addItem(food[math.random(#food)])
+	if math.random(1, 100) >= 50 then
```
ranisalt: You changed the chance from 50% to 51% by replacing `>` with `>=`. Also, as the chance is 50%, you can use `math.random(1, 2) == 1` instead.

Why did you change the order, exactly?

raymondtfr: Not exactly, ``math.random(1, 100)`` is a closed interval [a, b], so that [1, 100]  (has a ``b - 1 + 1`` which is ``100 - 1 + 1`` that is 100 different numbers (including 100), so ``>=`` could be read in a reversed order: ``50 <= math.random(1, 100)``, so if the random value was **50** we would get ``50 <= 50`` (b - a + 1) which is 1/2.
If I used ``<`` it would be a open interval (b - a), so it would exclude one element.

I had done it wrong when I updated this script a while ago, so I have now corrected the change.
The code is clearer with that statement moved below the ``if`` I guess.

ranisalt: Oh yeah, I'm more used to working with 0-based indexes and ranges and Lua patterns still confuse me :P

data/spells/scripts/support/find_person.lua
```diff
@@ -79,10 +79,8 @@ function onCastSpell(creature, variant)
 		elseif absValue < 2.4142 then
 			if distanceValue > 0 then
 				direction = offset.y > 0 and DIRECTION_NORTHWEST or DIRECTION_SOUTHEAST
-			elseif offset.x > 0 then
-				direction = DIRECTION_SOUTHWEST
 			else
-				direction = DIRECTION_NORTHEAST
+				direction = offset.x > 0 and DIRECTION_SOUTHWEST or DIRECTION_NORTHEAST
```
ranisalt: Though I usually complain about ternaries, this is a particularly nice usage 😋

data/spells/scripts/support/animate_dead_rune.lua
```diff
@@ -1,5 +1,5 @@
 function onCastSpell(creature, variant, isHotkey)
```
ranisalt: While you're at it, `isHotkey` is not needed here too.

data/spells/scripts/support/destroy_field_rune.lua
```diff
@@ -1,8 +1,7 @@
 function onCastSpell(creature, variant, isHotkey)
```
ranisalt: Also here.

data/spells/scripts/support/animate_dead_rune.lua
raymondtfr: But this is a rune spell #1020 , it was originally made for rune spells, so I just removed those from instant spells.

ranisalt: Oh, sure, didn't noticed it's a rune.

ranisalt:
Thanks!

Issue #278, opened 2013-12-10
onEquip executes two times, when we equip item. [$5]


djseban:
Hi
This bug exists in every version of TFS. When we equip item, which has custom lua script, function MoveEvents::onPlayerEquip is executed two times. For example we have lua script:
function onEquip(cid, item, slot)
     print(1)
end
After wearing item, output in TFS console is:
1
1
but should be:
1
##

gugahoa:
Right now onEquip can be called three times, two for checking and one to run the script. I would like to suggest adding the variable isCheck to onEquip/onDeEquip. Any thoughts? I already did this in a fork, if needed I can make a PR.

brunominervino:
@ranisalt @djarek , maybe 1.3 milestone?

djarek:
Fixed in #2058

gugahoa:
How do I claim my lunch? Ops, I mean the bounty

djarek:
Ask @Znote .

Znote:
![image](https://cloud.githubusercontent.com/assets/1552144/24285937/ef85100a-1072-11e7-8d29-3754c1ef76f0.png)

PR #2058, opened 2017-01-06
Proposed fix to #278

gugahoa:
#278 
Add isCheck variable to onEquip to differentiate between when it's being called from queryAdd and when it's being called from postAdd.
It's added to onDeEquip too, but since it's always called from postRemove, isCheck will always be false

### DIFF
--- a/src/movement.cpp
+++ b/src/movement.cpp
@@ -340,7 +340,7 @@ uint32_t MoveEvents::onPlayerDeEquip(Player* player, Item* item, slots_t slot)
 	if (!moveEvent) {
 		return 1;
 	}
-	return moveEvent->fireEquip(player, item, slot, true);
+	return moveEvent->fireEquip(player, item, slot, false);
 }
 
 uint32_t MoveEvents::onItemMove(Item* item, Tile* tile, bool isAdd)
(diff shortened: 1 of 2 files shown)


src/movement.cpp
```diff
@@ -820,16 +820,16 @@ bool MoveEvent::executeStep(Creature* creature, Item* item, const Position& pos)
 uint32_t MoveEvent::fireEquip(Player* player, Item* item, slots_t slot, bool boolean)
 {
 	if (scripted) {
-		return executeEquip(player, item, slot);
+		return executeEquip(player, item, slot, boolean);
```
djarek: While you're at it, you could change the name of the boolean variable to something that makes more sense.

gugahoa: Done 😃

src/movement.cpp
```diff
@@ -817,19 +817,19 @@ bool MoveEvent::executeStep(Creature* creature, Item* item, const Position& pos,
 	return scriptInterface->callFunction(4, true);
 }
 
-uint32_t MoveEvent::fireEquip(Player* player, Item* item, slots_t slot, bool boolean)
+uint32_t MoveEvent::fireEquip(Player* player, Item* item, slots_t slot, bool isCheck)
```
ninjalulz: While you're at it, can't you rename the variable to something more meaningful? Perhaps something that contains the keyword 'abilities'. 🤔

gugahoa: Sorry, I can't think of anything more meaningful than isCheck to a boolean that is true when the function is called only to check the return value :(

ninjalulz: Okay, haha 😄

gugahoa:
Review status? @djarek \cc @ranisalt

djarek:
Thanks!

PR #2061, opened 2017-01-08
Check tile before calling function

gugahoa:
This will fix a segmentation fault when a player logs in on top of a onStep event that deals damage.
Now that onStepIn was moved to queryAdd, if a player logs in, he'll receive damage before being placed on the map so tile can be null

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2,7 +2,7 @@
  * The Forgotten Server - a free and open-source MMORPG server emulator
  * Copyright (C) 2017 Mark Samman <mark.samman@gmail.com>
  *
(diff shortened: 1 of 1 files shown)


src/player.cpp
```diff
@@ -2,7 +2,7 @@
  * The Forgotten Server - a free and open-source MMORPG server emulator
  * Copyright (C) 2017 Mark Samman <mark.samman@gmail.com>
  *
- * This program is free software; you can redistribute it and/or modify
+ * This program is free tisoftware; you can redistribute it and/or modify
```
Kamenuvol: wrong place hehe

gugahoa: Ops

Issue #2024, opened 2016-12-15
[Crash] Creature::executeConditions


ninjalulz:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Use an [onPrepareDeath script](http://pastebin.com/cuUcsBDP) which removes one condition or more with the help of Creature.removeCondition. However you can bypass the crash if you replace L9 with condition:setTicks(0), but it doesn't remove the condition immediately.
2. Kill your target with a dot (damage-over-time) spell or rune.

### Expected behaviour
Removes the condition safely.

### Actual behaviour
Server crashes. It might not always crash for you, but it did for me at least.

### Stacktrace
http://pastebin.com/cQKYN9kW

### Environment
macOS 10.12 (Sierra) / Ubuntu 15.10 (Wily Werewolf).

wgrriffel:
i can confirm this.

wgrriffel:
i think this block https://github.com/otland/forgottenserver/blob/master/src/creature.cpp#L1302-L1311 `it` is a dangling pointer, since it was deleted on `Creature::removeCondition`, not sure tho.

vankk:
The fix for this issue is:

`void Creature::executeConditions(uint32_t interval)
{
    auto it = conditions.begin(), end = conditions.end();
     while (it != end) {
        Condition* condition = *it;
        if(!condition) {
            it = conditions.erase(it);
            continue;
        } else if (!condition->executeCondition(this, interval)) {
            ConditionType_t type = condition->getType();

            it = conditions.erase(it);

            condition->endCondition(this);
            delete condition;

            onEndCondition(type);
        } else {
            ++it;
        }
    }
}`

gugahoa:
I can't reproduce it no matter how many times I try, could you share the crash report if possible?

ninjalulz:
@gugahoa You can view it by clicking on the pastebin link above.

gugahoa:
Ops, I expected a wall of text somewhere in the issue hahaha

gugahoa:
Could reliable reproduce this: Step on fire field and let the tick damage kill you

Eternal-Scripts:
[...]

PR #2062, opened 2017-01-08
Fix #2054

gugahoa:
Iterate over a copy of conditions during conditions execution to prevent iterator invalidation.
During ConditionDamage::executeCondition, Lua scripts can someway or another delete a condition invalidating the iterator. For example the onPrepareDeath script at #2054

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1292,13 +1292,16 @@ Condition* Creature::getCondition(ConditionType_t type, ConditionId_t conditionI
 
 void Creature::executeConditions(uint32_t interval)
 {
-	auto it = conditions.begin(), end = conditions.end();
+	ConditionList conditionsCopy;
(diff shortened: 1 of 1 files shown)


marksamman:
Have you confirmed that this actually works? Using the iterator of the copy to erase from the original list seems fragile.

gugahoa:
I'll test more thoroughly to give a more solid answer

gugahoa:
Tested against the issue it's fixing, and when a conditions kills someone or when it doesn't.

vankk:
I still have crashes with #2024 even with this PR.

gugahoa:
@vankk could you test again? It's working as intended here, what steps are you following?

gugahoa:
Dropped in favor of: https://github.com/Mkalo/forgottenserver/commit/746ee5d0a18a605c89771214b61980d9a4016711

vankk:
@gugahoa with the fix that @Mkalo provided it was been fixed.

PR #2063, opened 2017-01-08
Change default schema town_id

wgrriffel:
only for convince, since tfs default map there is no such townid.


Kamenuvol:
btw, we should update tfs map somehow

DSpeichert:
@wgrriffel Why is this needed? Is the town ID when added by RME starting from 1?

WibbenZ:
@DSpeichert Yes it starts from 1.
It would also help when it comes to newbies who get this error;
https://github.com/otland/forgottenserver/blob/06bf9c9ec16ed0d3e9313e40b1266b68a67159cb/src/iologindata.cpp#L372-L376

Also as @wgrriffel said, the default map starts at 1 and with RL maps it would be Venore.

Issue #2065, opened 2017-01-08
onStepOut / closingdoor.lua bug


Zbizu:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. create gate of expertise
2. set action id e.g 1001
3. use and step out of it

### Expected behaviour
level door should close

### Actual behaviour
it stays open because
```tile:getCreatureCount()```
returns 1 instead of 0 in closingdoor.lua (movement)

### Environment
same as in #2059, probably caused by recent changes in sources

**Comment by PrinterLUA** [comment]
Since the latest commit using return values onStepIn and onStepOut #2049

So instead of using teleportTo to "block" someone from enter. We should return false instead.

**Comment by Zbizu** [comment]
@PrinterLUA blocking someone from enter works fine
stepping out is the problem because door doesn't close

**Comment by Zbizu** [comment]
workaround
```
local function tpOthers(oldPos, newPos)
	local c = Tile(oldPos):getCreatures()
	for i = 1, #c do
		c[i]:teleportTo(newPos)
	end
end

function onStepOut(creature, item, position, fromPosition)
	local tile = Tile(position)
	local newPosition = {x = position.x + 1, y = position.y, z = position.z}
	local query = Tile(newPosition):queryAdd(creature)
	if query ~= RETURNVALUE_NOERROR or query == RETURNVALUE_NOTENOUGHROOM then
		newPosition.x = newPosition.x - 1
		newPosition.y = newPosition.y + 1
		query = Tile(newPosition):queryAdd(creature)
	end

	if query == RETURNVALUE_NOERROR or query ~= RETURNVALUE_NOTENOUGHROOM then
		addEvent(tpOthers, 100, item:getPosition(), newPosition)
		local it = tile:getItems()
		for i = 1, #it do
			if it[i]:getWeight() > 0 then
				it[i]:moveTo(newPosition)
			end
		end
		local f = tile:getFieldItem()
		if f then
			f:remove()
		end
	end

	item:transform(item.itemid - 1)
	return true
end
```

Zbizu:
tested it now on a fresh tfs, no longer bugged

PR #2069, opened 2017-01-08
onStep return value matters + side effect fixes

### LINKED ISSUE #2067: Creature::updateMapCache triggering onStepIn event
Because onStepIn was moved to Tile::queryAdd, Creature::updateMapCache is triggering it generating nasty side effects such as creature receiving damage when caching the map and there's a magic field in range.

How to reproduce:
1. Create a fire field (1487)
2. Summon a rat anywhere close to it

gugahoa:
Closes #2067 #2065


src/creature.cpp
```diff
@@ -390,17 +390,13 @@ void Creature::onRemoveTileItem(const Tile* tile, const Position& pos, const Ite
 	}
 }
 
-void Creature::onCreatureAppear(Creature* creature, bool isLogin)
```
WibbenZ: Is there even a need to keep the bool?

src/movement.cpp
```diff
@@ -295,15 +295,15 @@ MoveEvent* MoveEvents::getEvent(const Tile* tile, MoveEvent_t eventType)
 	return nullptr;
 }
 
-uint32_t MoveEvents::onCreatureMove(Creature* creature, const Tile* tile, MoveEvent_t eventType)
+uint32_t MoveEvents::onCreatureMove(Creature* creature, const Tile* tile, const Position& fromPos, MoveEvent_t eventType)
```
WibbenZ: Can't you use fromPosition?

src/luascript.h
```diff
@@ -240,7 +240,7 @@ class LuaScriptInterface
 		bool pushFunction(int32_t functionId);
 
 		static int luaErrorHandler(lua_State* L);
-		bool callFunction(int params);
+		bool callFunction(int params, bool defaultReturn = false);
```
conde2: I Think that should be true for backwards compatibility.

gugahoa: It's true, but only when called from the onStep functions, for the rest it should be false

gugahoa:
FLAG_PATHFINDING is the right one to use on ::queryAdd as it is what's passed when it's called only to cache, find path and so on, things that wouldn't trigger onStepIn even if ::queryAdd returned true because it would never call ::postAdd or ::postRemove)
But for the FLAG_NOLIMIT I completely agree with you that there should be a FLAG_IGNORESTEPOUT
If it's needed I can add it

conde2:
What still left to do in this pull request?

ranisalt:
Rebasing.

conde2:
Is this guy active ? If not I can rebase it and references the credits to him.

ranisalt:
Yes, he is. In fact he is a friend of mine and I'll notify him now.

gugahoa:
Rebased. For now I can't test it against the bugs reported in the previous merge, if anyone could test it before me I would appreciate the help a lot. Thanks!

data/talkactions/scripts/reload.lua
```diff
@@ -8,9 +8,6 @@ local reloadTypes = {
 	["channel"] = { targetType = RELOAD_TYPE_CHAT, name = "chatchannels" },
 	["chatchannels"] = { targetType = RELOAD_TYPE_CHAT, name = "chatchannels" },
 
-	["command"] = { targetType = RELOAD_TYPE_COMMANDS, name = "commands" },
-	["commands"] = { targetType = RELOAD_TYPE_COMMANDS, name = "commands" },
-
```
Mkalo: Already fixed in #2172. Those commits looks pretty messy now maybe remake?

ranisalt: It's because he merged instead of rebasing. It won't count as changed lines if merged, but it's ugly anyway.

PR #2070, opened 2017-01-08
Updated Party spells

raymondtfr:
* Created the *metatable function* ``Creature:addPartyCondition(combat, variant, condition, baseMana)`` in ``lib/spells.lua`` out of the code shared by all 4 spells.
* Renamed variable ``membersList`` to ``members`` and ``affectedList`` to ``affectedMembers``.
* Removed variable ``count`` (a placeholder for ``#affectedMembers``).

### DIFF
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -322,7 +322,7 @@ function Creature:addPartyCondition(combat, variant, condition, baseMana)
 		return false
 	end
 
-	local mana = math.ceil((math.pow(0.9, #affectedMembers - 1) * baseMana) * #affectedMembers)
+	local mana = math.ceil(math.pow(0.9, #affectedMembers - 1) * baseMana * #affectedMembers)
 	if self:getMana() < mana then
 		self:sendTextMessage(MESSAGE_STATUS_SMALL, Game.getReturnMessage(RETURNVALUE_NOTENOUGHMANA))
 		position:sendMagicEffect(CONST_ME_POFF)
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -296,3 +296,57 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
 	target:addCondition(condition)
 	return true
 end
+
+function Creature:addPartyCondition(combat, variant, condition, baseMana)
+	local position = self:getPosition()
+	local party = self:getParty()
+	if not party then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	local members = party:getMembers()
+	members[#members + 1] = party:getLeader()
+
+	if #members == 1 then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	local affectedMembers = {}
+	for _, member in ipairs(members) do
+		if member:getPosition():getDistance(position) <= 36 then
+			affectedMembers[#affectedMembers + 1] = member
+		end
+	end
+
+	if #affectedMembers <= 1 then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	local mana = math.ceil((math.pow(0.9, #affectedMembers - 1) * baseMana) * #affectedMembers)
+	if self:getMana() < mana then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, Game.getReturnMessage(RETURNVALUE_NOTENOUGHMANA))
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	if not combat:execute(self, variant) then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, Game.getReturnMessage(RETURNVALUE_NOTPOSSIBLE))
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	self:addMana(baseMana - mana, false)
+	self:addManaSpent(mana - baseMana)
+
+	for _, member in ipairs(affectedMembers) do
+		member:addCondition(condition)
+	end
+	return true
+end
+
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -309,12 +309,6 @@ function Creature:addPartyCondition(combat, variant, condition, baseMana)
 	local members = party:getMembers()
 	members[#members + 1] = party:getLeader()
 
-	if #members == 1 then
-		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
-		position:sendMagicEffect(CONST_ME_POFF)
-		return false
-	end
-
 	local affectedMembers = {}
 	for _, member in ipairs(members) do
 		if member:getPosition():getDistance(position) <= 36 then
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -298,17 +298,17 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
 end
 
 function Creature:addPartyCondition(combat, variant, condition, baseMana)
-	local position = self:getPosition()
 	local party = self:getParty()
 	if not party then
-		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
-		position:sendMagicEffect(CONST_ME_POFF)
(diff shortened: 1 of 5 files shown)


data/spells/lib/spells.lua
```diff
@@ -258,6 +258,16 @@ RANGE = {
 [16] = {1, 2},   [17] = {1, 2},   [18] = {1, 2}, [19] = {1, 2}
 }
 
+COORDINATES_AREA_CIRCLE5X5 = {
+{x = 0, y = 0},
+{x = 1, y = 0}, {x = 2, y = 0}, {x = 3, y = 0}, {x = 4, y = 0}, {x = 5, y = 0},
```
WibbenZ: This looks really messy to me

raymondtfr: Yes, it does, anyway this is far more complicated than I thought it would be, such as checking if the combat area effect actually hit a party member (if blocked by walls, etc...).

data/spells/lib/spells.lua
```diff
@@ -296,3 +306,74 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
...
+				return true
+			elseif casterPosition.x - axis.x == targetPosition.x and casterPosition.y - axis.y == targetPosition.y then
+				return true
+			elseif casterPosition.x + axis.x == targetPosition.x and casterPosition.y - axis.y == targetPosition.y then
+				return true
+			end
+		end
+	end
+	return false
+end
```
ranisalt: This is just a horrible way to use plain old areas, literally. You could iterate over the spell area instead. I don't understand why you changed this.

data/spells/lib/spells.lua
```diff
@@ -296,3 +306,74 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
...
+	local position = self:getPosition()
+	local party = self:getParty()
+	if not party then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	local membersList = party:getMembers()
+	membersList[#membersList + 1] = party:getLeader()
```
ranisalt: Isn't `table.insert` more commonly used?

data/spells/lib/spells.lua
WibbenZ: Well you can atleast clean the table up abit.
As I told you before, making it shorter isen't always better.

We have to keep in mind that people might have to change your code at some time and it will help ALOT if it's clean and easy to read, if I found a bug and tracked it down to this table I would just scrap it because of how messy it is, creating more work for no reason.

raymondtfr: I'll just leave the party condition function here and remove the others related to the bug fix (this table included), as it wouldn't take care of all the possible cases.

data/spells/lib/spells.lua
raymondtfr: I don't understand what you mean ehhe, anyway I'll remove this function and open an issue as I can't think of a way to properly and fully fix the bug I mentioned.

data/spells/lib/spells.lua
raymondtfr: It's been like this since tfs 1.0, although ``table.insert`` states better what it does, and in this case I think it would be cool to use it (performance related) but probably not in line 348??

data/spells/lib/spells.lua
ranisalt: The idea of abstracting a long, repeated code from a lot of functions is good. You just didn't figure a good way to work with the coordinates 😛

raymondtfr: I thought I could get rid of this bug just by using Cartesian coordenates (that's why I had to create that table) and checking if a party member was on any square of the combat area (I just added the first quadrant of the original area to that table because all others Q's were symmetric) and also checking if the member was on the same floor level.

I thought about using tile, target combat callbacks but that wouldn't work because of the *required mana formula* which is based on number of affected members (the formula must be used only once to get the correct value.)

data/spells/lib/spells.lua
```diff
@@ -296,3 +296,57 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
...
+	if not party then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	local members = party:getMembers()
+	members[#members + 1] = party:getLeader()
+
+	if members == nil or type(members) ~= 'table' or #members <= 1 then
```
ranisalt: This test is very strange. `getMembers()` will always return a table if `party` is valid, which it is at this point. You shouldn't need to check: if it's nil, because it isn't; it it's a table, because it is; if length is 0 or 1, because it is not (AFAIK a party must have 2 or more players).

data/spells/lib/spells.lua
```diff
@@ -296,3 +296,57 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
...
+		end
+	end
+
+	if #affectedMembers <= 1 then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "No party members in range.")
+		position:sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	local mana = math.ceil((0.9 ^ (#affectedMembers - 1) * baseMana) * #affectedMembers)
```
ranisalt: You should use math.pow here, ^ is not a good operator.

ranisalt:
> Replaced creature:sendCancelMessage (compat) with creature:sendTextMessage function.

The former is not from compat, so please revert tihis change.

data/spells/lib/spells.lua
```diff
@@ -296,3 +296,57 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
 	target:addCondition(condition)
 	return true
 end
+
+function Creature:addPartyCondition(combat, variant, condition, baseMana)
+	local position = self:getPosition()
```
ranisalt: Throughout the function, you will never need this variable more than once. You can replace all instances of `position` with `self:getPosition()` and save 1 variable.

data/spells/lib/spells.lua
raymondtfr: It's been a long time since I've used Party feature in Tibia and by that I mean I don't know if it is possible to just have one person at a Party (i.e. a player invites another player (which accepts the request), and then the leader (or the invitee) leaves the party, and the user which didn't leave receives the leadership and now its a party of just 1 member).

Although I tested in TFS, and as you said it is only possible to have 2 or more members in a party, so I'll just change it accordingly.

data/spells/lib/spells.lua
raymondtfr: We use ``position`` in 
```lua
	local affectedMembers = {}
	for _, member in ipairs(members) do
		if member:getPosition():getDistance(position) <= 36 then
			affectedMembers[#affectedMembers + 1] = member
		end
	end
```
so I guess it would make more sense if I move this variable declaration to just above ``local affectedMembers = {}``

raymondtfr:
I replaced ``creature:sendCancelMessage`` with ``creature:sendTextMessage`` because I find it better because it does not obfuscates code, and what I meant with it about compat, is that it is used in this function located in compat:
```lua
function doPlayerSendCancel(cid, text) local p = Player(cid) return p ~= nil and p:sendCancelMessage(text) or false end
```
which is in a way serving as a compatability function, but indirectly.

data/spells/lib/spells.lua
ranisalt: Yes, that would be better 😄

ranisalt:
IMHO `sendCancelMessage` expresses *intent*, i.e. it *means* we are sending a message cancelling an action. Of course half of the functions could be removed if we just copied and pasted their code everywhere but it's about *intention* here, and if one wants to change all cancel messages' style it is done in one line.

It has also hides uses of `Game.getReturnMessage` which is a pretty strange function to call in a script.

ranisalt:
IMHO `sendCancelMessage` expresses *intent*, i.e. it *means* we are sending a message cancelling an action. Of course half of the functions could be removed if we just copied and pasted their code everywhere but it's about *intention* here, and if one wants to change all cancel messages' style it is done in one line.

It has also hides uses of `Game.getReturnMessage` which is a pretty strange function to call in a script.

That's more about coding style than real utility of the code.

raymondtfr:
But using ``sendTextMessage`` allow us to maintain uniformity across all data files.

For example, we are limited to only one kind of text type when using ``sendCancelMessage`` (i.e. ``MESSAGE_STATUS_SMALL``), so if we ever wanted to use another text type we would be forced to use ``sendTextMessage`` instead. Of course we could update ``Player.sendCancelMessage`` and ``doPlayerSendCancel`` (compat) to accomplish this, but it'll just do the same thing as the original function.

I'll revert it anyhow (also in #2057 ).

ranisalt:
Thanks!

PR #2071, opened 2017-01-09
Move house management function to Lua scripts

ranisalt:
This moves house management spells (aleta sio, aleta som, aleta grav and alana sio) to Lua scripts.

aleta sio and aleta som are working great, but I'm not sure about aleta grav. It inserts (and deletes, when erased) multiple rows at `house_lists` table, but ingame it works properly and it should not be related to these changes.

alana sio also works great but GM players can't kick themselves as they have the `CanEditHouses` flag set and [House::kickPlayer](https://github.com/otland/forgottenserver/blob/master/src/house.cpp#L160) disallows. I could fix while I'm at it.

### DIFF
--- a/data/spells/scripts/house/edit_door_list.lua
+++ b/data/spells/scripts/house/edit_door_list.lua
@@ -1,27 +1,11 @@
-local function getNextPosition(player)
-	local position = player:getPosition()
-	local direction = player:getDirection()
-
-	-- a player can only be facing cardinal directions
-	if direction == DIRECTION_NORTH then
-		position.y = position.y - 1
-	elseif direction == DIRECTION_SOUTH then
-		position.y = position.y + 1
-	elseif direction == DIRECTION_WEST then
-		position.x = position.x - 1
-	else
-		position.x = position.x + 1
-	end
-	return position
-end
-
 function onCastSpell(player, variant)
 	local house = player:getTile():getHouse()
 	if not house then
 		return false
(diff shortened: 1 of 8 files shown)


raymondtfr:
I am not sure but I think I saw @marksamman mentioning somewhere about moving **commands** and **house spells** to *talkactions*.

Kamenuvol:
I believe they are spells because they are shown in spellbook. But even if he want to move to talkactions, it's easier to change now.

ranisalt:
Currently, those 4 house commands are spells even in Tibia, I don't see the point in moving them to talkactions unless they change to be more like commands, e.g. `/guests` or `/kick playerName`. Internally spells are talkactions, too.

There is a PR (#1990) moving the remaining commands (`/raid`, `/reload` and `!sellhouse`) to talkactions.

data/spells/scripts/house/edit_door_list.lua
```diff
@@ -0,0 +1,33 @@
+local function getNextPosition(player)
```
ninjalulz: I think you should remove this redundant function, and use Position.getNextPosition instead.

ranisalt: Sure! Didn't see it already existed in libs.

ranisalt:
Let's discuss this spell vs talkaction later, open an issue if you want.

PR #2072, opened 2017-01-09
Remove player from attack list upon death (Fixes a yellow skull issue)

ninjalulz:
This should eliminate the possibility of receiving a yellow skull after killing another player (if none of them has any skull).

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -624,6 +624,7 @@ void Creature::onDeath()
 
 	Creature* mostDamageCreature = nullptr;
 
+	const Player* targetPlayer = getPlayer();
 	const int64_t timeNow = OTSYS_TIME();
 	const uint32_t inFightTicks = g_config.getNumber(ConfigManager::PZ_LOCKED);
 	int32_t mostDamage = 0;
@@ -638,8 +639,12 @@ void Creature::onDeath()
 
 			if (attacker != this) {
 				uint64_t gainExp = getGainedExperience(attacker);
-				if (Player* player = attacker->getPlayer()) {
-					Party* party = player->getParty();
+				if (Player* attackerPlayer = attacker->getPlayer()) {
+					if (targetPlayer && attackerPlayer->hasAttacked(targetPlayer)) {
+						attackerPlayer->removeAttacked(targetPlayer);
+					}
+
+					Party* party = attackerPlayer->getParty();
 					if (party && party->getLeader() && party->isSharedExperienceActive() && party->isSharedExperienceEnabled()) {
 						attacker = party->getLeader();
 					}
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -639,10 +639,7 @@ void Creature::onDeath()
 			if (attacker != this) {
 				uint64_t gainExp = getGainedExperience(attacker);
 				if (Player* attackerPlayer = attacker->getPlayer()) {
-					Player* targetPlayer = getPlayer();
-					if (targetPlayer && attackerPlayer->hasAttacked(targetPlayer)) {
-						attackerPlayer->removeAttacked(targetPlayer);
-					}
+					attackerPlayer->removeAttacked(getPlayer());
 
 					Party* party = attackerPlayer->getParty();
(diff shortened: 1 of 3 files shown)


src/creature.cpp
```diff
@@ -638,8 +639,12 @@ void Creature::onDeath()
 
 			if (attacker != this) {
 				uint64_t gainExp = getGainedExperience(attacker);
-				if (Player* player = attacker->getPlayer()) {
-					Party* party = player->getParty();
+				if (Player* attackerPlayer = attacker->getPlayer()) {
+					if (targetPlayer && attackerPlayer->hasAttacked(targetPlayer)) {
```
ranisalt: Get the target player here instead of up above where it is long unneeded yet.

ninjalulz: Fine, I'll move it xD

src/player.cpp
```diff
@@ -3751,6 +3751,20 @@ void Player::addAttacked(const Player* attacked)
 	attackedSet.insert(attacked->guid);
 }
 
+void Player::removeAttacked(const Player* attacked)
+{
+	if (!attacked || attacked == this) {
+		return;
+	}
+
+	auto it = std::find(attackedSet.begin(), attackedSet.end(), attacked->guid);
```
ranisalt: A set is implemented as a red-black tree and it's faster to traverse with `attackedSet.find(attacked->guid)` than the generic `std::find`.

src/player.cpp
```diff
@@ -3751,6 +3751,20 @@ void Player::addAttacked(const Player* attacked)
...
+	if (it == attackedSet.end()) {
+		return;
```
ranisalt: You can save a few lines if you use `!= attackedSet.end()` and erase in here.

ranisalt: Is there an issue reporting this bug?

ranisalt:
Is there an issue reporting this bug?

ninjalulz:
No, I don't think there is (I'm unable to find one at least).

src/player.cpp
```diff
@@ -3751,6 +3751,18 @@ void Player::addAttacked(const Player* attacked)
 	attackedSet.insert(attacked->guid);
 }
 
+void Player::removeAttacked(const Player* attacked)
+{
+	if (!attacked || attacked == this) {
+		return;
+	}
+
+	auto it = attackedSet.find(attacked->guid);
```
ranisalt: Note that `hasAttacked` that is called before this function already performs this check. You probably don't need to check `hasAttacked` then.

src/creature.cpp
```diff
@@ -638,8 +638,13 @@ void Creature::onDeath()
 
 			if (attacker != this) {
 				uint64_t gainExp = getGainedExperience(attacker);
-				if (Player* player = attacker->getPlayer()) {
-					Party* party = player->getParty();
+				if (Player* attackerPlayer = attacker->getPlayer()) {
+					Player* targetPlayer = getPlayer();
+					if (targetPlayer && attackerPlayer->hasAttacked(targetPlayer)) {
+						attackerPlayer->removeAttacked(targetPlayer);
+					}
```
ranisalt: See my comment below, I think those 4 lines above might be reduced to `attackerPlayer->removeAttacked(getPlayer())`, but please double check.

ranisalt:
I think you just solved a dangling pointer issue :)

src/player.cpp
gugahoa: I think it's better to be safe than sorry, as later it can be used in other context if need be

src/creature.cpp
ninjalulz: Yep! You are absolutely right, I didn't really pay attention to that.

src/player.cpp
ranisalt: @gugahoa he doesn't need to check if `hasAttacked` prior to calling `removeAttacked`, not the other way around

ranisalt:
Thanks!

src/player.cpp
gugahoa: Oh, sorry, I misunderstood your comment haha

PR #2075, opened 2017-01-11
Fix #2074 double door listing on houses

### LINKED ISSUE #2074: Multiple database entries for the same house door list

1. Own a guild hall
2. Add a player to a door list with aleta grav
3. Save the server

### Expected behaviour
A single entry in the `house_lists` table should be used to store that door name list.

### Actual behaviour
Multiple entries with the same content are stored

Imgur

ranisalt:
Changing from list to set seems to fix as the same door is no longer double inserted. Also it reduces complexity of searching for a door to logarithmic.

Fixes #2074.

### DIFF
--- a/src/house.h
+++ b/src/house.h
@@ -21,6 +21,7 @@
 #define FS_HOUSE_H_EB9732E7771A438F9CD0EFA8CB4C58C4
 
 #include <regex>
+#include <set>
 
 #include "container.h"
 #include "housetile.h"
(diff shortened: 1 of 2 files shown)


src/house.h
```diff
@@ -234,7 +235,7 @@ class House
 		Container transfer_container{ITEM_LOCKER1};
 
 		HouseTileList houseTiles;
-		std::list<Door*> doorList;
+		std::set<Door*> doorSet;
```
djarek: std::unordered_set?

ranisalt: I'm not sure the overhead is worth it. Most houses don't even have internal doors, and guild halls generally between 4 and 8 AFAIK.

ranisalt:
Thanks!  🤣

PR #2076, opened 2017-01-11
Add spell methods to Lua

WibbenZ:
New class:

```
Spell(id)
Spell(name)
Spell(words)
```

New methods:

```
Spell:getName()
Spell:getWords()
Spell:getLevel()
Spell:getMagicLevel()
Spell:getManaCost()
Spell:getManaPercent()
Spell:getSoulCost()
Spell:isPremium()
Spell:isLearnable()

Player:getSpellCount()
Player:canCast(spell)
```

Moved getPlayerInstantSpellInfo to compat
Moved getPlayerInstantSpellCount to compat


src/luascript.cpp
```diff
@@ -12175,6 +12171,133 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
 	return 1;
 }
 
+// Spells
+int LuaScriptInterface::luaSpellCreate(lua_State* L)
+{
+	// Spell(words, name or id)
+	InstantSpell* spell;
+	if (isString(L, 2)) {
```
Mkalo: isString uses lua_isstring which will return true for numbers and strings, you should first check if it's number using isNumber and then check with isString. I don't even think this is working for numbers as that else if will never get triggered. 

```cpp
if (isNumber(L, 2)) {
	spell = g_spells->getInstantSpellById(getNumber<uint32_t>(L, 2));
} else if (isString(L, 2)) {
	const std::string& stringArgument = getString(L, 2);
	spell = g_spells->getInstantSpellByName(stringArgument);
	if (!spell) {
		spell = g_spells->getInstantSpell(stringArgument); 
	}
}
```

src/luascript.cpp
```diff
@@ -12175,6 +12171,133 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+		const std::string& stringArgument = getString(L, 2);
+		spell = g_spells->getInstantSpellByName(stringArgument);
+		if (spell == nullptr) {
+			spell = g_spells->getInstantSpell(stringArgument);
+		} else if (isNumber(L, 2)) {
+			spell = g_spells->getInstantSpellById(getNumber<uint32_t>(L, 2));
+		}
+	} else {
```
Mkalo: Above change to remove this else.

src/luascript.cpp
```diff
@@ -12175,6 +12171,133 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
 	return 1;
 }
 
+// Spells
+int LuaScriptInterface::luaSpellCreate(lua_State* L)
+{
+	// Spell(words, name or id)
+	InstantSpell* spell;
```
Mkalo: InstantSpell* spell = nullptr;

src/spells.cpp
```diff
@@ -234,17 +234,11 @@ uint32_t Spells::getInstantSpellCount(const Player* player) const
...
-	uint32_t count = 0;
-	for (const auto& it : instants) {
-		InstantSpell* instantSpell = it.second;
-		if (instantSpell->canCast(player)) {
-			if (count == index) {
-				return instantSpell;
-			}
-			++count;
-		}
+	auto& it = std::advance(instants.begin(), spellId);
```
WibbenZ: @ranisalt im totally lost here, I tried making it a ref but it still dosn't wanna work. xD

Mkalo: std::advance just advance the iterator, do it like:
```cpp
auto& it = instants.begin():
std::advance(it, spellId);
```

src/spells.cpp
```diff
@@ -236,7 +236,8 @@ uint32_t Spells::getInstantSpellCount(const Player* player) const
 
 InstantSpell* Spells::getInstantSpellById(uint32_t spellId)
 {
-	auto& it = std::advance(instants.begin(), spellId);
+	auto& it = instants.begin();
+	std::advance(it, spellId);
```
Mkalo: I'm not sure but if you use advance past the container's end. It will most likely crash when you do it->second. Maybe change it to a for loop (std::advance and std::next do that)

src/spells.cpp
```diff
@@ -234,17 +234,12 @@ uint32_t Spells::getInstantSpellCount(const Player* player) const
...
-	for (const auto& it : instants) {
-		InstantSpell* instantSpell = it.second;
-		if (instantSpell->canCast(player)) {
-			if (count == index) {
-				return instantSpell;
-			}
-			++count;
-		}
+	auto& it = instants.begin();
+	std::advance(it, spellId);
```
ranisalt: You can't catch a reference to rvalue, so `it` is invalid just as it's assigned as the return from `begin` is destroyed. The Travis build broke because of this but Appveyor didn't because VS compiler is a big pile of shit.

You should use `auto it = std::next(instants.begin(), spellId)`.

Better yet, since the spell list does not change too often (it is only completely reloaded, but never changed), it can be replaced with a vector and it would be both easier to check if given spell exists (by checking the `index <= size()`)  and getting the value (`return vector[index]` vs iterator mess).

src/spells.cpp
Mkalo: Something like:
```cpp
InstantSpell* Spells::getInstantSpellById(uint32_t spellId)
{
	uint32_t count = 0; 
	for (auto it = instants.begin(); it != instants.end(); it++, count++) {
		if (count == spellId) {
			return it->second;
		}
	}
	return nullptr;
}```

Comment (ranisalt) [inline src/spells.cpp]: That kills the point of the change. Since he checks if the iterator is past the end right below this line, it's fine.

File: src/spells.cpp (hunk 4 above)
Comment (ranisalt) [inline src/spells.cpp]: Don't make it a reference, you can't ref to "function returns" (technical name is rvalue). Also, use `std::next` instead.

File: src/spells.cpp (hunk 5 above)
Comment (Mkalo) [inline src/spells.cpp]: As I said, std::advance or std::next past the end of the container will give you undefined behavior:
```cpp
int main() {
	std::map<int32_t, int32_t> test = {
		{1, 2},
		{3, 4},
		{5, 6}
	};
	auto it = std::next(test.begin(), 12);
	if (it != test.end()) {
		std::cout << it->second << std::endl;
	}
	return 0;
}
```
This is printing 6 for me compiling with gcc.

ranisalt: I just [read here](http://en.cppreference.com/w/cpp/iterator/advance) that:

> The behavior is undefined if the specified sequence of increments or decrements would require that a non-incrementable iterator (such as the past-the-end iterator) is incremented, or that a non-decrementable iterator (such as the front iterator or the singular iterator) is decremented.

So it's probably smart to use `std::next(test.begin(), std::max(n, test.size()))`.

Mkalo: Guess you meant std::min :)

ranisalt: Yep. I tested with min and it works ok.

Mkalo:
What about a Game.getInstantSpells() method and a Player:getInstantSpells() method?
Should return table with Spell objects.

src/spells.cpp
WibbenZ: Sure w/e you want :P 
But my plan with this PR was just to add basic support (also refering to mkalo's comment below) making it easier to add new stuff related to the spell class.
Ill do the changes tomorow.

ranisalt: Ofc, just fix it and I can merge it. Changing it is for another PR.

ranisalt:
Thanks!

WibbenZ:
Oh thanks for the changes :+1:

PR #2077, opened 2017-01-12
Update spells lib

raymondtfr:
By the way, I have also found a pretty straightforward formula to use with ``listType == 1`` from a guy I discussed with about this same subject in *tibia.wikia*, so I may update ``creature:addDamageCondition`` (you would still be able to use rounds table) and #1940 to use it.

### DIFF
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -259,6 +259,10 @@ RANGE = {
 }
 
 function Creature:addDamageCondition(target, conditionType, listType, damage, time, rounds)
+	if self:isImmune(conditionType) then
+		return false
+	end
+
 	local condition = Condition(conditionType)
 	condition:setParameter(CONDITION_PARAM_OWNER, self:getId())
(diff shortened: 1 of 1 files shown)


data/spells/lib/spells.lua
```diff
@@ -259,6 +259,10 @@ RANGE = {
 }
 
 function Creature:addDamageCondition(target, conditionType, listType, damage, time, rounds)
+	if self:isImmune(conditionType) then
```
ninjalulz: A typo.

raymondtfr: I can not see a typo anywhere in this line??

ninjalulz: Is it supposed to check if the caster (self) is immune to this conditionType?

raymondtfr: Oh, I see now, you are right, it should be the target lol. By the way I also update the script to use ``math.pow`` instead of ``^``.

ranisalt:
Thanks!

PR #2079, opened 2017-01-13
Use correct tick interval for bleeding & poison conditions

raymondtfr:


ranisalt:
I couldn't find any reference to this. Where did you see it ticks every 4s?

raymondtfr:
The damage caused by these condition types should be added to the target each 4 seconds, as it is on vanilla.

raymondtfr:
@ranisalt Any update on this?

ranisalt:
It was missing a solid reference, but I found it here http://tibia.wikia.com/wiki/Inflict_Wound

raymondtfr:
I wrote that descriptions 😄 . It was part of a research I made regarding the known *utori-like spells*, which I also used to update them. Though that description is a bit confusing, due to changes some other collaborators made.

Such condtions may have different time period for different monsters, but the majority is 4 seconds.

PR #2080, opened 2017-01-13
new player defense formula based on t-s

raymondtfr:
**New Defense Formula based on tibia-stats research**
* Defense formula now uses [Fist Fighting skill](https://github.com/otland/forgottenserver/pull/2080/files#diff-dc52f162280289dd2d4d97d82658362dR336) *if neither weapon nor shield is equipped*.
* No matter if your [weapon's defense](https://github.com/otland/forgottenserver/pull/2080/files#diff-dc52f162280289dd2d4d97d82658362dR343) is higher than your shield's, instead, your [defense from shield](https://github.com/otland/forgottenserver/pull/2080/files#diff-dc52f162280289dd2d4d97d82658362dR348) will be used.
- No matter which **fight mode** you activated **(Balanced or Offensive)**, **both will act as Defensive** ``(defenseFactor == 1.0f)`` unless you are actively attacking ``((OTSYS_TIME() - lastAttack) < getAttackSpeed())`` a creature.

**Comparison**
- Shielding skill = 100
- Shield defense = 37
- Weapon defense = 31 +3 (extra defense)

| **Defense Factor** | **New Formula** | **Current Formula** | **Diff. range** |
| :-: | :-: | :-: | :-: |
| **Defensive** | 163 | 144 | +19 |
| **Balanced** | 122 | 87 | +35 |
| **Offensive** | 81 | 72 | +9 |

This is a remake of #1581 , and a few others haha.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -360,8 +360,7 @@ int32_t Player::getDefense() const
 		}
 	}
 
-	defenseValue = static_cast<int32_t>(((((defenseSkill / 4.) + 2.23) * defenseValue) * 0.15) * getDefenseFactor());
-	return static_cast<int32_t>(static_cast<float>(defenseValue * vocation->defenseMultiplier));
+	return (defenseSkill / 4. + 2.23) * defenseValue * 0.15 * getDefenseFactor() * vocation->defenseMultiplier;
 }
 
 float Player::getAttackFactor() const
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -333,32 +333,38 @@ void Player::getShieldAndWeapon(const Item*& shield, const Item*& weapon) const
 
 int32_t Player::getDefense() const
 {
-	int32_t baseDefense = 5;
-	int32_t defenseValue = 0;
-	int32_t defenseSkill = 0;
-	int32_t extraDefense = 0;
-	float defenseFactor = getDefenseFactor();
+	int32_t defenseSkill = getSkillLevel(SKILL_FIST);
+	int32_t defenseValue = 7;
 	const Item* weapon;
 	const Item* shield;
 	getShieldAndWeapon(shield, weapon);
 
 	if (weapon) {
-		defenseValue = baseDefense + weapon->getDefense();
-		extraDefense = weapon->getExtraDefense();
+		defenseValue = weapon->getDefense() + weapon->getExtraDefense();
 		defenseSkill = getWeaponSkill(weapon);
 	}
 
-	if (shield && shield->getDefense() >= defenseValue) {
-		defenseValue = baseDefense + shield->getDefense() + extraDefense;
+	if (shield) {
+		defenseValue = weapon != nullptr ? shield->getDefense() + weapon->getExtraDefense() : shield->getDefense();
 		defenseSkill = getSkillLevel(SKILL_SHIELD);
 	}
 
 	if (defenseSkill == 0) {
-		return 0;
+		switch (fightMode) {
+			case FIGHTMODE_ATTACK:
+			case FIGHTMODE_BALANCED:
+				return 1;
+
+			case FIGHTMODE_DEFENSE:
+				return 2;
+
+			default:
+				return 0;
+		}
 	}
 
-	defenseValue = static_cast<int32_t>(defenseValue * vocation->defenseMultiplier);
-	return static_cast<int32_t>(std::ceil((static_cast<float>(defenseSkill * (defenseValue * 0.015)) + (defenseValue * 0.1)) * defenseFactor));
+	defenseValue = static_cast<int32_t>(((((defenseSkill / 4.) + 2.23) * defenseValue) * 0.15) * getDefenseFactor());
+	return static_cast<int32_t>(static_cast<float>(defenseValue * vocation->defenseMultiplier));
 }
 
 float Player::getAttackFactor() const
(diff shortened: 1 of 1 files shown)


src/player.cpp
```diff
@@ -333,32 +333,38 @@ void Player::getShieldAndWeapon(const Item*& shield, const Item*& weapon) const
...
+		switch (fightMode) {
+			case FIGHTMODE_ATTACK:
+			case FIGHTMODE_BALANCED:
+				return 1;
+
+			case FIGHTMODE_DEFENSE:
+				return 2;
+
+			default:
+				return 0;
```
ranisalt: Is it ever possible to have a fight mode different from the previous two? Can you remove this default or does it break?

ranisalt:
This new balanced mode is nice!

src/player.cpp
raymondtfr: Not in the tibia client. I added it because 'cause I thought of *OTClient*, although it is not really needed. I'll remove it.

raymondtfr:
Well, I guess this formula (and others, such as armor reduction, etc...) has been used since Tibia 7.9 or later, although I think tibia-stats researcher Gorak made his research on 2012 if I am not wrong, so it's been a long time.

src/player.cpp
ranisalt: Yes, it's better, if anyone wants to add more fight modes they have enough knowledge to edit the server too :)

src/player.cpp
```diff
@@ -333,32 +333,35 @@ void Player::getShieldAndWeapon(const Item*& shield, const Item*& weapon) const
...
+
+			case FIGHTMODE_DEFENSE:
+				return 2;
+		}
 	}
 
-	defenseValue = static_cast<int32_t>(defenseValue * vocation->defenseMultiplier);
-	return static_cast<int32_t>(std::ceil((static_cast<float>(defenseSkill * (defenseValue * 0.015)) + (defenseValue * 0.1)) * defenseFactor));
+	defenseValue = static_cast<int32_t>(((((defenseSkill / 4.) + 2.23) * defenseValue) * 0.15) * getDefenseFactor());
+	return static_cast<int32_t>(static_cast<float>(defenseValue * vocation->defenseMultiplier));
```
ranisalt: You don't need that much parenthesis around multiplications and you won't need static casts here. Use
```cpp
defenseValue = ((defenseSkill / 4. + 2.23) * defenseValue * 0.15 * getDefenseFactor());
return defenseValue * vocation->defenseMultiplier;
```
As setting `defenseValue`will cast it to int32 anyway and `vocation->defenseMultiplier` is a float and will cast `defenseValue` to float anyway.

raymondtfr: But won't we be diffing from our code base? (not casting)

Throughout the our code base I see we using static_cast to ensure the correct type is returned.

Yep, I usually use that much parenthesis to emphasize precedence, but I'll need to use ``std::floor`` in ``defenseValue = ((defenseSkill / 4. + 2.23) * defenseValue * 0.15 * getDefenseFactor());`` to ensure it has no decimal digits.

ranisalt: There are some moments when upcasting (smaller to larger int) is needed to ensure it doesn't overflow. When you divide an integer (or multiply by something <1), it will not overflow. It also probably doesn't overflow by adding small numbers such as 2.23 here, and certainly doesn't overflow after dividing by 4.

Another need for cast is when you want floating divising between two ints, where one of them must be cast to float, but you are dividing by a float here.

You will also downcast where a number comes in a larger format but the result must be shorter or to fast floor a floating point number.

This piece of code suffers none of this: ints are being automatically casted to float where needed, and there is no multiplication/sum that will ever overflow.

Also, as the return type is `int32_t`, the return is automatically casted, so `return static_cast<...>(...);` is redundant.

raymondtfr: As this formula does not need that much precision, should I use ``4.f`` instead of ``4.`` and ``0.15f`` instead of ``0.15`` then?

By the way, I realized that there is no need to use ``std::floor`` as it will be implicitly cast to ``int32_t`` as you said (?) anyway, so I will remove it.

ranisalt: No, it shouldn't make that of a difference to use float or double. I just measured it and there is no performance difference.

You can definitely remove the outermost pair of parenthesis as it's a multiplication anyway.

ranisalt: I'd even use `(defenseSkill + 9) * defenseValue * 0.0375 * getDefenseFactor() * vocation->defenseMultiplier;` that is shorter and yields almost the same results (less than 1% higher). I multiplied 2.23 by 4 and moved the division outside the parenthesis, then merging with 0.15

raymondtfr: But that won't be the same formula.

kornholi: I agree with @ranisalt. You can even change it to `(0.3333 + 0.0375 * defenseSkill) * defenseValue * getDefenseFactor() * vocation->defenseMultiplier` which looks a lot more natural.

ranisalt: @raymondtfr the point here is the original formula is not full of magic numbers like that, it's just standard deviation, but as T-S editors are probably not statistics professionals, they don't try to smooth functions 🌵

ranisalt:
Thanks!

peonso:
https://github.com/otland/forgottenserver/blob/master/src/creature.cpp#L802

should be `damage -= uniform_random(0, defense);`

PR #2081, opened 2017-01-13
Moved spell Levitate to Lua

raymondtfr:


WibbenZ:
Already started to work on it and you don't have to use moveCreature, you can do it with teleportTo, if im not mistaken the only return value is no error or not posible anyways.

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,40 @@
+function onCastSpell(creature, variant)
+	local currentPos = creature:getPosition()
+	local destPos = creature:getPosition()
+	destPos:getNextPosition(creature:getDirection())
+
+	local ret = RETURNVALUE_NOTPOSSIBLE
```
WibbenZ: Use returnValue insted of ret

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,40 @@
+function onCastSpell(creature, variant)
+	local currentPos = creature:getPosition()
```
WibbenZ: Use currentPosition insted of pos, same with destPos

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,40 @@
...
+	local destPos = creature:getPosition()
+	destPos:getNextPosition(creature:getDirection())
+
+	local ret = RETURNVALUE_NOTPOSSIBLE
+	local param = variant:getString():lower()
+	if param == "up" then
+		if currentPos.z ~= 8 then
+			local tile = Tile(currentPos.x, currentPos.y, currentPos.z - 1)
+			if tile == nil or (tile:getGround() == nil and not tile:hasFlag(TILESTATE_IMMOVABLEBLOCKSOLID)) then
```
WibbenZ: if not tile etc

vankk: Not sure if the check should be like what @WibbenZ told. I saw a few scripts in the datapack that use == nil.

raymondtfr: Yep, but they are not consistent throughout all datapack files, so it is probably ok...? @ranisalt @marksamman .

ranisalt: We should try to achieve consistence. I vow for `not var` instead of `var == nil`.

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,36 @@
+function onCastSpell(creature, variant)
+	local currentPosition = creature:getPosition()
+	local destinationPosition = creature:getPosition()
+	destinationPosition:getNextPosition(creature:getDirection())
```
ranisalt: `getNextPosition` returns the next position instead of changing in place. `local destination = current:getNextPosition(creature:getDirection())` instead.

raymondtfr: But the function ``Position:getNextPosition`` modifies the caller's position ``self`` fields, instead of returning them. So I do what you did then ``currentPosition`` will be modified and instead will hold the next position.

We could prevent that if:

```lua
function Position:getNextPosition(direction, steps)
	local offset = Position.directionOffset[direction]
	if offset then
		steps = steps or 1
		self.x = self.x + offset.x * steps
		self.y = self.y + offset.y * steps
	end
end
```

should be modified to return them, instead of modifying **self** fields.

```lua
function Position:getNextPosition(direction, steps)
	local offset = Position.directionOffset[direction]
	return offset and Position(self.x + offset.x * steps or 1, self.y + offset.y * steps, self.z)
end
```

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,36 @@
...
+
+	local returnValue = RETURNVALUE_NOTPOSSIBLE
+	local parameter = variant:getString():lower()
```
raymondtfr: I don't there is a need to keep consistency between talkactions and spells, in any case, it would be even better if we renamed talkaction's parameter ``param`` to ``parameter``.

data/spells/scripts/support/levitate.lua
WibbenZ: I totally agree with that @raymondtfr, this function has made me scratch my head alot of times...
But it does not matter in this PR, if you wanna get that change done before this PR is done then do that, otherwise do it "the right way" as I told you above, you change the variable by setting nextPosition?.z to another value.

But IMO we should update that function aswell as all the scirpts that uses that function & inform @ninjalulz so he can update orts.

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,36 @@
...
+	if parameter == "up" and currentPosition.z ~= 8 then
+		local tile = Tile(currentPosition.x, currentPosition.y, currentPosition.z - 1)
```
raymondtfr: I don't understand xD.

WibbenZ: That is what I ment :p

data/spells/scripts/support/levitate.lua
ranisalt: Wait what. It's totally wrong, it should return the next position.

I'll merge this PR and later think about fixing that.

raymondtfr:
Rebased against master.

vankk:
@ranisalt This PR isn't ready to be merged? Only needs be rebased..

data/spells/scripts/support/levitate.lua
ranisalt: @WibbenZ is it ok to you now?

ranisalt:
There's still a question by @WibbenZ even though he approved it.

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,36 @@
...
+			if tile and tile:getGround() and not tile:hasFlag(bit.bor(TILESTATE_IMMOVABLEBLOCKSOLID, TILESTATE_FLOORCHANGE)) then
+				returnValue = Game.moveCreature(creature, tile, bit.bor(FLAG_IGNOREBLOCKITEM, FLAG_IGNOREBLOCKCREATURE))
+			end
+		end
+	elseif parameter == "down" and currentPosition.z ~= 7 then
+		local tile = Tile(destinationPosition)
+		if not tile or not tile:getGround() and not tile:hasFlag(TILESTATE_BLOCKSOLID) then
+			tile = Tile(destinationPosition.x, destinationPosition.y, destinationPosition.z + 1)
+
+			if tile and tile:getGround() and not tile:hasFlag(bit.bor(TILESTATE_IMMOVABLEBLOCKSOLID, TILESTATE_FLOORCHANGE)) then
```
ranisalt: This if is duplicated from above and can be extracted.

WibbenZ:
Yes seemw fine to me, he fixed that bug idk why the comment is still there tho :p

data/spells/scripts/support/levitate.lua
raymondtfr: I'll and rebase and take a final look at this.
Last time I tried removing duplicated code I did something wrong, but I'll try again.

raymondtfr:
@ranisalt Now I'll be trying to get rid of that duplicated code, maybe making a function out of it (and try to give a meaningful name to it, of course)

data/spells/scripts/support/levitate.lua
```diff
@@ -1,30 +1,26 @@
...
-				returnValue = creature:move(nil, tile, bit.bor(FLAG_IGNOREBLOCKITEM, FLAG_IGNOREBLOCKCREATURE))
-			end
-		end
-	elseif parameter == "down" and currentPosition.z ~= 7 then
-		local tile = Tile(destinationPosition)
-		if not tile or not tile:getGround() and not tile:hasFlag(TILESTATE_BLOCKSOLID) then
-			tile = Tile(destinationPosition.x, destinationPosition.y, destinationPosition.z + 1)
+		local flag = parameter == "up" and TILESTATE_IMMOVABLEBLOCKSOLID or TILESTATE_BLOCKSOLID
+		local floorChange = parameter == "up" and -1 or 1
+		local tile = Tile(parameter == "up" and {x = fromPosition.x, y = fromPosition.y, z = fromPosition.z + floorChange} or toPosition)
```
WibbenZ: Use Position(...) insted of {x, y, z}

data/spells/scripts/support/levitate.lua
```diff
@@ -0,0 +1,32 @@
+local function levitate(creature, parameter)
+	local fromPosition = creature:getPosition()
+
+	if parameter == "up" and fromPosition.z ~= 8 or parameter == "down" and fromPosition.z ~= 7 then
```
ranisalt: You can store the direction on a variable to avoid having to query `parameter == "up"` all the time. Not sure how fast Lua is with strings, but it is definitely faster with booleans.

src/luascript.cpp
```diff
@@ -7327,19 +7327,32 @@ int LuaScriptInterface::luaCreatureGetPathTo(lua_State* L)
 
 int LuaScriptInterface::luaCreatureMove(lua_State* L)
 {
+	// creature:move(direction, tile[, flags = 0])
```
ranisalt: There should be 2 signatures, `creature:move(direction)` and `creature:move(tile[, flags=0])` , not a single.

raymondtfr: I made it so that we would be forced to use ``direction`` even if we were targeting the form ``creature:move(tile[, flags = 0])``, so I guess I'll update it to welcome both signatures logic.

Please assist me if I don't do it right xD.

data/spells/scripts/support/levitate.lua
raymondtfr: Isn't it a trivial thing to do? i.e. I didn't even need to create the variable ``floorChange`` to hold the change of floor levels, but I chose to create it for clarity.

In the case your mentioning it does not add up for readability (if I understood you correctly.)

ranisalt: It's not so trivial to compare strings, it is generally bound by the shortest in O(n), can have cache and aligning issues and whatnot, if that's what you meant. I mean, you could have two `local UP = 1` and `local DOWN = -1` "constants" and use them.

```lua
local floorChange = (param == 'up' and UP) or (param == 'down' and DOWN) or 0
if floorChange == 0 then
    return RETURNVALUE_NOTPOSSIBLE
end

local tile = Tile(floorChange == UP and Position(fromPosition.x, fromPosition.y, fromPosition.z + floorChange) or toPosition)
if not tile or not tile:getGround() and not tile:hasFlag(floorChange == UP and TILESTATE_IMMOVABLEBLOCKSOLID or TILESTATE_BLOCKSOLID) then
```
And so on. Of course, if you feel it's not clear enough, it's just a suggestion 😛

raymondtfr: Well, this is the way I refactored the previous code, it is as simple as it could get.
If people are familiar with vanilla, then they'll know what is the use of this spell, function...

The string itself already states the purpose of the spell, either *go up one floor* or *go down one floor*, so I think it is unnecessary to add new constants.

What I could do is to optimize a little, defining some variables only when needed, which would kill readibility a little - which was the main reason I chose to kept these variables, such as ``flag`` and ``floorChange``.

By removing some variables, it could be like this:
```lua
local function levitate(creature, parameter)
	local fromPosition = creature:getPosition()

	if parameter == "up" and fromPosition.z ~= 8 or parameter == "down" and fromPosition.z ~= 7 then
		local toPosition = creature:getPosition()
		toPosition:getNextPosition(creature:getDirection())

		local tile = Tile(parameter == "up" and Position(fromPosition.x, fromPosition.y, fromPosition.z - 1) or toPosition)
		if not tile or not tile:getGround() and not tile:hasFlag(parameter == "up" and TILESTATE_IMMOVABLEBLOCKSOLID or TILESTATE_BLOCKSOLID) then
			tile = Tile(toPosition.x, toPosition.y, toPosition.z + (parameter == "up" and -1 or 1))

			if tile and tile:getGround() and not tile:hasFlag(bit.bor(TILESTATE_IMMOVABLEBLOCKSOLID, TILESTATE_FLOORCHANGE)) then
				return creature:move(tile, bit.bor(FLAG_IGNOREBLOCKITEM, FLAG_IGNOREBLOCKCREATURE))
			end
		end
	end
	return RETURNVALUE_NOTPOSSIBLE
end

	local returnValue = levitate(creature, variant:getString())
	if returnValue ~= RETURNVALUE_NOERROR then
		creature:sendCancelMessage(returnValue)
		return false
	end

	creature:getPosition():sendMagicEffect(CONST_ME_TELEPORT)
	return true
end

```

PR #2085, opened 2017-01-14
Remove compat function from swimming.lua

WibbenZ:
doSetCreatureOutfit is not a compat function, no need to use it in the
datapack.


ranisalt:
I'm pretty sure you don't need to specify all attributes, only `lookType`, can you check?

WibbenZ:
No need to check it, if we don't modify them we will still use our "old" colors.

ranisalt:
But the swimming looktype has no color settings, it just won't apply

WibbenZ:
I was wrong, it worked as you said.
Updated the PR.

ranisalt:
TOLD YA

Thanks!

PR #2086, opened 2017-01-15
Add Game.getVersion (table)

### LINKED ISSUE #1584: Ability to get the client version from a Lua method, new "drop the two-handed"
Well even if I atleast think the activity has been low here (hopefully this changes soon :p) 

But since Ninja has released this; https://otland.net/threads/8-60-the-forgotten-server-1-2.236489/
We should some sort of function to check if it's the 8.6 or latest version, this way we can do as I have spoken about before; one script for all client version.

The reason this function is needed is mainly because of things like mounts etc that 8.6 does not utilize.
We can either return a boolean function or even better an integer that gives us the version, this way we can check if the version is 10.70 or 10.90 etc

Another thing is the "new" system where a certain game drops the shield or w/e and insted allows the move of a two-handed weapon, this way you save time insted of first having do drop the shield, then move the new weapon to replace your old weapon.

Id like the community input before I or someone else implements these things.

WibbenZ:
Returns a table containing:
minVersion
maxVersion
aswell as the string value, if you some reason need that.

Semi fixes #1584


ranisalt:
I'm not sure it makes sense to maintain a function only useful for niche modifications of the source. I mean, request a pull to ninja's repository if it's useful there, not here.

Anyway, it should be named `getClientVersion` as it's not the game version.

WibbenZ:
But if I open a PR there then ppl will get errors when runnings scripts on this version.
The point of this function is to be able to create scripts like annihilator where they are the same on all versions, but insted of having 100 diffrent scripts we can change certain things like item ids, rewards etc with this function.

It can also be used for ex getting the client version if you want that in your motd or w/e so it's still gonna be useful for those using the latest version.
It's a useful function if you ask me.

But getClientVersion to me sounds like the only version we allow, this way we can get the min, max aswell as the string / message from that function while it still makes sence.

ranisalt:
Looks good to me. It's just this repo, as a reference implementation, should have usage for all Lua functions, so if you can provide a working example (e.g. a /version talkaction) it would be more useful.

WibbenZ:
Well anything can be added to this tbh, the good thing as I statet is that we can use the exact same script on all platforms, no matter the version.
It might not be 100% needed but will make it easier for people to keep 8-6 or w/e while staying updated on the latest PRs.

But at the same time im not saying this function is needed to accept PR, it's more a "good thing if you do".

ranisalt:
Ok, just rename to `getClientVersion` then.

ranisalt:
Hmmm just noticed that you can use `player:getClient()` to check a player client version. It works exactly the same, sending `CLIENT_VERSION_MIN`. Isn't that enough?

WibbenZ:
Yes you could use that, but that would make it hard if you ex change a globalevent or something where we don't have a player userdata.

It could ofc work for actions etc, but then we only get the version that the player is using, seems smarter to be able to check all versions, take shadowcores as an example, they had like 10.10 to 10.40.

So it does not send the min version, it sends the version that the player uses.

ranisalt:
Oh, nice. It could be the start of some larger version range :)
Thanks!

PR #2087, opened 2017-01-15
Moved conjure spells to Lua

raymondtfr:
All magic effects were researched.


ranisalt:
Next step is KILL THEM RUNES! Ah, since only players can (or at least make sense) conjure stuff, you can rename the parameter to `player` and add the method to the `Player` metatable instead.

You should move to `scripts/conjure`, as other folders use infinitive verbs.

raymondtfr:
To keep consistency I think it is better to not rename ``onCastSpell`` parameters, users should have enough information to know about this :eyes:

Kamenuvol:
conjure is better than conjuring for the folder's name

raymondtfr:
@Kamenuvol It should be the past participle (or gerund), so it will be a noun, the exceptions are: attack, party, house, etc... where they are grammatically already a noun.

So conjure -> conjuring, heal -> healing, etc.

Btw, this is gonna take longer than I expected, I'm organizing spells.xml (it'll be much more clearer to read then) and also renaming spells name (for updated ones), removing unnecessary tags (i.e. the tag ``aggressive`` by default is already ``true``, so we could remove all of them from attack spells (not breaking-code-proof, will just provide better readibility) or use them in all attack spells (this way is breaking-code-changes proof)), etc..

data/spells/lib/spells.lua
```diff
@@ -348,3 +348,36 @@ function Creature:addPartyCondition(combat, variant, condition, baseMana)
...
+		self:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
+		self:getPosition():sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	if item:hasAttribute(ITEM_ATTRIBUTE_DURATION) then
+		item:decay()
+	end
+
+	self:getPosition():sendMagicEffect(effect)
```
ranisalt: Suggestion: use `effect or CONST_ME_MAGIC_GREEN` and you will have a default value for effect you can omit on all files. It seems all conjuring has this same effect.

raymondtfr: I still have not researched any effect, but I expect to soon, so only then I'll have a say if we would really need a default, but in any case I can add it if the same effect is being used by the majority of the scripts (after the research).

I'm still updating spells.xml, soon I'll mark this as concluded xD.

ranisalt:
IMHO the defaults should be the most common options, i.e. convention over configuration. The majority of spells is aggressive, so by default aggressive is true; the majority of spells do not take a parameter, so by default it expects 0 parameters; etc.

raymondtfr:
So I guess I'll be removing the tag ``prem="0"`` from spells that do not require premium (based on vanilla), since its default value is ``false``.

I'll also be removing the tag ``aggressive="1"`` since its default value is already set to ``true``. I also tried removing this very same xml tag, but set to 0 ``aggressive="0"`` from support, healing spells, but it overwrites those from the scripts, so I had to keep them.

I'm also thinking about removing all occurrences of the tag ``needlearn="0"`` since its default value is also ``false``.

ranisalt:
Just make sure those are the defaults. Even though most spells are premium-only and all are needed to learn on Tibia, nobody enforces this restriction on OT servers.

Kamenuvol:
IMO learnable default should be false and add to config.lua an option to make all not needing to learn
OR make default directly in Lua
spellLearnableDefault = true/false

raymondtfr:
The only thing I have to do now is to resolve the conflicts, and update all conjuring spell effects that I can (I'm not pacc on vanilla), if I cannot then I'll add some form of a comment to that spell (so I can update them later on an upcoming Testserver) and remove [WIP] tag from the title, or maybe @ninjalulz or someone else could.

raymondtfr:
It seems to me that all conjuring runes spells probably have the same effect ``CONST_ME_MAGIC_RED``, but I am not sure.

I also think the effect for conjuring a Enchanted Spear is right (green shimmer), and the others ammo (blue shimmer) and less sure about Enchant Staff.

conde2:
Why not have parameters for rune in spells.xml ?
Like we have for exiva, we could have itemid and count as parameter, so we don't need to have multiple files doing the same stuff

raymondtfr:
@ranisalt  This is ready for merge I guess.

I'll try and update effects (if there's any incorrect) in the next testserver.

WibbenZ:
@raymondtfr Are you sure that there are diffrent effects for diffrent runes?
If you want I can check the other ones when I get time, but I can check all on rp, ms but only up to like 30? on eds.

raymondtfr:
I think the effect is the same for all conjuring runes, but nonetheless I tested all of them I could (except for premium ones), and they were all red shimmer.

You can check above in the list which effects weren't researched (``effect: ?``).

WibbenZ:
Did a couple atleast, didnt feel like running around with a low lvl so someone else has to do the eds ones
```
 Disintegrate Rune effect: red shim
 Energy Bomb Rune effect: red shim
 Fireball Rune effect: red shim
 Great Fireball Rune effect: red shim
 Lightest Magic Missile Rune effect: red shim
 Magic Wall Rune effect: red shim
 Soulfire Rune effect: red shim
 Thunderstorm Rune effect: red shim
 
 
 Conjure Bolt effect: blue shim
 Conjure Piercing Bolt effect: blue shim
 Conjure Power Bolt effect: blue shim
 Conjure Sniper Arrow effect: blue shim
 Enchant Spear effect: green shim
 Enchant Staff effect: green shim
```

joseluis2g:
`Icicle Rune: red shim
Paralyze Rune: red shim
Wild Growth Rune: red shim
Stone Shower Rune : red shim`

raymondtfr:
WibbenZ you meant **Light Magic Missile Rune** (instead of **Lightest**), right? Because I can not find a way to learn that spell in vanilla 😕 (Practise Magic Missile Rune that conjurs a Lightest Magic Missile Rune).

Thanks, I'll update the last and make sure they match the ones you guys have researched.

WibbenZ:
Ye that text is copied from you msg, but yes one was wrong forgot to write that ;)

raymondtfr:
Removed some conjuring spells from the list, which is probably not possible to learn them in vanilla (light stone shower/lightest missile spells, and practise magic missile - which conjures a lightest magic missile rune.)

So now we have only 3 runes with effects to be researched: **Animate Dead/Holy Missile/Poison Bomb runes** - and I am pretty sure they are all the same, so I'll just update the code with the piece of code in my most recent comment.

Kamenuvol:
i think it is a kind of "unbalanced" only runes have default effect, but it would be too specific make default for ammunitions for example

raymondtfr:
I don't think you understood the purpose of a default effect for runes.

The purpose of it is that the effect for all spells that conjures a rune is *probably* the same ``CONST_ME_MAGIC_RED``, so I added the commit ``add default effect for runes`` so that we can omit that parameter for conjuring spells that conjures runes... other than that, these effects were researched by me and other contributors as you can see in the comments above, so it is not *unbalanced*.

raymondtfr:
Rebased against master

@ranisalt @djarek @DSpeichert

raymondtfr:
Well, I thought *conjuring* was the right one... should I change it to *conjuration* then?

djarek:
Seems they are [synonyms](http://www.thefreedictionary.com/conjuration). Doesn't matter really, you can leave the name as is.

ranisalt:
Well, the movie name is The Conjuring so I guess it's more common

ranisalt:
Months on and there's no merge conflict on a 400+ line diff commit. Seems our codebase is getting less entangled 👍 

Thanks, @raymondtfr!

djarek:
The development is just getting slow :D.

PR #2089, opened 2017-01-15
Move illusion to Lua

WibbenZ:
Tested and works just like it did when it was in C++
Did not remove any of the functions from the Spell class since they are
still in use in other codes.

### DIFF
--- a/data/spells/scripts/support/illusion.lua
+++ b/data/spells/scripts/support/illusion.lua
@@ -1,0 +1,33 @@
+local condition = Condition(CONDITION_OUTFIT)
+condition:setTicks(180000)
+
+function onCastSpell(creature, variant)
+	if not creature:isPlayer() then
+		return false
+	end
+
+	local returnValue = RETURNVALUE_NOERROR
+
+	local monsterType = MonsterType(variant:getString())
+	if not monsterType then
+		returnValue = RETURNVALUE_CREATUREDOESNOTEXIST
+	end
+
+	if returnValue == RETURNVALUE_NOERROR then
+		if not getPlayerFlagValue(creature, PlayerFlag_CanIllusionAll) or not monsterType:isIllusionable() then
+			returnValue = RETURNVALUE_NOTPOSSIBLE
+		end
+	end
+
+	if returnValue ~= RETURNVALUE_NOERROR then
+		creature:sendCancelMessage(returnValue)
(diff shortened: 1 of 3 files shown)


data/spells/scripts/support/illusion.lua
```diff
@@ -0,0 +1,33 @@
...
+
+	local returnValue = RETURNVALUE_NOERROR
+
+	local monsterType = MonsterType(variant:getString())
+	if not monsterType then
+		returnValue = RETURNVALUE_CREATUREDOESNOTEXIST
+	end
+
+	if returnValue == RETURNVALUE_NOERROR then
+		if not getPlayerFlagValue(creature, PlayerFlag_CanIllusionAll) or not monsterType:isIllusionable() then
```
TheSumm: I think this could be added as an "else if" branch in line 14. As a consequence, the check for returnValue == RETURNVALUE_NOERROR is not necessary.

data/spells/scripts/support/illusion.lua
```diff
@@ -0,0 +1,29 @@
...
+	if not creature:isPlayer() then
+		return false
+	end
+
+	local returnValue = RETURNVALUE_NOERROR
+
+	local monsterType = MonsterType(variant:getString())
+	if not monsterType then
+		returnValue = RETURNVALUE_CREATUREDOESNOTEXIST
+	elseif not getPlayerFlagValue(creature, PlayerFlag_CanIllusionAll) or not monsterType:isIllusionable() then
```
ninjalulz: Change **or not** to **and not** ^^

ranisalt:
I'll wait until @ninjalulz and @TheSumm give their approval of changes.

WibbenZ:
@ninjalulz Yes we need the if statment for getPlayerFlagValue

ninjalulz:
Why would that be necessary if I may ask?

WibbenZ:
That function can't be used on monsters etc or am I wrong?

ninjalulz:
It will simply throw an error saying 'Player not found', and return false. This spell is not used by monsters anyway.

ranisalt:
Oh my god appveyor is so utterly slow

WibbenZ:
Yeah haha, 3h seems like a decent compiling time :P

brunominervino:
Maybe appveyor have stopped?

ranisalt:
Thanks!

PR #2092, opened 2017-01-15
Add option to choose DB

aspiraboo:
PT-BR
Olá, com esse commit, se torna possível escolher o DB, no momento testei com 2 DB: MySQL e PostgreSQL.
Acho necessário desenvolver a opção para conexão com o DB via ODBC,  SQLite e MariaDB.

Considerações: É necessário a verificação para a melhora do código.
Esté o meu primeiro commit para o forgottenserver.

ENG
Hi, this commit, is possible choose DB, It was tested with 2 DB: MySQL and PostgreSQL.
I think it's necessary develope to connect DB with ODBC, SQLite and MariaDB.

Considerations: I think it's necessary verification is required for code improvement.
This is my first commit for forgottenserver.


yrpen:
TFS work with MariaDB... I know, I tested master few months ago.
MariaDB is fork of mySQL.

yrpen:
BTW Take a look at Roadmap of TFS 1.0. Developers removed support for SQLite. I guess they had good reason.

https://github.com/otland/forgottenserver/wiki/Changelog-1.0

aspiraboo:
yes, one problem is schema.sql, the lexical schema is not equals in others DB.
delete or change table, do you need to modify all others.

aspiraboo:
why appveyor fail to build?

https://ci.appveyor.com/project/kornholi/forgottenserver/build/1.0.116/job/gsfk8name9erw431

wgrriffel:
@aspiraboo that's because you didn't specify to compiler where is librarys, normaly i'd do something like
`cl -Zi ..\src\otserv.cpp /link path\to\lib.lib`, I don't use visual studio IDE shit, but it must be defined somewhere in /vc14/, peek there.

ranisalt:
MariaDB API is fully compatible with MySQL, it's the philosophy. I use MariaDB since most Linux distros don't ship proprietary software.

SQLite would be a great possibility to fire up a server with less boilerplate and resources than MySQL is.

wgrriffel:
@ranisalt I'm not aware if he did included correctly, but his dependencies inclusion is over 150 mb, this is unviable to include in tfs-sdk, i guess.

raymondtfr:
Although I concur on having SQLite (less resources as @ranisalt pointed out, and maybe easier to setup), but this something already removed from **1.0**, so I don't understand the meaning of this PR, aside from more dependencies.

ranisalt:
The problem is not the extra dependencies but having direct dependencies. This adds a direct dependency on another set of packages which is not viable. Instead, a solution that doesn't depend on a single database server (e.g. ODBC) is much better.

Mkalo:
@djarek I have a question kinda related to this PR but not quite.
Why are we using mysql-connector-c instead of mysql-connector-cpp? I was about to implement prepared statements and make a PR then I saw that we are using the C library that does support prepared statements but its so ugly so I would rather not.

Like: 

https://github.com/hholzgra/connector-c-examples/blob/master/mysql_stmt_bind_result.c
vs
https://github.com/anhstudios/mysql-connector-cpp/blob/master/examples/prepared_statement.cpp

djarek:
@Mkalo 
I'm not sure, most likely historical reasons.

aspiraboo:
Possibility of use:
Name: sqlapi++: shareware - http://www.sqlapi.com - OS: Windows; Linux -
Support BD:
Oracle Database Server; Microsoft SQL Server; Sybase; DB2; Informix; InterBase/Firebird; SQLBase; MySQL; MariaDB; PostgreSQL; ODBC; SQLite; Sybase SQL Anywhere

Name: SOCI: free - http://soci.sourceforge.net/index.html - OS: Windows; Linux -
Support BD:
DB2; Firebird; MySQL; ODBC (generic backend); Oracle; PostgreSQL; SQLite3

Name: OTL: free - http://otl.sourceforge.net - OS: ¿? -
Support BD:
DB2; ODBC; Oracle

Name: SQLPP11: free - https://github.com/rbock/sqlpp11 - OS: Windows; Linux -
Support BD:
MySQL; sqlite3; PostgreSQL; "ODBC: experimental"

Name: CppDB: free - http://cppcms.com/sql/cppdb/index.html - OS: Windows; Linux -
Support BD:
MySQL; PostgreSQL; Sqlite3; ODBC

DSpeichert:
What exactly is the use case for adding support for other database engines?
Are there any features that are useful? Performance considerations?

Because adding such a feature that requires considerable maintenance in code for no apparent reason is not a good design decision. That's why it was removed in the first place, because almost everyone used MySQL/MariaDB anyway.

ranisalt:
IMHO, as I already stated before, SQLite is much superior than MySQL for prototyping, as its a piece of cake to setup (just point to the file) and resetting a database is as simple as copying a clean sample over. This makes it much simpler to test the server.

Other than that I don't see a point as database servers are pretty much the same for that TFS needs.

DSpeichert:
There is some differences between sqlite and MySQL, e.g. triggers (don't exist in sqlite). It may be useful for prototyping but simply making it possible to use sqlite means people will use it in production just because it's easier. This leads to the necessity for AAC scripts to adapt and also forget about triggers (if any use them) and of course have lower performance. Besides, concurrent use of sqlite database....

DSpeichert:
As discussed above, there is really no advantage to adding PostgreSQL support, especially as a direct dependency.
SQLite (even with its limitations as explained above) would make first-time use of forgottenserver easier. PostgreSQL is so similar to MySQL that it does not bring in any extra value.
In terms of this PR, the preferred way to add such capability (SQLite and maybe PostgreSQL) would be via DBAL.

ranisalt:
Hey, it's 2018, we have Docker for setting up databases cleanly and, for the most part and for what we use, PostgreSQL is compatible with the queries too.

PR #2093, opened 2017-01-16
More updates on Party spells

raymondtfr:
* Updated spells lib function to use the ``RETURNVALUE`` equivalent of ``"No party members in
range"`` in ``sendCancelMessage``.
* Switched function metatable from Creature to Player.
* Updated spell to use correct area (AREA_CIRCLE3X3).


ranisalt:
Thanks!

Issue #950, opened 2014-08-14
[QUESTION] Offline training rate


edumntg:
Is offline training rate multiplied by rateSkill? I readed the code and i didn't found something about the rate being multiplied by rateSkill.

PR #2095, opened 2017-01-16
Move isInArray to Lua

WibbenZ:


raymondtfr:
I think isInArray has been in C++ this whole time probably due to performance concerns, so I don't think it would be a good idea moving it to Lua.

Also as we can see in your code, implementing it its pretty straightforward, so ``isInArray`` must be in C++ for a reason.

WibbenZ:
Yeah you might be right @marksamman

ninjalulz:
I ran a quick test of 10 million iterations through the table _questDoors_ (defined in global.lua) looking for itemId 3551, and the C++ implementation were actually slower.

**Code**
```Lua
local itemId = 3551
for i = 1, 10000000 do
    if isInArray(questDoors, itemId) then
        --
    end
end
```
**C++**
```Lua
2.22s
2.39s
2.22s
2.30s
2.23s

avg: 2.272s
```
**Lua**
```Lua
1.24s
1.31s
1.25s
1.25s
1.24s

avg: 1.258s
```

ranisalt:
As I said previously, any time you switch between the interpreter and C++ code, it has a performance impact, so that's probably where you're getting the hit. Another possibility is that LuaJIT (if you're using) is noticing the table is not being modified in the inner loop and hoisting the condition outside e.g. JITed code is:
```
if isInArray(...) then
    for i = ... do
        --
    end
end
```

That said, performance is not a concern here, as long as complexity maintains and the code is pretty simple, only basic operations and `pairs`.

WibbenZ:
We have always used isInArray so I didn't think about using table.contains (which a lot of people use in other apps).
And if we do change it we will have to create a compat function / rename for it.

IMO we should keep isInArray, 99% of the people know about it and have no problems using it.

raymondtfr:
Just add it to table metatable as ranisalt suggested (``table:find(value)``, ``table:contains`` or ``table:has`` ) and add it to compat  in compat ❓

WibbenZ:
Sure ill do that

@raymondtfr What is the error you are getting?

raymondtfr:
```lua
Lua Script Error: [Action Interface]
data/actions/scripts/other/spellbook.lua:onUse
data/actions/scripts/other/spellbook.lua:7: attempt to index local 'spell' (a boolean value)
stack traceback:
        [C]: in function '__index'
        data/actions/scripts/other/spellbook.lua:7: in function <data/actions/scripts/other/spellbook.lua:1>
```

WibbenZ:
Are you sure you update your files?
I can't reproduce it, but I had to download from nightlies since my VS is acting up

raymondtfr:
I have just unpacked clean datakpack files (downloaded now) to a different folder and downloaded nightlies ``forgottenserver-31672c24d832f75bc2b67d4e9a103b56a2d39c2f-b951-Win32-Release`` and it happened too.

WibbenZ:
Humm, im using forgottenserver-03483aacbddec173517601f5482f64d15946df14-b950-Win32-Release
So #950
Gonna check it again

WibbenZ:
flash

raymondtfr:
The heck :confused: . I am using the item spellbook (id 2175) btw, anyway I'm gonna take a nap, it must be me brains lol.

WibbenZ:
Yeah thats pretty strange, let's see if someone else reports it :P

Riverlance:
Here is okay, I think:

image


Tested on latest TFS (pulled it right now and recompiled).

brunominervino:
@Riverlance can you try one of these changes and confirm us if you got same error?
http://pastebin.com/BJCGn9Jt

WibbenZ:
@brunominervino The code we are talking about has already been fixed in a PR from @ranisalt , it was not related to this PR.
This code works just fine.

brunominervino:
@WibbenZ Sorry, I haven't understood. :) Ignore my comment hehe

ranisalt:
Should be even faster now without the type check :)

ranisalt:
Thanks!

PR #2096, opened 2017-01-16
More ways to refer to spells.xml tags

raymondtfr:
This'll provide a much better readibility and organization of spells.xml.

### DIFF
--- a/src/spells.cpp
+++ b/src/spells.cpp
@@ -453,11 +453,11 @@ bool Spell::configureSpell(const pugi::xml_node& node)
 		secondaryGroupCooldown = pugi::cast<uint32_t>(attr.value());
 	}
 
-	if ((attr = node.attribute("lvl"))) {
+	if ((attr = node.attribute("level")) || (attr = node.attribute("lvl"))) {
 		level = pugi::cast<uint32_t>(attr.value());
 	}
 
-	if ((attr = node.attribute("maglv"))) {
+	if ((attr = node.attribute("magiclevel")) || (attr = node.attribute("maglv"))) {
 		magLevel = pugi::cast<uint32_t>(attr.value());
 	}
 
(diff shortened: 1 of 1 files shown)


ranisalt:
Add new options later in the check since they will be less common for a while.

raymondtfr:
Including ``cooldown``?

By the way, do you think it is worth to add a third option for it, such as ``spellcooldown``? because we have ``group`` then it has ``groupcooldown``, so similarly we have ``spellid`` then would make sense to have ``spellcooldown``.

By the way, I have been giving some thought about making spells.xml better. Do you think instead of using start/end tags ``instant`` or ``rune`` we could go ahead and just use ``spells -> spell``? In a way, I think this would remove a lot of code that would be required to load ``rune`` and ``instant`` so we could go with uniformity and just use ``spells -> spell``, and instead add a new tag named ``type`` which the attribute could either be ``rune`` or ``instant`` so that we could identify which it is. What you think?

Not that I am suggesting I would do that haha, I haven't enough experience.

Kamenuvol:
instant spells are talkactions
runes (items) are actions
i think they must stay separated

ranisalt:
Agree with @Kamenuvol, they are essentially different and it's just another way of doing the same with more characters.

`spellcooldown` is not useful since every attribute refers to the spell. `cooldown` is the spell's cooldown and `groupcooldown` is the spell's group cooldown.

Kamenuvol:
About spells organization, I think we should add spells in vocation list, instead of vocations in spell list but will make a giant incompatibility. But, it is a idea for 2.0

raymondtfr:
I moved these new options to the beginning of the ``if``s conditions, since this PR been here for a long time already.
@ranisalt @djarek

ranisalt:
Thx

PR #2097, opened 2017-01-16
Fix /reload not returning the correct text message

WibbenZ:


kornholi:
uh, why are you calling `std::terminate` in the first place?

WibbenZ:
@djarek Told me to use it since if spells fail monsters will fail ~
https://github.com/otland/forgottenserver/pull/1520

wgrriffel:
@WibbenZ did you checked these supposed dangling pointers anyhow? maybe this check is a waste of time.

ranisalt:
@wgrriffel it will happen if you e.g. change spells.xml by removing something and reload only spells, not monsters. A monster has a vector of `spellBlock_t` which itself has a pointer to a spell that, if removed, becomes dangling as the spell will be deleted.

wgrriffel:
@ranisalt well noticed, since i assume this is not a concern of this pr, instead just calling std::terminate @WibbenZ i think you should make an exception handler throw it with an error message, by this way user at least knows why server was termined.

WibbenZ:
@wgrriffel That should not be needed since we can still see the console.
flash

WibbenZ:
Seeing as we haven't gotten any other issues about this;
1. Does no one except the devs use the latest version?
2. Is no one reporting errors anymore?

Mkalo:
@WibbenZ I'm still getting the errors when compiling, I don't know why this didn't get merged yet.

WibbenZ:
Probbly because the owners don't have time, but im surprised that we don't have any new issues about it... xD

ranisalt:
I don't see where this is going to get. Resumedly, you make the server crash & exit whenever a script fails to load, but this is not good behavior. When e.g. a monster fails to load it's script, it should just reset scripting status, warn on console and go ahead.

Additionally, I don't feel good about changing code because a compiler doesn't know how to compile, if that's VS-exclusive.

Mkalo:
@ranisalt But he is just removing/adding returns, this is not a big change to fix a vs-exclusive compile warning.

You are all talking about std::terminate but the issue is not about that. If you disagree with the std::terminate you can open a new issue and then a new pull request. This is to fix unreacheable code warning only.

WibbenZ:
The crash and exit is what @djarek told me to use (there is a link to PR above).
And I agree with him, if spells fails monsters should fail aswell.

And to be fair this "issue" was a bug created by me, if we break a switch we have nothing to return, the code won't continue.
If linux dosn't care about it or VS does has nothing to do with it, the code is useless.

src/baseevents.cpp
```diff
@@ -74,6 +74,7 @@ bool BaseEvents::loadFromXml()
 
 		if (!success || !registerEvent(event, node)) {
 			delete event;
+			return false;
```
Mkalo: I don't see the reason of this return as it will stop loading in half because an error, thats bad imo, or you load it all and report errors or you load nothing and terminate, the first is as it always was so I wouldn't change that.

src/monsters.cpp
```diff
@@ -198,6 +198,7 @@ bool Monsters::loadFromXml(bool reloading /*= false*/)
 			} else {
 				std::cout << "[Warning - Monsters::loadMonster] Can not load script: " << scriptEntry.second << std::endl;
 				std::cout << scriptInterface->getLastLuaError() << std::endl;
+				return false;
```
Mkalo: Same as above.

src/baseevents.cpp
WibbenZ: The point is that the commit needs a boolean value return that indicates whether the reload was successful or not, if the reload fails it should return false not true (aka it shouldn't say everything is OK when it isen't).
And the same with this issue, the old system did not require a boolean value for a return value, but this one does to check if monsters / spells should be loaded aswell.

This is a thing I forgot when I created the pr and not checking every return :P

Mkalo: Well yea but this is not changing only for spells, I believe more classes use this method, I could be wrong tho.

WibbenZ: Yes but it dosn't seem to break anything, it would always return true before this change.

wgrriffel:
@ranisalt 
> Additionally, I don't feel good about changing code because a compiler doesn't know how to compile, if that's VS-exclusive.
VS is the only opition provided for windows users, you should care about if it knows how to compile, btw it knows how to compile, it only warns you that there is unreachable code, this pr doesn't fix an error but a wanrning.

ranisalt:
Windows users can use CMake + MinGW, it's just not on the wiki.

I meant that this PR changes the behavior of the code, like what was pointed that some events will stop loading at all on the first error.

wgrriffel:
@ranisalt yes they use cmake if they know cmake, cmake build files in this repository does NOT support windows file system, it never would find the dependencies to generate any build file.

WibbenZ:
@ranisalt 
Isn't it correct that when we use std::terminate no more code after that is executed (in this case return false?)
Why would otherwise VS throw a warning about it?

ranisalt:
Not exactly, it's just the default action. You can install a different `std::terminate` behavior, but we don't do it, so you can assume it aborts the execution immediately.

src/baseevents.cpp
ranisalt: Wait there, failing here just means the spell could not be loaded, not that loading the XML list failed. There's a difference between loading the XML and loading the Lua script.

WibbenZ:
So in this case this should be merged since linux should not have any problems with it and VS is happy?

Mkalo:
I think a better solution than std::terminate if spells reload fail is create a new class for monster spells and that wont be reloadable only when the monsters reload they would be reloaded, that way if a monster spell fail to load we just don't load that monster instead of closing.

WibbenZ:
Feel free to improve it, I don't have the time for that right now, got enough things going on atm :P 
The only thing I wanted out off this PR was to remove the warning, but you do have a fair point that we shouldn't terminate if it's posible to avoid it, maybe by validating the reload, if it fails skip that monster and use the old code and then just send an error.

That way a server can continue even with a "bugged script", allowing users to pretty much code on a server with 100+ players insted of using a test server :P 
Atleast with monsters and spells.

ranisalt:
A monster could have a weak pointer to a spell so when a spell ceases to exist you should just check the pointer hasn't expired. That is not related to this PR.

Mkalo:
@ranisalt I think this solution is overkill, we would basically need to change everything about spells to use shared pointers just for this reason.

src/baseevents.cpp
WibbenZ: So @ranisalt @djarek what should we do?
As of right now they don't return anything = you will think it reloaded without errors.

Should I add a boolean that will be returned as Makalo said?

ranisalt:
Thanks!

PR #2098, opened 2017-01-16
Fix spellbooks etc not loading

WibbenZ:


src/spells.cpp
```diff
@@ -236,9 +236,12 @@ uint32_t Spells::getInstantSpellCount(const Player* player) const
...
 {
-	auto it = std::next(instants.begin(), std::min<uint32_t>(spellId, instants.size()));
-	if (it != instants.end()) {
-		return it->second;
+	uint32_t count = 0;
+	for (const auto& it : instants) {
+		if (count == spellId) {
+			return it.second;
+		}
+		++count;
```
ranisalt: This is a revert, is it really an issue?

WibbenZ: Yes that was the thing causing the issue what I could find (except the nil check in the Lua script)
If you can find a way to do it in a good way and a working way I can revise the PR.
Otherwise this is the way to go, so we atleast have a stable master for now.

ranisalt:
I don't think this is the way to tackle this issue. I'm coding a more robust solution and will PR later.

PR #2100, opened 2017-01-18
Datapack consistent coding style

wgrriffel:


data/npc/lib/npcsystem/modules.lua
```diff
@@ -700,7 +700,7 @@ if Modules == nil then
 				i = i + 1
 			end
 
-			if name ~= nil and container ~= nil and itemid ~= nil and cost ~= nil then
+			if name and container and itemid and cost then
```
TheSumm: Not related to your changes, but I think "container" is never set. It might be a bug in the npc system (itemid is set in 2 cases).

data/npc/lib/npcsystem/npcsystem.lua
```diff
@@ -51,7 +51,7 @@ if NpcSystem == nil then
 	-- Gets an npcparameter with the specified key. Returns nil if no such parameter is found.
 	function NpcSystem.getParameter(key)
 		local ret = getNpcParameter(tostring(key))
-		if (type(ret) == 'number' and ret == 0) or ret == nil then
+		if (type(ret) == 'number' and ret == 0) or not ret then
```
TheSumm: `ret == nil` is redundant. It can be dropped completely since `return ret` will still return nil in the case that `ret` is `nil`.

From what I noticed `getNpcParameter` returns either a string or nil, so `type(ret)` will never be `'number'`.

data/npc/lib/npcsystem/modules.lua
wgrriffel: @TheSumm https://github.com/otland/forgottenserver/blob/master/data/npc/lib/npcsystem/modules.lua#L678

TheSumm: Yes, it is never changed afterwards.

wgrriffel: @TheSumm oh, i see, actually this code is legacy, i can't have any idea what was/is it purposes, maybe some really old tfs version had a usage? this npc system needs to be completely reworked, i'd say 1.3 roadmap, i'll left as is, the purpose os this pr is make code looks more consistent in datapack not fix anything, anyway well noticed.

PR #2102, opened 2017-01-18
Move rune functions to Lua scripts

ranisalt:
Moved the remaining two rune functions to Lua scripts to unbloat the engine. Runes now don't have any predefined functions so I also removed the loading function, which now only returns `false` from base class `Event`.

Tested both but I won't push to master so someone can review before I merge.


data/spells/scripts/support/chameleon.lua
```diff
@@ -0,0 +1,16 @@
+local condition = Condition(CONDITION_OUTFIT)
+condition:setTicks(200000)
+
+function onCastSpell(creature, variant, isHotkey)
+	local item = Tile(variant:getPosition()):getTopDownItem()
```
ninjalulz: I wouldn't recommend that. This will throw an error (_attempt to use a nil value_) unless you place the item directly on the ground.

wgrriffel:
@ranisalt I've noticed a ton of changes to Lua API and nothing pushed to wiki, assuming you don't access to push changes there, are you or anyone else registering these changes? Lua API is badly documented but at least had all metatables/methods registered there.

WibbenZ:
Well only the owners have access to the wiki, if it's posible it should be open for PRs aswell.
That way we can keep it updated when we do changes related to the API

wgrriffel:
since after Dec. github support [Github Pages](https://pages.github.com/) for repositories, what do you guys think reborn from the ashes /docs folder  @ranisalt @otland? contributors could push commits to help with tfs documentation furthermore it could still as an online wiki through otland.github.io

ranisalt:
I can edit the wiki but Github wiki system is just so awful it's not possible to open issues or PRs there.

I support moving documentation to a separate branch with a more powerful documentation engine.

wgrriffel:
@ranisalt as i said, you can set up a Github Page pointing to /docs folder as index.html or index.md, it has default web layouts to .md but i think they're awful, i rather a simple one, but docs folder must exists at master branch

ranisalt:
I'll do it later and update everything I can in a `docs` branch, then merge to master.

wgrriffel:
@ranisalt i could help you, you're doing it with HTML5 or what?

ranisalt:
@wgrriffel Github Pages actually supports plain markdown so it is simple to convert from the wiki: https://github.com/blog/2233-publish-your-project-documentation-with-github-pages

ranisalt:
@wgrriffel Github Pages actually supports plain markdown so it is simple to convert from the wiki: https://github.com/blog/2233-publish-your-project-documentation-with-github-pages

data/spells/scripts/support/chameleon.lua
PrinterLUA: Also should it not be possible to use the Rune inside your BP?

raymondtfr:
@ranisalt I am going to test this. But why (#2168).

ranisalt:
No need for more C++ code that is harder to maintain and fix it we can do in plain Lua.

raymondtfr:
Sorry, I must've been a bit too lazy, I meant this PR: [Move Convince Creature to Lua](#2168) .

ranisalt:
That PR has a bit different algorithm but achieves the same. I'll fix this for the sake of having a single PR.

data/spells/scripts/support/chameleon.lua
```diff
@@ -0,0 +1,16 @@
...
+	if not item or item.itemid == 0 or not isMoveable(item.uid) then
```
raymondtfr: I remember far back somewhere in some discussions where Mark stated that we were probably going to get rid of such fields (``item.itemid``, ``item.type``, etc), instead we should be using ``item:getId()`` everywhere xD.

Does ``item:getType():isMoveable()`` achieve the same as ``isMoveable(item.uid)``, if it doesn't you could also replace ``item.uid`` with ``item:getUniqueId()``...

data/spells/scripts/support/convince_creature.lua
```diff
@@ -0,0 +1,31 @@
+function onCastSpell(creature, variant, isHotkey)
+	local target = Creature(variant:getNumber())
+	if not target or not target:isMonster() then
+		creature:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
+		creature:getPosition():sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	local monsterType = target:getType()
+	if not getPlayerFlagValue(creature, PlayerFlag_CanConvinceAll) then
```
raymondtfr: You should probably get rid of such ``getPlayerFlagValue`` statements, we should not care about these flags particularities.

data/spells/scripts/support/convince_creature.lua
```diff
@@ -0,0 +1,31 @@
...
+	local manaCost = target:getType():getManaCost()
+	if creature:getMana() < manaCost and not getPlayerFlagValue(creature, PlayerFlag_HasInfiniteMana) then
+		creature:sendCancelMessage(RETURNVALUE_NOTENOUGHMANA)
+		creature:getPosition():sendMagicEffect(CONST_ME_POFF)
+		return false
+	end
+
+	creature:addMana(-manaCost)
+	creature:addManaSpent(manaCost)
+
```
raymondtfr: I don't think there's a need for a new line here xD.

data/spells/scripts/support/chameleon.lua
Codinablack: This particular script might be an opportunity to use the new itemtype methods like :isUseable()

raymondtfr:
I have just tested **Convince Creature** rune on vanilla.

* If the target is not convinceable, the correct cancel message should be: **"Sorry, not possible."** (``RETURNVALUE_NOTPOSSIBLE``)

* If the target already has two or more *summons*, then the correct cancel message should be: **"You cannot control more creatures."**, leading to a new ``ReturnValue`` constant, i.e. ``RETURNVALUE_YOUCANNOTCONTROLMORECREATURES``.

* If the target already has a master (i.e. the target is a summon) the correct cancel message should be: **"You may not attack this person."**, leading to a new ``ReturnValue`` constant, i.e. ``RETURNVALUE_YOUMAYNOTATTACKTHISPERSON``.

* The correct tile effect sent to the caster should be: ``CONST_ME_MAGIC_BLUE``.

ranisalt:
> If the target already has a master (i.e. the target is a summon)

Does this make sense on PvP worlds? I remember summon selling services where a mage would summon a monk and a knight would convince it for endless training.

raymondtfr:
Unfortunately, I cannot test such mechanism in a PvP world, but I will update it if I am able to do such researches in the future. It makes sense not to add that now.

data/spells/spells.xml
```diff
@@ -494,11 +494,11 @@
 	<rune group="healing" spellid="5" name="Ultimate Healing Rune" id="2273" allowfaruse="1" charges="1" level="24" magiclevel="4" cooldown="2000" groupcooldown="2000" aggressive="0" needtarget="1" blocktype="solid" script="healing/ultimate_healing_rune.lua" />
 
 	<!-- Summon Runes -->
-	<rune group="support" spellid="83" name="Animate Dead Rune" id="2316" allowfaruse="1" charges="1" level="27" magiclevel="4" cooldown="2000" groupcooldown="2000" blocktype="solid" script="support/animate_dead_rune.lua" />
-	<rune group="support" spellid="12" name="Convince Creature Rune" id="2290" allowfaruse="1" charges="1" level="16" magiclevel="5" cooldown="2000" groupcooldown="2000" needtarget="1" blocktype="solid" function="convince" />
+  <rune group="support" spellid="83" name="Animate Dead Rune" id="2316" allowfaruse="1" charges="1" level="27" magiclevel="4" cooldown="2000" groupcooldown="2000" blocktype="solid" script="support/animate_dead_rune.lua" />
```
ranisalt: Argh fuck github editor

PR #2103, opened 2017-01-19
Add damage condition formula for listType == 1 in spells lib

raymondtfr:
This also removes the old way (using table) of mimicking this very same formula.

### DIFF
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -275,7 +275,7 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
 		local value, n = math.floor(damage), 1
 		while value >= 1 do
 			condition:addDamage(1, time or 4000, -value)
-			value, n = math.floor(damage * (2.75 ^ (-0.05 * n)) + 0.35), n + 1
+			value, n = math.floor(damage * math.pow(2.75, -0.05 * n) + 0.35), n + 1
 		end
 	elseif listType == 2 then
 		for _ = 1, rounds do
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -273,7 +273,7 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
 		end
 	elseif listType == 1 then
 		local value, n = math.floor(damage + 0.35), 1
-		while value >= 1 do
+		while value > 0 do
 			condition:addDamage(1, time or 4000, -value)
 			value, n = math.floor(damage * math.pow(2.75, -0.05 * n) + 0.35), n + 1
 		end
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -250,7 +250,7 @@ CORPSES = {
 FIELDS = {1487,1488,1489,1490,1491,1492,1493,1494,1495,1496,1500,1501,1502,1503,1504}
 
 function Creature:addDamageCondition(target, conditionType, listType, damage, time, rounds)
-	if target:isImmune(conditionType) then
+	if damage <= 0 or target:isImmune(conditionType) then
 		return false
 	end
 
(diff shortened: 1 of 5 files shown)


raymondtfr:
@ranisalt Any word on this?

data/spells/lib/spells.lua
```diff
@@ -281,15 +272,14 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
...
+		while value > 0 do
+			value = math.floor(damage * math.pow(2.75, -0.05 * n) + 0.35)
 
-		while damage > 0 do
-			condition:addDamage(rounds[damage] and math.random(rounds[damage][1], rounds[damage][2]) or 1, time or 4000, -damage)
-			damage = damage - (damage > 21 and math.floor(damage / 20) + math.random(0, 1) or 1)
+			if value > 0 then
+				condition:addDamage(1, time or 4000, -value)
+				n = n + 1
+			end
```
ranisalt: Github diff formatter is just horrible, heck.

data/spells/scripts/attack/envenom.lua
```diff
@@ -6,8 +6,8 @@ combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ANI_EARTH)
 function onTargetCreature(creature, target)
 	local min = (creature:getLevel() / 80) + (creature:getMagicLevel() * 0.55) + 6
 	local max = (creature:getLevel() / 80) + (creature:getMagicLevel() * 0.75) + 7
-	local damage = math.random(math.floor(min), math.floor(max))
-	creature:addDamageCondition(target, CONDITION_POISON, 1, target:isPlayer() and math.floor(damage / 2 + 0.5) or damage)
+	local damage = math.random(math.floor(min) * 1000, math.floor(max) * 1000) / 1000
```
ranisalt: What's the difference here? You should floor the result of `math.random(min, max)` btw.

data/spells/scripts/attack/inflict_wound.lua
```diff
@@ -17,8 +17,8 @@ function onTargetCreature(creature, target)
 	local skill = getHighestSkillLevel(creature)
 	local min = (creature:getLevel() / 80) + (skill * 0.2) + 2
 	local max = (creature:getLevel() / 80) + (skill * 0.4) + 2
-	local damage = math.random(math.floor(min), math.floor(max))
-	creature:addDamageCondition(target, CONDITION_BLEEDING, 1, target:isPlayer() and math.floor(damage / 4 + 0.5) or damage)
+	local damage = math.random(math.floor(min) * 1000, math.floor(max) * 1000) / 1000
```
ranisalt: Same as above and probably below.

data/spells/scripts/attack/envenom.lua
raymondtfr: 3e3577d281a5779baa176a55c482e9c9c3c5aa53. As a result, we should not apply ``floor`` to it.

ranisalt: It does not make sense. `min` and `max` are floating point values, and thus have precision of more than 3 digits certainly.

Multiplying by 1000, then dividing by 1000, is the same as not and just adds rounding errors (which do not matter anyway). The damage must be an integer in the end.

Since both ends of the range are floats, the random result can be anything within the range, with the precision you claim for.

raymondtfr: The values of ``min`` and ``max`` are the values generated by the min and max formulas. As a way of keeping the definition of min and max formulas cleaner, I chose not to use ``floor`` on the same line, and instead I decided to go for:
```lua
local damage = math.random(math.floor(min), math.floor(max))
```
instead of:
```lua
local min = math.floor((creature:getLevel() / 80) + (creature:getMagicLevel() * 0.55)) + 6
local max = math.floor((creature:getLevel() / 80) + (creature:getMagicLevel() * 0.75)) + 7
-- then
local damage = math.random(min, max)
```

So we should not use the values generated by the min and max formulas as a way of acquiring such precision.

So, now we get two integer values from min and max using ``floor`` on ``math.random(math.floor(min), math.floor(max))``, but this logarithmic formula requires atleast 3 digits of precision to have the behavior we're looking for for such damage conditions (bleeding, and poison originally), so I multiply both min and max integer values by 1000 and then I divided that by 1000 (permille).

While it is true that both min and max may sometimes generate numbers with 3 digits of precision, it'll by limited due to the fact that the player's level and magic level is a integer, i.e. it will not cover this precision.
```lua
for i = min, max, 0.001 do
    print(i) -- each value incremented by 0.001, covering all values with 3 digits of precision
end
```

And this tell us that every time a user decide to add the damage condition using this list type, he should know (if he wants the proper behavior of this formula) that he will need perform that same operation I did in all monster spells that uses this formula ``math.random(math.floor(min) * 1000, math.floor(max) * 1000) / 1000`` and that he should not apply ``floor`` to it.

raymondtfr:

PR #2109, opened 2017-01-20
Fix #1319

### LINKED ISSUE #1319: Error in Vocation Attackspeed Property [$10]
Hello, I have tested the attackspeed property from vocations.xml, and decrease it to 100, for example.
It only work when I attack the creature while I moving my character, if I stand, attack is set to default.

If you don't understand, watch this movie please, have only 20 seconds:
https://youtu.be/8vnhZzNQXVk

## <bountysource-plugin>
[bountysource](https://www.bountysource.com/issues/12707638-error-in-vocation-attackspeed-property)
</bountysource-plugin>

Mkalo:
This should fix the issue with attackspeed under 1000.

Edit: Basically the same as #1305 the only difference is it will work for fist as well.


brunominervino:
@Mkalo I will test when I back to my home. Start a solution on bountysource.

brunominervino:
@Mkalo I have some delay when I'm attacking running.
I have made some changes and got more performance. Maybe you can improve my changes:
https://github.com/brunominervino/forgottenserver/commit/da0e653b6e0c0718c85c2c03bca3bdab78733d3b

Mkalo:
@brunominervino The delay when running is intended as seen [here].(https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L1212-L1213)

This cancels the next scheduled action task. If this doesn't happen in the real tibia what we could do is instead of using setNextActionTask to schedule the next check we create a new method and use that.

brunominervino:
@Mkalo do you remember old tfs 0.4? My ideia is set something like this, making possible create 4fun servers like "high exp" servers.

Mkalo:
@brunominervino For that you can remove [this](https://github.com/otland/forgottenserver/pull/2109/files#diff-dc52f162280289dd2d4d97d82658362dR3156) line and replace [this](https://github.com/Mkalo/forgottenserver/blob/48f8bca432880b7d6febe65af3033935fdd2a985/src/player.cpp#L3144-L3150) for `result = weapon->useWeapon(this, tool, attackedCreature);` should give you the effect you want, that way the attackspeed will always be constant no matter movement/actions of the player.

brunominervino:
@Mkalo this changes affect attack speed only when I move the character. If I stay on any position, attacks increase to 1s.

zydent:
@brunominervino @Mkalo I would like to be able to put the attack fast of Aries SVN version 0.6.0, but my knowledge in c ++ is null.

I think brunominervino looks for something like this.
**Example;**
https://gfycat.com/DirtyDefenselessDevilfish

brunominervino:
@zydent yes, This is ideal for 4fun servers.

Mkalo:
@brunominervino, @zydent Alright I have pushed the changes, just place `classicAttackSpeed = true` in config.lua and with this changes it should work as you guys want.

WibbenZ:
Add classicAttackSpeed to the config, you shouldn't have to look in the source code to see what you can add to config.lua

brunominervino:
@ranisalt this pr is able to merge?

zydent:
@Mkalo Thank you very much, it's perfect.

diegorodriguesvieira:
@ranisalt this PR is able to merge?

src/configmanager.h
```diff
@@ -40,6 +40,7 @@ class ConfigManager
 			WARN_UNSAFE_SCRIPTS,
 			CONVERT_UNSAFE_SCRIPTS,
 			CLASSIC_EQUIPMENT_SLOTS,
+			CLASSIC_ATTACKSPEED,
```
ranisalt: `ATTACK_SPEED`

ranisalt:
Can you test if the classic attack speed set to true overloads the task scheduler or something? I'm not sure how to do it, it probably has to be tested with a lot of players and a very fast attack.

My point is that since this change does not impact if you use multiples of 1000, it can just be the default behavior if it does not overload the scheduler i.e. not worth maintaining 2 functions if 1 is fine.

Znote:
Forgot to squash, but did a rebase. Still learning GIT. 👍

Roniox12:
Hello how, to do it by using this distro https://github.com/TwistedScorpio/OTHire ? E.g in configmanager.cpp there are all on "m_confInteger" not "boolean" and in luascript.cpp there aren't any line with "registerEnumIn". Thanks in advance !

ranisalt:
> Hello how, to do it by using this distro https://github.com/TwistedScorpio/OTHire ?

Ask there, not here

PR #2110, opened 2017-01-20
Improve damage message building

ranisalt:
- Lazy build message to spectators
- Use stringstream instead of string concatenation

From what I could test, it is working as expected, but I couldn't test complex scenarios with loads of spectators, parties, guild wars, etc.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4023,17 +4023,11 @@ bool Game::combatChangeHealth(Creature* attacker, Creature* target, CombatDamage
 		}
 
 		if (message.primary.color != TEXTCOLOR_NONE || message.secondary.color != TEXTCOLOR_NONE) {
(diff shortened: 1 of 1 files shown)


raymondtfr:
@ranisalt There is also the mana counterpart of this code.

ranisalt:
Yes, the reason is I really didn't see them 🤣 can you point them out?

raymondtfr:
[This](https://github.com/ranisalt/forgottenserver/blob/cf338400562ca88643faa8f271d67edcffbb0ecd/src/game.cpp#L3820), [this](https://github.com/ranisalt/forgottenserver/blob/cf338400562ca88643faa8f271d67edcffbb0ecd/src/game.cpp#L3918), and [this](https://github.com/ranisalt/forgottenserver/blob/cf338400562ca88643faa8f271d67edcffbb0ecd/src/game.cpp#L4152).

PR #2112, opened 2017-01-21
Use better types and names for date fields

ranisalt:
This aims to make the database stricter and better typed by using time fields for time values e.g. creation times, expiration times. It enables lots of (My)SQL functions to be used on those fields (such as `NOW`, `DATE_ADD`, etc) and reduce the friction for AAC development, as often web frameworks handle time fields internally and having them as integers prohibits this automatic management of time.

Additionally, I renamed time fields to use standard `<action>_at` and `last_<action>` names, also a widespread pattern for frameworks.

Implications are `FROM_UNIXTIME` and `UNIX_TIMESTAMP` functions are needed to convert timestamp to datetime back and forth while querying the database. I made those changes where I could find them.

The migration is rather huge, but simple.


Mkalo:
This is not backward compatible tho as it will break scripts that fetched dates without using UNIX_TIMESTAMP in the query.

DSpeichert:
@ranisalt Most likely the schema.sql should never be updated. If you import the schema.sql to an empty database and then the server tries to run the migrations, it will fail.

EPuncker:
couldn't we have at least the data type change? 😛

PR #2113, opened 2017-01-21
Fix direction when changing floors

t0kenz:
... while standing on 3 stackable objects

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -769,9 +769,9 @@ ReturnValue Game::internalMoveCreature(Creature* creature, Direction direction, 
 	if (player && !diagonalMovement) {
 		//try go up
 		if (currentPos.z != 8 && creature->getTile()->hasHeight(3)) {
-			Tile* tmpTile = map.getTile(currentPos.x, currentPos.y, currentPos.getZ() - 1);
+			Tile* tmpTile = map.getTile(currentPos.x, currentPos.y, currentPos.z - 1);
 			if (tmpTile == nullptr || (tmpTile->getGround() == nullptr && !tmpTile->hasFlag(TILESTATE_BLOCKSOLID))) {
-				tmpTile = map.getTile(destPos.x, destPos.y, destPos.getZ() - 1);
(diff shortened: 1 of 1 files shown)


Kamenuvol:
But you can change floors going to <-- and it should be looking to the left

t0kenz:
oh yeah, didn't think about it let me fix it :-)

src/game.cpp
```diff
@@ -763,18 +763,20 @@ ReturnValue Game::internalMoveCreature(Creature* creature, Direction direction,
...
 	Position destPos = getNextPosition(direction, currentPos);
+	Player* player = creature->getPlayer();
 
 	bool diagonalMovement = (direction & DIRECTION_DIAGONAL_MASK) != 0;
-	if (creature->getPlayer() && !diagonalMovement) {
+	if (player && !diagonalMovement) {
 		//try go up
 		if (currentPos.z != 8 && creature->getTile()->hasHeight(3)) {
-			Tile* tmpTile = map.getTile(currentPos.x, currentPos.y, currentPos.getZ() - 1);
+			Tile* tmpTile = map.getTile(currentPos.x, currentPos.y, currentPos.z - 1);
```
ranisalt: Don't change this line as it's not related to the issue.

src/game.cpp
```diff
@@ -763,18 +763,20 @@ ReturnValue Game::internalMoveCreature(Creature* creature, Direction direction,
...
 			if (tmpTile == nullptr || (tmpTile->getGround() == nullptr && !tmpTile->hasFlag(TILESTATE_BLOCKSOLID))) {
-				tmpTile = map.getTile(destPos.x, destPos.y, destPos.getZ() - 1);
+				tmpTile = map.getTile(destPos.x, destPos.y, destPos.z - 1);
```
ranisalt: Same as above.

Kamenuvol:
nice name for a commit 😆

ranisalt:
Thanks!

I felt like the boss now 🤣

PR #2116, opened 2017-01-21
Fix issue with summon_creature

Mkalo:
This fixes the issue when summoning non convinceable monsters.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -6733,11 +6733,13 @@ int LuaScriptInterface::luaCreatureSetMaster(lua_State* L)
 		return 1;
 	}
 
-	Creature* master = getCreature(L, 2);
-	bool force = getBoolean(L, 3, false);
-	if (master) {
-		if (force) {
+	if (Creature* master = getCreature(L, 2)) {
+		if (getBoolean(L, 3, false)) {
 			if (Monster* summon = creature->getMonster()) {
+				if (Creature* oldMaster = summon->getMaster()) {
+					oldMaster->removeSummon(summon);
+				}
+
 				master->addSummon(summon);
 				summon->setFollowCreature(nullptr);
 				summon->setAttackedCreature(nullptr);
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -6726,7 +6726,7 @@ int LuaScriptInterface::luaCreatureGetMaster(lua_State* L)
 
 int LuaScriptInterface::luaCreatureSetMaster(lua_State* L)
 {
-	// creature:setMaster(master)
+	// creature:setMaster(master[, force = false])
 	Creature* creature = getUserdata<Creature>(L, 1);
 	if (!creature) {
 		lua_pushnil(L);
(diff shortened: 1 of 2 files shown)


src/luascript.cpp
```diff
@@ -6726,16 +6726,30 @@ int LuaScriptInterface::luaCreatureGetMaster(lua_State* L)
...
 	Creature* creature = getUserdata<Creature>(L, 1);
 	if (!creature) {
 		lua_pushnil(L);
 		return 1;
 	}
 
 	Creature* master = getCreature(L, 2);
+	bool force = getBoolean(L, 3, false);
 	if (master) {
-		pushBoolean(L, creature->convinceCreature(master));
```
WibbenZ: Have you tested other scripts that use this function to make sure it doesn't break it?
I tested adding the Creature::addSummon(Creature* creature) to Lua (as you pretty much did here) but that still didn't fix the bug.
Tested an older src from nightlies and had the same issue with the old C++ code.

https://github.com/otland/forgottenserver/commit/de73b47e3bd66007584a2017a7481213fc9c22af

Mkalo:
I reverted the commits about this change and there was no bug before. I have tested this PR as well and it fixes the bug.

src/luascript.cpp
Mkalo: It won't break it, the default for the force argument is false. It will work exactly the same when force is not set to true.

WibbenZ:
flash

With this update I can still attack summons, both fire elemental and wasps.
And as I said, if I revert the Lua port I can still attack them and that makes me think this is not related to the Lua port nor this function.

@brunominervino could you test this fix aswell to make sure it works for you?

Mkalo:
@WibbenZ That is not the issue. You have understood it wrong. Login in a normal character (not god) and do utevo res "wasp. It will spawn a wasp but wont be able to convince it because setMaster is using convinceCreature internally. It works for gods because they have the flag to convince any creature.

WibbenZ:
Oh okay nvm then

Kamenuvol:
I don't understand what is being forced. It shouldn't be the default behaviour for setMaster?

Mkalo:
@Kamenuvol Well I didn't wanna change it to keep backward compatibility, the thing that is being forced is because it doesn't check if its either summonable or convinceable (probably how the method should've worked from the beginning).

comedinha:
the old code only use addsummon
https://github.com/otland/forgottenserver/commit/de73b47e3bd66007584a2017a7481213fc9c22af#diff-72726bca5c006a50f5f53f0a15423ceaL931
if player have attacking other monster and use utevo res, no have bug?

WibbenZ:
https://github.com/otland/forgottenserver/blob/master/src/creature.cpp#L1124

comedinha:
@WibbenZ the change is on
+				summon->setFollowCreature(nullptr);
 +				summon->setAttackedCreature(nullptr);
 +				summon->updateTargetList();
 +				summon->updateIdleStatus();
but update target list, possibly no have bugs

WibbenZ:
No ide if its called from one of the other ones, you can check.

Mkalo:
The idea of having this is setMaster can be called in another context like if the monster is already attacking a creature.

Kamenuvol:
setMaster should change the master without any checks

ranisalt:
> setMaster should change the master without any checks

I agree with @Kamenuvol here. If a scripter wants to check if the creature is convinceable, just run `creature:isConvinceable()` before.

Mkalo:
Closed due to #2131

PR #2119, opened 2017-01-22
Add Position:isWalkable()

Riverlance:
This function Position:isWalkable() is very useful for be used.

@WibbenZ

### DIFF
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -18,29 +18,29 @@ function Position:getNextPosition(direction, steps)
 	end
 end
 
-function Position:moveUpstairs()
-	local isWalkable = function (position)
-		local tile = Tile(position)
-		if not tile then
-			return false
-		end
-
-		local ground = tile:getGround()
-		if not ground or ground:hasProperty(CONST_PROP_BLOCKSOLID) then
-			return false
-		end
-
-		local items = tile:getItems()
-		for i = 1, tile:getItemCount() do
-			local item = items[i]
-			local itemType = item:getType()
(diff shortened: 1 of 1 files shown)


Kamenuvol:
Tile is walkable or not, not position

Riverlance:
Ooh... I see. How can I fix this in here. Let me see.

data/lib/core/position.lua
```diff
@@ -18,29 +18,29 @@ function Position:getNextPosition(direction, steps)
...
 end
 
-function Position:moveUpstairs()
-	local isWalkable = function (position)
-		local tile = Tile(position)
-		if not tile then
-			return false
-		end
+function Position:isWalkable()
+	local tile = Tile(position)
```
WibbenZ: Remove this line, use Tile:isWalkable() on line 21.
Then replace tile with self on all functions

ex (line 23)
```
if not self then
...
end
```

Riverlance: How can I cancel this pull request? I will do another.

WibbenZ: You can probbly do changes here: https://github.com/Riverlance/forgottenserver/tree/patch-1
And they should be updated here when you push things there...

Option 2 install the git programs and hope for the best (that is what I do everytime, works about 70% of the times haha)

Option 3, below the comment thingi you should have an option to close the PR.

Riverlance:
@WibbenZ How can I edit 2 or more files and do only 1 pull request on GitHub?

WibbenZ:
As I said, use this link and edit the files there: https://github.com/Riverlance/forgottenserver/tree/patch-1
But you deleted the branch so idk if you can do it unless you restore the branch and open this PR again.

PR #2120, opened 2017-01-22
Add Tile:isWalkable()

Riverlance:
@WibbenZ I need to add changes on position.lua also for this works. How do I do this?

### DIFF
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -37,7 +37,7 @@ function Position:moveUpstairs()
 
 			local position = self + Position.directionOffset[direction]
 			toTile = Tile(position)
-			if tilePosition and tilePosition:isWalkable() then
+			if toTile and toTile:isWalkable() then
 				swap(self, position)
 				return self
 			end
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -19,28 +19,6 @@ function Position:getNextPosition(direction, steps)
 end
 
 function Position:moveUpstairs()
-	local isWalkable = function (position)
-		local tile = Tile(position)
-		if not tile then
-			return false
-		end
-
-		local ground = tile:getGround()
-		if not ground or ground:hasProperty(CONST_PROP_BLOCKSOLID) then
-			return false
-		end
-
-		local items = tile:getItems()
-		for i = 1, tile:getItemCount() do
-			local item = items[i]
-			local itemType = item:getType()
-			if itemType:getType() ~= ITEM_TYPE_MAGICFIELD and not itemType:isMovable() and item:hasProperty(CONST_PROP_BLOCKSOLID) then
-				return false
-			end
-		end
-		return true
-	end
-
 	local swap = function (lhs, rhs)
 		lhs.x, rhs.x = rhs.x, lhs.x
 		lhs.y, rhs.y = rhs.y, lhs.y
@@ -50,14 +28,16 @@ function Position:moveUpstairs()
 	self.z = self.z - 1
 
 	local defaultPosition = self + Position.directionOffset[DIRECTION_SOUTH]
-	if not isWalkable(defaultPosition) then
+	local tileDefaultPosition = Tile(defaultPosition)
+	if not tileDefaultPosition or not tileDefaultPosition:isWalkable() then
 		for direction = DIRECTION_NORTH, DIRECTION_NORTHEAST do
 			if direction == DIRECTION_SOUTH then
 				direction = DIRECTION_WEST
 			end
 
 			local position = self + Position.directionOffset[direction]
-			if isWalkable(position) then
+			local tilePosition = Tile(position)
+			if tilePosition and tilePosition:isWalkable() then
 				swap(self, position)
 				return self
 			end
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -28,15 +28,15 @@ function Position:moveUpstairs()
 	self.z = self.z - 1
 
 	local defaultPosition = self + Position.directionOffset[DIRECTION_SOUTH]
-	local tileDefaultPosition = Tile(defaultPosition)
-	if not tileDefaultPosition or not tileDefaultPosition:isWalkable() then
+	local toTile = Tile(defaultPosition)
+	if not toTile or not toTile:isWalkable() then
 		for direction = DIRECTION_NORTH, DIRECTION_NORTHEAST do
 			if direction == DIRECTION_SOUTH then
 				direction = DIRECTION_WEST
 			end
 
 			local position = self + Position.directionOffset[direction]
-			local tilePosition = Tile(position)
+			toTile = Tile(position)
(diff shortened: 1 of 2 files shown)


Kamenuvol:
Use position:getTile():isWalkable, not sure if is getTile(). It's something like that.

Riverlance:
It worked for me 100%.
And I followed the TFS model.

Kamenuvol:
No, you should commit locally and push to the remote. It will automatically update this pull request.

Riverlance:
Take a look. I tested here and it worked for me. 👍

data/lib/core/position.lua
```diff
@@ -50,14 +28,16 @@ function Position:moveUpstairs()
 	self.z = self.z - 1
 
 	local defaultPosition = self + Position.directionOffset[DIRECTION_SOUTH]
-	if not isWalkable(defaultPosition) then
+	local tileDefaultPosition = Tile(defaultPosition)
```
WibbenZ: Use toTile here insted,

WibbenZ:
And I told you before to use this link if you aren't using the github software on your computer: https://github.com/Riverlance/forgottenserver

From there go to the file and hit the edit button.

data/lib/core/position.lua
Riverlance: But there is 2 Tile(pos) on this code.
Why it should be only there?
Also, I can't find any toTile() function.

WibbenZ: No insted of tileDefaultPosition use:
local toTile = Tile(defaultPosition)

And then you can overwrite toTile with Tile(position) on line 39

Riverlance:
@WibbenZ Fixed what you said. 😄

Riverlance:
@WibbenZ I will add a check if the tile exists, okay? Wait a little.

Riverlance:
Forgot to change these vars, sorry.

Riverlance:
@WibbenZ Can I improve this like tile:isWalkable(params)?
Params will be used for things like ignore creatures, teleports, protection zones, stairs, etc.
If params == nil, it will work normally doing no checks.

Kamenuvol:
Do what you want 😄

WibbenZ:
I agree with kamen
As long as you check that variable quickly to avoid loading code that isent needed

joseluis2g:
@Riverlance 
```
if not params
```

Riverlance:
Added these flags:
> 1. protection
> 2. floorChange
> 3. house
> 4. teleport
> 5. magicField
> 6. creatures
> 7. players
> 8. monsters
> 9. npcs

Any suggestions? 😄

data/lib/core/tile.lua
```diff
@@ -32,3 +32,44 @@ function Tile.relocateTo(self, toPosition)
 
 	return true
 end
+
+function Tile.isWalkable(self, check)
+  local ground = self:getGround()
```
WibbenZ: This is the downside of OTC imo, ppl might start out with scripting on OTC and then think we use spaces, but we should try to keep a consistency in our code style, that is why my project uses tabs xD

So to my review, use tabs insted of spaces, most of the OT stuff uses tabs.
Most coding is done via spaces, since you can modify your editor to show a space as 2 spaces (while you then press tab the next editor might show it as 5 spaces insted of forcing people to use 2 spaces).

If people wanna argue about using parentheses around if statment etc that is a diffrent story tho :P

data/lib/core/tile.lua
```diff
@@ -32,3 +32,44 @@ function Tile.relocateTo(self, toPosition)
...
+    if check.protection and self:hasFlag(TILESTATE_PROTECTIONZONE) or
+      check.floorChange and self:hasFlag(TILESTATE_FLOORCHANGE) or
+      check.house and self:getHouse() ~= nil then
+      return false
+    elseif check.creatures or check.players or check.monsters or check.npcs then
+      local creatures = self:getCreatures()
+      for i = 1, self:getCreatureCount() do
+        local creature = creatures[i]
+        if check.creatures and creature:isCreature() or
+					check.players and creature:isPlayer() or
```
WibbenZ: IMO these shoud be one tab away, to make it more clean.
````
if ...
	...
	...
	...
	then
	
end
```

Riverlance:
@WibbenZ Fixed spaces. :)

Kamenuvol:
some suggestions for check: 
actionid, uniqueid (check all items on the tile)
and storage value

Kamenuvol:
sorry, my suggestions can't be made because it's not saying not about being walkable or not 😞 haha

Riverlance:
@Kamenuvol I think this makes sense, because you could add an actionid/uniqueid for avoid the player passage through this tile, so it wont be walkable. But the storage does not. 😄

data/lib/core/tile.lua
```diff
@@ -33,19 +33,43 @@ function Tile.relocateTo(self, toPosition)
...
-		local itemType = item:getType()
-		if itemType:getType() ~= ITEM_TYPE_MAGICFIELD and not itemType:isMovable() and item:hasProperty(CONST_PROP_BLOCKSOLID) then
-			return false
-		end
-	end
-	return true
+  if check then
+    if check.protection and self:hasFlag(TILESTATE_PROTECTIONZONE) or
+      check.floorChange and self:hasFlag(TILESTATE_FLOORCHANGE) or
+      check.house and self:getHouse() ~= nil then
```
vankk: Not need the ~= nil

Kamenuvol:
@Riverlance yeah, but how does the function know if that specific actionid or uniqueid allow the creature to walk? it depends on movement scripts and it is not possible to check it

Riverlance:
@Kamenuvol What do you think?

image

data/lib/core/tile.lua
Riverlance: I see. I copied this from getTileInfo(position) of compat.lua.
I will change this, but I suggest you to remove there also. :)

ninjalulz:
This implementation of yours is too inconvenient. It would be better if you pass a callback where users can set their own condition(s).

e.g.

```Lua
function Tile.isWalkable(self, condition)
    if condition and condition(self) then
        return true
    end

    -- Default Code
end

-- An example
tile:isWalkable(function(tile) return tile:hasFlag(TILESTATE_PROTECTIONZONE) end)
```

Kamenuvol:
I would put item:hasProperty first in line 67, so if it pass doesnt need to check even if check is true

Riverlance:
@ninjalulz I don't think is too inconvenient as you said, but I like your idea also. I did by this way because these checks I did is basically for see if something is walkable or not, that's why I did this way. I used all these checks on my server and they were very useful, so I thought it would be for others also.

@Kamenuvol Sure. :)

Riverlance:
image


@Kamenuvol @WibbenZ @ninjalulz

Riverlance:
Added tileCondition and itemCondition. :)

PR #2121, opened 2017-01-23
Rename overly abbreviated field names on database

ranisalt:
There are some tables which have columns named `pid` and `sid` and that does not describe anything. `pid` generally refers to the process ID and, in this context, can be the player ID too, but in fact means the container parent ID. `sid` is in the best case the sloth from Ice Age and was pretty tricky to understand but I named it `slot_id` instead.

I also renamed some variables using the same names that were close to the queries.

### DIFF
--- a/data/migrations/19.lua
+++ b/data/migrations/19.lua
@@ -1,3 +1,7 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 20 (renaming abbreviated fields)")
+	for _, table in ipairs({"player_depotitems", "player_inboxitems", "player_items"}) do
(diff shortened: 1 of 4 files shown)


jo3bingham:
LGTM FWIW

Kamenuvol:
Only now I am thinking why does mailbox slot id still exists? delete it!

ranisalt:
> Only now I am thinking why does mailbox slot id still exists? delete it!

Because all tables are handled by a single `saveItems` method that relies on tables having `parent_id` and `slot_id` fields. Also, it works as ordering of items, so items don't get randomly arranged on your mailbox, but sorted by time of arrival.

EPuncker:
@DSpeichert what about adding this to 1.4? 👀

luanluciano93:
This pull request could be approved before the version update ...

PR #2123, opened 2017-01-23
Fix 'force' of Player::toggleMount

Riverlance:
@WibbenZ @Kamenuvol :)

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -4049,31 +4049,35 @@ void Player::setCurrentMount(uint8_t mount)
 	addStorageValue(PSTRG_MOUNTS_CURRENTMOUNT, mount);
 }
 
-bool Player::toggleMount(bool mount)
+bool Player::toggleMount(bool mount, bool force /*= false*/)
 {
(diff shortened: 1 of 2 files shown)


jo3bingham:
Build failure is because you forgot to update the function declaration in player.h.

Also, maybe I'm failing to comprehend, but I don't understand the purpose of this pr. Feel free to explain, if you don't mind.

Riverlance:
Well, I made this possible to toggle the mount by forcing (without the checks).
So, if it forces, it will stops only if the mount id is incorrect.
If the mount id is correct, it will force the mount change (without checking if is in Protection Zone, exhaustion, and so on).

> If you still want this, I add the change on player.h. Else, I understand also. :)

jo3bingham:
Yes, but as it stands, force will always be false unless the user modifies the source directly. It would make much more sense if you added a Lua function that took advantage of this.

Riverlance:
Hmm... if someone teach me by Team Viewer about how I can send my commits changes of my project as a pull request in here, I will be able to send some of my mount functions. xd

PR #2126, opened 2017-01-25
Lazily load monsters from XML

ranisalt:
TFS currently has 664 monsters and 7 NPCs but only 59 of them actually spawn on map. Loading all monsters and NPCs is therefore unnecessary, so lazy loading them on demand has benefits to startup time.

On my machine, it brings down the time to load monsters from c. 1.16s to 0.46s with the default map. It should be good for servers with low amount of spawns but high monster count (e.g. lots of event monsters & NPCs) while not a significant difference otherwise.

Motivation: most monsters either have numerous spawns, in which case the penalty of lazy loading is amortized to be constant, or are single, evented spawns, in which case the server benefits from not loading them on startup.

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -166,6 +166,7 @@ bool MonsterType::createLootContainer(Container* parent, const LootBlock& lootbl
 
 bool Monsters::loadFromXml(bool reloading /*= false*/)
 {
+	unloadedMonsters = {};
 	pugi::xml_document doc;
 	pugi::xml_parse_result result = doc.load_file("data/monster/monsters.xml");
 	if (!result) {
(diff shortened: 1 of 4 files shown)


WibbenZ:
Cool keep the good work up :)
Btw you missed to edit the road map when you added the idle npcs

src/monsters.cpp
```diff
@@ -1182,10 +1181,16 @@ void Monsters::loadLootContainer(const pugi::xml_node& node, LootBlock& lBlock)
 
 MonsterType* Monsters::getMonsterType(const std::string& name)
 {
-	auto it = monsters.find(asLowerCaseString(name));
+	std::string lowerCaseName = asLowerCaseString(name);
 
+	auto it = monsters.find(lowerCaseName);
 	if (it == monsters.end()) {
-		return nullptr;
+		auto it = unloadedMonsters.find(lowerCaseName);
```
Mkalo: Change this var name or the other, VS complains about it.
`1>..\src\monsters.cpp(1188): warning C4456: declaration of 'it' hides previous local declaration`

ranisalt: I think I'll name it `it2` as other pieces of code do.

conde2:
I will still be able to use MonsterType in unloaded monsters ? If not then it is not backwards compatible.

Some server may use MonsterType(monsterNme) before actually adding the monster to the game.

vankk:
I agree with @conde2

Kamenuvol:
Loading only a list of monsters should be enough to use MonsterType

conde2:
In my opinion there is not problem if we spend a little more time in the startup and save processing before, another point is: what if a player wants to summon a non loaded monster ?

ranisalt:
It will work. All functions use `Monsters::getMonsterType` that loads a monster if currently unloaded.

djarek:
@ranisalt 
Is there something left to be done in this PR?

ranisalt:
No, @Mkalo didn't change his review but I fixed what he pointed out. I'm merging.

PR #2129, opened 2017-01-26
Updated Party class to notify the leader if sharing experience is possible

raymondtfr:
All these changes below have been researched on vanilla.

- Fixed the formula of minimum level required to share experience.
- Now the leader will be notified (after enabling shared experience) whenever the level spread of the party is too wide, which improves gaming awareness..

### DIFF
--- a/src/party.cpp
+++ b/src/party.cpp
@@ -446,7 +446,7 @@ SharedExperienceState_t Party::canUseSharedExperience(const Player* player) cons
 		}
 	}
 
-	uint32_t minLevel = static_cast<int32_t>(std::ceil((static_cast<float>(highestLevel) * 2) / 3));
+	uint32_t minLevel = static_cast<uint32_t>(highestLevel * 2 / 3.f);
 	if (player->getLevel() < minLevel) {
 		return SHAREDEXPERIENCE_STATE_ACTIVATED_LEVELSPREADTOOWIDE;
 	}
--- a/src/party.cpp
+++ b/src/party.cpp
@@ -395,15 +395,29 @@ bool Party::setSharedExperience(Player* player, bool sharedExpActive)
 	this->sharedExpActive = sharedExpActive;
 
 	if (sharedExpActive) {
-		this->sharedExpEnabled = canEnableSharedExperience();
+		sharedExpEnabled = canEnableSharedExperience();
+	}
 
-		if (this->sharedExpEnabled) {
+	switch (sharedExpState) {
+		case SHAREDEXPERIENCE_STATE_ACTIVE: {
 			leader->sendTextMessage(MESSAGE_INFO_DESCR, "Shared Experience is now active.");
-		} else {
-			leader->sendTextMessage(MESSAGE_INFO_DESCR, "Shared Experience has been activated, but some members of your party are inactive.");
+			break;
 		}
-	} else {
-		leader->sendTextMessage(MESSAGE_INFO_DESCR, "Shared Experience has been deactivated.");
+
+		case SHAREDEXPERIENCE_STATE_ACTIVATED_LEVELSPREADTOOWIDE: {
+			leader->sendTextMessage(MESSAGE_INFO_DESCR, "Shared Experience has been activated, but the level spread of your party is too wide.");
+			break;
+		}
+
+		case SHAREDEXPERIENCE_STATE_ACTIVATED_MEMBERSINACTIVE: {
+			leader->sendTextMessage(MESSAGE_INFO_DESCR, "Shared Experience has been activated, but some members of your party are inactive.");
+			break;
+		}
+
+		case SHAREDEXPERIENCE_STATE_DEACTIVATED: {
+			leader->sendTextMessage(MESSAGE_INFO_DESCR, "Shared Experience has been deactivated.");
+			break;
+		}
 	}
 
 	updateAllPartyIcons();
@@ -419,10 +433,10 @@ void Party::shareExperience(uint64_t experience, Creature* source/* = nullptr*/)
 	leader->onGainSharedExperience(shareExperience, source);
 }
 
-bool Party::canUseSharedExperience(const Player* player) const
+SharedExperienceState_t Party::canUseSharedExperience(const Player* player) const
 {
 	if (memberList.empty()) {
-		return false;
+		return SHAREDEXPERIENCE_STATE_ACTIVATED_MEMBERSINACTIVE;
 	}
 
 	uint32_t highestLevel = leader->getLevel();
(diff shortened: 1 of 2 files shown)


src/party.cpp
```diff
@@ -432,8 +434,9 @@ bool Party::canUseSharedExperience(const Player* player) const
 		}
 	}
 
-	uint32_t minLevel = static_cast<int32_t>(std::ceil((static_cast<float>(highestLevel) * 2) / 3));
```
raymondtfr: Can anyone explain me why we are casting this expression here to ``int32_t`` and then assigning to a ``uint32_t``? @kornholi @ranisalt @djarek

Mkalo: I'm pretty sure it was a typo.

ranisalt: There are lots of errors like this throughout the code. I'd even use `auto` here as the type is very explicit.

src/party.cpp
```diff
@@ -432,8 +434,9 @@ bool Party::canUseSharedExperience(const Player* player) const
...
+	uint32_t minLevel = static_cast<int32_t>(static_cast<float>(highestLevel * 2 / 3.));
```
ranisalt: You don't need to cast to float as the division already casts it.

src/party.cpp
```diff
@@ -432,8 +434,9 @@ bool Party::canUseSharedExperience(const Player* player) const
...
 	if (player->getLevel() < minLevel) {
+		this->hasRequiredLevelToShare = false;
```
ranisalt: No need for `this->`

src/party.cpp
ranisalt: You don't need to cast to float as the division already casts it.

src/party.cpp
```diff
@@ -419,7 +421,7 @@ void Party::shareExperience(uint64_t experience, Creature* source/* = nullptr*/)
 	leader->onGainSharedExperience(shareExperience, source);
 }
 
-bool Party::canUseSharedExperience(const Player* player) const
+bool Party::canUseSharedExperience(const Player* player)
```
ranisalt: A query method should always be `const`. It doesn't make sense to mutate the object on a `canSomething` method.

(yes, queryAdd and similars are doing the wrong thing but ¯\_(ツ)_/¯)

src/party.cpp
raymondtfr: That is because we are comparing it to ``player->getLevel()``, so I think it serves that purpose (i.e. we should know that ``player->getLevel()`` returns a ``uint32_t`` inferring from ``minLevel``'s type) 

Should I change it to ``auto`` nonetheless?

src/party.cpp
raymondtfr: I remove ``const`` frm this member-function because I needed to modify the data member ``hasRequiredLevelToShare`` inside of it.

src/party.cpp
raymondtfr: ``3.f?``

src/party.cpp
raymondtfr: Okay, I just used it because we would avoid having variables with the same name if it ever happened with this one, which is unlikely.

raymondtfr: Should I also refrain from using ``this->`` in ``Party::setSharedExperience``?

src/party.cpp
raymondtfr: How would we know which condition it would have failed on?

Kamenuvol: using the function directly there as it returns a bool

raymondtfr: I don't understand...

ranisalt: `else if (!canShareExperience(...))`

Kamenuvol: sorry, i didnt read the entire function canUseSharedExperience, there are more "exits" from it than I thought
maybe you can use return values like other queries and create enum for it

raymondtfr: But we would not know which ``if`` condition was evaluated to ``false`` in ``Party::canUseSharedExperience``, it could be either because of [this](https://github.com/raymondtfr/forgottenserver/blob/bd6c492d60f00bbb258042f5a6ffd0e3dc03c9d7/src/party.cpp#L426), or [this](https://github.com/raymondtfr/forgottenserver/blob/bd6c492d60f00bbb258042f5a6ffd0e3dc03c9d7/src/party.cpp#L438) (which is what really matter for us), or [this](https://github.com/raymondtfr/forgottenserver/blob/bd6c492d60f00bbb258042f5a6ffd0e3dc03c9d7/src/party.cpp#L443), or [this](https://github.com/raymondtfr/forgottenserver/blob/bd6c492d60f00bbb258042f5a6ffd0e3dc03c9d7/src/party.cpp#L450), or even [this](https://github.com/raymondtfr/forgottenserver/blob/bd6c492d60f00bbb258042f5a6ffd0e3dc03c9d7/src/party.cpp#L455).

So we could be sending that message description even if the member had the required level to share experience within a party.

Mkalo: I don't like the way you're doing this tho, I would rather have a new ReturnValue than this and then use:

```cpp
ReturnValue returnValue = canEnableSharedExperience();
this->sharedExpEnabled = returnValue == RETURNVALUE_SHAREDENABLE ? true : false;
leader->sendTextMessage(MESSAGE_INFO_DESCR, getReturnMessage(returnValue));
```
Something like that.

raymondtfr: But that would require additional code that is nowhere needed, the way I did follows the current pattern used between these two functions.

ranisalt: It's not a problem to have more code if this translates to better readability and quality.

raymondtfr:
@Mkalo  Would you mind trying to better introduce these changes the way you described? I'm not putting too much effort on programming right now (or ever :s). I tried several times :sweat: to design it the way you described several times (past and now) but I couldn't do it. I, we would appreciate it.

@ranisalt

src/party.cpp
```diff
@@ -399,6 +399,8 @@ bool Party::setSharedExperience(Player* player, bool sharedExpActive)
 
 		if (this->sharedExpEnabled) {
 			leader->sendTextMessage(MESSAGE_INFO_DESCR, "Shared Experience is now active.");
+		} else if (!this->hasRequiredLevelToShare) {
```
ranisalt: Here, you should call `canUseSharedExperience` instead of setting a variable. It's not that costly anyway. This way you wouldn't need to change the constness of the function.

src/party.cpp
ranisalt: No, the way you did is ok. The static cast is implicit, though, but keep it that way :)

src/party.h
```diff
@@ -75,7 +75,7 @@ class Party
 		bool isSharedExperienceEnabled() const {
 			return sharedExpEnabled;
 		}
-		bool canUseSharedExperience(const Player* player) const;
+		bool canUseSharedExperience(const Player* player);
```
ranisalt: It should take a reference instead, since it does not make sense for a player to be null here.

src/party.cpp
```diff
@@ -432,8 +434,9 @@ bool Party::canUseSharedExperience(const Player* player) const
 		}
```
ranisalt: I suggest you use `std::max` above, but feel free to not mess with it.

raymondtfr:
@ranisalt I have tried one last time to make it the way @Mkalo described, though it may look even worse than what I had done before lol. I'm not even sure I should use a ``enum`` (should I explictly add a type for it?) or new *return values*, can you please review this? again? xD.

I tried my best to not cause any side effects.

src/party.cpp
```diff
@@ -395,15 +395,29 @@ bool Party::setSharedExperience(Player* player, bool sharedExpActive)
 	this->sharedExpActive = sharedExpActive;
 
 	if (sharedExpActive) {
-		this->sharedExpEnabled = canEnableSharedExperience();
+		sharedExpEnabled = canEnableSharedExperience();
```
ranisalt: This will change the parameter value, not the class attribute!

raymondtfr: I don't quite understand what you mean xD. ``canEnableSharedExperience()`` has been modified in order to set the current state of shared experience, letting us know which message to send.

ranisalt: Are you sure? `canEnableSharedExperience` is a const function and does not modify the party.

raymondtfr:
Closing this for now as I can not find a clean solution atm.

PR #2131, opened 2017-01-27
Fix setMaster() behaviour

Kamenuvol:
It may break some code.

addSummon and removeSummon were removed because they should only set master not dropLoot and LossSkill, so I made a setMaster that only sets or unsets (add or remove) the monster to the master.

The plan is remove convince creature from the sources to Lua. This is the first step.

Please, test it!

### DIFF
--- a/data/lib/core/creature.lua
+++ b/data/lib/core/creature.lua
@@ -91,9 +91,9 @@ function Creature:addSummon(monster)
 		return false
 	end
 
-	summon.setMaster(self)
-	summon.setDropLoot(false)
-	summon.setSkillLoss(false)
+	summon:setMaster(self)
+	summon:setDropLoot(false)
+	summon:setSkillLoss(false)
 
 	return true
 end
@@ -104,9 +104,9 @@ function Creature:removeSummon(monster)
 		return false
 	end
 
-	summon.setMaster(nil)
-	summon.setDropLoot(true)
-	summon.setSkillLoss(true)
+	summon:setMaster(nil)
+	summon:setDropLoot(true)
+	summon:setSkillLoss(true)
 
 	return true
 end
--- a/data/lib/core/creature.lua
+++ b/data/lib/core/creature.lua
@@ -91,9 +91,11 @@ function Creature:addSummon(monster)
 		return false
 	end
 
-	summon:setMaster(self)
+	summon:setTarget(nil)
+	summon:setFollowCreature(nil)
 	summon:setDropLoot(false)
 	summon:setSkillLoss(false)
+	summon:setMaster(self)
 
 	return true
 end
@@ -104,9 +106,11 @@ function Creature:removeSummon(monster)
 		return false
 	end
 
-	summon:setMaster(nil)
+	summon:setTarget(nil)
+	summon:setFollowCreature(nil)
 	summon:setDropLoot(true)
 	summon:setSkillLoss(true)
+	summon:setMaster(nil)
 
 	return true
 end
--- a/data/lib/core/creature.lua
+++ b/data/lib/core/creature.lua
@@ -84,3 +84,29 @@ function Creature:setItemOutfit(item, time)
 
 	return true
 end
+
+function Creature:addSummon(monster)
+	local summon = Monster(monster)
+	if not summon then
+		return false
+	end
+
+	summon.setMaster(self)
+	summon.setDropLoot(false)
+	summon.setSkillLoss(false)
+
+	return true
+end
+
+function Creature:removeSummon(monster)
+	local summon = Monster(monster)
+	if not summon or summon:getMaster() ~= self then
+		return false
+	end
+
+	summon.setMaster(nil)
+	summon.setDropLoot(true)
+	summon.setSkillLoss(true)
+
+	return true
+end
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -413,7 +413,7 @@ void Creature::onRemoveCreature(Creature* creature, bool)
 	onCreatureDisappear(creature, true);
 	if (creature == this) {
 		if (master && !master->isRemoved()) {
-			master->removeSummon(this);
+			this->setMaster(nullptr);
 		}
 	} else if (isMapLoaded) {
 		if (creature->getPosition().z == getPosition().z) {
@@ -674,7 +674,7 @@ void Creature::onDeath()
 	death(lastHitCreature);
 
 	if (master) {
-		master->removeSummon(this);
+		this->setMaster(nullptr);
 	}
 
 	if (droppedCorpse) {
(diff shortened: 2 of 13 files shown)


src/creature.cpp
```diff
@@ -1121,24 +1121,17 @@ void Creature::onGainExperience(uint64_t gainExp, Creature* target)
...
-	if (cit != summons.end()) {
-		creature->setDropLoot(false);
-		creature->setLossSkill(true);
-		creature->setMaster(nullptr);
-		creature->decrementReferenceCounter();
-		summons.erase(cit);
+void Creature::setMaster(Creature* creature) {
+	master = creature;
+	if (creature) {
+		this->incrementReferenceCounter();
```
ranisalt: I think you don't need `this` here

src/creature.cpp
```diff
@@ -1121,24 +1121,17 @@ void Creature::onGainExperience(uint64_t gainExp, Creature* target)
...
+		creature->summons.push_back(this);
+	} else {
+		auto cit = std::find(summons.begin(), summons.end(), this);
+		if (cit != summons.end()) {
+			this->decrementReferenceCounter();
```
ranisalt: Same as above

src/creature.cpp
```diff
@@ -1121,24 +1121,17 @@ void Creature::onGainExperience(uint64_t gainExp, Creature* target)
...
-		creature->setMaster(nullptr);
-		creature->decrementReferenceCounter();
-		summons.erase(cit);
+void Creature::setMaster(Creature* creature) {
+	master = creature;
+	if (creature) {
+		this->incrementReferenceCounter();
+		creature->summons.push_back(this);
+	} else {
+		auto cit = std::find(summons.begin(), summons.end(), this);
```
ranisalt: Rename to `it` as it's not really const.

Mkalo: This is really wrong tho, why are you trying to find "this" in it's on summons container?

src/creature.cpp
Mkalo: You forgot to change Line 46, as setMaster is now responsible for decrementing. https://github.com/Kamenuvol/forgottenserver/blob/88a47bfe04dc8b9b08019f0afbe176e52f9cdbdb/src/creature.cpp#L46

Mkalo: Also this may break things when using setMaster(nullptr), if its counter is decremented to 0 it will be deleted and if there is places in the source code that it doesn't expect that it will break.

Kamenuvol:
I wonder if I can move setMaster to Monster since we can't have other type of creature as summon

src/creature.cpp
```diff
@@ -1121,25 +1120,26 @@ void Creature::onGainExperience(uint64_t gainExp, Creature* target)
...
-	creature->setDropLoot(false);
-	creature->setLossSkill(false);
-	creature->setMaster(this);
-	creature->incrementReferenceCounter();
-	summons.push_back(creature);
-}
+bool Creature::setMaster(Creature* newMaster) {
+	if (!newMaster && !master) {
+		return false;
+	}
```
ranisalt: Why is this? Both checks below will be skipped anyway.

src/monster.cpp
```diff
@@ -933,11 +933,13 @@ void Monster::onThinkDefense(uint32_t interval)
...
 				const Position& summonPos = getPosition();
-
-				addSummon(summon);
-
+				summon->setDropLoot(false);
+				summon->setLossSkill(false);
+				summon->setMaster(this);
 				if (!g_game.placeCreature(summon, summonPos, false, summonBlock.force)) {
-					removeSummon(summon);
+					summon->setDropLoot(false);
```
ranisalt: You're already setting drop loot to false just above.

src/creature.cpp
Kamenuvol: Because I need the option with both not exist at the same time in order to return false.
returning false means that what the scripter wanted to do has not worked

src/monster.cpp
```diff
@@ -933,11 +933,13 @@ void Monster::onThinkDefense(uint32_t interval)
 			Monster* summon = Monster::createMonster(summonBlock.name);
 			if (summon) {
 				const Position& summonPos = getPosition();
```
Kamenuvol: The purpose of this PR is not that, but I will check

src/creature.cpp
```diff
@@ -413,7 +412,7 @@ void Creature::onRemoveCreature(Creature* creature, bool)
 	onCreatureDisappear(creature, true);
 	if (creature == this) {
 		if (master && !master->isRemoved()) {
-			master->removeSummon(this);
+			this->setMaster(nullptr);
```
ranisalt: No need for `this->`

src/creature.cpp
```diff
@@ -674,7 +673,7 @@ void Creature::onDeath()
 	death(lastHitCreature);
 
 	if (master) {
-		master->removeSummon(this);
+		this->setMaster(nullptr);
```
ranisalt: No need for this->

src/monster.cpp
```diff
@@ -931,16 +931,15 @@ void Monster::onThinkDefense(uint32_t interval)
...
-				} else {
+			if (summon){
+				if (g_game.placeCreature(summon, getPosition(), false, summonBlock.force)) {
+					summon->setDropLoot(false);
+					summon->setLossSkill(false);
+					summon->setMaster(this);
 					g_game.addMagicEffect(getPosition(), CONST_ME_MAGIC_BLUE);
 					g_game.addMagicEffect(summon->getPosition(), CONST_ME_TELEPORT);
+				} else {
+					delete summon;
```
ranisalt: Won't this crash? I mean, maybe somewhere somewhat is storing this summon, as you placed it on the map.

src/monster.cpp
```diff
@@ -931,16 +931,15 @@ void Monster::onThinkDefense(uint32_t interval)
...
 			Monster* summon = Monster::createMonster(summonBlock.name);
-			if (summon) {
-				const Position& summonPos = getPosition();
-
-				addSummon(summon);
-
-				if (!g_game.placeCreature(summon, summonPos, false, summonBlock.force)) {
-					removeSummon(summon);
-				} else {
+			if (summon){
```
ranisalt: Space before brace

src/monster.cpp
Kamenuvol: placeCreature returns false if is not placed, so there is no reference, and as it was created just before, I see no problem on deleting it

Kamenuvol: btw, why do we still have a global g_game instead of a singleton with static methods?

ranisalt: Globals are shit and singletons are globals. It would be much better to hold a reference to `Game` wherever it's needed.

Kamenuvol:
May I move it to Monster class? And now it's ok?

ranisalt:
Looks good to me, but I can't really test it, I'll wait for more feedback.

Kamenuvol:
Now, I finally tested what I've done. It works well, I need only to adjust the functions of summon in Lua because summons are giving loot (xp probably too, but tested on God only).
I do not know if I change in this pull request or after accept it.

ranisalt:
If it broke in this PR, fix it here.

data/lib/core/creature.lua
```diff
@@ -84,3 +84,29 @@ function Creature:setItemOutfit(item, time)
...
 	return true
 end
+
+function Creature:addSummon(monster)
+	local summon = Monster(monster)
+	if not summon then
+		return false
+	end
+
+	summon.setMaster(self)
```
WibbenZ: summon:function()

src/luascript.cpp
```diff
@@ -6825,6 +6827,20 @@ int LuaScriptInterface::luaCreatureSetDropLoot(lua_State* L)
...
 
+int LuaScriptInterface::luaCreatureSetSkillLoss(lua_State* L)
+{
+	// creature:setSkillLoss(skillLoss)
+	Creature* creature = getUserdata<Creature>(L, 1);
+	if (creature) {
+		creature->setSkillLoss(getBoolean(L, 2));
+		pushBoolean(L, true);
+	}
+	else {
```
WibbenZ: Move the else up a line } else {

Kamenuvol:
Something very wrong happened in last commits. I've just summoned a demon skeleton and it started to attack me. 😆

data/lib/core/creature.lua
Kamenuvol: thanks, it was exactly the problem I said before

Kamenuvol:
I already have Convince Rune in Lua
https://github.com/Kamenuvol/forgottenserver/commit/08e38baef39c387a76ae294ea76ce01879be2fb6
But it depends on this pull request.

ranisalt:
Is this PR ready to merge?

Kamenuvol:
I think so

ranisalt:
So be it 👍

PR #2133, opened 2017-01-28
Fix pushMovement of creature:teleportTo

Riverlance:
Note: Ignore the movements.xml changes.


Mkalo:
I don't think that is an issue tho. It was correct, the pushMovement is going to have effect in internalTeleport pushMove parameter, that if is just to turn the character after teleporting when pushMove is false.

Riverlance:
Ooh, tested again and it worked.

But I found what I missed.
Well, I think is better when you teleport a creature with `creature:teleportTo(position, **false**)`, it **should not** update his outfit direction once the pushMovement is false meaning that will only "update" its position without move properly.

PR #2134, opened 2017-01-28
add new text color & renamed existing one

raymondtfr:
Vanilla now uses two different colors to differ from restoring mana/hit points.

The new text color used for healing/recovering hit points is: ``TEXTCOLOR_PASTELRED``
Original sample from vanilla:
pastel_red_small_254_101_101

I renamed ``TEXTCOLOR_PURPLE`` to the most approximated one, which is *electric purple*.

I found the names with help of websites such as [this](http://www.htmlcsscolor.com/), and a few others, and while some would suggest that the original sample could be **bittersweet (254, 111, 94)** I did more researches and realized that **pastel red (255, 105, 97)** is far more approximated to the **original (254, 101, 101)**.


ranisalt:
According to ColorHexa it's called [soft red](http://www.colorhexa.com/fe6565)

WibbenZ:
Compat? :>

raymondtfr:
Do you think it would also make sense to rename ``TEXTCOLOR_WHITE_EXP`` to ``TEXTCOLOR_WHITE``? because I don't think we should append a name to a text color just to point out we're using it as the color for the experience, they should only contain the color name ⁉️

raymondtfr:
Closing this since I updated another similar PR with these changes.

Issue #2122, opened 2017-01-23
[Question] - CONDITION_PARAM_SUBID


conde2:
We have two different SUBID for haste and strong haste, shouldn't they be the same ? So if you cast strong haste it will overwrite haste spell ?

raymondtfr:
No. It has to be different subid's for each different spell, so that you can identify which spell you cast.

conde2:
SUBID is for finding specific conditions when using getCondition and to not overwrite conditions, if you cast haste and strong haste you will have 2 conditions running at the same time, what is not meant to

raymondtfr:
I am pretty sure if a condition of the same type is added to a creature then it will overwrite the former, otherwise it would not make sense to have subids at all.

Yes, I meant *spell* as a spell that adds a condition to a creature.

raymondtfr:
@conde2 You're actually right here. I mislead the real purpose of subids... conditions with different subids can coexist, which would change the default behavior of spells, I'll make a PR to remove subids and let users decide if they want that or not.

PR #2136, opened 2017-01-30
Removed conditions subId's from potions script

raymondtfr:
This reverts the changes I made in (#1927) adding subIds to potions script (to avoid overlapping of conditions of the same type, keeping the same behavior before I introduced these changes).
#2122

### DIFF
--- a/data/actions/scripts/other/potions.lua
+++ b/data/actions/scripts/other/potions.lua
@@ -1,18 +1,15 @@
 local berserk = Condition(CONDITION_ATTRIBUTES)
-berserk:setParameter(CONDITION_PARAM_SUBID, 7)
 berserk:setParameter(CONDITION_PARAM_TICKS, 10 * 60 * 1000)
 berserk:setParameter(CONDITION_PARAM_SKILL_MELEE, 5)
 berserk:setParameter(CONDITION_PARAM_SKILL_SHIELD, -10)
 berserk:setParameter(CONDITION_PARAM_BUFF_SPELL, true)
 
 local mastermind = Condition(CONDITION_ATTRIBUTES)
-mastermind:setParameter(CONDITION_PARAM_SUBID, 8)
 mastermind:setParameter(CONDITION_PARAM_TICKS, 10 * 60 * 1000)
 mastermind:setParameter(CONDITION_PARAM_STAT_MAGICPOINTS, 3)
 mastermind:setParameter(CONDITION_PARAM_BUFF_SPELL, true)
 
 local bullseye = Condition(CONDITION_ATTRIBUTES)
-bullseye:setParameter(CONDITION_PARAM_SUBID, 9)
 bullseye:setParameter(CONDITION_PARAM_TICKS, 10 * 60 * 1000)
 bullseye:setParameter(CONDITION_PARAM_SKILL_DISTANCE, 5)
 bullseye:setParameter(CONDITION_PARAM_SKILL_SHIELD, -10)


conde2:
Potion does need subids because they can coexist

ranisalt:
Damn, this is getting me confused.

conde2:
@ranisalt 

Take this example:

  local A = Condition(CONDITION_ATTRIBUTES)
  A:setParameter(CONDITION_PARAM_SUBID, 1)
  A:setParameter(CONDITION_PARAM_TICKS, 1000)
  A:setParameter(CONDITION_PARAM_SKILL_MELEE, 5)

  local B = Condition(CONDITION_ATTRIBUTES)
  B:setParameter(CONDITION_PARAM_SUBID, 2)
  B:setParameter(CONDITION_PARAM_TICKS, 1000)
  B:setParameter(CONDITION_PARAM_SKILL_MELEE, 5)


Cast A then Cast B
or
Cast B then Cast A
Final Melee = +10


  local A = Condition(CONDITION_ATTRIBUTES)
  A:setParameter(CONDITION_PARAM_TICKS, 1000)
  A:setParameter(CONDITION_PARAM_SKILL_MELEE, 5)

  local B = Condition(CONDITION_ATTRIBUTES)
  B:setParameter(CONDITION_PARAM_TICKS, 1000)
  B:setParameter(CONDITION_PARAM_SKILL_MELEE, 7)

Cast A then Cast B
Final Melee = +7

Cast B and Cast A
Final Melee = +5

Kamenuvol:
7+5 = 13?

conde2:
Just a lack o attention, but you got it.

Kamenuvol:
I got it 😆 
There will be a lot of work to test all combinations of spells to see if they are cummulative or substitutive

raymondtfr:
@conde2 There is no need to add subIds for potions (they are defaulted to 0, meaning all conditions have the same ``subId`` if it is not set), and we shouldn't because if we do that they will overlap, which should not be the case.

@ranisalt Having **different** ``subId``'s allows the condition to overlap with another conditions of the same type. Just treat this as having the same default behavior (which had no problems) before I introduced these changes.

conde2:
@raymondtfr

You need subids for potions otherwise spells are going to overwrite the potions effect.

raymondtfr:
You are not getting it. If you have **different subids** they will overlap, having **equal subids** will not let them overlap.

subids are 0 by default, so they  won't overlap.
For example, prior to these changes, if you use a Berserk Potion and then you cast Blood Rage your skill will also be affected by this spell and it'll increase, but the duration fo the spell runs out, and the skill becomes your normal base skill plus the berserk potion.

If you have different subids (like now) and you use a berserk potion and then you cast Blood Rage spell the effect of Blood Rage in this case won't be applied.

If you have different subIds -- let's say for spells Recovery (utura) and Intense Recovery (utura gran) - and you first cast utura, and then utura gran - then both conditions will overlap and you will see your character healing over 20 and 40 hitpoints at the same time. But if you have default subId (0) then if you cast utura and then utura gran - the condition of utura gran will overwrite utura's one, and only the condition of utura gran will be applied,

conde2:
I know how CONDITIONS works, that's why I pointed this issue LOL. I'm just saying that as I remember in vanilla you can have both potion and spell effects overlapping. But I haven't tested in vanilla to be sure.

conde2:
Also this pull request still wrong since FOOD uses Condtion Regenaration, so after casting recovery you will break gain health/mana ticks

raymondtfr:
Just tested it, and it doesn't. Merging this PR will revert these changes I made back in the past and will keep the same default behavior as before, it is not wrong. If you want any of those conditions to act exactly like vanila, you'll have to it yourself (its the user job), and as I said before, you won't achieve that with subIds only, might as hell not even help you trying to achieve what're implying because I had already done such researches in private testservers, and it would only add useless code.

Kamenuvol:
And there is a question about monsters conditions too, when are poisoned by a scorpion, even if you are poisoned again by larva or snake, only the scorpion poison still counts. There should be a priority parameter also.

raymondtfr:

PR #2138, opened 2017-01-31
Reload talkaction shouldn't be displayed in chat

raymondtfr:
This keeps old behaviour of not showing words, params in game chat.


ranisalt:
Good eye! Talkactions still have this gotcha...

Issue #1959, opened 2016-10-23
Question: couple questions about server code


w19787:
Dear,

```
  i have couple questions about the server solution. if you can provide me some thoughts will be deeply appreciated!!
```
1. how can we make sure the quality of the server code? do we plan to develop unit test cases in future? what is the framework will be used?
2. in one of dicussion thread, i was told that this project will plan to remove the boost and other dependencies. Is this true? boost seems to be a solid soluiton, why do we plan to replace it?
3. The code of message exchange between server and client is using raw number instead of enum value. it is not helpful to read and maintain. is it any reason for this?

wgrriffel:
The only reason to drop boost from dependencies is that because is an additional step when using VS, also it is not been used that much since mutex, thread and shared_ptr have been added to STL, i think it will be replaced by STL after [c++17](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/n4612.pdf).

w19787:
how about boost asio? any plan to replace with other networking? i am not using VS, so i am fine with boost :)

wgrriffel:
take a look at #1759

ranisalt:
@w19787 ASIO can be used standalone, without depending on the entire boost library, but we still have a dependency on a lockfree stack and boost provides a good one, and there's not really much options out there.

As for 1, it would be very nice to have at least some regression tests to assure new features do not introduce bugs to old ones. I suggest gtest as it's easily integrable with cmake (example: https://github.com/ranisalt/hoist)

w19787:
thanks for all replies.  i don't have big picture of this project, but would like to add couple suggestions:

1. it will be very nice to add unit test framework
2. it might be good to use google protobuf for serializaiton/unserialization
3. ORM which could make our code base more clean
4. log system. boost seems to be a solid log system.

if you have different opinon, will apprieciated if you share it. it will help me to learn!!

ranisalt:
> ORM which could make our code base more clean

There's odb which seems great but would require a huge refactoring. I could do it, if wanted.

w19787:
Personal point of view, i like ORM. need inputs from project owners to plan whether this could be one of project road map.

Issue #1942, opened 2016-10-07
Question about serialization/unserialization


w19787:
just wonder any reason that forgotten server is not using protobuf or boost serialization solution. it seems to me a lot of extra work to create serialize/unserialze for each object.

can someone show me some thoughts? thanks in advance.

gugahoa:
One of the items on the roadmap is moving away from boost and other dependencies, it maybe because of that. 
Is those solution a lot better than the one being used?

kornholi:
What objects are you talking about? TFS is also older than protobuf.

w19787:
from my personal view, the current serialization/unserialization of item/items object is not easy to maintain. [i am new to this project, likely i am wrong]

it seems that more and more game design houses intend to use protobuf. Some analysis show it is very effective.

DerzeGuilherme:
It is difficult to change to a more efficient model. But it's a valid idea.

w19787:
@gugahoa why are we planning to move away from boost? what is plan for networking framework?

PR #2139, opened 2017-01-31
[wip] use protocol buffers for item (un)serialization

ranisalt:
This is work in progress as I still have to do lots of tests with complex scenarios, but so far it seems good.

I refactored item serialization and unserialization to the database to use Protocol Buffers instead of maintaining our own stream facilities. Protobuf is very optimized, has a nice prototype syntax and should allow our focus to keep on the game, easing the expansion of attributes and item types.

When writing a new specialized item type (e.g. depot lockers, house doors, teleports - basically everything that inherits `Item`) one should write `unserializeAttr(const tfs::Item&)` and `serializeAttr(tfs::Item*) const` functions, to respectively decode and encode the particularities of that item (e.g. teleports load/store destination coordinates).

From my early tests, it also reduced transaction times, allowing both faster loading and saving of items. Currently, it only works for database storing, as it would require refactoring even the binary formats for OTB and OTBM to work with Protobuf too.

Relates to #1942 and #1959, as suggested by @w19787

*code is not really large, it's mostly cmake modules to find deps*


ranisalt:
Why is Ubuntu so damn hard to install and use packages? On Arch it's just `pacman -S protobuf` and not a single thing fails.

wgrriffel:
@ranisalt it's supposed to work like that either, maybe cmake script or wrong package/version?

ranisalt:
omg it worked

kornholi:
Why not flatbuffers? Also I'm not sure how hard it will be to get this working on windows

ranisalt:
@kornholi didn't knew it yet. I guess the SDK would need to include the protobuf compiler for it to work on Windows.

ranisalt:
@kornholi it turns out `flatbuffers` isn't available in any Ubuntu distro nor PPA. I'll try to make a PPA with all TFS direct dependencies, but I'm not an Ubuntu user so it should take some time.

src/bed.cpp
```diff
@@ -67,18 +69,33 @@ Attr_ReadValue BedItem::readAttr(AttrTypes_t attr, PropStream& propStream)
 	return Item::readAttr(attr, propStream);
 }
 
-void BedItem::serializeAttr(PropWriteStream& propWriteStream) const
+bool BedItem::unserializeAttr(const tfs::Item& pbItem)
```
djarek: I think it would be nice if the serialization code were placed in a subnamespace of tfs instead - I plan to wrap our code into the tfs namespace, so I wouldn't want any collisions.

ranisalt: This is great. I'll put it under the `protobuf` or `flatbuffers` namespace later.

PR #2140, opened 2017-01-31
Storing TalkAction objects.

Mkalo:
It was marked as TODO, so...


ranisalt:
Thanks! All "lists of something" should hold instances instead of pointers actually.

I'll even replace with a vector as it's never mutated after loading.

http://xkcd.com/371/

Mkalo:
@ranisalt Hmm you didn't replace with a vector tho.

ranisalt:
I'll do it on master later.

PR #2143, opened 2017-01-31
[WIP] Store System

brunominervino:
TODO:
- [x] Split getMountByName()
- [x] Able to return disabled reason OnRender Method ([here](https://github.com/brunominervino/forgottenserver/blob/ebd6346b88795db0b98b25d15d64af3f33fc9733/src/protocolgame.cpp#L2129-L2132))
- [ ] Improve outfit function
- [ ] Set best method to block store inbox usage

Maybe, static.otland set default images for this feature?
[Store Images](https://mega.nz/#!NVhnCZbA!TxOdOzt2VLesrZrIuNX-2pbz9g8AW1O5cmiXmj4KRzc)


data/events/scripts/player.lua
```diff
@@ -86,6 +86,25 @@ function Player:onLookInShop(itemType, count)
 end
 
 function Player:onMoveItem(item, count, fromPosition, toPosition, fromCylinder, toCylinder)
+	-- Store Inbox
```
Mkalo: That here will not work for containers inside the store inbox.

data/store/scripts/outfits.lua
```diff
@@ -0,0 +1,79 @@
+-- Load outfits
+local outfits = {}
+
+for line in io.lines('data/XML/outfits.xml') do
```
Mkalo: This is ugly, it would be better to have a Game.getOutfits.

data/events/scripts/player.lua
wgrriffel: that is not the purpose of lua player events, it's a matter of customization, your system should not have any dependency on Lua api

src/enums.h
```diff
@@ -121,6 +121,32 @@ enum SpellGroup_t : uint8_t {
...
 };
 
+enum StoreOfferState_t : uint8_t {
+	STORE_OFFERSTATE_NONE = 0,
+	STORE_OFFERSTATE_NEW = 1,
+	STORE_OFFERSTATE_SALE = 2,
+	STORE_OFFERSTATE_TIMED = 3,
+};
+
+enum StoreOfferType_t : uint8_t {
```
wgrriffel: client do expect these offer types?

data/events/scripts/player.lua
brunominervino: Really, I didn't test this.

src/game.cpp
```diff
@@ -4919,6 +4921,92 @@ void Game::playerDebugAssert(uint32_t playerId, const std::string& assertLine, c
 	}
 }
 
+void Game::playerPurchaseStoreOffer(uint32_t playerId, uint32_t offerId, const std::string& param) {
+	Player* player = getPlayerByID(playerId);
+	if (!player) {
+		return;
+	}
+
+	auto offer = g_store->getOfferById(offerId);
```
wgrriffel: in this case and below auto typing is confusing, at least to me, explicit it as int is better imo

data/events/scripts/player.lua
Mkalo: There is also the problem of equipping the item, lets say you ahve a shield in your hand and you buy another shield in the store, if you move the shield in the store inbox it will replace with your equipped one and that shouldn't happen.

data/store/scripts/outfits.lua
brunominervino: :sweat:
Help me with this, I'm sharing my time at work to do some contributions, is very little time :(

src/game.cpp
```diff
@@ -5624,7 +5754,7 @@ bool Game::reload(ReloadTypes_t reloadType)
 				std::cout << "[Error - Game::reload] Failed to reload spells." << std::endl;
 				std::terminate();
 				return false;
-			} else if (!g_monsters.reload()) {
+			} if (!g_monsters.reload()) {
```
wgrriffel: why?

data/events/scripts/player.lua
brunominervino: The reality is that I took advantage this script from an otland post and did not do all the testing if you have something ready to share with us, I'll appreciate it. :scream:

src/luascript.cpp
```diff
@@ -8742,10 +8764,20 @@ int LuaScriptInterface::luaPlayerSendOutfitWindow(lua_State* L)
 
 int LuaScriptInterface::luaPlayerAddMount(lua_State* L)
 {
-	// player:addMount(mountId)
+	// player:addMount(mountId or mountName)
```
wgrriffel: split these mount stuff in two pr.

src/game.cpp
brunominervino: Resharper make this change, I didn't even see

src/enums.h
brunominervino: https://github.com/brunominervino/forgottenserver/blob/03e1fa4ef45cb58468b5aa03f869f3d4644d906e/src/protocolgame.cpp#L2177

src/luascript.cpp
brunominervino: It's really necessary?

data/store/scripts/outfits.lua
```diff
@@ -0,0 +1,79 @@
...
+    local sex = tonumber(line:match('type.-=[\'"](%d-)[\'"]'))
+    local lookType = tonumber(line:match('looktype.-=[\'"](.-)[\'"]'))
+    local name = line:match('name.-=[\'"](.-)[\'"]')
+    local enabled = line:match('enabled.-=[\'"](.-)[\'"]') == 'yes'
+    local unlocked = line:match('unlocked.-=[\'"](.-)[\'"]') == 'yes'
+
+    if enabled then
+        local outfit = outfits[name:lower()]
```
WibbenZ: Spaces in this file

data/store/scripts/outfits.lua
WibbenZ: Or even better an Outfit class so we don't have to send huge tables when we might just want the sex of a citizen outfit.

src/luascript.cpp
wgrriffel: imo, yes, it's better for review discussion, and for source control

WibbenZ: Since you are changing a function I would say yes.
Mark ones told me to do it, since if we get a problem related to this function we should be able to search for "mount(s)" and not have to check every PR.

src/protocolgame.cpp
```diff
@@ -1972,6 +2038,161 @@ void ProtocolGame::sendMarketDetail(uint16_t itemId)
...
+	msg.addByte(0xE1);
+	msg.add<uint32_t>(offerId);
+	msg.addByte(offerType);
+
+	writeToOutputBuffer(msg);
+}
+
+void ProtocolGame::sendCoinBalance() {
+	NetworkMessage msg;
+	msg.addByte(0xF2);
```
WibbenZ: Seeing to the amount of bytes here, you should write down what they are.
Also you should make a new line for each new header that you send.

src/store.cpp
```diff
@@ -0,0 +1,258 @@
...
+			icons = explodeString(attr.as_string(), ";");
+		}
+
+		categories.emplace_back(name, state, description, icons);
+		StoreCategory& category = categories.back();
+
+		for (auto offerNode : categoryNode.children()) {
+			StoreOffer& offer = category.createOffer(runningId++);
+
+			if ((attr = offerNode.attribute("name"))) {
```
WibbenZ: You don't need 2 parentheses here, if (attr = ..()) is enough IMO

src/store.cpp
```diff
@@ -0,0 +1,258 @@
...
+				}
+
+				std::string script = attr.as_string();
+				if (scriptInterface->loadFile("data/store/scripts/" + script) == 0) {
+					offer.scriptInterface = scriptInterface.get();
+					offer.renderEvent = scriptInterface->getEvent("onRender");
+					offer.buyEvent = scriptInterface->getEvent("onBuy");
+				} else {
+					std::cout << "[Warning - Store::loadFromXml] Can not load script: " << script << std::endl;
+					std::cout << scriptInterface->getLastLuaError() << std::endl;
```
WibbenZ: If the reload fails you should return false, not true.
Otherwise the reload function will tell the player that the reload was successful.

conde2:
Amazing work @brunominervino  !

Thanks

src/luascript.cpp
brunominervino: #2144

data/events/scripts/player.lua
WibbenZ: Wrong, events was created so we had a place to put the "old" C++ code when we moved it.
Idk why ppl seem to think that we should have everything in C++ when we can place it in Lua.

So IMO nothing wrong there.

vankk:
The bytes is currently updated for our version, 1097-1098?

src/store.cpp
WibbenZ: @brunominervino Apparently travis didn't like that so add the parentheses back again, sorri :p

brunominervino:
@conde2 thank you bro!

@vankk yes!!

PrinterLUA:
Before i review any code, you should consider to remove the word "Tibia" since it's a trademark word.

config.lua
```diff
@@ -29,6 +29,10 @@ statusTimeout = 5000
 replaceKickOnLogin = true
 maxPacketsPerSecond = 25
 
+-- Store configuration
+storeImagesUrl = "http://static.tibia.com/images/store/"
```
Tarliton: this might be a no-no

WibbenZ: Ye change it to the ip located in config or something but keep the /images/.. if someone wants to use that URL.

vankk: @brunominervino should explain somewhere that in the folder _/images/store/_ needs **have a folder named 64,** and the images needs to be **64x64** as well.

brunominervino: @vankk when this pr are ready to merge I'll create a documentation on Wiki.

brunominervino:
@PrinterLUA I completely forgot this. Already changed.

Kamenuvol:
Now, there are some funny names there 😆

data/events/scripts/player.lua
wgrriffel: @WibbenZ server working stuff should stay in binaries and left lua as a easily customization/server setup or changing the default behaviour, never server working stuff, datapack are supossed to be an example for people build their own datapack, which means server should not have dependency on datapack scripts, otherwise datapack is not an example for a datapack but a server working stuff, and it force people to use default tfs datapack.

brunominervino:
@Kamenuvol have any suggest? I'm not creative hahaha

config.lua
```diff
@@ -29,6 +29,10 @@ statusTimeout = 5000
 replaceKickOnLogin = true
 maxPacketsPerSecond = 25
 
+-- Store configuration
+storeImagesUrl = "http://static.tibia.solutions/store/"
```
WibbenZ: Tibia is the word you should remove, not .com.
Tbh for now just use the ip variable and then /store/ or smth xD

Remember to add http://

src/configmanager.cpp
```diff
@@ -140,6 +140,7 @@ bool ConfigManager::load()
 	string[LOCATION] = getGlobalString(L, "location", "");
 	string[MOTD] = getGlobalString(L, "motd", "");
 	string[WORLD_TYPE] = getGlobalString(L, "worldType", "pvp");
+	string[STORE_IMAGES_URL] = getGlobalString(L, "storeImagesUrl", "http://static.tibia.solutions/images/store/");
```
WibbenZ: Same here

brunominervino: Done!

src/luascript.cpp
vankk: Should remove it from this pull request since you already created one for it.

brunominervino: Until not it is approved, if I remove it, it will give an error.

data/events/scripts/player.lua
WibbenZ: I get what you are saying but we should keep as much in Lua as posible.
Having some codes in Lua can be faster than having them in c++.

Another reason for that is so we can modify things without needing to recompile everytime.

And gl using a 1.1 datapack with 1.3 or even 1.2 without doing changes.
Thats a thing we need to do, either as new prs are approved or once every year or so.

Either update your server files or stay where you are.

data/store/scripts/outfits.lua
```diff
@@ -0,0 +1,79 @@
+-- Load outfits
+local outfits = {}
+
+for line in io.lines('data/XML/outfits.xml') do
+    local sex = tonumber(line:match('type.-=[\'"](%d-)[\'"]'))
+    local lookType = tonumber(line:match('looktype.-=[\'"](.-)[\'"]'))
+    local name = line:match('name.-=[\'"](.-)[\'"]')
+    local enabled = line:match('enabled.-=[\'"](.-)[\'"]') == 'yes'
+    local unlocked = line:match('unlocked.-=[\'"](.-)[\'"]') == 'yes'
```
ranisalt: Regexing a XML file: please DON'T YOU EVER DARE.

[Here's why](http://stackoverflow.com/a/1732454/1940360).

data/events/scripts/player.lua
wgrriffel: > we should keep as much in Lua as posible.

i don't think so, imo we should increase lua coverage over server default behaviour, not move server to lua, it would be a disaster, we really shouldn't
> Having some codes in Lua can be faster than having them in c++.

even once in a blue moon cases, isn't a matter of server performance
> And gl using a 1.1 datapack with 1.3 or even 1.2 without doing changes.

that's why wiki provide server version for lua implementation, you're not suppose to use datapack 1.1 on 1.2, but any 1.2 datapack should be fully compatible with 1.2 server

WibbenZ: Well that is what we have and most likely? are still aiming for.
Im not saying we should move everything, but as much as posible.
There really is no point to keep useless things that we might want to change in C++ when we can even make the code run faster by having it in Lua, no matter what your opinons are :p 

Correct, the source we are writing right now (1.3) should not be 100% compat with 1.2.
There for I don't understand why you wrote;
```
never server working stuff, datapack are supossed to be an example for people build their own datapack, which means server should not have dependency on datapack scripts, otherwise datapack is not an example for a datapack but a server working stuff, and it force people to use default tfs datapack.
```

You should have to update your scripts, in no way should we care that a datapack has to be updated in order to implement new things.

How ever I think it was Mark who said when I wanted to update the gender tags in outfits.xml that we should only update things that we feel needed or that we can have a better use out of it.
But that still tells us that we can and should move and rename things to / in Lua.

wgrriffel: > There for I don't understand why you wrote;

i said that because of your quote `And gl using a 1.1 datapack with 1.3 or even 1.2 without doing changes.`, which i never said all datapacks should be compatible with server.
> You should have to update your scripts, in no way should we care that a datapack has to be updated in order to implement new things.

as i said then datapack drop the status a "showcase" to user and become a dependency,(i mean scripts in datapack, not configuration files) nonetheless, how you ensure tfs new scripts is fully compatible with user custom script systems?

WibbenZ: Ehm I wrote that becuase of what you said, just read above haha

Not sure what you mean but imo a datapack should be set to a version ex 1.3 and you should have to use 1.3 for that datapack(talking about stable releases, latest atm is 1.2)
If we wanna get compat we should not do any new Lua changes, revert everything we have moved to Lua, updates to XML files etc that way you can use datapacks from 1.1 to 1.3.
If we revert metatags we can include 0.x aswell as 1.0.

Is that what you want?
A datapack should have a version: 1.2
If you wanna use 1.3 you have to update the files that we changed during 1.3 or stay with 1.2.

wgrriffel: @WibbenZ what i meant, a user that has a custom datapack with according to lua api version (interfaces and methods), its datapack may work without any imposed script by server, such this one

WibbenZ: It may but we shouldn't care about that when it comes to new things, as I said a 1.2 datapack isent supposed to work with any other version, if it does great but we already created compat issues with 1.2 so there is no point in keeping things back when they anyways have to do updates.

conde2: This is a new functionality, doesn't matter which data pack version you have. You will need to change it !! Guys I think it is better to start a discussion about this in OTLand, every time someone answers here all participants are notified.

wgrriffel: backwards compatibility is not the point here, show me any other server system that are breakable by modifing in a script, there is none because it's not design project have been doing, thats the main point.

conde2: I do agree with @wgrriffel in this point, if you can break the system/functionality changing the script, then it is better to keep it at sources, but again, it is debatable in some cases.

WibbenZ: But you can break any server by modifying a script.
The amount of damage might differ but you could crash 1.1+ by not knowing how to handle userdata values.

You can also break 0.x servers via Lua scripts.
The point is to make people learn, we cant make everything noob proof without having a billion checks for everything the server handles.

I agree that this has gotten out of hand and it would be better to talk abt this on otland to see what ppl think.

wgrriffel: > But you can break any server by modifying a script.

what you meant was any tfs right? out there is more gaming servers that use scripting api, for example dota2 server lua api, it's safe enough to valve host any mod developed by community on their own machine, that's a good example of a well done lua api.
have a look at developer.valvesoftware.com
> The amount of damage might differ but you could crash 1.1+ by not knowing how to handle userdata values.

when user turn off script security conversions, server assume he knows what he is doing, btw i think this could be changed, out of topic tho, nothing going to arg it here.

> You can also break 0.x servers via Lua scripts.

idk why you are referring to unsupported and legacy version, if you want my opnion 0.x was horrible, bad design and kludges over whole code.

>The point is to make people learn, we cant make everything noob proof without having a billion checks for everything the server handles.

it's not being 'noob proof' but beeing stable.

@conde2 

> every time someone answers here all participants are notified.

sorry for the inconvenience, but it may be discussed here imo, it's about project, to avoid this kind of discussion in futher i think someone with merge access should write a contribution guide, btw there is a left panel that you can always unsubscribe from PRs you're not interested in.

WibbenZ: Tfs is a server I hope you know that? ;)
I know alot of games allow public codes, vut they have alot of checks for that code insted or review / check them.
A thing we can't do.

I could crash 1.x with the security checks enabled, and you also seem to think 1.1 is still supported?
1.2 and 1.3 are the only supported version as of right now.
But I agree that 0.x is a complete failure compared to 1.x.

But they are the same thing, either you know what you are doing or your a noob.
A person sith experience that dosnt know what a function does would look it up before using it.
While a noob would hope for the best.

As I said TFS should not be noob proof as we would have to modify every function and class pretty much = waste & more cpp overhead.
Let ppl learn by their mistakes, I know I did.

wgrriffel: > Tfs is a server I hope you know that? ;)
wtf? reason why i wrote that is that you said ANY SERVER (any server include tfs, or dota server and whos know how many more) so gl breaking valve's servers.

i think you still not get the point, the point of my comments, i'm repeating myself, but ok look, you agree that player walking event shouldn't stop working because of bad stepin/stepout event? so why store inbox should?

WibbenZ: I semi agree with you, as I have said, if a movement script bugged that the player now uses it should bug.
If there is a problem with the store inbox it should bug.

Having 50% of a logic working dosnt seem like a good ide to me.
But it might be better if we ask @djarek and @ranisalt

ranisalt: @wgrriffel not pending to any side, but comparing to Dota is too shallow here. TFS IS meant to be MUCH more flexible, as it does not accept any user custom script as Dota does (as you mentioned, I don't play it), so all scripts are to be validated by the server owner.

That said, I'd like more of the core moved to Lua even though it's easily "crashable" by a bad scripter. Store inbox being a Lua event would be easy to e.g. notify a player if something happens, or add more/remove some checks. To me, it's a very valid tradeoff.

You guys should open a new issue to discuss that in a more generic way, this review is getting way too large and hard to follow.

wgrriffel: @ranisalt you're mistaken, it's a fair comparation, valve distribute their server same as tfs, you can configure it to have your own server, ofc they don't provide source code for engine, whici means you can have a running dota server up running and connect throught official client, just like counter-strike 1.6 if you remember, then user has to validade their scripts as well, my point was they trust pretty much in their lua api, they provide their host with runnign server to your scripting, that what i meant, anyway it gonne too off topic now, the problem never was "crashable" but without onMoveItem event the store will not work properly, it's my problem with that. tho im ok with a new event to onMoveStoreInbox, as i said before user should be able to change default behaviour, whats the point of engine rely on scripting? that's awful.

Mkalo: I wanna also point out the event Player:onLook, this same discussion could be applied to it but in the end it was moved to Lua so I think this should be in Lua too.

brunominervino:
I'm really compromised to release this new feature. But, I do not have the time I would like to dedicate myself for this improvements and to test them.
I would like the help of all of you who share your knowledge with this repository, I believe that there is not much to finish.
A TODO list has been inserted in the description, I'm open to suggestions.

Thank you guys.

data/events/scripts/player.lua
ninjalulz: I would probably opt for a source modification here instead of relying on 'onMoveItem'. An example would be to create a new class called 'StoreInbox' which mimics the behavior of the normal Inbox. It wouldn't let the player to put item(s) inside the store inbox, nor its child containers, and even replacing shields wouldn't work.

ninjalulz:
I think you should rename 'onRender' to something else, and perhaps even change the data type (from boolean to string).

ranisalt:
I'll cleanup code from mounts as soon as I get home

brunominervino:
@ninjalulz I thought of something like that, `if (value != null)` the offer is disabled and takes the string as reason, right?

data/events/scripts/player.lua
brunominervino: Right, good idea.

Mkalo:
What about splitting the changes for the store inbox functionality? (moving it to the source like @ninjalulz mentioned)

brunominervino:
@Mkalo I'm agree with this.

brunominervino:
Now, reason will display on details:
Disabled Reason

WibbenZ:
Im not that familiar with the store system but shouldnt they just be grayd out if you already bought the product?

brunominervino:
@WibbenZ is working:
image

src/game.cpp
```diff
@@ -4919,6 +4921,92 @@ void Game::playerDebugAssert(uint32_t playerId, const std::string& assertLine, c
...
+	std::string description(player->getName() + " transferred to " + recipient);
+
+	IOAccount::addCoins(player->getAccount(), -static_cast<int32_t>(amount));
+	g_store->onTransactionCompleted(player->getAccount(), -static_cast<int32_t>(amount), description);
+
+	IOAccount::addCoins(tmpPlayer->getAccount(), amount);
+	g_store->onTransactionCompleted(tmpPlayer->getAccount(), amount, description);
+
+	if (tmpPlayer->isOffline()) {
+		delete tmpPlayer;
```
joseluis2g: player should be saved here, or am I wrong?

brunominervino: Why? If player is offline don't have any data to save. If he is online, when logout, he is be saved.

joseluis2g: Just saying, saw the game market code after reading the ninja comment and when you receive something on market if you offline then you will be saved.

joseluis2g: @brunominervino 
Also coins are transfered so there yes there is data to save.

brunominervino: @joseluis2g you're right, thanks for point this.

ranisalt:
You definitely should base it on #2010 for serving the images.

brunominervino:
Anynone can confirm in what version "new" flag has implemented?
Look:
image

src/creature.h
```diff
@@ -43,9 +43,10 @@ enum slots_t : uint8_t {
 	CONST_SLOT_FEET = 8,
 	CONST_SLOT_RING = 9,
 	CONST_SLOT_AMMO = 10,
+	CONST_SLOT_STORE_INBOX = 11,
 
 	CONST_SLOT_FIRST = CONST_SLOT_HEAD,
-	CONST_SLOT_LAST = CONST_SLOT_AMMO,
+	CONST_SLOT_LAST = CONST_SLOT_STORE_INBOX,
```
ranisalt: Wait, no. Slots are meant for body dressing, it should not be touched.

securmk:
@brunominervino Hi, I have an issue. Store opens but when I press a category such as Items or mounts my char logs out. It doesn't debug client, doesn't show any error on server console nor does anything. Just kick my char out and I return to char list. Also I noted that when I transfer coins and the "you have successfully gifted tibia coins" message the client does the same thing, kicks me out. Last error I noticed is that the transaction type doesn't work. It should send something like this "account_id, type, coins, description, timestamp" But it keeps sending this "account_id,,coins, description, timestamp" I downloaded and compiled, nothing was modified by me. Is there any way to solve these problems?

src/creature.h
Mkalo: But the store inbox is a slot. That's how tibia made it.

brunominervino: @ranisalt how @Mkalo told. Store Inbox is 11th slot. Works like a normal backpack with some conditions.

brunominervino:
@securmk I'll do some tests today to improvement the system.

securmk:
@brunominervino Would you send me or upload a binary with store? I'm not using my computer (can't compile) and I would like to test this system. Thanks in advance.

joseluis2g:
@brunominervino take a look
http://www.tibia.com/news/?subtopic=newsarchive&id=3681&fbegind=7&fbeginm=1&fbeginy=2015&fendd=6&fendm=2&fendy=2017&flist=11111111

vankk:
I think that we should care only with the XP boost. The gold poncher and gold converter is an item that can be created by scripters by using onMoveItem.

securmk:
I think I can work on the gold pouch and gold converter, but I just can't compile this server because I'm not using my computer and can't download compiler. Would someone send me the binary so I can try to make it?

src/creature.h
TheSumm: Just like the "Purse" previously, the store inbox is a slot. It is hidden (so that the player doesn't think he can move it) and can be opened through a button.

brunominervino:
Now working:
image

data/store/scripts/templeteleport.lua
```diff
@@ -0,0 +1,23 @@
...
+	local templePosition = player:getTown():getTemplePosition()
+	if player:getPosition():getDistance(templePosition) < MIN_DISTANCE then
+		return false, "You can not purchase this transportation service this close to your temple."
+	end
+
+	return true
+end
+
+function onBuy(player, offer)
+	player:teleportTo(templePosition)
```
securmk: Hi, I was testing the system and it gives me an error on this line when purchasing a temple teleport. Console error is here:

`Lua Script Error: [Store Interface] 
data/store/scripts/templeteleport.lua:onBuy
attempt to index a nil value
stack traceback:
	[C]: at 0x010851cea0
	[C]: in function 'teleportTo'
	data/store/scripts/templeteleport.lua:21: in function <data/store/scripts/templeteleport.lua:20>`

src/enums.h
securmk: I think there's an  error with type 2. It debugs my client. Is setting state to "2" enough? Maybe I need to add something else to the xml

data/store/scripts/templeteleport.lua
brunominervino: Tested here and all working...

raymondtfr: I have began testing this system and this is something that also happened to me when I bought this.

ninjalulz: Replace **templePosition** (the local variable was declared in the onRender callback, which makes it inaccessible) with **player:getTown():getTemplePosition()**.

brunominervino:
I forgort to add on change onRender method, thanks! I'll fix on next
commit.

Em 7 de fev de 2017 7:30 AM, "ninjalulz" <notifications@github.com>
escreveu:

data/store/scripts/templeteleport.lua
brunominervino: I forgot to fix this when I have update onRender method. Thank you guys.

brunominervino: @ninjalulz done. Thank you bro.

joseluis2g: I think the changes that @raymondtfr writed here are the most important, you should change the code to the new tfs code-style

securmk:
@brunominervino I found this in the sources:

`msg.add<uint16_t>(100); // base xp gain rate
msg.add<uint16_t>(0); // xp voucher
msg.add<uint16_t>(0); // low level bonus
msg.add<uint16_t>(0); // xp boost
msg.add<uint16_t>(100); // stamina multiplier (100 = x1.0)`

and 

`msg.add<uint16_t>(0); // xp boost time (seconds)
msg.addByte(0); // enables exp boost in the store`

is the xp boost already programmed in the sources?

MatheusGrilo:
Hey, just saying that if you add an item to sell without image, the client crashes after you try to "go to next image",,, Look:
Error
(I don't know how to copy the stuffs on "error report", lol)
Part 1
Part 2

Mkalo:
@MatheusGrilo There is fix for this, you should have images for all offers, you could create a transparent image to use if you'd like having offers without an image.

brunominervino:
@MatheusGrilo how pointed by @Mkalo, all offers should have images available.

MatheusGrilo:
@Mkalo I'm just saying that a var can be added like
```
if(empty(icons))
icons="noimage.png"
```
(ofc it's just an example)

brunominervino:
@MatheusGrilo all images generate cache, it's not a good option, since all offers need a image. But, you can use setting this image on xml of your own store.

src/creature.h
SavageCrown: its need to be a slot because you can use the items/potions stored there (when you bought on the store)

SavageCrown:
Found a bug.
ex: 
player A, have 500 coins. 
Player B, have 0 coins.
player A, transfer 500 coins to player B
player A, on the store still the balance of 500coins until you close and open the store again.
player B, get the 500coins. 
player A, can transfer the 500 coins again. (10 times) 
Player A, close and open the store.
Player A, balance of tibia coins -5000
Player B, balance of tibia coins 5000

[https://fotos.subefotos.com/3f6c4527a3b3cec8aacfcfd0ab8e4de5o.png](img)

joseluis2g:
@SavageCrown thats because the code is updating the coin balance when you do some kind of transaction, you can easily add it.

SavageCrown:
like you said using this func at the end of the parseTransferCoins
updateCoinBalance();

brunominervino:
I'll back to do changes on store system in few days. Now I'm very busy on my personal work. I'm apologize for can't help you now!

GoularPink:
If player no have freecapacity and buy, the item disappear and discount money

src/protocolgame.cpp
```diff
@@ -1977,6 +2043,161 @@ void ProtocolGame::sendMarketDetail(uint16_t itemId)
...
+
+void ProtocolGame::updateCoinBalance() {
+	NetworkMessage msg;
+	msg.addByte(0xF2);
+	msg.addByte(0x00);
+
+	writeToOutputBuffer(msg);
+
+	g_dispatcher.addTask(
+		createTask(std::bind([](ProtocolGame_ptr client) {
```
pzaj2: Hey guys, I'm not very familiar with The forgotten server structure, but has been given a "task" to fix this system you work on here. I just went through pieces of code and I wonder if it's such a good idea to pass ProtocolGame_ptr to the dispatcher??

djarek: @pzaj2 
Not sure if it's what you're asking, but only a shared_ptr owning the ProtocolGame object is passed here.

@brunominervino
std::bind call is redundant in this line. Drop it and just capture the shared_ptr in the lambda

drakylucas:
for who want to use it even before the merge, it's working but there is a bug (https://github.com/otland/forgottenserver/pull/2143#issuecomment-287465755)

to fix it, I changed 
`void Player::sendStorePurchaseCompleted(const std::string& message) {
	if (client) {
		client->sendStorePurchaseCompleted(message);
	}
}`

to

`void Player::sendStorePurchaseCompleted(const std::string& message) {
	if (client) {
		client->sendStorePurchaseCompleted(message);
		client->updateCoinBalance();
	}
}`


I need to do some tests yet, but so far it's ok.

src/enums.h
infister: According to OtClient:

```
int price = msg->getU32();
        int highlightState = msg->getU8();
        if(highlightState == 2 && g_game.getFeature(Otc::GameIngameStoreHighlights) && g_game.getClientVersion() >= 1097) {
            int saleValidUntilTimestamp = msg->getU32();
            int basePrice = msg->getU32();
        }
```

Znote:
Might be time to resurrect this soon, preferably in a new and fresh pull request.

brunominervino:
I agree with you @Znote

andersonfaaria:
@brunominervino 

> * Set best method to block store inbox usage

This worked fine here, please change the code and I think this can be merged already
image

FelypeBastos:
Hey guys, pretty nice job here! I followed all the steps but ingame I just see Fetchimg items and nothing appear. Categories are loaded without problems. Any advice how I manage to fix this?

DamianZ96:
There is also a bug where if player doesn't have enough cap they will not get the item in their store inbox, but points will be deducted. It can be fixed by simply extending data/store/scripts/items.lua with the following,

	if player:getFreeCapacity() < itemType:getWeight() then
		player:sendStoreError(STORE_ERROR_PURCHASE, "You do not have enough capacity to purchase this item.")
		return false
	end

lucasgrizante:
This pull request has already 5 years of existence.
Whats is pending? Tfs already moved to tibia 12 and we still don't have the native shop system

EPuncker:
> Whats is pending?

someone willing to open a new PR and finish what is left, specially the requested changes 😅

joseluis2g:
I can work on this

omarcopires:
If this request was a child, this person was already going to college.

PR #2144, opened 2017-02-01
Mounts by Name

brunominervino:


src/luascript.cpp
```diff
@@ -8740,43 +8740,65 @@ int LuaScriptInterface::luaPlayerSendOutfitWindow(lua_State* L)
 	return 1;
 }
 
-int LuaScriptInterface::luaPlayerAddMount(lua_State* L)
-{
-	// player:addMount(mountId)
+int LuaScriptInterface::luaPlayerAddMount(lua_State* L) {
+	// player:addMount(mountId or mountName)
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
```
ranisalt:
```cpp
if (!player) {
    lua_pushnil(L);
    return 1;
}
```

And dedent 1 level the rest.

src/luascript.cpp
```diff
@@ -8740,43 +8740,65 @@ int LuaScriptInterface::luaPlayerSendOutfitWindow(lua_State* L)
...
 	return 1;
 }
 
-int LuaScriptInterface::luaPlayerRemoveMount(lua_State* L)
-{
-	// player:removeMount(mountId)
+int LuaScriptInterface::luaPlayerRemoveMount(lua_State* L) {
+	// player:removeMount(mountId or mountName)
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
```
ranisalt: Same as above.

src/mounts.cpp
```diff
@@ -61,6 +63,15 @@ Mount* Mounts::getMountByID(uint8_t id)
 	return it != mounts.end() ? &*it : nullptr;
 }
 
+Mount* Mounts::getMountByName(const std::string& name) {
+	auto it = std::find_if(mounts.begin(), mounts.end(), [&name](const Mount& mount) {
+		return boost::iequals(name, mount.name);
```
ranisalt: `strcasecmp` isn't enough here? Also a range for would fit better here.

brunominervino:
@ranisalt can you confirm this change to `strcasecmp`? I can't test this now.

src/mounts.cpp
```diff
@@ -65,7 +63,7 @@ Mount* Mounts::getMountByID(uint8_t id)
 
 Mount* Mounts::getMountByName(const std::string& name) {
 	auto it = std::find_if(mounts.begin(), mounts.end(), [&name](const Mount& mount) {
-		return boost::iequals(name, mount.name);
+		return strcasecmp(name.c_str(), mount.name.c_str());
```
ranisalt: `strcasecmp` returns 0 if strings are equal, so you need to use `return strcasecmp(...) == 0`.

Yes, glibc sucks a lot.

brunominervino: I have read docs before and still forgot this, hahaha.
Done @ranisalt.

ranisalt:
Sweet. Will merge as soon as someone else approves.

jo3bingham:
LGTM. Though, I agree with @ranisalt that a range-based for loop would be better. It should also be applied to the other get methods in this class.
```
Mount* Mounts::getMountByName(const std::string& name)
{
	auto mountName = name.c_str();
	for (auto& it : mounts) {
		if (strcasecmp(mountName, it.name.c_str()) == 0) {
			return &it;
		}
	}

	return nullptr;
}
```

src/luascript.cpp
```diff
@@ -8740,43 +8740,67 @@ int LuaScriptInterface::luaPlayerSendOutfitWindow(lua_State* L)
...
-		uint8_t mountId = getNumber<uint8_t>(L, 2);
-		pushBoolean(L, player->tameMount(mountId));
-	} else {
+	if (!player) {
 		lua_pushnil(L);
+		return 1;
 	}
+
+	uint8_t mountId;
+	if (isString(L, 2)) {
```
Mkalo: Did you test this? This is making the method not work for number as isString uses lua_isstring and
```cpp
int lua_isstring (lua_State *L, int index);

Returns 1 if the value at the given acceptable index is a string or a number (which is always convertible to a string), and 0 otherwise
```

src/luascript.cpp
```diff
@@ -8740,43 +8740,67 @@ int LuaScriptInterface::luaPlayerSendOutfitWindow(lua_State* L)
...
-		uint8_t mountId = getNumber<uint8_t>(L, 2);
-		pushBoolean(L, player->untameMount(mountId));
-	} else {
+	if (!player) {
 		lua_pushnil(L);
+		return 1;
 	}
+
+	uint8_t mountId;
+	if (isString(L, 2)) {
```
Mkalo: Same as above.

src/luascript.cpp
```diff
@@ -8740,43 +8740,67 @@ int LuaScriptInterface::luaPlayerSendOutfitWindow(lua_State* L)
...
 	const Player* player = getUserdata<const Player>(L, 1);
 	if (!player) {
 		lua_pushnil(L);
 		return 1;
 	}
 
-	uint8_t mountId = getNumber<uint8_t>(L, 2);
-	Mount* mount = g_game.mounts.getMountByID(mountId);
+	Mount* mount = nullptr;
+	if (isString(L, 2)) {
```
Mkalo: Same as above

src/luascript.cpp
ranisalt: Yes, he should test for number instead, and just assume it's a string otherwise.

brunominervino:
@jo3bingham thank you!

@ranisalt @Mkalo Now, first check if is a number and assume it's a string otherwise.

ranisalt:
Thanks and welcome to the repository 😉 I'm following your progress and dedication closely.

brunominervino:
@ranisalt Thank you bro.

PR #2145, opened 2017-02-01
Fix memory leak in talkaction

Mkalo:
That was my mistake when changing it to list of objs instead of pointers.

### DIFF
--- a/src/talkaction.cpp
+++ b/src/talkaction.cpp
@@ -61,7 +61,9 @@ Event* TalkActions::getEvent(const std::string& nodeName)
 
 bool TalkActions::registerEvent(Event* event, const pugi::xml_node&)
 {
-	talkActions.push_front(*static_cast<TalkAction*>(event)); // event is guaranteed to be a TalkAction
+	TalkAction* talkAction = static_cast<TalkAction*>(event) // event is guaranteed to be a TalkAction
+	talkActions.push_front(*talkAction);
+	delete talkAction;
 	return true;
 }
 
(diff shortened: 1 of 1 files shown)


src/talkaction.cpp
```diff
@@ -61,7 +61,9 @@ Event* TalkActions::getEvent(const std::string& nodeName)
 
 bool TalkActions::registerEvent(Event* event, const pugi::xml_node&)
 {
-	talkActions.push_front(*static_cast<TalkAction*>(event)); // event is guaranteed to be a TalkAction
+	TalkAction* talkAction = static_cast<TalkAction*>(event); // event is guaranteed to be a TalkAction
+	talkActions.push_front(*talkAction);
```
ranisalt: `std::move(*talkaction)` might use less resources than copying here. Also I'd use an `unique_ptr` for clean pointer handling.

```
auto talkaction = std::unique_ptr<Talkaction>(*event);
talkactions.push_front(std::move(*talkaction.get()));
```

src/talkaction.cpp
```diff
@@ -61,7 +61,8 @@ Event* TalkActions::getEvent(const std::string& nodeName)
 
 bool TalkActions::registerEvent(Event* event, const pugi::xml_node&)
 {
-	talkActions.push_front(*static_cast<TalkAction*>(event)); // event is guaranteed to be a TalkAction
+	auto talkAction = std::unique_ptr<TalkAction>(static_cast<TalkAction*>(event)); // event is guaranteed to be a TalkAction
+	talkActions.push_front(std::move(*talkAction.get()));
```
ranisalt: Just noticed you can use `*talkAction`, `.get()` is redundant [(source)](http://en.cppreference.com/w/cpp/memory/unique_ptr/operator*)

Kamenuvol: I'd rather use get to emphasize is not a normal pointer

ranisalt:
Thanks!

PR #2148, opened 2017-02-03
Create a hashing backend for multiple algorithms

ranisalt:
This adds a higher level hashing backend with support for different algorithms, heavily based on [Django framework's](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/) which is very reliable. A password must be stored in the database with the following format:

```
<algorithm>$<iterations>$<base64 salt>$<base64 hash>
```

Example:

```
pbkdf2_sha256$30000$8CvkBnHLIGIU$6HdK2DDTZve8j4S+KEH4xiczzM2dWlFUDYhnj9wSO8Q=
```

For using the (included) PBKDF2 w/ SHA256 hash function with 30k iterations. This, of course, is not created by the server but by the account manager. The notable exception being the current SHA1 hasher, automatically used when the password contains no $ character (as SHA1 is hex encoded) and has exactly 40 bytes.

To add a new hasher, inherit from the `hashers::Hasher` class and add to the `hashers` vector in hashers.cpp.

I intend to make a new function to harden the password whenever the `iterations` parameter (or whatever your hasher might use as work factor) is upgraded, thus automatically increasing safety of password hashes. This is too based on what Django does.

*The password field on the accounts table MUST be changed to `varchar(128)` or longer*, but I'm not sure I should do it in this PR directly.


src/hashers.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <cryptopp/hex.h>
+#include <cryptopp/pwdbased.h>
+#include <cryptopp/sha.h>
+
+#include "hashers.h"
+#include "tools.h"
+
+namespace hashers {
+
+static std::unique_ptr<Hasher> hashers[] = {
```
djarek: I'd prefer std::array over raw arrays - the former does not decay to a pointer unless you specifically want it.

src/hashers.cpp
```diff
@@ -0,0 +1,112 @@
...
+	if (encoded.size() == 40 && it == std::string::npos) {
+		return getHasher("sha1");
+	}
+
+	return getHasher(encoded.substr(0, it));
+}
+
+std::string SHA1Hasher::encode(const std::string& input, const std::string&) const
+{
+	static CryptoPP::SHA1 hasher;
```
djarek: Are CryptoPP:SHA1 objects thread-safe?

src/tools.cpp
```diff
@@ -207,15 +97,15 @@ std::string generateToken(const std::string& key, uint32_t ticks)
 	iKeyPad.append(message);
 
 	// hmac first pass
-	message.assign(transformToSHA1(iKeyPad));
+	message.assign(hasher.encode(iKeyPad));
```
djarek: Might as well use the HMAC implementation from CryptoPP.

src/hashers.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+	// ancient versions of TFS used plain SHA1 passwords
+	if (encoded.size() == 40 && it == std::string::npos) {
+		return getHasher("sha1");
+	}
+
+	return getHasher(encoded.substr(0, it));
+}
+
+std::string SHA1Hasher::encode(const std::string& input, const std::string&) const
```
djarek: If we detect SHA1 in the database we should print a big scary warning. Not every time the hasher is used, but perhaps once, at startup.

ranisalt: Yeah, I thought about that. And it is also possible to force updating to PBKDF2.

src/tools.cpp
ranisalt: Will do, but on a separate PR.

src/hashers.cpp
ranisalt: Not that I'm aware of, no. Might use `thread_local` here?

ranisalt:
I'd like to have feedback from aac maintainers like @Znote @gesior @Avuenja too to see which hashing algorithms they'd like to use (namely bcrypt as they use PHP) and if the formatting is OK. I can provide help if needed even on another language.

ranisalt:
I mistyped an username and forgot the guys from devaac and I can't edit on my phone, if someone can mention them.

wgrriffel:
@DevelopersPL

Znote:
@ranisalt 
bcrypt is supported as of PHP 5.5 by default, would this suffice?

http://php.net/manual/en/function.password-hash.php

> PASSWORD_BCRYPT - Use the CRYPT_BLOWFISH algorithm to create the hash. This will produce a standard crypt() compatible hash using the "$2y$" identifier. The result will always be a 60 character string, or FALSE on failure.

```php
/**
 * Note that the salt here is randomly generated.
 * Never use a static salt or one that is not randomly generated.
 *
 * For the VAST majority of use-cases, let password_hash generate the salt randomly for you
 */
$options = [
    'cost' => 11,
    //'salt' Manual configuration of salt is deprecated as of PHP 7.0
];
echo password_hash("rasmuslerdorf", PASSWORD_BCRYPT, $options)."\n";
```

> password_hash() returns the algorithm, cost and salt as part of the returned hash. Therefore, all information that's needed to verify the hash is included in it. This allows the verify function to verify the hash without needing separate storage for the salt or algorithm information.

```php
// See the password_hash() example to see where this came from. (loaded from DB)
$hash = '$2y$07$BCryptRequires22Chrcte/VlQH0piJtjXl.0t1XkA8pw9dMXTpOq';

// Loaded from user input
$password = 'rasmuslerdorf';

if (password_verify($password, $hash)) {
    echo 'Password is valid!';
} else {
    echo 'Invalid password.';
}
```

If this pull request gets merged to master, I will shortly implement official support for it on https://github.com/Znote/ZnoteAAC

src/hashers.cpp
djarek: It might be worth checking how expensive the construction of these objects is. If it's cheap we might not have to deal with statics at all.

ranisalt:
@Znote can't use PHP hash pattern because it limits to just bcrypt variants, but you should be able to easily convert between this format and PHP bcrypt with some string manipulation.

I'll still have to code bcrypt on Blowfish since CryptoPP doesn't have a full bcrypt impl.

src/hashers.cpp
ranisalt: I read CryptoPP is safe on class-level, so changing to non-static should be fine.

src/hashers.cpp
ranisalt: @djarek Django's hashing system auto upgrades a password hash when it gets weak (n. of iterations < default iterations), but TFS is the server and I'm not sure it should mess with passwords, even though upgrading the cost while not changing the hash should not be a problem.

What do you think? Should I make an auto-upgrading password algorithm?

marksamman:
Stronger algorithms require >100ms computation time (and rightfully so), occupying the dispatcher or network thread for so long is not ideal. Is this addressed? That's the main reason we still use SHA-1. We could easily strengthen it by adding salt, pepper, and also running multiple rounds of SHA-1 and it would still be feasible to run in the network thread.

As for library and algorithm, I would prefer libsodium over crypto++, and use crypto_pwhash (Argon2i). It looks like it's going to be added as a core extension in PHP 7.2: https://wiki.php.net/rfc/libsodium

src/hashers.cpp
djarek: IMO doing password upgrading should be related to password changes and is a web apps job.

ranisalt:
I didn't use libsodium as the version present on Ubuntu 14.04 is awfully archaic and didn't have Argon2 yet. When something on the like of #2157 is available we can push the Ubuntu version on Travis forward to 16.04 or higher and use a more up-to-date version of libsodium.

ranisalt:
Ooops accidentally closed.

djarek:
[This](https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html) might be relevant to this PR.

ranisalt:
@djarek well it certainly speeds up the need to move avay from SHA1 as soon as possible.

src/hashers.cpp
DSpeichert: I think password hash upgrades could be done in here. With the HTTP API coming in, TFS may become the only place where the hash is actually computer. Definitely you'd get the hash upgraded sooner if you have a chance to do it on every login instead of waiting for the player to log in on the website (which they might never need to do).

ranisalt: Agreed @DSpeichert, actually that would be a great design of an API 👍

gugahoa:
It's still not as scary as it sounds (specially for OTs). No need to hurry.

ranisalt:
*finally* there's a PPA with up-to-date libsodium, previously there was only up to 1.0.8 which didn't have Argon2. I replaced Crypto++ PBKDF2 with Argon2 from Sodium and added another Scrypt hasher as a concept of how it should work and be easy to add new hashers.

Next step is to auto upgrade (rehash) old passwords.

appveyor.yml
```diff
@@ -19,6 +19,7 @@ environment:
 
 install:
   - ps: if (-not (Test-Path tfs-sdk-3.2)) { Start-FileDownload https://static.otland.net/dl/tfs-sdk-3.2.zip; 7z x tfs-sdk-3.2.zip }
+  - ps: Install-Package cryptopp
```
ranisalt: F**k, forgot that. Sorry Windows.

src/hashers.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+std::string Argon2Hasher::encode(const std::string& input, const std::string& salt) const
+{
+	return encode(input, salt, default_time_cost, default_memory_cost);
+}
+
+std::string Argon2Hasher::encode(const std::string& input, const std::string& salt, std::size_t timeCost, std::size_t memoryCost) const
+{
+	char out[crypto_pwhash_argon2i_STRBYTES];
+	if (crypto_pwhash_argon2i_str(out, input.c_str(), input.size(), timeCost, memoryCost) == 0) {
```
ranisalt: I could use only `crypto_pwhash_str` here but there's no clear guarantee from libsodium that Argon2 will stay forever here, so just in case it changes.

jo3bingham:
@gugahoa is correct, the SHA-1 collision that was announced today is only obtainable under set pre-conditions and doesn't warrant immediate action. However, it's always good practice to use stronger encryption when feasible.

Like @marksamman said, we have to be careful using stronger encryption since it occupies the network thread, but SHA-256 isn't that much slower (30% at the most) than SHA-1, and would be the ideal candidate to replace SHA-1. I see no reason to support more encryption methods (especially even slower ones), but that's just my opinion. More importantly, we should promote good [Password Strength](https://xkcd.com/936/)

DSpeichert:
Looking at the current state of PHP (considering it's the primary language used in AACs for TFS), we can easily move to sha256 but we can't use Argon2i. Even if the latter comes in PHP 7.2, that would rule out the use of all previous PHP versions. I guess sha256 should be good enough for the time being and also fast enough.

marksamman:
For password hashing, sha256 does not offer much more strength than sha1.
We don't use sha1 in a way that collisions, or even preimage attacks would
compromise the security of TFS. The worst thing that can happen is that
someone can find a different password to login to your account than your
actual password, and without even knowing the hash, that's not feasible
over the network. If the attacker already knows the hash, you have bigger
problems.

Adding salt and pepper does much more than upgrading to sha256, we could
even run multiple iterations of sha1 to make computations some factors more
expensive. I don't think upgrading to sha256 or any other hash not meant
for passwords is worth introducing a new dependency for.

DSpeichert:
Adding salt is a good idea and effective against rainbow tables. Having pepper in config.lua makes the database less "portable" in case someone loses/forgets/deletes config.lua. That's a potential usability drawback.

ranisalt:
Multiple rounds of SHA1 is essentially PBKDF2 that I had before 😛 

Those specialized password algorithms have automatic salting, meaning they generate cryptographically safe salts and store with each password hash.

While I know SHA1 is reasonably safe for hiding in a database, this is actually a principle of computer security. You should never trust that your password (or, in this case, your hashes) are perfectly hidden in a safe, and optimally it should be okay for an attacker to have access to every hash in the database and still be unable to crack them. This is not the case for MD5, which we already deprecated, and in a near future might not be the case for SHA1 too, especially because SHA1 is designed to be fast and not secure.

On the other hand, those password hashes are intentionally very slow, hashing a password taking time in a matter of hundreds of milliseconds or even more than a second if tuned with more conservative parameters, while you can compute pretty much millions of SHA1 hashes in a second, even thousands of millions in a GPU.

It's not very hard to crack open a server database. Most servers use dumb passwords and misconfigure their webservers - it was a very common attack with XML servers to access the accounts/players via the website, you might remember, and it was even worse because plain passwords were stored.

SHA1 is good for other uses, such as MOTD hashing (yet `std::hash` is probably enough for that, as it doesn't change that much and at most there is 1 hash stored and 1 hash compared, ever) and is still required for token authentication, so it shouldn't be removed.

mackerel225:
If you want to use any SHA-type algorithm, it might be better to just add more functionality to the current one, rather than sacrificing speed when implementing SHA256/512. Otherwise, I would replace it completely.

On top of that, there are 2 popular hashes which could be incorporated as well. First one has already been mentioned; Argon2 is a popular solution when it comes to passwords, but there is also an alternative called blake2.

The latter has been optimised for 64 bit platforms and I believe we could benefit from it, if the results from https://blake2.net/ are correct. The speed is much higher compared to current SHA-1, whilst ensuring the security is on similar level that of SHA-3.

But that's not it! Should we expect even better performance, perhaps we should incorporate its improvement, Blake3: 

https://github.com/BLAKE3-team/BLAKE3
https://www.infoq.com/news/2020/01/blake3-fast-crypto-hash/

However, this should definitely be tested as one of the users had complained about c++ implementation being slower than blake2: https://kdave.github.io/blake3-vs-blake2-in-btrfs/ (see personal addendum)

ranisalt:
BLAKE is not an alternative to Argon2; it is meant for fast hashing to verify integrity of data. Argon2, on the other hand, is meant for slow hashing to protect passwords against brute forcing.

mackerel225:
Hello @ranisalt 

In terms of protection against brute force argon2 is a way to go, just like you said. My thinking was that because we use SHA-1, which is not ideal for passwords either. We could at least replace it with something faster: blake2/blake3 if all we care about is speed. This could be used for something like MOTD (message of the day) as it appears it's currently using SHA-1.

There is also type SHA-256 in commits, unless we have a really good reason to implement it, I would re-read this https://github.com/zcash/zcash/issues/706#issuecomment-187807410

Also, @ranisalt I would like to test the argon2 functionality at some point, do you think the changes made in this PR are up to date? Or is there anything you would suggest changing as it's seems to be 3 years old now.

Let me know what you think

ranisalt:
@mackerel225 in this PR I changed MOotD to use std::hash, which returns a 32 bit integer, and that should be enough since MotD is trusted input and does not require integrity or high collision avoidance.

You should be able to use it by fixing the conflicts, as it is contained in its own file it does not interact with the rest of the server.

Erza:
What's the status of this now? I saw 2 users asking about other hashing algorithms than SHA1 on OTLand lately:
https://otland.net/threads/cryptography-an-introduction.263166/#post-2646935
https://otland.net/threads/bcrypt-instead-of-sha1.274821

It's 2021 and finally time to get moving with this (4 year old!) PR.

yamaken93:
> Stronger algorithms require >100ms computation time (and rightfully so), occupying the dispatcher or network thread for so long is not ideal. Is this addressed? That's the main reason we still use SHA-1. We could easily strengthen it by adding salt, pepper, and also running multiple rounds of SHA-1 and it would still be feasible to run in the network thread.
> 
> As for library and algorithm, I would prefer libsodium over crypto++, and use crypto_pwhash (Argon2i). It looks like it's going to be added as a core extension in PHP 7.2: https://wiki.php.net/rfc/libsodium

I think its pretty clear we need helper threads and some mean to let do some asynchronous and then comeback.

ranisalt:
I'll try something this weekend

yamaken93:
> I'll try something this week
Add me on discord so we can debate some ideas: yamaken#6490

gesior:
I don't think we can use any strong hashing on OTSes. DDoSes are much bigger problem than database hacks and data breaches.
About people selling databases or using own OTS databases to hack people - these people start OTS with plain password implementation.

All we can do is add salt to each account. I don't know why it's not here, as it was already implemented in TFS 0.4.

If we tune algorithm to execute for 100 ms, 80 IPs reloging 1 time per second will kill 8 core CPU. Adding multiple 'workers' won't help much.
We will tune it to execute 100 ms on CPU, but on GPU - which hackers will use to break passwords - it will be muuuch faster.

PR #2149, opened 2017-02-03
Change all event containers to store objects

Mkalo:
Well first I've changed the Event pointers to use smart pointers, the
second step would be to change the registerEvent functions to be
something like TalkActions::registerEvent (being careful to copy or move
depending on the event if it allows duplicates etc.)

But before doing that I would like some opinion on that, like if it does
even have a chance of being merged.


ranisalt:
I do appreciate it. There's a land of undefined behavior and segmentation fault on naked pointers.

src/baseevents.cpp
```diff
@@ -61,15 +61,19 @@ bool BaseEvents::loadFromXml()
...
+		bool success;
+
 		pugi::xml_attribute scriptAttribute = node.attribute("script");
 		if (scriptAttribute) {
 			std::string scriptFile = "scripts/" + std::string(scriptAttribute.as_string());
-			event->checkScript(basePath, scriptsName, scriptFile) && event->loadScript(basePath + scriptFile);
+			success = event->checkScript(basePath, scriptsName, scriptFile) && event->loadScript(basePath + scriptFile);
 		} else {
-			event->loadFunction(node.attribute("function"));
+			success = event->loadFunction(node.attribute("function"));
```
ranisalt: Functions are generally being removed, it would be better to wait until they are removed so you won't need this book :)

src/creatureevent.cpp
```diff
@@ -57,17 +57,17 @@ std::string CreatureEvents::getScriptBaseName() const
...
 }
 
-Event* CreatureEvents::getEvent(const std::string& nodeName)
+std::unique_ptr<Event> CreatureEvents::getEvent(const std::string& nodeName)
 {
 	if (strcasecmp(nodeName.c_str(), "event") != 0) {
 		return nullptr;
 	}
-	return new CreatureEvent(&scriptInterface);
+	return std::unique_ptr<Event>(new CreatureEvent(&scriptInterface));
```
ranisalt: `return {new ...};` is cleaner

src/creatureevent.cpp
```diff
@@ -57,17 +57,17 @@ std::string CreatureEvents::getScriptBaseName() const
...
 }
 
-bool CreatureEvents::registerEvent(Event* event, const pugi::xml_node&)
+bool CreatureEvents::registerEvent(std::unique_ptr<Event>&& event, const pugi::xml_node&)
 {
-	CreatureEvent* creatureEvent = static_cast<CreatureEvent*>(event); //event is guaranteed to be a CreatureEvent
+	CreatureEvent* creatureEvent = static_cast<CreatureEvent*>(event.release()); //event is guaranteed to be a CreatureEvent
```
ranisalt: Why not unique ptr here too?

Mkalo: Yea, I will change all those pointers to unique ptr when I start changing the containers. I let it like that for now.

src/actions.cpp
```diff
@@ -76,17 +76,17 @@ std::string Actions::getScriptBaseName() const
...
 	}
-	return new Action(&scriptInterface);
+	return std::unique_ptr<Event>(new Action(&scriptInterface));
 }
 
-bool Actions::registerEvent(Event* event, const pugi::xml_node& node)
+bool Actions::registerEvent(std::unique_ptr<Event>&& event, const pugi::xml_node& node)
 {
-	Action* action = static_cast<Action*>(event); //event is guaranteed to be an Action
+	Action* action = static_cast<Action*>(event.release()); //event is guaranteed to be an Action
```
djarek: Violates http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#a-namerf-consumeaf18-for-consume-parameters-pass-by-x-and-stdmove-the-parameter ?

src/globalevent.cpp
```diff
@@ -60,17 +60,17 @@ void GlobalEvents::clear()
...
 }
 
-Event* GlobalEvents::getEvent(const std::string& nodeName)
+std::unique_ptr<Event> GlobalEvents::getEvent(const std::string& nodeName)
 {
 	if (strcasecmp(nodeName.c_str(), "globalevent") != 0) {
 		return nullptr;
 	}
-	return new GlobalEvent(&scriptInterface);
+	return std::unique_ptr<Event>(new GlobalEvent(&scriptInterface));
```
djarek: ```return {new ...};```

src/spells.cpp
```diff
@@ -122,31 +122,33 @@ std::string Spells::getScriptBaseName() const
...
 	return nullptr;
 }
 
-bool Spells::registerEvent(Event* event, const pugi::xml_node&)
+bool Spells::registerEvent(std::unique_ptr<Event>&& event, const pugi::xml_node&)
 {
-	InstantSpell* instant = dynamic_cast<InstantSpell*>(event);
+	InstantSpell* instant = dynamic_cast<InstantSpell*>(event.get());
 	if (instant) {
+		event.release();
```
djarek: Don't do this because it looks like a leak at first glance.

src/creatureevent.cpp
djarek: Conversion constructor is explicit.

DSpeichert:
@djarek Please re-review if you have time.

src/weapons.h
```diff
@@ -62,7 +62,7 @@ class Weapons final : public BaseEvents
 class Weapon : public Event
 {
 	public:
-		explicit Weapon(LuaScriptInterface* interface) : Event(interface) {}
```
djarek: Why did you remove the `explicit` specifier?

Mkalo: Not really sure, I'll re add them

src/actions.cpp
```diff
@@ -108,14 +99,14 @@ bool Actions::registerEvent(Event* event, const pugi::xml_node& node)
 		uint16_t iterId = fromId;
 		uint16_t toId = pugi::cast<uint16_t>(toIdAttribute.value());
 
-		auto result = useItemMap.emplace(iterId, action);
+		auto result = useItemMap.emplace(iterId, *action);
```
djarek: Why aren't you moving the action, like in line 86?

Mkalo: Because in this part, its the fromid and toid code, I need to make a copy of the action for everytime it's inserted. Not ideal but I don't know how to do it better with a map.

PR #2157, opened 2017-02-05
Dockerization for TFS

Tarjei400:


Tarjei400:
I don't have idea what do you mean. Both Server and website can be dockerized and encapsulated in seperate containers. F.i server container has redirected ports 7171 and 7172 so you can actualy connect to server through localhost ip.

brunominervino:
@Tarjei400 Forget my comment, I ended up confusing the gdb port with apache. Sorry.

Taking advantage of this as a sugestion, would not it be better to take advantage of docker to install the webserver, like apache or maybe nginx?

Tarjei400:
I don't think its a good idea. On necronia we make seperate docker container for website project. It makes deployment, and debugining much easier

brunominervino:
Thinking well, I also think that is not a good idea, since there is no official AAC of this repository with established prerequisites.

Tarjei400:
Hmm I guess I will try to mess around with docker compsoser, however I dont know if this will support remote debugging from a container.

gugahoa:
I was going to comment this. Docker compose is a great way to do what's being done in bash here. And if it's working right now, nothing stops it from working with docker compose

ranisalt:
> I was going to comment this. Docker compose is a great way to do what's being done in bash here. And if it's working right now, nothing stops it from working with docker compose

Yes, I intend to convert into separate containers for server, database and maybe AAC (even though they can be pretty different from language to libs)

ranisalt:
I'm working on the `docker` branch and already managed to use on Travis: https://travis-ci.org/otland/forgottenserver/builds/199839354

Next step is to use compose to fire up a clean mysql server and run a startup check after every build to ensure at least basic functionality.

ranisalt:
> Do not you think it's best to start the docker on standard port 80? Since many of the otadmins use their hosts just to host only tfs + website?

The AAC should be as isolated as possible, with only CRUD connection to the database preferably over sockets and run behind a reverse proxy such as Nginx, if not embedded into Apache. The server and the AAC would better run separate to prevent file leaking.

DSpeichert:
First of all, it is considered best practice to run exactly one service per container.

I think we are confusing or mixing 3 types of docker containers that are potentially useful to TFS:
* a container to build TFS (which is what this PR includes and what @ranisalt is working on) - unnecessarily big to run TFS (all dev tools), only suitable for building TFS
* a container to run TFS (base OS with libraries; you mount the binary + data pack) - risky, libs may not match
* a container to run TFS with TFS (OS + libraries + TFS; you mount only datapack) - safest to use, container release = TFS release

ranisalt:
@DSpeichert library versioning shouldn't be an issue as long as you use the same base image (e.g. use `ubuntu:xenial` for both building and running, but on separate containers)

DSpeichert:
We can set up automated builds with docker if that is a wanted feature. It would have to be popular enough to offset the work for maintaining another build recipe.

ranisalt:
You mean having a `tfs` image with the release tags? I'm not sure about that, it is very common to modify the sources so having packaged versions would not be that popular, only if we manage to make Docker would allow patching sources somehow.

DSpeichert:
I would say more people use TFS unmodified than modified. How many of those use Docker?
Those who do would gain easy access to updates by just updating the image.

I don't see anything that we can easily use in this PR. Having a Dockerfile would imply selecting a single guest OS. We have guides for compilation on many OSes. On Windows, it's more troublesome to get Docker to work to use xenial than to compile it directly on Windows using the wiki.

ranisalt:
I'd use Docker more for isolation than ease of build. Of course, it also provides additional advantages such as always using the latest LTS or consistent system and libs.

DSpeichert:
These days using a bash script to do what's being done here is not the recommended way, as @gugahoa mentioned there's docker-compose for that.

While we have established there is avenue for adding Docker into this project, we need to iron out the way to do it. Let's follow #2185 for further discussion.

PR #2158, opened 2017-02-05
Added new message types (white messages in channel)

Colandus:
This commit will add new message types:
```
MESSAGE_GUILD = 33
MESSAGE_PARTY_MANAGEMENT = 34
MESSAGE_PARTY = 35
```

All of which will send to a channel. Player.sendTextMessage has been modified to support them with the following usage:

`player:sendTextMessage(type, text, channelId)`

Where type must be one of the above (although not enforced, but won't work without them).

### DIFF
--- a/src/const.h
+++ b/src/const.h
@@ -212,6 +212,9 @@ enum MessageClasses : uint8_t {
 	MESSAGE_EVENT_DEFAULT = 30, /*White message at the bottom of the game window and in the console*/
 	MESSAGE_LOOT = 31,
 
+	MESSAGE_GUILD = 33, /* White message in channel (+ channelId) */
+	MESSAGE_PARTY_MANAGEMENT = 34, /* White message in channel (+ channelId) */
+	MESSAGE_PARTY = 35, /* White message in channel (+ channelId) */
 	MESSAGE_EVENT_ORANGE = 36, /*Orange message in the console*/
 	MESSAGE_STATUS_CONSOLE_ORANGE = 37,  /*Orange message in the console*/
 };
--- a/src/const.h
+++ b/src/const.h
@@ -212,9 +212,9 @@ enum MessageClasses : uint8_t {
 	MESSAGE_EVENT_DEFAULT = 30, /*White message at the bottom of the game window and in the console*/
 	MESSAGE_LOOT = 31,
 
-	MESSAGE_GUILD = 33, /* White message in channel (+ channelId) */
-	MESSAGE_PARTY_MANAGEMENT = 34, /* White message in channel (+ channelId) */
-	MESSAGE_PARTY = 35, /* White message in channel (+ channelId) */
+	MESSAGE_GUILD = 33, /*White message in channel (+ channelId)*/
+	MESSAGE_PARTY_MANAGEMENT = 34, /*White message in channel (+ channelId)*/
+	MESSAGE_PARTY = 35, /*White message in channel (+ channelId)*/
 	MESSAGE_EVENT_ORANGE = 36, /*Orange message in the console*/
 	MESSAGE_STATUS_CONSOLE_ORANGE = 37,  /*Orange message in the console*/
 };
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1367,6 +1367,9 @@ void LuaScriptInterface::registerFunctions()
 	registerEnum(MESSAGE_HEALED_OTHERS)
 	registerEnum(MESSAGE_EXPERIENCE_OTHERS)
 	registerEnum(MESSAGE_EVENT_DEFAULT)
+	registerEnum(MESSAGE_GUILD)
+	registerEnum(MESSAGE_PARTY_MANAGEMENT)
+	registerEnum(MESSAGE_PARTY)
 	registerEnum(MESSAGE_EVENT_ORANGE)
 	registerEnum(MESSAGE_STATUS_CONSOLE_ORANGE)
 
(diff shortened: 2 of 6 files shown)


src/luascript.cpp
```diff
@@ -8523,9 +8526,14 @@ int LuaScriptInterface::luaPlayerShowTextDialog(lua_State* L)
 int LuaScriptInterface::luaPlayerSendTextMessage(lua_State* L)
 {
 	// player:sendTextMessage(type, text[, position, primaryValue = 0, primaryColor = TEXTCOLOR_NONE[, secondaryValue = 0, secondaryColor = TEXTCOLOR_NONE]])
+	// player:sendTextMessage(type, text, channelId)
```
raymondtfr: Why did you not include the previous parameters in this comment?

Colandus: @raymondtfr They won't be used if you send with channelId.

src/luascript.cpp
```diff
@@ -8523,9 +8526,14 @@ int LuaScriptInterface::luaPlayerShowTextDialog(lua_State* L)
...
 	int parameters = lua_gettop(L);
 
 	TextMessage message(getNumber<MessageClasses>(L, 2), getString(L, 3));
+	if (parameters == 4) {
```
WibbenZ: Not sure if this is gonna work, let's say we use position then it will try to get the channelId.

Colandus: @WibbenZ No offence but it looks like you guys aren't reading the code. Position won't be considered unless params >= 6.

Kamenuvol:
Maybe we should use Chat userdata instead of id

Colandus:
I noticed that if the channel is closed, it will send a MESSAGE_INFO_DESCR instead (to Server Log) which would not be the desired behavior when attempting to send to a specific channel. And because there is currently no way to check if a channel is open, I decided it's best to not send the message if the channel isn't found

@marksamman Can you check if my implementation is fine? In chat.h/cpp mainly and perhaps luascript.cpp (the channel checks). I was hoping to get ChatChannel::hasUser into chat.h solely (which I assume would be best considering its small size) but I couldn't manage to.

src/luascript.cpp
```diff
@@ -8523,9 +8526,28 @@ int LuaScriptInterface::luaPlayerShowTextDialog(lua_State* L)
...
+		ChatChannel* channel = g_chat->getChannel(*player, channelId);
+		if (!channel || !channel->hasUser(*player)) {
+			pushBoolean(L, false);
+			return 1;
+		}
+
+		message.channelId = channelId;
+	}
+
 	if (parameters >= 6) {
```
ranisalt: `else if`

Colandus:
@ranisalt I thought on the contrary, you wouldn't prefer an else there so I didn't add it :p Anyway, you have a say concerning my previous question?

wgrriffel:
@Colandus 
no offence but you should get the basics of git, doing git through github spams the commit history and gives a vague comment to commit

comedinha:
MESSAGE_SAY = 1,
MESSAGE_WHISPER = 2,
MESSAGE_YELL = 3,
MESSAGE_PRIVATE_FROM = 4,
MESSAGE_PRIVATE_TO = 5,
MESSAGE_CHANNEL_MANAGEMENT = 6,
MESSAGE_CHANNEL = 7,
MESSAGE_CHANNEL_HIGHLIGHT = 8,
MESSAGE_SPELL = 9,
MESSAGE_NPC_FROM_START_BLOCK = 10,
MESSAGE_NPC_FROM = 11,
MESSAGE_NPC_TO = 12,
MESSAGE_GAMEMASTER_BROADCAST = 13,
MESSAGE_GAMEMASTER_CHANNEL = 14,
MESSAGE_GAMEMASTER_PRIVATE_FROM = 15,
MESSAGE_GAMEMASTER_PRIVATE_TO = 16,
MESSAGE_LOGIN = 17,
MESSAGE_ADMIN = 18,
MESSAGE_GAME = 19,
MESSAGE_GAME_HIGHLIGHT = 20,
MESSAGE_FAILURE = 21,
MESSAGE_LOOK = 22,
MESSAGE_DAMAGE_DEALED = 23,
MESSAGE_DAMAGE_RECEIVED = 24,
MESSAGE_HEAL = 25,
MESSAGE_EXP = 26,
MESSAGE_DAMAGE_OTHERS = 27,
MESSAGE_HEAL_OTHERS = 28,
MESSAGE_EXP_OTHERS = 29,
MESSAGE_STATUS = 30,
MESSAGE_LOOT = 31,
MESSAGE_TRADE_NPC = 32,
MESSAGE_GUILD = 33,
MESSAGE_PARTY_MANAGEMENT = 34,
MESSAGE_PARTY = 35,
MESSAGE_BARK_LOW = 36,
MESSAGE_BARK_LOUD = 37,
MESSAGE_REPORT = 38,
MESSAGE_HOTKEY_USE = 39,
MESSAGE_TUTORIAL_HINT = 40,
MESSAGE_THANKYOU = 41,
MESSAGE_MARKET = 42,
MESSAGE_MANA = 43,
MESSAGE_BEYOND_LAST = 44,

To see if show screen, show channel, is editable and need level (flash and tibia 11) https://github.com/comedinha/Flash-News/blob/master/scripts/tibia/chat/MessageMode.as#L19

Colandus:
@wgrriffel Yeah I know :p

src/luascript.cpp
```diff
@@ -8523,27 +8526,41 @@ int LuaScriptInterface::luaPlayerShowTextDialog(lua_State* L)
...
+		if (parameters >= 6) {
+			message.position = getPosition(L, 4);
+			message.primary.value = getNumber<int32_t>(L, 5);
+			message.primary.color = getNumber<TextColor_t>(L, 6);
+		}
+
+		if (parameters >= 8) {
+			message.secondary.value = getNumber<int32_t>(L, 7);
+			message.secondary.color = getNumber<TextColor_t>(L, 8);
+		}
```
ranisalt: Why did you separe this statement instead of chaining `else if`s?

Colandus: @ranisalt I didn't separate it, that's how it originally was.

src/const.h
```diff
@@ -212,6 +212,9 @@ enum MessageClasses : uint8_t {
 	MESSAGE_EVENT_DEFAULT = 30, /*White message at the bottom of the game window and in the console*/
 	MESSAGE_LOOT = 31,
```
brunominervino: Like this.

wgrriffel:
@Colandus try.github.io

src/luascript.cpp
ranisalt: Well, for code style, it should be two else ifs instead of this nested if for one less level of indentation. This pattern is much more used throughout the code, you could fix while at it.

Colandus: @ranisalt But that would break the code... That's why it wasn't so from the start and that's why I didn't do it.

ranisalt: Are you sure? Indeed you will have to swap them, so that `>= 8` is checked before `>= 6`, and each one on its own `else if` branch. The `else` branch is when not enough parameters were supplied and you can push nil and return 1 directly.

ranisalt:
Thanks!

PR #2161, opened 2017-02-09
Updated Creature.getClosestFreePosition smarter behavior

Colandus:
Creature.getClosestFreePosition should now begin from inside out, meaning if target position is blocked it will pick the closest radius rather than the furthest one (when using extended, as old code would do).

Also with the ability to go further than just radius 2 by setting maxRadius parameter as desired (which is adapted to accept boolean values for backward compatability).

**Edit:**
Also added "mustBeReachable". I was unsure about that one because I feel it might eat too much CPU. Opinions?
(sorry for my noobish methods of updating commit)

### DIFF
--- a/data/lib/core/creature.lua
+++ b/data/lib/core/creature.lua
@@ -1,27 +1,22 @@
-function Creature.getClosestFreePosition(self, position, extended)
-	local usePosition = Position(position)
-	local tiles = { Tile(usePosition) }
-	local length = extended and 2 or 1
+function Creature.getClosestFreePosition(self, position, maxRadius)
+	maxRadius = maxRadius or 1
 
-	local tile
-	for y = -length, length do
-		for x = -length, length do
-			if x ~= 0 or y ~= 0 then
-				usePosition.x = position.x + x
-				usePosition.y = position.y + y
-
-				tile = Tile(usePosition)
-				if tile then
-					tiles[#tiles + 1] = tile
-				end
-			end
-		end
+	-- backward compatability (extended)
+	if maxRadius == true then
+		maxRadius = 2
 	end
 
-	for i = 1, #tiles do
-		tile = tiles[i]
-		if tile:getCreatureCount() == 0 and not tile:hasProperty(CONST_PROP_IMMOVABLEBLOCKSOLID) then
-			return tile:getPosition()
+	for _ = 0, maxRadius do
+		for x = -radius, radius do
+			for y = -radius, radius do
+				if math.abs(x) == radius or math.abs(y) == radius then
+					local checkPosition = Position(position.x + x, position.y + y, position.z)
+					local tile = Tile(checkPosition)
+					if tile:getCreatureCount() == 0 and not tile:hasProperty(CONST_PROP_IMMOVABLEBLOCKSOLID) then
+						return checkPosition
+					end
+				end
+			end
 		end
 	end
 	return Position()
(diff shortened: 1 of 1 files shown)


ninjalulz:
Too bad you broke the function in the process, haha 😄

Colandus:
@ninjalulz Yeah, how?

ninjalulz:
It will stop at L10 since **radius** is an undefined value.

Colandus:
@ninjalulz Yeah thanks for pointing out. radius should be created by the loop, for some reason I changed it to _ no idea why. Silly me!

data/lib/core/creature.lua
```diff
@@ -1,29 +1,25 @@
...
-					tiles[#tiles + 1] = tile
+	for radius = 0, maxRadius do
+		for x = -radius, radius do
+			for y = -radius, radius do
+				if math.abs(x) == radius or math.abs(y) == radius then
+					local checkPosition = Position(position.x + x, position.y + y, position.z)
+					local tile = Tile(checkPosition)
+					if tile:getCreatureCount() == 0 and not tile:hasProperty(CONST_PROP_IMMOVABLEBLOCKSOLID) and
+						(not mustBeReachable or self:getPathTo(checkPosition)) then
+						return checkPosition
```
ranisalt: Please review if it's not possible to reduce levels of indentation inverting the if conditions and using `continue`.

data/lib/core/creature.lua
```diff
@@ -1,29 +1,25 @@
...
+		maxRadius = 2
+	end
 
-				tile = Tile(usePosition)
-				if tile then
-					tiles[#tiles + 1] = tile
+	for radius = 0, maxRadius do
+		for x = -radius, radius do
+			for y = -radius, radius do
+				if math.abs(x) == radius or math.abs(y) == radius then
```
ranisalt: You should alias `math.abs` to a local variable to avoid scope lookout everytime this is called (and it will be TONS of times, apparently).

ranisalt: Also this spiraling is suboptimal. Every iteration of the outermost loop you cycle through all positions already checked, only to skip them on the if. This is easily refactorable by using four variables instead and reduces complexity a lot.

data/lib/core/creature.lua
Colandus: @ranisalt No it's not and there's no continue in Lua you should know. Actually the Lua standard of indentation is 2 spaces (which is what I use and looks fine) especially due to the fact that Lua code is prone to look this way. Unless you want me to use a goto statement or recursion (would kill one loop).

data/lib/core/creature.lua
Colandus: @ranisalt "This is easily refactorable by using four variables instead and reduces complexity a lot." explain.

data/lib/core/creature.lua
ranisalt: Yeah, I'm more used to C++ than Lua and didn't look further. I'm not sure about function call overhead on Lua, I program in Python (where the overhead is huge) and JS (where the overhead is not significant, and indeed better than looping) so it should be analyzed too.

data/lib/core/creature.lua
ranisalt: Yeah, it's pretty hard to get a perfect spiraling loop :P a good start would be instead of looping from -radius to radius, using only `radius` and duplicating code (not a big issue considering reduced complexity).

Colandus:
Should be fine now! @ranisalt

jo3bingham:
You should really try to use better commit titles/messages other than "Update creature.lua".

Colandus:
@jo3bingham I'm noob :stuck_out_tongue_closed_eyes: And because I'm too lazy to read the git manual, it's either these noob commits or none at all. I don't say there is no embarrassment when clicking the save button.

PR #2165, opened 2017-02-10
Move remaining commands to Lua (remake of #1990)

ranisalt:
This is a remake of #1990 by @Mkalo as it became too hard to rebase and easier to just redo. It's here for review, instead of squashing and merging I'll push to master to keep both commits split for atomicity.


data/talkactions/scripts/sellhouse.lua
```diff
@@ -0,0 +1,19 @@
+function onSay(player, words, param)
+	local tradePartner = Player(param)
+	if not tradePartner or tradePartner == player then
+		player:sendCancelMessage("Trade player not found.")
+		return false
+	end
+
+	local house = Tile(player:getPosition()):getHouse()
```
Colandus: local house = player:getTile():getHouse()

ranisalt: Good eye.

PR #2167, opened 2017-02-11
new magic effect classes constants

raymondtfr:


src/const.h
```diff
@@ -108,6 +108,11 @@ enum MagicEffectClasses : uint8_t {
 	CONST_ME_YELLOWSMOKE = 168,
 	CONST_ME_GREENSMOKE = 169,
 	CONST_ME_PURPLESMOKE = 170,
+	CONST_ME_EARLY_THUNDER = 171,
+	CONST_ME_RAGIAZ_BONECAPSULE = 172,
+	CONST_ME_CRITICAL_DAMAGE = 173,
+
```
ranisalt: What about 174?

raymondtfr: It's empty.

ranisalt: So mark it just like above (see 159-166)

ranisalt:
Thanks!

PR #2168, opened 2017-02-12
Move Convince Creature to Lua

Kamenuvol:


ninjalulz:
Hm, this doesn't notify surrounding creatures about the change, and I think you might have fixed a respawn issue.

Kamenuvol:
I will fix the notification, and.. which respawn issue was fixed oO?

DSpeichert:
@ninjalulz Can you contribute the fix for surrounding creatures you mentioned?

DSpeichert:
We have merged #2102 that duplicates this behavior.

However, I see that this version of `convince_creature.lua` also checks for summons and removes them. If that is necessary you can open a new PR adding just those lines.

https://github.com/otland/forgottenserver/pull/2168/files#diff-5b7d3daa3f9be5f0d1e801abf0e79521R33

PR #2176, opened 2017-02-16
fix removeMoney unintended behavior

socket2810:
The variable removeCount was being incorrectly rounded up.
If a player had a stack of 100 gold coins, and you tried to remove 1 gold coin, they would then have a stack with 98 gold coins and another with 1.

This behavior can probably also be fixed by calling internalRemoveItem before addMoney, assuming the subsequent call to addMoney would reuse the initial stack of gold coins.


src/game.cpp
```diff
@@ -1483,7 +1483,7 @@ bool Game::removeMoney(Cylinder* cylinder, uint64_t money, uint32_t flags /*= 0*
 			money -= moneyEntry.first;
 		} else if (moneyEntry.first > money) {
 			const uint32_t worth = moneyEntry.first / item->getItemCount();
-			const uint32_t removeCount = (money / worth) + 1;
+			const uint32_t removeCount = (money + worth - 1) / worth;
```
ranisalt: Is this poor man's rounding? I think `std::ceil` has clearer intention.

Mkalo:
Merge status? @ranisalt

ranisalt:
Thanks!

PR #2178, opened 2017-02-17
IPv6 support

djarek:
Hey, it's 2017 so it's high time we implemented support for IPv6 connections. The server now listens on an IPv6 socket unless it's configured to only bind to a specific IPv4 address. The address is now stored as a varchar in the database. The internal representation of the address has been changed to the
asio-native one for simplicity.

### DIFF
--- a/data/migrations/19.lua
+++ b/data/migrations/19.lua
@@ -1,8 +1,8 @@
 function onUpdateDatabase()
 	print("> Updating database to version 20 (IPv6 support)")
-	db.query("ALTER TABLE `players` MODIFY `lastip` VARCHAR(45) NOT NULL")
-	db.query("ALTER TABLE `ip_bans` MODIFY `ip` VARCHAR(45) NOT NULL")
-	db.query("UPDATE `players` p  SET p.lastip=INET_NTOA(p.lastip)")
-	db.query("UPDATE `ip_bans` b  SET b.ip=INET_NTOA(b.ip)")
+	db.query("ALTER TABLE `players` MODIFY `lastip` VARBINARY(16) NOT NULL")
+	db.query("ALTER TABLE `ip_bans` MODIFY `ip` VARBINARY(16) NOT NULL")
+	db.query("UPDATE `players`  SET `lastip` = INET_ATON6(INET_NTOA(p.lastip))")
+	db.query("UPDATE `ip_bans`  SET `ip` = INET_ATON6(INET_NTOA(b.ip))")
 	return true
 end
--- a/data/migrations/19.lua
+++ b/data/migrations/19.lua
@@ -1,3 +1,8 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 20 (IPv6 support)")
+	db.query("ALTER TABLE `players` MODIFY `lastip` VARCHAR(45) NOT NULL")
+	db.query("ALTER TABLE `ip_bans` MODIFY `ip` VARCHAR(45) NOT NULL")
+	db.query("UPDATE `players` p  SET p.lastip=INET_NTOA(p.lastip)")
+	db.query("UPDATE `ip_bans` b  SET b.ip=INET_NTOA(b.ip)")
+	return true
 end
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -7391,7 +7391,7 @@ int LuaScriptInterface::luaPlayerGetIp(lua_State* L)
 	// player:getIp()
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
-		lua_pushstring(L, player->getIP().to_string().c_str());
+		pushString(L, player->getIP().to_string());
 	} else {
 		lua_pushnil(L);
 	}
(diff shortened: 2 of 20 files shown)


marksamman:
VARBINARY(16) with INET6_NTOA and INET6_ATON would be more space efficient, they require MySQL 5.6 or newer though.

kornholi:
Does the client support IPv6 yet?

djarek:
@kornholi 
OTClient does support IPv6 (I've tested it). Not sure about "the others".

src/luascript.cpp
```diff
@@ -7391,7 +7391,7 @@ int LuaScriptInterface::luaPlayerGetIp(lua_State* L)
 	// player:getIp()
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
-		lua_pushnumber(L, player->getIP());
+		lua_pushstring(L, player->getIP().to_string().c_str());
```
ranisalt: There's `pushString` to use with C++ strings.

src/player.h
```diff
@@ -1244,7 +1244,7 @@ class Player final : public Creature, public Cylinder
 		uint32_t nextStepEvent = 0;
 		uint32_t walkTaskEvent = 0;
 		uint32_t MessageBufferTicks = 0;
-		uint32_t lastIP = 0;
+		Connection::Address lastIP{};
```
ranisalt: `= {}`?

src/protocolstatus.cpp
```diff
@@ -43,11 +43,10 @@ enum RequestedInfo_t : uint16_t {
...
 void ProtocolStatus::onRecvFirstMessage(NetworkMessage& msg)
 {
-	uint32_t ip = getIP();
-	if (ip != 0x0100007F) {
-		std::string ipStr = convertIPToString(ip);
-		if (ipStr != g_config.getString(ConfigManager::IP)) {
-			std::map<uint32_t, int64_t>::const_iterator it = ipConnectMap.find(ip);
+	Connection::Address ip = getIP();
+	if (!ip.is_loopback()) {
+		if (ip != Connection::Address::from_string(g_config.getString(ConfigManager::IP))) {
```
ranisalt: This should probably be stored to avoid building the address object on every connection (of course, if it's not trivial).

src/server.cpp
```diff
@@ -167,14 +167,26 @@ void ServicePort::open(uint16_t port)
 	pendingStart = false;
 
 	try {
+		IPAddress address;
```
ranisalt: You could instantiate this as `any()` here instead of the else branch below. It's a trivial constructor.

src/server.cpp
```diff
@@ -167,14 +167,26 @@ void ServicePort::open(uint16_t port)
...
+		acceptor.reset(new TCPAcceptor(io_service, TCPEndpoint(address, serverPort)));
+		if (address.is_v6()) {
+			boost::asio::ip::v6_only option;
+			acceptor->get_option(option);
+			if (option) {
+				option = {false};
+				boost::system::error_code err;
+				acceptor->set_option(option, err);
+				if (err) {
+					std::cout << "[ServicePort::open] Warning: enabling IPv4 support failed: " << err.message() << std::endl;
```
ranisalt: `[Warning - ServicePort::open] Enabling...` as standard.

data/migrations/19.lua
```diff
@@ -1,3 +1,8 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 20 (IPv6 support)")
+	db.query("ALTER TABLE `players` MODIFY `lastip` VARCHAR(45) NOT NULL")
+	db.query("ALTER TABLE `ip_bans` MODIFY `ip` VARCHAR(45) NOT NULL")
+	db.query("UPDATE `players` p  SET p.lastip=INET_NTOA(p.lastip)")
```
ranisalt:
```
UPDATE `players`  SET `lastip`=INET_NTOA(`lastip`)
```

data/migrations/19.lua
```diff
@@ -1,3 +1,8 @@
...
+	db.query("UPDATE `ip_bans` b  SET b.ip=INET_NTOA(b.ip)")
```
ranisalt:
```
UPDATE `ip_bans`  SET `ip`=INET_NTOA(`ip`)
```

src/protocolstatus.cpp
djarek: Digging through the implementation of asio is quite hard, but it's non trivial, results in a syscall so I'll cache it in the Connection.

djarek:
I'm wondering - is it intended behavior that loadPlayer() does not use the column lastIP from the player table result?

joseluis2g:
this is really interesting

ranisalt:
@djarek what would you use that for? It probably shouldn't be fetched from DB too.

It's updated on `savePlayer` tho.

kornholi:
If you do this you also have to rework the IP ban system to work on subnets and ban at least /64 in the IPv6 case.

ranisalt:
@kornholi is right, and it's worth reading answers on http://serverfault.com/questions/258466/how-are-ip-based-rules-eg-bans-filters-affected-once-ipv6-becomes-the-standar

PR #2180, opened 2017-02-17
TFS-2179 Fix compiler warnings

jo3bingham:


djarek:
Thanks!

ranisalt:
Weren't there another compiler warning if the returns were missing? @WibbenZ

WibbenZ:
@ranisalt No, that was the main problem, I forgot to remove those 2, if you check https://github.com/jo3bingham/forgottenserver/blob/147098b135e4f1d92f5415c1326d7fae26635dd9/src/game.cpp#L5585
You can see I removed them there xD

But the problem is still there that ex monsters don't give any return values since it always returns true.

PR #2183, opened 2017-02-19
Fix item decay counting down if decay state is false

Vulcanx:


DSpeichert:
Hi @Vulcanx 
Can you please provide some detail into why this this a fix and possible side effects?
At least the commit message must be more meaningful to get merged.

Vulcanx:
@DSpeichert 
I should have been more descriptive, my bad, never made a PR before and it's a bit confusing.
Setting decay state to false with `item:setAttribute(ITEM_ATTRIBUTE_DECAYSTATE, 0)` works, but there was nothing in game.cpp to check if it was false or not, so it continued to decay unless the item wasn't decayable.
Adding this makes it check if the decay state is false, if it is then it will release item and will remove the item from decaying items, rather than letting it keep ticking.
There's probably a better fix for this, but it should work just fine for now and seems to work properly.

wgrriffel:
@Vulcanx some tips for your nexts contributions.
1: don't do patchs against master branch, your origin master branch are suppose to be synchronized to upstream master branch, one reason for that is that you can't open more than one PR at same time while all of your job are under master, you should create a new branch to each patch.
2: make more meaningful commits messages, it makes easier to track commits later.

PR #2185, opened 2017-02-23
Dockerization of build+mysql+server

ranisalt:
So, this is continuing #2157 and will have to be done in two steps:

- First, I'll merge the dockerfiles into master, so that we can have automated builds for both a compilation container and a server container that runs the compiled files. This is because pulling an Ubuntu image then installing packages is utterly slow and timeouts in Travis, and is probably suboptimal for usage.

- Then, after automated builds are working and we can pull an image from Docker Hub, and use in Travis to build and test the server. I added two separate dockerfiles, one for a compilation container that installs development files and is huge, and one for running the server which only installs dependencies, and is isolated.

So, this PR as is is not to be merged, and if it receives enough positive feedback I'll do as above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ sudo: required
 dist: trusty
 language: generic
 env:
-  global:
-    - CONTAINER=forgottenserver
   matrix:
     - LUAJIT=OFF CXX=g++
     - LUAJIT=ON  CXX=g++
@@ -11,9 +9,11 @@ env:
     - LUAJIT=ON  CXX=clang++
 services:
   - docker
-install: docker build -t ${CONTAINER} .
-before_script: mkdir ${TRAVIS_BUILD_DIR}/build
+before_script:
+  - export MYSQL_PASS="$(openssl rand -base64 32)"
+  - docker-compose up -d
 script:
-  - DOCKER_PARAMS="-e CXX=${CXX} -v ${TRAVIS_BUILD_DIR}:/opt -w /opt/build ${CONTAINER}"
-  - docker run ${DOCKER_PARAMS} cmake -DCMAKE_BUILD_TYPE=Release -DUSE_LUAJIT=${LUAJIT} ..
-  - docker run ${DOCKER_PARAMS} make -j2
+  - docker-compose run compile cmake -DUSE_LUAJIT=${LUAJIT} /source
+  - docker-compose run compile make -j2
+  - docker-compose run server --dry-run
+after_script: docker-compose down
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ services:
   - docker
 before_script:
   - export MYSQL_PASS="$(openssl rand -base64 32)"
-  - docker-compose up -d
+  - docker-compose up -d db
 script:
   - docker-compose run compile cmake -DUSE_LUAJIT=${LUAJIT} /source
   - docker-compose run compile make -j2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,19 @@
 sudo: required
 dist: trusty
-language: cpp
-compiler:
-  - clang
-  - gcc
+language: generic
 env:
-  - LUAJIT=OFF
-  - LUAJIT=ON
-cache: apt
-addons:
-  apt:
-    packages:
-      - libboost-dev
-      - libboost-system-dev
-      - libboost-iostreams-dev
-      - liblua5.2-dev
-      - libluajit-5.1-dev
-      - libmysqlclient-dev
-      - libpugixml-dev
-before_script:
-  - mkdir build && cd build
-  - cmake -DCMAKE_BUILD_TYPE=Release -DUSE_LUAJIT=${LUAJIT} ..
-script: make -j2
+  global:
+    - CONTAINER=forgottenserver
+  matrix:
+    - LUAJIT=OFF CXX=g++
+    - LUAJIT=ON  CXX=g++
+    - LUAJIT=OFF CXX=clang++
+    - LUAJIT=ON  CXX=clang++
+services:
+  - docker
+install: docker build -t ${CONTAINER} .
+before_script: mkdir ${TRAVIS_BUILD_DIR}/build
+script:
+  - DOCKER_PARAMS="-e CXX=${CXX} -v ${TRAVIS_BUILD_DIR}:/opt -w /opt/build ${CONTAINER}"
+  - docker run ${DOCKER_PARAMS} cmake -DCMAKE_BUILD_TYPE=Release -DUSE_LUAJIT=${LUAJIT} ..
+  - docker run ${DOCKER_PARAMS} make -j2
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,42 @@
+version: '2'
+
+services:
+  compile:
+    build:
+      context: ${PWD}/docker
+      dockerfile: Dockerfile.compile
+    environment:
+      - CXX
+    volumes:
+      - build:/build
+      - ${TRAVIS_BUILD_DIR}:/source
+    working_dir: /build
+  db:
+    environment:
+      - MYSQL_DATABASE=forgottenserver
+      - MYSQL_USER=forgottenserver
+      - MYSQL_PASSWORD=forgottenserver
+      - MYSQL_RANDOM_ROOT_PASSWORD=yes
+    image: mariadb
+    restart: always
+    volumes:
+      - ${TRAVIS_BUILD_DIR}/schema.sql:/docker-entrypoint-initdb.d/schema.sql
+  server:
+    build:
+      context: ${PWD}/docker
+      dockerfile: Dockerfile.server
+    depends_on:
+      - db
+    environment:
+      - MYSQL_HOST=db
+      - MYSQL_PASS=forgottenserver
+    volumes:
+      - build:/build
+      - ${TRAVIS_BUILD_DIR}/config.lua:/build/config.lua
+      - ${TRAVIS_BUILD_DIR}/data:/build/data
+    working_dir: /build
+
+volumes:
+  build:
+
+# vim: set et sw=2 ts=2:
--- a/docker/Dockerfile.compile
+++ b/docker/Dockerfile.compile
@@ -1,0 +1,14 @@
+FROM ubuntu:xenial
+
+RUN apt-get update && apt-get install -y \
+  cmake make \
+  clang g++ \
+  libboost-iostreams-dev libboost-system-dev \
+  libgmp-dev \
+  liblua5.2-dev libluajit-5.1-dev \
+  libmysqlclient-dev \
+  libpugixml-dev
+
+WORKDIR /build
+
+# vim: set et sw=2 ts=2 syntax=dockerfile:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,0 @@
-FROM ubuntu:xenial
-
-RUN apt-get update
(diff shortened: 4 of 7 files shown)


kornholi:
Why not use an Alpine base instead of Ubuntu? It's a lot smaller.

ranisalt:
I'd also like review from @gugahoa and @Tarjei400

kornholi:
[mariadb-client](https://pkgs.alpinelinux.org/package/v3.4/main/x86_64/mariadb-client ) [mariadb-clientlibs](https://pkgs.alpinelinux.org/package/v3.4/main/x86_64/mariadb-client-libs)?

ranisalt:
I'm pretty sure those packages are missing development headers.

Tarjei400:
I will take a look on that after work, I took a glance at it. Great job :)

.travis.yml
```diff
@@ -1,24 +1,19 @@
 sudo: required
 dist: trusty
```
Tarjei400: Why trusty, is it even needed in docker compose?

ranisalt: Not aure, it's just a change-the-least-possible basis really :P

DSpeichert:
@ranisalt While I see how your plan would work together (the MySQL container, the "build" container and the "run" container), I'm not convinced if this is a useful combination and therefore if it can have the potential of becoming popular.

First of all, what are the benefits of compiling TFS in Docker? If the environment was OS-invasive or difficult to create, we could publish a container for compilation. However, TFS built in such a container can only be run in the "run" container. Since you now require both of them at the same time (assuming you want to compile your TFS because you make changes), I don't see any benefit over having one container to do both these things (the counter-argument used to be the large size of the "build" container, but you still need it). Typically, the reason to build/run in a container would be if we didn't trust the source or wanted to limit its resources. I don't see that being quite applicable here.

I think that using a very specific environment in builds in Travis CI (if you used the pre-published docker "build" container) would actually make the testing less thorough. First, you picked the "trusty" flavor, which not only has older boost (and other) libraries, but is also less commonly (hopefully) installed recently. Also, you would be freezing the libraries to a specific version (where instead on Ubuntu you'd always be testing the newest incremental sub-version of the library - e.g. libboost-*). This may lead to the unlikely scenario where a build in Travis CI succeeds in Docker container but fails on modern/updated Ubuntu machine.

I would see the advantage of using Docker is simply the installation of all dependencies took too long in Travis CI, especially when you need to pull a lot of additional libraries for the HTTP feature in #2010. Instead of tightening the range of tested dependencies, I would be testing on Arch Linux (always newest libraries) in addition to Ubuntu (typically more "stable", older versions of libraries). Is Docker the solution to improving build times for Travis CI? Is that the primary motivation of doing this PR?

From my point of view, the usability of the addition of a Dockerfile would be if we had a pretty "standard" build/deployment type of scenario with minimal image size. Therefore, I would favor Alpine or Arch or any other OS that results in smallest binary size. The Dockerfile in such a case would be included so a user modifying the source would be using Docker "build" process to prepare OS, libraries, build the TFS, remove *-dev packages to decrease size and output the image with the binary inside which allows mounting /data and config.lua. I would expect the user to do this every time the modify source (and need to rebuild) and the output would be a portable container image with the binary and matching libs/OS that can be e.g. published in the private/public Docker hub and moved between Docker hosts.

I appreciate the possibility to set certain config.lua options from environment variables. It's really good for automation and I wish I had it earlier for the Hosting Service. However, I think the idea of doing a dry run with MySQL is a stretch. I assume it tests the datapack (errors would be thrown) and the database schema and upgrades (from data/migrations) but besides that it increases the complexity of the Travis CI build/test (if that was supposed to run there) assuming of course Travis even permits running multiple containers with Docker-compose.

gugahoa:
I think the combination here is good for the following purposes:
- Unified envirovment for building. This would help OT Developers that use Windows, for sure.
- Latest build on a public repository. Most of OT Developers does not modify the source, so a container that builds and runs is not ideal as they would prefer to skip the build step.
- Unified envirvoment for running. If everyone is running on the same envirovment, testing issues and confirming if they're really valid is simplified a lot.

For sure, we can have only one image and run the commands against it, but it would add another step for users that are seeing Docker for the first time.

I'm tempted to use this on my server, and will be experimenting over the weekend.

ranisalt:
> First of all, what are the benefits of compiling TFS in Docker?

Mostly it's much more standard, comparing installing Docker & spinning up containers vs. installing deps and compiling on every other OS out there. Also, it would provide sane, default isolation between server, database and possibly an AAC running together.

> Typically, the reason to build/run in a container would be if we didn't trust the source or wanted to limit its resources.

Not only. Once a build container finishes compiling the server, you can spin up multiple server containers with the same executable but varying (or not) config/data, so it's also easier to setup multiple servers (though I admit this was unintended side effect).


> you picked the "trusty" flavor

I actually use xenial image, the latest LTS. It could be `ubuntu:latest` for auto-updating LTS but I prefer a manual approach here. It's irrelevant now, as I'm preparing Alpine-based images that are much lighter and faster to build.

> The Dockerfile in such a case would be included so a user modifying the source would be using Docker "build" process to prepare OS, libraries, build the TFS, remove *-dev packages to decrease size and output the image with the binary inside which allows mounting /data and config.lua.

I'm not sure it works that way, removing -dev packages won't slim the image because Docker has to store all layers, including the ones containing -dev packages. That's the primary reason for split build+run images, especially because g++ and boost headers are elephants.

> However, I think the idea of doing a dry run with MySQL is a stretch. I assume it tests the datapack (errors would be thrown) and the database schema and upgrades (from data/migrations)

The dry run is not meant to be used elsewhere, only for CI here, as it "guarantees" the schema is correct and the default data pack and config.lua do not contain typos and simpler errors, such as file name issues. It simply loads everything, then shuts down, that's why its not even the default command.

DSpeichert:
@gugahoa 

> Unified envirovment for building. This would help OT Developers that use Windows, for sure.

I seriously doubt that anyone on Windows would prefer to use Docker to build/run TFS. MSVS is not that hard to set up now (read the updated guide on wiki, it's really straightforward, just 3 steps). On the other hand, last time I tried to get Docker to work on Windows, I failed. Also, when you build it in Docker you have to keep using it through docker.

> Latest build on a public repository. Most of OT Developers does not modify the source, so a container that builds and runs is not ideal as they would prefer to skip the build step.

This is actually completely opposite to what @ranisalt claimed in #2157. While I agree with you that most users don't modify the source code, I still think that a single container could provide the executable inside and be more convenient.

> Unified envirvoment for running. If everyone is running on the same envirovment, testing issues and confirming if they're really valid is simplified a lot.

While usually a unified environment is good, I think TFS is in a state where it's working flawlessly in many different environments and I don't think that a variation of environments is the source of any problems (not sure if we had bug reports originating from faulty/mis-matching libraries, I'd doubt that)

> For sure, we can have only one image and run the commands against it, but it would add another step for users that are seeing Docker for the first time.

I don't understand your point.

@ranisalt 
> Mostly it's much more standard, comparing installing Docker & spinning up containers vs. installing deps and compiling on every other OS out there. Also, it would provide sane, default isolation between server, database and possibly an AAC running together.

I am not a huge fan of "standardizing" the environment here as I explained above. To me it's the first step in the wrong direction of only testing TFS in one environment. I do agree that isolation is good when TFS is run in production. I also would run TFS and MySQL in separate containers. At this point I simply believe that the "build" container is unnecessary and the docker-compose recipe shouldn't belong in the repo here (merely the Dockerfile for the TFS container). There are many ways operators could run TFS in Docker and making "universal" docker-compose recipe would most likely lead to bad assumptions.

> Not only. Once a build container finishes compiling the server, you can spin up multiple server containers with the same executable but varying (or not) config/data, so it's also easier to setup multiple servers (though I admit this was unintended side effect).

Which would be even easier if we had the executable inside the container image.

> I'm not sure it works that way, removing -dev packages won't slim the image because Docker has to store all layers, including the ones containing -dev packages. That's the primary reason for split build+run images, especially because g++ and boost headers are elephants.

You are right that you can't slim an image across layers. So if you tried to uninstall packages that already existed in xenial, you wouldn't free any space. However, when in a single Dockerfile (being a single layer) you install packages and then remove them, the removed packages won't be wasting any space. So that would be my preferred approach - do everything in one step and just leave the binary and libraries installed without the dev-tools.

If you get it to work on Alpine - that's even better.

A counter-argument would be that every time you rebuild docker image, you need to reinstall the dev tools only to uninstall them a minute later. That is true. However, if you really wanted to have a container just for builds, I would prefer to have another Docker container image named `forgottenserver-sdk` and the name pretty much says it all - it is the build environment without a binary/source/datapack inside. Still, I would have Travis CI build the first one and also have the first one published in Docker Hub for every commit and the second one would probably be more static (only updated as the SDK or lib packages in this case are updated).

ranisalt:
> A counter-argument would be that every time you rebuild docker image, you need to reinstall the dev tools only to uninstall them a minute later.

Thinking again, for a production container it is not that bad as it won't really change that much, you'll only redownload and rebuild if you change the source code. Datapack and config.lua can be loaded with volumes, as I do here.

> I would prefer to have another Docker container image named forgottenserver-sdk and the name pretty much says it all - it is the build environment without a binary/source/datapack inside. Still, I would have Travis CI build the first one and also have the first one published in Docker Hub for every commit and the second one would probably be more static (only updated as the SDK or lib packages in this case are updated).

That is actually a very good idea. I was thinking on maintaining a `sdk` tag with the latest development container and `latest` tag with the latest commit build.

DSpeichert:
Yes, tags would be great for that purpose.

Also, as a bonus, I would make the "production" image detect if the path to config.lua and /data is empty and if it is, it could copy the default datapack in its place. That's a simple bash script to embed in Dockerfile. Default datapack isn't large and it would save the hassle of finding a matching datapack if you use a specific (not newest) container image.

docker/Dockerfile.compile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:xenial
+
+RUN apt-get update && apt-get install -y \
+  cmake make \
+  clang g++ \
+  libboost-iostreams-dev libboost-system-dev \
+  libgmp-dev \
+  liblua5.2-dev libluajit-5.1-dev \
```
Tarjei400: I dont know if its just me or luajit is completely ignore with setup like this. Cmake uses lua 5.2 only anyway I think

ranisalt: To me it is actually inverse, if I compile with LuaJIT installed it ignores Lua52

Tarjei400: Hmm yea I had bad cmake file now its for me like you say.

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+      context: ${PWD}/docker
+      dockerfile: Dockerfile.compile
+    environment:
+      - CXX
+    volumes:
+      - build:/build
+      - ${TRAVIS_BUILD_DIR}:/source
+    working_dir: /build
+  db:
+    environment:
```
Tarjei400: On Necronia I made it this way to wait till build container is eventualy gone : 
1. Linked server container with build container, then added this code to entry point script

###########################
##Check if server is build, is so postpone start up
###########################
BUILD_CONTAINER=necronia-build

ping -c 1 $BUILD_CONTAINER;
while [ "$?" -eq 0 ]; do echo 'Server is still building...'; sleep 15; ping -c 1 $BUILD_CONTAINER; done;

DSpeichert:
@ranisalt Do we still need any of this? Or is the current `Dockerfile` superseding this?

ranisalt:
Partly, the Dockerfile is a part but docker-compose is still useful, but it's better to start again than rebase :P

PR #2187, opened 2017-02-26
Refactored and moved to its own method code responsible for sending VIP entries

elderapo:
Refactored and moved to its own method code responsible for sending VIP entries.

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2675,19 +2675,17 @@ void ProtocolGame::sendVIP(uint32_t guid, const std::string& name, const std::st
 	writeToOutputBuffer(msg);
 }
 
-void ProtocolGame::sendVIPEntries() {
+void ProtocolGame::sendVIPEntries()
+{
 	const std::forward_list<VIPEntry>& vipEntries = IOLoginData::getVIPEntries(player->getAccount());
 
 	for (const VIPEntry& entry : vipEntries) {
-		VipStatus_t vipStatus;
+		VipStatus_t vipStatus = VIPSTATUS_ONLINE;
 
 		Player* vipPlayer = g_game.getPlayerByGUID(entry.guid);
 
 		if (!vipPlayer || vipPlayer->isInGhostMode() || player->isAccessPlayer()) {
 			vipStatus = VIPSTATUS_OFFLINE;
-		}
-		else {
-			vipStatus = VIPSTATUS_ONLINE;
 		}
 
 		sendVIP(entry.guid, entry.name, entry.description, entry.icon, entry.notify, vipStatus);
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2424,35 +2424,7 @@ void ProtocolGame::sendAddCreature(const Creature* creature, const Position& pos
 	//player light level
 	sendCreatureLight(creature);
 
-	const std::forward_list<VIPEntry>& vipEntries = IOLoginData::getVIPEntries(player->getAccount());
-
-	if (player->isAccessPlayer()) {
-		for (const VIPEntry& entry : vipEntries) {
-			VipStatus_t vipStatus;
-
-			Player* vipPlayer = g_game.getPlayerByGUID(entry.guid);
-			if (!vipPlayer) {
-				vipStatus = VIPSTATUS_OFFLINE;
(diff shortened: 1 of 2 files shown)


ranisalt:
Why do we use a forward_list? It is worse than a vector space-wise and slower because iteration is not over contiguous memory. It should be change because the collection is not changed.

src/protocolgame.cpp
```diff
@@ -2703,6 +2675,25 @@ void ProtocolGame::sendVIP(uint32_t guid, const std::string& name, const std::st
 	writeToOutputBuffer(msg);
 }
 
+void ProtocolGame::sendVIPEntries() {
```
jo3bingham: Move `{` to new line.

src/protocolgame.cpp
```diff
@@ -2703,6 +2675,25 @@ void ProtocolGame::sendVIP(uint32_t guid, const std::string& name, const std::st
...
+	const std::forward_list<VIPEntry>& vipEntries = IOLoginData::getVIPEntries(player->getAccount());
+
+	for (const VIPEntry& entry : vipEntries) {
+		VipStatus_t vipStatus;
```
jo3bingham: May as well assign `VIPSTATUS_ONLINE` to `vipStatus` here, and remove the `else` statement below.

src/protocolgame.cpp
```diff
@@ -2703,6 +2675,25 @@ void ProtocolGame::sendVIP(uint32_t guid, const std::string& name, const std::st
...
+
+		Player* vipPlayer = g_game.getPlayerByGUID(entry.guid);
+
+		if (!vipPlayer || vipPlayer->isInGhostMode() || player->isAccessPlayer()) {
+			vipStatus = VIPSTATUS_OFFLINE;
+		}
+		else {
```
jo3bingham: If you prefer to keep this `else` statement (see my comment above), please move it up to the previous line (`} else {`).

djarek:
@ranisalt 
I agree about the forward list, however, I think that should be a separate patch.

jo3bingham:
@ranisalt I'm going to assume that, whomever decided to use forward_list here did so because we only ever traverse forward through this list (which saves space since it only points to the next element and it doesn't contain a size member).

ranisalt:
@jo3bingham I don't think so, a size member is only as big as a pointer, as the function is inlined, so there is absolutely no advantage on having a forward list over a vector *in this case* if the list ever has an element. See:

For 0 elements: forward_list has 1 pointer, vector has 1 pointer and 1 size_t
For 1 element: forward_list has 2 pointers and 1 T, vector has 1 pointer, 1 T and 1 size_t (the same)
For 2 or + elements: forward_list has N+1 pointers and N T, vector has 1 pointer, N T and 1 size_t (vector has N-1 less pointer-sized elements)

Of course, this totally disregard other internal variables, I'm only considering the simplest possible implementation.

djarek:
@ranisalt 
You forgot to mention that general-purpose allocators are pretty bad space-wise in allocating small blocks of memory (in a good allocator you'll have 1 pointer of overhead per allocated block, however, you usually can't allocate less than a fixed size, IIRC in glibc it's 16 bytes). This makes node based containers look really bad both space and performance wise when container traversal is more frequent than insertion/deletion. There are few places were lists *might* be better than contiguous memory storage - when used as queues(pop/push_back/front) or when you do splicing. There are a few places in tfs where forward_list is used to save space when the list itself is heap allocated and represents a container that's usually empty. These should probably be replaced by a "vector" that stores most of its control block along with the data.

src/protocolgame.h
```diff
@@ -225,6 +225,7 @@ class ProtocolGame final : public Protocol
 
 		void sendUpdatedVIPStatus(uint32_t guid, VipStatus_t newStatus);
 		void sendVIP(uint32_t guid, const std::string& name, const std::string& description, uint32_t icon, bool notify, VipStatus_t status);
+		void ProtocolGame::sendVIPEntries();
```
djarek: Remove ProtocolGame:: from this line

jo3bingham:
@ranisalt I was only pointing out a possible reason why someone would think using a forward_list here would be ideal. I agree that the use of a vector would be better.

djarek:
Thanks!

PR #2188, opened 2017-03-03
Removed spell Levitate from C++

raymondtfr:
Made this new PR as requested by @ranisalt in #2081 .

Should the **actual** removing of the spell levitate be here? or just the function loading part (as you indeed told in #2081 ?), I think I have mislead this.

### DIFF
--- a/src/spells.cpp
+++ b/src/spells.cpp
@@ -935,15 +935,8 @@ bool Levitate(const InstantSpell*, Creature* creature, const std::string& param)
 bool InstantSpell::loadFunction(const pugi::xml_attribute& attr)
 {
 	const char* functionName = attr.as_string();
-	if (strcasecmp(functionName, "levitate") == 0) {
-		function = Levitate;
-	} else {
-		std::cout << "[Warning - InstantSpell::loadFunction] Function \"" << functionName << "\" does not exist." << std::endl;
-		return false;
-	}
-
-	scripted = false;
-	return true;
+	std::cout << "[Warning - InstantSpell::loadFunction] Function \"" << functionName << "\" does not exist." << std::endl;
+	return false;
 }
 
 bool InstantSpell::playerCastInstant(Player* player, std::string& param)
@@ -1140,8 +1133,6 @@ bool InstantSpell::internalCastSpell(Creature* creature, const LuaVariant& var)
 {
 	if (scripted) {
 		return executeCastSpell(creature, var);
-	} else if (function) {
-		return function(this, creature, var.text);
 	}
 
 	return false;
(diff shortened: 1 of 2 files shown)


djarek:
@raymondtfr 
I think that this PR should remove all the dead code related to the Levitate spell, so it's fine as is.

ranisalt:
I meant you should keep the code for the `loadFunction` function just issuing a warning (i.e. remove the last if but keep the rest).

src/spells.cpp
```diff
@@ -884,68 +884,6 @@ bool InstantSpell::configureEvent(const pugi::xml_node& node)
...
-}
-
-}
-
-bool InstantSpell::loadFunction(const pugi::xml_attribute& attr)
-{
-	const char* functionName = attr.as_string();
-	if (strcasecmp(functionName, "levitate") == 0) {
-		function = Levitate;
-	} else {
```
ranisalt: What I said is to keep this function here, remove the if and just keep the else branch here:

```cpp
bool InstantSpell::loadFunction(const pugi::xml_attribute& attr)
{
	const char* functionName = attr.as_string();
	std::cout << "[Warning - InstantSpell::loadFunction] Function \"" << functionName << "\" does not exist." << std::endl;
	return false;
}
```

The rest is ok and you can add to the other PR.

src/spells.cpp
```diff
@@ -1140,8 +1078,6 @@ bool InstantSpell::internalCastSpell(Creature* creature, const LuaVariant& var)
 {
```
ranisalt: The changes below here can also be undone.

src/spells.cpp
raymondtfr: Even though it doesn't have any function now (when Levitate is moved to Lua.)?

WibbenZ:
Yes for ppl who havent updated.
But IMO print that the functions have been moved to Lua or smth, if we for some reason add C++ functions again we can edit the string.

src/spells.cpp
ranisalt: Yes, then you can open a PR removing leftovers if you want, or I can do it as soon as I merge.

ranisalt:
I don't think it will be moved back to C++ ever, but we can overload the `loadFunction` in the base `Event` class to issue a warning.

src/spells.cpp
raymondtfr: Ok, then I updated the other PR with removing only the function Levitate enclosed in a namespace.

And here I'll just do what you suggested.

ranisalt:
As both PRs and a review was pending weren't compiling I went ahead and commited to master with fixes.

PR #2189, opened 2017-03-04
Move WaitingList attributes to private implementation

ranisalt:
This is a different idiom I intend to apply on classes with huge loads of attributes (e.g. `Player` and `Monster`). It already is applied somewhat in `NetworkMessage` too.

Moving attributes to a implementation pointer has the effect of reducing pollution of global namespace at the cost of one more indirection in memory access.

There are also small fixes on attribute and return types.

### DIFF
--- a/src/waitlist.cpp
+++ b/src/waitlist.cpp
@@ -29,7 +29,8 @@ extern Game g_game;
 
 namespace {
 
-struct Wait {
+struct Wait
+{
 	constexpr Wait(std::size_t timeout, uint32_t playerGUID) :
 			timeout(timeout), playerGUID(playerGUID) {}
 
@@ -59,9 +60,10 @@ std::size_t getTimeout(std::size_t slot)
 	return WaitingList::getTime(slot) + 15;
 }
 
-}
+} // namespace
 
-struct WaitListInfo {
+struct WaitListInfo
+{
 	WaitList priorityWaitList;
 	WaitList waitList;
 
@@ -82,8 +84,8 @@ struct WaitListInfo {
 	}
 };
 
-
-WaitingList& WaitingList::getInstance() {
+WaitingList& WaitingList::getInstance()
+{
 	static WaitingList waitingList;
 	return waitingList;
 }
--- a/src/waitlist.cpp
+++ b/src/waitlist.cpp
@@ -26,26 +26,69 @@
 extern ConfigManager g_config;
 extern Game g_game;
 
-WaitListIterator WaitingList::findClient(const Player* player, uint32_t& slot)
-{
-	slot = 1;
-	for (auto it = priorityWaitList.begin(), end = priorityWaitList.end(); it != end; ++it) {
-		if (it->playerGUID == player->getGUID()) {
-			return it;
-		}
-		++slot;
-	}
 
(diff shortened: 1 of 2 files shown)


src/waitlist.cpp
```diff
@@ -26,26 +26,69 @@
...
 	}
+}
+
+std::size_t getTimeout(std::size_t slot)
+{
+	//timeout is set to 15 seconds longer than expected retry attempt
+	return WaitingList::getTime(slot) + 15;
+}
+
+}
```
djarek: Add comment //namespace to mark the end of an anonymous namespace

src/waitlist.cpp
```diff
@@ -26,26 +26,69 @@
 extern ConfigManager g_config;
 extern Game g_game;
 
-WaitListIterator WaitingList::findClient(const Player* player, uint32_t& slot)
+
+namespace {
+
+struct Wait {
```
djarek: Brace to next line

src/waitlist.cpp
```diff
@@ -26,26 +26,69 @@
...
+				return {it, slot};
+			}
+		}
+		return {waitList.end(), slot};
 	}
-	return waitList.end();
+};
+
+
+WaitingList& WaitingList::getInstance() {
```
djarek: Brace to next line

src/waitlist.cpp
```diff
@@ -95,36 +134,25 @@ bool WaitingList::clientLogin(const Player* player)
...
 
-	slot = priorityWaitList.size();
+	slot = info->priorityWaitList.size();
 	if (player->isPremium()) {
-		priorityWaitList.emplace_back(OTSYS_TIME() + (getTimeout(slot + 1) * 1000), player->getGUID());
+		info->priorityWaitList.emplace_back(OTSYS_TIME() + (getTimeout(slot + 1) * 1000), player->getGUID());
 	} else {
-		slot += waitList.size();
-		waitList.emplace_back(OTSYS_TIME() + (getTimeout(slot + 1) * 1000), player->getGUID());
+		slot += info->waitList.size();
```
ranisalt: Hey, I think this is wrong. Why is the free account player slot added the size of the free waitlist? Shouldn't it be added the size of the PREMIUM waitlist, i.e. the first free player slot is immediately after the last premium player?

djarek: Perhaps we should use http://en.cppreference.com/w/cpp/container/priority_queue ?:D

ranisalt: Not really related, but a list specifically has constant complexity for every single operation we use (`emplace_back` and `erase`), even though it is horribly managed on memory.

A priority queue would do little more than two queues here, just spare a pointer and a size value.

ranisalt: Well, I'll keep it for later

(and skip your approval 😝 )

djarek: I was just asking about the queue. Any performance gains or losses are unlikely to have any significant effect here, so w/e is simpler should be used ;).

ranisalt: Yeah, of course, a priority queue seems to be the optimal container if the queue wasn't jsut so specific. Indeed using a PQ could allow for more complex algorithms for queueing :)

PR #2191, opened 2017-03-07
Fix #2190

### LINKED ISSUE #2190: Summons crashing

1. Summon any monster
2. Logout

### Expected behaviour
Player logout and summon disappear.

### Actual behaviour
Crash

### Environment
Latest build of TFS 1.3

Related to #2173 but it is crashing for another reason.

Backtrace:
http://pastebin.com/z6nM8Jsp

Mkalo:
This should stop the crashes. There might be a better solution.
Fix for #2190

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1131,15 +1131,16 @@ bool Creature::setMaster(Creature* newMaster) {
 		newMaster->summons.push_back(this);
 	}
 
-	if (master) {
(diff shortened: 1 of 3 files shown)


djarek:
Yeah there is a better [solution](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization), but that's enough work for 4-5 PRs  ;).

PR #2195, opened 2017-03-09
Create clang-format config file

djarek:
Due to inconsistencies in code style we have introduce tooling to end
their existence once and for all. I've tried to mimic existing style as closely as possible to minimize required changes. I propose we reformat gradually to reduce the chaos that occurs :D.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,89 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: 0
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: false
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterClass:      true
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+ColumnLimit:     120
+CommentPragmas:  '^ IWYU pragma:'
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeCategories:
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
+    Priority:        2
+  - Regex:           '^(<|"(gtest|isl|json)/)'
+    Priority:        3
+  - Regex:           '.*'
+    Priority:        1
+IndentCaseLabels: true
+IndentWidth:     -2
+IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: None
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left
+ReflowComments:  true
+SortIncludes:    true
+SpaceAfterCStyleCast: false
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        2
+UseTab:          Always
+...
--- a/.clang-format
+++ b/.clang-format
@@ -36,7 +36,7 @@ BreakBeforeBinaryOperators: None
 BreakBeforeBraces: Custom
 BreakBeforeTernaryOperators: true
 BreakConstructorInitializersBeforeComma: false
-ColumnLimit:     120
+ColumnLimit:     160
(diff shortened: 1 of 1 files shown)


.clang-format
```diff
@@ -0,0 +1,89 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        2
```
ranisalt: What does that mean?

.clang-format
```diff
@@ -0,0 +1,89 @@
...
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left
+ReflowComments:  true
+SortIncludes:    true
```
ranisalt: We usually don't do that, mostly we sort as:
- PCH
- STL
- system headers
- project headers

.clang-format
```diff
@@ -0,0 +1,89 @@
...
+IndentWidth:     -2
+IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: None
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
```
ranisalt: Remove this ObjC configuration as it's not applied.

.clang-format
```diff
@@ -0,0 +1,89 @@
...
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+ColumnLimit:     120
```
ranisalt: Why? I usually keep it below 80 for two tiled screens side by side.

ranisalt:
It should make the minimal amount of changes to the current source code, did you apply against our current source?

ranisalt:
I also think we should adequate code to the [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines)

.clang-format
djarek: I think all indentation in clang-format is calculated using spaces and if you choose to use tabs you have to specify how many tabs to put per indent. Long story short if this isn't smaller or equal to indent width weird things happen.

.clang-format
djarek: The ordering of includes can be configured in IncludeCategories. Though you're right, this is likely to break the build in some cases.

.clang-format
djarek: My experience so far with the 80 character limit is that most people try to work around it by choosing shorter identifier names. Overall I agree that short columns are easier to read but the gain is far outweighed by the fact that people produce stuff like:
```
ret = add_chr_to_srv_inc_c(s, c, TRUE, FALSE, TRUE);
```
I've also noticed that 120 made people worry less about fitting in the line and care more about code readability. My findings may be skewed due to small test group and the fact that people who follow the 80 character rule are quite often very anal and religious about it (to the point where common sense and logic can't touch them).

djarek:
Ok, no [problem](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#a-namerl-camelanl10-avoid-camelcase) :D Introducing more checks from core guidelines(clang-tidy) will have to be done in a separate PR. I did run it agains some shorter source files and it look sort of fine. I'll try to run it on some abomination like protocolgame and see what happens :D

.clang-format
ranisalt: Well, that's why I'm so pedantic with names on PRs... but I agree.

.clang-format
ranisalt: Can't it be removed with similar effect?

djarek:
I've checked the config on protocolgame.cpp and had to make some minor adjustments.

ranisalt:
Why don't we base on LLVM?

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+  AfterFunction:   true
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     true
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
```
ranisalt: We use the Stroustroup, in general.

djarek:
```
# BasedOnStyle:  LLVM
```
The template was generated based on LLVM, but I made changes to reflect the current style. Not sure if this is what you were asking for.

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+ColumnLimit:     120
+CommentPragmas:  '^ IWYU pragma:'
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
```
ranisalt: I'd like that we set it to true and follow. It definitely adds up readability.

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
```
ranisalt: This isn't used.

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+ColumnLimit:     120
+CommentPragmas:  '^ IWYU pragma:'
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat:   false
```
ranisalt: This isn't used, can be removed if defaults to true.

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+    Priority:        1
+IndentCaseLabels: true
+IndentWidth:     4
+IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: None
+PenaltyBreakBeforeFirstCallParameter: 19
```
ranisalt: Magic number?

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
```
ranisalt: Isn't that too harsh for current code base?

.clang-format
```diff
@@ -0,0 +1,86 @@
...
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        4
+UseTab:          Always
```
ranisalt: I prefer `ForIndentation`, for we use spaces for alignment after indentation.

ranisalt:
@djarek but enabling that setting won't base on LLVM and we only need to change what we have different, instead of having the entire format file?

djarek:
@ranisalt 
I'd prefer using one of the predefined styles, but that seems a little extreme :D. I'll see if choosing a base and overriding only some works properly.

ranisalt:
I personally don't fully like any of them :/

.clang-format
djarek: Stroustrup places else on a separate line:
```
if () {
}
else {
}

.clang-format
djarek: Looks better with 20. Can't get it less magical :D

.clang-format
djarek: No mercy to long lines.

djarek:
@ranisalt Anything else?

PR #2197, opened 2017-03-10
Remove lastWalkthrough* limiting

ranisalt:
It does not help limiting walkthrough attempts and, with the current codebase, uses a `const_cast` that invokes undefined behavior. As of #1764, it's better to remove this than try workarounds such as `mutable`.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -711,23 +711,7 @@ bool Player::canWalkthrough(const Creature* creature) const
 	}
 
 	const Item* playerTileGround = playerTile->getGround();
-	if (!playerTileGround || !playerTileGround->hasWalkStack()) {
-		return false;
-	}
-
-	Player* thisPlayer = const_cast<Player*>(this);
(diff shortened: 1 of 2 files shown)


conde2:
Just curious... What this was used for ?

ranisalt:
Rate limiting. It just plain ignores trying to walk through more than once every 2 seconds on the same tile. That's just so micro, nano, yocto optimization that it's not even worth the extra lots of variables.

djarek:
I guess this rate limiting is quite pointless. I some wanna has issues with people spamming commands they should consider bumping down the packetsPerSecond config value.

WibbenZ:
Might be totally off, but dosn't this remove the push back when trying to merge into the same tile as another player (pz)?
If so, then it's to 99% there because to mimic "how it should be".

Mkalo:
I agree with @WibbenZ, this is not only to rate limit. That's how it works in "you know what".

PR #2198, opened 2017-03-12
Change onManaChange event.

### LINKED ISSUE #2159: onManaChange(...) - cannot block part of damage with mana shield
I tried to make script that block 50% of spell damage. For health it works fine, but then I tried to make it work for players with mana shield:

```
function onManaChange(creature, attacker, manaChange, origin)
	if origin == ORIGIN_SPELL then
		manaChange = manaChange / 2
	end
	return manaChange
end
```

as you can read in this line:
https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L3984

It will execute 'onHealthChange' with health reduced by number of mana I blocked and deal that damage (with origin NONE) to.. mana as player still got manashield.

On my server I changed that line to:

`damage.primary.value = healthChange - manaDamage;`

but it will probably give bug with weapons that deal 'element damage'.

I hope you guys can re-think whole onHealthChange and onManaChange code and make it work in every possible scenario.

Mkalo:
Now onManaChange have the same parameters as onHealthChange and expects 4 return values as onHealthChange. 
That makes it more flexible to check damage type and secondary damage info.

If the combat is COMBAT_MANADRAIN:
Positive values for primaryDamage and secondaryDamage are going to heal your mana, while negative values are going to decrease your mana.
Else:
The damage values are always positive and its going to decrease your mana.

This could be fixed with a COMBAT_MANAHEALING, but it could have side effects in old Lua scripts that uses COMBAT_MANADRAIN and expect it to heal mana.

Closes #2159


conde2:
While you are at this could you add manaChange when casting spells ?

src/creatureevent.cpp
```diff
@@ -546,15 +545,22 @@ void CreatureEvent::executeManaChange(Creature* creature, Creature* attacker, in
 		lua_pushnil(L);
 	}
 
-	lua_pushnumber(L, manaChange);
-	lua_pushnumber(L, origin);
+	lua_pushnumber(L, damage.primary.value);
+	lua_pushnumber(L, damage.primary.type);
+	lua_pushnumber(L, damage.secondary.value);
+	lua_pushnumber(L, damage.secondary.type);
+	lua_pushnumber(L, damage.origin);
```
ranisalt: Create `pushDamage` function.

src/creatureevent.cpp
```diff
@@ -546,15 +545,22 @@ void CreatureEvent::executeManaChange(Creature* creature, Creature* attacker, in
...
+	if (scriptInterface->protectedCall(L, 7, 4) != 0) {
+		LuaScriptInterface::reportError(nullptr, LuaScriptInterface::popString(L));
 	} else {
-		manaChange = LuaScriptInterface::getNumber<int32_t>(L, -1);
+		damage.primary.value = LuaScriptInterface::getNumber<int32_t>(L, -4);
+		damage.primary.type = LuaScriptInterface::getNumber<CombatType_t>(L, -3);
+		damage.secondary.value = LuaScriptInterface::getNumber<int32_t>(L, -2);
+		damage.secondary.type = LuaScriptInterface::getNumber<CombatType_t>(L, -1);
+
+		lua_pop(L, 4);
```
ranisalt: Create `popDamage` function.

src/creatureevent.cpp
```diff
@@ -546,15 +541,13 @@ void CreatureEvent::executeManaChange(Creature* creature, Creature* attacker, in
...
-	lua_pushnumber(L, origin);
+	LuaScriptInterface::pushCombatDamage(L, damage);
 
-	if (scriptInterface->protectedCall(L, 4, 1) != 0) {
-		LuaScriptInterface::reportError(nullptr, LuaScriptInterface::getString(L, -1));
+	if (scriptInterface->protectedCall(L, 7, 4) != 0) {
+		LuaScriptInterface::reportError(nullptr, LuaScriptInterface::popString(L));
 	} else {
-		manaChange = LuaScriptInterface::getNumber<int32_t>(L, -1);
+		damage = LuaScriptInterface::getCombatDamage(L);
```
ninjalulz: Wouldn't it be possible to use this for executeHealthChange as well?

DSpeichert: Absolute values are being used there, so I don't think so.

PR #2203, opened 2017-03-19
Reformat all the things

djarek:
Added support for clang-format (```make clang-format```) and implemented integration with travis. Travis builds will now fail if the PR is improperly formatted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache: apt
 addons:
   apt:
     packages:
+      - clang-format-3.8
       - libboost-dev
       - libboost-system-dev
       - libboost-iostreams-dev
@@ -19,6 +20,7 @@ addons:
       - libmysqlclient-dev
       - libpugixml-dev
 before_script:
+  - ./verify-format.sh
   - mkdir build && cd build
   - cmake -DCMAKE_BUILD_TYPE=Release -DUSE_LUAJIT=${LUAJIT} ..
 script: make -j2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ endif()
 find_package(Boost 1.53.0 COMPONENTS system iostreams REQUIRED)
 
 add_subdirectory(src)
+include(ClangTools)
 add_executable(tfs ${tfs_SRC})
 
 include_directories(${MYSQL_INCLUDE_DIR} ${LUA_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${PUGIXML_INCLUDE_DIR} ${GMP_INCLUDE_DIR})
--- a/cmake/ClangTools.cmake
+++ b/cmake/ClangTools.cmake
@@ -1,0 +1,4 @@
+find_program(CLANG_FORMAT NAMES "clang-format")
+
+message(STATUS "clang-format found: ${CLANG_FORMAT}")
+add_custom_target(clang-format COMMAND ${CLANG_FORMAT} -i -style=file ${tfs_SRC} ${tfs_HDR})
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -75,4 +75,6 @@ set(tfs_SRC
 	${CMAKE_CURRENT_LIST_DIR}/weapons.cpp
 	${CMAKE_CURRENT_LIST_DIR}/wildcardtree.cpp
 	PARENT_SCOPE)
+file(GLOB_RECURSE tfs_HDR *.h)
 
+set(tfs_HDR ${tfs_HDR} PARENT_SCOPE)
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -32,8 +32,7 @@ extern Spells* g_spells;
 extern Actions* g_actions;
 extern ConfigManager g_config;
 
-Actions::Actions() :
-	scriptInterface("Action Interface")
+Actions::Actions() : scriptInterface("Action Interface")
 {
 	scriptInterface.initState();
 }
@@ -86,7 +85,7 @@ Event* Actions::getEvent(const std::string& nodeName)
 
 bool Actions::registerEvent(Event* event, const pugi::xml_node& node)
 {
-	Action* action = static_cast<Action*>(event); //event is guaranteed to be an Action
+	Action* action = static_cast<Action*>(event); // event is guaranteed to be an Action
 
 	pugi::xml_attribute attr;
 	if ((attr = node.attribute("itemid"))) {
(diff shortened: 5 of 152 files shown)


.travis.yml
```diff
@@ -19,6 +20,8 @@ addons:
       - libmysqlclient-dev
       - libpugixml-dev
 before_script:
+  - ./verify-format.sh
   - mkdir build && cd build
   - cmake -DCMAKE_BUILD_TYPE=Release -DUSE_LUAJIT=${LUAJIT} ..
-script: make -j2
+script:
```
ranisalt: Why?

CMakeLists.txt
```diff
@@ -16,7 +16,6 @@ if (CMAKE_COMPILER_IS_GNUCXX)
 endif()
 
 include(FindCXX11)
-
```
ranisalt: Why?

cmake/ClangTools.cmake
```diff
@@ -0,0 +1,10 @@
+find_program(CLANG_FORMAT NAMES "clang-format")
+
+if(CLANG_FORMAT)
+  message(STATUS "clang-format found: -i -style=file ${tfs_HDR}")
+  add_custom_target(
+    clang-format
+    COMMAND ${CLANG_FORMAT}
+    -i -style=file ${tfs_SRC} ${tfs_HDR}
+    )
+endif()
```
ranisalt: I don't think you should check for clang-format executable. Just let it fail if the executable doesn't exist.

src/CMakeLists.txt
```diff
@@ -75,4 +75,6 @@ set(tfs_SRC
 	${CMAKE_CURRENT_LIST_DIR}/weapons.cpp
 	${CMAKE_CURRENT_LIST_DIR}/wildcardtree.cpp
 	PARENT_SCOPE)
+file(GLOB_RECURSE tfs_HDR *.h)
```
ranisalt: If we use glob for headers, we should use for source files too, for the sake of consistency.

src/actions.cpp
```diff
@@ -110,14 +109,16 @@ bool Actions::registerEvent(Event* event, const pugi::xml_node& node)
 
 		auto result = useItemMap.emplace(iterId, action);
 		if (!result.second) {
-			std::cout << "[Warning - Actions::registerEvent] Duplicate registered item with id: " << iterId << " in fromid: " << fromId << ", toid: " << toId << std::endl;
+			std::cout << "[Warning - Actions::registerEvent] Duplicate registered item with id: " << iterId << " in fromid: " << fromId << ", toid: " << toId
+			          << std::endl;
```
ranisalt: SEXY!

src/actions.h
```diff
@@ -28,68 +28,68 @@ using ActionFunction = std::function<bool(Player* player, Item* item, const Posi
...
-		virtual ReturnValue canExecuteAction(const Player* player, const Position& toPos);
-		virtual bool hasOwnErrorHandler() {
-			return false;
-		}
-		virtual Thing* getTarget(Player* player, Creature* targetCreature, const Position& toPosition, uint8_t toStackPos) const;
+	virtual ReturnValue canExecuteAction(const Player* player, const Position& toPos);
+	virtual bool hasOwnErrorHandler()
+	{
+		return false;
+	}
```
ranisalt: Try to allow short functions as one-liners, it reduces clutter.

src/ban.h
```diff
@@ -20,15 +20,18 @@
...
 };
 
-struct ConnectBlock {
-	constexpr ConnectBlock(uint64_t lastAttempt, uint64_t blockTime, uint32_t count) :
-		lastAttempt(lastAttempt), blockTime(blockTime), count(count) {}
+struct ConnectBlock
+{
+	constexpr ConnectBlock(uint64_t lastAttempt, uint64_t blockTime, uint32_t count) : lastAttempt(lastAttempt), blockTime(blockTime), count(count)
+	{
+	}
```
ranisalt: :( this is so ugly

src/combat.cpp
```diff
@@ -1280,21 +1267,11 @@ void AreaCombat::setupArea(int32_t length, int32_t spread)
...
-		{0, 0, 8, 7, 6, 5, 5, 5, 6, 7, 8, 0, 0},
-		{0, 0, 0, 8, 7, 6, 6, 6, 7, 8, 0, 0, 0},
-		{0, 0, 0, 0, 8, 8, 7, 8, 8, 0, 0, 0, 0},
-		{0, 0, 0, 0, 0, 0, 8, 0, 0, 0, 0, 0, 0}
-	};
+	int32_t area[13][13] = {{0, 0, 0, 0, 0, 0, 8, 0, 0, 0, 0, 0, 0}, {0, 0, 0, 0, 8, 8, 7, 8, 8, 0, 0, 0, 0}, {0, 0, 0, 8, 7, 6, 6, 6, 7, 8, 0, 0, 0},
+	                        {0, 0, 8, 7, 6, 5, 5, 5, 6, 7, 8, 0, 0}, {0, 8, 7, 6, 5, 4, 4, 4, 5, 6, 7, 8, 0}, {0, 8, 6, 5, 4, 3, 2, 3, 4, 5, 6, 8, 0},
+	                        {8, 7, 6, 5, 4, 2, 1, 2, 4, 5, 6, 7, 8}, {0, 8, 6, 5, 4, 3, 2, 3, 4, 5, 6, 8, 0}, {0, 8, 7, 6, 5, 4, 4, 4, 5, 6, 7, 8, 0},
+	                        {0, 0, 8, 7, 6, 5, 5, 5, 6, 7, 8, 0, 0}, {0, 0, 0, 8, 7, 6, 6, 6, 7, 8, 0, 0, 0}, {0, 0, 0, 0, 8, 8, 7, 8, 8, 0, 0, 0, 0},
+	                        {0, 0, 0, 0, 0, 0, 8, 0, 0, 0, 0, 0, 0}};
```
ranisalt: This is pretty fucked up now

src/const.h
```diff
@@ -341,10 +318,11 @@ enum TextColor_t : uint8_t {
 	TEXTCOLOR_NONE = 255,
 };
 
-enum Icons_t {
+enum Icons_t
+{
 	ICON_POISON = 1 << 0,
 	ICON_BURN = 1 << 1,
-	ICON_ENERGY =  1 << 2,
+	ICON_ENERGY = 1 << 2,
```
ranisalt: How come @WibbenZ didn't see it yet

src/ban.h
djarek: Reducing column limit would probably help here.

ranisalt: I mean, it should be `constexpr ConnectBlock(uint64_t lastAttempt, uint64_t blockTime, uint32_t count) : lastAttempt(lastAttempt), blockTime(blockTime), count(count) {}` on a single line or `{}` on the next line.

src/actions.h
djarek: I'll set it so that inline function definitions will be one liners (if they fit).

src/combat.cpp
djarek: Should be better now.

src/CMakeLists.txt
djarek: We use the glob only for clang-format (and clang-tidy in the future). Using it for the build is a bad idea (screws up cmake's change detection system AFAIK).

Leo32onGIT:
Although its generic enough of an integration to have reference value, it'll just sit here forever.
The redundant branch is now clutter too, should be cleaned up _(make use of [readme](https://github.com/otland/forgottenserver/tree/clang_format))_ or removed.

PR #2205, opened 2017-03-20
Fix monsters melee conditions.

### LINKED ISSUE #1372: Poisonous enemies poisoning at first hit
Like the title says, i tested with snakes and poison spiders and they poison at the first hit, even if they fail.
screenshot_1
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/16607478-poisonous-enemies-poisoning-at-first-hit?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F392151&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mkalo:
Only apply melee conditions on successful hit.

This fixes #1372.

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -499,7 +499,7 @@ void Combat::CombatHealthFunc(Creature* caster, Creature* target, const CombatPa
 	}
 
 	if (g_game.combatChangeHealth(caster, target, damage)) {
-		CombatConditionFunc(caster, target, params, nullptr);
+		CombatConditionFunc(caster, target, params, &damage);
 		CombatDispelFunc(caster, target, params, nullptr);
 	}
 }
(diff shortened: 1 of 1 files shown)


ranisalt:
Ty m8

src/combat.cpp
```diff
@@ -520,8 +520,12 @@ void Combat::CombatManaFunc(Creature* caster, Creature* target, const CombatPara
 	}
 }
 
-void Combat::CombatConditionFunc(Creature* caster, Creature* target, const CombatParams& params, CombatDamage*)
+void Combat::CombatConditionFunc(Creature* caster, Creature* target, const CombatParams& params, CombatDamage* data)
 {
+	if (params.origin == ORIGIN_MELEE && data && data->primary.value == 0 && data->secondary.value == 0) {
```
raymondtfr: Shouldn't this first check if the *attacker* was a monster? otherwise, this may have unintended side-effects?

Mkalo: I don't see why this would have side-effects, I don't think we have weapons that apply conditions for players.

peonso:
A hit that deals no damage can poison. If the damage is blocked by shield there is no poison, but it can still poison if the residual damage not blocked by shield is blocked by armor and there is no damage.

PR #2207, opened 2017-03-20
Fix monsters walking over fields.

Mkalo:
Removed the "strong" condition for going through fields as this is not accurate. Replaced it with flags in monsters:
"canwalkonenergy", "canwalkonfire", "canwalkonpoison".

All monsters have these flags defaulted to 1 unless specified in the XML.

Added a new condition for the monster to go through a field:
Monsters that can't go through a certain field type will ignore it if it has been attacked while he is random stepping (no available reachable target).

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3999,8 +3999,6 @@ bool Game::combatChangeHealth(Creature* attacker, Creature* target, CombatDamage
 			return true;
 		}
 
-		target->drainHealth(attacker, realDamage);
-
 		if (spectators.empty()) {
 			map.getSpectators(spectators, targetPos, true, true);
 		}
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3999,6 +3999,8 @@ bool Game::combatChangeHealth(Creature* attacker, Creature* target, CombatDamage
 			return true;
 		}
 
+		target->drainHealth(attacker, realDamage);
+
 		if (spectators.empty()) {
 			map.getSpectators(spectators, targetPos, true, true);
 		}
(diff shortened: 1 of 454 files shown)


joseluis2g:
Hello @Mkalo, thanks for this.
Btw its not working as it should, you can see my data to be sure.
Monster will only walk on fields if they have been attacked, you should remove the hasCondition check to make it work as intended.

src/game.cpp
```diff
@@ -4001,6 +4001,15 @@ bool Game::combatChangeHealth(Creature* attacker, Creature* target, CombatDamage
 		}
 
 		target->drainHealth(attacker, realDamage);
+		if (realDamage > 0) {
+			if (Monster* targetMonster = target->getMonster()) {
```
EvulMastah: Wouldn't this logic fit more in Monster::drainHealth?

Mkalo: Yes, you're right. I'll move it.

GoularPink:
Hello,
"**All monsters have these flags defaulted to 1 unless specified in the XML.**"

I dont specified on XML the War Golem and I just tested, he dont walk on fire field.. why?

MatheusGrilo:
@GoularPink https://github.com/Mkalo/forgottenserver/blob/d34fe86403b2d9b8184a2e202d722e9310446a6e/src/monsters.h#L147-L149

joseluis2g:
> "canwalkonenergy", "canwalkonfire", "canwalkonpoison".
> All monsters have these flags defaulted to 1 unless specified in the XML.
just read my friend

thexamx:
Some monsters on cipsoft servers have weird behaviour. Like "vexclaw" he has like 50%(maybe not) chance of entering field, sometimes he will dont enter the field and sometimes he is just rushing at you. Will you think about adding this feature too? Great work btw ;D

thexamx:
Found a bug.
Summons(from monsters) are not random stepping while they cant pass the field, they are just standing in front of the wall looking at the enemy.

PR #2220, opened 2017-03-27
Try and use link-time optimizations

ranisalt:
In a similar fashion to how `-std=c++11` is enabled, now try and load `-flto` which does [link-time optimization](https://gcc.gnu.org/onlinedocs/gccint/LTO.html). Should generate smaller and potentially faster code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ if (CMAKE_COMPILER_IS_GNUCXX)
 endif()
 
 include(FindCXX11)
+include(FindLTO)
 
 # Find packages.
 find_package(GMP REQUIRED)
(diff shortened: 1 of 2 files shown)


djarek:
I'd like this to be more thoroughly tested(too bad we don't have tests!) before it is merged - LTO can cause lurking UB to surface.

ranisalt:
Yeah, that's why I PRd and not just pushed. Though I didn't find anything wrong yet, I'd like to test it more.

DSpeichert:
@djarek Unless we have a server here somewhere with players on it that we can use to test, I'd say that we can merge that and count on bug reports coming in in case anything goes wrong.

djarek:
What about LTO in MSVC?

DSpeichert:
Closest similar option seems to be [/GL (Whole Program Optimization)](https://msdn.microsoft.com/en-us/library/0zza0de8.aspx) which is already turned on for the "Release" build.

PR #2232, opened 2017-04-16
Fix party spells

raymondtfr:
* Fixed a **bug** where the former code was getting the *affected party members* **by distance from the caster member to other party members**. The affected party members are now gotten **through the positions of the area of effect of these spells**.

In addition of the bug-fix I also took the time to remove subIds of these spells (cause using them makes the spell behavior to differ from that of vanilla) . Removing these subIds keeps the same default behavior before I introduced these changes in #2004.

### DIFF
--- a/data/spells/lib/spells.lua
+++ b/data/spells/lib/spells.lua
@@ -301,7 +301,11 @@ function Creature:addDamageCondition(target, conditionType, listType, damage, ti
 	return true
 end
 
-function Player:addPartyCondition(combat, variant, condition, baseMana)
+function Player:addPartyCondition(combat, variant, positions, condition, baseMana)
+	if not combat:execute(self, variant) then
+		return false
+	end
+
 	local party = self:getParty()
 	if not party then
 		self:sendCancelMessage(RETURNVALUE_NOPARTYMEMBERSINRANGE)
@@ -312,38 +316,38 @@ function Player:addPartyCondition(combat, variant, condition, baseMana)
 	local members = party:getMembers()
 	members[#members + 1] = party:getLeader()
 
-	local position = self:getPosition()
 	local affectedMembers = {}
 	for _, member in ipairs(members) do
-		if member:getPosition():getDistance(position) <= 36 then
-			affectedMembers[#affectedMembers + 1] = member
+		local memberPosition = member:getPosition()
+		for _, position in ipairs(positions) do
+			if memberPosition == position then
+				affectedMembers[#affectedMembers + 1] = member
+			end
 		end
 	end
 
-	if #affectedMembers <= 1 then
+	if #affectedMembers == 1 then
 		self:sendCancelMessage(RETURNVALUE_NOPARTYMEMBERSINRANGE)
-		position:sendMagicEffect(CONST_ME_POFF)
+		self:getPosition():sendMagicEffect(CONST_ME_POFF)
 		return false
 	end
 
-	local mana = math.ceil(math.pow(0.9, #affectedMembers - 1) * baseMana * #affectedMembers)
+	local mana = math.ceil(#affectedMembers * math.pow(0.9, #affectedMembers - 1) * baseMana)
 	if self:getMana() < mana then
 		self:sendCancelMessage(RETURNVALUE_NOTENOUGHMANA)
-		position:sendMagicEffect(CONST_ME_POFF)
+		self:getPosition():sendMagicEffect(CONST_ME_POFF)
 		return false
 	end
-
-	if not combat:execute(self, variant) then
-		self:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
-		position:sendMagicEffect(CONST_ME_POFF)
-		return false
-	end
-
-	self:addMana(baseMana - mana, false)
-	self:addManaSpent(mana - baseMana)
 
 	for _, member in ipairs(affectedMembers) do
 		member:addCondition(condition)
 	end
+
+	for _, position in ipairs(positions) do
+		position:sendMagicEffect(CONST_ME_MAGIC_BLUE)
+	end
+
+	self:addMana(-mana)
+	self:addManaSpent(mana)
 	return true
 end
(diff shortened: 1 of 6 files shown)


raymondtfr:
A proper fix should be a complete refactor of spells, allowing us to add any or configure any spell we wanted to, based on vanilla changes, in a way that we could always use the same interface for all spells, i.e.  ``onCastSpell``, or maybe separate callback functions for each spell operation (area, effect, damage, etc).

The new script type of spells in OpenTibiaServer master (I think it is called revscriptsys?) is a good idea (although not entirely made in a greater way), but the ``spells`` code/callbacks implementation shouldn't be in Lua, but in the Lua C++ API (is that how I say it? is it possible?).

Unfortunately, I haven't been so active studying programming anymore, and even if I did I wouldn't have the experience needed to do this, all I got is the thoughts/procedures of it, how I would do.

Basically what I fixed is this:
Whoever first created the script code for those Party spells (they are seen in OpenTibiaServer files too), thought he got the idea right, because you know, ``tibia.wikia`` says (i.e. for the spell Train Party):

**Raises the Fist Fighting, sword, axe, club and distance fighting skills of party members in a range of 36 square meters around the caster by 3 for 2 minutes.**

So he/she thought understood it like this (taken from 1.0):
```lua
	local affectedList = {}
	for _, pid in ipairs(membersList) do
		if(getDistanceBetween(getCreaturePosition(pid), pos) <= 36) then
			table.insert(affectedList, pid)
		end
end
```

But it works like any other spell, that is, the party members affected by this spell, hence knowing how many members were affected, should be those that are standing on the tiles on which the spell area was sent on. The area for this spell is 3x3, so it has a total of 36 squares (not including the caster tile, which should total it to 37), so whoever did this did it solely based on ``tibia.wikia`` informations, which aren't always reliable, hence whoever did this did it incorrectly by thinking we should get the affected members through the distance from the caster member to the other members (i.e. the use of ``(getDistanceBetween(getCreaturePosition(pid), pos) <= 36``).

diath:
I'm not saying the system is perfect but you could achieve the similar results (based on the letter part of your comment) by using two combats with tile callbacks to count the number of players effected by an area using an invisible effect then actually execute the spell cast part.

raymondtfr:
@diath Thank you for your feedback, you actually gave me a new idea, I'll soon post here a different script using a new approach, based on your feedback.

By the way, while I was deciding how to make these scripts it seemed to me that having a table of *positions* in the case of this **tile callback** or a table of *targets* in **target callback** gives much more control than executing such callback once for each target/position. With such tables, I would not need to use *addEvent* although we would still have problems with **saying spell words, even if the cast requirements were not meet** this being the reason for me to modifying codes in C++ side.

raymondtfr:
I think this is now ready, there was no need for source modifications.

raymondtfr:
@djarek @ranisalt

data/spells/lib/spells.lua
```diff
@@ -312,38 +316,39 @@ function Player:addPartyCondition(combat, variant, condition, baseMana)
 	local members = party:getMembers()
 	members[#members + 1] = party:getLeader()
 
-	local position = self:getPosition()
 	local affectedMembers = {}
 	for _, member in ipairs(members) do
-		if member:getPosition():getDistance(position) <= 36 then
-			affectedMembers[#affectedMembers + 1] = member
+		local memberPosition = member:getPosition()
+
```
raymondtfr: Should I remove this line? or does it improve readability a bit?

ranisalt: Barely. At least be consistent: if you add a line here, add a line before the for above too.

djarek:
@raymondtfr 
The github rendering engine has changed a little while ago which messed up the entire wiki:/.

raymondtfr:
Would it be wise to move duplicated code from scripts to the library function @djarek @ranisalt @gugahoa ? i.e.

<details>
 <summary>Code</summary>

```lua
function Player:addPartyCondition(combat, variant, condition, baseMana)
	local party = self:getParty()
	if not party then
		self:sendCancelMessage(RETURNVALUE_NOPARTYMEMBERSINRANGE)
		self:getPosition():sendMagicEffect(CONST_ME_POFF)
		return false
	end

	local positions = {}
	function onTargetTile(creature, position)
		positions[#positions + 1] = position
	end

	combat:setCallback(CALLBACK_PARAM_TARGETTILE, "onTargetTile")

	if not combat:execute(self, variant) then
		return false
	end

	local members = party:getMembers()
	members[#members + 1] = party:getLeader()

	local affectedMembers = {}
	for _, member in ipairs(members) do
		local memberPosition = member:getPosition()
		for _, position in ipairs(positions) do
			if memberPosition == position then
				affectedMembers[#affectedMembers + 1] = member
			end
		end
	end

	if #affectedMembers == 1 then
		self:sendCancelMessage(RETURNVALUE_NOPARTYMEMBERSINRANGE)
		self:getPosition():sendMagicEffect(CONST_ME_POFF)
		return false
	end

	local mana = math.ceil(#affectedMembers * math.pow(0.9, #affectedMembers - 1) * baseMana)
	if self:getMana() < mana then
		self:sendCancelMessage(RETURNVALUE_NOTENOUGHMANA)
		self:getPosition():sendMagicEffect(CONST_ME_POFF)
		return false
	end

	for _, member in ipairs(affectedMembers) do
		member:addCondition(condition)
	end

	for _, position in ipairs(positions) do
		position:sendMagicEffect(CONST_ME_MAGIC_BLUE)
	end

	self:addMana(-mana)
	self:addManaSpent(mana)
	return true
end
```
</details>

ranisalt:
Provided that it is duplicated because it is a template code, yes. If it is duplicated by chance, then I don't think so. Where, exactly?

raymondtfr:
I forgot to pinpoint the duplicated code. I have updated the code from the previous comment. Here it is (used in all four scripts):
```lua
local positions

function onTargetTile(creature, position)
	positions[#positions + 1] = position
end

combat:setCallback(CALLBACK_PARAM_TARGETTILE, "onTargetTile")

	positions = {}
	return creature:addPartyCondition(combat, variant, positions, condition, 120)
end
```

raymondtfr:

PR #2236, opened 2017-04-23
remove unnecessary code

raymondtfr:


marksamman:
Why is it unnecessary?

ranisalt:
I don't think it is unnecessary. It's fallback for when the spell isn't properly configured and you can actually omit the spell group if it is attack or healing spell.

marksamman:
Yeah, I think this is a sane default if a group hasn't been specified. We could avoid doing this if "none" has explicitly been specified as the group though (if that makes sense and that's why @raymondtfr opened this PR).

raymondtfr:
I think it is unnecessary because it will force the spell to always have a default cooldown group (icon) assigned. But what if we want to have no effect displayed for a determined cooldown icon (this can be reverted in spells.xml)?

raymondtfr:
Closing this as this may seem *unnecessary*.

PR #2242, opened 2017-05-08
Fixed variable out of scope in first wand/rod script

jerryskye:
Line 122:  `player:addItem(items[vocationId], 1)` wasn't resulting in adding the item, because variable `vocationId` was out of scope, confined only to the first branch of the if statement.


ranisalt:
Thanks!

PR #2244, opened 2017-05-09
Non-pushable walking NPC

pzaj2:
This will allow marking NPCs as non-pushable even if they can walk on their own. "pushable" XML attribute is optional.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -738,7 +738,7 @@ void Game::playerMoveCreature(Player* player, Creature* movingCreature, const Po
 			}
 
 			Npc* movingNpc = movingCreature->getNpc();
-            if (movingNpc && (!movingNpc->isPushable() || !Spawns::isInZone(movingNpc->getMasterPos(), movingNpc->getMasterRadius(), toPos))) {
+			if (movingNpc && (!movingNpc->isPushable() || !Spawns::isInZone(movingNpc->getMasterPos(), movingNpc->getMasterRadius(), toPos))) {
 				player->sendCancelMessage(RETURNVALUE_NOTENOUGHROOM);
 				return;
 			}
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -738,7 +738,7 @@ void Game::playerMoveCreature(Player* player, Creature* movingCreature, const Po
 			}
 
 			Npc* movingNpc = movingCreature->getNpc();
-			if (movingNpc && (!movingNpc->isPushable() || !Spawns::isInZone(movingNpc->getMasterPos(), movingNpc->getMasterRadius(), toPos))) {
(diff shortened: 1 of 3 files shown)


jo3bingham:
Spacing.

pzaj2:
Alright, you're right. My editor was showing them just fine. Fixed (I believe).

src/npc.cpp
```diff
@@ -153,6 +154,12 @@ bool Npc::loadFromXml()
 		baseSpeed = 100;
 	}
 
+        if((attr = npcNode.attribute("pushable"))) {
+            pushable = attr.as_bool();
+        } else {
```
EvulMastah: Not needed. It is set to true by default in `reset()`.

Also, space after `if` and correct spacing here (use tabs instead of spaces).

pzaj2: here you go.

src/game.cpp
```diff
@@ -738,7 +738,7 @@ void Game::playerMoveCreature(Player* player, Creature* movingCreature, const Po
 			}
 
 			Npc* movingNpc = movingCreature->getNpc();
-			if (movingNpc && !Spawns::isInZone(movingNpc->getMasterPos(), movingNpc->getMasterRadius(), toPos)) {
+			if (movingNpc && (!movingNpc->isPushable() || !Spawns::isInZone(movingNpc->getMasterPos(), movingNpc->getMasterRadius(), toPos))) {
```
ranisalt: You can use De Morgan and change to `!(movingNpc->isPushable() && Spawns::isInZone(...))` instead.

src/npc.h
```diff
@@ -118,7 +118,7 @@ class Npc final : public Creature
 		}
 
 		bool isPushable() const final {
-			return walkTicks > 0;
+			return walkTicks > 0 && pushable;
```
ranisalt: Test for pushability first, `pushable && walkTicks > 0`. Actually, since `walkTicks` is unsigned, you could test `!= 0` instead.

pzaj2: You're right about the test. I will change the order and use non-equality test then.

src/game.cpp
pzaj2: :+1: will do

pzaj2: I think you may be right. I will analyze it again and check if this could be removed.

pzaj2: You're right. The redundant check is removed.

pzaj2: Seems like this check was redundant at first, removed it.

src/npc.h
pzaj2: done

ranisalt:
Thanks!

PR #2250, opened 2017-05-11
make ProtocolGame::login use loadPlayerById instead of loadPlayerByName

pzaj2:
Since the ID was already loaded in preloadPlayer earlier in this function, why not use ID instead of name? Not only it seems more correct to load it using PK - even if `name` column is unique and indexed - but is better for performance.

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -126,7 +126,7 @@ void ProtocolGame::login(const std::string& name, uint32_t accountId, OperatingS
 			return;
 		}
 
-		if (!IOLoginData::loadPlayerByName(player, name)) {
+		if (!IOLoginData::loadPlayerById(player, player->getGUID())) {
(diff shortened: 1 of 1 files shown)


pzaj2:
Just wanted to point out that this happens in various other places as well. Should I look for them and apply similar modifications?

djarek:
Just wondering - is the ```loadPlayerByName()``` function actually needed? :D

pzaj2:
@djarek 
IIRC it is used by Mailbox, isn't it? To load the receiving user from database :)

HeavenIsLost:
I also vote for allowing the loadPlayerByName function to exists even if its not used anywhere. It will may be useful in the future and it may be used by custom tfs forks.

marksamman:
Thanks!

ranisalt:
We would move out of `loadSomethingByAttribute` functions once using some smart database mapping.

PR #2253, opened 2017-05-13
allow varying (random) haste / paralyze for monsters

pzaj2:
Since there are some monsters in Tibia that has varying haste (paralyze?) strength, I think it would be nice to have that available in monsters via XML. This will have no impact on current monsters with constant change speed. The behaviour of this spell will only change for changespeed="0" - it will now become haste, not paralyze (which I think would be more correct anyway).

Feel free to bring your suggestions. I'm aware it may require some refactoring.

I don't yet have the monsters, I need to look for information on what monsters actually have this ability.


src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+
+				if(speedChangeMin < -1000) {
+					speedChangeMin = -1000;
+				}
+
+				speedChangeMax = speedChangeMin;
+			} else {
+				attr = node.attribute("speedchangemin");
+
+				if(attr) {
```
djarek: space after if

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+					speedChangeMin = pugi::cast<int32_t>(attr.value());
+
+					if(speedChangeMin < -1000) {
```
djarek: space after if

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+						speedChangeMin = -1000;
+					}
+
+					speedChangeMax = speedChangeMin;
+				}
+
+				attr = node.attribute("speedchangemax");
+
+				if(attr) {
```
djarek: space after if

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+					speedChangeMax = pugi::cast<int32_t>(attr.value());
+
+					if(speedChangeMax < -1000) {
```
djarek: space after if

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+						speedChangeMax = -1000;
+					}
+				}
+
+				if(!hasSameSign(speedChangeMin, speedChangeMax)) {
```
djarek: space after if

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+					std::cout << "[Error - Monsters::deserializeSpell] - " << description << " - invalid speed change range: " << name << std::endl;
+					delete combat;
+					return false;
+				}
+
+				if(speedChangeMin > speedChangeMax) {
```
djarek: space after if

pzaj2: I have this feeling that you'd like to kill me now reading this and having in mind that it's not the first time. I usually use "format indentations in file" and this removes these spaces :( and my IDE has no setting that forces space after if. Anyway, fixed.

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
 			combat->setParam(COMBAT_PARAM_TYPE, COMBAT_HEALING);
 			combat->setParam(COMBAT_PARAM_AGGRESSIVE, 0);
 		} else if (tmpName == "speed") {
-			int32_t speedChange = 0;
+			int32_t speedChangeMin = 0;
```
ninjalulz: IMO, minSpeedChange & maxSpeedChange has a better ring to it. 😄

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
 			if ((attr = node.attribute("duration"))) {
 				duration = pugi::cast<int32_t>(attr.value());
 			}
 
 			if ((attr = node.attribute("speedchange"))) {
-				speedChange = pugi::cast<int32_t>(attr.value());
-				if (speedChange < -1000) {
-					//cant be slower than 100%
-					speedChange = -1000;
+				speedChangeMin = pugi::cast<int32_t>(attr.value());
```
ninjalulz: std::max can be used here
```cpp
std::max<int32_t>(-1000, pugi::cast<int32_t>(attr.value()));
```

src/tools.cpp
```diff
@@ -346,6 +346,15 @@ void trimString(std::string& str)
 	str.erase(0, str.find_first_not_of(' '));
 }
 
+bool hasSameSign(int32_t x, int32_t y, bool neutralZero /* = true */)
+{
+	if(neutralZero) {
```
ninjalulz: Don't you think this if-statement is quite redundant at the moment?

src/monsters.cpp
```diff
@@ -431,31 +431,66 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+
+				if (speedChangeMin < -1000) {
+					speedChangeMin = -1000;
+				}
+
+				speedChangeMax = speedChangeMin;
+			} else {
+				attr = node.attribute("speedchangemin");
```
ninjalulz: Do the assignment within the if statement for consistency

src/tools.cpp
pzaj2: Well, perhaps it's redundant at the moment, but I'm afraid someone may use it in the future with neutralZero set to false and it will give unexpected behaviour.

Thank you for your suggestions, I've made changes based on them, apart from this one. I think it would be better to leave it to avoid confusion when used in the future.

pzaj2:
I was also thinking about merging `speedChange` & `minSpeedChange` into one `if` statement, but then it would require an additional bool variable to control whether it's `speedChange` or `minSpeedChange` that has been handled (since the code for both is basically the same) and it seems not very elegant solution. What do you think? Something like:
```Cpp
bool isSpeedRange = false;

if ((attr = node.attribute("speedchange")) || (attr = node.attribute("minSpeedChange") && isSpeedRange = true)) {
    minSpeedChange = std::max(-1000, pugi::cast<int32_t>(attr.value()));
    maxSpeedChange = minSpeedChange;
}
```

Then instead of `else if` it would be:

```Cpp
if (isRangeSpeed && (attr = node.attribute("minSpeedChange")))
```

src/tools.cpp
EvulMastah: You are still missing a space after if tho. :)

pzaj2: @EvulMastah 
oh come on! Nobody spotted it there! Just corrected :)

EvulMastah:
Try tricking CI to run again. For example: squash your commits and push with force.

ranisalt:
I am against it. To me, it seems you are programming through XML, and we have Lua precisely to avoid that. It's a regression.

src/monsters.cpp
```diff
@@ -431,31 +431,48 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+				}
+
+				if (!hasSameSign(minSpeedChange, maxSpeedChange)) {
+					std::cout << "[Error - Monsters::deserializeSpell] - " << description << " - invalid speed change range: " << name << std::endl;
+					delete combat;
+					return false;
+				}
+
+				if (minSpeedChange > maxSpeedChange) {
+					std::swap(minSpeedChange, maxSpeedChange);
```
ranisalt: `std::tie(minSpeedChange, maxSpeedChange) = std::minmax(minSpeedChange, maxSpeedChange);`, though you should only issue a warning if there are no side effects.

src/monsters.cpp
```diff
@@ -431,31 +431,48 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
 			if ((attr = node.attribute("duration"))) {
 				duration = pugi::cast<int32_t>(attr.value());
 			}
 
 			if ((attr = node.attribute("speedchange"))) {
-				speedChange = pugi::cast<int32_t>(attr.value());
-				if (speedChange < -1000) {
-					//cant be slower than 100%
-					speedChange = -1000;
+				minSpeedChange = std::max(-1000, pugi::cast<int32_t>(attr.value()));
```
ranisalt: minSpeedChange = maxSpeedChange = std::max(-1000, pugi::cast<int32_t>(attr.value()));

src/monsters.cpp
```diff
@@ -431,31 +431,48 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+				maxSpeedChange = minSpeedChange;
+			} else if ((attr = node.attribute("minspeedchange"))) {
+				minSpeedChange = std::max(-1000, pugi::cast<int32_t>(attr.value()));
+				maxSpeedChange = minSpeedChange;
+
+				attr = node.attribute("maxspeedchange");
+
+				if (attr) {
```
ranisalt: `if ((attr = node.attribute("maxspeedchange"))) {`

src/monsters.cpp
```diff
@@ -431,31 +431,48 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
 			if ((attr = node.attribute("speedchange"))) {
-				speedChange = pugi::cast<int32_t>(attr.value());
-				if (speedChange < -1000) {
-					//cant be slower than 100%
-					speedChange = -1000;
+				minSpeedChange = std::max(-1000, pugi::cast<int32_t>(attr.value()));
+				maxSpeedChange = minSpeedChange;
+			} else if ((attr = node.attribute("minspeedchange"))) {
+				minSpeedChange = std::max(-1000, pugi::cast<int32_t>(attr.value()));
+				maxSpeedChange = minSpeedChange;
```
ranisalt: Same as above.

src/monsters.cpp
```diff
@@ -431,31 +431,48 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+
+				attr = node.attribute("maxspeedchange");
+
+				if (attr) {
+					maxSpeedChange = std::max(-1000, pugi::cast<int32_t>(attr.value()));
+				}
+
+				if (!hasSameSign(minSpeedChange, maxSpeedChange)) {
```
ranisalt: No need for this new function, `(minSpeedChange ^ maxSpeedChange) < 0` means signals differ.

ranisalt:
`hasSameSign` is definitely not needed, just bitwise-xor numbers and check most significant bit. I still doubt it should be in C++.

kito2:
I am not approving this too, cause on RL creatures do paralyze.

The important fact is the paralyze now is applied on speed and not on base speed, that would be a good improvement.

Leo32onGIT:
IMO this should be closed
Author hasn't addressed @ranisalt requests

> `hasSameSign` is definitely not needed, just bitwise-xor numbers and check most significant bit. I still doubt it should be in C++.

If you want to use dynamic paralyze amounts use a monster spell script and make it use that spell instead.

This just seems to be a custom way of doing it via xml.
If someone has a niche need for that rather than through lua, then they can refer to this PR.

TheSumm:
In reality every single monster uses this feature. I don't think being forced to create an additional spell for each monster and cluttering your datapack that way is justified for something as simple as a variable paralyze condition.

Leo32onGIT:
☝️

> In reality every single monster uses this feature. I don't think being forced to create an additional spell for each monster and cluttering your datapack that way is justified for something as simple as a variable paralyze condition.

I'll concede that extra functionality is extra functionality.
Rebased and submitted new PR.

PR #2254, opened 2017-05-14
Added missing pops

DanielChabrowski:
Small mistake I found which actually caused lua_close to complain since I have more than 50 missing keys in config.lua.


djarek:
Thanks!

PR #2267, opened 2017-06-12
Remove monster immune effect

OPfeajoief13:
Monsters shouldn't receive magic effects if they're immune to the condition.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3611,7 +3611,7 @@ bool Game::combatBlockHit(CombatDamage& damage, Creature* attacker, Creature* ta
 			addMagicEffect(targetPos, CONST_ME_POFF);
 		} else if (blockType == BLOCK_ARMOR) {
 			addMagicEffect(targetPos, CONST_ME_BLOCKHIT);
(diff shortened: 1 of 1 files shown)


wgrriffel:
why?

OPfeajoief13:
Dragon Lords for example shouldn't receive blockhit effect if they walk on a magic field they're resistant to (fire).

wgrriffel:
when i used to play monsters actually did make a block effect from elements that they are immune, did this changed?

OPfeajoief13:
To be honest I'm not 100% sure, what I do know is that if a Dragon Lord with race fire walks in a firefield (type fire) it does not show any effect whatsoever.

Maybe it should be like:

if monster->getRace() == item->getMagicfield(){
No effect
}

OPfeajoief13:
I will test this soon, maybe someone can confirm this for me

joseluis2g:
I tested with some mobs on tibia and if monster is immune to fire the spark still showing
https://www.youtube.com/watch?v=v1grapvsY90

OPfeajoief13:
It's indeed false.

Issue #2255, opened 2017-05-15
[BUG] /reload monsters will double monsters strength


Strutzz:
Found issue when reloading monsters. It seems to be doubling the strength of all monsters.
Easiest to see by summoning a training monk and a demon. have the demon attack the training monk and then type in '/reload monsters'  demons attack will be stronger the more you reload.

Kamenuvol:
At least is not memory leaking nor invalidating pointers...

Strutzz:
Just an update it looks like its not 'doubling' the monsters attack, but it is making it attack twice causing a combo.

M4GO:
i have this bug its doubling the monsters loots too

> Strutzz commented 2 hours ago
Just an update it looks like its not 'doubling' the monsters attack, but it is making it attack twice causing a combo

Strutzz:
The fix that @M4GO posted indeed fixes the issue... It also fixes another bug of reloading spells causing the server to crash.. ZZZZZZZZZZ

M4GO:
@Strutzz  change the credits bug fixed by @DanielChabrowski

marksamman:
Fixed in 967d82305e7bdbc54a5dbe92c45bd9caee736117

PR #2268, opened 2017-06-13
Fix monster reload

DanielChabrowski:
Forgot to make a pull request. It fixes the issue #2255 of doubling monster's attacks and loot during reload.


marksamman:
Thanks!

PR #2276, opened 2017-07-01
Update modules.lua

luanluciano93:
I have doubts in two functions:
• doAddContainerItem(container, itemid, subType) - Container(container):addItemEx(Item(itemid)) ?
• player:say(text, type, false, pcid, player:getPosition()) - pcid or player2 ?

### DIFF
--- a/data/npc/lib/npcsystem/modules.lua
+++ b/data/npc/lib/npcsystem/modules.lua
@@ -926,20 +926,22 @@ if Modules == nil then
 		local backpack = 1988
 		local totalCost = amount * shopItem.buy
 		if inBackpacks then
-			totalCost = isItemStackable(itemid) == TRUE and totalCost + 20 or totalCost + (math.max(1, math.floor(amount / getContainerCapById(backpack))) * 20)
+			totalCost = ItemType(itemid):isStackable() == TRUE and totalCost + 20 or totalCost + (math.max(1, math.floor(amount / ItemType(backpack):getCapacity())) * 20)
 		end
 
+		local player = Player(cid)
 		local parseInfo = {
-			[TAG_PLAYERNAME] = getPlayerName(cid),
+			[TAG_PLAYERNAME] = player:getName(),
 			[TAG_ITEMCOUNT] = amount,
 			[TAG_TOTALCOST] = totalCost,
 			[TAG_ITEMNAME] = shopItem.name
 		}
 
-		if getPlayerMoney(cid) < totalCost then
+		if player:getMoney() < totalCost then
 			local msg = self.npcHandler:getMessage(MESSAGE_NEEDMONEY)
 			msg = self.npcHandler:parseMessage(msg, parseInfo)
-			doPlayerSendCancel(cid, msg)
+			player:sendCancelMessage(msg)
+			player:sendCancelMessage(msg)
 			return false
 		end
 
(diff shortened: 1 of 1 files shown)


data/npc/lib/npc.lua
```diff
@@ -11,33 +11,34 @@ function msgcontains(message, keyword)
...
 	local amount = amount or 1
 	local subType = subType or 0
 	local item = 0
-	if isItemStackable(itemid) then
+	if ItemType(itemid):isStackable() then
 		if inBackpacks then
-			stuff = doCreateItemEx(backpack, 1)
-			item = doAddContainerItem(stuff, itemid, math.min(100, amount))
+			stuff = Game.createItem(backpack, 1)
+			item = doAddContainerItem(stuff, itemid, math.min(100, amount)) -- ???
```
WibbenZ: Why did you add "-- ???"
And why did you not change the compat function? https://github.com/otland/forgottenserver/blob/master/data/lib/compat/compat.lua#L583-L594

data/npc/lib/npc.lua
```diff
@@ -11,33 +11,34 @@ function msgcontains(message, keyword)
...
 
 	local a = 0
 	if inBackpacks then
-		local container, b = doCreateItemEx(backpack, 1), 1
+		local container, b = Game.createItem(backpack, 1), 1
 		for i = 1, amount do
-			local item = doAddContainerItem(container, itemid, subType)
-			if table.contains({(getContainerCapById(backpack) * b), amount}, i) then
-				if doPlayerAddItemEx(cid, container, ignoreCap) ~= RETURNVALUE_NOERROR then
+			local item = doAddContainerItem(container, itemid, subType) -- ???
```
WibbenZ: Same thing here as above;
"
Why did you add "-- ???"
And why did you not change the compat function? https://github.com/otland/forgottenserver/blob/master/data/lib/compat/compat.lua#L583-L594
"

data/npc/lib/npc.lua
```diff
@@ -56,33 +57,37 @@ function doNpcSellItem(cid, itemid, amount, subType, ignoreCap, inBackpacks, bac
 end
 
 local func = function(cid, text, type, e, pcid)
-	if isPlayer(pcid) then
-		doCreatureSay(cid, text, type, false, pcid, getCreaturePosition(cid))
-		e.done = TRUE
+	local player = Creature(cid)
```
WibbenZ: Move this to below ```if player2 then```

data/npc/lib/npc.lua
```diff
@@ -56,33 +57,37 @@ function doNpcSellItem(cid, itemid, amount, subType, ignoreCap, inBackpacks, bac
...
+		player:say(text, type, false, pcid, player:getPosition())
+		e.done = true
 	end
 end
 
 function doCreatureSayWithDelay(cid, text, type, delay, e, pcid)
-	if isPlayer(pcid) then
-		e.done = FALSE
+	local player = Player(pcid)
+	if player then
```
WibbenZ: Since player is only used once you can use ```if Player(pcid) then```

data/npc/lib/npcsystem/modules.lua
```diff
@@ -986,35 +988,36 @@ if Modules == nil then
...
-			doPlayerSendCancel(cid, msg)
+			player:sendCancelMessage(msg)
 			self.npcHandler.talkStart[cid] = os.time()
 			return false
 		end
 	end
 
 	-- Callback for requesting a trade window with the NPC.
-	function ShopModule.requestTrade(cid, message, keywords, parameters, node)
+	function ShopModule.requestTrade(cid, message, keywords, parameters, node)
```
WibbenZ: Remove trailing tab

DSpeichert:
The source repository seems to be removed and we can't accept that as-is.
I'm closing this and we will replicate some of those changes in other PRs.

PR #2282, opened 2017-07-17
Missing distance weapons attributes

### LINKED ISSUE #1909: Missing item attribute for distance weapons?
### Actual behaviour

We have this bow.

```
<item id="2456" article="a" name="bow">
        <attribute key="weight" value="3100" />
        <attribute key="weaponType" value="distance" />
        <attribute key="slotType" value="two-handed" />
        <attribute key="ammoType" value="arrow" />
        <attribute key="skilldist" value="3" />
        <attribute key="range" value="6" />
    </item>
```

when we add "ammoType" then it doesnt show skill dist in description:
16:57 You see a bow (Range:6).
else after removing only this attribute it shows "distance fighting +3"
16:58 You see a bow (distance fighting +3).
### Expected behaviour

It should show item attribute everytime

thexamx:
Fix #1909

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -873,14 +873,12 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
 
 				begin = false;
 			} else {
-
 				int32_t attack, defense, extraDefense;
 				if (item) {
 					attack = item->getAttack();
 					defense = item->getDefense();
 					extraDefense = item->getExtraDefense();
-				}
-				else {
+				} else {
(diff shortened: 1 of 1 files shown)


GoularPink:
Hi, very good, but could not be written only in 1 parenthesis?

**no**
(Range:5, Atk+9, Hit%+4) (distance fighting +3). 

**yes**
(Range:5, Atk+9, Hit%+4, distance fighting +3).

thexamx:
@GoularPink done ;)

PR #2283, opened 2017-07-17
Fixing enabling party shared in protection zone

### LINKED ISSUE #2270: Party Shared Experience - bug with enabling in protection zone

1. Enter protection zone tile and try to enable shared experience after you attacked monsters and got any condition which is sustaining your condition in fight.

### Expected behaviour
You should be able to enable shared experience immediately.

### Actual behaviour
You cannot enable shared experience, because the server is thinking that u got the "condition in fight" even if you are in protection zone where you can logout when you want.

nekiro:
Fix #2270

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4827,7 +4827,8 @@ void Game::playerEnableSharedPartyExperience(uint32_t playerId, bool sharedExpAc
 	}
 
 	Party* party = player->getParty();
-	if (!party || player->hasCondition(CONDITION_INFIGHT)) {
+	Tile* playerTile = player->getTile();
+	if (!party || (player->hasCondition(CONDITION_INFIGHT) && playerTile && !playerTile->hasFlag(TILESTATE_PROTECTIONZONE))) {
 		return;
 	}
 
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4827,8 +4827,7 @@ void Game::playerEnableSharedPartyExperience(uint32_t playerId, bool sharedExpAc
 	}
 
 	Party* party = player->getParty();
-	Tile* playerTile = player->getTile();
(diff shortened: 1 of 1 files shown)


wgrriffel:
is this the right approach? shouldn't condition 'peace' (idk how it's called) overwrite condition in fight? imo it's a condition bug and should be fixed there

thexamx:
I checked how it is in Tibia, and it's as it should be. Protection zone does not override the conidtion in fight.
Example. You have condition in fight, because you attacked something recently. You are going to protection zone, you can log out etc, but if you leave this zone immadetiately you still have condition in fight.

ranisalt:
But did you check enabling party share while infight and peace conditions are both active? That's the point of this change.

OPfeajoief13:
Pacified?

thexamx:
@ranisalt 
If you meant having condition in fight and being in protection zone(peace condition, thats how I understand your message) then yeah I tested it.

If you got condition in fight then it is active even when you enter protection zone.

ranisalt:
No, I asked if you BOTH have infight ticks AND is in a protection zone you can enable party sharing, or if you need to wait until infight ends.

thexamx:
You do not need to wait until condition in fight ends, you should be able to enable shared even with in fight condition when you are in protection zone,

jo3bingham:
Technically, you can drop the "InFight" check and just check if the player is in a PZ zone.

thexamx:
@jo3bingham I can't, because when I drop it, then player will be able to enable shared only in protection zone. You should be able to enable shared, when you got no condition infight or when you are in protection zone. We are ignoring condition infight check if we are in protection zone.

No condition infight and not in protection zone:
https://gyazo.com/a7042da671b99c7b5d840a1d7a1fb5be

Condition infight and in protection zone:
https://gyazo.com/bc039c09bf1eb7f48f1f34157748d1c0

No condition infight and in protection zone:
https://gyazo.com/987f8385f1f84e4376f734d4dba92621

jo3bingham:
@thexamx oh yes, good call.

PR #2296, opened 2017-08-04
Use base magic level instead of total for condition percent

wgrriffel:
this may cause player to earn more magic points than indeed when affected by conditions, like wearing items or casting spells that increase magic points before get affected by a condition


ranisalt:
Sorry, accidentally clicked @marksamman name.

ranisalt:
Appveyor has gone full retard. Thanks!

PR #2297, opened 2017-08-04
Fix #2264

### LINKED ISSUE #2264: Missing call to DatabaseTasks::flush()
I believe that DatabaseTasks::flush() should be called Game::saveGameState(). Why? Some queries without a callback(like a insert) may be waiting to be executed by DatabaseTasks.

HeavenIsLost:
Fix #2264, I'm also insecure about the need to call taskLock.lock() and taskLock.unlock()
@djarek @ranisalt

### DIFF
--- a/src/databasetasks.cpp
+++ b/src/databasetasks.cpp
@@ -85,6 +85,13 @@ void DatabaseTasks::runTask(const DatabaseTask& task)
 
 void DatabaseTasks::flush()
 {
+	taskLock.lock();
+	flushInternal();
+	taskLock.unlock();
+}
+
+void DatabaseTasks::flushInternal()
+{
 	while (!tasks.empty()) {
 		runTask(tasks.front());
 		tasks.pop_front();
@@ -95,7 +102,7 @@ void DatabaseTasks::shutdown()
 {
 	taskLock.lock();
 	setState(THREAD_STATE_TERMINATED);
-	flush();
+	flushInternal();
 	taskLock.unlock();
 	taskSignal.notify_one();
 }
--- a/src/databasetasks.cpp
+++ b/src/databasetasks.cpp
@@ -85,11 +85,6 @@ void DatabaseTasks::runTask(const DatabaseTask& task)
 
 void DatabaseTasks::flush()
 {
-	flushInternal();
-}
-
-void DatabaseTasks::flushInternal()
-{
 	std::unique_lock<std::mutex> guard{ taskLock };
 	while (!tasks.empty()) {
 		auto task = std::move(tasks.front());
@@ -105,6 +100,6 @@ void DatabaseTasks::shutdown()
 	taskLock.lock();
 	setState(THREAD_STATE_TERMINATED);
 	taskLock.unlock();
-	flushInternal();
+	flush();
 	taskSignal.notify_one();
 }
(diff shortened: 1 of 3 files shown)


src/databasetasks.cpp
```diff
@@ -85,6 +85,13 @@ void DatabaseTasks::runTask(const DatabaseTask& task)
...
 {
+	taskLock.lock();
+	flushInternal();
+	taskLock.unlock();
+}
+
+void DatabaseTasks::flushInternal()
+{
 	while (!tasks.empty()) {
 		runTask(tasks.front());
```
djarek: This may deadlock if the callback attempts to query another async database operation.

HeavenIsLost: How to fix @djarek

HeavenIsLost: Btw if the lock inside the flush function can deadlock then the DatabaseTasks::shutdown() may make a deadlock as both functions(DatabaseTasks::flush() and DatabaseTasks::shutdown()) are called from the dispatcher thread. Also the query callback will be only called in the next dispatcher cycle(g_dispatcher.addTask(createTask(std::bind(task.callback, result, success)));) so how it will cause a deadlock?

djarek: Okay, you're right, I forgot the callbacks are not called directly. Still, you're performing database IO while holding a mutex, so that's a bad idea. You should use a unique_lock in this case:
```
std::unique_lock<std::mutex> guard{taskLock}
while (!tasks.empty()) {
    auto task = std::move(tasks.front())
    tasks.pop_front();
    guard.unlock();
    runTask(task);
    guard.lock();
}
```

HeavenIsLost: Why its a bad idea? I will update the pull request with your code, anything more?

djarek: It's good practice: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cp43-minimize-time-spent-in-a-critical-section

A mutex prevents any other threads from accessing the object. If, for some reason that other thread is the dispatcher thread, you'd block it from execution. While it's unlikely here, it's better to just avoid the potential of this occuring.

src/databasetasks.h
```diff
@@ -47,6 +47,7 @@ class DatabaseTasks : public ThreadHolder<DatabaseTasks>
 		void threadMain();
 	private:
 		void runTask(const DatabaseTask& task);
+		void flushInternal();
```
ranisalt: I'd prefer if this function would just be on the cpp. You don't need to expose it.

HeavenIsLost: How? A normal void function with DatabaseTask as the single parameter?

ranisalt: With the tasks queue, I believe, but yes, a free void function.

Also, you are double locking the taskLock now, no?

HeavenIsLost: @djarek did i added your fix correctly?

djarek: I think you need to get rid of the locking in shutdown because as it is right now, it will hang on shutdown.

HeavenIsLost: @djarek check my last commit, i have also test the call to g_databaseTasks.flush() and g_databaseTasks.shutdown()

PR #2302, opened 2017-08-15
Transform Database class into namespace

ranisalt:
Since the Database is a singleton, it can be a namespace instead, for almost the same syntax and no need to control instances through a `getInstance` method. It also moves some private methods to the implementation file rather than header, thus exposing less stuff.

The result is a large cleanup of over 100 hundred lines less :)

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -26,14 +26,58 @@
 
 extern ConfigManager g_config;
 
-Database::~Database()
+namespace Database {
+
+MYSQL* handle = nullptr;
+std::recursive_mutex databaseLock;
+uint64_t maxPacketSize = 1048576u;
+
+namespace {
+
+void disconnect()
 {
 	if (handle != nullptr) {
 		mysql_close(handle);
 	}
 }
 
-bool Database::connect()
+bool beginTransaction()
+{
+	if (!executeQuery("BEGIN")) {
+		return false;
+	}
+
+	databaseLock.lock();
+	return true;
+}
+
+bool rollback()
+{
+	if (mysql_rollback(handle) != 0) {
+		std::cout << "[Error - mysql_rollback] Message: " << mysql_error(handle) << std::endl;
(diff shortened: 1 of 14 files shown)


src/database.cpp
```diff
@@ -288,8 +310,28 @@ bool DBInsert::execute()
...
+		Database::rollback();
+	}
+}
+
+bool DBTransaction::begin() {
+	state = STATE_START;
+	return Database::beginTransaction();
+}
+
+bool DBTransaction::commit() {
```
wgrriffel: shouldn't all these function brackets be new lined?

src/databasetasks.cpp
```diff
@@ -71,11 +71,11 @@ void DatabaseTasks::runTask(const DatabaseTask& task)
 	bool success;
 	DBResult_ptr result;
 	if (task.store) {
-		result = db.storeQuery(task.query);
+		result = Database::storeQuery(task.query);
```
djarek: The async database tasks used a different db connection (yes, we had two instances of  a class that is supposed to be a singleton...) than the dispatcher, so you've kinda killed their usefulness with this patch :/,

src/database.cpp
ranisalt: Ooops.

src/databasetasks.cpp
ranisalt: Now I see it. It shouldn't have a singleton-like interface then, but rather be instantiated and referenced where needed. What is the advantage of having multiple connections? It could be a connection pool instead.

ranisalt: An option is having a `thread_local` mysql handle, so each thread that wants to perform queries sets up it's own connection.

DSpeichert:
Given that a singleton results in incorrect behavior as pointed out by @djarek, I'm closing this now as there's nothing here to merge. A connection pool could be a good idea.

PR #2305, opened 2017-08-17
Fix Creature functions isContainer and isTeleport missing

Riverlance:


data/lib/core/creature.lua
```diff
@@ -31,7 +31,7 @@ function Creature.getPlayer(self)
 	return self:isPlayer() and self or nil
 end
 
-function Creature.isItem(self)
```
DSpeichert: Why did you change order of these two?

Issue #2293, opened 2017-08-03
Missing Itemtype functions


Codinablack:
Ok so I can't do much in the world of c++ right now but I can do some things and one thing I would like to do is to have a couple more itemtype functions available for everyone on a stock tfs. 

Itemtype(..):getAmmoType()
Itemtype(..):isGroundTile()

There are a few others that may be nice to have but aren't really that necessary as there are other ways of finding this information such as

Itemtype(..):isBlocking()
Itemtype(..)isMagicField()
Itemtype(..)isUseable()
Itemtype(..)isHangable() -- _removed_
Itemtype(..)isRotateable() -- _removed_
Itemtype(..):isPickupable()
Itemtype(..):hasWalkStack() -- _removed_
ItemType(..):getCorpseType() -- **added**

I can make each one of these and put in a pull request to have it merged, however I figure most of these aren't already there as they were deemed not necessary. I strongly request the top two atleast and I ask here what of the others would be accepted if I put in the pull request?

WibbenZ:
IMO they would all kinda be good to have, really annoying when you don't have a function that you need :p

ranisalt:
Please do. Most of those functions are actually much similar to others, you might get away copying them and changing the function to call.

Also, I think as both are constant values you can push as attributes instead of methods.

Codinablack:
I guess I will ultimately have to make some alterations because I believe a couple aren't working correctly, hasWalkStack(), isPickupable(), the two I notice right now, isUsuable(), getAmmoType(), isGroundTile(), isMagicField(), isBlocking(), are all for sure working fine as far as I can tell. Still more testing, if anyone see's my errors please go ahead and point them out, I'll fix them when I get a chance tomorrow or later, It is late now I am going to bed

thexamx:
@Codinablack 
About pickupable did you test like this:
to items.h
```
bool isPickupable() const {
	return (pickupable);
}
```
`pushBoolean(L, itemType->isPickupable());`

Codinablack:
@thexamx Thank you for that, I will definitely use it 👍

Codinablack:
The isHangable() method has been removed from the list due to isHangable already being a bool value for itemType, and would require either renaming that bool in many places in source, or figuring a better name for the method. As is, I feel that would be better made in lua anyways. 

The isRotateable() method as been removed from the list for all the same reasons as isHangable()

[...]

PR #2306, opened 2017-08-20
Missing ItemType Methods

Codinablack:
-added some new itemType methods
-altered isRune boolean in items.h for consistency

### DIFF
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -1,4 +1,4 @@
-/**
+﻿/**
  * The Forgotten Server - a free and open-source MMORPG server emulator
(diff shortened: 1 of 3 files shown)


Codinablack:
This list 

https://github.com/otland/forgottenserver/issues/2293

src/luascript.h
```diff
@@ -1,4 +1,4 @@
-/**
+﻿/**
```
ranisalt: What happened here? I can't find any diff.

ranisalt:
Thanks!

src/luascript.h
ranisalt: It's git drunk again.

PR #2308, opened 2017-08-23
Fix #1909

### LINKED ISSUE #1909: Missing item attribute for distance weapons?

thexamx:
I would like to request review.
This pull request should fix #1909


ranisalt:
Please rebase your code against master as it seems it's incorporating changes from other commits.

thexamx:
@ranisalt is it good now? I'm very noob in github commands ;d

ranisalt:
Yes! LGTM.

wgrriffel:
@thexamx github doesn't have 'commands', what has 'commands' are git, github is where you host git repositories

PR #2309, opened 2017-08-24
Support adding guild ranks to access lists, fix #2240

### LINKED ISSUE #2240: [bug] TFS 1.3 House invite list doesn't parse guild ranks
1. Add RankName@GuildName to any house list (guest, subowner, door)

### Expected behaviour
Only players that have a rank name == RankName should be able to enter.

Reference for additional functionality:
 http://tibia.wikia.com/wiki/House#The_House_Spells

### Actual behaviour
Anybody in the guild can enter, regardless of rank.

### Environment
house.cpp -> line 420

Adds the entire guild instead of only guild members with mentioned rank.

```
void AccessList::parseList(const std::string& list)
{
	playerList.clear();
	guildList.clear();
	expressionList.clear();
	regExList.clear();
	this->list = list;
	if (list.empty()) {
		return;
	}

	std::istringstream listStream(list);
	std::string line;

	while (getline(listStream, line)) {
		trimString(line);
		trim_left(line, '\t');
		trim_right(line, '\t');
		trimString(line);

		if (line.empty() || line.front() == '#' || line.length() > 100) {
			continue;
		}

		toLowerCaseString(line);

		std::string::size_type at_pos = line.find("@");
		if (at_pos != std::string::npos) {
			addGuild(line.substr(at_pos + 1));
		} else if (line.find("!") != std::string::npos || line.find("*") != std::string::npos || line.find("?") != std::string::npos) {
			addExpression(line);
		} else {
			addPlayer(line);
		}
	}
}
```
### Example

Player1 is a Rank1 of Guild1.
Player2 is a Rank2 of Guild1.

Add Rank1@Guild1 to invite list.
Rank1 and Rank2 can enter.

Only Player1 should be allowed to enter. Player2 should receive "Not invited."

ranisalt:


ranisalt:
@jerryb1988 @Codinablack

src/ioguild.h
```diff
@@ -20,7 +20,7 @@
 #ifndef FS_IOGUILD_H_EF9ACEBA0B844C388B70FF52E69F1AFF
 #define FS_IOGUILD_H_EF9ACEBA0B844C388B70FF52E69F1AFF
 
-struct Guild;
+class Guild;
```
ranisalt: Both making no semantic difference here, yet clang is annoyed. Holy fsck.

PR #2312, opened 2017-08-25
Add clang-format

DanielChabrowski:
As we were talking in #2307, it would be good idea to use clang-format to format our code. It is somewhat based on my personal preferences so I would like you to suggest modifications if any.  I wanted to make this PR clean so the actual changes are not included in here. 

For know it won't pass the CI so we will have to push changes first. Because of sorting includes it also won't compile but I will make a PR of a fix in a second. If we decide to merge this, it has to go as the last one.

PR #2313, opened 2017-08-25
Moved slots_t to enums.h

DanielChabrowski:
A change I was talking about in #2312. Sorted includes would result in a compilation error from items or movements - I don't remember. I don't see any reason why it shouldn't be in enums.h.


DanielChabrowski:
@EvulMastah It would work but honestly I don't even know the difference between types of content that are in these two files. Both have enums, only enums has also structs.

ranisalt:
Enums and consts should be defined on the files where they make sense. I think those separate files (and definitions) are legacy from otserv.

DanielChabrowski:
So you guys wanna move it to const.h or leave it in enums.h?

How about dividing those definitions into separate server-side-only and client-side-only files?

DSpeichert:
I agree with @ranisalt, this doesn't seem like a move in the right direction.

Issue #2279, opened 2017-07-07
/i (create item bug)


rudger:
### Before creating an issue, please ensure:
- [ ] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [ ] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Using command: /i 2160 100 a few times.

### Expected behaviour
There should be a few stacks of 100 crystal coins under you or the script should tell you that it's an invalid itemid or that you would need to add a ','.

### Actual behaviour
A none valid itemid is created, when you do the command a few times and step on the tile a creature dissapears.

### Environment
In-game.

thexamx:
correct spelling is `/i 2160, 100`

OPfeajoief13:
I know

PR #2314, opened 2017-08-26
Fix create_item param check

Codinablack:
-- Player would created bugged item if no comma was used.
-- closes https://github.com/otland/forgottenserver/issues/2279

### DIFF
--- a/data/talkactions/scripts/create_item.lua
+++ b/data/talkactions/scripts/create_item.lua
@@ -11,6 +11,10 @@ function onSay(player, words, param)
 		return false
 	end
 
+	if not param:find(",") then
+		return false
+	end
+
 	local split = param:split(",")
 
 	local itemType = ItemType(split[1])


EvulMastah:
Wouldn’t it be better to make the count optional?

Codinablack:
Actually the count was already optional until the changes I made here, so I get what you are saying and I didn't think about that...

Anyone else got a better idea how to fix it?

slavidodo:
### Scenario.
1- trim the param string out of spaces.
2- explode using a comma, let's say, into a table called split.
3- in a tmp variable, remove the spaces from the first value of the table. if it can be converted to a number then continue to step 4. If not, then rush to step 5.
4- If the variable contains spaces, then explode the tmp variable into 2 variables, item id/count. You are done to give the item. If not, it's just an item id. the split[2] should be the count if given.
5- It's a string which should be the item name, test if that is correct and then the split[2] is the count. If not, then explode it. Supposing the last value of that table is the count, then the rest of the table is the name. Join them using a space. Test again if it can produce an item type. If not, then return false. If given an item type. then create an item using the name and the last value of the new table.

---
### Scenario in a code.
```lua
-- feel free to reduce the tmp variables and beatify the code
-- code is not tested, it's just to prove the idea
	params = params:trim()
	
	local itemType, count
	local split, tmp1, tmp2, tmp3 = params:split(","), split[1], tmp1:gsub(" ", ""), tonumber(tmp2)
	if tmp3 then -- is number
		if tmp1 == tmp2 then -- doesn't contain spaces
			itemType = ItemType(tmp1)
			count = split[2] and split[2] or 1
			-- tmp1: item id
			-- split[2]: count (or nil then count = 1)
		else
			tmp3 = tmp1:split(" ")
			itemType = ItemType(tmp3[1])
			count = tmp3[2]
			-- tmp3[1]: item id
			-- tmp3[2]: count
		end
	else -- is string
		itemType = ItemType(tmp3)
		if ItemType(tmp3) then
			count = split[2] and split[2] or 1
			--tmp3: item name
			-- split[2]: count
		else
			tmp3 = tmp3:split(" ")
			count = tmp3[#tmp3]
			tmp3:remove(#tmp3)
			itemType(tmp3:concat(" "))
			-- tmp3[1-~]: item name
			-- tmp3[~]: count
		end
	end
	
	if not itemType then
		return false -- we couldn't find an item
	end
	
	...
end
```

Codinablack:
@slavidodo bro, very nice. Thanks man. I'll try to have the talkaction rescripted and committed before the end of the week. I didn't even know string function trim() existed, that's nice. I'll take your code, clean it up as you said, remove the comments and probably rename a couple variables and give it a run. 

I learn something new everyday, today @slavidodo was my teacher 👍

PR #2316, opened 2017-08-28
Remove subid from potions script

raymondtfr:
Removed subIds from potions script so that conditions don't overlap.


ranisalt:
Thx

PR #2317, opened 2017-08-30
Fix #2266 Floor change bug

### LINKED ISSUE #2266: Levitate Bug (game.cpp)
Game.cpp ---> Game::internalMoveCreature

Can't **levitate using trophy stands/parcels/boxes/etc**. multiple times in a row. I can levitate up as many times as I want. But when there are 2 or more levitate spots in a row (a good example is the big wall, west of Svargrond, with about 5 levitate spots in a row).

Seems to be an error when the hasHeight() is 3 after levitating down once.
Levitating up works perfectly fine, multiple times in a row.
But levitating down stops after the first levitate down.

DanielChabrowski:
I wrote it some time ago, it seems to be fixing the issue - players can walk down from 3+ parcels onto 3+ parcels. It would be nice if another person confirmed that first.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -778,10 +778,12 @@ ReturnValue Game::internalMoveCreature(Creature* creature, Direction direction, 
 					}
 				}
 			}
-		} else {
-			//try go down
+		}
+		
+		//try go down
+		if (currentPos.z != 7 && currentPos.z == destPos.z) {
 			Tile* tmpTile = map.getTile(destPos.x, destPos.y, destPos.z);
-			if (currentPos.z != 7 && (tmpTile == nullptr || (tmpTile->getGround() == nullptr && !tmpTile->hasFlag(TILESTATE_BLOCKSOLID)))) {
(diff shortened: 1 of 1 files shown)


thexamx:
https://gyazo.com/6a48aec82ed38d6cd918915c603a307a
Looks like its fixing this issue.

DanielChabrowski:
@thexamx Are you talking about giving player CONDITION_PACIFIED so that he can't use spells/runes or some kind of movement exhaust?

thexamx:
@DanielChabrowski Yes, pacified too, but it looks to spammy IMO, you can just go up and down without any delay immediately. I have no idea how it is on vanila, because I don't have pacc, so I cannot check.

src/game.cpp
```diff
@@ -778,10 +778,12 @@ ReturnValue Game::internalMoveCreature(Creature* creature, Direction direction,
 					}
 				}
 			}
-		} else {
-			//try go down
+		}
+
```
WibbenZ: Trailing otherwise it looks good

PR #2318, opened 2017-09-01
Source reorganization - database, encryption, networking

DanielChabrowski:
For now just moving files, no custom cmake targets.


ranisalt:
Use relative includes instead. `#include "../database/something.h"` is better than magic resolution.

djarek:
@ranisalt
Includes relative to source root are fine IMO.

DanielChabrowski:
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

ranisalt:
I personally don't like it, it adds more places to search for header files, but that's fine if you back it up :P

@DanielChabrowski it's now like we follow that guide throughout the code

DanielChabrowski:
@ranisalt We don't, but it's a good reference point if you're uncertain about something, at least for me.

ranisalt:
Sure, that's true.

My 2 cents is that it doesn't help as long as our source code is as entangled as it currently is. We should abstract some stuff, make the code more generic so that we reduce the amount of includes, forward declare as much as possible because there's no advantage in separating directories if you have to include a lot of stuff in other directories anyway.

Also, pull requests modifying file organization can break a lot of other PRs, so they should be accepted after much more care.

DanielChabrowski:
Ye, it doesn't really help the compiler but it's definitely a step forward.

Until relevant PRs are not merged we can work on directory structure, then merge it with clang-format and possibly clang-tidy, cppcheck. After that we can work on untangling the code and dividing cmake into smaller targets.

PR #2321, opened 2017-09-04
Changing how blessing-related methods are treated

t0kenz:
Usage shouldn't depend on knowing the underlying implementation,
therefore I suggest with this pr that we make it more trivial to use
them in the future.


EvulMastah:
Why not use http://en.cppreference.com/w/cpp/utility/bitset while you're at it?

src/luascript.cpp
```diff
@@ -12634,4 +12634,4 @@ void LuaEnvironment::executeTimerEvent(uint32_t eventIndex)
 	for (auto parameter : timerEventDesc.parameters) {
 		luaL_unref(luaState, LUA_REGISTRYINDEX, parameter);
 	}
-}
+}
```
ranisalt: Revert it

src/player.h
```diff
@@ -1335,4 +1335,4 @@ class Player final : public Creature, public Cylinder
 		friend class ProtocolGame;
 };
 
-#endif
+#endif
```
ranisalt: Revert it

ranisalt:
If you want to change it, then do use bitset as @EvulMastah suggested.

You have made it impossible to set multiple blessings at once.

Kamenuvol:
we should have a bitset using storages maybe

t0kenz:
@ranisalt Yes, I'm aware of this. What do you suggest instead? I still don't agree with having to know the implementation in order to use a method. 

@EvulMastah Regarding bitset, sure that's something we might consider. The reason I didn't include it in this pr was due to past experience with ranisalt wanting me to split up things. But I guess this time it'd be reasonable to have it in the same pr.

DSpeichert:
@t0kenz This is changing the Lua interface and should come with backward-compatibility layer if possible.

If not obvious from the discussion here, the request is for a bitset options as has been suggested.

I'm closing this as it looks abandoned over the years, but do feel free to reopen if you intend to update this PR.

PR #2322, opened 2017-09-04
Moved function getCreatureCondition to compat

raymondtfr:
Remake of #2040.


ranisalt:
Good one! Waiting for the CI to pass.

Issue #2320, opened 2017-09-04
Crash DoTargetCombatCondition


infister:
### Before creating an issue, please ensure:
- [+] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [+] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. create StepIn movement with doTargetCombatCondition
2. step 2 times on it

### Expected behaviour
Not crash

### Actual behaviour
Crash

### Environment
Ubuntu 16

``` Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000000065 in ?? ()
[Current thread is 1 (Thread 0x7f3380cdc700 (LWP 13194))]
(gdb) bt
#0  0x0000000000000065 in ?? ()
#1  0x000000000073fe8c in std::default_delete<Condition const>::operator()(Condition const*) const ()
#2  0x000000000073ee81 in std::unique_ptr<Condition const, std::default_delete<Condition const> >::~unique_ptr() ()
#3  0x000000000073c9d4 in void __gnu_cxx::new_allocator<std::unique_ptr<Condition const, std::default_delete<Condition const> > >::destroy<std::unique_ptr<Condition const, std::default_delete<Condition const> > >(std::unique_ptr<Condition const, std::default_delete<Condition const> >*) ()
#4  0x0000000000739537 in void std::allocator_traits<std::allocator<std::unique_ptr<Condition const, std::default_delete<Condition const> > > >::destroy<std::unique_ptr<Condition const, std::default_delete<Condition const> > >(std::allocator<std::unique_ptr<Condition const, std::default_delete<Condition const> > >&, std::unique_ptr<Condition const, std::default_delete<Condition const> >*) ()
#5  0x00000000007355c4 in std::_Fwd_list_base<std::unique_ptr<Condition const, std::default_delete<Condition const> >, std::allocator<std::unique_ptr<Condition const, std::default_delete<Condition const> > > >::_M_erase_after(std::_Fwd_list_node_base*, std::_Fwd_list_node_base*) ()
#6  0x0000000000730a7c in std::_Fwd_list_base<std::unique_ptr<Condition const, std::default_delete<Condition const> >, std::allocator<std::unique_ptr<Condition const, std::default_delete<Condition const> > > >::~_Fwd_list_base() ()
[...]

PR #2325, opened 2017-09-06
Fix multiple delete called on condition in Lua

DanielChabrowski:
Fixing #2320


ranisalt:
Good one! I'd still use vector though.

PR #2326, opened 2017-09-06
Renamed Combat::setCondition

DanielChabrowski:
Combat::setCondition suggests that only one condition per combat is supported while in reality it adds it to the list of conditions.

I added an alias to compat.lua since it would break existing scripts but I would prefer a clean solution without it or at least with all ~65 setCondition changed in basic Lua scripts.

+ missing return in NpcEventsHandler::onCreatureAppear


ranisalt:
Don't just rename, you'll break old code. Add an alias `setCondition` that prints to the screen it should be fixed by the scripter.

DanielChabrowski:
@ranisalt I already did, without the message tho. Currently I added a warning which will spam the shit out of the console if the old function is used.

data/lib/compat/compat.lua
```diff
@@ -111,11 +111,19 @@ end
 createCombatObject = Combat
 setCombatArea = Combat.setArea
 setCombatCallback = Combat.setCallback
-setCombatCondition = Combat.addCondition
 setCombatFormula = Combat.setFormula
 setCombatParam = Combat.setParameter
+addCombatCondition = Combat.addCondition
```
ranisalt: Put in alphabetic order, below line 111

ranisalt:
Thanks!

PR #2332, opened 2017-09-10
Fix indentation typo & items id signature

wgrriffel:


src/items.cpp
```diff
@@ -111,7 +111,7 @@ FILELOADER_ERRORS Items::loadFromOtb(const std::string& file)
 		return ERROR_INVALID_FORMAT;
 	}
 
-	for(auto & itemNode : root.children) {
+	for (auto & itemNode : root.children) {
```
ranisalt: While at it, remove that space before &

wgrriffel: @ranisalt i have had fixed it before but somehow i rebased and it gone away :rofl:

ranisalt:
Thanks!

PR #2334, opened 2017-09-10
Enhanced monster spells

raymondtfr:
The main reason for this update is to get rid of the way some monster spells are scripted, **looping through combat table objects to be executed**.

I have also updated some spells to use ``onTargetCreature`` callbacks and ``creature:addDamageCondition``, because it gives much more control.


raymondtfr:
Resolved conflicts and squashed similar commits.

diath:
Aren't some of these spells broken? For instance the electrify spells damage the caster rather than the target?

raymondtfr:
I don't think so (tested most of them), the ``addDamageCondition`` adds the damage condition to the target (first argument).

diath:
>the addDamageCondition adds the damage condition to the target (first argument)

If that's the case then never mind what I said.

PR #2335, opened 2017-09-11
Fixing StairHoop delay for spells/runes

### LINKED ISSUE #2330: No StairHoop delay for attacking spells/runes
- [+] This is a bug in the software that resides in this repository, and not a
- [+] This issue is reproducible without changes to the code in this repository

1. Jump on stairs and immediately cast a attacking rune or spell

### Expected behaviour
Should be exhausted for 2s.

### Actual behaviour
You may cast everything, except autoattacking (it seems to be blocked correctly)

thexamx:
This pull request should fix #2330.
Player should not be able to use attacking spells or runes, immediately after stair jump.

### DIFF
--- a/src/spells.cpp
+++ b/src/spells.cpp
@@ -560,6 +560,12 @@ bool Spell::playerSpellCheck(Player* player) const
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


djarek:
Thanks!

PR #2338, opened 2017-09-13
Condition muted should be saved with other persitent conditions

### LINKED ISSUE #2044: Reseting the mute's condition at death

1. Mute a player
2. Kill the player

### Expected behaviour
The player should keep muted.

### Actual behaviour
When the player dies it reset the condition and the player can talk in the Channel which he was muted.

### Environment
The test was maded in Windows 7 - 64 bits.

thexamx:
Fix #2044

### DIFF
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -266,7 +266,7 @@ bool Condition::isPersistent() const
 		return false;
 	}
 
-	if (!(id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT || condition->getType() == CONDITION_MUTED)) {
+	if (!(id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT || conditionType == CONDITION_MUTED)) {
 		return false;
 	}
 
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -266,7 +266,7 @@ bool Condition::isPersistent() const
 		return false;
 	}
 
-	if (!(id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT)) {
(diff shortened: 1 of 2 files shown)


src/iologindata.cpp
```diff
@@ -661,7 +661,7 @@ bool IOLoginData::savePlayer(Player* player)
 	//serialize conditions
 	PropWriteStream propWriteStream;
 	for (Condition* condition : player->conditions) {
-		if (condition->isPersistent()) {
+		if (condition->isPersistent() || condition->getType() == CONDITION_MUTED) {
```
DanielChabrowski: Wouldn't the isPersistent function be a perfect place for that condition type check?

thexamx: Do you mean
```
if (!(id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT || getType() == CONDITION_MUTED)) {
	return false;
}
```
?

DanielChabrowski: Yep.

src/condition.cpp
```diff
@@ -266,7 +266,7 @@ bool Condition::isPersistent() const
 		return false;
 	}
 
-	if (!(id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT)) {
+	if (!(id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT || condition->getType() == CONDITION_MUTED)) {
```
DanielChabrowski: `condition->getType()`
to
`conditionType`

thexamx: I'm sorry, I did fast change and didnt notice.

djarek:
Thanks!

PR #2339, opened 2017-09-14
Moved Creature:addDamageCondition to core

raymondtfr:
Users should also be able to use such metatable method in other kinds of scripts (e.g. Claw of the Noxious Spawn, Poison Arrow, Viper Star, etc.)


diath:
Perhaps you could create some constants for the listType variable rather than using magic numbers.

raymondtfr:
Added such constants, they make it more clear to use that method, although very long to read hah.
I'll update spells to use use such constants later.

@diath

ranisalt:
You'll need to rebase after #2326

ranisalt:
Sweet!

PR #2343, opened 2017-09-20
Fix #2021

### LINKED ISSUE #2021: Rift carpets / void carpets not being saved as house items

1. Place a rift/void carpet (not folded) inside a house
2. Save and restart server

### Expected behaviour
The carpet should be saved and loaded, just like the other items inside the house.

### Actual behaviour
The carpet is not saved or loaded (it is not a moveable item). I'm only talking about those two carpets here, but it probably happens with those new carpets from the Tibia Store.

I fixed this for the rift carpet but I'm not making a pull request, since I'm a noob (sorry for the mess from now on).
**The ID used for the void carpet, in my server, is 25392**:
In _IOMapSerialize::loadItem_
I complemented the conditional:
`if (iType.moveable || !tile) {`
became
`if (iType.moveable || !tile || id == 25392)`
In _IOMapSerialize::saveTile_
I complemented the conditional:
`if (!(it.moveable || item->getDoor() || (item->getContainer() && !item->getContainer()->empty()) || it.canWriteText || item->getBed())) {`
became
`if (!(it.moveable || item->getID() == 25392 || item->getDoor() || (item->getContainer() && !item->getContainer()->empty()) || it.canWriteText || item->getBed())) {`

vankk:
This should fix #2021 

The only thing that we need is to set in items.xml the attribute unmoveableditem with the value that is equal to 1.

Tested and using in my servers.


src/items.h
```diff
@@ -208,6 +208,7 @@ class ItemType
 		int32_t extraDefense = 0;
 		int32_t armor = 0;
 		uint16_t rotateTo = 0;
+		uint16_t unmoveabledItem = 0;
```
DanielChabrowski: Shouldn't that be a boolean? What the hell "unmoveabled" means and is it telling you that it can be saved in a house or will it confuse ppl with the difference between "moveable" member?

Kamenuvol:
It should be a forceSave attribute

vankk:
I vote for _forceSave_. _isFurniture_ sounds vague.

DanielChabrowski:
`isFurniture` would only confuse as to why carpet is a furniture and chair suddenly isn't.
`forceSave` is a little vague as well - what save, player save?

I would opt for `forceHouseSave` or something similar.

src/items.h
vankk: I wasn't using as a boolean and that worked as well too, but I'll move it to a boolean.

ranisalt:
But aren't chairs saved with a house, too?

EvulMastah:
In TFS 0.4 this attribute was called `forceSerialize` and you could enable it with an attribute name of `forceSerialize`/`forceSerialization`/`forceSave` in items.xml. I guess that fits well, since the class that is saving items is called IOMapSerialize.

src/enums.h
```diff
@@ -89,6 +89,7 @@ enum itemAttrTypes : uint32_t {
 	ITEM_ATTRIBUTE_CHARGES = 1 << 20,
 	ITEM_ATTRIBUTE_FLUIDTYPE = 1 << 21,
 	ITEM_ATTRIBUTE_DOORID = 1 << 22,
+	ITEM_ATTRIBUTE_UNMOVEABLEDITEM = 1 << 23,
```
thexamx: why not just make this attribute, ITEM_ATTRIBUTE_CARPET?

src/items.h
thexamx: Same as @DanielChabrowski 
IMO it should be boolean.
Also I would change the name to isCarpet.

DSpeichert:
Should we instead call it `ITEM_ATTR_FORCE_SAVE` and make it boolean?

It's weird to have both "movable" and "unmovable" properties on an item.

ninjalulz:
Implemented in https://github.com/otland/forgottenserver/commit/1f528718b5666674a1901b7b7bf36d488d28e6e2

PR #2347, opened 2017-09-28
Small adjustment to combatChangeMana and combatChangeHealth

nekiro:
IMO combatChangeMana and combatChangeHealth shouldn't be executed when the damage or manachange is 0. There is no reason to execute it. Correct me if I'm wrong.


vankk:
I don't think we should approves that and I'll use me as an example: in my servers I uses combatChangeHealth and mana to always increase the damage, always increase the value damage in 300.

What I mean is, what if someone want increase a value even it is in zero with a magic number or even uses if the value is zero then increase?

DSpeichert:
This would be a change of behavior that is not a clear improvement. It's also hardly affecting performance to execute that every time.
Per @vankk's case, it may limit functionality.

PR #2350, opened 2017-10-04
Fix inconsistencies between declaration and definition arg names

marksamman:
This pull request fixes inconsistencies in the argument names between declarations and definitions. There shouldn't be any semantic changes.


src/ban.h
```diff
@@ -51,7 +51,7 @@ class IOBan
 {
 	public:
 		static bool isAccountBanned(uint32_t accountId, BanInfo& banInfo);
-		static bool isIpBanned(uint32_t ip, BanInfo& banInfo);
+		static bool isIpBanned(uint32_t clientip, BanInfo& banInfo);
```
ranisalt: `clientIp`?

src/combat.cpp
```diff
@@ -506,17 +506,17 @@ void Combat::CombatHealthFunc(Creature* caster, Creature* target, const CombatPa
 	}
 }
 
-void Combat::CombatManaFunc(Creature* caster, Creature* target, const CombatParams& params, CombatDamage* data)
+void Combat::CombatManaFunc(Creature* caster, Creature* target, const CombatParams& params, CombatDamage* damage)
 {
-	assert(data);
-	CombatDamage damage = *data;
-	if (damage.primary.value < 0) {
+	assert(damage);
```
ranisalt: While at this, why is this assert here? It doesn't help and the server would crash anyway the next line.

PR #2352, opened 2017-10-04
Fix statements that check if unsigned types are less than zero

marksamman:
cppcheck also claims this if anyone wants to have a look at it:
```
[src/fileloader.cpp:117]: (style) Condition '!lastEscaped' is always true
```

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1353,7 +1353,7 @@ int64_t Creature::getStepDuration() const
 	int32_t stepSpeed = getStepSpeed();
 	if (stepSpeed > -Creature::speedB) {
 		calculatedStepSpeed = floor((Creature::speedA * log((stepSpeed / 2) + Creature::speedB) + Creature::speedC) + 0.5);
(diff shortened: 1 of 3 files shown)


src/creature.cpp
```diff
@@ -1353,7 +1353,7 @@ int64_t Creature::getStepDuration() const
 	int32_t stepSpeed = getStepSpeed();
 	if (stepSpeed > -Creature::speedB) {
 		calculatedStepSpeed = floor((Creature::speedA * log((stepSpeed / 2) + Creature::speedB) + Creature::speedC) + 0.5);
-		if (calculatedStepSpeed <= 0) {
```
ranisalt: This one could be replaced with a `std::max`, no?

ranisalt:
I see cppcheck confused on that line because 0xFD (`Node::ESCAPE`) is always larger than any char value, and it doesn't handle the static cast properly on analysis.

I watched it on gdb and `lastEscaped` equals `true` once loading the default map.

PR #2362, opened 2017-10-16
* check if eventList is empty to prevent an invalid task in scheduler thread code

### LINKED ISSUE #2361: Possible crash in scheduler code when the server is shutting down
I was debugging my server with visual studio and after i close the console window to close the server i had a crash here:
https://github.com/otland/forgottenserver/blob/master/src/scheduler.cpp#L47
As the console window had the "Shutting down..." message i think Scheduler::shutdown() has been called before the crash. Inside Scheduler::shutdown() eventList priority queue and eventIds becomes empty. 
Its possible that "ret = eventSignal.wait_until(eventLockUnique, eventList.top()->getCycle());" is waiting for a task to be executed while the Scheduler::shutdown() is called by dispatcher thread and then this code "SchedulerTask* task = eventList.top();" will get a invalid task and make the crash possible.

My fix to this issue is to change this:
`if (ret == std::cv_status::timeout) {`
To this:
`if ((ret == std::cv_status::timeout) && !eventList.empty()) {`

@marksamman @djarek @ranisalt

HeavenIsLost:
An attempt to fix #2361 
@djarek @marksamman @ranisalt

### DIFF
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -35,7 +35,7 @@ void Scheduler::threadMain()
 		}
 
 		// the mutex is locked again now...
-		if ((ret == std::cv_status::timeout) && !eventList.empty()) {
+		if (ret == std::cv_status::timeout && !eventList.empty()) {
 			// ok we had a timeout, so there has to be an event we have to execute...
 			SchedulerTask* task = eventList.top();
 			eventList.pop();
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -36,6 +36,7 @@ void Scheduler::threadMain()
 
 		// the mutex is locked again now...
 		if (ret == std::cv_status::timeout) {
+		if ((ret == std::cv_status::timeout) && !eventList.empty()) {
 			// ok we had a timeout, so there has to be an event we have to execute...
 			SchedulerTask* task = eventList.top();
 			eventList.pop();
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -35,7 +35,6 @@ void Scheduler::threadMain()
 		}
 
 		// the mutex is locked again now...
-		if (ret == std::cv_status::timeout) {
 		if ((ret == std::cv_status::timeout) && !eventList.empty()) {
 			// ok we had a timeout, so there has to be an event we have to execute...
 			SchedulerTask* task = eventList.top();


src/scheduler.cpp
```diff
@@ -35,7 +35,7 @@ void Scheduler::threadMain()
 		}
 
 		// the mutex is locked again now...
-		if (ret == std::cv_status::timeout) {
+		if ((ret == std::cv_status::timeout) && !eventList.empty()) {
 			// ok we had a timeout, so there has to be an event we have to execute...
```
ranisalt: What about this comment? Is it wrong, then?

HeavenIsLost: I made a mistake, i quick copied the fix from own src and its why there this extra "if (ret == std::cv_status::timeout) {" that will prevent the source to be compiled(its why the pr has two commits, the second one is to fix the stupid mistake in the first one). About the comment "the mutex is locked again now..." its not wrong. The problem lies on "eventSignal.wait_until". in the documentation http://en.cppreference.com/w/cpp/thread/condition_variable/wait_until its says "The thread will be unblocked when notify_all() or notify_one() is executed, or when the absolute time point timeout_time is reached. It may also be unblocked spuriously. When unblocked, regardless of the reason, lock is reacquired and wait_until exits." So while the eventSignal.wait_until is waiting the cycles the mutex is unlocked and then its possible to Scheduler::shutdown() clean both lists(eventIds and eventList).

HeavenIsLost: In tfs 1.1 there is an extra check that does not exists in tfs 1.2 code: "if (ret == std::cv_status::timeout && threadState != THREAD_STATE_TERMINATED) {" also in tfs 0.4 code: "if(!action && Scheduler::m_threadState != Scheduler::STATE_TERMINATED)" Maybe the old code was made to prevent this issue and in some rework it has been removed? @djarek

djarek: I'm too lazy to check the history, but this might've been broken when code related to thread management was abstracted away.

src/scheduler.cpp
```diff
@@ -35,7 +35,7 @@ void Scheduler::threadMain()
 		}
 
 		// the mutex is locked again now...
-		if (ret == std::cv_status::timeout) {
+		if ((ret == std::cv_status::timeout) && !eventList.empty()) {
```
ranisalt: You added redundant parenthesis.

HeavenIsLost: @ranisalt so you want me to remove the ()? i thought it was needed as there is two checks in the same IF statement.

ranisalt: It is not. Logical expressions have lower priority than equality or unary expressions, so it would parse as expected.

HeavenIsLost:
@ranisalt ready to merge

andersonfaaria:
Why this isn't merged yet?

ranisalt:
Thanks!

PR #2368, opened 2017-10-30
Fix party extraExpRate

raymondtfr:
This data member is [initialized](https://github.com/otland/forgottenserver/pull/2368/files#diff-bca280ff47a20e24267a1c6b5caf6139L96) on Party class instantiation and _possibly_ assigned the same value [``0.20f``](https://github.com/otland/forgottenserver/pull/2368/files#diff-c5af55d870b9bc2502667c495bad41f7L381), once a new member [leaves](https://github.com/raymondtfr/forgottenserver/blob/31a16d765b403767eba0d7c83bedce9be81c817e/src/party.cpp#L130) or [join](https://github.com/raymondtfr/forgottenserver/blob/31a16d765b403767eba0d7c83bedce9be81c817e/src/party.cpp#L221) a party.

Am I right about this? 🙄

### DIFF
--- a/src/party.cpp
+++ b/src/party.cpp
@@ -375,11 +375,7 @@ void Party::updateVocationsList()
 	}
 
 	size_t size = vocationIds.size();
-	if (size > 1) {
-		extraExpRate = static_cast<float>(size * (10 + (size - 1) * 5)) / 100.f;
-	} else {
-		extraExpRate = 0.20f;
-	}
+	extraExpRate = size > 1 ? static_cast<float>(size * (10 + (size - 1) * 5)) / 100.f : 0.20f;
 }
 
 bool Party::setSharedExperience(Player* player, bool sharedExpActive)
--- a/src/party.cpp
+++ b/src/party.cpp
@@ -127,7 +127,7 @@ bool Party::leaveParty(Player* player)
 	player->sendTextMessage(MESSAGE_INFO_DESCR, "You have left the party.");
 
 	updateSharedExperience();
-	updateVocationsList();
+	updateSharedExperienceBonus();
 
 	clearPlayerPoints(player);
 
@@ -218,7 +218,7 @@ bool Party::joinParty(Player& player)
 
 	player.removePartyInvitation(this);
 	updateSharedExperience();
-	updateVocationsList();
+	updateSharedExperienceBonus();
 
 	const std::string& leaderName = leader->getName();
 	ss.str(std::string());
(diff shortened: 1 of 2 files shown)


raymondtfr:
Is there any reason for [this](https://github.com/otland/forgottenserver/pull/2368/commits/836cd68396099d8b94e5ad1a2d35ab6d964c79de#diff-bca280ff47a20e24267a1c6b5caf6139L24)?

ranisalt:
> Is there any reason for this?

Yes, someone put it there, fast forward a few months and it's not needed anymore 😝 remove if you feel like

src/party.cpp
```diff
@@ -375,11 +375,7 @@ void Party::updateVocationsList()
 	}
 
 	size_t size = vocationIds.size();
-	if (size > 1) {
-		extraExpRate = static_cast<float>(size * (10 + (size - 1) * 5)) / 100.f;
-	} else {
-		extraExpRate = 0.20f;
-	}
+	sharedExpBonus = size > 1 ? static_cast<float>(size * (10 + (size - 1) * 5)) / 100.f : 0.20f;
```
ranisalt: Replace this ternary with `std::min(0.2f, <big function here>)` .

I think you don't even need the float cast since division by float is always float, too. And I wouldn't even use `float` since it is casted to `double` everywhere.

raymondtfr: Should I change it's data type to ``double`` then?

I think the ``#include "monsters.h"`` is because someone was probably going to add a missing check regarding the shared experience mechanism, in which it should not apply the bonus to the ``shareExperience`` variable in ``Party::shareExperience`` in case the target creature (a monster) does not yield atleast ``20`` experience points.

I'll remove it nonetheless, but I may research this with more caution in vanilla, and I'll make a PR for it. But I think it is something like this:
```C++
void Party::shareExperience(uint64_t experience, Creature* source/* = nullptr*/)
{
	uint64_t shareExperience;
	if (source) {
		MonsterType* monsterType = g_monsters.getMonsterType(source:getName());
		if (monsterType && monsterType->getExperience() < 20) {
			shareExperience = experience
		}
	} else {
		shareExperience = static_cast<uint64_t>(std::ceil((static_cast<double>(experience) * (sharedExpBonus + 1)) / (memberList.size() + 1)))
	}

	for (Player* member : memberList) {
		member->onGainSharedExperience(shareExperience, source);
	}
	leader->onGainSharedExperience(shareExperience, source);
}
```

ranisalt: Yes, sorry, I mistyped `min` for `max`, it should be the latter.

You can safely change to `double`, as far as I am concerned there are no drawbacks.

PR #2371, opened 2017-11-10
Avoid magic rope be casted outside of a hole.

vankk:
This change will avoid the spell magic rope _AKA_ exani tera be casted outside of a hole.

### DIFF
--- a/data/spells/scripts/support/magic_rope.lua
+++ b/data/spells/scripts/support/magic_rope.lua
@@ -3,16 +3,18 @@ function onCastSpell(creature, variant)
 	position:sendMagicEffect(CONST_ME_POFF)
 
 	local tile = Tile(position)
-	if table.contains(ropeSpots, tile:getGround():getId()) or tile:getItemById(14435) then
-		tile = Tile(position:moveUpstairs())
-		if tile then
-			creature:teleportTo(position, false)
-			position:sendMagicEffect(CONST_ME_TELEPORT)
-		else
-			creature:sendCancelMessage(RETURNVALUE_NOTENOUGHROOM)
-		end
-	else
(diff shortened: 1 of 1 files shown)


ranisalt:
I'm not sure that's how it works. I remember falling into a hole without rope spot and being able to escape with exani tera.

data/spells/scripts/support/magic_rope.lua
```diff
@@ -3,16 +3,18 @@ function onCastSpell(creature, variant)
...
-	if table.contains(ropeSpots, tile:getGround():getId()) or tile:getItemById(14435) then
-		tile = Tile(position:moveUpstairs())
-		if tile then
-			creature:teleportTo(position, false)
-			position:sendMagicEffect(CONST_ME_TELEPORT)
-		else
-			creature:sendCancelMessage(RETURNVALUE_NOTENOUGHROOM)
-		end
-	else
+	if not table.contains(ropeSpots, tile:getGround():getId()) or not tile:getItemById(14435) then
```
TheSumm: The inverse of `table.contains(ropeSpots, tile:getGround():getId()) or tile:getItemById(14435)` is `not table.contains(ropeSpots, tile:getGround():getId()) and not tile:getItemById(14435))`

TheSumm:
@ranisalt You cannot use exani tera if unless you are standing on any kind of rope spot. Otherwise no mana will be used, you will poff and receive the message "Sorry, not possible."

ranisalt:
Hmm, sure. My mind is playing tricks :P

data/spells/scripts/support/magic_rope.lua
vankk: Thanks Summ. Good eye! :)

Dagst:
Did you test it?
Does not work in the hole

PR #2374, opened 2017-11-21
Fix luaGameReload not working with RELOAD_TYPE_ALL param

diath:
RELOAD_TYPE_ALL is equal to 0 which means Game.reload would never get past that check effectively making the type unusable.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -4283,11 +4283,6 @@ int LuaScriptInterface::luaGameReload(lua_State* L)
 {
 	// Game.reload(reloadType)
 	ReloadTypes_t reloadType = getNumber<ReloadTypes_t>(L, 1);
-	if (!reloadType) {
-		lua_pushnil(L);
-		return 1;
-	}
-
 	if (reloadType == RELOAD_TYPE_GLOBAL) {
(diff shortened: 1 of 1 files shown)


src/luascript.cpp
```diff
@@ -4283,11 +4283,6 @@ int LuaScriptInterface::luaGameReload(lua_State* L)
 {
 	// Game.reload(reloadType)
 	ReloadTypes_t reloadType = getNumber<ReloadTypes_t>(L, 1);
-	if (!reloadType) {
```
EvulMastah: Default behaviour of Game::reload is to reload everything. Wouldn't it be better to introduce some validation here?

diath: I think Game::reload should check whether the reloadType is valid or not by explicitly handling the RELOAD_TYPE_ALL case and showing a warning for the default case of a switch and reporting the error back but since @ranisalt said that Game::reload should default to RELOAD_TYPE_ALL in #1520 I don't think that's the desired behavior. This PR simply addresses the problem of not being able to pass RELOAD_TYPE_ALL to luaGameReload function.

marksamman:
Thanks!

PR #2377, opened 2017-12-02
complement to getTileInfo

psychonaut04:
Without it we need to do a hasFlag after already getting the ret if we wanna check for pvp and nopvp flags too after calling the function.


WibbenZ:
The compat file should not be updated with things like this.
Unless 0.2 had this implemented (does not) it should be as is, the compat file is just there to help when you convert scripts, not to run the server.

PR #2379, opened 2017-12-06
Setting default player capacity to 400

vamo89:
Editing schema.sql, setting default player capacity to 400, like the official tibia.


DSpeichert:
This needs a migration too. Default would be different in existing databases.

PR #2385, opened 2017-12-18
Fixed displaying Premium Account (status) in client

slawkens:
After Mark update Premium Status stopped showing in Tibia Client.

This push fixes it.
Now client displays correct Premium Status (either free or premium with days).

### DIFF
--- a/src/protocollogin.cpp
+++ b/src/protocollogin.cpp
@@ -107,7 +107,7 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
 		output->addByte(1);
 		output->add<uint32_t>(0);
 	} else {
-		output->addByte(0);
+		output->addByte(account.premiumDays > 0 ? 1 : 0);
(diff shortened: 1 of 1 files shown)


DSpeichert:
Can you link to the commit that broke it?

slawkens:
In this commit there were some changes how packet were sent to the client:
https://github.com/otland/forgottenserver/commit/2cc16851e69d9dd9ad8a80c9add3312de9d2d890#diff-4aa7a24bdaebd6ae3f1ee9500e6b47e0

I think it doesn't work since then. I saw other posts on otland about it.

DSpeichert:
image
Is OTClient showing wrong number of days left or was this also not working properly? I should only have 2 days left per database.

slawkens:
As you can see I've only changed one byte from 0 to 1 (if premium). Nothing with days. So it might be something with OTClient.

Tested only with official Tibia Client 10.98.

PR #2386, opened 2017-12-18
Store towns list in database

slawkens:
To be used by AAC creators

### DIFF
--- a/data/globalevents/scripts/startup.lua
+++ b/data/globalevents/scripts/startup.lua
@@ -36,4 +36,11 @@ function onStartup()
 		until not result.next(resultId)
 		result.free(resultId)
 	end
+
+	-- store towns in database
+	db.query("TRUNCATE TABLE `towns`")
+	for i, town in ipairs(Game.getTowns()) do
+		local position = town:getTemplePosition()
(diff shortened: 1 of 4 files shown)


DSpeichert:
Looks good. Are there any other attributes for towns besides name and position?

slawkens:
No. Only those you mentioned.

From towns.h:
```
	private:
		uint32_t id;
		std::string name;
		Position templePosition;
```

Also it might be useful to have foreign keys on houses.town_id and players.town_id, but then I'll need to rewrite that lua part cause it will throw an error.

Znote:
Why not use asyncquery here in startup.lua?

Should the code in startup.lua that populates the db also be present in migrations/20.lua? 
After the table has been created. There might be a future migration that fix table dependencies and return key constraint errors. (like fix players.town_id or houses.town_id in migration/25), which probably runs before startup.lua. 
Does the tibia client have a limit on town name length?

DSpeichert:
What if the map changes? How do we handle updates?

Znote:
@DSpeichert Thats something to bring up if we connect tables key PK-FK constraints (Not this PR). Currently that connection is not there, and startup.lua truncates and reimports the table on every startup, any changes in map will be reflected in the db.

PR #2389, opened 2017-12-23
Blood rage spell fix

nekiro:
According to tibia wikia:
>Concentrating on the fight, a knight will neglect his defense being unable to block any attack with his shield.

Blood rage spell shouldn't set shielding to 0. It should disable your shield defense, so you cant block any attack with it.


joseluis2g:
Really nice, thanks.

src/condition.cpp
```diff
@@ -396,6 +397,8 @@ bool ConditionAttributes::startCondition(Creature* creature)
 		return false;
 	}
 
+	creature->setUseDefense(disableDefense);
```
DSpeichert: Shouldn't that be `creature->setUseDefense(!disableDefense);` ?

src/condition.cpp
```diff
@@ -508,6 +511,10 @@ void ConditionAttributes::endCondition(Creature* creature)
 			player->sendStats();
 		}
 	}
+
+	if (disableDefense) {
+		creature->setUseDefense(!disableDefense);
```
DSpeichert: This should be `creature->setUseDefense(!disableDefense);` or `creature->setUseDefense(true);`

nekiro: Yes, you are right. I got confused. because I changed canUseDefense to true by default and forgot about these.

First comment is right too.

nekiro: But second should be `creature->setUseDefense(disableDefense);` right?

DSpeichert: But inside that `if` it is guarantee that `disableDefense` is true, so you can just set it to `true` directly, without using variable.

PR #2392, opened 2017-12-25
Fixing creating ItemType with empty string

### LINKED ISSUE #2328: Creating ItemType with empty string
- [+ ] This is a bug in the software that resides in this repository, and not a
- [+ ] This issue is reproducible without changes to the code in this repository

1. local itemType = ItemType("")
2. itemType:getId() will be different from 0, as it will point to first item without name:
`<item id="9060" />`

### Expected behaviour
It should return error that there is no such itemtype

### Actual behaviour
It returns item without name.

nekiro:
This pull request should fix #2328.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -10717,6 +10717,11 @@ int LuaScriptInterface::luaItemTypeCreate(lua_State* L)
 		id = Item::items.getItemIdByName(getString(L, 2));
 	}
 
+	if (id == 0) {
+		lua_pushnil(L);
+		return 1;
+	}
+
 	const ItemType& itemType = Item::items[id];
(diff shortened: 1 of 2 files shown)


src/luascript.cpp
```diff
@@ -10717,6 +10717,11 @@ int LuaScriptInterface::luaItemTypeCreate(lua_State* L)
 		id = Item::items.getItemIdByName(getString(L, 2));
```
ranisalt: You should add the check above this line instead. If string is empty, don't even call the function.

nekiro: Yes I know, but `uint16_t Items::getItemIdByName(const std::string& name)` should not return empty name item too.

PR #2393, opened 2017-12-27
Fixing getting useItem

nekiro:
This pull request is related to #1851, and #1970.

I tested this with cipsoft client. All seems to work correctly, test it please. I could forget some code, because it's late.

### DIFF
--- a/src/tile.cpp
+++ b/src/tile.cpp
@@ -1601,7 +1601,7 @@ bool Tile::isMoveableBlocking() const
 	return !ground || hasFlag(TILESTATE_BLOCKSOLID);
 }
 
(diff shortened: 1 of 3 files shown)


src/tile.cpp
```diff
@@ -1601,22 +1601,16 @@ bool Tile::isMoveableBlocking() const
 	return !ground || hasFlag(TILESTATE_BLOCKSOLID);
 }
 
-Item* Tile::getUseItem() const
+Item* Tile::getUseItem(const int32_t& index) const
```
djarek: Pass by value. `(int32_t index)`

PR #2394, opened 2017-12-27
Fix shovel use on bloody sand ID 9059

### LINKED ISSUE #2073: Use shovel on blood
Can't use shovel on ground if have blood there:

http://prntscr.com/dtxbmb

DSpeichert:
Fixes #2073


nekiro:
This issue should be fixed when we commited #2393 for the stackpos, because cipsoft client ignores blood spots when sending item stackpos.

DSpeichert:
This sand is not actually supposed to be used with shovel.

PR #2395, opened 2017-12-27
Adding "boosts" introduced in Tibia 10.94

nekiro:
Added life leech, mana leech and critical hit support.
You can add them to player with condition or set attribute to item in items.xml. This pr does not support unique item attributes which means you can't set these attributes to one unique item. You can do it only for items. I can add support for attributes too, if it's needed.

If you are not familiar with Tibia, here is how it works:
Mana and hp leech just converts x% of damage dealt with y chance to creature and return it to player health or mana (as healing). Critical hit increases damage dealt to creature by x% with y chance.


ranisalt:
LGTM. Isn't it part of imbuement system?

Olimpotibia:
@nekiro , Tested and i discover some strange thing or maybe is my tfs have some outdated, when im using full boost liveleech on some monsters for example single targer works perfect, but also with more of two creatures the Health points absorb is x2=2creatures, that is correct?

nekiro:
That’s old and deprecated commit, if you want to test go to #2497

Issue #2018, opened 2016-12-04
Boost exception while loading files


wgrriffel:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Game.loadMap("") to a nonexistent path or rename map/items otb
2. Crash

It will crash if path or file doest exist.

### Expected behaviour
Should display error output instead and not crash

### Actual behaviour
`theforgottenserver.exe!boost::throw_exception<std::ios_base::failure>(const std::ios_base::failure & e) Line 69`

### Environment
Win7/Debian 8

djarek:
https://github.com/otland/forgottenserver/blob/master/src/iomap.cpp#L72
I think the construction of the loader should be wrapped with a try ... catch block, an error should be printed, and the exception should probably be rethrown. Could you confirm the same happens if the item.otb file is missing?

wgrriffel:
@djarek yes it does the same, exeption thrown at `theforgottenserver.exe!boost::throw_exception<std::ios_base::failure>(const std::ios_base::failure & e) Line 69`

wgrriffel:
@djarek even so, don't you think it's an overkill terminate the server for this reason? i mean for files that are not a dependency of server, such as loading a new map with Game.loadMap.

wgrriffel:
@dalkon btw don't you think it's an overkill throw an exeption because of a bad file path? i'm not about core files loaded on server startup, im about simply map loaded in Game.loadMap

djarek:
I guess it's safe to rework the error handling in the OTB parser, because failure in OTB tree construction does not alter the map data structures.

Issue #2396, opened 2017-12-27
Migrate to C++14


djarek:
We should migrate to C++14 to get all the language sugar available. The migration should be fairly smooth, because currently all platforms have support for this version of the standard.

ranisalt:
I guess we could rethink on #1439

PR #2397, opened 2017-12-27
Catch the exception when loading map fragments in lua

djarek:
The server should not crash if the loading process failed, since the server is left in a valid state.

Resolves: #2018

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -3991,7 +3991,15 @@ int LuaScriptInterface::luaGameLoadMap(lua_State* L)
 {
 	// Game.loadMap(path)
 	const std::string& path = getString(L, 1);
-	g_dispatcher.addTask(createTask(std::bind(&Game::loadMap, &g_game, path)));
+	g_dispatcher.addTask(createTask( [path]() {
+		try {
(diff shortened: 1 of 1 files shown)


src/luascript.cpp
```diff
@@ -3991,7 +3991,15 @@ int LuaScriptInterface::luaGameLoadMap(lua_State* L)
 {
 	// Game.loadMap(path)
 	const std::string& path = getString(L, 1);
-	g_dispatcher.addTask(createTask(std::bind(&Game::loadMap, &g_game, path)));
+	g_dispatcher.addTask(createTask( [path]() {
```
ranisalt: We don't usually put spaces between ( and [, and I think `path` should be passed by reference.

djarek: You can't pass path by reference, because the temporary returned from getString() could expire before the lambda is executed.

ranisalt: Sure, but you could move it then

djarek: Is move capture allowed in C++11?

ranisalt: One more to #2396

PR #2401, opened 2017-12-30
Add 32bit to appveyor

WibbenZ:


ranisalt:
Not sure why you closed it, but for discussion: it is worthless, recent Windows versions do not even have x86-32 variants, Arch Linux dropped support for it and x86-32 is on the road of extinction.

WibbenZ:
Well it could be re-opened if we can get it to work.
The only reason is because some users with ex Windows 7 32bit would like to have it.

And ex ORTS has problems running 64bit on Windows

DSpeichert:
It works without any issues, the trick is that the platform is called `Win32` instead of `x86`. But it slows down build and therefore Appveyor feedback.

ranisalt:
> The only reason is because some users with ex Windows 7 32bit would like to have it.

That's a very specific target audience, Windows 32-bit version users. Steam survey shows that's less than 2% total, and that's between gamers, our count is probably less than 1% - not worth the extra build time.

PR #2402, opened 2017-12-31
Fixing #2056

### LINKED ISSUE #2056: Reload chatchannels, bug?
When reloading chatchannels, you are unable to write to any channel such as Help/World Chat/Gamemaster etc until you close and reopen them again (or relog).

Is this a bug or are all channels purposefully wiped from memory when reloading?

nekiro:
This commit should fix #2056

### DIFF
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -294,21 +294,34 @@ bool Chat::load()
 		return false;
 	}
 
-	std::forward_list<uint16_t> removedChannels;
-	for (auto& channelEntry : normalChannels) {
-		ChatChannel& channel = channelEntry.second;
-		channel.onSpeakEvent = -1;
-		channel.canJoinEvent = -1;
-		channel.onJoinEvent = -1;
-		channel.onLeaveEvent = -1;
-		removedChannels.push_front(channelEntry.first);
-	}
-
 	for (auto channelNode : doc.child("channels").children()) {
-		ChatChannel channel(pugi::cast<uint16_t>(channelNode.attribute("id").value()), channelNode.attribute("name").as_string());
-		channel.publicChannel = channelNode.attribute("public").as_bool();
-
+		uint16_t channelId = pugi::cast<uint16_t>(channelNode.attribute("id").value());
+		std::string channelName = channelNode.attribute("name").as_string();
+		bool isPublic = channelNode.attribute("public").as_bool();
 		pugi::xml_attribute scriptAttribute = channelNode.attribute("script");
+		
+		auto it = normalChannels.find(channelId);
+		if (it != normalChannels.end()) {
+			ChatChannel& channel = it->second;
+			channel.publicChannel = isPublic;
+			channel.name = channelName;
+
+			if (scriptAttribute) {
+				if (scriptInterface.loadFile("data/chatchannels/scripts/" + std::string(scriptAttribute.as_string())) == 0) {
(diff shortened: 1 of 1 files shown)


src/chat.cpp
```diff
@@ -294,21 +294,34 @@ bool Chat::load()
...
+		auto it = normalChannels.find(channelId);
+		if (it != normalChannels.end()) {
+			ChatChannel& channel = it->second;
+			channel.publicChannel = isPublic;
+			channel.name = channelName;
+
+			if (scriptAttribute) {
+				if (scriptInterface.loadFile("data/chatchannels/scripts/" + std::string(scriptAttribute.as_string())) == 0) {
+					channel.onSpeakEvent = scriptInterface.getEvent("onSpeak");
+					channel.canJoinEvent = scriptInterface.getEvent("canJoin");
```
DanielChabrowski: Shouldn't we readd users to the channel? If 'canJoin' changed and there are different conditions for joining there could be players that shouldn't be there. Also, you could expect some values set in 'onJoin' that wouldn't be present after a reload because the event is not executed for players that are already in the channel

nekiro: Yeah, you are right. I didn't think about this, just pushed commit, review it please.

src/chat.cpp
```diff
@@ -294,21 +294,42 @@ bool Chat::load()
...
+				}
+			}
+
+			UsersMap tempUserMap;
+			tempUserMap.insert(channel.users.begin(), channel.users.end());
+			channel.users.clear();
+			for (const auto& pair : tempUserMap) {
+				channel.addUser(*pair.second);
+			}
+			tempUserMap.clear();
```
djarek: Clearing the map explicitly is redundant if it's just going to be destroyed.

src/chat.cpp
```diff
@@ -294,21 +294,42 @@ bool Chat::load()
...
+					channel.onSpeakEvent = scriptInterface.getEvent("onSpeak");
+					channel.canJoinEvent = scriptInterface.getEvent("canJoin");
+					channel.onJoinEvent = scriptInterface.getEvent("onJoin");
+					channel.onLeaveEvent = scriptInterface.getEvent("onLeave");
+				} else {
+					std::cout << "[Warning - Chat::load] Can not load script: " << scriptAttribute.as_string() << std::endl;
+				}
+			}
+
+			UsersMap tempUserMap;
```
djarek: Copying map entries when you're just going to clear the source makes no sense
```cpp
UsersMap tempUserMap = std::move(channel.users);
```

src/chat.cpp
```diff
@@ -294,21 +294,39 @@ bool Chat::load()
...
-
 	for (auto channelNode : doc.child("channels").children()) {
-		ChatChannel channel(pugi::cast<uint16_t>(channelNode.attribute("id").value()), channelNode.attribute("name").as_string());
-		channel.publicChannel = channelNode.attribute("public").as_bool();
-
+		uint16_t channelId = pugi::cast<uint16_t>(channelNode.attribute("id").value());
+		std::string channelName = channelNode.attribute("name").as_string();
+		bool isPublic = channelNode.attribute("public").as_bool();
 		pugi::xml_attribute scriptAttribute = channelNode.attribute("script");
+
```
WibbenZ: You created another trailing? :P

PR #2403, opened 2017-12-31
Ensure UTF8 character set in DB

### LINKED ISSUE #2175: Database charset must be UTF-8
- [ ] This issue is reproducible without changes to the code in this repository

1. Create database with character set different than `utf8` (e.g. MariaDB defaults to `latin1`)
2. Import schema.sql

### Expected behaviour
Schema tables should be created on the database.

### Actual behaviour
It fails with error message `ERROR 1071 (42000) at line 15: Specified key was too long; max key length is 767 bytes`.

It happens because characters on `latin1` use 4 bytes on MySQL, and there is an unique key on `players.name` (and others, maybe) which is 255 characters, thus 1020 bytes, hitting the 767 bytes limit on keys. It's not an issue on UTF-8 as it uses 3 bytes per character, thus 765 bytes just short of the limit.

We should either manually set the column character set to `utf8` on every indexed column or set it to the entire database, which is a better option IMHO since utf8 is ubiquitous and should prevent other problems with the database.

DSpeichert:
Fixes #2175


schema.sql
```diff
@@ -300,35 +300,35 @@ CREATE TABLE IF NOT EXISTS `player_items` (
...
+) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
 CREATE TABLE IF NOT EXISTS `server_config` (
   `config` varchar(50) NOT NULL,
   `value` varchar(256) NOT NULL DEFAULT '',
   PRIMARY KEY `config` (`config`)
-) ENGINE=InnoDB;
+) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
 INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '20'), ('motd_hash', ''), ('motd_num', '0'), ('players_record', '0');
```
Znote: Update db version in this insert query

slawkens:
Just a question - are there any drawbacks when doing this?

I'm asking cause I would like to do the same with MyAAC.

Will there be any side-effects when converting forum posts, newses, and any other data that may contain some incompatible characters?

ranisalt:
> Will there be any side-effects when converting forum posts, newses, and any other data that may contain some incompatible characters?

Might be, if your texts contain any non-ASCII characters, but the database usually handles it properly if you add a migration query.

slawkens:
As migration query you mean something like this? (taken from this pull request)

`ALTER TABLE account_ban_history CONVERT TO CHARACTER SET utf8`

Or would it be better to keep old user databases as they was, and only force new installations to use new utf8? What do you think? Or are there any cases that I didn't take into consideration?

PR #2407, opened 2018-01-02
Wrap System introduced in 10.92

### LINKED ISSUE #2017: constructionkits not work

nekiro:
Should close #2017 and #2031 

Wrapping does not work like rotating the item, cipsoft uses it like "attribute" one item wrap into many. 
To achieve that, we are using custom attribute system which is already implemented in sources.

I'm also crediting the first pull request by @Zbizu. I used some of his code as a base.

Example use of this system:
```
local item = player:addItem(26054, 1)
if item then
	item:setCustomAttribute("unWrapId", 25393)
	item:setCustomAttribute("wrapId", 26054) or item:getId()
end

### CODE REVIEW

File: data/events/scripts/player.lua [hunk 1]
```diff
@@ -273,3 +273,27 @@ function Player:onGainSkillTries(skill, tries)
 	end
 	return tries * configManager.getNumber(configKeys.RATE_SKILL)
 end
+
+function Player:onWrapItem(item)
+	local position = item:getPosition()
+	local tile = Tile(position)
```
WibbenZ: local tile = Tile(item:getPosition())

src/protocolgame.cpp
```diff
@@ -926,6 +927,14 @@ void ProtocolGame::parseHouseWindow(NetworkMessage& msg)
 	addGameTask(&Game::playerUpdateHouseWindow, player->getID(), doorId, id, text);
 }
 
+void ProtocolGame::parseWrapItem(NetworkMessage& msg)
+{
+	Position pos = msg.getPosition();
+	uint16_t spriteId = msg.get<uint16_t>();
+	uint8_t stackpos = msg.getByte();
```
WibbenZ: stackPos

src/protocolgame.cpp
```diff
@@ -926,6 +927,14 @@ void ProtocolGame::parseHouseWindow(NetworkMessage& msg)
 	addGameTask(&Game::playerUpdateHouseWindow, player->getID(), doorId, id, text);
 }
 
+void ProtocolGame::parseWrapItem(NetworkMessage& msg)
+{
+	Position pos = msg.getPosition();
```
WibbenZ: Make it a reference

data/events/scripts/player.lua
```diff
@@ -273,3 +273,26 @@ function Player:onGainSkillTries(skill, tries)
 	end
 	return tries * configManager.getNumber(configKeys.RATE_SKILL)
 end
+
+function Player:onWrapItem(item)
+	local tile = Tile(item:getPosition())
+	if tile and not tile:getHouse() then
```
TheSumm: This should be `not tile or not tile:getHouse()`

data/events/scripts/player.lua
```diff
@@ -273,3 +273,26 @@ function Player:onGainSkillTries(skill, tries)
...
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "You can only wrap and unwrap items inside a house.")
```
TheSumm: The official message is "You can only wrap and unwrap this item inside a house."

data/events/scripts/player.lua
TheSumm: Currently, if the tile does not exist (for some reason) it will skip this block and might execute `item:moveTo(tile)`.

TheSumm:
You could include a function to create a virtual item which also sets the attributes e.g. `Game.createFurnitureKit(furnitureKitId, unwrapId)`.

@conde2 "decoration kits", which is the item used whenever an item is wrapped, cannot be sold in the market. However, normal "furniture kits" (which are currently scripted in actions) would need to be hardcoded with a table (without using attributes).

src/events.cpp
```diff
@@ -787,3 +789,30 @@ void Events::eventPlayerOnGainSkillTries(Player* player, skills_t skill, uint64_
 
 	scriptInterface.resetScriptEnv();
 }
+
+bool Events::eventPlayerOnWrapItem(Player* player, Item* item)
```
ninjalulz: You may as well use `void` here

conde2:
Sorry, didn't know it couldn't be sold on market.

data/events/scripts/player.lua
```diff
@@ -273,3 +273,26 @@ function Player:onGainSkillTries(skill, tries)
 	end
 	return tries * configManager.getNumber(configKeys.RATE_SKILL)
 end
+
+function Player:onWrapItem(item)
+	local tile = Tile(item:getPosition())
+	if not tile or not tile:getHouse() then
+		self:sendTextMessage(MESSAGE_STATUS_SMALL, "You can only wrap and unwrap this item inside a house.")
```
ninjalulz: Use `sendCancelMessage` for consistency 😄

src/protocolgame.cpp
```diff
@@ -926,6 +927,14 @@ void ProtocolGame::parseHouseWindow(NetworkMessage& msg)
 	addGameTask(&Game::playerUpdateHouseWindow, player->getID(), doorId, id, text);
 }
 
+void ProtocolGame::parseWrapItem(NetworkMessage& msg)
+{
+	const Position& position = msg.getPosition();
+	uint16_t spriteId = msg.get<uint16_t>();
+	uint8_t stackPos = msg.getByte();
+	addGameTaskTimed(DISPATCHER_TASK_EXPIRATION, &Game::playerWrapItem, player->getID(), position, stackPos, spriteId);
```
djarek: Any particular reason why this task is timed?

nekiro: Nope, this code is copied from the old wrap furniture pr, also every other "useItem" etc is timed, so thats why probably.

EPuncker:
why was this closed? I've just tried it and it is perfectly working:
0636fbf0-8005-4128-b466-9f6997631ed8

nekiro:
@EPuncker from time to time I get very unmotivated, when almost none of my prs are getting merged. Reopening.

EPuncker:
I'm using this for almost a month and everything is working great so far!

EPuncker:
@nekiro there are two small things missing that I've noticed just now:
- only house owner should be able to wrap/unwrap wrapable items
- the wrap/unwrap items shouldn't be able to be moved outside the house (tiles)

Znote:
Conflicting files
data/events/events.xml
src/events.cpp
src/events.h

nekiro:
I broke that pr, will open new. Closing.

PR #2408, opened 2018-01-02
Adding creature:setHealth() method

nekiro:
I think this method is really useful when you want to set creature health without showing it (silently).

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -7087,7 +7087,7 @@ int LuaScriptInterface::luaCreatureSetHealth(lua_State* L)
 		return 1;
 	}
 
-	creature->health = std::max<int32_t>(0, getNumber<uint32_t>(L, 2));
+	creature->health = std::min<int32_t>(getNumber<uint32_t>(L, 2), creature->healthMax);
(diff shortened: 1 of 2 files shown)


src/luascript.cpp
```diff
@@ -7077,6 +7078,26 @@ int LuaScriptInterface::luaCreatureGetHealth(lua_State* L)
...
+int LuaScriptInterface::luaCreatureSetHealth(lua_State* L)
+{
+	// creature:setHealth(health)
+	Creature* creature = getUserdata<Creature>(L, 1);
+	if (!creature) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	creature->health = std::max<int32_t>(0, getNumber<uint32_t>(L, 2));
```
ninjalulz: Shouldn’t it update `healthMax` as well in case `health` exceeds `healthMax`?

nekiro: Actually it should be `creature->health = std::min<int32_t>(creature->health, getNumber<uint32_t>(L, 2));`

Znote:
Please update: https://github.com/otland/forgottenserver/wiki/Metatable:Creature

PR #2411, opened 2018-01-04
Remove promoted vocations from weapons.xml

bpawel10:


nekiro:
What is fixing this "fix"?

bpawel10:
Sorcerers and druids can still use their wands and rods once they buy a promotion.

nekiro:
They can use it right now too.

bpawel10:
So why weapons for knights and paladins have two vocations?

nekiro:
Probably, because of this code:
https://github.com/otland/forgottenserver/blob/master/src/weapons.cpp#L205

bpawel10:
Okay, so let's remove elite knights and royal paladins instead to unify it.

DSpeichert:
@bpawel10 You can change this PR to remove the promoted vocations and we'll merge it.

Kamenuvol:
I think they are there because showInDescription tag. They should not appear and the default is to appear.

DSpeichert:
@Kamenuvol will they still appear even if the vocation is not listed at all?

PR #2412, opened 2018-01-04
Adding player:getZone()

nekiro:
Currently to get current player zone, we need to get the position and then tile player is standing on. I think we should have a function for this, which would shorten scripts and simplify getting current player zone in lua.


Znote:
Update: https://github.com/otland/forgottenserver/wiki/Metatable:Creature

PR #2415, opened 2018-01-06
 Handle >7 players on 1 tile

DSpeichert:
Fixes #1950


src/protocolgame.cpp
```diff
@@ -517,19 +517,31 @@ void ProtocolGame::GetTileDescription(const Tile* tile, NetworkMessage& msg)
 		for (auto it = items->getBeginTopItem(), end = items->getEndTopItem(); it != end; ++it) {
 			msg.addItem(*it);
 
-			if (++count == 10) {
+			count++;
```
ranisalt: I cry in bed when I see `count++` instead of `++count`

DSpeichert: Why do you cry?

PR #2416, opened 2018-01-06
Ship default config.lua as config.lua.dist

DSpeichert:
If config.lua.dist exists and config.lua does not,
we will copy config.lua.dist to config.lua on lauch.


ranisalt:
No need to close files because `fstream` implements RAII.

Dagst:
What is the advantage of having the config.lua.dist being that it will create the config.lua? It would not be easier to leave the config.lua alone. I'm not criticizing, I just want to understand.

Kamenuvol:
@Dagst Now, it ignores your personal config when commiting.

Znote:
@Dagst No more git conflict when pulling in recent changes because you didnt reset config.lua.

PR #2422, opened 2018-01-08
Allow players to cast spells while being muted

nekiro:
Players should be able to cast spells even when they are muted from talking.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3297,15 +3297,15 @@ void Game::playerSay(uint32_t playerId, uint16_t channelId, SpeakClasses type,
 
 	player->resetIdleTime();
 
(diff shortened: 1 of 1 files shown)


raymondtfr:
Isn't this a workaround? I mean, isn't there somewhere in our code base a place where spell words are still counting as mute ticks?

Correct me if I'm wrong, I didn't check this right now, its just some déjà-vu whn I tried to do this same task, but never finished it.

nekiro:
@raymondtfr 
You cant spam spells to the point when you get muted, because you got exhaust. Workaround? Hmm. That's how the code should look from the start IMO.

raymondtfr:
I must be very mislead about this never mind it xD.

PR #2424, opened 2018-01-08
Attempt to fix Player.getInstantSpells

ninjalulz:


ranisalt:
How does this change anything relevant to the issue?

WibbenZ:
@ranisalt For some reason this fixed the issue.
Not sure why, tested without the changes (spellbook broke) and with Ninjas change it started working again.

marksamman:
My guess is that the pointer that was being pushed into the "spells"-vector was pointing to the second value in the iterator and not the actual InstantSpell object, thus being overwritten every iteration and ultimately all the inserted InstantSpell pointers were pointing to the last iterated InstantSpell object.

ranisalt:
That is hilarious

PR #2428, opened 2018-01-12
Fix Player.getInstantSpells

Mkalo:
There is no need to copy instant spells there.


ranisalt:
That's exactly reverting the previous related PR which claims to fix Player.getInstantSpells.

nekiro:
@ranisalt no, its not. It has small changes.

Issue #2206, opened 2017-03-20
Design a flatbuffers based map format


djarek:
# Design goals
- achieve faster map loading time
- keep the map size within the order of magnitude of OTBM
- be compatible with #2139

# Other requirements
- provide conversion tools
- implement support for the new format in map editors ?

ranisalt:
Can you provide an OTBM format specification for reference? I think what is currently specified in the source code is very lacking and too superficial about what can be contained in an OTBM file.

djarek:
This is a very generic spec. Too bad the original author didn't write an RFC :D.
```
Escape = 0xFD
Start = 0xFE
End = 0xFF
NodeTag = Escape | Start | End
OTBM_ROOTV1 = 0x01
NodeType = OTBM_ROOTV1 | 0x02 | ... | 0x10
EscapedByte = (0xFD 0xFD) | (0xFD 0xFE) | (0xFD 0xFF)
UnescapedByte = 0x00 | 0x01 | ... | 0xFC
PropertyStream = {EscapedByte | UnescapedByte}
Node = Start NodeType [PropertyStream] {Node} End
Identifier = "OTBM" | "OTBI"
RootHeaderVersion = u32le
RootHeaderWidth = u16le
RootHeaderHeight = u16le
RootHeaderMajorItemsVersion = u32le
RootHeaderMinorItemsVersion = u32le
RootHeader = RootHeaderVersion RootHeaderWidth RootHeaderHeight RootHeaderMajorItemsVersion RootHeaderMinorItemsVersion
RootNode = Start OTBM_ROOTV1 {Node} End
OTBFile = Identifier RootNode 
```

conde2:
Wouldn't worth rework OTB aswell when doing this changes ? It could totally be done in XML/LUA or anything more friendly than what we have now.

djarek:
@conde2 
If you mean the items file, I'm not sure why it's in binary format now, if there are no size issue, then it could be merged with items.xml. As for xml maps? Nope, not going to happen.

conde2:
Yes, I was talking about the items.otb, it would be awesome have it  merged with items.XML.
About the map I know it is not viable doing it in XML, i was strictly speaking about items.

ranisalt:
Expect items.xml to grow orders of magnitude in size since binary formats can store much more info on a single byte compared to text formats.

Kamenuvol:
[...]

PR #2429, opened 2018-01-13
[WIP] Update variables data types

raymondtfr:


src/item.cpp
```diff
@@ -751,7 +751,7 @@ void Item::serializeAttr(PropWriteStream& propWriteStream) const
 
 	if (hasAttribute(ITEM_ATTRIBUTE_DEFENSE)) {
 		propWriteStream.write<uint8_t>(ATTR_DEFENSE);
-		propWriteStream.write<int32_t>(getIntAttr(ITEM_ATTRIBUTE_DEFENSE));
+		propWriteStream.write<uint32_t>(getIntAttr(ITEM_ATTRIBUTE_DEFENSE));
```
ranisalt: This is backwards-incompatible, do take care. It is better to wait until #2206 for items is ready.

ranisalt:
Keep in mind it is not necessarily advantageous to have unsigned values for non-negative fields e.g. `Position`'s `x`, `y` and `z`. Sometimes you will want to subtract and handle overflows, which are not easily detectable with unsigned values.

src/item.cpp
raymondtfr: Oh okay... so should I put this PR on hold for now?

ranisalt: You might keep working on it for science, but I would call it waste of time *for now*.

DSpeichert:
While these values may not make sense to be negative, this looks like a change for the sake of change as there is no other explanation attached. Unless the change brings any benefits, it's not worth merging.

PR #2431, opened 2018-01-16
Accept GUID for Player constructor

Vulcanx:


src/luascript.cpp
```diff
@@ -7554,7 +7554,12 @@ int LuaScriptInterface::luaPlayerCreate(lua_State* L)
 	// Player(id or name or userdata)
```
conde2: You should also change to
// Player(id or guid or name or userdata)

PR #2444, opened 2018-02-02
Shoplist update upon proper situations

### LINKED ISSUE #2443: [Question] Why shoplist sending is mandatory in player's post functions if link is LINK_NEAR?
### Question Desciption
https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L2979

slavidodo:
shoplist shouldn't be updated within link_near or link_parent

Closes #2443

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2922,7 +2922,7 @@ void Player::postAddNotification(Thing* thing, const Cylinder* oldParent, int32_
 		g_moveEvents->onPlayerEquip(this, thing->getItem(), static_cast<slots_t>(index), false);
 	}
 
-	bool requireListUpdate = true;
+	bool requireListUpdate = false;
 
 	if (link == LINK_OWNER || link == LINK_TOPPARENT) {
 		const Item* i = (oldParent ? oldParent->getItem() : nullptr);
@@ -2976,7 +2976,7 @@ void Player::postRemoveNotification(Thing* thing, const Cylinder* newParent, int
 		g_moveEvents->onPlayerDeEquip(this, thing->getItem(), static_cast<slots_t>(index));
 	}
 
-	bool requireListUpdate = true;
+	bool requireListUpdate = false;
 
 	if (link == LINK_OWNER || link == LINK_TOPPARENT) {
 		const Item* i = (newParent ? newParent->getItem() : nullptr);


DSpeichert:
@slavidodo Can you explain the context of this change with examples?

slavidodo:
This function updates inventory shop items client-wise. By default, every item moved/added/removed inside or outside player's inventory causes an update. This should only happen when the item is inside the player's container.

DSpeichert:
@slavidodo Maybe I am forgetting something but wouldn't this be true only when the option to include items from inventory is disabled? I don't remember exactly how it works now but I think there was a switch/checkbox to prevent the shop from letting you sell your e.g. armor, which could be disabled. In that case, don't we need link_near?

slavidodo:
But definitely not the item from the ground! I may be wrong tho.

DSpeichert:
Okay, so for this to be properly considered, you'll need to explain your thought process a bit more. Not because I think you're wrong but because I don't know enough about it.

In which case do you think it the shoplist is being updated for items on the ground being moved?
Also, why is it important to fix this? Is there a performance issue related to this? That would just help prioritize.

jo3bingham:
@DSpeichert I'll try to help explain what @slavidodo is trying to fix.

In `Tile::postAddNotification`, `Player::postAddNotification` is triggered for each spectating player: https://github.com/otland/forgottenserver/blob/020ed5fd36518bbd41894cbdd4478bfa5993260b/src/tile.cpp#L1343-L1349
Same thing in `Tile::postRemoveNotification` for `Player::postRemoveNotification`.

In `Player::postRemoveNotification`, `requireListUpdate` defaults to `true` and only changes if link is that of `LINK_OWNER` or `LINK_TOPPARENT`. However, tile notifications always passes `LINK_NEAR` so `requireListUpdate` will always be true. If the player is currently trading with an NPC, and `requireListUpdate` is true, then `updateSaleShopList` will be invoked:
https://github.com/otland/forgottenserver/blob/020ed5fd36518bbd41894cbdd4478bfa5993260b/src/player.cpp#L2950-L2952

That means, if you're trading with an NPC and another player starts dropping items on a tile near you (or picking them up) then the server is going to send an updated shop list packet to your client even though your inventory has not changed. There's no ill side-effects other than it being completely unnecessary and a waste of IO space.

PR #2449, opened 2018-02-22
Fix #2440 UID Reset

### LINKED ISSUE #2440: item:getUniqueID() returns multiple UIDs for the same item.
- [x ] This is a bug in the software that resides in this repository, and not a
- [x ] This issue is reproducible without changes to the code in this repository

### Description
Calling item:getUniqueID() on an item with a UID of 0 will always return a new UID.
It should set the items uniqueID to the new ID
This is a problem because calling
item:getUniqueID()
item:getUniqueID()
item:getUniqueID()
Will return a different UID every time for the same item.

1. In a lua script call item:getUniqueID() on an item with a UID of 0 (99% of map)
2. In a lua script call item:getUniqueID() on the same item

### Expected behaviour
First Call
The item has a UID of 0
The next available UID is 65542
The item now has a UID of 65542
The function returns 65542
Second Call
The item has a UID of 65542
The function returns 65542

### Actual behaviour
First Call
The item has a UID of 0
The next available UID is 65542
The UID is added to the UID map and not used again
The item does not get the UID 65542 (stays 0)
The function returns 65542
Second Call
The item has a UID of 0
The next available UID is 65543
The UID is added to the UID map and not used again
The item does not get the UID 65543 (stays 0)
The function returns 65543

### Environment
TFS 1.2
Windows 10
Visual Studio 2015

comedinha:
Fix error on localMap reset everytimes and is impossible get the actual uniqueId of created items.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -66,7 +66,6 @@ void ScriptEnvironment::resetEnv()
 	callbackId = 0;
 	timerEvent = false;
 	interface = nullptr;
-	localMap.clear();
 	tempResults.clear();
 
 	auto pair = tempItems.equal_range(this);


marksamman:
Wouldn't this result in dangling pointers and a memory leak?

comedinha:
I think it would be a problem if there is a clean, because always clean and the next time will increase the counter
https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L128

It would be faster if memory failure occurs because the counter get the max value.

marksamman:
Sorry, I don't understand what you're trying to say. If you have a use case where the counter is overflowing, consider submitting a pull request where you change it from u32 to u64, but before you do that, please explain what kind of script is causing it to overflow.

CristoferMartins:
That is something that bothers me a lot. Items don't have a reliable item id like creatures. If we want to use the item in the future(addevent/storage), we can't use the Item() constructor since the uid is not reliable or simply useless.

PR #2452, opened 2018-03-04
Update copyright for 2018

vankk:
It is time to we move on to 2018.


slawkens:
I just found something useful regarding copyright year on the web.

https://softwareengineering.stackexchange.com/questions/130478/what-copyright-date-for-an-update-to-an-open-source-project-from-last-year

and

http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/

tl;dr
It states that 'copyright' year in source files is meant to be the year of publication, **not** an actual year.
Correct me if I'm wrong.

Issue #2446, opened 2018-02-08
Eternal character list means eternal premium time


TwistedScorpio:
Premium time should be updated even when a player logs in.
On my server I've had people get free premium time by never closing the character list, hence not calling Game::updatePremium,

okk3:
Hello,

I totally get your point but I still think the premium should be updated on server start.

You can remove premium account days via login, but that's just a feature that doesn't seem that important to have on the master repository.

Kindest Regards,
okke

TwistedScorpio:
Updating premium time upon server boot means that if server boots and a player has 2 hours left of premium, how is the server then going to know the premium time expired? It has to be on upon player and account login.

Kamenuvol:
every login it should add event with time left duration

soul4soul:
On Cipbia you can stay logged in on one character until server save if your premmy expires during the day. Once your premmy expires relogging to the same or another character on the account will cause you to loose premmy even if you keep the character list up. To me that means they check premmy time on login.

firdaus3318:
Apa fungsi serfer? Tolong jelaskan kepada saya

Issue #2436, opened 2018-01-20
[Suggestion] premend instead of premdays and lastday


slawkens:
I suggest removing _premdays_ and _lastday_ columns from the table accounts and replace them with _premend_. Why? Its easier to maintain, understand and doesn't require ANY database updates when servers starts or when account is loaded. Its then easier to calculate from database when Premium Account ends, without performing any arithmetical operations on the both values to get the current time.

Also, OpenTibia and many other distros based on it are already using premend, so its some kind of standard.

And finally: it uses less memory, as its only one column instead of two 😃 

What do you guys think?

ranisalt:
If I had a server, I would store premium time on a separate table to track all premium time spells.

That said, I agree that it's much better. `premdays` and `lastday` come from a time it was much harder to handle time on C++, but since `<chrono>` points and durations it is not an issue anymore.

pzaj2:
How far do you think these changes should reach? Only `premiumDays` or perhaps all things involving time arithmetics?

PR #2453, opened 2018-03-05
Update premium time on login

nekiro:
Attempt to "fix"  #2446
I don't know if that's how it should be done, but that's my attempt. Let me know what you think.

### DIFF
--- a/src/protocollogin.cpp
+++ b/src/protocollogin.cpp
@@ -66,9 +66,6 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
 		output->addByte(0);
 	}
 
-	//Update premium days
-	Game::updatePremium(account);
-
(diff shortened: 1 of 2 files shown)


slawkens:
If #2436 will be done, then updating premium time on login won't be needed anymore.

Till then, it's a good solution (when it works - didn't test).

src/protocollogin.cpp
```diff
@@ -66,9 +66,6 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
 		output->addByte(0);
 	}
 
-	//Update premium days
-	Game::updatePremium(account);
```
DSpeichert: If you remove this, the first login during a day may show incorrect premium status because time is not recalculated before character list is sent.

Issue #1556, opened 2015-09-04
getPathTo is not working as expected


HeavenIsLost:
I'm trying to move a monster from his position to his master position(spawn position). It works if spawn position is closer(20~25 sqms), but if if its not inside this range getPathTo will just fail(return false).

test code:
http://hastebin.com/ebazadabox.avrasm

djarek:
I haven't delved too deep into the map code, so I only have a high-level understanting of it. Pathing is computed by using the A* algorithm (https://en.wikipedia.org/wiki/A*_search_algorithm). MAX_NODES is the maximal number of nodes that are taken into consideration, so doubling them increases the number of vertices in the graph that represents all possible paths. So, in the worst case the performance of the path search is going to drop by a factor of 4.

marksamman:
This is working as intended, the current algorithm is optimized for max 512 nodes. If you increase max nodes, the algorithm may not be optimal and you should consider rewriting it using a priority queue.

PR #2464, opened 2018-04-28
 Optimized Pathfinding

Flatlander57:
Used Distance to Optimize Pathfinding
More Info Here:
https://otland.net/threads/optimizing-tfs-pathfinding.257868/

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1573,7 +1573,7 @@ bool Creature::isInvisible() const
 
 bool Creature::getPathTo(const Position& targetPos, std::forward_list<Direction>& dirList, const FindPathParams& fpp) const
 {
-	return g_game.map.getPathMatching(*this, dirList, FrozenPathingConditionCall(targetPos), fpp);
+	return g_game.map.getPathMatching(*this, targetPos, dirList, FrozenPathingConditionCall(targetPos), fpp);
 }
 
 bool Creature::getPathTo(const Position& targetPos, std::forward_list<Direction>& dirList, int32_t minTargetDist, int32_t maxTargetDist, bool fullPathSearch /*= true*/, bool clearSight /*= true*/, int32_t maxSearchDist /*= 0*/) const
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -561,7 +561,7 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
 	return tile;
 }
 
-bool Map::getPathMatching(const Creature& creature, std::forward_list<Direction>& dirList, const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const
+bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::forward_list<Direction>& dirList, const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const
 {
 	Position pos = creature.getPosition();
 	Position endPos;
@@ -588,7 +588,7 @@ bool Map::getPathMatching(const Creature& creature, std::forward_list<Direction>
 
 	AStarNode* found = nullptr;
 	while (fpp.maxSearchDist != 0 || nodes.getClosedNodes() < 100) {
-		AStarNode* n = nodes.getBestNode();
+		AStarNode* n = nodes.getBestNode(targetPos);
 		if (!n) {
 			if (found) {
 				break;
@@ -771,17 +771,30 @@ AStarNode* AStarNodes::createOpenNode(AStarNode* parent, uint32_t x, uint32_t y,
 	return node;
 }
 
-AStarNode* AStarNodes::getBestNode()
+AStarNode* AStarNodes::getBestNode(const Position& targetPos)
 {
 	if (curNode == 0) {
 		return nullptr;
 	}
 
+	int32_t bestNodeF = 1000000;
+	uint32_t bestNode = 0;
+	int32_t diffNode = 0;
+
 	int32_t best_node_f = std::numeric_limits<int32_t>::max();
 	int32_t best_node = -1;
 	for (size_t i = 0; i < curNode; i++) {
-		if (openNodes[i] && nodes[i].f < best_node_f) {
-			best_node_f = nodes[i].f;
+		const int_fast32_t Sx = nodes[0]->x;
+		const int_fast32_t Sy = nodes[0]->y;
+		const int_fast32_t Cx = nodes[i]->x;
+		const int_fast32_t Cy = nodes[i]->y;
+		int32_t SdiffX = std::abs(targetPos.x - Sx);
+		int32_t SdiffY = std::abs(targetPos.y - Sy);
+		int32_t NdiffX = std::abs(targetPos.x - Cx);
+		int32_t NdiffY = std::abs(targetPos.y - Cy);
+		diffNode = ((nodes[i].f+(((NdiffX-SdiffX)*5)+((NdiffY-SdiffY)*5)))+(std::max(NdiffX, NdiffY)*5));
+		if (openNodes[i] && diffNode < best_node_f) {
+			best_node_f = diffNode;
 			best_node = i;
 		}
 	}
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -781,10 +781,10 @@ AStarNode* AStarNodes::getBestNode(const Position& targetPos)
 	int32_t best_node_f = std::numeric_limits<int32_t>::max();
 	int32_t best_node = -1;
 	for (size_t i = 0; i < curNode; i++) {
-		const int_fast32_t Sx = nodes[0]->x;
-		const int_fast32_t Sy = nodes[0]->y;
-		const int_fast32_t Cx = nodes[i]->x;
-		const int_fast32_t Cy = nodes[i]->y;
+		const int_fast32_t Sx = nodes[0].x;
+		const int_fast32_t Sy = nodes[0].y;
+		const int_fast32_t Cx = nodes[i].x;
+		const int_fast32_t Cy = nodes[i].y;
 		int32_t SdiffX = std::abs(targetPos.x - Sx);
 		int32_t SdiffY = std::abs(targetPos.y - Sy);
 		int32_t NdiffX = std::abs(targetPos.x - Cx);
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -777,10 +777,7 @@ AStarNode* AStarNodes::getBestNode(const Position& targetPos)
 		return nullptr;
 	}
 
-	int32_t bestNodeF = 1000000;
-	uint32_t bestNode = 0;
 	int32_t diffNode = 0;
-
 	int32_t best_node_f = std::numeric_limits<int32_t>::max();
 	int32_t best_node = -1;
 	for (size_t i = 0; i < curNode; i++) {
--- a/src/map.h
+++ b/src/map.h
@@ -56,7 +56,7 @@ class AStarNodes
 		AStarNodes(uint32_t x, uint32_t y);
 
 		AStarNode* createOpenNode(AStarNode* parent, uint32_t x, uint32_t y, int_fast32_t f);
-		AStarNode* getBestNode();
+		AStarNode* getBestNode(const Position& targetPos);
 		void closeNode(AStarNode* node);
 		void openNode(AStarNode* node);
 		int_fast32_t getClosedNodes() const;
@@ -251,8 +251,8 @@ class Map
 
 		const Tile* canWalkTo(const Creature& creature, const Position& pos) const;
 
-		bool getPathMatching(const Creature& creature, std::forward_list<Direction>& dirList,
-		                     const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const;
+		bool getPathMatching(const Creature& creature, const Position& targetPos, std::forward_list<Direction>& dirList,
+							const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const;
 
 		std::map<std::string, Position> waypoints;
 


DSpeichert:
Do you have any benchmarks before/after to quantify the improvement coming from this change?

Flatlander57:
I could benchmark it, but I can tell you it gets insanely better in most cases, and it's worse in 0 cases.

marksamman:
This is probably better in terms of performance for most cases since it adds a heuristic for prioritizing nodes closer to the destination (which is usually the best path). It sacrifices quality by potentially taking paths that cost more in other terms than distance though.

I believe the algorithm would be even faster by storing the additional distance cost in the "f"-variable so that it doesn't have to be re-calculated every time getBestNode() is called.

I'm not sure sacrificing quality for the performance gains here is worth it though, but it would be interesting to see some animations on how the best nodes are selected in some example cases compared to the current code, especially where the shortest path is not the best path.

Flatlander57:
I would never say my code is the perfect optimization of pathfinding in tibia.

But it is a fact that it is better than TFS current pathfinding in just about all situations possible.

marksamman:
I'm in favor of adding a distance heuristic to the algorithm. I'm just hoping we can find the optimal value for the distance multiplier where we can get as close results to the current algorithm as possible so that we gain the performance without sacrificing quality. Have you tested other values than 5?

Flatlander57:
Can you give me an example of something I should test, where you think normal TFS would perform better or choose a better path than mine?

Or are you stating that I should create additional code to make it even better than normal TFS in more ways?

CristoferMartins:
@Flatlander57 i just want to mention this https://github.com/otland/forgottenserver/issues/1556

"This is working as intended, the current algorithm is optimized for max 512 nodes. If you increase max nodes, the algorithm may not be optimal and you should consider rewriting it using a priority queue."

This feature would be useful if we want to make monsters walk a long way.

marksamman:
This pull request does not extend the max nodes AFAIK, so switching to a priority queue would most likely result in decreased performance.

I'm not sure if TFS will choose a better path, but it would never choose a worse path. In terms of performance, the current algoritm might perform better for very short distances since getBestNode does less work. I'm asking for anyone to test this in various cases to gather feedback, I assume you've already done so, so if nobody has any objections I suppose this can be merged (although I think it would be better if the math in getBestNode was precalculated so we get more performance at the cost of temporary slightly increased memory use). Personally I would prefer to see comparisons in path selection and performance with different values than 5 to see if 5 is the optimal value or if we can find something better (both in terms of path quality and performance).

Flatlander57:
The 5 value is simple, it means it will check longer-paths for 5 sqm before checking farther paths.

For example, if you are behind a wall, and your "target" is on the other side, it will check along the wall in both directions for 5 sqm, before it checks 1 sqm in the "wrong" direction.

If you change this to lower than 5 it will more often "waste" time checking wrong directions. 
On the other hand, since the screen-size of tibia is 11x15 (it is often the case that a door or corner will be within 5 sqm of you allowing you to get closer to your target).

Every sqm closer you get to your target, you get 5 more "free" sqm to check before checking a direction away from your target. So basically the more success the pathfinding has in closing the distance, the more nodes it gives that path before checking other paths.

If you go into an open area (no blocking walls or doodads) with the current version of TFS normal pathing, and try to "Use" the top corner of your screen, or "use" a shovel on the ground on the top corner of your screen, TFS will be unable to find the path to that top-corner. This is due to it checking all the nodes on the screen and reaching the limit (which is lower than 512 for the player-onUse pathing).

Though, you may be right that setting a value:
nodes[i].f = cost of walking to this node position
nodes[i].other = distance cost of this node

Then just add these 2 values together when doing the bestNode.

CristoferMartins:
@Flatlander57 does your algorithm handles long paths(like 50 sqms distant position) better than the current one? And how it will performe when MAX_NODES is 1024? Thanks in advance.

Flatlander57:
@CristoferMartins It will handle MAX_NODES = 1024 the same.

As for long paths (like 50 sqm) it is insanely better than the current one. To the point of, if you have 10 creatures following you at 50 sqm right now with Max_Nodes set to 99999999 or whatever, the server will just be a constant frozen laggy mess. If you add in the above, it would be playable.

I do not suggest increasing Max_Nodes. this feature was put-in to make it so your server does not lag, If you do increase it, just know you are increasing the chance that pathfinding will cause lag on your server.

CristoferMartins:
@Flatlander57 thank you, i will probably will make MAX_NODES = 1024 and i will add a special flag to FindPathParams to allow the use of 1024 nodes in the path finding.

marksamman:
Looks like we got some data here: https://otland.net/threads/optimizing-tfs-pathfinding.257868/#post-2502478. I would prefer if someone could run tests with our codebase though.

joseluis2g:
@marksamman 
What does this repo needs to start again with the client updates? Whats your opinion? I can really help with it (and hope that you guys also help).

marksamman:
@joseluis2g The requirement is that it is compatible with OTClient, if you update OTClient first, then you can update TFS to support the new client version. Please open a new issue if you'd like to discuss this matter since it is not related to this pull request in any way.

Flatlander57:
I have had a friend help me with doing our own testing.

Below you can see the results:
https://otland.net/threads/optimizing-tfs-pathfinding.257868/#post-2505691

marksamman:
Thanks! That looks good. Unless some other maintainer wants to go ahead and merge this already, I'll try to spend some time optimizing getBestNode by pre-calculating the new heuristic so that we can have the best of both worlds before merging it myself.

marksamman:
Here's my patch. It's much faster when there's no path to the destination, and also faster to discover a valid path, it might be too aggressive though, someone should test if it's affecting the quality of the paths.

```patch
diff --git a/src/creature.cpp b/src/creature.cpp
index 336141d3..f0744531 100644
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1573,7 +1573,7 @@ bool Creature::isInvisible() const
 
 bool Creature::getPathTo(const Position& targetPos, std::forward_list<Direction>& dirList, const FindPathParams& fpp) const
 {
-	return g_game.map.getPathMatching(*this, dirList, FrozenPathingConditionCall(targetPos), fpp);
+	return g_game.map.getPathMatching(*this, targetPos, dirList, FrozenPathingConditionCall(targetPos), fpp);
 }
 
 bool Creature::getPathTo(const Position& targetPos, std::forward_list<Direction>& dirList, int32_t minTargetDist, int32_t maxTargetDist, bool fullPathSearch /*= true*/, bool clearSight /*= true*/, int32_t maxSearchDist /*= 0*/) const
diff --git a/src/map.cpp b/src/map.cpp
index 660fd9c8..e4ff5e05 100644
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -561,7 +561,7 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
 	return tile;
 }
 
-bool Map::getPathMatching(const Creature& creature, std::forward_list<Direction>& dirList, const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const
+bool Map::getPathMatching(const Creature& creature, Position targetPos, std::forward_list<Direction>& dirList, const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const
 {
 	Position pos = creature.getPosition();
 	Position endPos;
@@ -682,7 +682,9 @@ bool Map::getPathMatching(const Creature& creature, std::forward_list<Direction>
 				nodes.openNode(neighborNode);
 			} else {
 				//Does not exist in the open/closed list, create a new node
-				neighborNode = nodes.createOpenNode(n, pos.x, pos.y, newf);
+				neighborNode = nodes.createOpenNode(n, pos.x, pos.y, newf,
+					((std::abs(targetPos.x - pos.x) + std::abs(targetPos.y - pos.y)) * 10)
+				);
 				if (!neighborNode) {
 					if (found) {
 						break;
@@ -736,6 +738,7 @@ bool Map::getPathMatching(const Creature& creature, std::forward_list<Direction>
 	return true;
 }
 
+
 // AStarNodes
 
 AStarNodes::AStarNodes(uint32_t x, uint32_t y)
@@ -753,7 +756,7 @@ AStarNodes::AStarNodes(uint32_t x, uint32_t y)
 	nodeTable[(x << 16) | y] = nodes;
 }
 
-AStarNode* AStarNodes::createOpenNode(AStarNode* parent, uint32_t x, uint32_t y, int_fast32_t f)
+AStarNode* AStarNodes::createOpenNode(AStarNode* parent, uint32_t x, uint32_t y, int_fast32_t f, int_fast32_t g)
 {
 	if (curNode >= MAX_NODES) {
 		return nullptr;
@@ -768,20 +771,22 @@ AStarNode* AStarNodes::createOpenNode(AStarNode* parent, uint32_t x, uint32_t y,
 	node->x = x;
 	node->y = y;
 	node->f = f;
+	node->g = g;
 	return node;
 }
 
 AStarNode* AStarNodes::getBestNode()
 {
-	if (curNode == 0) {
-		return nullptr;
-	}
-
 	int32_t best_node_f = std::numeric_limits<int32_t>::max();
 	int32_t best_node = -1;
 	for (size_t i = 0; i < curNode; i++) {
-		if (openNodes[i] && nodes[i].f < best_node_f) {
-			best_node_f = nodes[i].f;
+		if (!openNodes[i]) {
+			continue;
+		}
+
+		int32_t cost = nodes[i].f + nodes[i].g;
+		if (cost < best_node_f) {
+			best_node_f = cost;
 			best_node = i;
 		}
 	}
diff --git a/src/map.h b/src/map.h
index 9a1275c9..95e040bd 100644
--- a/src/map.h
+++ b/src/map.h
@@ -41,7 +41,7 @@ static constexpr int32_t MAP_MAX_LAYERS = 16;
 struct FindPathParams;
 struct AStarNode {
 	AStarNode* parent;
-	int_fast32_t f;
+	int_fast32_t f, g;
 	uint16_t x, y;
 };
 
@@ -55,7 +55,7 @@ class AStarNodes
 	public:
 		AStarNodes(uint32_t x, uint32_t y);
 
-		AStarNode* createOpenNode(AStarNode* parent, uint32_t x, uint32_t y, int_fast32_t f);
+		AStarNode* createOpenNode(AStarNode* parent, uint32_t x, uint32_t y, int_fast32_t f, int_fast32_t g);
 		AStarNode* getBestNode();
 		void closeNode(AStarNode* node);
 		void openNode(AStarNode* node);
@@ -251,9 +251,8 @@ class Map
 
 		const Tile* canWalkTo(const Creature& creature, const Position& pos) const;
 
-		bool getPathMatching(const Creature& creature, std::forward_list<Direction>& dirList,
+		bool getPathMatching(const Creature& creature, Position targetPos, std::forward_list<Direction>& dirList,
 		                     const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const;
-
 		std::map<std::string, Position> waypoints;
 
 		QTreeLeafNode* getQTNode(uint16_t x, uint16_t y) {
```

Flatlander57:
10 points per sqm is twice as aggressive as my use of 5, but it really shouldn't cause many issues.
I had a friend test it, and your code is more optimized and faster than mine.

You won't see any issues using * 10 with normal tibia screen size and monsters that can "see" 11 sqm. But over long distances (50-100sqm pathing) 5 seems to find the path faster more often when there are many blocking obstacles, dead ends, and "wrong" paths. (not always, everything is situational)

You want to have the lowest "cost" with the most effect. I think 5 is your best bet, but i'll need to run a LOT of benchmarks in a LOT of environments to actually show a meaningful difference.

marksamman:
It behaves quite similarly to your formula though in the way it picks best nodes, I couldn't make much sense of using distance from the start point that your formula uses, mine only looks at the manhattan distance to the destination. If you change it to multiply by 5, it would be slower at picking the best path compared to your formula. Your formula uses two different data points (some math that involves the distance from start point + the longer distance of the x or y-axis IIRC) that it multiplies by 5 and adds together, sometimes the sum of the additions result in a negative value.

Benchmarks/test cases that also measure quality of path would be great, but if we can just get rid of the getBestNode() performance overhead in this pull request it's good to go. I did some benchmarks to measure performance and mine performs almost as good as the current algorithm when it can't find a path, and usually as good (and sometimes significantly faster) than your algorithm when it finds a path. I didn't measure quality or test many edge cases though. It would be best to test paths where there are tiles that result in additional cost (see getTileWalkCost or w/e the name is) that should be avoided, e.g. a monster shouldn't destroy another monster if there's a short enough path around the monster.

CristoferMartins:
pathfind
With @marksamman code i can use the item the image(broken bottle) and then server will find the path while with the normal tfs pathing code or ciptibia path the msg "there is no way" will appear.

DSpeichert:
@Flatlander57 do you intend to incorporate Mark's changes? I am in favor of incremental improvements, so the work of teaching monsters to walk around each other instead of killing themselves could go in a separate PR.

Flatlander57:
@DSpeichert I do plan to implement Mark's changes in a way.

I am using a custom version of TFS I started working on 7 years ago. So it's a 0.X version of TFS that is heavily edited.

I know TFS 1.X is way better in almost every way, but I just can't currently justify putting the time in to put all my edits into the new TFS. I simply wanted to pass on my optimization to the new TFS, it's the one edit I had made that actually helps normal tibia servers. All my other Edits change the core game mechanics of Tibia (new lighting system, new creature AI, aggro system, etc) 

As for monsters killing each other rather than walking around them.., from what I can tell, if you set a monster to canPushCreatures=1 then they will always ignore creatures completely (and push them if they are on a tile they walk on).

Our creatures do not currently "prefer" to walk around, they will always plow through all other creatures they are capable of moving through.

CristoferMartins:
"this merge request might introduce cases where monsters will prefer killing other monsters instead of walking around them." @marksamman I believe the destroy/push should be a const(or a config.lua parameter) where the server should should settle this behavior for his own taste.

Flatlander57:
If you are going to make a config for it, make it a monster config, where you can decide for each monster what they prefer.
on the flag canPushCreatures = "0 or 1"
We could change this to:
canPushCreatures = "0-100"
The higher the number the more situations where the creature will push/kill another creature.

DSpeichert:
@Flatlander57 any updates here?

Flatlander57:
@DSpeichert I've been using Mark's version of the code (with some edits) forever now.

It works beautifully, though I did change it from

`((std::abs(targetPos.x - pos.x) + std::abs(targetPos.y - pos.y)) * 10)`
to
`((std::abs(targetPos.x - pos.x) + std::abs(targetPos.y - pos.y)) * 15)`

DSpeichert:
Thanks for the feedback. Could you update our PR with that code and summarize the changes it makes?

I think it's OK if the behavior changes a bit as long as we have a clear description of how it changes.

Flatlander57:
Honestly I don't have the time to update the code.
The code in my current server also has edits to allow monsters to path up/down stairs. So the pathfinding is much differing than default TFS.

Just use Mark's Code, change the 10 to a 15. And the only changes will be monsters may or may not push each other out of the way as often, or walk over fire-fields as often. (Most likely you'll see no noticeable change in monster AI)

Good luck!

PR #2466, opened 2018-05-01
Handle token time parsing exceptions

mic-kul:
The current implementation of token time parsing can lead to a server crash using crafted packet.

`std::stoul` throws 2 types of exceptions: 
- one on an empty string and/or string not containing any numbers
- another if the parsed number would overflow the `long` type
http://en.cppreference.com/w/cpp/string/basic_string/stoul

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -281,7 +281,16 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
 	std::string& accountName = sessionArgs[0];
 	std::string& password = sessionArgs[1];
 	std::string& token = sessionArgs[2];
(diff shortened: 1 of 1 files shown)


ranisalt:
I'd rather use `std::strtoul` and avoid exceptions altogether.

mic-kul:
@ranisalt Actually I was considering using it, but found this... https://github.com/otland/forgottenserver/commit/aed6af694aa4bd4e61c83a3d2391b5761f5d7944
You can simply revert that change

ranisalt:
Oh, you are right, I didn't remember the signature for `strtoull` used C-style strings. Considering that, LGTM

PR #2214, opened 2017-03-25
Guildhalls and guild bank.

Znote:
WIP #2213 Click on issue link for further description.

**Edit: 13. Jul 2020:** Feature implementation complete! 
**Edit: 15. Nov 2020:** Rebase with master to continue testing

#### Schema changes
* houses.type to determine if it is a house or guildhall. 
* guilds.balance to determine the guild balance. 
* guild_transactions table to log transactions between guildmember and the guild, as well as between guilds.
* Schema version update to ~25~ 27

#### data folder changes
* Lua migrations added to reflect schema changes.
* Changed 4 houses to be flagged as guildhalls.
image

#### Engine/C++/Lua changes
* ```house:getType()``` return house type enums. 1 = HOUSE_TYPE_NORMAL, 2 = HOUSE_TYPE_GUILDHALL 
* Update database `houses` table with house type
* Modify house owner to no longer be Player guid implicit, but either player.guid or guild.id depending on house type.
* Create bank NPC to support (unrestricted) deposit to and (restricted) withdraws from guild banks. And (restricted) transfers between guilds.
* Make bank NPC use `guild_transactions` and give player receipts for their deposits and withdrawals.

#### WIP Known issues / todo:
* Testing rent paying. (might have broken rent/payhouses)?
* Looking at house doors seems to have inconsistent behavior (Does not always display house data) **Edit:** Does not seem to be related to this PR, getting same issue with otland/master
* Testing selling guildhalls with !sellhouse
* Todo: ~~Transfer funds between guilds~~ Fixed 29.06.2020
* Todo: ~~guild_transaction table was a bit too ambigious, want to rename the guild columns~~ Fixed 29.06.2020
* Todo: ~~!buyhouse: account for guild:getBankBalance() and only for guild leaders~~ Fixed 12.07.2020
* Todo: ~~sell house account for guildhalls~~ Fixed 13.07.2020

> as of 10.7 update we have Receipts:
> Receipts are pieces of paper which confirm whether or not guild-related transactions were successful, and also provide details about said transactions. There are 2 types of receipts:
> 
> Receipt (Fail) - a receipt of a failed transaction
> Receipt (Success) - a receipt of a successful transaction
> 
> [[source](https://tibia.fandom.com/wiki/Receipt)]

#### Postoned for later PR/skipped:
[...]

PR #2467, opened 2018-05-15
Guild halls support

nekiro:
In this pull request I'm trying to add missing guild halls support.

What does this pr add?
- now every guild can have it's own residence, which should be guild hall, but isn't required, it's up you to check that, keep in mind that you cannot set it from lua script, this should be done by AAC (because AAC creates new guild entry in database)
- you can now call house:isGuildHall() to check whether it is guild hall or not
- loading houses from houses.xml, will now parse missing "guildhall" attribute
- houses now have one additional attribute called guildId, this attribute stores guild that owns this house, this should be used only with guild halls
- made tfs load ownerid from database to determine who is the guild leader, this will fail if we have more than one leader in a guild, but currently the only other option is to check guildrank if its number 3, because 3 is leader.

Probably forgot about something as always, so let me know if anything is missing, btw. I didn't test this code.

New enums:
HOUSE_TYPE_NORMAL
HOUSE_TYPE_GUILDHALL

New lua methods:
- guild:setHouseId() -- sets guild hall id
- guild:getHouseId() -- gets guild hall id
- house:getType() -- returns house type
- house:setGuildId() -- sets guild id
- house:getGuildId() -- gets guild id


vankk:
I think that you missed the in-game part, should has a condition in _House::setOwner_ that if the house is a guildHall and who is purchasing is a guild leader then should set was a guild house to the Guild. It may be good for thoses servers how uses !buyhouse command.

malucooo:
need replace line 275 on iomapserialize.cpp
DBResult_ptr result = db.storeQuery("SELECT `id`, `owner`, `paid`, `warnings` FROM `houses`");
for this:
DBResult_ptr result = db.storeQuery("SELECT `id`, `owner`, `paid`, `warnings`, `guildid` FROM `houses`");

thanks for advance!

schema.sql
```diff
@@ -207,6 +208,7 @@ CREATE TABLE IF NOT EXISTS `houses` (
   `highest_bidder` int(11) NOT NULL DEFAULT '0',
   `size` int(11) NOT NULL DEFAULT '0',
   `beds` int(11) NOT NULL DEFAULT '0',
+  `guildid` int(11) NOT NULL,
```
DSpeichert: should be guild_id

schema.sql
```diff
@@ -133,6 +133,7 @@ CREATE TABLE IF NOT EXISTS `guilds` (
   `ownerid` int(11) NOT NULL,
   `creationdata` int(11) NOT NULL,
   `motd` varchar(255) NOT NULL DEFAULT '',
+  `residence` int(11) NOT NULL,
```
DSpeichert: should be house_id

src/guild.h
```diff
@@ -69,6 +69,12 @@ class Guild
 		void setMotd(const std::string& motd) {
 			this->motd = motd;
 		}
+		uint32_t getResidenceId() const {
```
DSpeichert: Instead of "residence" use "house", it's an actual entity in DB and really it's a house.

nekiro:
@vankk there was a reason why I didnt touch that. Current tfs code is only loading guilds from database (these guilds are created by AAC), we are not saving the guild from tfs in any scenario. So when we set the house id attribute to guild in tfs source, it won't get saved. I can write additional function that would be called after setting house id to guild, but I don't know if this is right approach.

schema.sql
```diff
@@ -133,6 +133,7 @@ CREATE TABLE IF NOT EXISTS `guilds` (
   `ownerid` int(11) NOT NULL,
   `creationdata` int(11) NOT NULL,
   `motd` varchar(255) NOT NULL DEFAULT '',
+  `house_id` int(11) NOT NULL,
```
DSpeichert: You can't update schema.sql
It must be a migration.

schema.sql
```diff
@@ -207,6 +207,7 @@ CREATE TABLE IF NOT EXISTS `houses` (
   `highest_bidder` int(11) NOT NULL DEFAULT '0',
   `size` int(11) NOT NULL DEFAULT '0',
   `beds` int(11) NOT NULL DEFAULT '0',
+  `guild_id` int(11) NOT NULL,
```
DSpeichert: There's also this one here that should be a migration instead.

DSpeichert:
@Znote can you take a look at this vs #2214 ?

WibbenZ:
LGTM

Znote:
Sorry, this got hidden in my backlog. Gonna take a proper look at it tonight/tomorrow.

Znote:
This vs #2214 
* This does not help an AAC understand what is and what isn't a guildhall. How do I differentiate between a normal house and a guildhall in the houses table (if they both are available)? This only adds an uneccesary reference? 
* Shouldn't it be possible to distinguish an available house from an available guildhall in-game?
* This adds 4 bytes overhead to each houses rows + 4 bytes overhead to all guild rows. Mine only add 1 byte overhead to houses rows. (+8 to guilds with guild_bank functionality)
* This only supports guild halls. Mine supports up to 128 different house group types, where 0 would be considered a house, 1 would be considered guildhalls, and 2+ for custom types. (etc 10+townid could be town hall rooms)? You can be more creative in Lua. 
* This does not support guild bank, mine does with an additional table and 8 byte bigint column in guilds row

@nekiro Have you looked at my request, do you think you could give the c++ code a go there? 
if you could use house type instead, wrap regular house functions to use type 0, guildhalls to use type 1, and create a generic function which checks for configured type?

How do I get the guildid <-> houseid reference without the reference? I go through the player id. (owner) 

**Sample: 1:**
All guildhalls, if guild_id is null then guildhall is available. (dont have a guild owner)
```sql
SELECT 
	`h`.`id`,
	`h`.`owner`,
	`g`.`id` AS `guild_id`
FROM `houses` AS `h` 
LEFT JOIN `guilds` AS `g` 
	ON `h`.`owner` = `g`.`ownerid`
	AND `h`.`type` = 1 
WHERE `h`.`type` = 1 
```

**Sample 2:**
All guilds, if house_id is null then guild does not have a guildhall
```sql
SELECT 
	`g`.`id`,
	`g`.`ownerid`,
	`h`.`id` AS `house_id`
FROM `guilds` AS `g`
LEFT JOIN `houses` AS `h`
	ON `g`.`ownerid` = `h`.`owner` 
	AND `h`.`type` = 1
```

data/migrations/23.lua
```diff
@@ -1,3 +1,6 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 24 (Guildhall support)")
+	db.query("ALTER TABLE `guilds` ADD `house_id` int(11) NOT NULL")
+	db.query("ALTER TABLE `houses` ADD `guild_id` int(11) NOT NULL")
+	return true
```
Znote:
```suggestion
	db.query("ALTER TABLE `houses` ADD `type` tinyint(3) NOT NULL DEFAULT '0'")
	return true
```

PR #2469, opened 2018-06-05
prevent crafted packets from inviting yourself to a party

Sickness-chan:
could potentially cause issues


DSpeichert:
"Potentially"?
Are there any tested issues this is preventing?

Sickness-chan:
on older tfs distributions like 0.3.6 and 0.4 this lead to a crash, tfs 1.3 does not seem to be affected as far as the crash goes, but you can still invite yourself and join your own party iirc, which should not be allowed

PR #2473, opened 2018-06-30
Handle SIGBREAK (Windows console close)

### LINKED ISSUE #1853: SetConsoleCtrlHandler
https://github.com/otland/forgottenserver/blob/master/src/otserv.cpp#L98
1. Log off or shut down
### Expected behavior

TFS should save and shutdown as if you had closed the console window.
### Actual behavior

TFS shuts down without saving.
From my tests it seems as though the function does not fire when a user logs off or does a shutdown, only when closing the console.
### Environment

Windows

DSpeichert:
SetConsoleCtrlHandler is not needed because SIGBREAK is also sent
on window close. The process is killed when the thread handling
SIGBREAK returns. That's why we must handle it synchronously
and wait for other threads to finish first. The Windows-defined
timeout is only 5 seconds but there is no way to change that.

https://docs.microsoft.com/en-us/windows/console/ctrl-c-and-ctrl-break-signals

Fixes #1277
Fixes #1853

### DIFF
--- a/src/signals.cpp
+++ b/src/signals.cpp
@@ -68,8 +68,9 @@ Signals::Signals(boost::asio::io_service& service) :
 	set.add(SIGUSR1);
 	set.add(SIGHUP);
 #else
-	// this must be a blocking call as Windows calls it in a new thread and terminates
-	// the process when the handler returns (or after 5 seconds, whichever is earlier)
+	// This must be a blocking call as Windows calls it in a new thread and terminates
+	// the process when the handler returns (or after 5 seconds, whichever is earlier).
+	// On Windows it is called in a new thread.
 	signal(SIGBREAK, dispatchSignalHandler);
 #endif
 
@@ -88,6 +89,9 @@ void Signals::asyncWait()
 	});
 }
 
+// On Windows this function does not need to be signal-safe,
+// as it is called in a new thread.
+// https://github.com/otland/forgottenserver/pull/2473
 void Signals::dispatchSignalHandler(int signal)
 {
 	switch(signal) {
@@ -107,6 +111,7 @@ void Signals::dispatchSignalHandler(int signal)
 #else
 		case SIGBREAK: //Shuts the server down
 			g_dispatcher.addTask(createTask(sigbreakHandler));
+			// hold the thread until other threads end
 			g_scheduler.join();
 			g_databaseTasks.join();
 			g_dispatcher.join();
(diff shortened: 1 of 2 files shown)


src/signals.cpp
```diff
@@ -63,6 +67,10 @@ Signals::Signals(boost::asio::io_service& service) :
 #ifndef _WIN32
 	set.add(SIGUSR1);
 	set.add(SIGHUP);
+#else
+	// this must be a blocking call as Windows calls it in a new thread and terminates
+	// the process when the handler returns (or after 5 seconds, whichever is earlier)
+	signal(SIGBREAK, dispatchSignalHandler);
```
djarek: There's one problem - dispatchSignalHandler is not signal-safe, invoking it may result in a deadlock (and loss of state).

djarek: As far as I understand SIGBREAK, if you use the standard API (the `signal` function), the handler will be invoked on an existing thread(as opposed to `SetConsoleCtrlHandler` which spawns a new thread that runs the handler) in an asynchronous manner. The deadlock can occur if a thread is interrupted while holding a `mutex` and the handler attempts to lock that mutex. Only signal-safe functions and lockfree atomic operations are allowed on mutable shared state.

DSpeichert: You are right about this being the expected behavior. However, in my tests, it appears that in fact a new thread is created. If the handler was running in an existing thread, it would either block entirely or it wouldn't need to "join" other threads. Yet, it does, because Windows is creating this new thread AND watching for it to finish (and then kills the rest of the process, if any).

image
Above is a debugger showing threads at a breakpoint set on the `Signals::dispatchSignalHandler`.  The `KernelBase.dll thread` doesn't exist during normal runtime.

The handling of SIGINT looks the same, and the [documentation](https://msdn.microsoft.com/en-us/library/ms811896.aspx?f=255&MSPPError=-2147217396#ucmgch09_topic3) specifically mentions:
> Note   When a Ctrl+C interrupt occurs, Win32 operating systems generate a new thread to handle the interrupt. This can cause a single-thread application, such as one ported from UNIX, to become multithreaded, potentially resulting in unexpected behavior.

So I believe this is a correct implementation. Also, the reason I don't exit the process with a code (e.g. `ExitProcess(0)`) is because after the thread returns nothing, it the "default" handler will return the special "window closed code":
> The program '[9944] theforgottenserver.exe' has exited with code -1073741510 (0xc000013a).

Found online:
> Exit Code 0xC000013A means that the application terminated as a result of a CTRL+C or  closing command prompt window

djarek: Windows development is harder than I thought... anyway, could you leave a mention of this fact in a comment (that dispatchSignalHandler does not need to be signal-safe on Windows)?

PR #2474, opened 2018-07-01
add onItemMoved event

Shawak:
I needed that

### DIFF
--- a/data/events/scripts/player.lua
+++ b/data/events/scripts/player.lua
@@ -116,7 +116,6 @@ function Player:onMoveItem(item, count, fromPosition, toPosition, fromCylinder, 
 end
 
 function Player:onItemMoved(item, count, fromPosition, toPosition, fromCylinder, toCylinder)
-	return true
 end
 
 function Player:onMoveCreature(creature, fromPosition, toPosition)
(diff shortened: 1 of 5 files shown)


data/events/scripts/player.lua
```diff
@@ -115,6 +115,10 @@ function Player:onMoveItem(item, count, fromPosition, toPosition, fromCylinder,
 	return true
 end
 
+function Player:onItemMoved(item, count, fromPosition, toPosition, fromCylinder, toCylinder)
+	return true
```
TheSumm: The return value is not necessary

src/events.cpp
```diff
@@ -500,6 +502,40 @@ bool Events::eventPlayerOnMoveItem(Player* player, Item* item, uint16_t count, c
...
 
+void Events::eventPlayerOnItemMoved(Player* player, Item* item, uint16_t count, const Position& fromPosition, const Position& toPosition, Cylinder* fromCylinder, Cylinder* toCylinder)
+{
+	// Player:onItemMoved(item, count, fromPosition, toPosition) or Player.onItemMoved(self, item, count, fromPosition, toPosition, fromCylinder, toCylinder)
+	if (info.playerOnItemMoved == -1) {
+		return;
+	}
+
+	if (!scriptInterface.reserveScriptEnv()) {
+		std::cout << "[Error - Events::eventPlayerOnMoveItem] Call stack overflow" << std::endl;
```
TheSumm: Error message reads the wrong event type.

PR #2476, opened 2018-07-06
Use SIMD auto-vectorization for XTEA

ranisalt:
This patch enables auto-vectorization for XTEA encryption and decryption, exploring vector (SSE, AVX) instructions to be auto-enabled by the compiler. Should render up to 16x faster encryption/decryption throughput on AVX2, 8x on AVX and 4x on SSE/Neon.

Things to consider:
- Maybe find another way to enable without `#ifdef` hell
- Code can probably be cleaner

On my machine (third-gen i7), it's performing up to 8x better than current algorithm.

### DIFF
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -22,6 +22,7 @@
 #include "protocol.h"
 #include "outputmessage.h"
 #include "rsa.h"
+#include "xtea.h"
 
 extern RSA g_RSA;
 
@@ -60,37 +61,14 @@ OutputMessage_ptr Protocol::getOutputBuffer(int32_t size)
 
 void Protocol::XTEA_encrypt(OutputMessage& msg) const
 {
-	const uint32_t delta = 0x61C88647;
-
 	// The message must be a multiple of 8
-	size_t paddingBytes = msg.getLength() % 8;
+	size_t paddingBytes = msg.getLength() % 8u;
 	if (paddingBytes != 0) {
 		msg.addPaddingBytes(8 - paddingBytes);
 	}
 
 	uint8_t* buffer = msg.getOutputBuffer();
-	const size_t messageLength = msg.getLength();
-	size_t readPos = 0;
-	const uint32_t k[] = {key[0], key[1], key[2], key[3]};
-	while (readPos < messageLength) {
-		uint32_t v0;
-		memcpy(&v0, buffer + readPos, 4);
-		uint32_t v1;
-		memcpy(&v1, buffer + readPos + 4, 4);
-
-		uint32_t sum = 0;
-
-		for (int32_t i = 32; --i >= 0;) {
-			v0 += ((v1 << 4 ^ v1 >> 5) + v1) ^ (sum + k[sum & 3]);
-			sum -= delta;
-			v1 += ((v0 << 4 ^ v0 >> 5) + v0) ^ (sum + k[(sum >> 11) & 3]);
-		}
-
-		memcpy(buffer + readPos, &v0, 4);
-		readPos += 4;
-		memcpy(buffer + readPos, &v1, 4);
-		readPos += 4;
-	}
+	xtea::encrypt(buffer, msg.getLength(), key);
 }
 
 bool Protocol::XTEA_decrypt(NetworkMessage& msg) const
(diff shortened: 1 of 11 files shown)


src/protocol.cpp
```diff
@@ -60,37 +61,14 @@ OutputMessage_ptr Protocol::getOutputBuffer(int32_t size)
...
-			sum -= delta;
-			v1 += ((v0 << 4 ^ v0 >> 5) + v0) ^ (sum + k[(sum >> 11) & 3]);
-		}
-
-		memcpy(buffer + readPos, &v0, 4);
-		readPos += 4;
-		memcpy(buffer + readPos, &v1, 4);
-		readPos += 4;
-	}
+	xtea::encrypt(buffer, msg.getLength(), buffer, key);
```
ranisalt: I am still to avoid this double sent parameter `buffer`

TechrlzDEV:
I'm having the same travis problem to compile this pull request.

src/xtea.cpp
```diff
@@ -0,0 +1,213 @@
...
+    0xafd9d683, 0x4e11503c, 0xec48c9f5,
+    0x8a8043ae, 0x28b7bd67, 0xc6ef3720,
+}};
+
+template<size_t BLOCK_SIZE>
+void XTEA_encrypt(
+        uint8_t data[BLOCK_SIZE * 8],
+        const key& k)
+{
+    auto words = reinterpret_cast<uint32_t*>(data);
```
djarek: Note that this `reinterpret_cast` is Undefined Behavior, because it breaks strict aliasing.

src/xtea.cpp
```diff
@@ -0,0 +1,213 @@
...
+        words[2 * i + 1] = right[i];
+    }
+}
+
+template<size_t BLOCK_SIZE>
+void XTEA_decrypt(
+        uint8_t data[BLOCK_SIZE * 8],
+        const key& k)
+{
+    auto words = reinterpret_cast<uint32_t*>(data);
```
djarek: Note that this `reinterpret_cast` is Undefined Behavior, because it breaks strict aliasing.

ranisalt: What if I copy the data to another buffer with proper type and `memcpy`?

ranisalt:
> I'm having the same travis problem to compile this pull request.

Fixed, but WIndows tooling still sux :stuck_out_tongue_closed_eyes:

TechrlzDEV:
Error	1	error C2143: syntax error : missing ';' before 'std::array<uint32_t,0x021>'	C:\src\xtea.cpp	32	1	theforgottenserver

kornholi:
Could you add some benchmark code and actual performance results? XTEA was/is an actual bottleneck with lots of players so thanks for the PR.

ranisalt:
> Could you add some benchmark code and actual performance results?

I'm preparing some code for it.

src/xtea.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include <assert.h>
+
+namespace xtea {
+
+namespace {
+
+constexpr uint32_t delta = 0x9E3779B9;
+
+template<size_t BLOCK_SIZE>
+void XTEA_encrypt(uint8_t data[BLOCK_SIZE * 8], const key& k)
```
djarek: Out of curiosity, has the change from reinterpret_cast to filling byte by byte affected the generated assembly?

src/xtea.cpp
```diff
@@ -0,0 +1,213 @@
...
+#endif // __x86_64__
+
+    for (auto i = 0u; i < length; i += 8u) {
+        XTEA_encrypt<1>(data + i, k);
+    }
+}
+
+void decrypt(uint8_t* data, size_t length, const key& k)
+{
+    assert(length % 8 == 0);
```
wgrriffel: this is bad for server availability, if someone throw a unaligned packet to decryption, server will abort

src/xtea.cpp
ranisalt: Hmmm I didn't bother to check, but the performance and binary size didn't change. I guess the assembly is pretty much the same, but better guarantee no aliasing issues right?

src/xtea.cpp
ranisalt: It is removed from generated code with `-DCMAKE_BUILD_TYPE=Release` which should definitely be used on production, but I think the server already aligns everything to 8 bytes and the check is redundant.

vankk: What about the others build type? It should be removed too?

src/xtea.cpp
```diff
@@ -0,0 +1,213 @@
...
+    {
+        constexpr auto step = 2u * 8u;
+        const auto blocks = (length & ~(step - 1));
+        for (auto i = 0u; i < blocks; i += step) {
+            XTEA_encrypt<2>(data + i, k);
+        }
+        data += blocks;
+        length -= blocks;
+    }
+#endif // __x86_64__
```
jo3bingham: Since the only difference between `#ifdef` blocks is the block size, you could trim this down to:
```cpp
#ifdef __AVX512F__
    const auto block_size = 128u;
#endif // __AVX512F__

#ifdef __AVX__
    const auto block_size = 32u;
#endif // __AVX__

#if defined(__SSE__) || defined(__ARM_FEATURE_SIMD32)
    const auto block_size = 8u;
#endif // __SSE__ || __ARM_FEATURE_SIMD32

#ifdef __x86_64__
    const auto block_size = 2u;
#endif // __x86_64__

    constexpr auto step = block_size * 8u;
    const auto blocks = (length & ~(step - 1));
    for (auto i = 0u; i < blocks; i += step) {
        XTEA_encrypt<block_size>(data + i, k);
    }
    data += blocks;
    length -= blocks;
```
Other than that, the latest changes LGTM.

src/xtea.cpp
ranisalt: If you are debugging your server, you probably want to get assertion errors.

src/xtea.cpp
ranisalt: Kinda, since each ifdef still enables the next one, as instruction sets include the previous ones. That is, `__AVX512F__` also enables `__AVX__` and `__SSE__`. I am still to test if the difference is significant, though.

ranisalt: Well, doing as you suggest slows by less than 5%, which is bearable IMHO.

jo3bingham: Personally, I prefer speed over code reduction. I would rather keep that 5% in this case.

DSpeichert: 5% is what new processors offer each year, so it seems like people care.

ranisalt: Managed to make a cleaner code with some preprocessing and templates.

ranisalt:
Added a new `Perfomance` CMake build type (generate with `cmake -DCMAKE_BUILD_TYPE=Performance`) that enables maximum optimization. `Release` build type only optimizes up to SSE, which yields double speed anyway.

ranisalt:
Enjoy!

PR #1375, opened 2015-05-23
Moved shared experience mechanism to lua

edumntg:
Moved the shared experience mechanism to events and also the shared
experience formula was fixed.

PR #2478, opened 2018-07-08
Moved shared experience formula to Lua

DSpeichert:
Co-authored-by: edumntg <edumntg@gmail.com>

Closes: #1375 

This is a rebased and re-formatted version of the old PR #1375 where the original author doesn't seem to want to update it.


data/events/scripts/party.lua
```diff
@@ -9,3 +9,28 @@ end
 function Party:onDisband()
 	return true
 end
+
+function Party:onShareExperience(exp)
+	local sharedExperienceMultiplier = 1.20 --20%
+	local vocationsIds = {}
+	
+	table.insert(vocationsIds, self:getLeader():getVocation():getId())
+	local vocationId = self:getLeader():getVocation():getId()
```
ranisalt: This one could be used on the line above to avoid function calls.

ranisalt:
I'd move the entire `shareExperience` function to Lua, but that's a start.

diath:
Won't that increase the multiplier for each promotion level as well? The source version used Vocation::getFromVocation to get the ID whereas this uses just Vocation::getId. Vocation::getFromVocation would be flawed as well though if someone had multi-level promotions, ideally there should be a function to get the base vocation ID.

PrinterLUA:
https://github.com/orts/server/blob/master/data/lib/core/vocation.lua

data/events/scripts/party.lua
```diff
@@ -9,3 +9,27 @@ end
...
+	local vocationsIds = {}
+
+	local vocationId = self:getLeader():getVocation():getId()
+	if vocationId ~= VOCATION_NONE then
+		table.insert(vocationsIds, vocationId)
+	end
+
+	for _, member in ipairs(self:getMembers()) do
+		vocationId = member:getVocation():getId()
+		if not isInArray(vocationsIds, vocationId) and vocationId ~= VOCATION_NONE then
```
WibbenZ: table.contains, isInArray is a compat function for some reason :P

WibbenZ:
LGTM

data/lib/core/vocation.lua
```diff
@@ -0,0 +1,7 @@
+function Vocation.getBase(self)
```
vankk: The variable name sounds vague - `v`.

data/events/scripts/party.lua
```diff
@@ -9,3 +9,27 @@ end
 function Party:onDisband()
 	return true
 end
+
+function Party:onShareExperience(exp)
+	local sharedExperienceMultiplier = 1.20 --20%
+	local vocationsIds = {}
+
+	local vocationId = self:getLeader():getVocation():getBase()
```
ninjalulz: Add `:getId()` at the end (on L23 as well).

data/events/scripts/party.lua
```diff
@@ -9,3 +9,27 @@ end
...
+			table.insert(vocationsIds, vocationId)
+		end
+	end
+
+	local size = #vocationsIds
+	if size > 1 then
+		sharedExperienceMultiplier = 1.0 + ((size * (5 * (size - 1) + 10)) / 100)
+	end
+
+	return (exp * sharedExperienceMultiplier) / (self:getMembers() + 1)
```
ninjalulz: Missing length operator (#) before `self:getMembers()`

data/events/scripts/party.lua
```diff
@@ -9,3 +9,27 @@ end
 function Party:onDisband()
 	return true
 end
+
+function Party:onShareExperience(exp)
+	local sharedExperienceMultiplier = 1.20 --20%
+	local vocationsIds = {}
+
+	local vocationId = self:getLeader():getVocation():getBase():getId()
```
ranisalt: That is a lot of indirections. Can someone with a large server benchmark?

DSpeichert: It is but I doubt anyone with a large server (if they are even using TFS 1.X) would just drop that in and test. I don't think that happens. I think we may find out in the long term and if that proves to be a problem, improve it.

edumntg:
I din't know the forgottenserver community was still active

PR #2480, opened 2018-07-12
Fix AppVeyor artifact patterns

DSpeichert:
Fixes #2479


kornholi:
Is there an actual reason to have the platform in the filename?

DSpeichert:
@kornholi Makes it more explicit what you are downloading from AppVeyor. You know what arch the dlls must be too.

Issue #2147, opened 2017-02-03
onLogout should have second parameter isForced!


Colandus:
onLogout event should have additional parameter isForced. This is necessary in case you have logic that expects player not to be logged out after preventing logout by returning false.

I would create a pull request if I wasn't noob with git :(

EPuncker:
@DSpeichert so if we are not implementing it at all after two attempts we might as well close this issue? I don't see any use for forced as well

DSpeichert:
I think the other PRs had a different meaning of forced, no? I read the 3 PRs and each really expected different meaning of "not forced" and "forced". I'll close this until there is a solid design proposal on what each should do.

PR #2482, opened 2018-07-13
Add secondary parameter "forced" to logout event

Leohige:
#2147

### DIFF
--- a/data/creaturescripts/scripts/logout.lua
+++ b/data/creaturescripts/scripts/logout.lua
@@ -1,4 +1,4 @@
-function onLogout(player)
+function onLogout(player, forced)
 	local playerId = player:getId()
 	if nextUseStaminaTime[playerId] then
 		nextUseStaminaTime[playerId] = nil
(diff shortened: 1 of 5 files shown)


vankk:
I don't think that is necessary because if you don't want the player logout just return **false** and if you want he logout just return **true**. I don't see the point of adding a new argument.

slawkens:
With this parameter, you can distinguish if the logout was forced (normal logout) or not forced (kicked by the anti-afk system). So I find it actually useful.

And in TFS 0.4 this parameter is present too, so I am for merging this.

vankk:
I still don't understand the logic of the needs of distinguish if it was forced or not. In my experience as a developer, if I restrict the player logout it is because there is something that he is doing in this particully moment that he shouldn't logout - even if it was by anti idle/afk.

slawkens:
> I haven't looked at the code but my immediate understanding would be that "forced" means kicked by e.g. anti-afk system, and not-forced would be normal logout - which is the opposite of what was explained. In other words, this change is confusing and the forced should be at least renamed to something better.

You are right. It was confusing for me too, I am not even sure if that's what this parameter means in TFS 1.2, as I took a look only at TFS 0.4. And it was always 'true', except kick after inactivity - anti-afk - where it is set to 'false'.

ranisalt:
When would it be relevant to know if a player was kicked or logged out willingly?

PR #2484, opened 2018-07-15
Codestyle fix

Sickness-chan:


src/items.cpp
```diff
@@ -190,25 +190,25 @@ const std::unordered_map<std::string, WeaponType_t> WeaponTypesMap = {
...
-	{"rum", FLUID_RUM },
-	{"swamp", FLUID_SWAMP },
-	{"tea", FLUID_TEA },
-	{"mead", FLUID_MEAD },
+	{"milk", FLUID_MILK},
+	{"mana", FLUID_MANA},
+	{"life", FLUID_LIFE},
+	{"oil", FLUID_OIL},
+	{"urine", FLUID_URINE},
+	{"coconut",	FLUID_COCONUTMILK},
```
djarek: Can you remove the extra space here as well?

Sickness-chan: no because there is actually no extra space there, it's a space like the others too, just github displaying it as tab for some reason

i had actually noticed too and checked in the file but the file is fine, just github acting up

WibbenZ: It's actually a tab, but the editor shows it as a space.
Send another commit where you replace the tab with a space that will fix the issue.

Sickness-chan: weird because yesterday i even copied the space/tab and checked the charcode compared to a space, maybe the editor trolled me too

WibbenZ:
Thanks! :)

PR #2485, opened 2018-07-15
Boosts name changes and 2 new functions

nekiro:
These names were too long and I think this way they will be better looking, also when implementing case switch item loading, you guys forgot about critical hit and rest of boosts. That's pretty much it.

Two new functions:
- player:getSpecialSkill(specialSkillType)
- player:addSpecialSkill(specialSkillType, value)


raymondtfr:
I think every instance of "special skills" should be renamed to "imbuing/imbuement skill" since thats what they actually are, e.g. "getImbuementSkill()" or "getImbuingSkill", etc.

joseluis2g:
Some items got crit stats and arent imbuiments so I dont know if thats correct.

raymondtfr:
Can you paste here a tibia.wikia link to one of those items?

Anyhow, it should be "getSpecialSkillLevel", etc; if imbuing doesn't make sense then.

joseluis2g:
Wand of darkness is the one that is working until last version but the first release of this items was also with Yana npc who changed gold tokens for charged/overcharged items with those stats (crit, leech, etc)
https://www.tibiawiki.com.br/wiki/Wand_of_Darkness
http://tibia.wikia.com/wiki/Updates/10.94

raymondtfr:
Hmm.. thats just an item with predefined imbuing skills, an item doesn't necessarily needs to go through the process of "imbuing" to be known as a weapon with imbuing skill, as long as it has any those new skills (life/mana leech and critical hit) it should be considered an imbuing skill.

It would make sense to think of it as a "special skill" only if there's no certainty as of how further we're gonna implement those new changes/updates, since on vanilla they have uniformed this new feature (removing weapons with predefined "boost", removing enchanting, that wand is an exception - but still has imbuing attributes).

nekiro:
@raymondtfr They are additional skills/stats in player skill window, doesnt matter that they are connected with imbuement system on real tibia (but not only) or not. I see no reason to call them imbuement skill, because its simply not imbuement skill.

rottylunia:
do not add this., tfs is based on otcleint, otcleitn dont even display info about critical chance, mana leech so why its needed?

if u gojnna update this then update client to tibia 11, because u are still trying to mimic real tibia but the main developer (mark) said tfs is now for otclient so maybe do some good features for otclient instead. for example we could readd magic texts like in older tibia clients.

ranisalt:
I think they are still too long. We should use `enum class` to properly scope those, not the C-style namespace we use.

Kamenuvol:
addSpecialStat, addStat, addAttribute (current one), protobuf things
Critical and leech are Item attributes that change Player stats. No need to make special function for it.

src/items.cpp
```diff
@@ -95,10 +95,10 @@ const std::unordered_map<std::string, ItemParseAttributes_t> ItemParseAttributes
 	{"magicpointspercent", ITEM_PARSE_MAGICPOINTSPERCENT},
 	{"criticalhitchance", ITEM_PARSE_CRITICALHITCHANCE},
 	{"criticalhitamount", ITEM_PARSE_CRITICALHITAMOUNT},
-	{"hitpointsleechchance", ITEM_PARSE_HITPOINTSLEECHCHANCE},
-	{"hitpointsleechamount", ITEM_PARSE_HITPOINTSLEECHAMOUNT},
-	{"manapointsleechchance", ITEM_PARSE_MANAPOINTSLEECHCHANCE},
-	{"manapointsleechamount", ITEM_PARSE_MANAPOINTSLEECHAMOUNT},
+	{"lifeleechchance", ITEM_PARSE_LIFELEECHCHANCE},
```
DSpeichert: Shouldn't this be called `hp` instead of `life`?

nekiro: well originallly by cipsoft its named life leech, so I dont know.

DSpeichert: Okay. I'll give it a few more days for others to look at.

PR #2488, opened 2018-07-26
Fix #2486

### LINKED ISSUE #2486: Possible bug inside Game::internalRemoveItem
https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L1343

The call to item->onRemoved() is done when test parameter is true even if the item has not been removed(since its a test to remove the item). I also found other places where "ReleaseItem(item)" is called butitem->onRemoved() is not called leading to a inconsistency if we want to use item->onRemoved() as a callback for when a item is truly removed from the game.

CristoferMartins:
Since the test parameter is only used when internalRemoveItem is called by trade code i think nothing bad happened so far but i want to fix this since its still a bug.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -1334,13 +1334,13 @@ ReturnValue Game::internalRemoveItem(Item* item, int32_t count /*= -1*/, bool te
 		cylinder->removeThing(item, count);
 
 		if (item->isRemoved()) {
+			item->onRemoved();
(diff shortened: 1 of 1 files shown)


GoularPink:
hi, there we would also need to add the item->onRemoved();   too?

https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L1174

PR #2496, opened 2018-08-29
Load RSA private key from PEM file

### LINKED ISSUE #2216: Custom RSA key loader
The current solution has bad security (does anyone remember to change their RSA key?).

## Requirements
- The keys should be loadable from PEM files generated by ```openssl rsa``` CLI tool
- Support for encrypted keys is optional (would be nice to have)
- ```config.lua``` must specify the absolute/global path of the key file
- If the path is missing, default keys should be loaded
- If default keys have been loaded and server is in production mode (#2215) the server should fail to start

ranisalt:
Implements logic to load PEM-encoded RSA private key on server start.

Currently lacks:
- Setting file path (is it really needed, though?)
- Detecting if using default key
- Some form of documentation on how to create keys (there's plenty on the web, though)
- Windows (sorry, can't do it)

Closes #2216.


src/rsa.cpp
```diff
@@ -21,72 +21,54 @@
...
+	CryptoPP::Integer m{reinterpret_cast<uint8_t*>(msg), 128};
+	auto c = pk.CalculateInverse(prng, m);
+	c.Encode(reinterpret_cast<uint8_t*>(msg), 128);
+}
 
-	mpz_clear(c);
-	mpz_clear(m);
+void RSA::loadPEM(const std::string& filename)
+{
+	static std::string header = "-----BEGIN RSA PRIVATE KEY-----";
```
djarek: const? (`string_view` would be a better type here)

ranisalt: @djarek `string_view` is C++17 only :(

DSpeichert:
@ranisalt anything else to do here?

ranisalt:
I had no feedback if anything from the list of missing things I added is wanted (I personally don't care about them) and I'd like someone to test it on Windows.

DSpeichert:
* Path doesn't really need to be customized but I'm thinking if it wouldn't fit in `data/` better
* Detecting if using default key isn't helpful until there's a common practice of changing default key. There isn't, so such warning would confuse most people.
* Documentation isn't required now, it's not like we have it for everything anyway
* Works on Windows.

ranisalt:
I prefer it at the root folder since the data folder is more about world content rather than server configuration. Briefly, I think it belongs besides config.lua instead of map and scripts.

ranisalt:
Thanks!

PR #2497, opened 2018-09-01
Combat and boosts changes

### LINKED ISSUE #2094: doAreaCombatCondition(cid, pos, area, condition, effect) crashes TFS

1. Use the function from the title above in any spell.
2. TFS should crash when the spell is cast for the second time.

### Expected behaviour
The function should not crash TFS.

### Actual behaviour
The function crashes TFS.

### Environment

TFS 1.3

nekiro:
To apply these changes #2485 needs to be merged.

What does this pr brings?
- completely rewritten combat.cpp, instead of bunch of functions, we are now using only two "internal" ones.
one main function doCombat() with target and position parameter, which leads to doTargetCombat() and doAreaCombat()
- two new functions for boosts:
Combat::checkCriticalHit() which is used to check if the combat is critical
Combat::checkLeech which is used to check if we can leech combat.
- now we do not use doTargetCombatHealth, doTargetCombatMana ... in lua side anymore, only two functions for everything:
doTargetCombat(cid, target, type, min, max, effect)
doAreaCombat(cid, type, pos, area, min, max, effect)
- doAreaDispel/doAreaCondition ... were removed, because they are deprecated in my opinion.
- some bugs with boosts, ex. criticals weren't calculated properly, when player was in combat with another player, are fixed.

I'm not 100% sure if everything is working and fine, so I will appreciate review and testing this pull request. (I tested it myself tho)

Resolve #2094


ranisalt:
I would put make those static void functions free, and enclose in a namespace.

DSpeichert:
@nekiro Please rebase for conflicts

DSpeichert:
@nekiro Looks like we got conflicts again?

nekiro:
> @nekiro Looks like we got conflicts again?

I see no conflicts right now, but this pr has to be tested before merged

Olimpotibia:
Great

nekiro:
Resolved the conflicts, can it be merged finally?

src/combat.cpp
```diff
@@ -731,7 +876,54 @@ void Combat::CombatFunc(Creature* caster, const Position& pos, const AreaCombat*
...
+					}
+
+					if (success) {
+						if (params.origin != ORIGIN_MELEE && damageCopy.primary.value != 0 && damageCopy.secondary.value != 0) {
+							for (const auto& condition : params.conditionList) {
+								if (caster == creature || !creature->isImmune(condition->getType())) {
+									Condition* conditionCopy = condition->clone();
+									conditionCopy->setParam(CONDITION_PARAM_OWNER, caster->getID());
+
+									//TODO: infight condition until all aggressive conditions has ended
```
DSpeichert: Is there something missing?

nekiro: If you mean that „todo” then it’s comment from the base tfs code

Leo32onGIT: @DSpeichert 
This should of been merged ages ago.
Leech amounts are busted without it.

nekiro: I can probably add later that todo, wont put all of it in one pr.

Leo32onGIT:
I think this needs backwards compatibility with doTargetCombatHealth() so old scripts don't break.

Maybe all this needs is a couple lines in lib/compat?

PR #2501, opened 2018-09-25
Add error message when failing to load RSA key

Vulcanx:


ranisalt:
While at it, you could check if any crypto++ function we call can throw recoverable errors.

Issue #2499, opened 2018-09-17
[Question] Combat Areas on functions


vankk:
I was developing a new method of spells and I found an interessing issue that I can't define a new Combat Area inside a function, [it should be on load](https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L11527-L11531).

For example [on this script](https://github.com/otland/forgottenserver/blob/master/data/spells/scripts/attack/eternal_winter.lua) it will works perfectly, and [on this one](https://pastebin.com/mr4UKDpt) will return the error on Console.

My question is divided into two parts:
**1-** Why do we need this method to avoid the Combat Area inside the functions?
**2-** What could happen if comment the condition statament? (Overflow, underflow on ScriptEvenId?)

Or may I doing it wrongly? 

Thanks.

Regards,
vankk.

ranisalt:
Did you try? I don't think anything bad would happen.

vankk:
@ranisalt I tested removing the condition statement and it will returns that Area isn't founded - so, we cannot define inside functions a new Combat Area, only if it was previous loaded.

ranisalt:
What happens? Can you post backtrace and logs?

vankk:
The TFS just crashed with a SIGSEGV. Here is the [backtrace simple](https://pastebin.com/zn5Y4R6K), aslo here is the [backtrace full](https://pastebin.com/iApQTFxz). 

The previous backtrace was from Ubuntu 14.04 64 bits, the Spell that I used [was this one](https://pastebin.com/mr4UKDpt).

PR #2503, opened 2018-10-01
Make AreaCombat::copyArea static

marksamman:
AreaCombat::copyArea doesn't have to be an instance method, the instance pointer is omitted in the call by making it static.

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -1111,7 +1111,7 @@ void AreaCombat::getList(const Position& centerPos, const Position& targetPos, s
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ubap:
Good catch,
For me it looks like this method should be moved to some util class, however I'm not really into theforgottenserver yet.

ranisalt:
It can be a free function inside `combat.cpp` to avoid it being declared on other files it is never used and allow it to be inlined.

ubap:
Thanks for your response

vankk:
Mark, this PR has any relation with #2499 ?

PR #2504, opened 2018-10-01
Fix inconsistent parameter names

marksamman:
Client IP was written as "clientIp" and "clientip", renamed all occurrences with "clientIP"

Renamed "className" to "globalName" in luascript.h to match the name used in the cpp file

Renamed "i, l" to "data, length" in xtea.cpp to match the names used in xtea.h

### DIFF
--- a/src/ban.cpp
+++ b/src/ban.cpp
@@ -24,15 +24,15 @@
 #include "databasetasks.h"
 #include "tools.h"
 
-bool Ban::acceptConnection(uint32_t clientip)
+bool Ban::acceptConnection(uint32_t clientIP)
 {
 	std::lock_guard<std::recursive_mutex> lockClass(lock);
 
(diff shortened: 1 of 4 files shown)


ubap:
Hi,
Good changes.
One question, is the name "clientIP "in correct naming convention format for theforgottenserver? I'm not really into theforgottenserver naming conventions, but often camelCase is used
https://en.wikipedia.org/wiki/Camel_case

In camel case it would be "clientIp".

Finally, maybe there is some code quality tool to check for all incorrect names?

Best regards!

marksamman:
It's currently mixed, my preference is to use the case that the abbreviation or acronym is most commonly written in and defer to camel case when it's a regular word. Both clientIp and clientIP are considered camel case.

I used a static analyzer to find these, it didn't spot any other mismatches.

PR #2505, opened 2018-10-01
Don't check if unsigned values are less than zero

marksamman:
Unsigned values cannot be less than zero, so it's enough to check if getItemCount() equals zero in these two cases.


ubap:
Hi @marksamman ,
Code looks good, I have a few questions
Is the performance better? Does it produce less warnings? Whats the benefit of the change?

ranisalt:
> Is the performance better?

No, at least not significantly.

> Does it produce less warnings?

Maybe, I know gcc has some warnings on integer signedness (it warns if you test unsigned < 0).

> Whats the benefit of the change?

Semantics only. It's actual implementation depends on processor, e.g. on MIPS it would probably generate a `beq $zero` (branch on equal to zero) for both checks, as the compiler can infer it is never less than zero.

PR #2510, opened 2018-10-05
Fix field attribute parsing in items.xml

marksamman:
The values of ticks, start and count were always discarded because they're reset on next iteration unless I'm missing something. I'm not sure how/if this ever worked, please have a look that this does not have unintended side-effects, I suspect that some users may have written their items.xml with workarounds for this problem. "start > 0" would never be true when reading the damage key.

### DIFF
--- a/src/items.cpp
+++ b/src/items.cpp
@@ -1155,12 +1155,10 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 						it.combatType = combatType;
 						it.conditionDamage.reset(conditionDamage);
 
+						uint32_t ticks = 0;
+						int32_t start = 0;
+						int32_t count = 1;
 						for (auto subAttributeNode : attributeNode.children()) {
-							uint32_t ticks = 0;
-							int32_t damage = 0;
-							int32_t start = 0;
-							int32_t count = 1;
-
 							pugi::xml_attribute subKeyAttribute = subAttributeNode.attribute("key");
 							if (!subKeyAttribute) {
 								continue;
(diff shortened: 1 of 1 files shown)


marksamman:
Looks like this was broken in #2481.

ubap:
Looks good code wise but can't test it right now

DSpeichert:
Indeed it looks like these changes match the original way it worked, here's the commit part that broke it:
https://github.com/otland/forgottenserver/commit/709fb3464a43865d6fc0d90206a6c2d4714fea09#diff-cdbcf3a5db86a021022fed1a9f317741L805

It's not entirely clear at first what's the intended algorithm here, but after analyzing that I noticed that most likely for a field item definition like this:
```xml
	<item id="1492" article="a" name="fire field">
		<attribute key="replaceable" value="0" />
		<attribute key="field" value="fire">
			<attribute key="damage" value="20" />
			<attribute key="ticks" value="10000" />
			<attribute key="count" value="7" />
			<attribute key="damage" value="10" />
		</attribute>
	</item>
```

The idea is that `ticks` and `count` applies to the second `damage` (=10). The first `damage` attribute is not a mistake but is added with implied 0 ticks and 1 count.
between #2481 and this PR, we were losing the `start`, `ticks`, and `count` attributes, always resetting them like @marksamman noted.

waqmaz93:
> Indeed it looks like these changes match the original way it worked, here's the commit part that broke it: [709fb34#diff-cdbcf3a5db86a021022fed1a9f317741L805](https://github.com/otland/forgottenserver/commit/709fb3464a43865d6fc0d90206a6c2d4714fea09#diff-cdbcf3a5db86a021022fed1a9f317741L805)
> 
> It's not entirely clear at first what's the intended algorithm here, but after analyzing that I noticed that most likely for a field item definition like this:
> 
> ```
> 	<item id="1492" article="a" name="fire field">
> 		<attribute key="type" value="magicfield" />
> 		<attribute key="replaceable" value="0" />
> 		<attribute key="field" value="fire">
> 			<attribute key="damage" value="20" />
> 			<attribute key="ticks" value="10000" />
> 			<attribute key="count" value="7" />
> 			<attribute key="damage" value="10" />
> 		</attribute>
> 	</item>
> ```
> 
> The idea is that `ticks` and `count` applies to the second `damage` (=10). The first `damage` attribute is not a mistake but is added with implied 0 ticks and 1 count. between #2481 and this PR, we were losing the `start`, `ticks`, and `count` attributes, always resetting them like @marksamman noted.

this is not working properly.
field should deal 20 dmg and then 7 times 10 dmg every 10 seconds.
instead it deals 25 and 30 dmg and then 7 times 10 dmg every 10 seconds.

PR #2514, opened 2018-11-17
Missions within the same storage

rookgaard:
While using the same storage in multiple missions in one quest having:
- `storage = 3` first mission is completed - OK
- `storage = 4` first mission disappear, second shows - not OK

Adding `ignoreendvalue="true"` in first mission causes:
- `storage = 3` first mission is completed - OK
- `storage = 4` first mission is still completed, second shows - OK

### DIFF
--- a/data/XML/quests.xml
+++ b/data/XML/quests.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <quests>
 	<quest name="Example Quest I" startstorageid="1001" startstoragevalue="1">
-		<mission name="Example Mission 1" storageid="1001" startvalue="1" endvalue="3">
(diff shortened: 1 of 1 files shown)


DSpeichert:
Can you give more background information?
What does `ignoreendvalue` normally do?

TheSumm:
Setting `ignoreendvalue` will cause the mission to show up in the quest log if the player's storage value is greater than or equal to the start value. If the storage value is greater than the end value it will still show the mission (the last description of the mission), whereas without `ignoreendvalue` the mission would disappear again.

 Generally, if you wanted to do a quest with multiple missions you would use a separate storage value for each mission, so the example quest is a bad example.
It is however also valid to have quests where only the current mission shows up. That's up to the user.

DSpeichert:
I've merged this because it's a bad example anyway.

PR #2516, opened 2018-11-28
Halt instant exit when failing to load the server

Vulcanx:
annoying to have to use a console to view the errors since it instantly closes if you have an error while loading server resources

### DIFF
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -93,6 +93,7 @@ int main(int argc, char* argv[])
 	g_scheduler.join();
 	g_databaseTasks.join();
 	g_dispatcher.join();
+	std::cin.get();
 	return 0;
 }
 


ranisalt:
That's annoying for automatic restarters. I think it is user's responsibility, not the server's.

On Linux, you can set up a file to redirect server output. Windows command prompt sucks completely.

DSpeichert:
Correct, interactivity breaks any auto-restarter.
Wouldn't that change also break scheduled shutdown, e.g. on server save?

jo3bingham:
>On Linux, you can set up a file to redirect server output. Windows command prompt sucks completely.

You can do the same on Windows...

@Vulcanx I can't approve of this change for the reasons stated above. On Windows, you shouldn't be running the TFS executable directly (i.e., double-clicking on it) anyway, you should be running it from Command Prompt (or PowerShell, or whatever). That way, if the server goes down unexpectedly, the output will still be there.

Open Command Prompt and `cd` to the directory where your server executable is (e.g., `cd C:\forgottenserver\`) then run it (e.g., `./theforgottenserver-x64.exe`). If you'd rather redirect the output to a file you can do so with `>` or `>>`. `>` overwrites the file, and `>>` appends to it (e.g. `./theforgottenserver-x64.exe >> log.txt` will redirect all output to log.txt in the same directory as your server executable and will create the file if it doesn't exist). If you need to shutdown your server manually, you just need to go back to your command prompt and press `Ctrl+c`, that will send an interruption signal (`SIGINT`) to the server and terminate it.

DSpeichert:
Closing this to make it clear that this is not a proper solution and can't be merged per reasons given above.

Issue #2518, opened 2019-01-13
Changefloor from items.xml can't read second direction


gesior:
### Steps to reproduce
1. Try to walk on Pyramid ramp at corner:
```
        <item id="1553" article="a" name="ramp">
		<attribute key="floorchange" value="north" />
		<attribute key="floorchange" value="west" />
	</item>
```
2. It moves you to other tile on same floor.

### Expected behaviour
It should teleport to upper floor, to north-west (X-1, Y-1, Z+1)

### Actual behaviour
It teleports to upper floor, to west (X-1, Y, Z+1). It moves to 'stairs' tile (on upper floor), which make player go downstairs again.

I found that it's because of that code:
https://github.com/otland/forgottenserver/blob/b88a13af044366bd5102495527388299337a482b/src/items.cpp#L671
```
it.floorChange = it2->second;
```
It should be:
```
it.floorChange |= it2->second;
```

DSpeichert:
@gesior Sounds sane. Care to submit PR?

gesior:
@DSpeichert Done

PR #2519, opened 2019-01-14
#2518 changefloor overwrites first parameter from items.xml

gesior:
Fix floorchange loading from items.xml
https://github.com/otland/forgottenserver/issues/2518


marksamman:
Thanks!

PR #2522, opened 2019-01-30
player storage values as string

rookgaard:
Hello. I would like to see in current TFS nice feature which I found back in 0.4 - possibility to store player storage values not only as integer, but as string. If my change has no chance to merge master, please at least verify the code, if I have not made gross errors.


src/luascript.cpp
```diff
@@ -8524,9 +8524,24 @@ int LuaScriptInterface::luaPlayerGetStorageValue(lua_State* L)
...
+	std::string strValue;
+	if (player->getStorageValue(key, strValue)) {
+		int32_t intValue;
+		try {
+			intValue = stoi(strValue);
+		} catch (std::invalid_argument& e) {
+			intValue = 0;
+		} catch (std::out_of_range& e) {
+			intValue = 0;
+		} catch (...) {
```
ranisalt: If you are using `catch (...)`, why the two above `catch` statements doing the same?

src/player.cpp
```diff
@@ -621,8 +621,19 @@ uint16_t Player::getLookCorpse() const
...
+void Player::addStorageValue(const uint32_t key, const std::string& strValue, const bool isLogin/* = false*/)
 {
+	int32_t value;
+	try {
+		value = stoi(strValue);
+	} catch (std::invalid_argument& e) {
+		value = 0;
+	} catch (std::out_of_range& e) {
+		value = 0;
+	} catch (...) {
```
ranisalt: Same as above. Also, never use pokemon exceptions.

src/player.cpp
```diff
@@ -638,11 +649,22 @@ void Player::addStorageValue(const uint32_t key, const int32_t value, const bool
...
+		std::string oldStrValue;
+		getStorageValue(key, oldStrValue);
+
+		try {
+			oldValue = stoi(oldStrValue);
+		} catch (std::invalid_argument& e) {
+			oldValue = 0;
+		} catch (std::out_of_range& e) {
+			oldValue = 0;
+		} catch (...) {
```
ranisalt: Once more.

src/player.cpp
```diff
@@ -638,11 +649,22 @@ void Player::addStorageValue(const uint32_t key, const int32_t value, const bool
 		}
 	}
 
-	if (value != -1) {
+	if (strValue != "-1") {
```
ranisalt: Dangerous. Using strings instead of integers will change the "undefined" value, "-1" should not be a special string and can be surprising. You should check for emptiness instead.

src/player.cpp
```diff
@@ -656,11 +678,11 @@ void Player::addStorageValue(const uint32_t key, const int32_t value, const bool
 	}
 }
 
-bool Player::getStorageValue(const uint32_t key, int32_t& value) const
+bool Player::getStorageValue(const uint32_t key, std::string& value) const
 {
 	auto it = storageMap.find(key);
 	if (it == storageMap.end()) {
-		value = -1;
+		value = "-1";
```
ranisalt: Same as two above.

src/player.cpp
```diff
@@ -3660,8 +3682,11 @@ void Player::genReservedStorageRange()
 {
 	//generate outfits range
 	uint32_t base_key = PSTRG_OUTFITS_RANGE_START;
+	std::ostringstream ss;
 	for (const OutfitEntry& entry : outfits) {
-		storageMap[++base_key] = (entry.lookType << 16) | entry.addons;
+		ss.str("");
+		ss << ((entry.lookType << 16) | entry.addons);
+		storageMap[++base_key] = ss.str();
```
ranisalt: Go straight with `std::to_string`.

src/player.cpp
```diff
@@ -4073,16 +4098,29 @@ GuildEmblems_t Player::getGuildEmblem(const Player* player) const
...
+	std::string strValue;
+	if (getStorageValue(PSTRG_MOUNTS_CURRENTMOUNT, strValue)) {
+		int32_t value;
+		try {
+			value = stoi(strValue);
+		} catch (std::invalid_argument& e) {
+			value = 0;
+		} catch (std::out_of_range& e) {
+			value = 0;
+		} catch (...) {
```
ranisalt: Same again.

src/player.cpp
```diff
@@ -4073,16 +4098,29 @@ GuildEmblems_t Player::getGuildEmblem(const Player* player) const
...
 	}
 	return 0;
 }
 
 void Player::setCurrentMount(uint8_t mountId)
 {
-	addStorageValue(PSTRG_MOUNTS_CURRENTMOUNT, mountId);
+	std::ostringstream ss;
+	ss << +mountId;
+	addStorageValue(PSTRG_MOUNTS_CURRENTMOUNT, ss.str());
```
ranisalt: `std::to_string`

src/player.cpp
```diff
@@ -4181,13 +4223,26 @@ bool Player::untameMount(uint8_t mountId)
...
+		value = 0;
+	} catch (...) {
+		value = 0;
+	}
+
 	value &= ~(1 << (tmpMountId % 31));
-	addStorageValue(key, value);
+	std::ostringstream ss;
+	ss << value;
+	addStorageValue(key, ss.str());
```
ranisalt: `std::to_string`

src/player.cpp
```diff
@@ -4213,11 +4268,22 @@ bool Player::hasMount(const Mount* mount) const
...
 	}
 
+	int32_t value;
+	try {
+		value = stoi(strValue);
+	} catch (std::invalid_argument& e) {
+		value = 0;
+	} catch (std::out_of_range& e) {
+		value = 0;
+	} catch (...) {
```
ranisalt: Same.

src/quests.cpp
```diff
@@ -25,8 +25,19 @@
...
+
 	int32_t value;
-	player->getStorageValue(storageID, value);
+	try {
+		value = stoi(strValue);
+	} catch (std::invalid_argument& e) {
+		value = 0;
+	} catch (std::out_of_range& e) {
+		value = 0;
+	} catch (...) {
```
ranisalt: Same.

src/quests.cpp
```diff
@@ -63,11 +74,22 @@ bool Mission::isStarted(Player* player) const
...
 	}
 
+	int32_t value;
+	try {
+		value = stoi(strValue);
+	} catch (std::invalid_argument& e) {
+		value = 0;
+	} catch (std::out_of_range& e) {
+		value = 0;
+	} catch (...) {
```
ranisalt: Same.

src/quests.cpp
```diff
@@ -85,11 +107,22 @@ bool Mission::isCompleted(Player* player) const
...
```
ranisalt: Same.

src/quests.cpp
```diff
@@ -132,8 +165,21 @@ bool Quest::isStarted(Player* player) const
...
+
 	int32_t value;
-	if (!player->getStorageValue(startStorageID, value) || value < startStorageValue) {
+	try {
+		value = stoi(strValue);
+	} catch (std::invalid_argument& e) {
+		value = 0;
+	} catch (std::out_of_range& e) {
+		value = 0;
+	} catch (...) {
```
ranisalt: Same.

src/iologindata.cpp
```diff
@@ -846,7 +846,7 @@ bool IOLoginData::savePlayer(Player* player)
 	player->genReservedStorageRange();
 
 	for (const auto& it : player->storageMap) {
-		query << player->getGUID() << ',' << it.first << ',' << it.second;
+		query << player->getGUID() << ',' << it.first << ",'" << it.second << "'";
```
ranisalt: You should use `db.escapeString` here instead of manually quoting.

src/player.cpp
rookgaard: Using "-1" as the default value gives me the advantage that in the case of later projection to int I get an easy to compare value in lua/c++. What could I use instead?

ranisalt: Empty string

rookgaard:
I corrected all of the above-mentioned comments. However, I have a question regarding the preparation of int to string mapping. How would it look like and whom would it serve?

src/luascript.cpp
```diff
@@ -8524,9 +8524,22 @@ int LuaScriptInterface::luaPlayerGetStorageValue(lua_State* L)
...
 
 	uint32_t key = getNumber<uint32_t>(L, 2);
-	int32_t value;
-	if (player->getStorageValue(key, value)) {
-		lua_pushnumber(L, value);
+	std::string strValue;
+	if (player->getStorageValue(key, strValue)) {
+		int32_t intValue;
+		try {
+			intValue = stoi(strValue);
```
djarek: avoid accidental ADL - `std::stoi`

rookgaard: Thanks, I've changed

vankk:
I personally am totally against this PR. **We should keep only on numbers**, and not uses string on Storages. I'll use an example that is happening on now days with me. 

I'm hosting a server 8.60, that wasn't coded by myself - who coded that used alot of return string values, don't ask me why. Sometimes the values are lost, don't ask me why too. The player is doing a Task and after that goes to string only God knows why, and the progress are lost, I did countless tests and that is pretty fine, it don't have anything to envolves strings, but it keeps with that problem.

Why I'm telling? It is because we still can avoid this mistake. We don't need someone to go through what we've been through in the past, and we're going through the now days with the old versions. I think that **1.x is suppost to be modern and clean, not an old rev**. We should learn with our past and mistakes. I'm trully favor of we only keep with numbers on Storages.

ranisalt:
While I approve the changes, I am against merging for the reasons @vankk has pointed out. It could be kept alive as a patch for people transitioning from 0.4, though.

Znote:
What if we add a supporting string reference table as a way to determine storage value groups?

string <-> key(s) <-> value(s)

If I for instance create a system "football arena" that might require lots of different storage values to be used. 
I think a string to categorize and wrap the relevant storage values for my system would be easier to deal with than "pick a random number to assign these storage values for that your OT does not currently use". 

Something along the lines of:
```sql
#Even though lots of players have this key, the string could be unique, 
#and we determine the relevant player in the player_storage table
CREATE TABLE IF NOT EXISTS `player_storage_strings` (
  `key_str` VARCHAR(32) NOT NULL,
  `key_int` INT(11) NOT NULL, 
) ENGINE=InnoDB;
```

```lua
-- Set the #1 key in "Football Arena" value 3
player:setStorageValue("Football Arena", 1, 3) 
```

Want to figure out which keys are used in this group?
```lua
-- Return table of all matching storage ids from player_storage, and their values
player:getStorageValue("Football Arena") 

-- Returns the value of the #2 storage key in category "Football Arena".
player:getStorageValue("Football Arena", 2) 
```

The actual integer keys could be generated based on each OT servers availability in player_storage? Preferably with a pre-determined seed. (etc storage value 61000 - 71000 reserved for string key generations). 

And going oldschool and setting and retrieving values
```lua
-- We know this isn't a string, so skip that table and 
-- just retrieve directly from the player_storage table
player:getStorageValue(50123)

-- No string, so set player_storage (50123 value 5). 
player:setStorageValue(50123, 5)
```

1. player_storage will not get cluttered with lots of strings, which is a performance issue
2. No duplicate strings in database, quicker to search through
3. Solves the problem with unorganized key value pairs if used correctly
4. Can create compat file with full backwards compatibility

compat.lua:
```lua
-- Retrieve the first value in string group if key is a string
function getPlayerStorageValue(cid, key) 
	local p = Player(cid) 
	if type(key) == 'number' then
		-- Retrieve based on int
		return p and p:getStorageValue(key) or false 
	else 
		-- Retrieve first value of string group
		return p and p:getStorageValue(key, 1) or false 
	end
end
-- Set the first value in string group if key is a string
function setPlayerStorageValue(cid, key, value) 
	local p = Player(cid) 
	if type(key) == 'number' then
		return p and p:setStorageValue(key, value) or false 
	else 
		return p and p:setStorageValue(key, 1, value) or false 
	end
end
```

And suddenly, we have an improved version of the old system. Old string storage value scripts will still work, but if used the "1.3" way it is more organized, which would have solved the issue @vankk has. (if the original creator of that 8.6 server had this system in place).

rookgaard:
Hello. Sorry for my huge 3-months delay, but now I'm back with another approach. Instead of changing existing storageMap from int to string, I've added another map particularly for strings. Now, instead of mapping values in many places (such as quests) I'm doing it only during loading character from database. Whereas on LUA side `if (isNumber(L, 3))` will determine value type and assign it to proper map. To protect from duplicating keys when saving player to database, I also added the clearing of the "opposite" map when overwriting values. I hope that these changes will be better appreciated than the previous ones :)

data/migrations/24.lua
```diff
@@ -1,3 +1,6 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 25 (change type of `player_storage`.`value` from int to string)")
+
+	db.query("ALTER TABLE `player_storage` CHANGE `value` `value` VARCHAR(255) NOT NULL DEFAULT '0'")
```
djarek: I don't think storing all values as a string is a good idea. What if a user wants to store "123" as a string? Your code will force that value into an integer after a reload.

rookgaard: Yes, it will perform `value = stoi(strValue);` for each storage entry, but it will do it once during user connection. Would a better idea be an additional table dedicated to string storages?

Kamenuvol:
It would be better make everything string. And have a getStorageNumber to get and convert to number if necessary.

EvilHero90:
In all honesty, I'd rather have this done completly in lua with overloading functions, in such a way Znote had his idea in mind.
```lua
STORAGE_STRING_MAP = {}

Player.oldGetStorageValue = Player.getStorageValue
Player.oldSetStorageValue = Player.setStorageValue

function Player.getStorageValue(self, key)
  if type(key) == "string" then
    local value = STORAGE_STRING_MAP[key]
    if value then
      return self:oldGetStorageValue(value)
    else
      print("There is no key string: ".. key)
    end
  elseif type(key) == "number" then
    return self:oldGetStorageValue(key)
  end
  return nil
end

function Player.setStorageValue(self, key, value, int_key)
  if type(key) == "string" then
    local check = STORAGE_STRING_MAP[key]
    if check then
      return self:oldSetStorageValue(check, value)
    else
      if int_key then
        STORAGE_STRING_MAP[key] = int_key
        return self:oldSetStorageValue(int_key, value)
      else
        print("There is no int_key set for: ".. key)
      end
    end
  elseif type(key) == "number" then
    return self:oldSetStorageValue(key, value)
  end
  return nil
end

player:setStorageValue(1, 100)
print(player:getStorageValue(1))
player:setStorageValue("test", 200, 2) -- would insert (if not already existent) ["test"] = 2, into STORAGE_STRING_MAP
print(player:getStorageValue("test")) -- gives us the table int_key value of "test" from STORAGE_STRING_MAP
```
So in the end we would just make a table dump into a lua file and save the table every server save or so and on startup just a simple load (which it should do automaticly)

rookgaard:
@djarek @ranisalt I've synced branch with master. Is there anything I can do to get this PR accepted?

src/player.cpp
```diff
@@ -654,6 +654,17 @@ void Player::addStorageValue(const uint32_t key, const int32_t value, const bool
 	} else {
 		storageMap.erase(key);
 	}
+	storageStringMap.erase(key);
+}
+
+void Player::addStorageValue(const uint32_t key, const std::string value)
```
djarek, line 660:
```suggestion
void Player::addStorageValue(const uint32_t key, const std::string& value)
```

rookgaard:
Is there anything I can do more on this PR @djarek @ranisalt ?

ranisalt:
I would like someone else to review instead, as I am still unconvinced and strongly against this change, with alternatives proposed along this thread.

Zbizu (commented):
I believe that the drawbacks of having this outweigh the benefits.
I do not recall ever needing to use string as storage value.
Whatever you were needing it for, there is probably a better way to achieve this, like a pattern your strings could be generated with.

This PR has decisions label, but the decision was not made for nearly two years as both the author and the reviewers lost their interest. I do not want to sound rude, but I believe that it is time to close this PR.

PR #2525, opened 2019-02-05
[Fix #2323] Fix skull ticks overflow

### LINKED ISSUE #2323: Bug on SkullTicks
- [ +] This is a bug in the software that resides in this repository, and not a
- [+] This issue is reproducible without changes to the code in this repository

1. Add over 30 frags

In player.cpp and player.h:
```
void Player::checkSkullTicks(int32_t ticks)
{
	int32_t newTicks = skullTicks - ticks;
```
It should be replaced with

```
void Player::checkSkullTicks(int64_t ticks)
{
	int64_t newTicks = skullTicks - ticks;
```

It is because of variable ranges. If every frag is 1 day (86400 * 1000), then 30 frags will cross that range.

aymanm419:
●The reason for the change:

When a player's skull time exceeds 35 days, the skull time overflows as it's stored as milliseconds which has no better value than seconds

●Reproducing Bug

1-Let Player A gets 35 kills(which is equal of 35 days) 
2-35 days calculated in milliseconds -> 3024000000
3-Maximum value for int32 -> 2147483648
4-by calculating the difference = -876516352
5-By next login the player will have 0 skull time
6-Major problem Lied in [here](https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L3828)

●Changes:

Changed skull ticks from millisecond to seconds to avoid overflow and
preserve performance because there's no need for using milliseconds.

●Test:
1-Player A Killed Player B 35 times
2-Player A now has a black skull
3-Player A logged out showing the following information(skull time:3020086 seconds)
4-Player A logged in after few seconds showing the following information(skull time:3020082 seconds)

### DIFF
--- a/data/migrations/22.lua
+++ b/data/migrations/22.lua
@@ -1,3 +1,5 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 23 (Fix skulltime)")
+	db.query("ALTER TABLE players MODIFY skulltime BIGINT(20) NOT NULL DEFAULT '0'")
+	return true
 end
--- a/data/migrations/22.lua
+++ b/data/migrations/22.lua
@@ -1,5 +1,5 @@
 function onUpdateDatabase()
 	print("> Updating database to version 23 (Fix skulltime)")
-	db.query("ALTER TABLE players MODIFY skulltime BIGINT(20) NOT NULL DEFAULT '0'")
+	db.query("ALTER TABLE players MODIFY skulltime BIGINT(20) NOT NULL DEFAULT 0")
 	return true
 end
--- a/schema.sql
+++ b/schema.sql
@@ -44,7 +44,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `lastip` int(10) unsigned NOT NULL DEFAULT '0',
   `save` tinyint(1) NOT NULL DEFAULT '1',
   `skull` tinyint(1) NOT NULL DEFAULT '0',
-  `skulltime` int(11) NOT NULL DEFAULT '0',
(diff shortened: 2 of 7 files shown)


schema.sql
```diff
@@ -44,7 +44,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `lastip` int(10) unsigned NOT NULL DEFAULT '0',
   `save` tinyint(1) NOT NULL DEFAULT '1',
   `skull` tinyint(1) NOT NULL DEFAULT '0',
-  `skulltime` int(11) NOT NULL DEFAULT '0',
+  `skulltime` bigint(20) NOT NULL DEFAULT '0',
```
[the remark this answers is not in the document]
aymanm419: I did another commit to do as you proposed, Have I done it right this time?

aymanm419:
@djarek @marksamman, Can anyone approve/merge this pull request as I think it would be rather valuable to some people.

aymanm419:
Any update on this matter?

data/migrations/22.lua
```diff
@@ -1,3 +1,5 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 23 (Fix skulltime)")
+	db.query("ALTER TABLE players MODIFY skulltime BIGINT(20) NOT NULL DEFAULT '0'")
```
DSpeichert: Since it's a BIGINT, the `DEFAULT 0` should be unquoted (the `0` specifically).

aymanm419: I didn't really know that there was a difference, Thanks for pointing it out!,

PR #2526, opened 2019-02-05
getItemIdByName fix

rookgaard:
Hello. When server is parsing items.xml it uses `std::unordered_multimap` to store item names. That means if different items have the same name (shovel, fish, rope, torch) calling `Items::getItemIdByName` will return only one of these (#2524).

The thing is, on newer protocols it may cause little confusion and here is an example - https://github.com/otland/forgottenserver/blob/master/data/actions/scripts/tools/fishing.lua#L75. A player uses his fishing rod so on newer protocols (like that which support https://github.com/otland/forgottenserver/blob/master/data/items/items.xml) instead of fish (ID 2667) he will get ID 11163 (which also are fish but not moveable/eatable item). My solution is using `equal_range` and I'm sure in all cases returns item which script author had on his mind.

### DIFF
--- a/src/items.cpp
+++ b/src/items.cpp
@@ -1338,10 +1338,15 @@ const ItemType& Items::getItemIdByClientId(uint16_t spriteId) const
 
 uint16_t Items::getItemIdByName(const std::string& name)
 {
-	auto result = nameToItems.find(asLowerCaseString(name));
+	auto ids = nameToItems.equal_range(asLowerCaseString(name));
 
-	if (result == nameToItems.end())
+	if (ids.first == nameToItems.cend())
 		return 0;
 
-	return result->second;
+	NameMap::iterator id;
+	for (NameMap::iterator it = ids.first; it != ids.second; ++it) {
+		id = it;
+	}
+
+	return id->second;
 }


aymanm419:
If you are going to make a linear search through the unordered_multimap for finding the last item, wouldn't it be better to use an ordered_multimap with the help of binary search?

The time complexity of search would drop from O(Items * average length) to O(log(items) * average length) (worst case)

` auto range = --map.upper_bound(word);`
    `return (range->first == word ? range->second : 0)`

aymanm419:
I think that unordered_multimap was used in the first place because there was only one occurrence of such item, but when you have more than an occurrence and you want to fetch the last one,I suggest using an ordered_multimap, It would increase the time of starting up the server slightly but decrease the time of searching for an item greatly, since you would get rid of the linear search in equal_range

The following code shows an example of what I mean
`std::multimap<string,int> map = {{"fish",2667},{"fish",11163}};`
` auto range = --map.upper_bound("fish");`
`return (range->first == "fish" ? range->second : 0);`
This would drop the worst case from
codecogseqn

TheSumm:
My solution is using `equal_range` and I'm sure in all cases returns item which script author had on his mind.
How can you be sure about that? It's not alaway the case that the user intends to retrieve the item with the lower ID.
The real issue here is that `player:addItem("fish", 1)` is used in the script you linked. It's not safe to use that and I don't feel like `Player.addItem` should support adding items by name either.

rookgaard:
Thanks, I've opened another PR with change you suggested #2529

DSpeichert:
The engine should be printing a warning when item lookup by name is occurs where there are multiple items with the same name to choose from and one was arbitrarily chosen.

src/items.cpp
```diff
@@ -1338,10 +1338,15 @@ const ItemType& Items::getItemIdByClientId(uint16_t spriteId) const
...
+	auto ids = nameToItems.equal_range(asLowerCaseString(name));
 
-	if (result == nameToItems.end())
+	if (ids.first == nameToItems.cend())
 		return 0;
 
-	return result->second;
+	NameMap::iterator id;
+	for (NameMap::iterator it = ids.first; it != ids.second; ++it) {
+		id = it;
```
ranisalt: This for essentially does `id = ids.second` anyway, you don't need to iterate. I think you got confused with the code here.

Also, use `auto` for iterator types.

nekiro:
In my opinion we should get rid of using names to addItem or whatever other methods also has the possibility to do that. It's additional problem for no reason at all.

DSpeichert:
I'm closing this because it looks like the proposed change is not currently considered better and the code has unresolved changes requested (abandoned).
We've determined it's not safe to use item names in any case and IDs should be used in scripts.
Feel free to reopen if circumstances change and you still want to go ahead with this.

PR #2527, opened 2019-02-08
Updates paidUntil according to rentPeriod on buying a house

vitorbertolucci:
House rents were charged twice since the money was removed when using !buyhouse and paidUntil was set to 0, so the warnings would start counting right away.


marksamman:
Thanks!

PR #2532, opened 2019-02-21
store player direction

rookgaard:
I thought it would be nice to see the functionality known from 0.4 which kept information about the direction of the player's view after logging out.

### DIFF
--- a/data/migrations/23.lua
+++ b/data/migrations/23.lua
@@ -1,3 +1,5 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 24 (Add player direction)")
+	db.query("ALTER TABLE `players` ADD `direction` int(1) unsigned NOT NULL DEFAULT 2")
+	return true
 end
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -343,7 +343,7 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 	player->defaultOutfit.lookFeet = result->getNumber<uint16_t>("lookfeet");
 	player->defaultOutfit.lookAddons = result->getNumber<uint16_t>("lookaddons");
 	player->currentOutfit = player->defaultOutfit;
-	player->direction = static_cast<Direction> (result->getNumber<uint16_t>("direction"));
+	player->direction = static_cast<Direction> (result->getNumber<uint8_t>("direction"));
 
 	if (g_game.getWorldType() != WORLD_TYPE_PVP_ENFORCED) {
 		const time_t skullSeconds = result->getNumber<time_t>("skulltime") - time(nullptr);
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -233,7 +233,7 @@ bool IOLoginData::loadPlayerById(Player* player, uint32_t id)
 {
 	Database& db = Database::getInstance();
 	std::ostringstream query;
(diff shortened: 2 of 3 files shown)


data/migrations/23.lua
```diff
@@ -1,3 +1,5 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 24 (Add player direction)")
+	db.query("ALTER TABLE `players` ADD `direction` int(1) unsigned NOT NULL DEFAULT 2")
```
rookgaard: `DEFAULT 2` is in this case `DIRECTION_SOUTH` which is default value for creatures in https://github.com/otland/forgottenserver/blob/master/src/creature.h#L512

ranisalt: I would add it right next to outfit data.

src/iologindata.cpp
```diff
@@ -400,6 +400,8 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 		player->skills[i].percent = Player::getPercentLevel(skillTries, nextSkillTries);
 	}
 
+	player->direction = static_cast<Direction> (result->getNumber<uint16_t>("direction"));
```
ranisalt: Isn't 8 bits enough for direction?

data/migrations/23.lua
DSpeichert: I'm against using variables in migrations. We don't expect this one to change but in case it did, it would be quite bad.

rookgaard: I've moved `player->direction` closer outfit data.

src/iologindata.cpp
korejwo: I think could be enough, but I was based on lines like that https://github.com/otland/forgottenserver/blob/master/src/iologindata.cpp#L364 which could be also int8 (which already is in https://github.com/otland/forgottenserver/blob/master/src/position.h#L42). Should I change on `player->loginPosition.z` also?

rookgaard: I think could be enough, but I was based on lines like that https://github.com/otland/forgottenserver/blob/master/src/iologindata.cpp#L364 which could be also int8 (which already is in https://github.com/otland/forgottenserver/blob/master/src/position.h#L42). Should I change on `player->loginPosition.z` also?

src/iologindata.cpp
```diff
@@ -343,6 +343,7 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 	player->defaultOutfit.lookFeet = result->getNumber<uint16_t>("lookfeet");
 	player->defaultOutfit.lookAddons = result->getNumber<uint16_t>("lookaddons");
 	player->currentOutfit = player->defaultOutfit;
+	player->direction = static_cast<Direction> (result->getNumber<uint16_t>("direction"));
```
GustavCrypter: Direction is **`uint8`** not **`uint16`**.

rookgaard: As I said in post earlier I was based on other values just like `lookHead`, `lookBody`, `lookLegs`, `lookFeet` and `lookAddons` which you have 2 lines above. All of these are `uint8_t`, but still from DB is casted as `uint16_t`.

Anyway, I've changed `direction` to `uint8_t`.

marksamman: There may be issues with using uint8_t because it's an unsigned char. It's been a while since I worked in C++, but doing a lexical cast from a char type may turn it into an ASCII character rather than digits. The memory use you're saving on the stack to risk introducing this bug is not even worth having a discussion about, just leave it as uint16_t.

rookgaard: Ok, so I reverted my changes. That was my thoughts first that it might not work.

src/iologindata.cpp
rookgaard: As Mark said here https://github.com/otland/forgottenserver/pull/2532#discussion_r260477881 it should remain int16.

data/migrations/23.lua
rookgaard: @ranisalt How about that migration? It's ok to have numeric value as @DSpeichert said or should I change to `DIRECTION_SOUTH`?

ranisalt: I agree with @DSpeichert

PR #2534, opened 2019-02-24
Update poison_field_rune.lua

GoularPink:


data/spells/scripts/attack/poison_field_rune.lua
```diff
@@ -1,6 +1,6 @@
 local combat = Combat()
 combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_EARTHDAMAGE)
-combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ANI_ENERGY)
+combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ME_GREEN_RINGS)
```
WibbenZ: This isen't correct, it will use the "throwing knife" effect.
You can't mix CONST_ME and CONST_ANI.

The correct fix would be to use CONST_ANI_POISON

raymondtfr: The effect ``CONST_ANI_ENERGY`` (doesn't make sense - energy, poison - , I know) is the one as used in vanilla already (retested right now), so no reason for updating this, other than being a cosmetic change, tbh .

GoularPink:
It does not make sense, poison shoot energy, but in the tibia it is anyway, I thought it was wrong, this pull is not correct

http://prntscr.com/mpw0m3

Issue #2344, opened 2017-09-21
Decaying item stops decaying when you move it


nekiro:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Make item decay, for example make new item in items.xml
2. item:decay()
3. move it to any equipment slot or any other position that it is in currently

This is the item I used:
```
<item id="28826" article="a" name="diamond arrow">
	<attribute key="weight" value="80" />
	<attribute key="slotType" value="ammo" />
	<attribute key="attack" value="37" />
	<attribute key="weaponType" value="ammunition" />
	<attribute key="ammoType" value="arrow" />
	<attribute key="shootType" value="diamondarrow" />
	<attribute key="maxHitChance" value="95" />
	<attribute key="decayTo" value="0" />
	<attribute key="duration" value="3600" />
	<attribute key="showduration" value="1" />
</item>
```

### Expected behaviour
Item should still be decaying

### Actual behaviour
Decaying is stopped.

infister:
Is it in movements.xml with equip/deequip ?

thexamx:
No, it does meant to be. This item should decay from the start, when created.

infister:
I could not recreate this in my server. I was using an item: 

	<item id="2090" article="a" name="crystal key">
		<attribute key="type" value="key"/>
		<attribute key="weight" value="100"/>
        <attribute key="decayTo" value="0"/>
		<attribute key="duration" value="3600"/>
        <attribute key="showduration" value="1"/>
        <attribute key="slotType" value="ammo" />
	</item>

Item was created by NPC, decayed and I was switching it to ammo slot and off few times, logged in or out... It is not in movements too. And it still kept decaying with exact timer. Maybe when using it with a bow as weapon disables decaying?

thexamx:
Maybe because its not actually an ammunition? Also I have it registered in weapons.xml, because its ammunition.

infister:
Paste your weapons.xml entry please

thexamx:
https://github.com/otland/forgottenserver/blob/master/data/weapons/weapons.xml#L702
It's not different from any of these.

infister:
Paste yours entry... it’s important what action you have set there

infister:
[...]

PR #2541, opened 2019-03-12
Fix crash when removing conditions in pre-death script

### LINKED ISSUE #2024: [Crash] Creature::executeConditions

jo3bingham:
Fixes #2024. I could 100% repeatedly reproduce the issue with the script @ninjalulz supplied. With this fix I can no longer reproduce it.

Please review, @Znote.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1186,12 +1186,10 @@ void Creature::removeCondition(ConditionType_t type, bool force/* = false*/)
 			}
 		}
 
-		it = conditions.erase(it);
-
 		condition->endCondition(this);
-		delete condition;
-
 		onEndCondition(type);
+		toReleaseConditions.push_back(condition);
+		++it;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


marksamman:
Any idea what's causing the crash? Is there some use-after-free where the deleted condition is still referenced somewhere else? There's probably a better way to fix this without the additional memory overhead for every creature, but I'm fine with merging this until there's a better patch. LGTM

jo3bingham:
@marksamman You can ready my outline for what's causing the crash here: https://github.com/otland/forgottenserver/issues/2024#issuecomment-471744759

andersonfaaria:
@Znote I think there's one more that's even more critical than #2344, that is #1753. Once they are fixed we can go to 1.3 milestone. But personally I think the community should focus their efforts in finishing half-made PR's like questlog in Lua and then focus on both otclient and moving things to Lua.

PR #2544, opened 2019-03-15
Guildhalls support

nekiro:
In this pull request I'm trying to add missing guild halls support.

What does this pr add?
- now every guild can have it's own residence, which should be guild hall, but isn't required, it's up you to check that, keep in mind that you cannot set it from lua script, this should be done by AAC (because AAC creates new guild entry in database)
- you can now call house:isGuildHall() to check whether it is guild hall or not
- loading houses from houses.xml, will now parse missing "guildhall" attribute
- houses now have one additional attribute called guildId, this attribute stores guild that owns this house, this should be used only with guild halls
- made tfs load ownerid from database to determine who is the guild leader, this will fail if we have more than one leader in a guild, but currently the only other option is to check guildrank if its number 3, because 3 is leader.

Probably forgot about something as always, so let me know if anything is missing, btw. I didn't test this code.

New enums:
HOUSE_TYPE_NORMAL
HOUSE_TYPE_GUILDHALL

New lua methods:
- guild:setHouseId() -- sets guild hall id
- guild:getHouseId() -- gets guild hall id
- house:getType() -- returns house type
- house:setGuildId() -- sets guild id
- house:getGuildId() -- gets guild id


data/migrations/24.lua
```diff
@@ -1,3 +1,7 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 25 (Guildhall support)")
+	db.query("ALTER TABLE `guilds` ADD `house_id` int(11) NOT NULL DEFAULT '0'")
+	db.query("ALTER TABLE `houses` ADD `guild_id` int(11) NOT NULL DEFAULT '0'")
```
DSpeichert: A cross-reference on both ends is just looking for trouble. Which one wins when they contradict?
It's SQL, we should need just one. Can a house by owned by both "owner" and "guild_id"? If not, it's easier to catch that by using `guild_id` on the `houses` table.

src/ioguild.cpp
```diff
@@ -30,6 +30,8 @@ Guild* IOGuild::loadGuild(uint32_t guildId)
 	query << "SELECT `name` FROM `guilds` WHERE `id` = " << guildId;
```
Znote: Does this even work? you set houseId and OwnerId in line 33 and 34. But the query don't retrieve it from the SQL query?

This should solve it, it also takes @DSpeichert requested change into consideration:
```suggestion
	query << "SELECT `g`.`ownerid`, `g`.`name`, CASE WHEN `h`.`id` IS NULL THEN 0 ELSE `h`.`id` END AS `house_id` FROM `guilds` AS `g` LEFT JOIN `houses` AS `h` ON `g`.`id` = `h`.`guild_id` WHERE `g`.`id` = " << guildId << " LIMIT 1";
```

Beautified representation:
```sql
SELECT 
	`g`.`ownerid`, 
	`g`.`name`, 
	CASE WHEN `h`.`id` IS NULL 
		THEN 0 
		ELSE `h`.`id` 
	END AS `house_id` 

FROM `guilds` AS `g` 

LEFT JOIN `houses` AS `h` 
	ON `g`.`id` = `h`.`guild_id` 

WHERE `g`.`id` = " << guildId << " 
LIMIT 1
```
The case statement converts the output from "null" to 0 if no house_id was matched. Also renamed houses.id to house_id to avoid confusion down the line. 

Limit 1 is added to ensure that we only get 1 row back, could be a db bug where a guild has two houses, that would be problematic since its not accounted for.

data/migrations/24.lua
```diff
@@ -1,3 +1,7 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 25 (Guildhall support)")
+	db.query("ALTER TABLE `guilds` ADD `house_id` int(11) NOT NULL DEFAULT '0'")
```
Znote:
```suggestion
```
See my other suggested change in ioguild.cpp on how to load this with a left join instead.

src/ioguild.cpp
DSpeichert: Well, why not allow guilds to have more than one house? There's nothing stopping players from having multiple houses - nothing other than Lua.

Znote: Trying not to overcomplicate things and get a functional version merged asap. To support more guildhalls alot of changes has to be done in this pull request(right?), not just my suggested query change.

DSpeichert: From a pure SQL standpoint, there is nothing preventing that in future. We don't need it right now.

src/house.h
```diff
@@ -110,6 +110,11 @@ enum AccessHouseLevel_t {
 	HOUSE_OWNER = 3,
 };
 
+enum HouseType_t {
+	HOUSE_TYPE_NORMAL = 0,
+	HOUSE_TYPE_GUILDHALL = 1,
+};
```
ranisalt: I don't think you need to explicit the values here.

src/house.h
```diff
@@ -200,6 +205,19 @@ class House
...
+		}
+		HouseType_t getType() const {
+			return type;
+		}
+		void setGuildId(uint32_t guildId) {
+			this->guildId = guildId;
+		}
+		uint32_t getGuildId() const {
+			return guildId;
+		}
```
ranisalt: What about making those public attributes if getters/setters are no ops?

PR #2545, opened 2019-03-17
Fix #2410

### LINKED ISSUE #2410: updateIdleStatus invisible condition
I noticed that if, let's say a stalker, goes invisible with a relative high chance/rate (or even just longer duration) it will never become idle due to 

`if (conditions.empty()) {` 
located in Monster::updateIdleStatus().

Anyone have a creative solution to this "problem"?

nekiro:
Fix #2410


Znote:
Did travis github bot hang? It says the build is in progress here, but when clicking on details it says it passed.

djarek:
Travis was bought by a company that wants to milk all the money it can from it, so they fired all the productive people and the infrastructure is dying :).

src/monster.cpp
```diff
@@ -659,10 +659,19 @@ void Monster::setIdle(bool idle)
 void Monster::updateIdleStatus()
 {
 	bool idle = false;
-
-	if (conditions.empty()) {
-		if (!isSummon() && targetList.empty()) {
+	if (!isSummon()) {
+		if (conditions.empty() && targetList.empty()) {
 			idle = true;
+		} else if (!conditions.empty()) {
```
slavidodo: ``if (targetList.empty())`` sounds more reasonable.

nekiro: @slavidodo  I assume you meant else instead of that check at all, why would I check if targlist is empty if then monster could have no conditions at all.

nekiro: but I wouldnt change that at all, but probably first check should be or instead of and

slavidodo: If targetList is empty and condition list is empty the first case should have been applied, that means you are probably iterating through conditions when targetList is actually not empty and then idle status should never be modified.

Znote:
> I think it should be reliable that a condition ticks down regardless of it being aggressive or not.
> IMO the real problem is that monsters cast defensive spells even though they have no creatures in their target list: https://github.com/otland/forgottenserver/blob/master/src/monster.cpp#L900

What about etc summoned monks? Do they need a target for their healing to trigger?

TheSumm:
I doublechecked and it seems like the behaviour is different depending on the type of spell.
Invisibility is only cast if a monster has a target which also applies if the monster is a summon.
Healing spells are cast as long as the monster is not idle, thus summons will heal even if you are not fighting.

DSpeichert:
So what does that mean? Do we not need this change at all?

joseluis2g:
Why would you close this?

Znote:
@slavidodo Your review is requesting changes, could you clarify what has to be changed in order to get your approval?

src/monster.cpp
slavidodo: @Znote In the first case it checks if both lists are empty (targets and conditions).
the second case would only apply if targets **OR** conditions is not empty, so what he did is checking if conditions is not empty, but, both cases might actually meet. It's logical to check if targets is empty to make sure that we don't iterate over condition list when there are targets that will always stop the idle status.

Leo32onGIT: @nekiro My hunch is that is was intention as per: https://github.com/otland/forgottenserver/issues/2410#issuecomment-357823797

slavidodo: that's exactly what I mean, you cant set the monster to be idle when target list is not actually empty!

nekiro: please suggest the whole code snippet on how it should be written, my small brain cant handle what you say somehow.

nekiro: Oh I think I got it, because second check always sets idle status. I think doing like 

> if (hasAggressiveCondition) { idle = false; } 
would be fine right?

nekiro:
I believe its ready to be merged.

Issue #2248, opened 2017-05-11
[Question] New groups


vankk:
We should set more groups to _groups.xml_ to make more easier to the AAC creators, as admin as well. 

Currently in our datapack we only have player, gamemaster and god.

**We should a few groups for example:**
- Tutor
- Senior Tutor
- Community Manager

Also in the _groups.xml_, we should provide a comment that have a flag calculator to if someone wants modify the flags.

What you guys think?

Regards,
vankk.

jo3bingham:
In the [Roadmap](https://github.com/otland/forgottenserver/wiki/Roadmap), there is a bullet for making flags easier to maintain:
> XML tag for each flag in groups.xml with value 0 or 1 to make it more maintainable instead of a single int representing the value for all flags

As for your suggestion for new groups, that's fine as long as there's a general agreement on what flags should be set for each group. Otherwise, if each group has the same flags as players and we leave it up to server owners to decide what flags they want to set, then there's no point in adding them.

pzaj2:
@jo3bingham 

Just a comment regarding the roadmap task, wouldn't it be better to have empty tags for these flags? like:
```xml
<group ...>
    <flags>
        <cannotattackmonsters />
    </flags>
</group>
```

Wouldn't that be better? We could omit unused flags in particular group, making the file more concise. I actually started doing this myself already. So it would be good to know if that form of assigning flags is acceptable? I have this pretty much done, including BC by checking `flags` attribute first. Would probably require some refactoring, but I wasn't sure if any of boost::assign::map_list_of and high order macros are allowed to map PlayerFlag enum to std::map? (so that it can be done in a loop instead of using multiple if statements and allow adding new flags more easily).

Well, I will gladly handle this, is there any way you can "assign" me to this task or is it only for TFS developers?

WibbenZ:
@pzaj2 https://github.com/otland/forgottenserver/issues/1394

pzaj2:
@WibbenZ 
Why didn't I see that before! I assume this means "no" then! Fair enough then, since I've done it already, I will just release it on Otland, perhaps someone will make use of it anyway :)

Thanks for pointing me to this issue, it's quite old, didn't see it :+1:

PR #2546, opened 2019-03-18
Changing the way we parse group flags

nekiro:
Ideas taken from #2248, #1394

### DIFF
--- a/src/groups.cpp
+++ b/src/groups.cpp
@@ -81,8 +81,7 @@ bool Groups::load()
 		group.access = groupNode.attribute("access").as_bool();
 		group.maxDepotItems = pugi::cast<uint32_t>(groupNode.attribute("maxdepotitems").value());
 		group.maxVipEntries = pugi::cast<uint32_t>(groupNode.attribute("maxvipentries").value());
-		
-		group.flags = 0;
+		group.flags = pugi::cast<uint64_t>(groupNode.attribute("flags").value());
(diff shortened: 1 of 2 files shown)


DSpeichert:
It looks ugly. But I can't think of a better way in XML.
Maybe get rid of `<flags>`? Nothing else is a child of the "group" anyway.

ranisalt:
Maybe we could add an `inherit` option to copy from another group defaults.

nekiro:
@ranisalt that's a good idea, do we agree on that?

src/groups.cpp
```diff
@@ -37,10 +78,25 @@ bool Groups::load()
 		Group group;
 		group.id = pugi::cast<uint32_t>(groupNode.attribute("id").value());
 		group.name = groupNode.attribute("name").as_string();
-		group.flags = pugi::cast<uint64_t>(groupNode.attribute("flags").value());
 		group.access = groupNode.attribute("access").as_bool();
 		group.maxDepotItems = pugi::cast<uint32_t>(groupNode.attribute("maxdepotitems").value());
 		group.maxVipEntries = pugi::cast<uint32_t>(groupNode.attribute("maxvipentries").value());
+		
+		group.flags = 0;
```
marksamman: Remove this line and restore `group.flags = pugi::cast<uint64_t>(groupNode.attribute("flags").value());` for backwards compatibility.

PR #2547, opened 2019-03-20
Spells updates

raymondtfr:
- Added two Combat metatable functions to Core 
   - Use ``Combat:getPositions`` on ``Player:addPartyCondition``
   - Use ``Combat:getTargets`` on some spells
- Fixed some spells, and their groups and moved them to their proper folder
- Renamed some variables, and replaced some condition parameters with proper ones, and also fixed a typo on custom spell *Combustion*


data/spells/scripts/support/find_person.lua
```diff
@@ -63,12 +63,12 @@ function onCastSpell(creature, variant)
...
 	local distance = maxPositionDifference < 5 and DISTANCE_BESIDE or maxPositionDifference < 101 and DISTANCE_CLOSE or maxPositionDifference < 275 and DISTANCE_FAR or DISTANCE_VERYFAR
-	local message = messages[distance][level] or messages[distance]
+	local description = descriptions[distance][level] or descriptions[distance]
 	if distance ~= DISTANCE_BESIDE then
-		message = message .. " " .. directions[direction]
+		description = description .. " " .. directions[direction]
 	end
 
-	creature:sendTextMessage(MESSAGE_INFO_DESCR, target:getName() .. " " .. message .. ".")
+	creature:sendTextMessage(MESSAGE_INFO_DESCR, target:getName() .. " " .. description .. ".")
```
DSpeichert: This looks like a cosmetic change without any obvious reason. It's best to avoid these to not clutter history of editing this file.

raymondtfr: The reason for this change is that the variable ``message`` in ``target:getName() .. " " .. message .. "."`` is not the message itself, but a part of it - the description based on distance and floor levels - the whole sentence ``target:getName() .. " " .. description .. "."`` is what composes the **message**.

PR #2552, opened 2019-03-25
Unnecessary verification

souzajunior:
In this case, if the condition dosen't do nothing, we can negate that.


kornholi:
why? it just makes the logic harder to follow

DSpeichert:
Looks like subjective change, potentially "change for change".
There should be "I don't care" review option.

PR #2553, opened 2019-03-25
Player set/get cannotMove

### LINKED ISSUE #775: onMove event?

rookgaard:
Hello. I would like to present another nice functionality which went lost in 0.X series - possibility to freeze/unfreeze player. It's nice to have feature which can be used in many quests or systems like transport by wagons.

Fixes #775


src/creature.cpp
```diff
@@ -249,6 +249,11 @@ bool Creature::getNextStep(Direction& dir, uint32_t&)
 
 void Creature::startAutoWalk(const std::forward_list<Direction>& listDir)
 {
+	if (getPlayer() && getPlayer()->isNoMove()) {
```
DSpeichert: Rename `isNoMove()` to `canMove()` and use that throughout this patch (should default to true obviously).

korejwo: But I wanted to call `sendCancelWalk` only if player is freezed. So maybe it could be `canNoMove` instead of `isNoMove`?

rookgaard: But I wanted to call `sendCancelWalk` only if player is freezed. So maybe it could be `canNoMove` instead of `isNoMove`?

nekiro:
Please don’t use _variables. Also this is really bad, because if you freeze player and he will continously spam movement it will kick him, because he will exceed the packet per second limit. (Happens because of sendCancelWalk)

src/creature.cpp
DSpeichert: If `cannotMove` is already used in create.h, then it should be used for player as well.

rookgaard: Nomenclature changed.

rookgaard:
@nekiro I've removed `_` prefix. Anyway, what is difference between
```
		void setNoMove(bool _cannotMove)
		{
			cannotMove = _cannotMove;
			cancelNextWalk = true;
		}
```
and
```
		void setNoMove(bool cannotMove)
		{
			cannotMove = cannotMove;
			cancelNextWalk = true;
		}
```
if it's only function variable? Plus, what does this have to do with the connection limits if they apply only to packets sent by player not by server? Isn't the same situation when the player is standing at the door of the house to which he can not enter and keeps the arrow down while still receiving the message "You are not invited."?

nekiro:
Did you check if that happens before you wrote message?

rookgaard:
@nekiro Nope, but I removed `_` as you suggested. I wouldn't call myself a C++ programmer, so some of my questions may be ridiculous.

src/creature.h
```diff
@@ -408,6 +408,14 @@ class Creature : virtual public Thing
 		void setUseDefense(bool useDefense) {
 			canUseDefense = useDefense;
 		}
+		void setNoMove(bool canNotMove)
```
DSpeichert: `canNotMove` isn't correct, `cannot` is a single word.

by the way, I can't find `cannotMove` in creature.h like you said is there.

rookgaard: I meant field `bool cannotMove = false;`

Anyway, I've changed again function naming. How would you rate changes now?

ranisalt:
I'd call the variable `canMove`, default to `true` and call every method get/set `CanMove` instead, for the sake of consistency.

rookgaard:
Ok @ranisalt, I've changed to set/get, how would you rate changes now?

ranisalt:
LGTM, though I'd like to hear more from @nekiro point.

src/creature.cpp
```diff
@@ -249,6 +249,11 @@ bool Creature::getNextStep(Direction& dir, uint32_t&)
 
 void Creature::startAutoWalk(const std::forward_list<Direction>& listDir)
 {
+	if (getPlayer() && !getPlayer()->getCanMove()) {
```
nekiro:
```
Player* player = getPlayer();
if (player && player->getCanMove()) {
```
same with sendCancelWalk

src/creature.h
```diff
@@ -408,6 +408,14 @@ class Creature : virtual public Thing
 		void setUseDefense(bool useDefense) {
 			canUseDefense = useDefense;
 		}
+		void setCanMove(bool move)
+		{
+			canMove = move;
+			cancelNextWalk = true;
```
nekiro: if we set canMove to true, do we still cancelNextWalk?

soul4soul: I too was wondering about cancelNextWalk  always being set to true in this function. What is the reason for the code not being this or similar
```suggestion
		void setCanMove(bool move)
		{
			canMove = move;
			cancelNextWalk = !move;
```

EPuncker:
> I would really like to see how it behaves in-game, could you make a gif spamming move, when you are set to cannot move?

I've just tested it and indeed it kicks the player because of the maxPacketsPerSecond setting

DSpeichert:
Should a player be exempt from maxPacketsPerSecond kick when they are not able to walk?

nekiro:
> Should a player be exempt from maxPacketsPerSecond kick when they are not able to walk?

No, that shouldnt be done fully server side in the first place.

DSpeichert:
We can't really do it client-side, so are we merging it as-is or adding some workaround?

EPuncker:
closes #775 if merged

nekiro:
Do we even need it? I dont think thats something that should be in tfs repo as its mainly client sided.

rookgaard:
Many servers still use the Cipsoft client to play, so due to the fact that it is quite difficult to edit, in my opinion it is easier to "block" it server-side.

rookgaard:
@DSpeichert Is there anything else I need to change in my request?

src/creature.h
DSpeichert: I'm still curious about this.

DSpeichert (dismissed):
Looking at `canMove` and `setMove`, I think it would be better to use terms such as:
`isMovementBlocked`: true/false
`setMovementBlocked`: true/false

That, IMHO, is less ambiguous.

andersonfaaria:
what's pending for this to be merged so we can close #775

rookgaard:
@DSpeichert @nekiro Can you look at it in free time? I've committed changes you requested.

src/creature.h
```diff
@@ -408,6 +408,14 @@ class Creature : virtual public Thing
 		void setUseDefense(bool useDefense) {
 			canUseDefense = useDefense;
 		}
+		void setMovementBlocked(bool state)
+		{
```
EPuncker, line 412: code styling
```
void {
{
```

src/luascript.cpp
```diff
@@ -6923,6 +6925,14 @@ int LuaScriptInterface::luaCreatureIsHealthHidden(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaCreatureIsMovementBlocked(lua_State* L)
+{
+	// creature:isMovementBlocked()
+	const Creature* creature = getUserdata<const Creature>(L, 1);
+	creature ? pushBoolean(L, creature->isMovementBlocked()) : lua_pushnil(L);
```
EPuncker, line 6932: code style:

```
if (creature) {
		pushBoolean(L, creature->isMovementBlocked());
	} else {
		lua_pushnil(L);
	}
```

EPuncker:
I hope we could get this merged soon, from time to time people are always requesting this function on the forum

src/creature.h
Znote: My guess is to fix desync issues mid-action, this is called to toggle the states, but no matter which direction it toggles to, you dont want to give the user half a sqm headstart as it may create odd client behavior. So even if you just got permission to walk, the character pos gets a cancel byte one last time to set the timings straight.

Znote (commented):
Pending patch to PR @rookgaard has to merge: (in his repo) https://github.com/rookgaard/forgottenserver/pull/1

Znote (approved):

Znote:
> Do we even need it? I dont think thats something that should be in tfs repo as its mainly client sided.

Absolutely, lots of people has requested this. I also remember being rather desperate for someone to make something like this back in the day when I was hosting servers. This has been a frustration solved by uglier hacks all the time over the past 10 years.

src/creature.h
nekiro: @Znote but if player is already movement blocked then he cannot walk, so what desynchro?

rookgaard:
Is there anything I can do more on this PR @DSpeichert ?

EPuncker (changes requested):
to me there is just a small request, the addition of the entries in compat.lua for it to be compatible with the old version:

```
getCreatureNoMove(cid)
doCreatureSetNoMove(cid, block)
```

rookgaard:
@EPuncker I've added functions to `compat.lua`.

EPuncker (approved):

Znote (approved):

Issue #2550, opened 2019-03-25
Fix #2541 Client Crash


gregorecruzeiro:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Attack a monster, Warlock for example, after a short time the client crash
2.
3.

### Environment

Fix crash when removing conditions in pre-death script

After this fix, when dueling with monsters, after a certain time, the client about a crash, defaced the update and the crash on the client stopped.

logs client.
https://uploaddeimagens.com.br/imagens/erro_1-png-4ecbc314-d266-47e4-a060-576b2daab3d1

https://uploaddeimagens.com.br/imagens/erro_2-png-29f34445-d081-4046-9c67-d1822fa8edd4

https://uploaddeimagens.com.br/imagens/erro_3-png-f60ecc8e-8cec-4409-8860-75a2e27d03af

raymondtfr:
I can confirm this.

Znote:
#2541 @jo3bingham

jo3bingham:
Will have a look when I'm home from work this evening. The crash reason, and the creature tested on, implies this probably has to do with the invisibility condition.

raymondtfr:
Unfortunately, it still crashes @jo3bingham .

jo3bingham:
@raymondtfr thanks for testing. I'm struggling to see how my changes are causing the issue as all it does is change how the condition is removed from the array and deleted from memory. The original functionality of ending the condition is still there. I'll see what I can come up with.

nekiro:
Why isnt the change implemented into Game::cleanup? Same with toreleaseconditions vector

jo3bingham:
@nekiro because it doesn't need to be? Conditions are cleaned up in the `Creature` deconstructor.

nekiro:
I didnt mean it needs to be there, isnt it better to have all in one place?

nekiro:
Also if creatures dies with „pending” conditions, wouldnt that create memory leak?

jo3bingham:
> I didnt mean it needs to be there, isnt it better to have all in one place?

It just doesn't make sense to have it there. Conditions are tied to creature logic, not game logic.

> Also if creatures dies (creature object is deleted) with „pending” to delete conditions, wouldnt that create memory leak?

[...]

Issue #2549, opened 2019-03-23
Stealth ring bug


gregorecruzeiro:
### Before creating an issue, please ensure:
- [X ] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [X] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. When removing the stealth ring ID 2202 from the slot (ring) the player continues in the invisible condition, however the mosntros can attack him
2.
3.

### Expected behaviour

Back outfit before using the rin
### Actual behaviour

The player continues with invisible outfit

jo3bingham:
This is probably related to #2550.

PR #2554, opened 2019-03-26
Fix #2024, #2549, and #2550

### LINKED ISSUE #2024: [Crash] Creature::executeConditions

jo3bingham:
My fix for #2024 (#2541) introduced issues with conditions (#2549 and #2550). Commit 9c2f11c reverts my original fix, and commit e21523e implements a better way to resolve the original issue while fixing the new ones.

Basically, before executing conditions, the current condition list is copied to a new one, and that new list is iterated through instead. Checks are done against the original condition list to ensure the current iteration is valid.

@gregorecruzeiro and @raymondtfr, please test and confirm this fixes the mentioned issues.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1280,20 +1280,21 @@ Condition* Creature::getCondition(ConditionType_t type, ConditionId_t conditionI
 
 void Creature::executeConditions(uint32_t interval)
 {
-	auto it = conditions.begin(), end = conditions.end();
-	while (it != end) {
-		Condition* condition = *it;
+	ConditionList tempConditions{ conditions };
+	for (Condition* condition : tempConditions) {
+		auto it = std::find(conditions.begin(), conditions.end(), condition);
+		if (it == conditions.end()) {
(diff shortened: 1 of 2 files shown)


gregorecruzeiro:
@jo3bingham 
I tested it and it really solved the problem

raymondtfr:
I didn't get any more crashes @jo3bingham .

PR #2557, opened 2019-04-01
Renaming misleading variable names in playerAcceptTrade

nekiro:
Just a "quality of life update", renamed tradeItem1 and tradeItem2 to be less confusing.

Also I see no reason for these two variables at all, why dont we use player->tradeItem and tradePartner->tradeItem instead?

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -2637,10 +2637,10 @@ void Game::playerAcceptTrade(uint32_t playerId)
 	player->setTradeState(TRADE_ACCEPT);
 
 	if (tradePartner->getTradeState() == TRADE_ACCEPT) {
-		Item* tradeItem1 = player->tradeItem;
-		Item* tradeItem2 = tradePartner->tradeItem;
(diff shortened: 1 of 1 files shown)


djarek:
FYI, looks like you've got misconfigured email address in your git configuration, which causes github to be unable to attribute commits to you.

src/game.cpp
```diff
@@ -2637,50 +2637,42 @@ void Game::playerAcceptTrade(uint32_t playerId)
 	player->setTradeState(TRADE_ACCEPT);
 
 	if (tradePartner->getTradeState() == TRADE_ACCEPT) {
-		Item* tradeItem1 = player->tradeItem;
-		Item* tradeItem2 = tradePartner->tradeItem;
+		Item* playerTradeItem = player->tradeItem;
+		Item* tradePartnerItem = tradePartner->tradeItem;
```
ranisalt: This is not consistent. Either rename the first to `tradePlayerItem` (which is ugly) or the second to `partnerTradeItem` (:+1:)

Issue #2555, opened 2019-03-27
Revscriptsys


EvilHero90:
I didn't find this to be on a TODO list even not for 2.0
I'm wondering if there is still any kind of interest in it.
Summ and I have been working on it years ago.
It's an easy drag and drop system where the lua files get load on startup (or on occasional load)
It doesn't break backwards compatibility and works normaly alongside the other xml stuff.
It would give the ability to write long quests and such easily in one file for organisation manner.
I'll work towards a PR if there's some real interest, else the time ammount to rebase this on newest sources wouldn't be worth it.

It invokes metatables for:
* Action()
* CreatureEvent()
* GlobalEvent()
* MoveEvent()
* TalkAction()
* Weapon()
* Spell()

with their corresponding functions to be able to set them up properly.

Supports:
- [x] Actions
- [x] Creaturescripts
- [x] Globalevents
- [x] Movements
- [x] Talkactions
- [x] Monsters
- [x] Weapons (as of #2571)
- [x] Spells (as of #2584)

Delayed for 1.4/2.0:
- [ ] NPC (Extensive change and lots of testing required)

Here are some working examples:

**Action**
```lua
local shovel = Action()

local holes = {468, 481, 483}
function shovel.onUse(player, item, fromPosition, target, toPosition, isHotkey)
	if toPosition.x == CONTAINER_POSITION then
		return false
	end

	local tile = Tile(toPosition)
	if not tile then
		return false
	end

	local ground = tile:getGround()
	if not ground then
		return false
	end

	local groundId = ground:getId()
	if isInArray(holes, groundId) then
		ground:transform(groundId + 1)
		ground:decay()

		toPosition.z = toPosition.z + 1
		tile:relocateTo(toPosition)
	elseif groundId == 231 then
		local randomValue = math.random(1, 100)
		if randomValue == 1 then
			Game.createItem(2159, 1, toPosition)
		elseif randomValue > 95 then
			Game.createMonster("Scarab", toPosition)
		end
		toPosition:sendMagicEffect(CONST_ME_POFF)
	else
		return false
	end

	return true
end

shovel:id(2554)
shovel:register()
```
**Globalevents**
```lua
local shutdownAtServerSave = false
local cleanMapAtServerSave = false

local function serverSave()
	if shutdownAtServerSave then
		Game.setGameState(GAME_STATE_SHUTDOWN)
	else
		Game.setGameState(GAME_STATE_CLOSED)

		if cleanMapAtServerSave then
[...]

PR #2558, opened 2019-04-05
revscriptsys

EvilHero90:
creates a new folder called 'scripts' in the data folder.

lua files placed in the 'scripts' or sub folders will automaticly get registered on load with the right setup ofcourse.

Registering happens in the lua file itself, doesn't require xml settings.

lua files can be deactivated by placing a '#' infront of the filename ex: '#position.lua' it wont get loaded then.

console
It provides the normal lua interface error system

Currently supports:
- [x] Actions
- [x] Creaturescripts
- [x] Globalevents
- [x] Movements
- [x] Talkactions
- [x] Monsters

For a longer explanation and examples #2555 

Co-Authored-By: Sam <thesumm@users.noreply.github.com>

### DIFF
--- a/data/scripts/actions/spellbook.lua
+++ b/data/scripts/actions/spellbook.lua
@@ -1,0 +1,36 @@
+local spellbook = Action()
+
+function spellbook.onUse(player, item, fromPosition, target, toPosition, isHotkey)
+	local text = {}
+	local spells = {}
+	for _, spell in ipairs(player:getInstantSpells()) do
+		if spell.level ~= 0 then
+			if spell.manapercent > 0 then
+				spell.mana = spell.manapercent .. "%"
+			end
+			spells[#spells + 1] = spell
+		end
+	end
+
+	table.sort(spells, function(a, b) return a.level < b.level end)
+
+	local prevLevel = -1
+	for i, spell in ipairs(spells) do
+		if prevLevel ~= spell.level then
+			if i == 1 then
+				text[#text == nil and 1 or #text+1] = "Spells for Level "
+			else
+				text[#text+1] = "\nSpells for Level "
+			end
+			text[#text+1] = spell.level .. "\n"
+			prevLevel = spell.level
+		end
+		text[#text+1] = spell.words .. " - " .. spell.name .. " : " .. spell.mana .. "\n"
+	end
+
+	player:showTextDialog(item:getId(), table.concat(text))
+	return true
+end
+
+spellbook:id(2175, 6120, 8900, 8901, 8902, 8903, 8904, 8918)
+spellbook:register()
--- a/data/scripts/talkactions/position.lua
+++ b/data/scripts/talkactions/position.lua
@@ -1,0 +1,15 @@
+local talk = TalkAction("/pos")
+
+function talk.onSay(player, words, param)
+	if player:getGroup():getAccess() and param ~= "" then
+		local split = param:split(",")
+		player:teleportTo(Position(split[1], split[2], split[3]))
+	else
+		local position = player:getPosition()
+		player:sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, "Your current position is: " .. position.x .. ", " .. position.y .. ", " .. position.z .. ".")
+	end
+	return false
+end
+
+talk:separator(" ")
+talk:register()
--- a/src/actions.h
+++ b/src/actions.h
@@ -34,11 +34,53 @@ class Action : public Event
 		explicit Action(LuaScriptInterface* interface);
 
 		bool configureEvent(const pugi::xml_node& node) override;
-		bool loadFunction(const pugi::xml_attribute& attr) override;
+		bool loadFunction(const pugi::xml_attribute& attr, bool isScripted) override;
 
 		//scripting
 		virtual bool executeUse(Player* player, Item* item, const Position& fromPosition,
 			Thing* target, const Position& toPosition, bool isHotkey);
+
+		bool getAllowFarUse() const {
+			return allowFarUse;
+		}
+		void setAllowFarUse(bool v) {
+			allowFarUse = v;
+		}
+
+		bool getCheckLineOfSight() const {
+			return checkLineOfSight;
+		}
+		void setCheckLineOfSight(bool v) {
+			checkLineOfSight = v;
+		}
+
+		bool getCheckFloor() const {
+			return checkFloor;
+		}
+		void setCheckFloor(bool v) {
+			checkFloor = v;
+		}
+
+		std::vector<uint16_t> getItemIdRange() {
+			return ids;
+		}
+		void addItemId(uint16_t id) {
+			ids.emplace_back(id);
+		}
+
+		std::vector<uint16_t> getUniqueIdRange() {
+			return uids;
+		}
+		void addUniqueId(uint16_t id) {
+			uids.emplace_back(id);
+		}
+
+		std::vector<uint16_t> getActionIdRange() {
+			return aids;
+		}
+		void addActionId(uint16_t id) {
+			aids.emplace_back(id);
+		}
 
 		virtual ReturnValue canExecuteAction(const Player* player, const Position& toPos);
 		virtual bool hasOwnErrorHandler() {
@@ -51,9 +93,12 @@ class Action : public Event
 	private:
 		std::string getScriptEventName() const override;
 
-		bool allowFarUse;
-		bool checkFloor;
-		bool checkLineOfSight;
+		bool allowFarUse = false;
+		bool checkFloor = true;
+		bool checkLineOfSight = true;
+		std::vector<uint16_t> ids;
+		std::vector<uint16_t> uids;
+		std::vector<uint16_t> aids;
 };
 
 class Actions final : public BaseEvents
@@ -72,6 +117,8 @@ class Actions final : public BaseEvents
 		ReturnValue canUse(const Player* player, const Position& pos);
 		ReturnValue canUse(const Player* player, const Position& pos, const Item* item);
 		ReturnValue canUseFar(const Creature* creature, const Position& toPos, bool checkLineOfSight, bool checkFloor);
+
+		bool registerLuaEvent(Event* event);
 
 	private:
 		ReturnValue internalUseItem(Player* player, const Position& pos, uint8_t index, Item* item, bool isHotkey);
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -100,6 +100,21 @@ void MonsterType::createLoot(Container* corpse)
 	corpse->startDecaying();
 }
 
+void MonsterType::loadLoot(MonsterType* monsterType, LootBlock lootBlock)
+{
+	if (lootBlock.childLoot.empty()) {
+		bool isContainer = Item::items[lootBlock.id].isContainer();
+		if (isContainer) {
+			for (LootBlock child : lootBlock.childLoot) {
+				lootBlock.childLoot.push_back(child);
+			}
+		}
+		monsterType->info.lootItems.push_back(lootBlock);
+	} else {
+		monsterType->info.lootItems.push_back(lootBlock);
+	}
+}
+
 std::vector<Item*> MonsterType::createLootItem(const LootBlock& lootBlock)
 {
 	int32_t itemCount = 0;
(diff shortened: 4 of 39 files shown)


src/movement.h
```diff
@@ -90,9 +96,9 @@ class MoveEvents final : public BaseEvents
 		LuaScriptInterface scriptInterface;
 };
 
-using StepFunction = std::function<uint32_t(Creature* creature, Item* item, const Position& pos)>;
-using MoveFunction = std::function<uint32_t(Item* item, Item* tileItem, const Position& pos)>;
-using EquipFunction = std::function<uint32_t(MoveEvent* moveEvent, Player* player, Item* item, slots_t slot, bool boolean)>;
+typedef uint32_t(StepFunction)(Creature* creature, Item* item, const Position& pos);
```
djarek: `using StepFunction = uint32_t(Creature* creature, Item* item, const Position& pos)`

Same can be done with the rest of callbacks.

src/movement.h
```diff
@@ -132,29 +138,117 @@ class MoveEvent final : public Event
...
+			return wieldInfo;
+		}
+		void setWieldInfo(WieldInfo_t info) {
+			wieldInfo |= info;
+		}
 
-	private:
 		std::string getScriptEventName() const override;
 
+		static StepFunction StepInField;
```
djarek: These statics seem to be unused?

src/script.cpp
```diff
@@ -0,0 +1,83 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "otpch.h"
+
+#include "script.h"
+#include <boost/filesystem.hpp>
+using namespace std;
+using namespace boost::filesystem;
```
djarek: Please don't use `using namespace`, it can cause surprising, silent breakage. Namespace aliases are fine though:
`namespace fs = boost::filesystem`.

src/script.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+	const auto dir = current_path() / "\\data\\" / folderName / "\\";
+	if(!exists(dir) || !is_directory(dir)) {
+		cout << "[Warning - Scripts::loadScripts] Can not load folder 'scripts'." << endl;
+		return false;
+	}
+
+	cout << ">> Loading lua scripts" << endl;
+	recursive_directory_iterator endit;
+	typedef vector<path> vec;
```
djarek: What's the point of this alias?

src/movement.h
EvilHero90: this needs some deeper look into, whenever I try to adjust this, I'm either stuck not beeing able to use those functions properly from the luascripts.cpp or it throws me tons of errors

src/movement.h
EvilHero90: They're used in luascripts.cpp lines 13233-13248
```c++
        moveevent->stepFunction = moveevent->StepInField;
} else if (tmpStr == "stepout") {
	moveevent->setEventType(MOVE_EVENT_STEP_OUT);
	moveevent->stepFunction = moveevent->StepOutField;
} else if (tmpStr == "equip") {
	moveevent->setEventType(MOVE_EVENT_EQUIP);
	moveevent->equipFunction = moveevent->EquipItem;
} else if (tmpStr == "deequip") {
	moveevent->setEventType(MOVE_EVENT_DEEQUIP);
	moveevent->equipFunction = moveevent->DeEquipItem;
} else if (tmpStr == "additem") {
	moveevent->setEventType(MOVE_EVENT_ADD_ITEM);
	moveevent->moveFunction = moveevent->AddItemField;
} else if (tmpStr == "removeitem") {
	moveevent->setEventType(MOVE_EVENT_REMOVE_ITEM);
	moveevent->moveFunction = moveevent->RemoveItemField;
```

src/script.cpp
EvilHero90: will be fixed in the next commit

src/script.cpp
EvilHero90: I've used an already example for the parsing and this was still included, will be fixed in the next commit

src/movement.h
djarek: You're probably forgetting a `*` somewhere:
https://godbolt.org/z/re1qPL

src/movement.h
djarek: Why are they static then?

EvilHero90: to be able to link them properly I had to make them static, if you have a better idea feel free to let me know

djarek: I don't understand, you already have an instance of `MoveEvent` (`movevent`), so there's no need for them to be static, is there?

Also I don't think it even makes sense for them to be `static`, because it would limit any of these functions to just one per type?

EvilHero90: well for the userdata it's intended to just have an one type use
it would probably harm more if we could do something like this
```lua
local object = MoveEvent()
object.onEquip = function(...)
    return true
end
object.onAddItem = function(...)
   return false
end
```
This could crash the server easily with the wrong setup.

I'm completly open to whatever suggestion you have here.
However I've mostly just mimic'd the loadFunction from xml in this case, so the implementation is rather the same
```c++
bool MoveEvent::loadFunction(const pugi::xml_attribute& attr, bool isScripted)
{
	const char* functionName = attr.as_string();
	if (strcasecmp(functionName, "onstepinfield") == 0) {
		stepFunction = StepInField;
	} else if (strcasecmp(functionName, "onstepoutfield") == 0) {
		stepFunction = StepOutField;
	} else if (strcasecmp(functionName, "onaddfield") == 0) {
		moveFunction = AddItemField;
	} else if (strcasecmp(functionName, "onremovefield") == 0) {
		moveFunction = RemoveItemField;
	} else if (strcasecmp(functionName, "onequipitem") == 0) {
		equipFunction = EquipItem;
	} else if (strcasecmp(functionName, "ondeequipitem") == 0) {
		equipFunction = DeEquipItem;
	} else {
		if (!isScripted) {
			std::cout << "[Warning - MoveEvent::loadFunction] Function \"" << functionName << "\" does not exist." << std::endl;
			return false;
		}
	}

	if (!isScripted) {
		scripted = false;
	}
	return true;
}
```
The only difference was that with the functions having a namespace I couldn't call them from another cpp file which gave me trouble

djarek: Ehh, I just realized these are member function not variables... Can you change the declaration so that it's less confusing and actually looks like a function declaration?
```suggestion
		static uint32_t StepInField(Creature* creature, Item* item, const Position& pos);
```

EvilHero90: couldn't we get rid of all this then?
```c++
typedef uint32_t(StepFunction)(Creature* creature, Item* item, const Position& pos);
typedef uint32_t(MoveFunction)(Item* item, Item* tileItem, const Position& pos);
typedef uint32_t(EquipFunction)(MoveEvent* moveEvent, Player* player, Item* item, slots_t slot, bool boolean);
```

djarek: It's still used to define the callback pointer data members:
```		StepFunction* stepFunction;
		MoveFunction* moveFunction;
		EquipFunction* equipFunction;```
If they're used elsewhere still, they should stay, if not, I don't mind if they're removed.

Comment (EvilHero90) [reply src/movement.h]: fixed in [c73ed97](https://github.com/otland/forgottenserver/pull/2558/commits/c73ed9783e0c33312326a790475cd4eee4f61ab9)

File: src/movement.h (hunk 1 above)
Comment (EvilHero90) [reply src/movement.h]: fixed in [c73ed97](https://github.com/otland/forgottenserver/pull/2558/commits/c73ed9783e0c33312326a790475cd4eee4f61ab9)

**Comment by EvilHero90** [comment]
The basic system is now working without any problems so far, would be good to have a few more reviews before I expand to Monsters / NPC's and such.

File: data/scripts/actions/spellbook.lua [hunk 5]
```diff
@@ -0,0 +1,35 @@
...
+	for i, spell in ipairs(spells) do
+		local line = ""
+		if prevLevel ~= spell.level then
+			if i ~= 1 then
+				line = "\n"
+			end
+			line = line .. "Spells for Level " .. spell.level .. "\n"
+			prevLevel = spell.level
+		end
+		text = text .. line .. "  " .. spell.words .. " - " .. spell.name .. " : " .. spell.mana .. "\n"
```
djarek: You might wanna store the strings into a table and then call, `table.concat(t)`, because this currently creates a lot of temporary strings.

data/scripts/talkactions/position.lua
```diff
@@ -0,0 +1,15 @@
...
+		player:teleportTo(Position(split[1], split[2], split[3]))
+	else
+		local position = player:getPosition()
+		player:sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, "Your current position is: " .. position.x .. ", " .. position.y .. ", " .. position.z .. ".")
+	end
+	return false
+end
+
+talk:separator(" ")
+talk:register()
```
djarek: newline

EvilHero90: fixed in [9c085d8](https://github.com/otland/forgottenserver/pull/2558/commits/9c085d844c9bc73bed7f6056131be74ed4a6ce03)

data/scripts/actions/spellbook.lua
EvilHero90: This was actually the original spellbook lua which is implemented at the moment, just moved it to scripts to show how the lua file setup would look like.
However fixed what you suggested in [9c085d8](https://github.com/otland/forgottenserver/pull/2558/commits/9c085d844c9bc73bed7f6056131be74ed4a6ce03)

djarek: I think you can also apply the same technique when creating a line, but looks good anyway.

EvilHero90: I guess you mean something like this?
```lua
for i, spell in ipairs(spells) do
	if prevLevel ~= spell.level then
		if i == 1 then
			text[#text == nil and 1 or #text+1] = "Spells for Level "
		else
			text[#text+1] = "\nSpells for Level "
		end
		text[#text+1] = spell.level
		text[#text+1] = "\n"
		prevLevel = spell.level
	end
	text[#text+1] = spell.words
	text[#text+1] = " - "
	text[#text+1] = spell.name
	text[#text+1] = " : "
	text[#text+1] = spell.mana
	text[#text+1] = "\n"
end
```

djarek: Looks terrible :P, i wonder if this can be abstracted somehow, my lua-fu isn't good enough to tell you.

EvilHero90: could be done with another loop inside parsing 'spell' actually and then glue it together but then again it's kinda bad to setup and I'd rather not put another loop in for basicly no 'real' reason

src/actions.h
```diff
@@ -34,12 +34,54 @@ class Action : public Event
...
+			return checkFloor;
+		}
+		void setCheckFloor(bool v) {
+			checkFloor = v;
+		}
+
+		std::vector<uint16_t> getIdRange() {
+			return ids;
+		}
+		void addId(uint16_t id) {
```
LucasCPrazeres: Would not it be better, or more organized, to use the same function as the movements?
- addItemId
- addActionId
- addUniqueId
- ...

EvilHero90: good call, will do that with the next commit, thanks for pointing it out

EvilHero90: fixed in [24db279](https://github.com/otland/forgottenserver/pull/2558/commits/24db2794032fc0aa78bcddbc70163ea293a9feb8)

data/scripts/actions/spellbook.lua
andersonfaaria: Use my spellbook script instead, it's best optimized and the closest to the global one.
Tables requires twice the access number so it's not really a good approach. In this particular case the best option would be to parse all spells on startup and create stored versions for each class divided by level and magic level, this way you could easily retrieve cached versions without performing this whole loop every damn time. But since I'm too lazy to do this on startup and store this table in a lib file and reload it, use this "not the best but still very good" approach below:

`function onUse(player, item, fromPosition, target, toPosition, isHotkey)
	local text = ""
	local tlvl = {}
	local tml = {}

	for _, spell in ipairs(player:getInstantSpells()) do
		if spell.level ~= 0 or spell.mlevel ~= 0 then
			if spell.manapercent > 0 then
				spell.mana = spell.manapercent .. "%"
			end
			if spell.level > 0 then
				tlvl[#tlvl+1] = spell
			elseif spell.mlevel > 0 then
				tml[#tml+1] = spell
			end
		end
	end

	table.sort(tlvl, function(a, b) return a.level < b.level end)
	local prevLevel = -1
	for i, spell in ipairs(tlvl) do
		local line = ""
		if prevLevel ~= spell.level then
			if i ~= 1 then
				line = "\n"
			end
			line = line .. "Spells for Level " .. spell.level .. "\n"
			prevLevel = spell.level
		end
		text = text .. line .. "  " .. spell.words .. " - " .. spell.name .. " : " .. spell.mana .. "\n"
	end
	text = text.."\n"
	table.sort(tml, function(a, b) return a.mlevel < b.mlevel end)
	local prevmLevel = -1
	for i, spell in ipairs(tml) do
		local line = ""
		if prevLevel ~= spell.mlevel then
			if i ~= 1 then
				line = "\n"
			end
			line = line .. "Spells for Magic Level " .. spell.mlevel .. "\n"
			prevmLevel = spell.mlevel
		end
		text = text .. line .. "  " .. spell.words .. " - " .. spell.name .. " : " .. spell.mana .. "\n"
	end


	player:showTextDialog(item:getId(), text)
	return true
end
`

EvilHero90:
Feel free to keep suggesting stuff, if someone thinks I'm missing something just let me know

andersonfaaria:
I know this is a huge change by itself. But I think it would be very welcomed to rework ItemType/Item classes and also to put dropLoot logic in an event similar to how onGainExperience works for players today.

The main problems I have faced when developing my server was to create an unique attribute system parallel to the one that exists today, one that could allow to create a fire sword with more fire damage for example. The way I did it was copying all the ItemType attribute logic to Item as well, and upon instancing I  would check if there's an ItemType and if so I updated the values accordingly. Those values could be then manipulated freely for each instance.

As for the loot, I got it in a forum and reworked most part to work 100% in Lua but this version is very messy and repetitive, besides it requires onSpawn event to register dropLoot event in everyspawned monster, which breaks 100% of TFS logic. The best possible thing would be to make a clean version with a mix of both C++ and Lua logic that could allow us to manipulate chances, and items before they are put in corpse.

Anyway, what about questlog in Lua PR? what's preventing it to be merged?
I hate so much that people still try to keep backwards compatibility not matter what

raymondtfr:
I think we should consider creating a new branch for fully implementing revscriptsys and all the useful metatables and new events we can think of, as well as completely getting rid of XML by converting everything to revscriptsys to test its features and stability until it is considered stable... and maybe also implement OTFB map format, and maybe using a .json for instead of xml for items 🌮 .

mattyx14:
As for example this system will implement something similar to items.srv

https://github.com/TwistedScorpio/Nostalrius/blob/master/data/items/items.srv
to eliminate dependency on items.xml-otb

Znote:
There are alot of ideas spawning from this pull request, but currently it does not break backwards compatibility, and functions as a partial port of a few sample scripts. 

Lets keep it that way, once it gets merged we can send additional pull requests that "revscriptify" the repo, without sending this pull request into complete development hell. 

Since there is no issues merging this as is (once the remaining checklist tasks are done), we can branch off and work on suggestions provided by @raymondtfr, @andersonfaaria and @mattyx14 in separate branches and pull requests. 

At least thats what I think, get this "core" merged, then begin to play with innovation that expands on this. If said innovation takes time, no problem since this is already functioning as mods pack and can be utilized by the community right away.

EvilHero90:
> 
> 
> I think we should consider creating a new branch for fully implementing revscriptsys and all the useful metatables and new events we can think of, as well as completely getting rid of XML by converting everything to revscriptsys to test its features and stability until it is considered stable... and maybe also implement OTFB map format, and maybe using a .json for instead of xml for items 🌮 .

Those are my future plans for it, but before all that we need to make the transition for it and I have to create some xml parsers which automaticly transfers old xml/lua stuff into the new lua format, because taking away xml dependency will give a huge hit to people who have tons of custom stuff already, if we don't provide an auto parser so people have to do it by hand.

I'm not familiar with OTFB (would be nice to provide a link)
About .json I'm not really happy with that actually, I'd rather stick to one scripting language for the entire engine, I've already plans for items in lua which allow even virtualization and such, so we take away stricktness on items.

> 
> 
> As for example this system will implement something similar to items.srv
> 
> https://github.com/TwistedScorpio/Nostalrius/blob/master/data/items/items.srv
> to eliminate dependency on items.xml-otb

same as above.

> 
> 
> There are alot of ideas spawning from this pull request, but currently it does not break backwards compatibility, and functions as a partial port of a few sample scripts.
> 
> Lets keep it that way, once it gets merged we can send additional pull requests that "revscriptify" the repo, without sending this particular pull request into complete development hell.
> 
> Since there is no issues merging this as is (once the remaining checklist tasks are done), we can branch off and work on suggestions provided by @raymondtfr, @andersonfaaria and @mattyx14 in separate branches and pull requests.
> 
> At least thats what I think, get this "core" merged, then begin to play with innovation that expands on this. If said innovation takes time, no problem since this is already functioning as mods pack and can be utilized by the community right away.

I'm completly on your side, the entire revscriptsys for now will be completly vanilla, just a bridge between core and lua to register the events, anything beyond that can be done by future PR's once it's implemented

EvilHero90:
I have to strip down the check list, neither of the remaining can be easily ported without major rework which could most likely break stuff.
I'd rather have it this far merged and once we decide to step up and port it completly to lua only, then we can re work the other stuff aswell without worrying about xml dependency

src/script.cpp
```diff
@@ -0,0 +1,87 @@
...
+Scripts::~Scripts()
+{
+	scriptInterface.reInitState();
+}
+
+bool Scripts::loadScripts(std::string folderName, bool isLib)
+{
+	namespace fs = boost::filesystem;
+
+	const auto dir = fs::current_path() / "\\data\\" / folderName / "\\";
```
djarek: The separators aren't needed, are they? `operator/` inserts them as needed.

EvilHero90: Would need to test that again but if I recall right it had some trouble with different Unix based OS without them, VS atleast doesn't throw errors, I'll just make a commit and see if travis passes

src/monsters.cpp
```diff
@@ -610,6 +625,234 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+
+			ConditionType_t conditionType;
+			if (speedChange > 0) {
+				conditionType = CONDITION_HASTE;
+				combat->setParam(COMBAT_PARAM_AGGRESSIVE, 0);
+			} else {
+				conditionType = CONDITION_PARALYZE;
+			}
+
+			ConditionSpeed* condition = reinterpret_cast<ConditionSpeed*>(Condition::createCondition(CONDITIONID_COMBAT, conditionType, duration, 0));
```
djarek: Wouldn't a `static_cast` be sufficient here?

src/monsters.cpp
```diff
@@ -610,6 +625,234 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
+			condition->setFormulaVars(speedChange / 1000.0, 0, speedChange / 1000.0, 0);
+			combat->addCondition(condition);
+		} else if (tmpName == "outfit") {
+			int32_t duration = 10000;
+
+			if (spell->duration != 0) {
+				duration = spell->duration;
+			}
+
+			ConditionOutfit* condition = reinterpret_cast<ConditionOutfit*>(Condition::createCondition(CONDITIONID_COMBAT, CONDITION_OUTFIT, duration, 0));
```
djarek: Wouldn't a `static_cast` be sufficient here?

EvilHero90: missed those 2 out, will be fixed in next commit

EvilHero90: fixed in [8f44595](https://github.com/otland/forgottenserver/pull/2558/commits/8f44595e1ea4ea1fae95008ffe9befbcfbdcfb96)

src/monsters.cpp
EvilHero90: fixed in [8f44595](https://github.com/otland/forgottenserver/pull/2558/commits/8f44595e1ea4ea1fae95008ffe9befbcfbdcfb96)

raymondtfr:
Wiki must be updated because these changes requires ``vcpkg install boost-filesystem:x32-windows`` or ``vcpkg install boost-filesystem:x64-windows`` to be added to ``vcpkg`` Windows installation guides in order for it to compile (atleast in Visual Studio 2017). @djarek @Znote @EvilHero90

PR #2559, opened 2019-04-05
Correct way to random step

nekiro:
Monsters dont random step while no target like they should right now, this pr should fix the behaviour to the correct one.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1106,7 +1106,7 @@ bool Monster::getNextStep(Direction& direction, uint32_t& flags)
 
 	bool result = false;
 	if ((!followCreature || !hasFollowPath) && (!isSummon() || !isMasterInRange)) {
-		if (getWalkDelay() <= 0) {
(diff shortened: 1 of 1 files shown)


DSpeichert:
Maybe add a note of why this is the right formula? I'm not sure why it's 600.

nekiro:
There is no „right” nor released formula for that, but that number was the closesr I coild get by comparing cipsoft monsters.

renanluis90:
Why the code don't respect less numbers there? I made many tests to put the real behavior of delay, numbers less of 600, the code dont read

nekiro:
I did further testing and seems like monsters on rl tibia (cipsoft servers) are stepping each 1 second, doesnt matter what speed monster has. Updated the pr.

soul4soul:
> 
> 
> I did further testing and seems like monsters on rl tibia (cipsoft servers) are stepping each 1 second, doesnt matter what speed monster has. Updated the pr.

For random stepping when the creature doesn't have a target I agree.

PR #2566, opened 2019-04-17
rework on /reload

### LINKED ISSUE #2567: Crash when reload

1. Log on to an admin character
2. Try to /reload any reload type
3. Crash

### Expected behaviour
It should reload.

### Actual behaviour
It crashes.

Running VS Debugger shows me the error is in LuaScriptInterface::luaDeleteAction
The exception is:
```
Exception thrown: read access violation.
*(*(__imp_lua_touserdata(...))) was nullptr.```

### PR DESCRIPTION
**Pull request by EvilHero90**
Fix #2567
- `reload` works now as it has been working before
- There are some additions to it now, you can use `/reload scripts` for the sole reload of the scripts folder and `/reload libs` to reload the global.lua and scripts/lib/ folder.

### CODE REVIEW

**Comment by raymondtfr** [comment]
Is the decision for making ``/reload`` reload all files due to restrictions imposed by our current code base design or is it just a workaround?

**Comment by EvilHero90** [comment]
The problem is the design of revscriptsys doesn't have exclusive loading which means, an action script could be in either 'scripts/actions/' or he could be directly in 'scripts/' or wherever you want it, so I cannot specify the exact location of the files for the specific action.
It could even be combined in a file with a movement in a complete different folder.
That basicly means, if I did a '/reload actions' and it wipes all states for actions and we assume that only in 'scripts/actions' resides the action scripts we want to reload but then in fact we would have somewhere in a folder called like 'scripts/quests/' then it would mean we wouldn't load this file which could break an entire system of other events because it's not correctly loaded.
What I could 'maybe' do but it wouldn't make much sense as it would be almost the same, if you call let's say '/reload actions' it would only reload xml action files but still all files which reside in 'scripts' so we don't forget to load something on the run, then again this solution wouldn't bring much more onto the table rather than making it unnecessarily bulky.
Future plans for this would in all honesty be to remove /reload completly as it can harm more and a restart is less than 30 secs depending on your map anyway. 
On a live server it shouldn't be enabled anyway but that's something to be discussed once we proceed further.

**Comment by vankk** [comment]
I don't think the idea of reloading everything, sometimes on live it is very useful, must have a way to avoid this - if don't, the reload.lua should be reworked as well, sounds too vague.

**Comment by EvilHero90** [comment]
Ok go ahead, tell me a good reason as to why it's important to be able to perform reloads on a live server, if what you say is good I'll change my mind, meanwhile I still believe it's rather bad and can do more harm to your server then anything else.

**Comment by andersonfaaria** [comment]
> Ok go ahead, tell me a good reason as to why it's important to be able to perform reloads on a live server, if what you say is good I'll change my mind, meanwhile I still believe it's rather bad and can do more harm to your server then anything else.

seriously? isn't obvious? If my server has a problem will I have to restart it instead of making just a quickfix and reloading it? What about npcs/creatureevents that cause many harm being reloaded live?

**Comment by djarek** [comment]
@andersonfaaria I wouldn't recommend doing a reload in a production environment. The reload system is inherently unsafe and can break, if misused, by introducing danging pointers.

**Comment by EvilHero90** [comment]
Imagine you do a quickfix meanwhile while you reloaded some event broke and throws up errors now which wouldn't be there unless you reload, that's the time you'll have to do a restart anyway.
I'd rather do a restart and be sure everything works correctly after it then reloading and on the way the server might crash, worst case server didn't save for hours and it crashes, I doubt any of your players will find that to be cool.

File: src/actions.cpp [hunk 1]
```diff
@@ -186,9 +186,10 @@ bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node)
 	return false;
 }
 
-bool Actions::registerLuaEvent(Event* event)
+bool Actions::registerLuaEvent(Action* event)
 {
-	Action_ptr action{ static_cast<Action*>(event) }; //event is guaranteed to be an Action
+	Event_ptr ptr = Event_ptr(event);
+	Action_ptr action{ static_cast<Action*>(ptr.release()) }; //event is guaranteed to be an Action
```
EvilHero90: @djarek just a quick question, wouldn't be `ptr.reset()` here better as it does not memory leak if we do not call delete on the raw pointer aswell? it maybe makes sense to use `ptr.release` if the event is guaranteed to be loaded again but on a reload it would memory leak if for example I removed that script from my files before reloading or not?

djarek: `Action_ptr` is a `unique_ptr`, so there's no leak here. Also, if we did `reset()` here it would destroy the object and invalidate the pointer.

djarek: You can drop `ptr` and assign the pointer directly into a `unique_ptr` (since the cast is no longer required):
```suggestion
	Action_ptr action{ event };
```

EvilHero90: ok that makes sense now, been reading a bit about unique pointers just now, was curious if `ptr.release` was called once `action` gives up ownership or if it's right away thanks for clearing that up

EvilHero90: ahh ok because I already push an `Action*` pointer into the function, makes sense actually.

EvilHero90:
@djarek can you do a review, so we can get this merged and get rid of the reload crash

vankk:
> 
> 
> Ok go ahead, tell me a good reason as to why it's important to be able to perform reloads on a live server, if what you say is good I'll change my mind, meanwhile I still believe it's rather bad and can do more harm to your server then anything else.

A good example is what Anderson said previously; that I can make a quick fix for an huge error, without restart the server - restarting the server out of the server save can be harmful to the server and/or his reputation. I speak this as a server owner and I have alot of experience with players. 

I'd rather have the command working by the reloading what I want and don't use than not have the command and need at some point. What about if I don't want to reload globalevents, but need reload the npcs to fix a simple issue, why should reload everything? We would be forcing the server admins to restart their servers without needed. Sounds a regress for me, even with this awesome technology.

EvilHero90:
There has been some major re work, reload works like the old one now.
You can also use /reload scripts and /reload libs now.

src/baseevents.h
```diff
@@ -44,7 +44,15 @@ class Event
...
 
+		bool isFromLua() const {
+			return fromLua;
+		}
+		void setFromLua(bool b) {
+			fromLua = b;
+		}
+
 		bool scripted = false;
+		bool fromLua = false;
```
ranisalt: No need for `get`/`set` if the attribute is public and the methods are basic.

src/creatureevent.cpp
```diff
@@ -29,15 +29,21 @@ CreatureEvents::CreatureEvents() :
...
+		} else {
+			++it;
+		}
 	}
 
 	//clear lua state
-	scriptInterface.reInitState();
+	if (!fromLua) {
+		scriptInterface.reInitState();
+	}
```
ranisalt: Hm, this seems to be very repetitive. I suggest you move it to the parent and call it on derived classes. Also, create a function for the `for` above instead of repeating it everywhere.

src/creatureevent.cpp
```diff
@@ -101,7 +107,7 @@ bool CreatureEvents::registerLuaEvent(Event* event)
 		//if there was an event with the same that is not loaded
 		//(happens when realoading), it is reused
 		if (!oldEvent->isLoaded() && oldEvent->getEventType() == creatureEvent->getEventType()) {
-			oldEvent->copyEvent(creatureEvent);
+			oldEvent->copyEvent(creatureEvent.get());
```
ranisalt: Make `oldEvent` use a smart pointer instead of accessing the raw pointer.

src/actions.h
```diff
@@ -118,13 +118,13 @@ class Actions final : public BaseEvents
 		ReturnValue canUse(const Player* player, const Position& pos, const Item* item);
 		ReturnValue canUseFar(const Creature* creature, const Position& toPos, bool checkLineOfSight, bool checkFloor);
 
-		bool registerLuaEvent(Event* event);
+		bool registerLuaEvent(Action* event);
+		void clear(bool fromLua) final;
```
ranisalt: You mixed `final` and `override` everywhere. I believe you should use both.

EvilHero90: to be fair they where like that to begin with, I just moved them from `private:` to `public:` but yea I can change that

src/creatureevent.cpp
EvilHero90: this part can be moved that's ok but for the `for` loops they're actually not all the same so it wouldn't make much sense to move them to elswhere, as they already have their function which is explicit for them.

src/baseevents.h
EvilHero90: will do that

src/actions.h
ranisalt: Oh sure, do it while working around here :smile:

djarek:
Since you removed the gc finalizers, does this mean that the Event objects passed to lua are leaked on a reload/when no reference to them exists?

EvilHero90:
I was worried about that aswell but if they're declared as local they get overwritten once you reload which means gc picks them up, it's basicly like re addressing a table in lua
ex:
```lua
local ex = {1,2,3,4,5,6,7,8,9,0}
ex = {}
now it will expose {1,2,3,4,5,6,7,8,9,0} to gc as we re addressed the var with a new table
```
so atleast from the lua side there are no side effects.

djarek:
Sure, but full-userdata is opaque to lua, so this allocation: https://github.com/otland/forgottenserver/pull/2566/files#diff-52cc0839e1c346634d8dad954bb7b30eL13782 leaks on reload (the pointer on lua side is cleaned but not the memory pointed to by the pointer).

EvilHero90:
that's what you meant, yea that causes a leak atm, maybe try `delete` on `clear()` ?

djarek:
I'm pretty sure that may leave dangling pointers on the lua side (because you don't know if any references exist).

EvilHero90:
I think this should get merged so we get rid of the crash on reload, the memory leak just happens on /reload all or /reload scripts.
I'll keep working on it and try to solve this problem with handling this a different way but that might take some time.

Znote:
@Vulcanx @vankk @andersonfaaria Please test this and let us know how it works. This should now work as the old reload system you guys are in favor of.

vankk:
I don't have too much free time for the next two weeks, if no one has tested until then I test. Meanwhile I think we should remove the revscripsys from the master, and create a branch, so that won't harm any servers that may download the newest changes while we still got this issue.

Znote:
> I don't have too much free time for the next two weeks, if no one has tested until then I test. Meanwhile I think we should remove the revscripsys from the master, and create a branch, so that won't harm any servers that may download the newest changes while we still got this issue.

The people who crash their servers should test this pull request, if they don't it just shows how little they actually care. Reverting revscriptsys does not help because then they certainly wont test the fix to their problem.

EvilHero90:
First of all, I did not rewrite this incase of 3 people complaining about reload I could have cared less, I did this in respect of Znote asking me.
Branching out for what reason? if you guys wouldn't have complained to begin with this would have already been merged and the crash would have been gone.
Stuff always takes 5x times longer to be merged because everytime you want to make a change another clown jumps out of his box and complains.

In the meantime I'm waiting for this to get merged I could have already finished Spells / Weapons, which would have finished the elementary stuff to be in revscriptsys.

@Znote This crash occurs for everyone using the last commits right now, I've had several people testing it now they all said it was fixed (as I don't know their github names I cannot tag them here) so this can be merged without any problems and it gets rid of the crash.

ninjalulz:
I merged this pull request due to the following reasons:
 - Reload should **not** be used in a production environment
 - End-users should be using **TFS 1.2** if they want a stable engine that doesn't come with any unexpected changes.
 - If there were flaws with the initial **revscriptsys** implementation, then it shouldn't have been merged in the first place but here we are, so there's no reason for us not to merge this.

Lyuzera:
Why reload the entire data / scripts directory at once if we can reload an interface and references in data/scripts?
for example /reload actions
reload all the LUA files in data/actions/scripts and also in data/scripts/actions, so it would be more organized. I do not know about you, but revscriptsys can deprecate XML Interfaces.

EvilHero90:
> 
> 
> Why reload the entire data / scripts directory at once if we can reload an interface and references in data/scripts?
> for example /reload actions
> reload all the LUA files in data/actions/scripts and also in data/scripts/actions, so it would be more organized. I do not know about you, but revscriptsys can deprecate XML Interfaces.

https://github.com/otland/forgottenserver/pull/2566#issuecomment-484205801

PR #2568, opened 2019-04-21
movements crash fix

EvilHero90:
- This fixes a crash which occured when you have a script set for movements but not a function in movements.xml
- You can also now either set a function/script for equip/deequip event types or both at once.
ex: `<movevent event="Equip" itemid="2195" slot="feet" function="onEquipItem" script="boh.lua" />`
This would normaly apply all default stuff handling like movement speed (set in items.xml) in addition to the script file.


src/movement.cpp
```diff
@@ -923,11 +923,15 @@ bool MoveEvent::executeStep(Creature* creature, Item* item, const Position& pos)
...
 			if (executeEquip(player, item, slot, isCheck)) {
 				return 1;
-			} else {
-				return 0;
+			}
+		} else {
+			if (equipFunction(this, player, item, slot, isCheck) == 1) {
+				if (executeEquip(player, item, slot, isCheck)) {
+					return 1;
+				}
```
jo3bingham: Would anyone be against condensing this into a single if statement?
```c++
if (!equipFunction || equipFunction(this, player, item, slot, isCheck) == 1) {
    if (executeEquip(player, item, slot, isCheck)) {
        return 1;
    }
}
```

EvilHero90: I don't mind tbh, most of the time here it's rather done by a nested if, for better readability I guess, so I've just adapted to the convention of how it's mostly done.

ranisalt: It can be condensed even further.

EvilHero90: I guess you mean like
```cpp
if ((!equipFunction) || (equipFunction(this, player, item, slot, isCheck) == 1) && (executeEquip(player, item, slot, isCheck))) {
        return 1;
}
```

EvilHero90: I'm more for readability but if both of you want it to condensed then say it please and I'll do it.
Because this should get merged already, as it fixes a crash which occurs on movements right now.

ninjalulz: I'll merge this if you apply this change.

Also couldn't you just condense it to the likes of this?
```CPP
if ((!equipFunction || equipFunction(this, player, item, slot, isCheck) == 1) && executeEquip(player, item, slot, isCheck)) {
        return 1;
}
```

EvilHero90: actually no, because we would skip `executeEquip(player, item, slot, isCheck` if `!equipFunction` would be false.
I'll stick with what @jo3bingham suggested, if that's ok?

ninjalulz: Sure, that's fine

PR #2569, opened 2019-04-23
Moved drop loot mechanism to Lua

joseluis2g:
Moved drop loot mechanist to Lua.
Added a new class to events called Monster and a new method called onDropLoot.
Added party to Lua core.
Added a new function to party class called broadcastPartyLoot (based on sources).
Replicated (or tried?) getLootRandom and createLootItem to Lua.
Added a new function to container class called createLootContainer (based on sources).
Added two new values to enums MESSAGE_LOOT and MAX_LOOTCHANCE.
Added two new container methods (getContentDescription and getCorpseOwner).


src/luascript.cpp
```diff
@@ -6778,6 +6794,19 @@ int LuaScriptInterface::luaContainerGetItemCountById(lua_State* L)
...
 }
 
+int LuaScriptInterface::luaContainerGetContentDescription(lua_State* L)
+{
+	// container:getContentDescription()
+	Container* container = getUserdata<Container>(L, 1);
+	if (container) {
+		pushString(L, container->getContentDescription());
+	}
+	else {
```
jo3bingham: You'll want to move this `else` back up to the previous line.

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,52 @@
+function Monster:onDropLoot(corpse)
+	if (configManager.getNumber(configKeys.RATE_LOOT) == 0) then
+		corpse:decay()
+		return
+	end
```
jo3bingham: Newline after this for separation.

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,52 @@
...
+						if corpse:addItemEx(item) ~= RETURNVALUE_NOERROR then
+							print('[Warning] DropLoot:', 'Could not add loot item to corpse.')
+						end
+					else
+						print('[Warning] DropLoot:', 'Item is nil.')
+					end
+				end
+			end
+		end
+		if player then
```
jo3bingham: Newline before this for separation.

data/lib/core/container.lua
```diff
@@ -1,3 +1,35 @@
 function Container.isContainer(self)
 	return true
 end
+
+function Container.createLootContainer(self, lootblock)
+	if #lootblock == 0 then
+		return true
+	end
```
jo3bingham: Newline after this for separation.

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,52 @@
...
+		local monsterLoot = mType:getLoot()
+		for it = 1, #monsterLoot do
+			local loot = monsterLoot[it]
+			local itemList = createLootItem(monsterLoot[it])
+			if (#itemList > 0) then
+				for i = 1, #itemList do
+					local item = itemList[i]
+					-- check containers
+					if item:isContainer() then
+						if not item:createLootContainer(loot.childLoot) then
```
jo3bingham: In the original C++ code, the container item would be deleted from memory. You're just forgetting about it and leaving the object around creating a memory leak.

data/lib/core/container.lua
```diff
@@ -1,3 +1,35 @@
...
+	for i = 1, #lootblock do
+		if self:getSize() < self:getCapacity() then
+			local loot = lootblock[i]
+			local itemList = createLootItem(loot)
+			if (#itemList > 0) then
+				for i = 1, #itemList do
+					local item = itemList[i]
+					-- check containers
+					if item:isContainer() then
+						if not item:createLootContainer(loot.childLoot) then
```
jo3bingham: Same as above: In the original C++ code, the container item would be deleted from memory. You're just forgetting about it and leaving the object around creating a memory leak.

data/events/scripts/monster.lua
joseluis2g: oopsie, thanks joe for your revision.

joseluis2g:
item:remove() would be enought, isnt it? @jo3bingham

nekiro:
What is the reason to move it to lua?

jo3bingham:
> item:remove() would be enought, isnt it?

I don't believe so. Since you're not specifying a position in `Game.createItem`, the item is added to a temporary array within the scripting environment:
https://github.com/otland/forgottenserver/blob/f7f2928e796554b2098626f4068b3cff114447ba/src/luascript.cpp#L4291-L4292

The temporary items are cleaned up when the scripting environment is reset:
https://github.com/otland/forgottenserver/blob/f7f2928e796554b2098626f4068b3cff114447ba/src/luascript.cpp#L79-L87

However, I'm not sure when that happens. It may be the case that they get cleaned up without you needing to do anythin g. Would take someone with better knowledge of the scripting environment to answer that, or you could debug it yourself and find out (e.g., set some breakpoints in Visual Studio and attach to the running process of a debug build).

EvilHero90:
item:remove() would handle this on it's own, it'll call Game::internalRemoveItem(...) which releases the item then.

jo3bingham:
@EvilHero90 read my reply, `item:remove()` will not work in this case.

joseluis2g:
> What is the reason to move it to lua?

if gives you a ton of new posibilites.

EvilHero90:
`internalRemoveItem` has a part with 
```cpp
if (item->isRemoved()) {
	item->onRemoved();
	ReleaseItem(item);
}``` `item->onRemoved();` calls ->`ScriptEnvironment::removeTempItem(this);` which does exactly what you want :D

**Comment by jo3bingham** [comment]
@EvilHero90 `item->isRemoved()` won't actually be true in this case, though. The item's parent is set to a `VirtualCylinder`, `VirutalCylinder` overrides `Cylinder::removeThing` and does nothing, `Item::isRemoved` checks if the item has a parent and if the parent is removed, `VirtualCylinder::isRemoved` always returns `false`. So, `item->isRemoved()` will return `false` in this case, and `item->onRemoved()` won't be called.

**Comment by EvilHero90** [comment]
@jo3bingham I see what you mean now this should be rather fixed in internalRemoveItem, I'd consider this a bug which is not exclusively related to this PR.
You could also instead of creating the item with `Game.createItem` check if it's a container and pass the container as a parameter and then call `:addItem()` on the corpse or parent container itself.
Which would be more sufficient as you don't have to do steps twice.

**Comment by jo3bingham** [comment]
@EvilHero90 agreed.

**Comment by joseluis2g** [comment]
can you check the new changes? @EvilHero90 @jo3bingham

File: data/global.lua [hunk 7]
```diff
@@ -33,6 +33,48 @@ function getFormattedWorldTime()
 	return hours .. ':' .. minutes
 end
 
+function getLootRandom()
+	return math.random(0, MAX_LOOTCHANCE) / configManager.getNumber(configKeys.RATE_LOOT)
+end
+
+function createLootItem(item)
```
EvilHero90: You can remove this function now

data/lib/core/container.lua
```diff
@@ -1,3 +1,47 @@
...
+	local randvalue = getLootRandom()
+	if (randvalue < item.chance) then
+		if (ItemType(item):isStackable()) then
+			itemCount = randvalue % item.countmax + 1
+		else 
+			itemCount = 1
+		end
+	end
+	
+	while (itemCount > 0) do
```
EvilHero90: You don't need a while loop here, as you don't loop over something here
```lua
if itemCount > 0 then
```

data/lib/core/container.lua
```diff
@@ -1,3 +1,47 @@
...
+				tmpItem:setActionId(item.actionId)
+			end
+
+			if (item.text and item.text ~= "") then
+				tmpItem:setText(item.text)
+			end
+		end
+	end
+	return true
+end
```
EvilHero90: new line after the end

data/lib/core/party.lua
```diff
@@ -0,0 +1,10 @@
+function Party.broadcastPartyLoot(self, text)
+	self:getLeader():sendTextMessage(MESSAGE_INFO_DESCR, text)
+	local membersList = self:getMembers()
+	for i = 1, #membersList do
+		local player = membersList[i]
+		if player then
+			player:sendTextMessage(MESSAGE_INFO_DESCR, text)
+		end
+	end
+end
```
EvilHero90: new line after the end

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,37 @@
...
+		local text = ("Loot of %s: nothing (due to low stamina)"):format(mType:getNameDescription())
+		local party = player:getParty()
+		if party then
+			party:broadcastPartyLoot(text)
+		else
+			player:sendTextMessage(MESSAGE_LOOT, text)
+		end
+	end
+	corpse:decay()
+end
```
WibbenZ: Missing new line

data/lib/core/container.lua
WibbenZ: Missing new line

data/lib/core/party.lua
WibbenZ: Missing new line

data/events/scripts/monster.lua
EvilHero90: new line after end

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,37 @@
...
+	if not player or player:getStamina() > 840 then
+		local monsterLoot = mType:getLoot()
+		for i = 1, #monsterLoot do
+			local item = corpse:createLootItem(monsterLoot[i])
+			if not item then
+				print('[Warning] DropLoot:', 'Could not add loot item to corpse.')
+			end
+		end
+
+		if player then
```
WibbenZ: if not player (line 9) and now if player?

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,37 @@
...
+			local text = ("Loot of %s: %s"):format(mType:getNameDescription(), corpse:getContentDescription())
+			local party = player:getParty()
+			if party then
+				party:broadcastPartyLoot(text)
+			else
+				player:sendTextMessage(MESSAGE_LOOT, text)
+			end
+		end
+	else
+		local text = ("Loot of %s: nothing (due to low stamina)"):format(mType:getNameDescription())
```
WibbenZ: Kinda no use for this variable since you only use it once, move them to the function(s) insted.

data/global.lua
```diff
@@ -33,6 +33,48 @@ function getFormattedWorldTime()
...
+	local itemCount = 0
+	local randvalue = getLootRandom()
+	if (randvalue < item.chance) then
```
WibbenZ: Remove all parentheses

data/lib/core/container.lua
```diff
@@ -1,3 +1,47 @@
...
+
+		if tmpItem:isContainer() then
+			for i = 1, #item.childLoot do
+				if not tmpItem:createLootItem(item.childLoot[i]) then
+					tmpItem:remove()
+				end
+			end
+		end
+
+		itemCount = itemCount - n
```
EvilHero90: can be removed aswell, don't see you using it somewhere.

data/events/scripts/monster.lua
joseluis2g: not player or player:getStamina() > 840
if there is no player it should also create the loot.

data/events/scripts/monster.lua
joseluis2g: you mean variable text? I was thinking on move it outside.

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,38 @@
+function Monster:onDropLoot(corpse)
+	if configManager.getNumber(configKeys.RATE_LOOT) == 0 then
+		corpse:decay()
+		return
+	end
+
+	local player = Player(corpse:getCorpseOwner())
+	local mType = self:getType()
+	local text
```
EvilHero90: can be removed

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,38 @@
...
+		local monsterLoot = mType:getLoot()
+		for i = 1, #monsterLoot do
+			local item = corpse:createLootItem(monsterLoot[i])
+			if not item then
+				print('[Warning] DropLoot:', 'Could not add loot item to corpse.')
+			end
+		end
+
+		if player then 
+			text = ("Loot of %s: %s"):format(mType:getNameDescription(), corpse:getContentDescription())
```
EvilHero90: make it local

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,38 @@
...
+			local party = player:getParty()
+			if party then
+				party:broadcastPartyLoot(text)
+			else
+				player:sendTextMessage(MESSAGE_LOOT, text)
+			end
+		end
+	else
+		text = ("Loot of %s: nothing (due to low stamina)"):format(mType:getNameDescription())
```
EvilHero90: make it local

data/global.lua
```diff
@@ -33,6 +33,10 @@ function getFormattedWorldTime()
 	return hours .. ':' .. minutes
 end
 
+function getLootRandom()
```
EvilHero90: for better randomization you should use `math.randomseed(os.mtime())` before calling `math.random`

data/lib/core/container.lua
```diff
@@ -1,3 +1,44 @@
 function Container.isContainer(self)
 	return true
 end
+
+function Container.createLootItem(self, item)
```
EvilHero90: You need to check in here first if container is full, else it'll throw the warning message that it cannot add the item to the container.
```lua
        return true
end
```

joseluis2g: thanks @EvilHero90

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,37 @@
+function Monster:onDropLoot(corpse)
+	if configManager.getNumber(configKeys.RATE_LOOT) == 0 then
+		corpse:decay()
```
TheSumm: As far as I can tell the corpse is already decaying: https://github.com/otland/forgottenserver/blob/32688aceee19d175a72f61a65743613bb45288e5/src/creature.cpp#L722
So this would not be needed.

data/events/scripts/monster.lua
```diff
@@ -0,0 +1,37 @@
...
+	else
+		local text = ("Loot of %s: nothing (due to low stamina)"):format(mType:getNameDescription())
+		local party = player:getParty()
+		if party then
+			party:broadcastPartyLoot(text)
+		else
+			player:sendTextMessage(MESSAGE_LOOT, text)
+		end
+	end
+	corpse:decay()
```
TheSumm: Same as above

data/events/scripts/monster.lua
joseluis2g: yeah, when I was creating it I tested it and corspes were already decaying, so both decays were useless, anyways I did it just because old code was like that.
I'm going to remove it.

joseluis2g:
> Also remove all parentheses in Lua scripts.

can you check this my friend. @WibbenZ

Znote:
Awesome, please update the wiki for the merge.
https://github.com/otland/forgottenserver.wiki.git

joseluis2g:
> Awesome, please update the wiki for the merge.
> https://github.com/otland/forgottenserver.wiki.git
> 
> https://github.com/otland/forgottenserver/wiki/Script-Interface

Maybe my question is weird, but how can I update the wiki? 📦

Znote:
> > Awesome, please update the wiki for the merge.
> > https://github.com/otland/forgottenserver.wiki.git
> > https://github.com/otland/forgottenserver/wiki/Script-Interface
> 
> Maybe my question may be weird but, how can I update the wiki? 📦

Clone the wiki git url. Its like a git repository. Each file is a wiki page. Commit a change and push, I think it will automatically get updated without requiring pull request review if im not mistaken. 

Or just tell me what to add and where to add it, and I'll do it for you. My discord is Znote#0001

Znote:
@joseluis2g Docs/wiki is still missing on this

Sajgon:
if anyone implements only this PR be sure to also merge this: https://github.com/otland/forgottenserver/pull/2637

PR #2575, opened 2019-05-05
duplicate CreatureEvent crash on reload FIX

Lyuzera:
When we try to load/reload CreatureEvents with duplicate registers, the server crashes. Here is a definitive solution.

### DIFF
--- a/src/creatureevent.cpp
+++ b/src/creatureevent.cpp
@@ -87,6 +87,7 @@ bool CreatureEvents::registerEvent(Event_ptr event, const pugi::xml_node&)
 		//if not, register it normally
 		creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
 	}
+	return true;
 }
 
 bool CreatureEvents::registerLuaEvent(CreatureEvent* event)
@@ -116,6 +117,7 @@ bool CreatureEvents::registerLuaEvent(CreatureEvent* event)
 		//if not, register it normally
 		creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
 	}
+	return true;
 }
 
 CreatureEvent* CreatureEvents::getEventByName(const std::string& name, bool forceLoaded /*= true*/)
--- a/src/creatureevent.cpp
+++ b/src/creatureevent.cpp
@@ -68,6 +68,12 @@ bool CreatureEvents::registerEvent(Event_ptr event, const pugi::xml_node&)
 		return false;
 	}
 
+	auto it = creatureEvents.find(creatureEvent->getName());
+	if (it != creatureEvents.end()) {
+		std::cout << "[Warning - CreatureEvents::registerEvent] Duplicate registered name: " << creatureEvent->getName() << std::endl;
+		return false;
+	}
+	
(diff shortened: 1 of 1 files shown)


raymondtfr:
Have you applied #2566 ?

Lyuzera:
yes, I tested to see if there was the problem in the last master and confirmed that there have crashes.

EvilHero90:
Tell us how to reproduce it, so far I didn't have any kind of crashs while reloading and provide a gbd stack trace aswell.

Lyuzera:
just register an event in creaturescripts.xml or data/scripts/creaturescripts with another existing name, for example "PlayerLogin".

EvilHero90:
That's a problem on your side then, because you shouldn't register 2 events with the same name to begin with.
This should rather get an overall rework than just a slight workaround.

Lyuzera:
I duplicate a log only to see if there could be crashes, but for beginners, he would doesn't know how to identify, so a log is required.

Znote:
I'm in favor of merging this if it causes a reload crash to become just a reload warning to remind the devs they are doing it wrong.

Lyuzera:
@Znote yes, because there really is a crash problem.

EvilHero90:
change the functions to:
```cpp
bool CreatureEvents::registerEvent(Event_ptr event, const pugi::xml_node&)
{
	CreatureEvent_ptr creatureEvent{static_cast<CreatureEvent*>(event.release())}; //event is guaranteed to be a CreatureEvent
	if (creatureEvent->getEventType() == CREATURE_EVENT_NONE) {
		std::cout << "Error: [CreatureEvents::registerEvent] Trying to register event without type!" << std::endl;
		return false;
	}

	auto it = creatureEvents.find(creatureEvent->getName());
	if (it != creatureEvents.end()) {
		std::cout << "[Warning - CreatureEvents::registerEvent] Duplicate registered creature event with name: " << creatureEvent->getName() << std::endl;
		return false;
	} else {
		creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
	}

	return true;
}

bool CreatureEvents::registerLuaEvent(CreatureEvent* event)
{
	CreatureEvent_ptr creatureEvent{ event };
	if (creatureEvent->getEventType() == CREATURE_EVENT_NONE) {
		std::cout << "Error: [CreatureEvents::registerLuaEvent] Trying to register event without type!" << std::endl;
		return false;
	}

	auto it = creatureEvents.find(creatureEvent->getName());
	if (it != creatureEvents.end()) {
		std::cout << "[Warning - CreatureEvents::registerLuaEvent] Duplicate registered creature event with name: " << creatureEvent->getName() << std::endl;
		return false;
	} else {
		creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
	}

	return true;
}
```
We can fully skip old events if we check for existence at first anyway, so it doesn't make sense leaving it like that.

Lyuzera:
@EvilHero90 thanks a lot, now we have a warning log. I don't changed old events because I doesn't know that really is.

Znote:
Obsolete as of #2579

PR #2579, opened 2019-05-06
CreatureEvent duplicate crash fix

EvilHero90:
-  This fixes when you try to register more than one creature event with the same name, that it throws an error warning instead of crashing the server.

### DIFF
--- a/src/creatureevent.cpp
+++ b/src/creatureevent.cpp
@@ -68,26 +68,15 @@ bool CreatureEvents::registerEvent(Event_ptr event, const pugi::xml_node&)
 		return false;
 	}
 
-	std::unique_ptr<CreatureEvent> oldEvent(getEventByName(creatureEvent->getName(), false));
-	if (oldEvent) {
-		//if there was an event with the same that is not loaded
-		//(happens when realoading), it is reused
-		if (!oldEvent->isLoaded() && oldEvent->getEventType() == creatureEvent->getEventType()) {
-			oldEvent->copyEvent(creatureEvent.get());
-		}
-
+	auto it = creatureEvents.find(creatureEvent->getName());
+	if (it != creatureEvents.end()) {
+		std::cout << "[Warning - CreatureEvents::registerEvent] Duplicate registered creature event with name: " << creatureEvent->getName() << std::endl;
 		return false;
 	} else {
-		//if not, register it normally
-		auto it = creatureEvents.find(creatureEvent->getName());
-		if (it != creatureEvents.end()) {
-			it->second = *creatureEvent;
-		} else {
-			creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
-		}
-
-		return true;
+		creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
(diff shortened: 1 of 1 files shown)


src/creatureevent.cpp
```diff
@@ -68,26 +68,15 @@ bool CreatureEvents::registerEvent(Event_ptr event, const pugi::xml_node&)
...
-		//if not, register it normally
-		auto it = creatureEvents.find(creatureEvent->getName());
-		if (it != creatureEvents.end()) {
-			it->second = *creatureEvent;
-		} else {
-			creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
-		}
-
-		return true;
+		creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
```
djarek: Instead of doing a find and `emplace` you should just do the `emplace` and use the return value: https://en.cppreference.com/w/cpp/container/map/emplace
```
decltype(creatureEvents)::iterator it;
bool inserted;
std::tie(it, inserted) = creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
if (!inserted)
{
  // Print warning
}

return inserted;
```

EvilHero90: nice, I've searched exactly for this, I've found something similiar but as it seems it only works for c++ 17 and up, so I've dropped that idea pretty fast
```cpp
auto [it, b] = creatureEvents.try_emplace(creatureEvent->getName(), std::move(*creatureEvent));
```
However considering your code, wouldn't it make more sense to use try_emplace instead of emplace, so we don't create an object if key already exists?

jo3bingham: `try_emplace` is C++17, so we can't use that, but you can use `auto` and use the `second` value of the returned `pair`.
```C++
auto ret = creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
if (ret.second) {
    // print warning
}

return ret.second;
```

EvilHero90: ok this is really weird, we are using c++ 11 right?
this actually worked just fine for me using VS, it compiled and the code is just working as intended.
```cpp
auto result = creatureEvents.try_emplace(creatureEvent->getName(), std::move(*creatureEvent));
	if (!result.second) {
		std::cout << "[Warning - CreatureEvents::registerLuaEvent] Duplicate registered creature event with name: " << creatureEvent->getName() << std::endl;
	}

	return result.second;
```

djarek: `emplace` doesn't construct an object if the key already exists IIRC.

EvilHero90: ok, guys, we gotta agree on one of either, which should it be? :D

Lyuzera: I prefer ```emplace```, but you decide.

jo3bingham: `try_emplace` would be preferred, but it's definitely a c++17 feature, so stick with `emplace`.

@djarek according to cppreference:
> The element may be constructed even if there already is an element with the key in the container, in which case the newly constructed element will be destroyed immediately.

Issue #2570, opened 2019-04-25
Very rare server crash on creature events.


marcinho1994:
### Steps to reproduce
1. Create a monster and make it "active". That means letting the monster attack another monster with (SPELLS) *The target monster tested here, have an event onHealthChange.

2. Now remove the target monster with this code:
```
local function RemoveCreature(cid)
    local monster = Creature(cid)
    if not monster then
         return false
    end
    monster:remove()
end
```
3. Repeat the process a thousand times until you crash.

### Expected behaviour
Remove the monster without major problems

### Actual behaviour
If the monster is removed at the exact moment it is dropping a spell the server crashes. (Important this was only tested with spells in area).

### Environment
Linux - Ubuntu 18

**GDB LOGS:**
```
#0  Combat::getCombatArea (centerPos=..., targetPos=..., 
    area=area@entry=0x1605ac0, list=empty std::forward_list) at combat.cpp:88
No locals.
#1  0x000000000042f25e in Combat::CombatFunc(Creature*, Position const&, AreaCombat const*, CombatParams const&, std::function<void (Creature*, Creature*, CombatParams const&, CombatDamage*)>, CombatDamage*) (caster=caster@entry=
    0xe095680, pos=..., area=0x1605ac0, params=..., func=..., 
    data=data@entry=0x7ffff5d49ba0) at combat.cpp:729
        tileList = empty std::forward_list
        spectators = std::unordered_set with 4560616172927782688 elements = {
          [0] = 0xc0310a74d2851e74<error reading variable: Cannot access memory at address 0x2fa833b7401fa83>...}
        rangeX = <optimized out>
        rangeY = <optimized out>
        player = 0x0
        maxX = <optimized out>
        maxY = <optimized out>
#2  0x000000000042f635 in Combat::doCombatHealth (
    caster=caster@entry=0xe095680, position=..., area=<optimized out>, 
    damage=..., params=...) at combat.cpp:869
No locals.
#3  0x0000000000430ce9 in Combat::doCombat (this=0x111d490, caster=0xe095680, 
    position=...) at combat.cpp:835
        damage = {primary = {type = COMBAT_ICEDAMAGE, value = -223}, 
          secondary = {type = COMBAT_NONE, value = 0}, origin = ORIGIN_SPELL, 
          critical = false}
#4  0x00000000005c5744 in CombatSpell::castSpell (this=0x1127c00, 
    creature=0xe095680, target=<optimized out>) at spells.cpp:345
No locals.
[...]

PR #2582, opened 2019-05-09
Update spells.cpp | fix very rare crash bug.

marcinho1994:
Fixing a very rare bug. In very specific situations, when the creature or target is removed, it can cause a crash, because in this part of the code the methods "creature->getPosition() and target->getPosition()" are used without first checking that they exist. This causes a bug in combat.cpp when trying to use a position that does not exist.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -774,6 +774,7 @@ void Monster::doAttacking(uint32_t interval)
 
 	for (const spellBlock_t& spellBlock : mType->info.attackSpells) {
 		bool inRange = false;
+		
 		if (attackedCreature == nullptr) {
 			break;
 		}
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -774,8 +774,10 @@ void Monster::doAttacking(uint32_t interval)
 
 	for (const spellBlock_t& spellBlock : mType->info.attackSpells) {
 		bool inRange = false;
-
-		if (attackedCreature && canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
+		if (attackedCreature == nullptr) {
+			break;
+		}
+		if (canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
 			if (spellBlock.chance >= static_cast<uint32_t>(uniform_random(1, 100))) {
 				if (updateLook) {
 					updateLookDirection();
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -778,6 +778,7 @@ void Monster::doAttacking(uint32_t interval)
 		if (attackedCreature == nullptr) {
 			break;
 		}
+		
 		if (canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
 			if (spellBlock.chance >= static_cast<uint32_t>(uniform_random(1, 100))) {
 				if (updateLook) {
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -775,7 +775,7 @@ void Monster::doAttacking(uint32_t interval)
 	for (const spellBlock_t& spellBlock : mType->info.attackSpells) {
 		bool inRange = false;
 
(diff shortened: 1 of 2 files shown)


EvilHero90:
this would close #2570 just doing this so it gets linked through.

src/spells.cpp
```diff
@@ -283,6 +286,9 @@ bool CombatSpell::castSpell(Creature* creature)
 
 bool CombatSpell::castSpell(Creature* creature, Creature* target)
 {
+	if (!creature || !target) {
```
djarek: This is not a proper fix. `creature` can never be `nullptr`, because it's always passed `this` in `Monster`. 
`target` can only be `nullptr` if a `Monster`'s target changes to `nullptr` when casting a spell. The proper fix would be to recheck here: https://github.com/otland/forgottenserver/blob/master/src/monster.cpp#L778 if the target changed and is still valid.

raymondtfr: @djarek I don't know if this is considered to be more of a user error more than it is considered a bug, but the server crashes when you pass a ``Combat`` object instead of ``Creature`` in ``combat:execute(creature, variant)``, e.g. ``combat:execute(combat, variant)``.

If you pass other values, or if ``Creature`` is ``nil`` it won't crash or anything, but will crash if you pass a ``Combat`` object.

It'll only crash if you also include ``variant`` in the function.

nekiro: Combat.execute(combat, creature, variant) does not crash

ninjalulz: That's not what he implied either, `combat:execute(combat, variant)` would be the equivalent to `Combat.execute(combat, combat, variant)`

djarek: @raymondtfr it should not crash, but I don't think it's related to this PR.

raymondtfr: >it should not crash, but I don't think it's related to this PR.

I think I'm having déjà vu cause I think I'd already mentioned this sometime ago in another PR.

Should I open an issue for that? Thanks for the reply.

djarek: Yeah, please open an issue if it still occurs. The lua interface should (ideally) not cause crashes.

marcinho1994: Now analyzing the calls from the beginning it turns out that this method is called directly from monster.cpp, so the creature will always be valid. I had not done or analyzed this before, because for some reason I thought this method could be called not just by monster.cpp.
I'll make those changes when I get back from traveling.

ranisalt:
Maybe we should use more references and less pointers. Checking for `nullptr` everywhere is very, very weak.

nekiro:
> Maybe we should use more references and less pointers. Checking for `nullptr` everywhere is very, very weak.

or more smart pointers

src/monster.cpp
```diff
@@ -775,7 +775,7 @@ void Monster::doAttacking(uint32_t interval)
 	for (const spellBlock_t& spellBlock : mType->info.attackSpells) {
 		bool inRange = false;
 
-		if (canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
+		if (attackedCreature && canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
```
djarek: If `attackedCreature` becomes `nullptr`, there's no point in iterating further:
```suggestion
		if (attackedCreature = nullptr) {
			break;
		}
		if (canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
```

nekiro: Why not if (!attackedCreature) ?

djarek: Just a personal preference, there's no semantic difference.

marcinho1994: truth, changes already made. Thanks Djarek!

src/monster.cpp
```diff
@@ -774,7 +774,9 @@ void Monster::doAttacking(uint32_t interval)
 
 	for (const spellBlock_t& spellBlock : mType->info.attackSpells) {
 		bool inRange = false;
-
```
DSpeichert: This `if` should be separated by empty lines above and below like the rest of the code. Don't change the formatting.

PR #2583, opened 2019-05-09
weapon __newindex function

EvilHero90:
- Forgot to include this in the weapons handling commit


TheSumm:
I still don't understand why the metamethods are part of compat.lua.
Since they are a core part of the revscript system, it might be better to move them to `lib/core/scripts.lua` or `lib/core/scriptsystem.lua` or another fitting name for that matter.

PR #2584, opened 2019-05-10
Revscriptsys Spells

EvilHero90:
The basics are done, spells can be created by either.

instant:
```lua
local conjureRune = Spell(SPELL_INSTANT)

function conjureRune.onCastSpell(creature, variant)
	return creature:conjureItem(2260, 2304, 4)
end

conjureRune:name("Great Fireball")
conjureRune:words("adori mas flam")
conjureRune:level(30)
conjureRune:mana(530)
conjureRune:group("support")
conjureRune:soul(3)
conjureRune:isAggressive(false)
conjureRune:cooldown(2000)
conjureRune:groupCooldown(2000)
conjureRune:needLearn(false)
conjureRune:vocation("sorcerer", true)
conjureRune:vocation("master sorcerer")
conjureRune:register()
```
rune:
```lua
local combat = Combat()
combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_HEALING)
combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_MAGIC_BLUE)
combat:setParameter(COMBAT_PARAM_DISPEL, CONDITION_PARALYZE)
combat:setParameter(COMBAT_PARAM_TARGETCASTERORTOPMOST, true)
combat:setParameter(COMBAT_PARAM_AGGRESSIVE, false)

function onGetFormulaValues(player, level, magicLevel)
	local min = (level / 5) + (magicLevel * 3.2) + 20
	local max = (level / 5) + (magicLevel * 5.4) + 40
	return min, max
end

combat:setCallback(CALLBACK_PARAM_LEVELMAGICVALUE, "onGetFormulaValues")

local rune = Spell(SPELL_RUNE)

function rune.onCastSpell(creature, variant, isHotkey)
	return combat:execute(creature, variant)
end

rune:name("Test Rune")
rune:id(220)
rune:runeId(2275)
rune:level(20)
rune:magicLevel(5)
rune:needTarget(true)
rune:isAggressive(false)
rune:allowFarUse(true)
rune:charges(1)
rune:register()
```

todo:
- [x] convert all remaining `const_cast` into `dynamic_cast`
- [x] cleanup a few functions to make them smaller
- [x] `InstantSpell` and `RuneSpell` lua functions need to return nil, if they are not used appropriately instead of crashing the server (ex: using a `RuneSpell` function on an `InstantSpell`).

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1837,6 +1837,9 @@ void LuaScriptInterface::registerFunctions()
 
 	registerEnum(MAX_LOOTCHANCE)
 
+	registerEnum(SPELL_INSTANT)
+	registerEnum(SPELL_RUNE)
+
 	// _G
 	registerGlobalVariable("INDEX_WHEREEVER", INDEX_WHEREEVER);
 	registerGlobalBoolean("VIRTUAL_PARENT", true);
@@ -2764,6 +2767,32 @@ void LuaScriptInterface::registerFunctions()
 	// Spells
 	registerClass("Spell", "", LuaScriptInterface::luaSpellCreate);
 	registerMetaMethod("Spell", "__eq", LuaScriptInterface::luaUserdataCompare);
+
+	registerMethod("Spell", "onCastSpell", LuaScriptInterface::luaSpellOnCastSpell);
+	registerMethod("Spell", "register", LuaScriptInterface::luaSpellRegister);
+	registerMethod("Spell", "name", LuaScriptInterface::luaSpellName);
+	registerMethod("Spell", "id", LuaScriptInterface::luaSpellId);
+	registerMethod("Spell", "group", LuaScriptInterface::luaSpellGroup);
+	registerMethod("Spell", "cooldown", LuaScriptInterface::luaSpellCooldown);
+	registerMethod("Spell", "groupCooldown", LuaScriptInterface::luaSpellGroupCooldown);
+	registerMethod("Spell", "level", LuaScriptInterface::luaSpellLevel);
+	registerMethod("Spell", "magicLevel", LuaScriptInterface::luaSpellMagicLevel);
+	registerMethod("Spell", "mana", LuaScriptInterface::luaSpellMana);
+	registerMethod("Spell", "manaPercent", LuaScriptInterface::luaSpellManaPercent);
+	registerMethod("Spell", "soul", LuaScriptInterface::luaSpellSoul);
+	registerMethod("Spell", "range", LuaScriptInterface::luaSpellRange);
+	registerMethod("Spell", "premium", LuaScriptInterface::luaSpellPremium);
+	registerMethod("Spell", "enabled", LuaScriptInterface::luaSpellEnabled);
+	registerMethod("Spell", "needTarget", LuaScriptInterface::luaSpellNeedTarget);
+	registerMethod("Spell", "needWeapon", LuaScriptInterface::luaSpellNeedWeapon);
+	registerMethod("Spell", "needLearn", LuaScriptInterface::luaSpellNeedLearn);
+	registerMethod("Spell", "selfTarget", LuaScriptInterface::luaSpellSelfTarget);
+	registerMethod("Spell", "blocking", LuaScriptInterface::luaSpellBlocking);
+	registerMethod("Spell", "aggressive", LuaScriptInterface::luaSpellAggressive);
+	registerMethod("Spell", "vocation", LuaScriptInterface::luaSpellVocation);
+
+	// only for instant spells
+	registerMethod("Spell", "words", LuaScriptInterface::luaSpellWords);
 
 	registerMethod("Spell", "getManaCost", LuaScriptInterface::luaSpellGetManaCost);
 	registerMethod("Spell", "getSoulCost", LuaScriptInterface::luaSpellGetSoulCost);
@@ -13777,7 +13806,43 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
 // Spells
 int LuaScriptInterface::luaSpellCreate(lua_State* L)
 {
-	// Spell(words, name or id)
+	// Spell(words, name or id) to get an existing spell
+	// Spell(type) ex: Spell(SPELL_INSTANT) or Spell(SPELL_RUNE) to create a new spell
+	SpellType_t type = getNumber<SpellType_t>(L, 2);
+	if (type) {
+		switch (type) {
+			case SPELL_INSTANT: {
+				InstantSpell* spell = new InstantSpell(getScriptEnv()->getScriptInterface());
+				if (spell) {
+					spell->fromLua = true;
+					pushUserdata<InstantSpell>(L, spell);
+					setMetatable(L, -1, "Spell");
+					spell->type = SPELL_INSTANT;
+				} else {
+					lua_pushnil(L);
+				}
+				break;
+			}
+			case SPELL_RUNE: {
+				RuneSpell* spell = new RuneSpell(getScriptEnv()->getScriptInterface());
+				if (spell) {
+					spell->fromLua = true;
+					pushUserdata<RuneSpell>(L, spell);
+					setMetatable(L, -1, "Spell");
+					spell->type = SPELL_RUNE;
+				} else {
+					lua_pushnil(L);
+				}
+				break;
+			}
+			default: {
+				lua_pushnil(L);
+				break;
+			}
+		}
+		return 1;
+	}
+
 	InstantSpell* spell = nullptr;
 	if (isNumber(L, 2)) {
 		spell = g_spells->getInstantSpellById(getNumber<uint32_t>(L, 2));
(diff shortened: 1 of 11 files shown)


EvilHero90:
time for a review

raymondtfr:
Maybe you could add support for constructing new spells by string ``Spell(string)``, e.g. ``Spell("instant")`` and ``Spell("rune")``.?

EvilHero90:
> 
> 
> Maybe you could add support for constructing new spells by string `Spell(string)`, e.g. `Spell("instant")` and `Spell("rune")`?

Yea sure why not.

soul4soul:
revscriptsys monsters should support this spell system and drop the current system. Now that everything is so cleanly created and registered in lua there is no reason for 2 spells creation systems to exist. Monsters that need spells would use this format.

Znote:
> revscriptsys monsters should support this spell system and drop the current system. Now that everything is so cleanly created and registered in lua there is no reason for 2 spells creation systems to exist. Monsters that need spells would use this format.

We will wait until TFS 2.0 or at least 1.4 before breaking backward compatibility.

TheSumm:
> 
> 
> Maybe you could add support for constructing new spells by string `Spell(string)`, e.g. `Spell("instant")` and `Spell("rune")`?

I see absolutely no benefit in allowing that.

raymondtfr:
The benefit is just aesthetic, makes it clearer . You can now both use a constant or a string.

soul4soul:
> 
> 
> > revscriptsys monsters should support this spell system and drop the current system. Now that everything is so cleanly created and registered in lua there is no reason for 2 spells creation systems to exist. Monsters that need spells would use this format.
> 
> We will wait until TFS 2.0 or at least 1.4 before breaking backward compatibility. The current revscriptsys implementation is an alternative, not a replacement for how to code in the current iteration of TFS.

I'm not saying to touch or remove anything with the TFS XML monsters.  What I purpose is that the revscriptsys monsters should be updated to only use the spell system being proposed in this PR instead of what is implemented now.

EvilHero90:
> 
> 
> > > revscriptsys monsters should support this spell system and drop the current system. Now that everything is so cleanly created and registered in lua there is no reason for 2 spells creation systems to exist. Monsters that need spells would use this format.
> > 
> > 
> > We will wait until TFS 2.0 or at least 1.4 before breaking backward compatibility. The current revscriptsys implementation is an alternative, not a replacement for how to code in the current iteration of TFS.
> 
> I'm not saying to touch or remove anything with the TFS XML monsters. What I purpose is that the revscriptsys monsters should be updated to only use the spell system being proposed in this PR instead of what is implemented now.

The idea itself is good and I already had plans later on to do that, same with Loot actually, but bringing spells under one roof isn't that easy and might add some complexity for people trying to do spells for monsters then, so I'll have to carefully do that once everything stands and works.

src/luascript.h
```diff
@@ -1335,11 +1335,43 @@ class LuaScriptInterface
...
+		static int luaSpellBlocking(lua_State* L);
+		static int luaSpellAggressive(lua_State* L);
+		static int luaSpellVocation(lua_State* L);
+
+		// only for InstantSpells
+		static int luaSpellWords(lua_State* L);
+		static int luaSpellNeedDirection(lua_State* L);
+		static int luaSpellHasParams(lua_State* L);
+		static int luaSpellHasPlayerNameParam(lua_State* L);
+		static int luaSpellNeedCasterTargetOrDirection(lua_State* L);
```
raymondtfr: I think you should rename this to ``luaSpellCastOnTargetOrDirection`` since it means that it can be cast indirectly (no target selected) to the tile the caster is facing or directly onto the target (target is selected).

src/luascript.cpp
```diff
@@ -13777,66 +13812,952 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+	} else {
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+// only for InstantSpells
+int LuaScriptInterface::luaSpellNeedCasterTargetOrDirection(lua_State* L)
+{
+	// spell:needCasterTargetOrDirection(bool)
```
raymondtfr: This should be named to ``spell:castOnTargetOrDirection``.

src/spells.h
```diff
@@ -271,13 +424,29 @@ class RuneSpell final : public Action, public Spell
 		uint16_t getRuneItemId() const {
 			return runeId;
 		}
+		void setRuneItemId(uint16_t i) {
+			runeId = i;
+		}
+		bool getHasCharges() const {
```
raymondtfr: Can't you reuse ``charges``? e.g. ``charges > 0``.

src/spells.h
```diff
@@ -271,13 +424,29 @@ class RuneSpell final : public Action, public Spell
...
+			return hasCharges;
+		}
+		void setHasCharges(bool c) {
```
raymondtfr: Maybe you could also use ``hasCharges = charges > 0``?

src/luascript.cpp
```diff
@@ -13777,66 +13812,952 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+	} else {
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+// only for InstantSpells
+int LuaScriptInterface::luaSpellNeedDirection(lua_State* L)
+{
+	// spell:needDirection(bool)
```
raymondtfr: This should also probably be renamed to ``spell:castOnDirection`` because it is not that it needs direction but instead asserting (imperative) that it will cast the spell towards the direction the caster is facing, so it is not like it is a requirement.

src/luascript.cpp
```diff
@@ -13777,66 +13812,952 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+		}
+	} else {
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+// only for InstantSpells
+int LuaScriptInterface::luaSpellNeedDirection(lua_State* L)
+{
```
raymondtfr: ~This should also probably be renamed to spell:castOnDirection because it is not that it needs direction but instead asserting (imperative) that it will cast the spell towards the direction the caster is facing, so it is not like it is a requirement.~

Since this works both ways ``getting`` and ``setting`` it only makes sense to rename this and all others functions to just their main words, like ``spell:direction()``, ``spell:params()``, ``spell:blockingWalls()`` etc, like you mentioned in an issue if I recall correctly.

It would not be pretty to rename it to something like ``spell:setDirectionOrGetDirection`` haha.

src/luascript.cpp
```diff
@@ -13777,66 +13812,952 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
 // Spells
 int LuaScriptInterface::luaSpellCreate(lua_State* L)
 {
-	// Spell(words, name or id)
-	InstantSpell* spell = nullptr;
+	// Spell(words, name or id) to get an existing spell
+	// Spell(type) ex: Spell(SPELL_INSTANT) or Spell(SPELL_RUNE) to create a new spell
```
djarek: Wouldn't it be better if there was a separate function that returns existing spells? This way it's explicit in user code.

src/luascript.cpp
EvilHero90: The problem would be that it gives the implication that something like `:direction` would set or get the `direction` instead of expecting/returning a `boolean` value.
`is` or `need` tells pretty good what we expect to get/set from the function.
We should rather rename all boolean functions to this standard and have it as get/set already.

src/luascript.cpp
EvilHero90: That would break backward compatibility as `Spell` was already existing and was a returning function for instant Spells actually.
I think we should start to unify functions from now on, so we are not relying on something like `Item`, `ItemType` and keep it as `Item` so we rather have it returning whatever we push inside already.

ixeduh:
Why convert const_cast to dynamic_cast?

EvilHero90:
> 
> 
> Why convert const_cast to dynamic_cast?

Because `dynamic_cast` handles base/derived classes better, it leaves a `nullptr` if what you try to cast on isn't the object type you expect, it gives overhead but it also gives safety in this case of usage.
[link](https://www.quora.com/How-do-you-explain-the-differences-among-static_cast-reinterpret_cast-const_cast-and-dynamic_cast-to-a-new-C++-programmer)

src/spells.h
EvilHero90: done in [5296c38](https://github.com/otland/forgottenserver/pull/2584/commits/5296c38673bb57e34bf5fe3769265a42a8d1dc95) thanks

src/spells.h
EvilHero90: done in [5296c38](https://github.com/otland/forgottenserver/pull/2584/commits/5296c38673bb57e34bf5fe3769265a42a8d1dc95) thanks

src/luascript.cpp
```diff
@@ -13777,66 +13812,927 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+	} else {
+		lua_pushnil(L);
 	}
+	return 1;
+}
 
+int LuaScriptInterface::luaSpellId(lua_State* L)
+{
+	// spell:id(id)
+	Spell* spell = dynamic_cast<InstantSpell*>(getUserdata<InstantSpell>(L, 1));
```
djarek: This `dynamic_cast` seems redundant.

src/luascript.cpp
```diff
@@ -13777,66 +13812,927 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+				} else {
+					lua_pushnil(L);
+				}
+				break;
+			}
+			case SPELL_RUNE: {
+				RuneSpell* spell = new RuneSpell(getScriptEnv()->getScriptInterface());
+				if (spell) {
+					spell->fromLua = true;
+					pushUserdata<RuneSpell>(L, spell);
```
djarek: Technically speaking, pushing a userdata of `RuneSpell` type and retrieving a different type (even if it's the base of `RuneSpell`) is UB, because userdata is stored/retrieved from a `void*`. You should make sure that the same type is pushed and retrieved.

src/luascript.cpp
```diff
@@ -13777,66 +13812,927 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+	if (isString(L, 2)) {
+		std::string tmp = asLowerCaseString(getString(L, 2));
+		if (tmp == "instant") {
+			type = SPELL_INSTANT;
+		} else if (tmp == "rune") {
+			type = SPELL_RUNE;
+		}
+	}
+
+	if (type) {
```
djarek: This check seems a little weird, considering the fact we have a `switch` below, IMO you should just add a case to the switch that handles that value explicitly. This gets rid of one level of nesting, thus improving readability.

src/luascript.cpp
EvilHero90: that's why I'm using `dynamic_cast` on a direct `Spell` pointer, the `Spell` pointer itself gives UB, when casting onto `InstantSpell` or `RuneSpell` I get the right values just fine, I couldn't find another way without splitting `RuneSpells` and `InstantSpells` entirely, if you have something better in mind just let me know.

src/luascript.cpp
EvilHero90: well I tried to initialize as `InstantSpell* without `dynamic_cast` and if it was not the right `spellType` I tried to `dynamic_cast` sideways onto `RuneSpell*` that however didn't work as intended and gave sometimes UB.
I didn't feel like making more pointers, would just bulk up the entire code actually, yet again if you have a better solution for this then feel free to let me know

src/luascript.cpp
EvilHero90: Well as this function consists of either creating spells or returning spell userdata, I'm just checking here beforehand if it's a creation type and then use the switch to work accordingly, if it's not a type of creating it should skip the switch anyway.

src/luascript.cpp
djarek: You should always push and get a `Spell` and cast afterwards:
```
pushUserdata<Spell>(L, rune_spell);
Spell* sp = getUserdata<Spell>(L, -1);
```

src/luascript.cpp
djarek: Sure it's equivalent, but using just the `switch` is easier to read, because it has one level of indentation less.

src/luascript.cpp
djarek: The `dynamic_cast` doesn't save the day here if the pointer we retrieve is not an `InstantSpell`, the pointer we get is an "invalid pointer" (http://eel.is/c++draft/expr.static.cast#2).
`dynamic_cast<T*>(p)` requires that `p` is either `nullptr` or a valid pointer.

Here's an [example](https://godbolt.org/z/bfHaT0). Both clang and gcc optimize out the `if` check, because of UB caused by the `static_cast` in line 21.

src/luascript.cpp
```diff
@@ -13777,66 +13812,927 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+		switch (type) {
+			case SPELL_INSTANT: {
+				InstantSpell* spell = new InstantSpell(getScriptEnv()->getScriptInterface());
+				if (spell) {
```
djarek: This is always true (`new` throws, never returns `nullptr`).

src/luascript.cpp
```diff
@@ -13777,66 +13812,927 @@ int LuaScriptInterface::luaPartySetSharedExperience(lua_State* L)
...
+					setMetatable(L, -1, "Spell");
+					spell->spellType = SPELL_INSTANT;
+				} else {
+					lua_pushnil(L);
+				}
+				break;
+			}
+			case SPELL_RUNE: {
+				RuneSpell* spell = new RuneSpell(getScriptEnv()->getScriptInterface());
+				if (spell) {
```
djarek: This is always true (`new` throws, never returns `nullptr`).

EvilHero90: fixed in [9e888a9](https://github.com/otland/forgottenserver/pull/2584/commits/9e888a96bb55d97ef102cb18a8c91e78178c8ecf)

src/luascript.cpp
EvilHero90: fixed in [9e888a9](https://github.com/otland/forgottenserver/pull/2584/commits/9e888a96bb55d97ef102cb18a8c91e78178c8ecf)

src/luascript.cpp
EvilHero90: I cannot possibly remove it because when parameter is a string then type = 0 which results that the switch will throw it as `SPELL_UNDEFINED` which we don't want.

src/luascript.cpp
EvilHero90: fixed in [9e888a9](https://github.com/otland/forgottenserver/pull/2584/commits/9e888a96bb55d97ef102cb18a8c91e78178c8ecf)

src/luascript.cpp
EvilHero90: fixed in [9e888a9](https://github.com/otland/forgottenserver/pull/2584/commits/9e888a96bb55d97ef102cb18a8c91e78178c8ecf)

src/luascript.cpp
djarek:
```	switch (type) {
		case SPELL_INSTANT: {
			InstantSpell* spell = new InstantSpell(getScriptEnv()->getScriptInterface());
			if (spell) {
				spell->fromLua = true;
				pushUserdata<InstantSpell>(L, spell);
				setMetatable(L, -1, "Spell");
				spell->spellType = SPELL_INSTANT;
			} else {
				lua_pushnil(L);
			}
			return 1;
		}
		case SPELL_RUNE: {
			RuneSpell* spell = new RuneSpell(getScriptEnv()->getScriptInterface());
			if (spell) {
				spell->fromLua = true;
				pushUserdata<RuneSpell>(L, spell);
				setMetatable(L, -1, "Spell");
				spell->spellType = SPELL_RUNE;
			} else {
				lua_pushnil(L);
			}
			return 1;
		}
		case SPELL_UNDEFINED: {
			break;
		}
		default: {
			lua_pushnil(L);
			return 1;
		}
	}
```

EvilHero90: that doesn't work either, if I push now like `Spell(220)` it'll go to default and give me nil, which should actually give me the userdata of spell with id 220
I don't see one check with more indentation worse if it does exactly the job it should.

djarek: I think you're confused what `if (type) {` does...
It checks whether the enum is greater than 0. So `Spell(220)` would cause that condition to evaluate to `true`. Then it goes into the `switch` and jumps into the `default` case. Both snippets are equivalent.

EvilHero90: I'll just remove the switch it's just `SPELL_INSTANT` and `SPELL_RUNE` anyway and rather check at the beginning if `lua_gettop(L)` returns 1 and then `nil` it with an error message

EvilHero90:
final review please, but it's at the stage to get merged now.

EvilHero90:
LGTM

EvilHero90:
can we finaly get this merged, it has been 4 days since approvement.

raymondtfr:
Would it make sense to ask for the ability to be able to do something like:
```lua
local spell = Spell("instant")
spell:vocation("knight", "elite knight", "druid", "elder druid", "sorcerer", "master sorcerer", "paladin", "royal paladin")
```
instead of:
```lua
local spell = Spell("instant")
spell:vocation("knight")
spell:vocation("elite knight")
spell:vocation("druid")
spell:vocation("elder druid")
spell:vocation("sorcerer")
spell:vocation("master sorcerer")
spell:vocation("paladin")
spell:vocation("royal paladin")
```
when there is no need to make use of third argument ``showInDescription`` since they can only be used through a ``RuneSpell`` only and not from ``InstantSpell`` class?

@EvilHero90

raymondtfr:
Am I just confused or are these functions logic swapped?
```C++
int LuaScriptInterface::luaSpellCooldown(lua_State* L)
{
	// spell:cooldown(primaryGroupCd[, secondaryGroupCd])
	Spell* spell = getUserdata<Spell>(L, 1);
	if (spell) {
		if (lua_gettop(L) == 1) {
			lua_pushnumber(L, spell->getCooldown());
			lua_pushnumber(L, spell->getSecondaryCooldown());
			return 2;
		} else if (lua_gettop(L) == 2) {
			spell->setCooldown(getNumber<uint32_t>(L, 2));
			pushBoolean(L, true);
		} else {
			spell->setCooldown(getNumber<uint32_t>(L, 2));
			spell->setSecondaryCooldown(getNumber<uint32_t>(L, 3));
			pushBoolean(L, true);
		}
	} else {
		lua_pushnil(L);
	}
	return 1;
}
```
```C++
int LuaScriptInterface::luaSpellGroupCooldown(lua_State* L)
{
	// spell:groupCooldown(cooldown)
	Spell* spell = getUserdata<Spell>(L, 1);
	if (spell) {
		if (lua_gettop(L) == 1) {
			lua_pushnumber(L, spell->getGroupCooldown());
		} else {
			spell->setGroupCooldown(getNumber<uint32_t>(L, 2));
			pushBoolean(L, true);
		}
	} else {
		lua_pushnil(L);
	}
	return 1;
}
```
The spell metatable function ``spell:groupCooldown(cooldown)`` and its parameters should actually be ``spell:groupCooldown(primaryGroupCd[, secondaryGroupCooldown])`` and on the other hand ``spell:cooldown(cooldown)`` (cooldown that is bound to the spell itself "exhaustion"). Right?

@EvilHero90

EvilHero90:
> 
> 
> Would it make sense to ask for the ability to be able to do something like:
> 
> ```lua
> local spell = Spell("instant")
> spell:vocation("knight", "elite knight", "druid", "elder druid", "sorcerer", "master sorcerer", "paladin", "royal paladin")
> ```
> 
> instead of:
> 
> ```lua
> local spell = Spell("instant")
> spell:vocation("knight")
> spell:vocation("elite knight")
> spell:vocation("druid")
> spell:vocation("elder druid")
> spell:vocation("sorcerer")
> spell:vocation("master sorcerer")
> spell:vocation("paladin")
> spell:vocation("royal paladin")
> ```
> 
> when there is no need to make use of third argument `showInDescription` since they can only be used through a `RuneSpell` only and not from `InstantSpell` class?
> 
> @EvilHero90

How about:
```lua
spell:vocation("knight;paladin", "elite knight;royal paladin")
```
First param will get shown in description for `RuneSpells` by default and on an `InstantSpell` it could look like:
```lua
spell:vocation("sorcerer;druid;master sorcerer;elder druid")
```

> 
> 
> Am I just confused or are these functions logic swapped?
> 
> ```c++
> int LuaScriptInterface::luaSpellCooldown(lua_State* L)
> {
> 	// spell:cooldown(primaryGroupCd[, secondaryGroupCd])
> 	Spell* spell = getUserdata<Spell>(L, 1);
> 	if (spell) {
> 		if (lua_gettop(L) == 1) {
> 			lua_pushnumber(L, spell->getCooldown());
> 			lua_pushnumber(L, spell->getSecondaryCooldown());
> 			return 2;
> 		} else if (lua_gettop(L) == 2) {
> 			spell->setCooldown(getNumber<uint32_t>(L, 2));
> 			pushBoolean(L, true);
> 		} else {
> 			spell->setCooldown(getNumber<uint32_t>(L, 2));
> 			spell->setSecondaryCooldown(getNumber<uint32_t>(L, 3));
> 			pushBoolean(L, true);
> 		}
> 	} else {
> 		lua_pushnil(L);
> 	}
> 	return 1;
> }
> ```
> 
> ```c++
> int LuaScriptInterface::luaSpellGroupCooldown(lua_State* L)
> {
> 	// spell:groupCooldown(cooldown)
> 	Spell* spell = getUserdata<Spell>(L, 1);
> 	if (spell) {
> 		if (lua_gettop(L) == 1) {
> 			lua_pushnumber(L, spell->getGroupCooldown());
> 		} else {
> 			spell->setGroupCooldown(getNumber<uint32_t>(L, 2));
> 			pushBoolean(L, true);
> 		}
> 	} else {
> 		lua_pushnil(L);
> 	}
> 	return 1;
> }
> ```
> 
> The spell metatable function `spell:groupCooldown(cooldown)` and its parameters should actually be `spell:groupCooldown(primaryGroupCd[, secondaryGroupCooldown])` and on the other hand `spell:groupCooldown(cooldown)` should be `spell:cooldown(cooldown)` (cooldown that is bound to the spell itself "exhaustion"). Right?
> 
> @EvilHero90

Yes you are correctly idk how I could miss that, thanks for telling, will open a pr soon.

raymondtfr:
Maybe you could add a new core Lua file as ``data/lib/core/spell.lua`` and redefine ``Spell:vocation(self)`` as:
```lua
function Spell:vocation(...)
	-- It would first require "spell:getType()" being added first
	if self:getType() == SPELL_INSTANT then
		for _, vocation in ipairs({...}) do
			self:vocation(vocation)
		end
	else
		local vocation, showInDescription = ...
		self:vocation(vocation, showInDescription)
	end
end
```
In any case, I think ``spell:vocation(vocation, showInDescription)`` in sources should be updated to not take into account the third argument (showInDescription) if the spell type is ``instant`` since such description can only be shown in runes.

If what I suggested above or in my previous reply is not a good approach, yes, it would be good to have what you suggested in case my previous suggestions are not feasible or adds too much work. I just think using a semicolon may not be pretty (one of the things 1.0 differed from 0.4) for readability.

PR #2585, opened 2019-05-10
Missing drop loot changes

joseluis2g:
Fixed some missing changes related to stackable items, reported here: https://otland.net/threads/mobs-only-drop-one-of-each-item.264492/#post-2556086

### DIFF
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -10,8 +10,8 @@ function Container.createLootItem(self, item)
 	local itemCount = 0
 	local randvalue = getLootRandom()
 	if randvalue < item.chance then
-		if ItemType(item):isStackable() then
-			itemCount = randvalue % item.countmax + 1
+		if ItemType(item.itemId):isStackable() then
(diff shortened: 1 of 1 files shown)


data/lib/core/container.lua
```diff
@@ -10,8 +10,8 @@ function Container.createLootItem(self, item)
 	local itemCount = 0
 	local randvalue = getLootRandom()
 	if randvalue < item.chance then
-		if ItemType(item):isStackable() then
-			itemCount = randvalue % item.countmax + 1
+		if ItemType(item.itemId):isStackable() then
```
raymondtfr: The field name for an item's id is ``itemid`` not ``itemId`` so it becomes ``item.itemid`` and not ``item.itemId``.

joseluis2g: we are using getLoot function to call this new table.
you can check this here:
https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L12736
@raymondtfr

PR #2588, opened 2019-05-11
Revert "Fixes wrong position"

TheSumm:
Reverts otland/forgottenserver#2587

#2587 has been merged faster than anyone could possibly check it.
It breaks onStepOut scripts. The onStepOut scripts will trigger when you step onto the tile now.
Also, note that changing the value of the parameter `fromPosition` will break scripts that rely on it.


Znote:
Ahh my bad, it looked like a simple fix that wouldn't require proper testing so I pushed it ahead.

Issue #2586, opened 2019-05-10
onStepOut not passing correct position


wo11ven:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
```
function onStepOut(creature, item, position, fromPosition)
    
local toPositionTile = Tile(position):getGround()

local fromPositionTile = Tile(fromPosition):getGround()

print(toPositionTile.itemid)

print(fromPositionTile.itemid)

print(position == fromPosition)

end
```

### Expected behaviour
The values should be different if the origin and destination tiles are different.

### Actual behaviour
The values are the same even if origin and destination tiles are different .

### Environment
Latest commits.

wo11ven:
The same logic is used for both stepIn and stepOut, but it seems this doesn't work for stepOut:
```
	LuaScriptInterface::pushPosition(L, pos);
	LuaScriptInterface::pushPosition(L, creature->getLastPosition());
```
They end up being the same.

Znote:
Tagging @EvilHero90 as he has some recent experience with movements due to revscriptsys.

EvilHero90:
I think this problem has already been asked a while around aswell.
The problem is that in Tile.cpp `Tile::postRemoveNotification` we push the `tile` the player is standing on right now -> `g_moveEvents->onCreatureMove(creature, this, MOVE_EVENT_STEP_OUT);` `this` <- is the tile.
So when we try to get the `const Position& pos = tile->getPosition();` in `uint32_t MoveEvents::onCreatureMove` then it gives us the `tile` we are moving away from, so `fromPosition` and `position` are equal.
I wouldn't consider this a bug actually, it's just a limitation of `Tile::postRemoveNotification` not beeing able to get the new `tile` in the first place.

Znote:
Fix commit had some issues and got reverted, this Lua change in the steps to reproduce might be a workaround?
``lua
function onStepOut(creature, item, position, fromPosition)
	local toPositionTile = Tile(creature:getPosition()):getGround()
	local fromPositionTile = Tile(fromPosition):getGround()
	print(toPositionTile.itemid)
	print(fromPositionTile.itemid)
	print(creature:getPosition() == fromPosition)
end
```

wo11ven:
[...]

PR #2589, opened 2019-05-12
onStepOut position fix

EvilHero90:
- now `onStepOut` returns `nextPosition` instead of `fromPosition` `onStepOut(creature, item, position, nextPosition)`
- `position` -> where the creature steps off
- `nextPosition` -> where the creature steps onto


TheSumm:
`MoveEvents::onCreatureMove` is called after the creature has been moved to the new tile.
It should not be necessary to introduce `nextPosition`, since `creature->getPosition()` can be used.
By adding nextPosition/toPosition as the fourth parameter, it can be ensured that no scripts will break.

wo11ven:
You mean lua Scripts that people have created or existing scripts in this repo? If we add a fifth variable, we will have a "dead" one that might cause confusion.

EvilHero90:
I can only agree with what @wo11ven here just said it would cause more confusion.
Well, we have to agree that the way `onStepOut` is right now is not correct so it should be considered as a bug. It should have it's parameters fixed.
Vanilla files have been corrected, even tho we never used the `fromPosition` parameter somewhere anyway.
I don't see this really as a "breaking backwards" problem since you could say one of either parameters didn't return what the user expected it to be in the first place.

TheSumm:
IMO it's not confusing at the moment. The parameters do exactly what they say right now.
`position` is the position of the tile/item which the script is assigned to.
`fromPosition` is the position where you came from.
Granted `fromPosition` is not very useful in case of `onStepOut`, since it is identical to `position`. Nevertheless, it is exactly what I would expect from it.

The misconception in #2586 is, that `position` is referring to the new position, which is does not and would not after any of these changes.
Therefore, I see no reason to break backwards compatibility in order to change something which works as intended/expected.
For now I would either leave it as it is, since you can already use `creature:getPosition()` to obtain the new position, OR introduce `toPosition` as a new parameter to both `onStepIn` and `onStepOut`.

Once we are willing to break backwards compatibility, it might be the best idea to get rid of `position` and just use `fromPosition` and `toPosition`.

EvilHero90:
I don't mind about `fromPos` and `toPos` those where my first intentions anyway but as it would have caused a complete re work on all scripts with `onStepIn/onStepOut` I have refrained from it.
How about I'll make the changes according to `fromPos` and `toPos` and once 1.3 is released, we should merge this.

djarek (dismissed):

src/movement.cpp
```diff
@@ -940,8 +940,8 @@ bool MoveEvent::executeStep(Creature* creature, Item* item, const Position& pos)
 	LuaScriptInterface::pushUserdata<Creature>(L, creature);
 	LuaScriptInterface::setCreatureMetatable(L, -1, creature);
 	LuaScriptInterface::pushThing(L, item);
-	LuaScriptInterface::pushPosition(L, pos);
-	LuaScriptInterface::pushPosition(L, creature->getLastPosition());
+	LuaScriptInterface::pushPosition(L, eventType == MOVE_EVENT_STEP_OUT ? creature->getNextPos() : pos);
```
TheSumm:
```suggestion
	LuaScriptInterface::pushPosition(L, eventType == MOVE_EVENT_STEP_OUT ? creature->getPosition() : pos);
```

src/creature.h
```diff
@@ -507,6 +514,7 @@ class Creature : virtual public Thing
 		Outfit_t defaultOutfit;
 
 		Position lastPosition;
+		Position nextPosition;
```
TheSumm: `nextPosition` is not needed

PR #2590, opened 2019-05-12
Update copyright for 2019

LucasCPrazeres:


ranisalt:
[We should not do that every year](https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated)

> The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work.
>
> Therefore, if the claim is based on common law copyright (not formally registered), then the date should be the date of first publication.

ninjalulz:
That's mildly interesting. yet we have done it every year since 2014 👍

PR #2593, opened 2019-05-20
Do not pass product of skills rate by skill tries/mana spent when calling addSkillTries/addManaSpent

raymondtfr:


DSpeichert:
Here is the definition (it does not multiply itself):
https://github.com/otland/forgottenserver/blob/b7ef46dfc7ac4fad1bcea5e3604c9ef8c4350864/src/player.cpp#L1523-L1525


Here is potentially incorrect usage:
https://github.com/otland/forgottenserver/blob/b7ef46dfc7ac4fad1bcea5e3604c9ef8c4350864/src/weapons.cpp#L429-L432

https://github.com/otland/forgottenserver/blob/b7ef46dfc7ac4fad1bcea5e3604c9ef8c4350864/src/spells.cpp#L772-L777

https://github.com/otland/forgottenserver/blob/bcb86eac0655c58b8dc9a35c652f8953a0edd033/data/spells/scripts/support/convince_creature.lua#L32

https://github.com/otland/forgottenserver/blob/bcb86eac0655c58b8dc9a35c652f8953a0edd033/data/spells/scripts/support/summon_creature.lua#L46

https://github.com/otland/forgottenserver/blob/7de38e3e358844068ead0c940e07b0ed64c8f093/data/spells/lib/spells.lua#L288

Here is CORRECT example:
https://github.com/otland/forgottenserver/blob/f0f1b0d12523f243f46a159b25b2fc5ce300823e/data/lib/compat/compat.lua#L577

So it looks to me like most of magic level multiplier logic is broken and all of it should be fixed.

raymondtfr:
So what do you think we should do? Update the one from ``Player`` metatable to work like the one in ``compat`` and update the ``compat`` one accordingly, or leave it as it is and let the user decide that by doing it explictly?

TheSumm:
Actually `doPlayerAddManaSpent` is wrong, since skill tries are multiplied in https://github.com/otland/forgottenserver/blob/b7ef46dfc7ac4fad1bcea5e3604c9ef8c4350864/data/events/scripts/player.lua#L269-L278

raymondtfr:
Does that mean that we don't need to use ``configManager.getNumber(configKeys.RATE_MAGIC)`` when adding skill tries or mana spent?

DSpeichert:
Clearly I missed that it's mutated in the "hook". In that case `doPlayerAddManaSpent` is indeed wrong.

raymondtfr:
I have updated the PR to fix the problems pointed out.

@DSpeichert @TheSumm

PR #2594, opened 2019-05-26
Revscriptsys Monstertype event callbacks

EvilHero90:
This will let you be able to make event callbacks for monstertypes directly in the lua file

- `onThink(monster, interval)`
- `onAppear(monster, creature)`
- `onDisappear(monster, creature)`
- `onMove(monster, creature, fromPosition, toPosition)`
- `onSay(monster, creature, type, message)`

Those event callbacks will ofc be executed for every monster of that monstertype on your map.
In `data/monster/lua/#example.lua` is an example code on how to use it.


EvilHero90:
I'm not sure how much people have already tested this but so far I didn't encounter problems with it, so we can get this merged.

PR #2595, opened 2019-05-27
Change to correct type additem and removeitem

danielsalim1:


dudantas:
This generated a bug in onRemoveItem (apparently the solution is only related to onAddItem)

https://github.com/opentibiabr/otservbr-global/issues/1905

PR #2597, opened 2019-05-28
Fix possible query error using Visual Studio 2019

danielsalim1:
When insert to the database a value uint8_t it presents a special symbol, generating error in the console and consequently error when saving the player, this is a rather serious problem.

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -686,7 +686,7 @@ bool IOLoginData::savePlayer(Player* player)
 	query << "`posz` = " << loginPosition.getZ() << ',';
 
 	query << "`cap` = " << (player->capacity / 100) << ',';
-	query << "`sex` = " << player->sex << ',';
+	query << "`sex` = " << static_cast<uint16_t>(player->sex) << ',';
 
 	if (player->lastLoginSaved != 0) {
 		query << "`lastlogin` = " << player->lastLoginSaved << ',';


DSpeichert:
Please post the bad character code that's generated.

danielsalim1:
> Please post the bad character code that's generated.

I had to go back to get the error. When insert to the database a value uint8_t it presents a special symbol, see this photo.

Screenshot_2

ranisalt:
That special symbol is `'\x01'`, the number for male sex. With that said, please report that error upstream (VS 2019 tracker) and let us know their answer, as this is working as expected for compliant compilers.

diath:
@DSpeichert 
AFAIK the standard only mentions the size of fixed width integers, and not their underlying type, so the implementations are free to use a char type (assuming the size guarantees), in which case it matches the char overload which then converts it to the corresponding ASCII character (since typedefs are just aliases). It's not specific to MSVC either: https://godbolt.org/z/55t5GS

djarek:
@diath that's not exactly what's happening here, because this is more representatitve of what's going on in the original source: https://godbolt.org/z/7hhiGm
I suspect that in the original source, MSVC injected an implicit conversion from the `enum` to its underlying integer type, rather than a promotion to a regular (possibly unsigned) integer type.

diath:
@djarek 
Fair enough, though if I'm reading the correct section of the standard it seems that it's implementation defined behavior so it should be explicitly cast regardless (correct me if I'm wrong):
>A prvalue of an unscoped enumeration type whose underlying type is fixed (10.2) can be converted to a prvalue of its underlying type. Moreover, if integral promotion can be applied to its underlying type, a prvalue of an unscoped enumeration type whose underlying type is fixed can also be converted to a prvalueof the promoted underlying type.

djarek:
That's only part of the picture, because there's also the issue of overload resolution most likely causing the type coersion. I suspect this is caused by some (possibly unrelated) change in MSVC's stdlib implementation that affected how overload resolution works.

Issue #2603, opened 2019-06-02
Luajit include path using (vcpkg) in Windows [not a issue]


GauzStudio:
Hi, it is my first time writing an issue here so I am sorry if it  is not the right place...

I was recently compiling the sources using (vcpkg) in Windows and had to attach a include path to LUAJIT in vcpkg directory to the project config.

Just thought it might be interesting to point out that this could be possible an issue for another person too.

Thanks for the amazing project!

PR #2606, opened 2019-06-04
windows <luajit/lua.hpp> include

EvilHero90:
This fixes the problem on windows that lua.hpp cannot be found, because vcpkg by default installs luajit, which results in <luajit/lua.hpp>
#2603

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -19,7 +19,11 @@
 
 #include "otpch.h"
 
+#ifdef _WIN32
+#include <luajit/lua.hpp>
+#else
 #include <lua.hpp>
+#endif
 
 #include "configmanager.h"
 #include "game.h"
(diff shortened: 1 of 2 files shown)


marksamman:
Wouldn't this be problematic when compiling without LuaJIT on Windows?

EvilHero90:
In that sense yes, but as we currently provide/suggest luajit for vcpkg it would make more sense to change the include directories accordingly, as you'll have the `cannot find lua.hpp` problem without changing it manual, for people with prior knowledge that's not a big issue but I've seen a lot of threads about this at otland where newcomers had trouble because of that.

DSpeichert:
I'm wondering what's the reason to keep plain Lua and not just have LuaJIT as the only option.

Lyuzera:
I see it recently and I had to change manually.

djarek:
@DSpeichert LuaJIT is unmaintained.

PR #2612, opened 2019-06-25
Fix send yourself in vip list. #2611

danielsalim1:


DSpeichert:
Is this the behavior in RL Tibia?

danielsalim1:
I don't think so, but I there's no reason to add myself to the friend list... it doesn't make sense

ranisalt:
Why fix if it's not broken tho

Znote:
Think I have seen some OTs where a player could opt out of "online list", and to see if they were hidden they could add themselves to their own vip list. If their own name was red then they knew they were currently "hidden".

wgrr:
This PR is a regression; it had been reported as an issue back then (see #1852 and #1656), it's introducing it back.

EPuncker:
> This PR is a regression; it had been reported as an issue back then (see #1852 and #1656), it's introducing it back.

yes in 2016 it was possible to add yourself on the vip list, now it isn't

PR #2616, opened 2019-07-01
Fix to #2615. Slot type of weapons and getting defaults values in weapons by .lua

### LINKED ISSUE #2615: Errors in weapons by .lua, when equipping "two-handed" and the duration of weapons.
305/5000
Well, first I want to say that it was tested directly on the repository folder after cloning.
I have not changed anything at all on source.

There is a bug in the "two-handed" function and the "duration"
The "two-handed" returns this error when trying to equip (both hands are free):

There is a bug in the "two-handed" function and the "duration"
The "two-handed" returns this error when trying to equip (both hands are free):

For the duration, I put it in three examples.

Next:
This is the first, it seems to be normal setting the value at "150" equivalent to 2.5 minutes (everything ok as it is in print)

Already in this one does not beat:

This is totally without logic:

And taking the opportunity to thank the great content, simplified a lot.
Congratulations

souzajunior:
- Getting default values in get values.
- Fixed slot position in weapon:slotType
- Fixed duration of a weapon (uint32_t)

Fix #2615

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -15662,14 +15662,9 @@ int LuaScriptInterface::luaWeaponVocation(lua_State* L)
 		weapon->addVocWeaponMap(getString(L, 2));
 		weapon->setWieldInfo(WIELDINFO_VOCREQ);
 		std::string tmp;
-		bool showInDescription = false;
-		bool lastVoc = false;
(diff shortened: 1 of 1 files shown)


Znote:
@EvilHero90 Please review.

PR #2620, opened 2019-07-02
Revscriptsys: Weapon absorb percent all and by an element

souzajunior:
**Function**
weapon:absorbPercentAll(value)
weapon:absorbPercent(type, value)

**Example**
weapon:absorbPercentAll(15)
weapon:absorbPercent("fire", 10)

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -725,6 +725,16 @@ std::string getCombatName(CombatType_t combatType)
 	return "unknown";
 }
 
+CombatType_t getCombatTypeByName(std::string name)
+{
+	for (auto it = combatTypeNames.begin(); it != combatTypeNames.end(); ++it) {
+		if (it->second == name) {
+			return it->first;
+		}
+	}
+	return COMBAT_NONE;
+}
+
 Ammo_t getAmmoType(const std::string& strValue)
 {
 	auto ammoType = ammoTypeNames.find(strValue);
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -725,7 +725,7 @@ std::string getCombatName(CombatType_t combatType)
 	return "unknown";
 }
 
-CombatType_t getCombatTypeByName(std::string name)
+CombatType_t getCombatTypeByName(const std::string& name)
 {
 	for (auto it = combatTypeNames.begin(); it != combatTypeNames.end(); ++it) {
 		if (it->second == name) {
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -727,12 +727,16 @@ std::string getCombatName(CombatType_t combatType)
 
 CombatType_t getCombatTypeByName(const std::string& name)
 {
-	for (auto it = combatTypeNames.begin(); it != combatTypeNames.end(); ++it) {
-		if (it->second == name) {
-			return it->first;
-		}
-	}
-	return COMBAT_NONE;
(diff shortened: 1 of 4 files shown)


eduardodantas95:
Tested. 
All ok

src/tools.cpp
```diff
@@ -725,6 +725,16 @@ std::string getCombatName(CombatType_t combatType)
 	return "unknown";
 }
 
+CombatType_t getCombatTypeByName(std::string name)
```
djarek:
```suggestion
CombatType_t getCombatTypeByName(const std::string& name)
```

src/tools.cpp
```diff
@@ -725,6 +725,16 @@ std::string getCombatName(CombatType_t combatType)
...
+{
+	for (auto it = combatTypeNames.begin(); it != combatTypeNames.end(); ++it) {
```
djarek: Use `find_if`, makes it clearer you just want to find an element in the map:
```
auto it = std::find_if(combatTypeNames.begin(), combatTypeNames.end(),
  [&](std::pair<std::string, CombatType_t> const& pair)
  {
    return pair.second == name;
  });

if (it == combatTypeNames.end())
  return COMBAT_NONE;
else
  return it->first;
```

Alternatively, there should be a reverse-lookup map.

souzajunior: Great modification, ty. Just a correction:
`[&](std::pair<std::string, CombatType_t> const& pair)` 
modify to:
`[&](std::pair<CombatType_t, std::string> const& pair)`

src/tools.cpp
```diff
@@ -725,6 +725,20 @@ std::string getCombatName(CombatType_t combatType)
 	return "unknown";
 }
 
+CombatType_t getCombatTypeByName(const std::string& name)
+{
+	auto it = std::find_if(combatTypeNames.begin(), combatTypeNames.end(),
+		[&](std::pair<CombatType_t, std::string> const& pair)
+	{
```
djarek: One minor thing: please check if this is consistent with lambda formatting elsewhere. IIRC braces in lambda expressions should be attached to the first line of the expression.

souzajunior: It's working fine, I didn't see any problem with this modification. I update with your review, compiled and yeah, all right.

EvilHero90:
This entire thing was actually planned to be in Items/ItemType once it gets a complete workover with item virtualization and such, so maybe instead of having it strictly for weapons only you could adapt it to ItemType for now, so we don't have it double once ItemType gets a re work and is already included.

src/tools.cpp
djarek: I was referring to putting `{` in the previous line, because that is, I think how it's formatted elsewhere in the project.

souzajunior:
> This entire thing was actually planned to be in Items/ItemType once it gets a complete workover with item virtualization and such, so maybe instead of having it strictly for weapons only you could adapt it to ItemType for now, so we don't have it double once ItemType gets a re work and is already included.

Oh, okay, no problem. My friend told me it was not implemented, so I thought it was better to do a PR here first. Anyway, thank you for reply.

src/tools.cpp
souzajunior: Oooh, sorry, haha. I guess so, I'm still adapting to the project, Ty.

EvilHero90:
You don't need to close it, i just meant if you moved the code to itemtype instead it would perfectly fit for later on, so we don't have to do it twice :D

PR #2626, opened 2019-07-13
fix crash #2625

### LINKED ISSUE #2625: Server crash

1. Make a script with monsterType:getLoot() and execute an event of onDeath or prepareDeath (killing a monster, for example)

Script in talkactions: (example to reproduce this bug)
Screenshot_71

Calling the script ingame:
`!monster Demon`

**If I start my server and type !monster Demon, this works perfect, but when I kill a monster, bang, server down.
Same thing if I start the server, kill a monster and type !monster Hydra (the monster's name doesn't matter), server down again.**

### Expected behaviour
The execution of the script using mType:getLoot(). 

### Actual behaviour
Server crash (just close the console, without error message)

maattch:
fix #2625

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -12769,7 +12769,7 @@ int LuaScriptInterface::luaMonsterTypeGetLoot(lua_State* L)
 		return 1;
 	}
 
-	static const std::function<void(const std::vector<LootBlock>&)> parseLoot = [&](const std::vector<LootBlock>& lootList) {
+	static const std::function<void(const std::vector<LootBlock>&)> parseLoot = [L](const std::vector<LootBlock>& lootList) {
 		lua_createtable(L, lootList.size(), 0);
 
 		int index = 0;


trunksagain:
Hi, maattch! Doesn't worked for me, :/.

maattch:
> 
> 
> Hi, maattch! Doesn't worked for me, :/.

thats weird, i was having the same issue and this worked for me

src/luascript.cpp
```diff
@@ -12769,7 +12769,7 @@ int LuaScriptInterface::luaMonsterTypeGetLoot(lua_State* L)
 		return 1;
 	}
 
-	static const std::function<void(const std::vector<LootBlock>&)> parseLoot = [&](const std::vector<LootBlock>& lootList) {
+	static const std::function<void(const std::vector<LootBlock>&)> parseLoot = [L](const std::vector<LootBlock>& lootList) {
```
djarek: This is not a proper fix. Please move the code in this lambda into a separate function - `void pushLoot(lua_State* L, const std::vector<LootBlock>& loot)`

PR #2637, opened 2019-08-12
Fix getLootRandom, fix use of randomseed

### LINKED ISSUE #2632: Monster:onDropLoot / getLootRandom() not working correctly
- [✔] This is a bug in the software that resides in this repository, and not a
- [✔ ] This issue is reproducible without changes to the code in this repository

To make the following steps easy, I suggest changing groups.xml to allow GOD characters to kill monsters and get loot.
1. Pick one monster of your choosing and kill a lot of them.
-- For my tests I chose Hero and Dragon Lord and killed 250
2. Observe that the monster dropped no rare items, until finally one monster dropped every item in its loot list, including all of its rares.
-- I tested at both the default 2x loot rate and at 10x

### Expected behaviour
Loot distribution of rare items should be spread according to its chance in the its xml

### Actual behaviour
Rare items only seem to drop all at the same time.
Seems to only affect loot items with a chance < 700 (approx.)

### Environment
Replicated on both Ubuntu 16.04 & Windows 10

thatmichaelguy:
resolves #2632 
Solution provided by iryont

- Removed randomseed from ``startup.lua``
- Removed randomseed from ``getLootRandom`` function of ``global.lua``
- Added randomseed to line 1 of ``global.lua``

### DIFF
--- a/data/global.lua
+++ b/data/global.lua
@@ -1,3 +1,4 @@
+math.randomseed(os.time())
 dofile('data/lib/lib.lua')
 
 STORAGEVALUE_PROMOTION = 30018
@@ -34,7 +35,6 @@ function getFormattedWorldTime()
 end
 
 function getLootRandom()
-	math.randomseed(os.mtime())
 	return math.random(0, MAX_LOOTCHANCE) / configManager.getNumber(configKeys.RATE_LOOT)
 end
 
--- a/data/globalevents/scripts/startup.lua
+++ b/data/globalevents/scripts/startup.lua
@@ -1,5 +1,4 @@
 function onStartup()
-	math.randomseed(os.mtime())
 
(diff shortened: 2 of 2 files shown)


ranisalt:
Hey, have a look at my latest comment in parent issue. I argue that you should keep it in `startup.lua` instead.

thatmichaelguy:
I will change this PR to keep it in ``startup.lua`` if you wish. Afaik TFS only operates in one lua state, but is this likely to change in the near future?
If it is likely to change, then it should probably be in ``global.lua`` now to ensure it is not forgotten about. Not really a big deal though - I don't have a strong opinion on this at all, so if anybody does I'll update to satisfy.

EvilHero90:
I'm the same opinion as ranisalt, it should be in startup.lua, get it changed and it'll be merged.

Issue #2477, opened 2018-07-06
Slightly off spell formulae


Wirox:
Instead of
```lua
local min = (level / 5) + (...)
local max = (level / 5) + (...)
```
spell formulae should look like
```lua
local min = math.floor(level / 5) + (...)
local max = math.floor(level / 5) + (...)
```

DSpeichert:
Please provide source for the proper formula (reference).

Wirox:
https://tibia.wikia.com/wiki/Formulae#Spell.2FRune_Damage.2FHealing
It can also be confirmed very easily by checking min-max values of some spell on different levels.

raymondtfr:
@Wirox is right, although pointing to tibia.wikia as a reliable source regarding some formulas is not good. You can check tibia-stats for references.

e.g. If you've been checking lets say your "healing" (e.g. 49-99 Hp) for a long time on a specific character level e.g. "95" you'll notice that you'll only get to heal exactly (50-100 Hp) once you're character level is 100.

Wirox:
Tibia Wikia indeed isn't a reliable source of formulae – I just couldn't think of any better reference at that moment, and my own gathered data is too disordered to publish it anywhere.

DSpeichert:
In that case we will welcome a PR.

Wirox:
I'm sorry, but what exactly do you mean by 'PR'?

raymondtfr:
I'll be doing this soon. Even though it is very hard to guess if there's any rounding type being added to the second part of the formula - let alone confirming if they are 100% accurate - I'll be adding ``floor`` or ``round`` rounding methods to it.

I also intend to *maybe* adding a new formula callback which only passes ``Player`` userdata and make a new metatable function for it to use with all current spells formulas with uniformity in mind.

PR #2638, opened 2019-08-13
Player get formula values metatable function

raymondtfr:
I created a new Player metatable function which generates formula values for all current spells in our data pack.

I had to add a a new function, ``Combat.aggressive`` (some aggressive spells do negative damage while some that are not do positive damage, which will work for our current spell pack) as a way to keep ``Player.getFormulaValue`` simpler.

I [plan](https://github.com/otland/forgottenserver/pull/2584#issuecomment-498692989) to apply this to all spells once I update them to use the new _revscriptsystem_ first.

This also addresses #2477 .

~I realized I forgot to consider formulas which only use constant values not based off of variables such as level, skills, etc. I'll be updating this once I think of a way to keep it simpler still.~ (for these spells I'll just use ``setFormula`` directly in the Lua files.)


EvilHero90:
nicely done, I'll apply the last changes for spells which you suggested and the bugfix for cooldown, so you can start with it already.

DSpeichert:
Is this still WIP?

PR #2639, opened 2019-08-13
optimizations / bugfix on Spell

EvilHero90:
- cooldown and groupCooldown are now correctly working
- vocation supports several vocation strings now
ex: :vocation("sorcerer;true", "master sorcerer") will show "sorcerer" in description


EvilHero90:
forgot to tag #2638 as this is ground work for it.

Znote:
I can cast the spell I just made, even though I'm not a Knight:
image

PR #2647, opened 2019-08-26
Optimized Clean with optional Protection Zone Clean

EvilHero90:
Refactored the clean function, instead of iterating over max dimensions (ex: rl map 32k x 32k) we rather save the position of an item which is dropped onto the ground in a `std::set`.
The code also checks once an item gets removed from the ground and erases that position out of the `std::set` if there aren't any cleanable items on that position anymore.

This still needs a lot of testing, so far it has been a lot more performant comparing to the current clean system, I'm however not sure how big the impact might be with more players online at once.


src/game.h
```diff
@@ -540,6 +560,8 @@ class Game
 
 		std::map<uint32_t, BedItem*> bedSleepersMap;
 
+		std::map<Tile*, uint8_t> cleanTiles;
```
djarek: Looks like the value is unused, so this should be `std::set`

EvilHero90: was not aware of `std::set` thanks, will change it.

EvilHero90:
I have been testing a bit with a rl map (32k x 32k)
This has been the result of the current /clean function
cleannotoptimized
This has been the result after refactoring the code
cleanoptimized
The performence boost increases with the size of the map itself.
If the refactored code reaches around 50% of the max tiles existing of the map then it starts getting slower, that's why I said the bigger the map the better it performs, as you'll actually not be able to reach 230k tiles on a rl map (as an example) from clean to clean.

ranisalt:
Nice, but `cleanTiles` is VERY misleading, it is the opposite of what the actual content means. Try `tilesToClean`.

src/tile.cpp
```diff
@@ -464,6 +470,26 @@ void Tile::onRemoveTileItem(const SpectatorHashSet& spectators, const std::vecto
 	for (Creature* spectator : spectators) {
 		spectator->onRemoveTileItem(this, cylinderMapPos, iType, item);
 	}
+
+	if (!hasFlag(TILESTATE_PROTECTIONZONE)) {
+		auto it = getItemList();
+		if (it->size() == 0) { // it->empty() does not work for some strange reason here
```
ranisalt: Hey this has to be debugged. Deep inside stdlib `empty()` is written as `return size() == 0` so they must be the same.

src/tile.cpp
```diff
@@ -464,6 +470,26 @@ void Tile::onRemoveTileItem(const SpectatorHashSet& spectators, const std::vecto
...
+		bool ret = false;
+		for (auto toCheck : *it) {
+			if (toCheck->isCleanable()) {
+				ret = true;
+				break;
+			}
+		}
+
+		if (!ret) {
+			g_game.removeCleanTile(this);
```
ranisalt: Instead of doing this, you can short circuit inside `for` directly:

```cpp
for (auto toCheck : *it) {
	if (toCheck->isCleanable()) {
		g_game.removeCleanTile(this);
		break;
	}
}
```

src/tile.cpp
```diff
@@ -398,6 +398,12 @@ void Tile::onAddTileItem(Item* item)
 	for (Creature* spectator : spectators) {
 		spectator->onAddTileItem(this, cylinderMapPos);
 	}
+
+	if (!hasFlag(TILESTATE_PROTECTIONZONE)) {
+		if (item->isCleanable()) {
```
ranisalt: Merge these ifs, no need to have separate lines

src/map.cpp
```diff
@@ -943,68 +943,36 @@ void QTreeLeafNode::removeCreature(Creature* c)
...
 	}
 
+	size_t count = toRemove.size();
+	g_game.clearCleanTiles();
+
 	std::cout << "> CLEAN: Removed " << count << " item" << (count != 1 ? "s" : "")
-	          << " from " << tiles << " tile" << (tiles != 1 ? "s" : "") << " in "
-	          << (OTSYS_TIME() - start) / (1000.) << " seconds." << std::endl;
+		<< " from " << tiles << " tile" << (tiles != 1 ? "s" : "") << " in "
+		<< (OTSYS_TIME() - start) / (1000.) << " seconds." << std::endl;
```
ranisalt: I think your IDE has changed this lines and you didn't notice

src/map.cpp
```diff
@@ -943,68 +943,36 @@ void QTreeLeafNode::removeCreature(Creature* c)
...
-				}
-			}
-		} else {
-			for (auto childNode : node->child) {
-				if (childNode) {
-					nodes.push_back(childNode);
-				}
+	for (auto tileList : g_game.getCleanTiles()) {
+		++tiles;
+		for (Item* item : *tileList->getItemList()) {
```
ranisalt: Why not `auto` here too?

gesior:
All names MUST be changed:
`getCleanTiles` -> `getTilesToClean`
`addCleanTile` -> `addTileToClean`
`removeCleanTile` -> `removeTileToClean`
`clearCleanTiles` -> `clearTiles`
`cleanTiles` -> `tilesToClean`

EvilHero90:
Alright, if you want I'll give this a final touch so we can get this through, if you want we could also enable this as a side option (not removing the original code) but enable it with a config.lua option? @DSpeichert

src/map.cpp
```diff
@@ -943,68 +943,36 @@ void QTreeLeafNode::removeCreature(Creature* c)
...
+	for (auto item : toRemove) {
+		g_game.internalRemoveItem(item, -1);
+	}
 
 	if (g_game.getGameState() == GAME_STATE_MAINTAIN) {
 		g_game.setGameState(GAME_STATE_NORMAL);
 	}
 
+	size_t count = toRemove.size();
+	g_game.clearCleanTiles();
```
ranisalt, line 972: This really should happen before the game state is changed to normal.

DSpeichert:
@EvilHero90 This doesn't have an alternative. If this works well, it can be the only option.

Lyuzera:
@DSpeichert i use it on my server since november and no problems at the moment. My clean is every 1 hour.

src/map.cpp
EvilHero90: this is actually 2 tabs

src/tile.cpp
EvilHero90: this does logic wise not work, only if all items in the stack of the tile return false, then it should remove the tile, else it should still clean it.

src/tile.cpp
EvilHero90: "Can't select the function, due to non necessary rights" not sure why this is throwing as we use the same method for Creature* vectors aswell and it works there.

src/map.cpp
EvilHero90: moved it up, thanks

src/tile.cpp
EvilHero90: done that

infister:
To be honest this was already present in 0.4 TFS, worked very well.

src/tile.cpp
djarek: You need to forward the function you want to use here: https://github.com/otland/forgottenserver/blob/master/src/tile.h#L81 because we're using private inheritence

infister:
One thing I would like to recommend it to store positions instead of Tiles, because if, by any chance, tile get nulled, clean will crash the server.

EvilHero90:
> 
> 
> One thing I would like to recommend is to store positions instead of Tiles, because if, by any chance, tile get nulled, clean will crash the server.

This should actually not happen, because if the tile gets erased it should normaly call ```onRemoveTileItem``` which will then remove the ```*tile``` out of the unorderd_map, but for safety I'll make a tile check to verify if the tile even exists.

Lyuzera:
@EvilHero90 I think if cleanProtectionZones is true, it will clean items from houses that players moved recently. So you need to check the tile in the methods `Tile::onAddTileItem` and `Tile::onRemoveTileItem`.
```c
// if it's not a house tile..
if (!dynamic_cast<HouseTile*>(this) {

}
```

Lyuzera:
@Znote nice, there's really no need to check ```Tile::onRemoveTileItem```.

PR #2649, opened 2019-08-27
[revscriptsys] call clear() on action/movement vectors

EvilHero90:
- this will clear the temp vectors used to store the id/aid/uid/pos for the actions/movements after calling the register() method.


Znote:
Compiled and run fine, didnt find a proper way to test for memory leak tho (even before or after this). 
However this supposedly fixes it, so I will merge it. Server compiles, scripts run just fine. Changes seem pretty harmless.

PR #2653, opened 2019-09-01
Moved spells to revscriptsys

raymondtfr:
Moved spells to the new scripting system.
Removed custom spells folder
Removed custom spell "Blank Rune".
Fixed support runes file name (Convince Creature to Convince Creature Rune and Chameleon to Chameleon Rune)
Updated arithmetic constants and use suffix ``* 1000`` to make time constants more readable.
Fixed some requirements of player spells Light Stone Shower Rune and Lightest Missile Rune.

This was done manually so I really hope I haven't messed anything up. Since I did this manually I had the chance to go through all spell files again and try and find some errors, etc, which are described above. I decided to remove custom spells because they were not used anyway.


soul4soul:
Good work. I feel custom spell should be kept even if the only custom spell is "create blank rune".

Znote:
Does this remove the ability to add spells the old way? (data/spells)?

data/scripts/lib/create_functions.lua
```diff
@@ -1,2 +1,4 @@
 createFunctions(MonsterType) -- creates get/set functions for MonsterType
 createFunctions(Spell) -- creates get/set functions for Spell
+
+dofile('data/scripts/spells/lib/spells.lua')
```
EvilHero90: this doesn't need to be here, just move the file `spells.lua` into `scripts/lib` it will be automaticly load before anything else.

EvilHero90: You forgot to remove this `dofile` part in your latest commit

raymondtfr:
> 
> 
> Does this remove the ability to add spells the old way? (data/spells)?

No, it doesn't. But you won't be able to use codes from the library there, since it was moved to another folder.

Znote:
I imagine when we tag the next release, which will be the first official release that supports Revscriptsys, people won't be very familiar with it yet. Because of this I propose waiting until we release the next version (Estimated launch now in October) before we merge this in. 

But that we merge this (and your other revscriptify PR's) pretty much asap after the release and reviews are done. 

We plan to tag this release 1.4, making even numbers "official" release, and odd numbers (1.3, 1.5) development versions. Perhaps we can merge these PR's into 1.5, as long as backwards compatibility still works.

mattyx14:
> I imagine when we tag the next release, which will be the first official release that supports Revscriptsys, people won't be very familiar with it yet. Because of this I propose waiting until we release the next version (Estimated launch now in October) before we merge this in.
> 
> But that we merge this (and your other revscriptify PR's) pretty much asap after the release and reviews are done.
> 
> We plan to tag this release 1.4, making even numbers "official" release, and odd numbers (1.3, 1.5) development versions. Perhaps we can merge these PR's into 1.5, as long as backwards compatibility still works.

wait other actions are moved to revscripts

rottyrub:
this feels like a step backwards.
Revscripts are a good addition to TFS, but should it really completely remove XML? Having all spells registered in a single file makes it much more organized. Do you wanna locate a spell by word for example, then you go into the spells.xml file and find it. With only revscripts we need to search for the word in the entire spells folder. And making bulk changes like let's say adding a new vocation and registering existing spells for it, now you would have to go into every spell file and make the changes there instead.

ranisalt:
> then all the existing AACS/tools that parses all our XML files would also have to be adjusted with this new system

Regarding this, I am strongly against AAC's source of truth being something other than the server running itself. They should use the status protocol, which itself should have a path returning available spells. See #2010

Leo32onGIT:
> 
> 
> > then all the existing AACS/tools that parses all our XML files would also have to be adjusted with this new system
> 
> Regarding this, I personally am strongly against AAC's source of truth being something other than the server running itself. They should use the status protocol, which itself should have a path returning available spells. See #2010

Merging this, along-side other revscript systems in 1.4 seems to be the best answer here.

DSpeichert:
Closing this because at this point there is no benefit to forcefully converting XML systems into revscripts and @rottyrub makes great points.

We can have new spells via Revscripts in the datapack as examples but no need to remove existing ones if the XML way of doing spells still works.

PR #2655, opened 2019-09-05
Apply schema.sql migrations

Znote:
This applies schema migrations to schema.sql: 
21.lua #2403 (utf8 was missing from one table)
22.lua #2525 (skullticks time overflow)
23.lua #2532 (look direction)

Also created migration 24 which optimizes migration 23. #2532
Look direction should not be stored as an integer (4 bytes), tinyint (1 byte) is sufficient.

And bumps the db_version.

### DIFF
--- a/data/migrations/24.lua
+++ b/data/migrations/24.lua
@@ -1,3 +1,5 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 25 (Optimize player direction 4 -> 1 byte)")
+	db.query("ALTER TABLE `players` MODIFY COLUMN `direction` tinyint(1) unsigned NOT NULL DEFAULT 2")
+	return true
 end
--- a/schema.sql
+++ b/schema.sql
@@ -28,6 +28,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `looklegs` int(11) NOT NULL DEFAULT '0',
   `looktype` int(11) NOT NULL DEFAULT '136',
   `lookaddons` int(11) NOT NULL DEFAULT '0',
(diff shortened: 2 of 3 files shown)


Znote:
Also created migration 24 which optimizes migration 23. #2532 
Look direction should not be stored as an integer (4 bytes), tinyint (1 byte) is sufficient.

data/migrations/24.lua
```diff
@@ -1,3 +1,5 @@
 function onUpdateDatabase()
-	return false
-end
+	print("> Updating database to version 25 (Optimize player direction 4 -> 1 byte)")
```
DSpeichert: It says version 25 here but this is migration 24.

schema.sql
```diff
@@ -340,6 +339,8 @@ CREATE TABLE IF NOT EXISTS `towns` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
+INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '24'), ('motd_hash', ''), ('motd_num', '0'), ('players_record', '0');
```
DSpeichert: This says version 24.

data/migrations/24.lua
Znote: This is just the convention style @marksamman did. I just follow that. Check literally every other migration, they follow the same style. The schema is tagged as version 24, version 25 just returns false until someone builds that migration.

schema.sql
Znote: Yep, because this is migration 24.

Issue #1150, opened 2015-01-17
Packet Erro and Lag (Game not Responding)


walissonsantos:
http://oi62.tinypic.com/2rqk6wx.jpg

walissonsantos:
maxPacketsPerSecond = 50

walissonsantos:
Sorry, the bug was caused by excessive items in the mail box ... if you can limit it would be great. The player was with some 70 thousand label in the mail box when the player logava the entire server was with lag.

walissonsantos:
If anyone has interest in resolving this for me I can reward you ... Those interested leave a message.

TheSumm:
There's Bountysource for that.

walissonsantos:
Summ, please contact me on skype: waliss.on@hotmail.com

Znote:
Interesting scenario, aside from the easy solution which is to add a limit to reasonable numbers, perhaps there also are other optimizations we could to related to item storage?

From a technical perspective: 
When you open the mail inbox, do the server load all the items, or just the visible slot ids? Could excessive records in the db be the cause along with mass loading? (Even from cache, loads of unnecessary data would cause a bottleneck)? 

From select query standpoint, perhaps limit 100 (or whatever max visible slots are) could help. 
From table structure, perhaps adding an item_container table could help filter out relevant(visible) items. As a way to load items progressively, instead of loading a huge chunk of data in the start, when your only suppose to be able to view max 100 of them.

marksamman:
The items are loaded to memory when a player logs in, and saved to the database when a player logs out, there's no performance issue with using the inbox while logged in. With a lot of entries in the database, this results in huge INSERT queries. There are many ways to approach this problem, e.g. disabling fsync, max limit for items in inbox, only apply diffs on save, live queries. We will probably end up going with a max limit like there is on the depot chest.

I have barely looked into this issue so I'm not even sure if fsync is the bottleneck here, and even if it is, there are also other ways to work around that than disabling fsync. We could use async queries to insert inbox items and prevent the player from logging in again until all queries have returned success.

jo3bingham:
@marksamman If you add a max limit to the inbox how will you handle house items that get sent to the player's inbox when they sell it that happen to be over the max limit?

marksamman:
[...]

PR #2657, opened 2019-09-08
Issue 1150 fix

gesior:
3 changes to reduce problems described in #1150 
1. Cache player items in RAM. Allow LUA script to load all players items into RAM at server start. Save items in asynchronous thread with own database connection. Reduce relog time 30-50 times.
2. Do not save player items when he logouts. Wait 5 seconds to make sure, he does not login again. If he does fast relogs all time, items will be saved to database when he stops.
3. Save items in binary format. Reduce save time 10-15 times. Added database migration for it.

Added new options to config.lua.dist

    playerItemsCache = true
    playerItemsCacheSaveThreshold = 5000
    binaryPlayerItems = true

Added new LUA functions:

    PlayerCacheManager.clearCache()
    PlayerCacheManager.clearPlayerCache(guid)
    PlayerCacheManager.loadPlayerCache(guid)
    PlayerCacheManager.savePlayerCache(guid)


Znote:
Migration conflict fixed, but there is a compile error as well:
```bash
[ 43%] Building CXX object CMakeFiles/tfs.dir/src/iomapserialize.cpp.o
/home/travis/build/otland/forgottenserver/src/iologindata.cpp: In function ‘IOLoginData::loadItems(std::map<unsigned int, std::pair<Item*, unsigned int>, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, std::pair<Item*, unsigned int> > > >&, std::shared_ptr<DBResult>)’:
/home/travis/build/otland/forgottenserver/src/iologindata.cpp:1135:45: error: ‘type’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
    Item* item = Item::CreateItem(type, count);
                                             ^
/home/travis/build/otland/forgottenserver/src/iologindata.cpp:1135:45: error: ‘count’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
cc1plus: all warnings being treated as errors
make[2]: *** [CMakeFiles/tfs.dir/src/iologindata.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/tfs.dir/all] Error 2
make: *** [all] Error 2
```

gesior:
I cannot reproduce. Isn't it gcc 4.8.4 bug?

Tried it on gcc 6.3.0 (Debian 9 default) - compiled without errors.
Tried it on VS 2019 - compiled without errors.
I even found some 3 years old dedic with Debian 6, but version there was 4.9.2 and compiled without errors.

Isn't 'trusty' Travis release an Ubuntu 14.04 LTS?
`Ubuntu 14.04 LTS reaches end of life on April 30, 2019`

https://docs.travis-ci.com/user/reference/trusty/
`Trusty is EOL by Canonical, try updating to a newer image and see our Trusty to Xenial Migration Guide.`

Znote:
Right now our source code complies with `gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4`, and if you can make your pull request compatible with it, then you should do that. 

It might be a good idea to also create an issue question/suggestion to bump our travis build utility. I personally agree with you that we should elevate the minimum distribution requirements based on LTS and EOL cycles.

SaiyansKing:
@gesior
Znote results are actually to be expected because reading from "PropStream" can failed and leave you with uninitialized variable so it'll be whatever you have in your memory on the stack.
Checking if reading failed and don't do anything with the variable should be the "writing real C++ code" that you said.

P.S.
Why you didn't write the binary saving the way houses does it? You can speed up saving by several miliseconds and loading by up to 4-6x and it'll progress even more the more item you have because your code use std::map(O(log n)) which isn't needed when loading items, a little example relogging with 400k items:
[https://i.ibb.co/r4n77Py/untitled.png](url)

gesior:
@Znote 
Updated code to make it 100% compatible with 'trusty gcc'.
### DO NOT MERGE THIS PULL REQUEST
I will prepare new with cleaner commits (made from master) and muuch updated code.
I want to make code work like in houses after I read @SaiyansKing comment.

> Znote results are actually to be expected because reading from "PropStream" can failed and leave you with uninitialized variable so it'll be whatever you have in your memory on the stack.
> Checking if reading failed and don't do anything with the variable should be the "writing real C++ code" that you said.
> 
> P.S.
> Why you didn't write the binary saving the way houses does it? You can speed up saving by several miliseconds and loading by up to 4-6x and it'll progress even more the more item you have because your code use std::map(O(log n)) which isn't needed when loading items, a little example relogging with 400k items:
> [https://i.ibb.co/r4n77Py/untitled.png](url)

About first thing:
`read` returns false only in case when there is no more data. To block this issue I've added:
`while (itemsStream.size() >= 14) {`
but adding big 'if' with reading all parameters would be clearer. I will update this.

About second thing:
I did not read house saving code :(
I copied binary save code from Kasteria engine. Without generating big map and then linking parent-child pairs it should be much faster.


Is there any limit of 'bp in bp' in TFS? Without it, player could crash server by.. stackoverflow :smile:

SaiyansKing:
Yeah I know that there is this check in while loop but what I wanted to say is that compilers might think it can end up with uninitialized variable because they aren't that smart.

About limit of 'bp in bp' there aren't any and never was while there are some other functions that can end up in stackoverflow like "Container::updateItemWeight", if you want I rewrited the house-saving like process to a non-recursive way:
[http://oneclickpaste.com/106159/](http://oneclickpaste.com/106159/)

gesior:
@SaiyansKing 
From PHP doc:
https://www.php.net/goto
image
:smile:

PR #2661, opened 2019-09-22
resolve rsa trim footer, correct dockerfile init and configlua dis to…

### LINKED ISSUE #2660: RSA key.pem file footer blank

### Expected behaviour

docker needs be able to run with a docker run command.
RSA key have a whitespace in footer and it broken when server shows up.
gesior needs a key to work in configdist

### Actual behaviour

server does not go up when started by docker command

### Environment

Dockerfile and docker-compose

Riicksouzaa:
… work with gesior fix #2660


Znote:
Why are you changing the default values in config.lua?

Riicksouzaa:
I am developing a docker-compose integration.
for the website and server to be integrated.

Riicksouzaa:
But. Realy. That names i've included até not equals to this project.

I'm gona make another commit with correct values.

Srry.

Riicksouzaa:
@Znote changes done.

Znote:
Havent used docker before, do you need ip="server" hostname?
I guess passwordType is a Gesior thing? (We are hardcoded to sha1 and its misleading to give the user the assumption that he can change this?)

This may break other deployments that rely on consistent default values in config.lua.dist

I think you can run the sed command in dockerfile to manipulate the config.lua.dist file while copying it over to config.lua

replace:
```
COPY ./config.lua.dist ./config.lua
```

with:
```
RUN sed 's/ip = "127.0.0.1"/ip = "server"\npasswordType = "sha1"/' config.lua > config.lua.dist
```

That way, no changes at all has to be done in config.lua.dist?

Riicksouzaa:
Yep, i'm new into docker.

> I guess passwordType is a Gesior thing? (We are hardcoded to sha1 and its misleading to give the user the assumption that he can change this?)
This is news to me, I didn't know it was hardcoded.

In my project the user would be able to run tfs, maridd, ngnx, php to have a complete site / server environment going up in minutes.


See initial prototype. [docker-compose-gesior+tfs](https://github.com/Riicksouzaa/FerobraTfsDocker)

DSpeichert:
I don't know what you're trying to do here.

a) Every text file should by convention end with a newline character, I see no problem with TFS running and loading the key file we have now. And the trimming?
b) The docker image builds fine without the extra package you added.

Riicksouzaa:
> I don't know what you're trying to do here.
> 
> a) Every text file should by convention end with a newline character, I see no problem with TFS running and loading the key file we have now. And the trimming?
> b) The docker image builds fine without the extra package you added.

In my ambient when i run the docker command the server doesn't go up, the error is related to the footer.

```cpp
if (key.substr(key.size() - footer.size(), footer.size()) != footer) {
    throw std::runtime_error("Missing RSA private key footer.");
}
```
This part of code reads the size of file and the size of footer, when the pem file have a newline this creates a mismatch and returns this error, that's why i implemented the trimimg, i don't know that's right but i tried, and now work's fine on my ambient.

And the 'boost-dev' on Dockerfile fixes the error i get, `libboost_filesystem.so.1.71.0` when i try to run my tfs image.

DSpeichert:
I don't know what environment you're using but I can't reproduce either of these by building and running TFS in docker on an Arch Linux host right now.

DSpeichert:
I'm closing this because there is no need to make those changes, at least not in the default Dockerfile - as I've tested today. We can reopen this if you show how/when such changes are actually necessary.

PR #2663, opened 2019-09-28
Partially moved actions to revscriptsys

raymondtfr:


DSpeichert:
Currently, we are not interested in moving existing actions to revscripts.

PR #2668, opened 2019-10-05
Removing old functions

luanluciano93:


data/npc/lib/npc.lua
```diff
@@ -97,27 +98,29 @@ function doPlayerTakeItem(cid, itemid, count)
 end
 
 function doPlayerSellItem(cid, itemid, count, cost)
+	local player = Player(cid)
 	if doPlayerTakeItem(cid, itemid, count) == true then
```
EPuncker: player:removeItem here?

data/npc/lib/npcsystem/modules.lua
```diff
@@ -954,20 +955,24 @@ if Modules == nil then
 			local msg = self.npcHandler:getMessage(msgId)
 			parseInfo[TAG_ITEMCOUNT] = a
 			msg = self.npcHandler:parseMessage(msg, parseInfo)
-			doPlayerSendCancel(cid, msg)
+			player:sendCancelMessage(msg)
 			self.npcHandler.talkStart[cid] = os.time()
 
 			if a > 0 then
-				doPlayerRemoveMoney(cid, ((a * shopItem.buy) + (b * 20)))
+				if not player:removeMoney(((a * shopItem.buy) + (b * 20))) then
```
EPuncker: there is a ( ) that is not needed

data/npc/lib/npc.lua
```diff
@@ -56,68 +56,44 @@ function doNpcSellItem(cid, itemid, amount, subType, ignoreCap, inBackpacks, bac
...
-		else
-			return false
-		end
-	end
-
-	if count ~= 0 then
-		return false
-	end
-	return true
-end
```
DSpeichert: Do we need that in compat.lua now? Could it be used outside this file?

luanluciano93: Doesn't function player:removeItem() the same as this?

jo3bingham:
Please use more descriptive commit messages. You have 5 commits that share the same two, vague messages. The commit message should prepare the reviewer for what to expect. For example, ce8825f could have been something as simple as "Remove unnecessary parentheses". This tells the reviewer exactly what to expect in this commit. "Update modules.lua" could be anything, and then the reviewer is left to figure it out. Granted, your commits are simple, but it's still useful to get in the habit of leaving better commit messages.

data/npc/lib/npc.lua
DSpeichert: That's exactly my point, if they are the same, are you providing a compatibility alias/shim wrapper for scripts that exist somewhere and may be missing this function?

EPuncker: yeah @luanluciano93 add doPlayerTakeItem to compat.lua file returning player:removeItem()

data/npc/lib/npc.lua
```diff
@@ -56,68 +56,44 @@ function doNpcSellItem(cid, itemid, amount, subType, ignoreCap, inBackpacks, bac
...
 function doCreatureSayWithDelay(cid, text, type, delay, e, pcid)
-	if isPlayer(pcid) then
-		e.done = FALSE
+	if Player(pcid):isPlayer() then
+		e.done = false
 		e.event = addEvent(func, delay < 1 and 1000 or delay, cid, text, type, e, pcid)
 	end
 end
 
-function doPlayerTakeItem(cid, itemid, count)
```
DSpeichert: Add this to compat, doPlayerTakeItem is commonly used in NPC scripts across datapacks.

data/lib/compat/compat.lua
```diff
@@ -1297,3 +1297,10 @@ function createFunctions(class)
 		rawset(class, func[3], func[4])
 	end
 end
+
+function doPlayerTakeItem(cid, itemid, count)
+	if not Player(cid):removeItem(itemid, count) then
+		return false
+	end
+	return true
```
DSpeichert: Uh?
Why not just `return Player(cid):removeItem(itemid, count)`

data/lib/compat/compat.lua
```diff
@@ -1297,3 +1297,7 @@ function createFunctions(class)
 		rawset(class, func[3], func[4])
 	end
 end
+
+function doPlayerTakeItem(cid, itemid, count)
+	return Player(cid):removeItem(itemid, count) then
```
DSpeichert: Do you even test this code?

Znote:
```suggestion
	return Player(cid):removeItem(itemid, count)
```

PR #2673, opened 2019-10-08
Fix stacking creatures in a tile

slavidodo:
TFS limits creatures to 10 in a tile, so the creatures wouldn't appear in battle-list. While this wouldn't affect you at all, it can affect worlds where expert pvp is introduced or even in depots.

### DIFF
--- a/src/player.h
+++ b/src/player.h
@@ -718,6 +718,11 @@ class Player final : public Creature, public Cylinder
 				client->sendRemoveTileThing(pos, stackpos);
 			}
 		}
+		void sendRemoveTileCreature(const Creature* creature, const Position& pos, int32_t stackpos) {
+			if (client) {
+				client->sendRemoveTileCreature(creature, pos, stackpos);
+			}
+		}
 		void sendUpdateTile(const Tile* tile, const Position& pos) {
 			if (client) {
 				client->sendUpdateTile(tile, pos);
@@ -736,7 +741,7 @@ class Player final : public Creature, public Cylinder
 		}
 		void sendCreatureAppear(const Creature* creature, const Position& pos, bool isLogin) {
 			if (client) {
-				client->sendAddCreature(creature, pos, creature->getTile()->getStackposOfCreature(this, creature), isLogin);
+				client->sendAddCreature(creature, pos, creature->getTile()->getClientIndexOfCreature(this, creature), isLogin);
 			}
 		}
 		void sendCreatureMove(const Creature* creature, const Position& newPos, int32_t newStackPos, const Position& oldPos, int32_t oldStackPos, bool teleport) {
@@ -746,7 +751,7 @@ class Player final : public Creature, public Cylinder
 		}
 		void sendCreatureTurn(const Creature* creature) {
 			if (client && canSeeCreature(creature)) {
-				int32_t stackpos = creature->getTile()->getStackposOfCreature(this, creature);
+				int32_t stackpos = creature->getTile()->getClientIndexOfCreature(this, creature);
 				if (stackpos != -1) {
 					client->sendCreatureTurn(creature, stackpos);
 				}
@@ -787,7 +792,7 @@ class Player final : public Creature, public Cylinder
 			} else if (canSeeInvisibility()) {
 				client->sendCreatureOutfit(creature, creature->getCurrentOutfit());
 			} else {
-				int32_t stackpos = creature->getTile()->getStackposOfCreature(this, creature);
+				int32_t stackpos = creature->getTile()->getClientIndexOfCreature(this, creature);
 				if (stackpos == -1) {
 					return;
 				}
--- a/src/player.h
+++ b/src/player.h
@@ -800,7 +800,7 @@ class Player final : public Creature, public Cylinder
 				if (visible) {
 					client->sendAddCreature(creature, creature->getPosition(), stackpos, false);
 				} else {
-					client->sendRemoveTileThing(creature->getPosition(), stackpos);
+					client->sendRemoveTileCreature(creature, creature->getPosition(), stackpos);
 				}
 			}
 		}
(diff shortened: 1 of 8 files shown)


slavidodo:
It allows sending more than 10 creatures in a tile.

Eternal-Scripts:
To avoid debugs in normal client, increase the NETWORKMESSAGE_MAXSIZE to 65535
Btw, nice fix, works fine here.

src/protocolgame.cpp
```diff
@@ -2976,6 +2975,18 @@ void ProtocolGame::RemoveTileThing(NetworkMessage& msg, const Position& pos, uin
...
+void ProtocolGame::RemoveTileCreature(NetworkMessage& msg, const Creature* creature, const Position& pos, uint32_t stackpos)
+{
+	if (stackpos < 10) {
+		RemoveTileThing(msg, pos, stackpos);
+		return;
+	}
+
+	msg.addByte(0x6C);
+	msg.add<uint16_t>(0xFFFF);
+	msg.addByte(creature->getID());
```
Eternal-Scripts: Fix: msg.addU32(creature->getID());

Taken from https://github.com/edubart/otclient/blob/master/src/client/protocolgameparse.cpp#L2194

DSpeichert:
> To avoid debugs in normal client, increase the NETWORKMESSAGE_MAXSIZE to 65535

Why this number? Is this related to the change here?

Eternal-Scripts:
@DSpeichert Because this change allow to send more informations through the socket. (more than 10 creatures). Yes.
Otherwise, check how the tfs connection works and you will understand.

slavidodo:
There are rare cases where the network message size limit could be exceeded, indeed this PR could cause, though I don't think this PR should not change it!

@Eternal-Scripts feel free to open another PR!

ranisalt:
Why bother having a max size that large? Let's just use a vector with a decent preallocated size, measured by seeing what the most common message size is.

DSpeichert:
The limit of 10 was introduced for a reason, I'm curious what changed that "now" the client can handle more or somehow it won't crash the client.

slavidodo:
The code is ancient, this limit was removed back at 9.00 I think~
This was tested and works fine! Also it's introduced on several profit servers!

DSpeichert:
> The code is ancient, this limit was removed back at 9.00 I think~

There was a problem in 10.98 that was fixed in #2415 when this piece of code was last edited and I don't recall the client handling more characters back then well, unless there was an unrelated bug. 
I'm not saying your change is wrong but I want to know what changed that it works now. It looks like just reverting the limit.

slavidodo:
Was there any sources to prove that the client handles only 10 players instead of 7 back then?
This has been there since 9.00 as I suggested, so technically that means back in 2018 we could've supported more than 10 characters too!

Maybe also this was implemented when the client supported moving into a character on the same tile on non-pvp zones. So logically, it makes sense that the server sends all characters so that the client can list them without actually drawing them!

slavidodo:
So your'e saying battle list is not entirely updated?

Znote:
Yes, as shown in my 1st gif, I added 12 monsters, only 9 where visible in my battle list. (Before I relogged)

And as shown in my 2nd gif, 12 monsters was removed, but 2 of them was still in my battle list (Even though they don't exist anymore).

slavidodo:
OTClient test confirmed, works fine!

src/player.h
```diff
@@ -718,6 +718,11 @@ class Player final : public Creature, public Cylinder
 				client->sendRemoveTileThing(pos, stackpos);
 			}
 		}
+		void sendRemoveTileCreature(const Creature* creature, const Position& pos, int32_t stackpos) {
+			if (client) {
```
lgrossi, line 722: Can you please check if stackpos is -1 and not call client for those cases? I don't think the clients are designed to handle this -1.

src/protocolgame.cpp
```diff
@@ -2397,7 +2395,19 @@ void ProtocolGame::sendAddCreature(const Creature* creature, const Position& pos
...
+		// 2. since no stackpos will be send to the client about that creature, removing
+		//    it must be done with its id if its stackpos remains >= 10. this is done to
+		//    add creatures to battle list instead of rendering on screen
+		if (stackpos >= 10) {
+			// @todo: should we avoid this check?
+			if (const Tile* tile = creature->getTile()) {
+				sendUpdateTile(tile, pos);
+			}
+		} else {
+			// if stackpos is -1, the client will automatically detect it
```
lgrossi, line 2410: That's not true for otc.

Znote:
> Otc will throw errors for stackpos -1, imho the server should handle that. Also, as already said before, the network limit may lead to unexpected behaviors. Maybe a way of paginating it, sending the creatures in small bunch of packages or splitting the map description in floor description would make it more consistent without needing to increase the network limit to 65k.

If the original clients handle stackpos -1 as (monster/object exist, but don't render it on screen), then otclient should be patched. As this is a protocol compatibility issue with otc, not the server.

Zbizu:
I have an idea.

You can use the code from /ghost to hide the creature from the client if it's on a stacked tile somewhere below top x creatures and code unghost it in a new position if it leaves the stack.

Zbizu:
https://github.com/edubart/otclient/blob/c6d0fc0f71a3986ceb61dbff971a8ac63c8e1623/src/client/tile.cpp#L187

This is probably where it's handled in otclient, though I have no idea how to fix it

nekiro:
This is a good change, had the chance to test it recently. Otc side you simply delete these two lines 
https://github.com/edubart/otclient/blob/c6d0fc0f71a3986ceb61dbff971a8ac63c8e1623/src/client/tile.cpp#L212-L213
and it works just fine.

nekiro (approved):

DSpeichert (approved):

Codinablack:
For the record. This commit completely broke compatibility between 1.4x and OTC. And removing those two lines nekiro mentioned, which now look like this in mehah's

```cpp
if (size > g_gameConfig.getTileMaxThings())
    removeThing(m_things[g_gameConfig.getTileMaxThings()]);
    ```
    
    Just causes a crash, and doesn't even fix anything at all.
    
    Default client still works fine, but yeah, OTC is completely broken on compatibility with this commit, which might not concern you guys (as its not broken on default client), but its actually quite a massive issue for everyone in the community who wants to use 1.4x

PR #2674, opened 2019-10-10
Player::canSeeCreature check creature is null

Eternal-Scripts:
- Fix a random crash when the creature is null (0x0)

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -678,6 +678,10 @@ bool Player::canSee(const Position& pos) const
 
 bool Player::canSeeCreature(const Creature* creature) const
 {
+	if (!creature) {
+		return false;
+	}
+	
 	if (creature == this) {
 		return true;
 	}


DSpeichert:
You're probably just hiding a symptom of a larger problem - why is this function called on null in the first place?

Eternal-Scripts:
@DSpeichert i don't remember now why, but this check is important btw.
The alternative is track every place who use this function and put a check too, but in big systems this is recommended.

Lyuzera:
In many places where this method is called, it is checked first if the creature exists, but I think the verification inside the method is valid, because we don't know if it's checking everywhere before. It is an extra security.

Znote:
Is it possible to throw out a warning, possibly a staktrace before returning false if creature does not exist? 

That way we solve this problem, as well as enlighten the users of which process is causing this error. 
Is it a performance thingy that stops us from verifying function params? We just expect a function to be used correctly?

DSpeichert:
Let's find a proper fix instead.

PR #2683, opened 2019-10-20
FIX string.split doesn't works correctly with space

Lyuzera:
If you use spaces after the first separator, you may return a bad result.

this command for example..
```Lua
	local split = param:split(',')
	if split[1] == 'check' then
		local player = Player(split[2])
		if player then
			print('exists')
		else
			print('doesn\'t exists')
		end
	end
	return false
end
```
`/command check, Playername`
```Lua
-- doesn't exists
```
` /command check,Playername`
```Lua
-- exit: exists
```

So now we can use with spaces normally.

### DIFF
--- a/data/global.lua
+++ b/data/global.lua
@@ -50,6 +50,14 @@ end
 string.split = function(str, sep)
 	local res = {}
 	for v in str:gmatch("([^" .. sep .. "]+)") do
+		res[#res + 1] = v
+	end
+	return res
+end
+
+string.splitTrimmed = function(str, sep)
+	local res = {}
+	for v in str:gmatch("([^" .. sep .. "]+)") do
 		res[#res + 1] = v:trim()
 	end
 	return res
--- a/data/global.lua
+++ b/data/global.lua
@@ -50,7 +50,7 @@ end
 string.split = function(str, sep)
 	local res = {}
 	for v in str:gmatch("([^" .. sep .. "]+)") do
-		res[#res + 1] = v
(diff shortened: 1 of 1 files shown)


Znote:
Technically, split does exacly what is expected of it to do. And with this PR, it trims as well, which may be considered unintentional additional behavior. 

I would propose leaving split as it is, and instead create a new function `string:splitparam()` which contains this additional feature to the split function.

Lyuzera:
@Znote @DSpeichert I made changes, but..
see this part of add_skill.lua
```Lua
	-- Trim left
	split[2] = split[2]:gsub("^%s*(.-)$", "%1")
```
They had to adapt to work normally, this would not be necessary with string.split + trim()

DSpeichert:
So as an example of using this function, why don't you modify `add_skill.lua` to use it?

Lyuzera:
just remove
```Lua
	-- Trim left
	split[2] = split[2]:gsub("^%s*(.-)$", "%1")
```
replace it
```Lua
```
for it
```Lua
local split = param:splitTrimmed(",")
```
I realized that this could be improved when I added war system and tried using this command
`/war invite, guild`
The command stated that the guild did not exist because the following was happening:
```Lua
local guild = Guild( guildname) -- not trimmed
```
Now imagine creating a new command and having to use the default add_skill.lua method for many arguments that need to be trim. Very tiring, right?

DSpeichert:
I'm not asking how to do it, I'm asking you to modify your PR to do it.

PR #2685, opened 2019-10-21
otserv.cpp: Tidy and comment otserv

foxrother:
Code tidying and extensive commenting.

### DIFF
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -18,13 +18,9 @@
  */
 
 #include "otpch.h"
-
 #include "server.h"
-
 #include "game.h"
-
 #include "iomarket.h"
-
 #include "configmanager.h"
 #include "scriptmanager.h"
 #include "rsa.h"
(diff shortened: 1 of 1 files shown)


src/otserv.cpp
```diff
@@ -141,105 +108,111 @@ void mainLoader(int, char*[], ServiceManager* services)
...
-	//load vocations
+	// Load vocations from '/XML/vocations.xml'
 	std::cout << ">> Loading vocations" << std::endl;
 	if (!g_vocations.loadFromXml()) {
 		startupErrorMessage("Unable to load vocations!");
 		return;
 	}
 
-	// load item data
+	// TODO DOC
```
Znote: Stripping comment in favor of a placeholder comment?
Shouldn't the todo list be resolved before you commit a PR? Especially if each PR represent a file/section already.

foxrother: Hey! Thanks for giving your feedback.

The stripped comment states the obvious and does not provide any additional insight.

I marked those as "TODO" because I wanted contributors (including myself) to know what are the parts that can receive more attention (including my own).
Also, would you leave a comment on something that you don't fully understand? I'd rather mark it.

I mentioned in the forums that I'm focusing on the more fundamental parts of the code base right now.
I intend update those comments as I read their respective files, I simply can't do it in a heartbeat...

foxrother:
I hope I haven't startled anyone. 😛
I'm not trying to resist your feedback or anything, by the way...
I'm really just attempting to share some insights and experiment a bit with a healthy dose of reasonable arguments...
There's no problem if you guys don't agree with me after that's been done, I'd just like to ask you to be more explicit on what changes you'd like to be made. 😀

I'd also like to say that I'm not "moving lines around" for stats' sake.
I genuinely think that this kind of housekeeping is necessary.

PR #2686, opened 2019-10-21
tasks.h, tasks.cpp: Tidy and comment tasks

foxrother:
Tidy code and extensive commenting.

### DIFF
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -18,13 +18,9 @@
  */
 
 #include "otpch.h"
-
 #include "server.h"
-
 #include "game.h"
-
 #include "iomarket.h"
-
 #include "configmanager.h"
 #include "scriptmanager.h"
 #include "rsa.h"
@@ -37,20 +33,31 @@
 #include "script.h"
 #include <fstream>
 
-DatabaseTasks g_databaseTasks;
-Dispatcher g_dispatcher;
-Scheduler g_scheduler;
+extern Scripts* g_scripts; // Holds '*.lua' scripts from '/data'. Extern means it's defined elsewhere.
 
-Game g_game;
-ConfigManager g_config;
-Monsters g_monsters;
-Vocations g_vocations;
-extern Scripts* g_scripts;
-RSA g_RSA;
+DatabaseTasks g_databaseTasks;	// Runs on separate thread, handles database queries
+Scheduler g_scheduler;			// Runs on separate thread, handles planned executions
+Dispatcher g_dispatcher;		// Runs on separate thread, handles immediate executions
 
+Game g_game;			// Holds and mutates game state (map, players, monsters, NPCs, etc)
+ConfigManager g_config; // Holds configuration from '/config.lua'
+Monsters g_monsters;	// Holds monster definitions from /data/monster (XMLs)
+Vocations g_vocations;	// Holds vocations definitions from '/data/XML/vocations.xml'
+RSA g_RSA;				// Holds private key from '/key.pem' and decrypts data
+
+// Prevents 'main()' from continuing execution without 'mainLoader()'
 std::mutex g_loaderLock;
 std::condition_variable g_loaderSignal;
 std::unique_lock<std::mutex> g_loaderUniqueLock(g_loaderLock);
+
+/* Entry point 'main()' at the end of the file. */
+
+[[noreturn]] void badAllocationHandler()
+{
+	puts("Allocation failed, server out of memory.\nDecrease the size of your map or compile in 64 bits mode.\n");
+	getchar();
+	exit(-1);
+}
 
 void startupErrorMessage(const std::string& errorStr)
 {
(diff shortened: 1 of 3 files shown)


foxrother:
Sorry about the multiple edits. It's bery late and I'm tired. 😛

src/otserv.cpp
```diff
@@ -37,96 +33,67 @@
...
+#	endif
 	std::cout << STATUS_SERVER_NAME << " - Version " << STATUS_SERVER_VERSION << std::endl;
 	std::cout << "Compiled with " << BOOST_COMPILER << std::endl;
 	std::cout << "Compiled on " << __DATE__ << ' ' << __TIME__ << " for platform ";
 
-#if defined(__amd64__) || defined(_M_X64)
+#	if defined(__amd64__) || defined(_M_X64)
 	std::cout << "x64" << std::endl;
-#elif defined(__i386__) || defined(_M_IX86) || defined(_X86_)
+#	elif defined(__i386__) || defined(_M_IX86) || defined(_X86_)
```
Znote: Whats the deal with a random tab between the # and conditional declaration?

src/otserv.cpp
```diff
@@ -37,96 +33,67 @@
...
 #include <fstream>
 
-DatabaseTasks g_databaseTasks;
-Dispatcher g_dispatcher;
-Scheduler g_scheduler;
+extern Scripts* g_scripts; // Holds '*.lua' scripts from '/data'. Extern means it's defined elsewhere.
+
+DatabaseTasks g_databaseTasks;	// Runs on separate thread, handles database queries
+Scheduler g_scheduler;			// Runs on separate thread, handles planned executions
+Dispatcher g_dispatcher;		// Runs on separate thread, handles immediate executions
```
DSpeichert: The formatting here is suboptimal

src/otserv.cpp
foxrother: It's not random. The tab indents the directives into the same block as the inserted code would be.
We could also just indent the whole line, but then we wouldn't easily see where there's preprocessing.

src/otserv.cpp
foxrother: Hey! Thanks.

Care to explain how?
I think we should always see externs before everything else.
These are just declarations, also, so it doesn't really affect performance.

Also, I just didn't want to stick yet another line into the source for these short comments.

src/otserv.cpp
```diff
@@ -37,96 +33,67 @@
...
-ConfigManager g_config;
-Monsters g_monsters;
-Vocations g_vocations;
-extern Scripts* g_scripts;
-RSA g_RSA;
+Game g_game;			// Holds and mutates game state (map, players, monsters, NPCs, etc)
+ConfigManager g_config; // Holds configuration from '/config.lua'
+Monsters g_monsters;	// Holds monster definitions from /data/monster (XMLs)
+Vocations g_vocations;	// Holds vocations definitions from '/data/XML/vocations.xml'
+RSA g_RSA;				// Holds private key from '/key.pem' and decrypts data
```
foxrother: Can anyone help me figure out what's going on with indentation?
Git is messing them up somehow... 
image

foxrother: Maybe it's related to .editorconfig?
Apparently it's configured to use spaces, which can cause this sort of behavior.

src/otserv.cpp
foxrother: GitHub uses tab-size 8 instead of 4, that's the root of the problem.
https://github.com/otland/forgottenserver/blob/96153c4565d0ecf6f9537eb7cda5fa886f26e94f/src/otserv.cpp?ts=4
This is how the code _**actually**_ looks like.

src/otserv.cpp
foxrother: GitHub uses tabsize 8 instead of 4.
This is how the code actually looks like?
https://github.com/otland/forgottenserver/blob/96153c4565d0ecf6f9537eb7cda5fa886f26e94f/src/otserv.cpp?ts=4

foxrother: @Znote that might be causing indentation to look "random"

src/otserv.cpp
Znote: The # symbol is not indented into the correct scope level, which breaks the scope collapse functionality in text editors such as Sublime Text.
image

foxrother:
@Znote Okay, I'll put comments in separate lines. 😄 
I'd just like to point out that there are _very_ few places where they were inlined. At the start of otserv.cpp, for instance, they give a pretty straightforward overview of the architecture without too much scrolling and digging around.

Most of these "deletions" and "additions" happened due to the "tidying" part. Git can't track whether a whole function or class declaration/definition was moved.
One of the weird things in otserv.cpp was mainLoader's declaration on top of the entry-point and then its definition below the entry-point. Why not just declare and define mainLoader() _before_ main()? It doesn't affect the behavior of the program, it's easier do understand that mainLoader is used by main() and it's easier to find main at the end...
I am very sorry that you're unable to easily review these changes with straight-up blaming... What I can do is to chop this them into (even) smaller commits, if that's of any help.

On the TODO tags, I agree that, ideally, they shouldn't appear on a "stable" branch. But I do believe that they can be beneficial. They are yet another tool to aid change documentation (take a look at [this bot](https://github.com/apps/todo)).
If the master branch stands for "stable branch", maybe we could consider creating a separate branch for development/documentation.

Yes, I agree we do need proper documentation guidelines. Maybe what we don't agree is on what the goal is.
To me, the goal is to make it easier to understand the architecture, for _new contributors_ and maintainers alike.
'Properly documenting functions' will not benefit _both_ because due to its rigidity, it can't really provide enough detail, especially since we have so many global variables, numerous dependencies, etc.
So until there's a very thoughtful, concerted analysis on the whole architecture, I don't think it is the means to this particular goal... Now, I'm really interested on how you think the 'proper function documentation' could help in this regard. If you have ideas, please share them! 😅

Also, I think we should be having more conversations on how to attract more contributors and drive the quality of contributions up. This is an amazing project and it is, to me, very saddening to see it unknown/uninteresting to the wider game-making community.

Maybe what we consider priorities really differ too much, maybe I am projecting too much onto this project (no pun intended). But I feel like I should share the rationale behind my "efforts"...

PR #2687, opened 2019-10-24
add docker-compose.yml to quick start

danielflira:
- Fix boost-filesystem dependecy in original Dockerfile
- Add entrypoint.sh with automatic configuration of database
- Add docker-compose.yml for quick start of a new server

If you hado docker in swarm mode just run:

```bash
$ docker stack up -c docker-compose.yaml otserver
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,30 @@
+version: "3.6"
+
+services:
+  mysql:
+    image: mysql
+    environment:
+      MYSQL_ROOT_PASSWORD: ${TFS_MYSQL_ROOT_PASSWORD:-r00tp@ssword}
+      MYSQL_DATABASE: ${TFS_MYSQL_DATABASE:-forgottenserver}
+      MYSQL_USER: ${TFS_MYSQL_USER:-forgottenserver}
+      MYSQL_PASSWORD: ${TFS_MYSQL_PASS:-forgottenserver}
+    volumes:
+      - forgottenserver_db:/var/lib/mysql
+
+  forgottenserver:
+    image: otland/forgottenserver
+    environment:
+      TFS_MYSQL_HOST: ${TFS_MYSQL_HOST:-mysql}
+      TFS_MYSQL_DATABASE: ${TFS_MYSQL_DATABASE:-forgottenserver}
+      TFS_MYSQL_USER: ${TFS_MYSQL_USER:-forgottenserver}
+      TFS_MYSQL_PASS: ${TFS_MYSQL_PASS:-forgottenserver}
+    volumes:
+      - forgottenserver:/srv
+    ports:
+      - 7171:7171
+      - 7172:7172
+
+volumes:
+  forgottenserver:
+  forgottenserver_db:
+  
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+set -e 
+set -x
+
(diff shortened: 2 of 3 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,34 @@
...
+    mv config.lua.dist config.lua
+fi
+
+if [ "${@}" != "" ]; then
+    ${@}
+    exit $?
+fi
+
+if [ -f schema.sql ]; then
+    ln -s  /usr /usr/usr
```
DSpeichert: why?

entrypoint.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [ -f "config.lua.dist" ]; then
+    sed -i "s/mysqlHost =.*/mysqlHost = \"${TFS_MYSQL_HOST}\"/" config.lua.dist
+    sed -i "s/mysqlUser =.*/mysqlUser = \"${TFS_MYSQL_USER}\"/" config.lua.dist
+    sed -i "s/mysqlPass =.*/mysqlPass = \"${TFS_MYSQL_PASS}\"/" config.lua.dist
+    sed -i "s/mysqlDatabase =.*/mysqlDatabase = \"${TFS_MYSQL_DATABASE}\"/" config.lua.dist
+    sed -i "s/mysqlPort =.*/mysqlPort = ${TFS_MYSQL_PORT}/" config.lua.dist
+    sed -i "s/mysqlSock =.*/mysqlSock = \"${TFS_MYSQL_Sock}\"/" config.lua.dist
+
+    mv config.lua.dist config.lua
```
DSpeichert: Why move that file instead of copying?

entrypoint.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [ "${@}" != "" ]; then
+    ${@}
+    exit $?
+fi
+
+if [ -f schema.sql ]; then
+    ln -s  /usr /usr/usr
+    mysql -h "${TFS_MYSQL_HOST}" -u "${TFS_MYSQL_USER}" "${TFS_MYSQL_DATABASE}" "-p${TFS_MYSQL_PASS}" < schema.sql || true
+    mv schema.sql schema.sql.bak
```
DSpeichert: why?

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: "3.6"
+
+services:
+  mysql:
+    image: mysql
+    environment:
+      MYSQL_ROOT_PASSWORD: ${TFS_MYSQL_ROOT_PASSWORD:-r00tp@ssword}
```
DSpeichert: not the best default password

entrypoint.sh
danielflira: when I call mysql to automatically create database schema the program fails with some broken linked library path looking for it on /usr/usr/lib... so I linked usr inside usr to workaround the problem.

entrypoint.sh
danielflira: There is no special reason, I just made that way

entrypoint.sh
danielflira: schema.sql is on mounted volume, so first time when you run stack it create database schema automatically, after that don't do nothing, because file don't exist.

docker-compose.yml
danielflira: Could you suggest what is the best default password?

DSpeichert: https://mariadb.com/kb/en/authentication-plugin-unix-socket/

Xaekai:
User has since dropped their fork, signalling this PR is abandoned.

PR #2691, opened 2019-10-27
Clean up LuaScriptInterface::luaSpellVocation

marksamman:
The code was quite confusing and from a quick glance there seemed to be a potential out of bounds access there:
```
if (vocList.size() > 0) {	
	if (vocList[1] == "true") {
```
It turns out that the first if-statement was just unnecessary. I've cleaned up the code and also replaced the `== "true"` check with booleanString().

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -14365,6 +14365,7 @@ int LuaScriptInterface::luaSpellVocation(lua_State* L)
 
 	if (lua_gettop(L) == 1) {
 		lua_createtable(L, 0, 0);
+		int i = 0;
 		for (auto& voc : spell->getVocMap()) {
 			std::string name = g_vocations.getVocation(voc.first)->getVocName();
 			setField(L, std::to_string(++i).c_str(), name);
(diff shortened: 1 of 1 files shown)


marksamman:
I noticed that the previous implementation would set `false` if there wasn't a semi-colon in the string, is that intentional? I feel like defaulting to true would make more sense, but I don't like the design of handling arguments in this way at all.

raymondtfr:
> 
> 
> I noticed that the previous implementation would set `false` if there wasn't a semi-colon in the string, is that intentional? I feel like defaulting to true would make more sense, but I don't like the design of handling arguments in this way at all.

In recent tibia versions all runes now show vocation descriptions by default, so I'd say remove them altogether if we plan to update to newest protocols.

marksamman:
I see, so that bool is for showing vocation in description. It's not obvious to anyone using this API.

raymondtfr:
Yes, and it is only used for runes but that part of the code would run for any kind of spell if I'm not mistaken, even though "instant" does not make use of it.

PR #2692, opened 2019-10-27
Optimize getSpectators

marksamman:
This is usually at least twice as fast, but depending on amount of spectators it can be orders of magnitude faster.

Note that I have not tested the changes and it would be great if someone can test before this is merged, especially the parts that need to filter out duplicates.

### DIFF
--- a/src/house.h
+++ b/src/house.h
@@ -22,6 +22,7 @@
 
 #include <regex>
 #include <set>
+#include <unordered_set>
 
 #include "container.h"
 #include "housetile.h"
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -552,7 +552,7 @@ void Combat::CombatNullFunc(Creature* caster, Creature* target, const CombatPara
 	CombatDispelFunc(caster, target, params, nullptr);
 }
 
-void Combat::combatTileEffects(const SpectatorHashSet& spectators, Creature* caster, Tile* tile, const CombatParams& params)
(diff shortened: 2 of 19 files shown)


src/house.h
```diff
@@ -22,6 +22,7 @@
 
 #include <regex>
 #include <set>
+#include <unordered_set>
```
slavidodo: what's the reason behind this?

marksamman: unordered_set was previously implicitly imported through tile.h, since it is no longer imported in tile.h, it has to be imported in house.h because house.h uses unordered_sets.

CristoferMartins:
It good to be merged. The only suggestion i got is to inherit from std::vector or create a class to wrap it and then add the functions to remove duplicated and erase a certain creature there to avoid duplicated code.

marksamman:
I agree, but not sure what a good value for vector.reserve would be. If there are getSpectator calls with empty results, you wouldn't want to allocate additional memory. I suspect that it contains at least 1 player/creature most of the time, and maybe reserving something between 1 and 8 slots could be good, but I would rather have real data on most common spectator size values if anyone could gather that.

CristoferMartins:
As far i understand if you allocate 8 slots or 32 slots it does not matter since its a single call to allocation and the size in bytes of the allocation is not great(8x8 bytes or 32x8 bytes as pointer size in 64 bits system is 8 bytes). " I would rather have real data on most common spectator size values if anyone could gather that." this is the best approach, in the case we don't have this data i would pick 32 slots.

PR #2693, opened 2019-10-27
Increase Dispatcher throughput by not clearing spectator cache every task

marksamman:
This could also increase the spectator cache hit rate since it is no longer cleared as often, but still cleared when necessary.

Further improvements that could be made for a higher cache hit rate would be to only clear the cache in map nodes where a creature has moved (or been added/removed) rather than clearing the entire cache, but that requires an efficient way to determine affected cache entries.

### DIFF
--- a/src/tasks.cpp
+++ b/src/tasks.cpp
@@ -58,8 +58,6 @@ void Dispatcher::threadMain()
 				++dispatcherCycle;
 				// execute it
 				(*task)();
-
-				g_game.map.clearSpectatorCache();
 			}
 			delete task;
 		} else {


CristoferMartins:
"Further improvements that could be made for a higher cache hit rate would be to only clear the cache in map nodes where a creature has moved (or been added/removed) rather than clearing the entire cache, but that requires an efficient way to determine affected cache entries." The only thing i can think about is a structure which tells the area the cache was taken from and then when a creature moves check if creature is moving from/into the area. The real question is, an active server there is people hunting(and thus moving and using spells all the time) then the cache may be useless.

PR #2694, opened 2019-10-27
Improve players spectator cache hit rate

marksamman:
There is no need to clear the players spectator cache when a non-player Creature has been added/moved/removed. This comes at the cost of checking if the creature is a player in those functions though, but hopefully that is negligible.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -451,6 +451,10 @@ void Map::getSpectators(SpectatorHashSet& spectators, const Position& centerPos,
 void Map::clearSpectatorCache()
 {
 	spectatorCache.clear();
+}
+
+void Map::clearPlayersSpectatorCache()
+{
 	playersSpectatorCache.clear();
 }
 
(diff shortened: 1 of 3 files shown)


CristoferMartins:
Why you don't remove the call "g_game.map.clearSpectatorCache();" inside tasks.cpp? As far i understood only the spectatorCache will be cleared when a task is executed while playersSpectatorCache will be only cleared when a player move.

kygov:
> Why you don't remove the call "g_game.map.clearSpectatorCache();" inside tasks.cpp? As far i understood only the spectatorCache will be cleared when a task is executed while playersSpectatorCache will be only cleared when a player move. I think now both can rely on Tile functions to be cleared and the call inside tasks.cpp can be removed.

https://github.com/otland/forgottenserver/pull/2693

CristoferMartins:
@kygov thanks, i was kinda lost with two prs that could be one i guess.

PR #2697, opened 2019-10-27
Added shopping bag to modules

EPuncker:
- Added shopping bag to NPC "buy with backpacks" function as of Tibia 10.55 update
- Changed some commands to meta-table commands

### DIFF
--- a/data/npc/lib/npcsystem/modules.lua
+++ b/data/npc/lib/npcsystem/modules.lua
@@ -1089,7 +1089,7 @@ if Modules == nil then
 				return false
 			end
 
-			local a, b = doNpcSellItem(cid, shop_itemid[cid], shop_amount[cid], shop_subtype[cid], false, false, 23782)
+			local a, b = doNpcSellItem(cid, shop_itemid[cid], shop_amount[cid], shop_subtype[cid], false, false, backpackContainerId)
 			if a < shop_amount[cid] then
 				local msgId = MESSAGE_NEEDMORESPACE
 				if a == 0 then
--- a/data/npc/lib/npcsystem/modules.lua
+++ b/data/npc/lib/npcsystem/modules.lua
@@ -34,6 +34,8 @@ if Modules == nil then
 
 	StdModule = {}
 
+	backpackContainerId = 23782
+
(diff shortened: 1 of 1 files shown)


data/npc/lib/npcsystem/modules.lua
```diff
@@ -1116,7 +1116,7 @@ if Modules == nil then
 					return false
 				end
 				if shop_itemid[cid] == ITEM_PARCEL then
-					doNpcSellItem(cid, ITEM_LABEL, shop_amount[cid], shop_subtype[cid], true, false, 1988)
+					doNpcSellItem(cid, ITEM_LABEL, shop_amount[cid], shop_subtype[cid], true, false, 23782)
```
DSpeichert: Can you make it `local backpackID = 23782` here as well like the in the function above?
Or just make it a "global" backpackContainerId at the top of the file once for all functions to use.

EPuncker: like I did now?

DSpeichert: I think it looks good. Did you test that?

EPuncker: Just did, and not working, need to edit a few thins in npc.lua as well, I'll sort it out and send a fix

EPuncker:
> Should we add ITEM_SHOPPING_BAG be a const instead of a scripted declared itemid?
> https://github.com/otland/forgottenserver/blob/master/src/const.h#L463

I think so

data/npc/lib/npcsystem/modules.lua
EPuncker: for some odd reason the "buy with backpack" is adding an empty container (without my modified code) and I don't know how to fix it, and seems like nobody reported it until now:

wJJc6bA

if someone is able to confirm and fix it, I'll be grateful, you can even include the content from my PR and close it, I'm not able to fix it myself

EPuncker: pr #2699 fixes the issue, after it gets approved (I tested and its working), I'll make a new and clean PR with my changes, you can close this one now

PR #2701, opened 2019-10-28
Added shopping bag to modules

EPuncker:
- Added shopping bag to NPC "buy with backpacks" function as of Tibia 10.55 update
- Changed some commands to meta-table commands
- Added a few items into enums for future use


EPuncker:
and yeah I tested it :D

src/const.h
```diff
@@ -461,15 +461,35 @@ enum item_t : uint16_t {
 	ITEM_WILDGROWTH_SAFE = 11099,
 
 	ITEM_BAG = 1987,
+	ITEM_SHOPPING_BAG = 23782,
 
 	ITEM_GOLD_COIN = 2148,
 	ITEM_PLATINUM_COIN = 2152,
 	ITEM_CRYSTAL_COIN = 2160,
+	ITEM_FORGGOTEN_COIN = 24774, // in-game store currency
```
nekiro: I dont think thats something we gonna allow in main repository.

EPuncker: it is the official store currency, Tibia coins, but since we avoid using Tibia for names because of copyright issues, I named it Forgotten coin, it is already in items.xml as it

Kamenuvol: Forggoten is wrong

Sickness-chan: @EPuncker what he meant was that you had misspelled "forgotten", you wrote "forggoten".

raymondtfr: Just rename all instances to ``ITEM_STORE_COIN`` as it makes more sense. "Forgotten" is the emulators name.

data/npc/lib/npcsystem/modules.lua
```diff
@@ -945,7 +944,7 @@ if Modules == nil then
 		end
 
 		local subType = shopItem.subType or 1
-		local a, b = doNpcSellItem(cid, itemid, amount, subType, ignoreCap, inBackpacks, backpack)
+		local a, b = doNpcSellItem(cid, itemid, amount, subType, ignoreCap, inBackpacks, ITEM_SHOPPING_BAG)
```
jo3bingham: Is `ITEM_SHOPPING_BAG` declared somewhere in Lua where I can't find it? Or are you missing a `registerEnum(ITEM_SHOPPING_BAG)` call in `luascript.cpp`?

EPuncker: thanks

PR #2703, opened 2019-10-30
Use a single function to clear spectator cache [with a boolean]

slavidodo:
Just to keep the code clean and to avoid duplicates.


kornholi:
I personally don't think it's any cleaner or clearer...

slavidodo:
Both functions do the exact same thing except that one of them is clearing the cache of a sub-class of the other (player <-> creature).
What's ambiguous or unclean?

slavidodo:
I think you are right that it hides the functionality of cleaning players cache only!

PR #2705, opened 2019-10-31
Fix getMinMaxValues attackSkill with distance weapons

Oen44:
When player is using distance weapon that needs ammunition and have ammo item in ammunition slot, attackSkill value is 0, this is because ammunition item is checked first and these items don't have any skill types.


nekiro:
Shouldn't this always use tool as the one for attackSkill? Ammunition never gonna have any skill value

Oen44:
Make random spell that uses skill value, equip bow and arrow, add `print(skill)`, you will get 0. If you equip a spear, you get your Distance Fighting.
Since when using distance weapon with ammunition is not connected to Distance Fighting?

nekiro:
Ah I got it confused. Tool is the ammunition so it should always take item in consideration

nekiro (approved):
LGTM

EPuncker (approved):

PR #2710, opened 2019-11-07
Prey System

joseluis2g:
Prey System
- New const PREY_SLOTCOUNT
- 3 new return values related to Prey
- New enums called resource types and message dialog
- Edited function removeMoney to work with balance
- New value added to players table called "bonusrerollcount"
- Created new table called "player_preydata"
- Edited setStamina function to work with Prey
- Added 4 new functions to luascript
- Created prey cpp and h files
- New methods and parse functions added.
- New byte to sendBasicData related to Prey
- Two new functions to tools to select random values

You can also check t***a.fandom.com/wiki/Prey_System for more information.

All credits to @mkalo


EPuncker:
introduced in Tibia 11.02 🤔  I hope we can update client version soon

src/game.cpp
```diff
@@ -5446,6 +5469,56 @@ void Game::parsePlayerExtendedOpcode(uint32_t playerId, uint8_t opcode, const st
...
+		case RESOURCETYPE_BANK_GOLD: player->sendResourceData(RESOURCETYPE_BANK_GOLD, player->getBankBalance()); break;
+		case RESOURCETYPE_INVENTORY_GOLD: player->sendResourceData(RESOURCETYPE_INVENTORY_GOLD, player->getMoney()); break;
+		case RESOURCETYPE_PREY_BONUS_REROLLS: player->sendResourceData(RESOURCETYPE_PREY_BONUS_REROLLS, 0); break;
+		default: {
+			player->sendResourceData(RESOURCETYPE_BANK_GOLD, player->getBankBalance());
+			player->sendResourceData(RESOURCETYPE_INVENTORY_GOLD, player->getMoney());
+			player->sendResourceData(RESOURCETYPE_PREY_BONUS_REROLLS, player->getBonusRerollCount());
+			break;
+		}
+}
```
slavidodo: indentation

src/player.cpp
```diff
@@ -4529,3 +4547,304 @@ void Player::setGuild(Guild* guild)
...
+			if (currentPrey.timeLeft > 0) {
+				sendPreyTimeLeft(preySlotId);
+			} else {
+				changePreyDataState(preySlotId, STATE_SELECTION);
+			}
+		}
+	}
+}
+
+uint16_t Player::getPreyBonusLoot(MonsterType* mType)
```
kygov: It's unused.

src/const.h
```diff
@@ -589,4 +589,6 @@ static constexpr int32_t PSTRG_MOUNTS_CURRENTMOUNT = (PSTRG_MOUNTS_RANGE_START +
 
 #define IS_IN_KEYRANGE(key, range) (key >= PSTRG_##range##_START && ((key - PSTRG_##range##_START) <= PSTRG_##range##_SIZE))
 
+#define PREY_SLOTCOUNT 3
```
nekiro: Why?

src/luascript.cpp
```diff
@@ -8795,8 +8810,7 @@ int LuaScriptInterface::luaPlayerSetStamina(lua_State* L)
 	uint16_t stamina = getNumber<uint16_t>(L, 2);
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
-		player->staminaMinutes = std::min<uint16_t>(2520, stamina);
-		player->sendStats();
+		player->setStamina(stamina);
```
nekiro: what is this change? I dom't think it's related to prey system?

src/protocolgame.cpp
```diff
@@ -1334,13 +1365,111 @@ void ProtocolGame::sendBasicData()
 		msg.add<uint32_t>(0);
 	}
 	msg.addByte(player->getVocation()->getClientId());
+	msg.addByte(1);
```
nekiro: elaborate? I guess its the byte for enabling prey, I dont think prey should be always enabled.

EPuncker: 1 prey is enabled by default, 2 for premium and 3 purchasable

nekiro: no, its not, while you are no vocation, you arent able to use prey.

EPuncker: true, I forgot that

src/const.h
joseluis2g: its used a lot of times and I guess its better to have a control of the amount of prey slot players can use, some may change this to use more slots with OTC, Idk.

src/luascript.cpp
joseluis2g: Yes, it is, you can check setStamina function and there a few lines changed.

src/protocolgame.cpp
joseluis2g: Maybe this need a check to avoid player with no vocation using prey.

PR #2711, opened 2019-11-09
Challenge creature stops monsters from running away.

Znote:
When a creature, like Dragon or Troll gets low on health, (treshhold specified in runonhealth), they will try to run away from the target.

This PR stops that from happening while they are under the influence of challenge creature (exeta res - Elite Knight spell). 

I became aware of this issue here:
https://otland.net/threads/exeta-res.267163/

and @EPuncker linked to a merged PR that solves this issue in the OTservBR-Global repository. 
Originally committed by @LucasCPrazeres 
https://github.com/opentibiabr/OTServBR-Global/pull/276

The code looks clean so I figured I would try to apply it to TFS. 
Tested locally, it works. 

Verified from TibiaWiki that the spell is suppose to prevent creatures from running away at low health. https://tibia.fandom.com/wiki/Challenge

So figured why not create a PR to this repository as well.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -866,11 +866,11 @@ void Monster::onThinkTarget(uint32_t interval)
 		if (mType->info.changeTargetSpeed != 0) {
 			bool canChangeTarget = true;
 
-			if (targetExetaCooldown > 0) {
-				targetExetaCooldown -= interval;
+			if (challengeFocusDuration > 0) {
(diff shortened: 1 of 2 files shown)


Kamenuvol:
exeta in variable name looks bad.
try targetChallengeCooldown

Znote:
Hmmm, or perhaps `challengeFocusDuration` ?

PR #2712, opened 2019-11-09
Added remove total money function to NPCs

EPuncker:
As of 10.94 update, Premium players no longer need to bring any money to pay for services or wares, as the concerned NPCs now deducts any money missing from their inventory from their Bank Account. And after another update it was made available for free account as well. Code based on the one originally posted by @Mkalo on otland [here](https://otland.net/threads/npc-working-with-balance.245690/)

### DIFF
--- a/data/npc/lib/npcsystem/modules.lua
+++ b/data/npc/lib/npcsystem/modules.lua
@@ -80,7 +80,7 @@ if Modules == nil then
 				npcHandler:say("You are already promoted!", cid)
 			elseif player:getLevel() < parameters.level then
 				npcHandler:say("I am sorry, but I can only promote you once you have reached level " .. parameters.level .. ".", cid)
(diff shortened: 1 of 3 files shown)


EPuncker:
and yes I tested 
9Wo0BLn

data/npc/lib/npcsystem/modules.lua
```diff
@@ -968,10 +968,10 @@ if Modules == nil then
 		else
 			local msg = self.npcHandler:getMessage(MESSAGE_BOUGHT)
 			msg = self.npcHandler:parseMessage(msg, parseInfo)
-			if not player:removeMoney(totalCost) then
+			player:sendTextMessage(MESSAGE_INFO_DESCR, msg)
```
EPuncker: this was swapped because the MESSAGE_BOUGHT should be sent first

PR #2715, opened 2019-11-12
Update Server Save configuration to the config.lua file

marcomoa:
Co-Authored-By: Renato Foot Guimarães Costallat <costallat@hotmail.com>


DSpeichert:
What's the point of this?

marcomoa:
> What's the point of this?

Better organization and better reading.

delusionx:
If you're going to do something like this, you should probably also implement their function rather than just defining the config keys and leaving the **entire** implementation to the user.

PR #2717, opened 2019-11-13
Updated corpse size of a few monsters

EPuncker:
I guess cipsoft did it way back when they added creature products and we missed it, but since I'm free account that is the furthest I could go

### DIFF
--- a/data/items/items.xml
+++ b/data/items/items.xml
@@ -4801,7 +4801,7 @@
 	</item>
 	<item id="2806" article="a" name="dead troll">
 		<attribute key="weight" value="60000" />
-		<attribute key="containerSize" value="10" />
+		<attribute key="containerSize" value="24" />
 		<attribute key="decayTo" value="2810" />
 		<attribute key="duration" value="600" />
 		<attribute key="corpseType" value="blood" />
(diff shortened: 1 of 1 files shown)


marcomoa:
I think it's a way of wasting time analyzing this, I'd think of a faster way to do it.
For example: the body will have the number of slots according to the items that were dropped, if I drop 5 would have 5 slots, on an automatic basis.
I don't know how it works but I think it would save time.

Would that cause any problems?
Sorry for the bad english xD

Sickness-chan:
Where did you get the information from? I quickly checked tibia wiki for a few corpses and the information differs from both the old values, and your new values. Of course it's also possible that the wiki contains wrong information. But here are a few examples

https://tibia.fandom.com/wiki/Dead_Troll -> ID 2806 - Volume differs, and for the next decay stage (ID 2810), the volume is incorrect entirely, as it's not a container that can be opened.
https://tibia.fandom.com/wiki/Slain_Skeleton -> ID 2809 - Volume & weight differs, and for the next decay stage (ID 2922), the volume is missing entirely & the weight differs.

There's probably many more examples of this, are you be sure the information you provided is correct?

PR #2718, opened 2019-11-14
Fix monster callbacks, prevent revscript constructors in regular script interfaces

TheSumm:


TheSumm:
I have disabled the possibility to use revscript constructors in the old lua interfaces, since the distinction between the interfaces still exists there. It would be confusing to have a action script that registers another action/movement script inside of it.
Additionally, it messes with reloading scripts, as `/reload scripts` would unload such scripts but not reload them.
As far as I remember the constructors would not even work properly when used in regular script interfaces.

PR #2720, opened 2019-11-16
Set default channel type to avoid client crash

delusionx:
Should only need to use this on creature:say, type param is at the end of the positional param list, other methods which use type param have more positional params after type.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -7709,7 +7709,7 @@ int LuaScriptInterface::luaCreatureTeleportTo(lua_State* L)
 
 int LuaScriptInterface::luaCreatureSay(lua_State* L)
 {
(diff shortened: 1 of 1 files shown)


Znote:
Perhaps do a check to see if creature is player or npc, and do talktype_say instead of talktype_monster_say?

delusionx:
I feel like in the community there's more usage with monster_say rather than talktype_say, but it does make more sense to default to yellow text for creatures that have the option to talk in default chat. It's a matter of opinion though, I'd like more input from others so we can determine the best option for default talk type.

Xagul:
The default across all distros I have ever used in the past (including older TFS as well) has been the standard "say" (yellow text). For the sake of consistency I would say leave it that way... but... defaulting to monster say sure does sound like nice qol for my personal use case.

PR #2721, opened 2019-11-18
Support for tibia 11 - only via protocol game.

slavidodo:
tested through tibia 11.04 ~ 11.11
this is not supposed to be merged unless a decision is made about comparability with tibia 10 client.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -67,10 +67,10 @@ void Connection::close(bool force)
 	ConnectionManager::getInstance().releaseConnection(shared_from_this());
 
 	std::lock_guard<std::recursive_mutex> lockClass(connectionLock);
-	if (connectionState != CONNECTION_STATE_OPEN) {
+	if (connectionState == CONNECTION_STATE_DISCONNECTED) {
 		return;
 	}
-	connectionState = CONNECTION_STATE_CLOSED;
+	connectionState = CONNECTION_STATE_DISCONNECTED;
 
 	if (protocol) {
 		g_dispatcher.addTask(
@@ -108,6 +108,7 @@ void Connection::accept(Protocol_ptr protocol)
 {
 	this->protocol = protocol;
 	g_dispatcher.addTask(createTask(std::bind(&Protocol::onConnect, protocol)));
+	connectionState = CONNECTION_STATE_CONNECTING;
 
 	accept();
 }
@@ -119,14 +120,51 @@ void Connection::accept()
 		readTimer.expires_from_now(boost::posix_time::seconds(CONNECTION_READ_TIMEOUT));
 		readTimer.async_wait(std::bind(&Connection::handleTimeout, std::weak_ptr<Connection>(shared_from_this()), std::placeholders::_1));
 
-		// Read size of the first packet
-		boost::asio::async_read(socket,
-		                        boost::asio::buffer(msg.getBuffer(), NetworkMessage::HEADER_LENGTH),
-		                        std::bind(&Connection::parseHeader, shared_from_this(), std::placeholders::_1));
+		if (connectionState == CONNECTION_STATE_CONNECTING) {
+			// Read one-by-one ascii characters (new-line-terminated) to verify world name
+			boost::asio::async_read(socket,
+									boost::asio::buffer(msg.getBuffer(), 1),
+									std::bind(&Connection::parseProxyWorldNameIdentification, shared_from_this(), std::placeholders::_1));
+		} else {
+			// Read size of the first packet
+			boost::asio::async_read(socket,
+									boost::asio::buffer(msg.getBuffer(), NetworkMessage::HEADER_LENGTH),
+									std::bind(&Connection::parseHeader, shared_from_this(), std::placeholders::_1));
+		}
+
+		
 	} catch (boost::system::system_error& e) {
 		std::cout << "[Network error - Connection::accept] " << e.what() << std::endl;
 		close(FORCE_CLOSE);
 	}
+}
+
+void Connection::parseProxyWorldNameIdentification(const boost::system::error_code& error)
+{
+	std::lock_guard<std::recursive_mutex> lockClass(connectionLock);
+	readTimer.cancel();
+
+	if (error) {
+		close(FORCE_CLOSE);
+		return;
+	} else if (connectionState == CONNECTION_STATE_DISCONNECTED) {
+		return;
+	}
+
+	char b = msg.getBuffer()[0];
+	if (b == 0x0A) {
+		if (worldName != g_config.getString(ConfigManager::SERVER_NAME)) {
+			std::cout << convertIPToString(getIP()) << " disconnected for sending invalid world name." << std::endl;
+			close();
+			return;
+		}
+
+		connectionState = CONNECTION_STATE_CONNECTED;
+	} else {
+		worldName.push_back(b);
+	}
+
+	accept();
 }
 
 void Connection::parseHeader(const boost::system::error_code& error)
@@ -137,7 +175,7 @@ void Connection::parseHeader(const boost::system::error_code& error)
 	if (error) {
 		close(FORCE_CLOSE);
 		return;
-	} else if (connectionState != CONNECTION_STATE_OPEN) {
+	} else if (connectionState == CONNECTION_STATE_DISCONNECTED) {
 		return;
 	}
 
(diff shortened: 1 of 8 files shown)


lBaah:
In my opinion client 10 should not be supported anymore.

marcomoa:
I agree with @lBaah, I agree that protocol 10 should no longer be supported, wouldn't it be better to add the latest support already?

lBaah:
I'm going far with this and say not to use Cipsoft client anymore.
I think TFS should be developed sided with OTClient or Unity Client and so on. We need more freedom of development and allow more than RL servers happen.

This PR is a great way to start discussing that.

EPuncker:
> I'm going far with this and say not to use Cipsoft client anymore.
> I think TFS should be developed sided with OTClient or Unity Client and so on. We need more freedom of development and allow more than RL servers happen.
> 
> This PR is a great way to start discussing that.

TFS is already intended to be used with OTC, that is why the latest supported version is 10.98, because that is the version OTC stopped

lBaah:
> > I'm going far with this and say not to use Cipsoft client anymore.
> > I think TFS should be developed sided with OTClient or Unity Client and so on. We need more freedom of development and allow more than RL servers happen.
> > This PR is a great way to start discussing that.
> 
> TFS is already intended to be used with OTC, that is why the latest supported version is 10.98, because that is the version OTC stopped

In my view, TFS is being developed to keep Cipsoft clients compatible.

EPuncker:
> > > I'm going far with this and say not to use Cipsoft client anymore.
> > > I think TFS should be developed sided with OTClient or Unity Client and so on. We need more freedom of development and allow more than RL servers happen.
> > > This PR is a great way to start discussing that.
> > 
> > 
> > TFS is already intended to be used with OTC, that is why the latest supported version is 10.98, because that is the version OTC stopped
> 
> In my view, TFS is being developed to keep Cipsoft clients compatible.

mark words: https://github.com/otland/forgottenserver/pull/2464#issuecomment-392544351

src/connection.cpp
```diff
@@ -119,16 +120,53 @@ void Connection::accept()
...
+	if (b == 0x0A) {
+		if (worldName != g_config.getString(ConfigManager::SERVER_NAME)) {
+			std::cout << convertIPToString(getIP()) << " disconnected for sending invalid world name." << std::endl;
+			close();
+			return;
+		}
+
+		connectionState = CONNECTION_STATE_CONNECTED;
+	} else {
+		worldName.push_back(b);
```
slavidodo: use emplace_back

src/protocolgame.cpp
```diff
@@ -1007,6 +1010,13 @@ void ProtocolGame::parseEditVip(NetworkMessage& msg)
 	const std::string description = msg.getString();
 	uint32_t icon = std::min<uint32_t>(10, msg.get<uint32_t>()); // 10 is max icon in 9.63
```
EPuncker: is 10 still the max icon in client 11?

src/connection.cpp
pawliczka: for trivial type emplace_back ? :D

src/connection.cpp
```diff
@@ -119,16 +120,53 @@ void Connection::accept()
...
+		if (connectionState == CONNECTION_STATE_CONNECTING) {
+			// Read one-by-one ascii characters (new-line-terminated) to verify world name
+			boost::asio::async_read(socket,
+									boost::asio::buffer(msg.getBuffer(), 1),
+									std::bind(&Connection::parseProxyWorldNameIdentification, shared_from_this(), std::placeholders::_1));
+		} else {
+			// Read size of the first packet
+			boost::asio::async_read(socket,
+									boost::asio::buffer(msg.getBuffer(), NetworkMessage::HEADER_LENGTH),
+									std::bind(&Connection::parseHeader, shared_from_this(), std::placeholders::_1));
```
pawliczka: Prefer lambda instead of std::bind. std::bind can be hard to read and can result in larger object files and binaries due to type information that will not be produced by equivalent lambdas.

src/connection.cpp
```diff
@@ -119,16 +120,53 @@ void Connection::accept()
...
 
-		// Read size of the first packet
-		boost::asio::async_read(socket,
-		                        boost::asio::buffer(msg.getBuffer(), NetworkMessage::HEADER_LENGTH),
-		                        std::bind(&Connection::parseHeader, shared_from_this(), std::placeholders::_1));
+		if (connectionState == CONNECTION_STATE_CONNECTING) {
+			// Read one-by-one ascii characters (new-line-terminated) to verify world name
+			boost::asio::async_read(socket,
+									boost::asio::buffer(msg.getBuffer(), 1),
+									std::bind(&Connection::parseProxyWorldNameIdentification, shared_from_this(), std::placeholders::_1));
```
pawliczka: Prefer lambda instead of std::bind. std::bind can be hard to read and can result in larger object files and binaries due to type information that will not be produced by equivalent lambdas.

src/connection.cpp
pawliczka: BTW if you have no idea what you are doing always use push_back.
https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/LubvTXxs7k4/ThMsbV23DQAJ

EPuncker:
I tried logging in, but I can't get past the character selection, it just says that server is offline

slavidodo:
What is the client version (including build version) that you used?
How did you pass login information in the PHP script?

EPuncker:
> What is the client version (including build version) that you used?
> How did you pass login information in the PHP script?

the client version is 12 something (I couldn't find a 11 version) [this one](https://mega.nz/#!T2JQDIAY!BNTeG-DXrbsxrhJK61Srmg8MwOI6Ix_ByUnCMSmLkNw) 

and the login was sent by znote AAC [here](https://github.com/Znote/ZnoteAAC/blob/master/login.php#L55)

slavidodo:
This is supposed to work only wit 11.10 as a maximum. You can find it somewhere on otland!

Leo32onGIT:
I see high value in having branches that are only protocolgame differences of the latest major version of the cip client.

i.e.
- Branch for the latest version of 10.X
- Branch for the latest version of 11.XX
- Branch for the latest release of 12.XX until 13.0 is released.

This allows us to officially retain parity with whatever open-source client we wish to tether to;
But also have an ease-of-use pathway for those who want to use TFS with a specific client/protocol.

soul4soul:
> 
> 
> I see high value in having branches that are only protocolgame differences.
> Branches should only exist for the last major release of each protocol:
> 
> i.e.
> 
>     * Branch for the latest version of 10.X
> 
>     * Branch for the latest version of 11.XX
> 
>     * Branch for the latest release of 12.XX until 13.0 is released.
> 
> 
> This allows us to officially retain parity with whatever open-source client we wish to tether to;
> But also have an ease-of-use pathway for those who want to use TFS with a specific client/protocol.
> This seems to be the best compromise - to satisfy both parties.

Not saying it was done well or not but pyOT engine supports multiple protocols. It was possible to connect with 8.0 client and 9.0 client to the same server. The engine took care of the protocol differences. It was very cool system.  The biggest advantage of such a system was having a single code base.

There was discussions on what would happen if new IDs are used in the map and someone connected with an old client. The temporary solution that was used was to replace items with some predefined "mapping", items there couldn't be replaced weren't sent to the old clients. At the time the only "mapping" that existed was to replace all unknown floor titles with grass. Using clients newer then what was to used to map the server worked seamlessly.

I'd still be a fan of seeing a single engine supporting multiple protocols but perhaps it would make sense to force a single version to be specified in `config.lua`.

Leo32onGIT:
> The engine took care of the protocol differences. It was a very cool system. A big advantage of such a system was having a single code base.

I don't see this happening after having branches for each protocolgame state.
As of today, branches for [7.72](https://github.com/nekiro/forgottenserver/tree/7.72-downgrade) and [8.60](https://github.com/nekiro/forgottenserver/tree/8.6-downgrade) already exist. Having the same thing for 9.X+ just makes sense.

@nekiro really cool buddy :)

EPuncker:
@slavidodo any chance to make it possible to "play" with both 10.98 and 11 clients, so we can try using/adding 11 client functionalities (prey etc), while still being backward compatible, so maybe it wont take as long as tfs 2.0 to be implemented?

 what do you think @DSpeichert

DSpeichert:
I don't have the confidence this will be possible bug-free at this point. Probably not likely to happen.

slavidodo:
The key to maintaining this is having either OpenTibia-Unity/OTClient compatible with these versions.
Neither me have the confidence to say this is bug-free, since tibia 11 changed a lot of things and these were only what I caught, besides I need to push more tests and then I will resubmit this PR.

Zbizu:
outdated

PR #2728, opened 2019-11-27
Small rework on the code for closing a door with a creature in it

Sundance:
The code seemed to be unfinished, with this commit it's accurate to how it works on RL tibia.
In RL tibia it will move the creature in the following prioritization order:
east -> south -> west -> north
-> position where queryAdd() == 0 (non blocked walkable tile) without any creature
-> the previous with a creature
-> position with movable tile-blocking items
-> position with magic wall or wild growth

### DIFF
--- a/data/actions/scripts/other/doors.lua
+++ b/data/actions/scripts/other/doors.lua
@@ -1,3 +1,62 @@
+local positionOffsets = {
+	Position(1, 0, 0), -- east
+	Position(0, 1, 0), -- south
+	Position(-1, 0, 0), -- west
+	Position(0, -1, 0) -- north
+}
+
+--[[
+When closing a door with a creature in it findPushPosition will find the most appropriate
+adjacent position following a prioritization order.
+The function returns the position of the first tile that fulfills all the checks in a round.
+The function loops trough east -> south -> west -> north on each following line in that order.
+In round 1 it checks if there's an unhindered walkable tile without any creature.
+In round 2 it checks if there's a tile with a creature.
+In round 3 it checks if there's a tile blocked by a movable tile-blocking item.
+In round 4 it checks if there's a tile blocked by a magic wall or wild growth.
+]]
+local function findPushPosition(creature, round)
+	local pos = creature:getPosition()
+	for _, offset in ipairs(positionOffsets) do
+		local offsetPosition = pos + offset
+		local tile = Tile(offsetPosition)
+		if tile then
+			local creatureCount = tile:getCreatureCount()
+			if round == 1 then
+				if tile:queryAdd(creature) == 0 and creatureCount == 0 then
+					if tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz() then
+						return offsetPosition
+					elseif not tile:hasFlag(TILESTATE_PROTECTIONZONE) then
+						return offsetPosition
+					end
+				end
+			elseif round == 2 then
+				if creatureCount > 0 then
+					if tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz() then
+						return offsetPosition
+					elseif not tile:hasFlag(TILESTATE_PROTECTIONZONE) then
+						return offsetPosition
+					end
+				end
+			elseif round == 3 then
+				local topItem = tile:getTopDownItem()
+				if topItem then
+					if topItem:getType():isMovable() then
+						return offsetPosition
+					end
+				end
+			else
+				if tile:getItemById(ITEM_MAGICWALL) or tile:getItemById(ITEM_WILDGROWTH) then
+					return offsetPosition
+				end
+			end
+		end
+	end
+	if round ~= 4 then
+		return findPushPosition(creature, round + 1)
+	end
+end
+
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 	local itemId = item:getId()
 	if table.contains(questDoors, itemId) then
@@ -29,22 +88,16 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 	end
 
 	if table.contains(horizontalOpenDoors, itemId) or table.contains(verticalOpenDoors, itemId) then
-		local doorCreature = Tile(toPosition):getTopCreature()
-		if doorCreature then
-			toPosition.x = toPosition.x + 1
-			local query = Tile(toPosition):queryAdd(doorCreature, bit.bor(FLAG_IGNOREBLOCKCREATURE, FLAG_PATHFINDING))
-			if query ~= RETURNVALUE_NOERROR then
-				toPosition.x = toPosition.x - 1
-				toPosition.y = toPosition.y + 1
-				query = Tile(toPosition):queryAdd(doorCreature, bit.bor(FLAG_IGNOREBLOCKCREATURE, FLAG_PATHFINDING))
+		local creaturePositionTable = {}
+		local doorCreatures = Tile(toPosition):getCreatures()
+		if #doorCreatures > 0 then
+			for _, doorCreature in pairs(doorCreatures) do
+				local pushPosition = findPushPosition(doorCreature, 1)
+				table.insert(creaturePositionTable, {creature = doorCreature, position = pushPosition})
 			end
-
-			if query ~= RETURNVALUE_NOERROR then
-				player:sendTextMessage(MESSAGE_STATUS_SMALL, Game.getReturnMessage(query))
-				return true
+			for _, tableCreature in ipairs(creaturePositionTable) do
+				tableCreature.creature:teleportTo(tableCreature.position, true)
 			end
-
-			doorCreature:teleportTo(toPosition, true)
 		end
 
 		if not table.contains(openSpecialDoors, itemId) then
--- a/data/actions/scripts/other/doors.lua
+++ b/data/actions/scripts/other/doors.lua
@@ -23,18 +23,14 @@ local function findPushPosition(creature, round)
 		if tile then
 			local creatureCount = tile:getCreatureCount()
 			if round == 1 then
-				if tile:queryAdd(creature) == 0 and creatureCount == 0 then
-					if tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz() then
-						return offsetPosition
-					elseif not tile:hasFlag(TILESTATE_PROTECTIONZONE) then
+				if tile:queryAdd(creature) == RETURNVALUE_NOERROR and creatureCount == 0 then
+					if not tile:hasFlag(TILESTATE_PROTECTIONZONE) or (tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz()) then
 						return offsetPosition
 					end
 				end
 			elseif round == 2 then
 				if creatureCount > 0 then
-					if tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz() then
-						return offsetPosition
-					elseif not tile:hasFlag(TILESTATE_PROTECTIONZONE) then
+					if not tile:hasFlag(TILESTATE_PROTECTIONZONE) or (tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz()) then
 						return offsetPosition
 					end
 				end
@@ -52,7 +48,7 @@ local function findPushPosition(creature, round)
 			end
 		end
 	end
-	if round ~= 4 then
+	if round < 4 then
 		return findPushPosition(creature, round + 1)
 	end
 end
@@ -93,6 +89,10 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 		if #doorCreatures > 0 then
 			for _, doorCreature in pairs(doorCreatures) do
 				local pushPosition = findPushPosition(doorCreature, 1)
+				if not pushPosition then
+					player:sendCancelMessage(RETURNVALUE_NOTENOUGHROOM)
+					return true
+				end
 				table.insert(creaturePositionTable, {creature = doorCreature, position = pushPosition})
 			end
 			for _, tableCreature in ipairs(creaturePositionTable) do
--- a/data/lib/core/creature.lua
+++ b/data/lib/core/creature.lua
@@ -167,3 +167,11 @@ function Creature:addDamageCondition(target, type, list, damage, period, rounds)
 	target:addCondition(condition)
 	return true
 end
+
+
+function Creature:canAccessPz()
+	if self:isMonster() or (self:isPlayer() and self:isPzLocked()) then
+		return false
+	end
+	return true
+end
--- a/data/lib/core/creature.lua
+++ b/data/lib/core/creature.lua
@@ -168,7 +168,6 @@ function Creature:addDamageCondition(target, type, list, damage, period, rounds)
 	return true
 end
 
-
 function Creature:canAccessPz()
 	if self:isMonster() or (self:isPlayer() and self:isPzLocked()) then
 		return false


data/actions/scripts/other/doors.lua
```diff
@@ -1,3 +1,51 @@
+local positions = {
+	{x = 1, y = 0}, -- east
+	{x = -1, y = 1}, -- south
+	{x = -1, y = -1}, -- west
+	{x = 1, y = -1} -- north
+}
+
+local function moveToAptPos(creature, round)
```
Znote: I don't understand this function name. Could you make it more obvious as to what it does? What is Apt?

Sundance: You guessed correctly, Apt is another word for appropriate.
I'll argue that it shouldn't be changed as it's not an abbreviation, but it's own word.

data/actions/scripts/other/doors.lua
```diff
@@ -1,3 +1,51 @@
...
+	for k,v in pairs(positions) do
+		pos.x = pos.x + v.x
+		pos.y = pos.y + v.y
+		local tile = Tile(pos)
+		if tile then
+			if round <= 2 then
+				if tile:queryAdd(creature, bit.bor(FLAG_IGNOREBLOCKCREATURE, FLAG_PATHFINDING)) == 0 then
+					if tile:getCreatureCount() == 0 and round == 1 then
+						creature:teleportTo(pos, true)
+						return
```
ranisalt: You may `return creature:teleportTo` to pass on return value.

data/actions/scripts/other/doors.lua
DSpeichert: This isn't a clear function name, regardless of the word chosen. Maybe add a description of what and how it does what it does?

data/actions/scripts/other/doors.lua
```diff
@@ -1,3 +1,51 @@
...
+					return
+				end
+			end
+		end
+	end
+	if round ~= 4 then
+		return moveToAptPos(creature, round+1)
+	end
+	if creature:isPlayer() then
+		creature:sendTextMessage(MESSAGE_STATUS_SMALL, Game.getReturnMessage(2))
```
TheSumm: `creature:sendCancelMessage(RETURNVALUE_NOTENOUGHROOM)`

data/actions/scripts/other/doors.lua
```diff
@@ -1,3 +1,51 @@
+local positions = {
+	{x = 1, y = 0}, -- east
+	{x = -1, y = 1}, -- south
```
TheSumm: It would be cleaner if you just use the actual offsets here e.g. `Position(1, 0, 0) -- east`, `Position(0, 1, 0) -- south` etc.

data/actions/scripts/other/doors.lua
```diff
@@ -1,3 +1,51 @@
...
+	{x = -1, y = -1}, -- west
+	{x = 1, y = -1} -- north
+}
+
+local function moveToAptPos(creature, round)
+	local pos = creature:getPosition()
+	for k,v in pairs(positions) do
+		pos.x = pos.x + v.x
```
TheSumm: If you use the Position constructor (see above), you can construct the offset position like this:
`local offsetPosition = pos + v`

data/actions/scripts/other/doors.lua
```diff
@@ -1,3 +1,51 @@
+local positions = {
+	{x = 1, y = 0}, -- east
+	{x = -1, y = 1}, -- south
+	{x = -1, y = -1}, -- west
+	{x = 1, y = -1} -- north
+}
+
+local function moveToAptPos(creature, round)
+	local pos = creature:getPosition()
+	for k,v in pairs(positions) do
```
TheSumm: `for _, offset in ipairs(positions) do`

data/actions/scripts/other/doors.lua
```diff
@@ -1,3 +1,51 @@
...
+		pos.x = pos.x + v.x
+		pos.y = pos.y + v.y
+		local tile = Tile(pos)
+		if tile then
+			if round <= 2 then
```
TheSumm: Would be nice to have comments stating what is checked in each round.

data/actions/scripts/other/doors.lua
Sundance: The only way the function would get to that part is if a mapper blocked all 4 exits with immovable items (which I've yet to see any mapper do) so I removed it.

TheSumm:
The changes LGTM.

`moveToAptPos` should still be renamed. Maybe something like `moveTo(Free)AdjacentPosition` (without any abbreviations).

Another question: Does this account for multiple creatures being stacked in a door? e.g. in a protection zone or through other means

data/actions/scripts/other/doors.lua
Znote: What about the function name `pushFromDoor` or `pushCreatureFromDoor`?
Since this is only used in doors.lua, and this is strictly speaking, what the function does?
"AptPos" is a bad abbrevation, and "AppropriatePosition" is ambigious.

ranisalt: Apt is not an abbreviation, it has a meaning closer to "possible" rather than "appropriate". "Apt" is an apt word here, but maybe not appropriate :P

However I would rename it to `findKickPosition` and return the position instead of moving the player, and move inside `onUse` to avoid a function with multiple responsibilities.

Sundance: To be precise apt is a homonym so the meaning depends on the context in which it is used, either way I agree that it is ambiguous.

How about findPushPosition? Imo push fits better than kick when a creature is being relocated by a closing door.

To be sure, by "move inside onUse" do you mean I should nest findPushPosition inside onUse?
And thanks I didn't know about Curly's law before.

Sundance:
> The changes LGTM.
> 
> `moveToAptPos` should still be renamed. Maybe something like `moveTo(Free)AdjacentPosition` (without any abbreviations).
> 
> Another question: Does this account for multiple creatures being stacked in a door? e.g. in a protection zone or through other means

@TheSumm renamed it to findPushPosition and had it return the position instead.
As ranisalt mentioned the function was handling more 1 thing which is no bueno (Curly's law).
Stacked creatures are accounted for now.

Sundance:
@Znote I didn't understand that flag FLAG_IGNOREBLOCKCREATURE only ignores pushable creatures (I was using rats in my tests).
`if tile:queryAdd(creature, bit.bor(FLAG_IGNOREBLOCKCREATURE, FLAG_PATHFINDING)) == 0 then`
is at fault. When the script wants to push a monster (doesn't affect players or npcs) onto another creature with the "can't be pushed" flag `tile:queryAdd` will return 2 (RETURNVALUE_NOTENOUGHROOM)

True, the position wouldn't necessarily be the same for all creatures. Although I can't think of a scenario where a door that does not belong to a house would be placed next to a house tile, the following poses an issue:

Should houses be accounted for in the next edit as well?

The creature(s) should always have somewhere to go, the only case in which a position can't be found would be if all exits are blocked by immovable blocking items. In such a rare case I believe the mapping has to be questioned.

I have an update ready, just let me know about the houses.
I created a new function "Creature:canAccessPz()" for the pz checks should that be in a separate PR?
Should it even be a global function or just local?

MillhioreBT:
If you test it in this situation it produces a problem, because you do not verify if the position obtained is not equal to nil
image

So I made a small addition and it is already 100% error-free, I tested it in many situations, including the most absurd and it still does not produce any errors, I also added the verification so that a normal player cannot close a door if a player with privileges is at the door.

<details>

<summary>door.lua</summary>

```lua
local positionOffsets = {
	Position(1, 0, 0), -- east
	Position(0, 1, 0), -- south
	Position(-1, 0, 0), -- west
	Position(0, -1, 0) -- north
}

--[[
When closing a door with a creature in it findPushPosition will find the most appropriate
adjacent position following a prioritization order.
The function returns the position of the first tile that fulfills all the checks in a round.
The function loops trough east -> south -> west -> north on each following line in that order.
In round 1 it checks if there's an unhindered walkable tile without any creature.
In round 2 it checks if there's a tile with a creature.
In round 3 it checks if there's a tile blocked by a movable tile-blocking item.
In round 4 it checks if there's a tile blocked by a magic wall or wild growth.
]]
local function findPushPosition(creature, round)
	local pos = creature:getPosition()
	for _, offset in ipairs(positionOffsets) do
		local offsetPosition = pos + offset
		local tile = Tile(offsetPosition)
		if tile then
			local creatureCount = tile:getCreatureCount()
			if round == 1 then
				if tile:queryAdd(creature) == RETURNVALUE_NOERROR and creatureCount == 0 then
					if not tile:hasFlag(TILESTATE_PROTECTIONZONE) or (tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz()) then
						return offsetPosition
					end
				end
			elseif round == 2 then
				if creatureCount > 0 then
					if not tile:hasFlag(TILESTATE_PROTECTIONZONE) or (tile:hasFlag(TILESTATE_PROTECTIONZONE) and creature:canAccessPz()) then
						return offsetPosition
					end
				end
			elseif round == 3 then
				local topItem = tile:getTopDownItem()
				if topItem then
					if topItem:getType():isMovable() then
						return offsetPosition
					end
				end
			else
				if tile:getItemById(ITEM_MAGICWALL) or tile:getItemById(ITEM_WILDGROWTH) then
					return offsetPosition
				end
			end
		end
	end
	if round < 4 then
		return findPushPosition(creature, round +1)
	end
end

	local itemId = item:getId()
	if table.contains(questDoors, itemId) then
		if player:getStorageValue(item.actionid) ~= -1 then
			item:transform(itemId + 1)
			player:teleportTo(toPosition, true)
		else
			player:sendTextMessage(MESSAGE_EVENT_ADVANCE, "The door seems to be sealed against unwanted intruders.")
		end
		return true
	elseif table.contains(levelDoors, itemId) then
		if item.actionid > 0 and player:getLevel() >= item.actionid - actionIds.levelDoor then
			item:transform(itemId + 1)
			player:teleportTo(toPosition, true)
		else
			player:sendTextMessage(MESSAGE_EVENT_ADVANCE, "Only the worthy may pass.")
		end
		return true
	elseif table.contains(keys, itemId) then
		if target.actionid > 0 then
			if item.actionid == target.actionid and doors[target.itemid] then
				target:transform(doors[target.itemid])
				return true
			end
			player:sendTextMessage(MESSAGE_STATUS_SMALL, "The key does not match.")
			return true
		end
		return false
	end

	if table.contains(horizontalOpenDoors, itemId) or table.contains(verticalOpenDoors, itemId) then
		local creaturePositionTable = {}
		local doorCreatures = Tile(toPosition):getCreatures()
		if #doorCreatures > 0 then
				for _, c in pairs(doorCreatures) do
					if creature ~= c and c:getGroup():getAccess() then
						player:sendCancelMessage("You cannot close this door.")
						return true
					end
				end
			end
			for _, doorCreature in pairs(doorCreatures) do
				local pushPosition = findPushPosition(doorCreature, 1)
				if pushPosition then
					table.insert(creaturePositionTable, {creature = doorCreature, position = pushPosition})
				end
			end
			if #creaturePositionTable == 0 then
				player:sendCancelMessage(RETURNVALUE_NOTENOUGHROOM)
				return true
			end 
			for _, tableCreature in ipairs(creaturePositionTable) do
				tableCreature.creature:teleportTo(tableCreature.position, true)
			end
		end

		if not table.contains(openSpecialDoors, itemId) then
			item:transform(itemId - 1)
		end
		return true
	end

	if doors[itemId] then
		if item.actionid == 0 then
			item:transform(doors[itemId])
		else
			player:sendTextMessage(MESSAGE_EVENT_ADVANCE, "It is locked.")
		end
		return true
	end
	return false
end
```

</details>

soul4soul:
Good work improving doors. Even cip doors can become buggy when there is no room to move a creature. In [this video at the 9 minute mark](https://youtu.be/EPWovUBqFIU?t=535) a door is closed and the creature is stuck standing on the door. The creature should of been pushed north and stacked. You can see on the second attempt to close the door the creature is pushed north correctly. One way to solve that problem is to prevent the door from being closed when there is no possible room for the creature to be pushed onto.

MillhioreBT:
> Good work improving doors. Even cip doors can become buggy when there is no room to move a creature. In [this video at the 9 minute mark](https://youtu.be/EPWovUBqFIU?t=535) a door is closed and the creature is stuck standing on the door. The creature should of been pushed north and stacked. You can see on the second attempt to close the door the creature is pushed north correctly. One way to solve that problem is to prevent the door from being closed when there is no possible room for the creature to be pushed onto.

This no longer happens if you use the script **door.lua**
GIF 10-05-2020 01-36-12 p  m

Sundance:
> The creature should of been pushed north and stacked

Should've been pushed south* weird bug nonetheless 😮  

> If you test it in this situation it produces a problem, because you do not verify if the position obtained is not equal to nil

I did think about that but rationalized that such a rare situation would be the mappers fault.
Since it's just a few extra lines of code there's nothing to lose from adding it tho.
Ty for the code, applied your addition 👍

data/lib/core/creature.lua
```diff
@@ -167,3 +167,11 @@ function Creature:addDamageCondition(target, type, list, damage, period, rounds)
 	target:addCondition(condition)
 	return true
 end
+
+
```
DSpeichert, line 171: Rest of the file uses single-newline separation, remove one blank line

gigastar:
What about something like this.
[doors.zip](https://github.com/otland/forgottenserver/files/4672123/doors.zip)

Xaekai:
I can't believe you didn't name the function `eviction`

soul4soul:
> 
> 
> > Good work improving doors. Even cip doors can become buggy when there is no room to move a creature. In [this video at the 9 minute mark](https://youtu.be/EPWovUBqFIU?t=535) a door is closed and the creature is stuck standing on the door. The creature should of been pushed north and stacked. You can see on the second attempt to close the door the creature is pushed north correctly. One way to solve that problem is to prevent the door from being closed when there is no possible room for the creature to be pushed onto.
> 
> This no longer happens if you use the script **door.lua**
> GIF 10-05-2020 01-36-12 p m

Is that picture before or after the change? To mirror cipbia the monster should be pushed north and stack on top of the other demon.

Juanzin420:
i just made these changes and it didnt work, the creature gets pushed if i close the door while it is walking to the door sqm, but if its already standing there it doesnt work

https://gyazo.com/92b42853f791e6203d234663c4597965

https://gyazo.com/58b4849639fffa453f407dd6cdcffd8b

would u know how to help me @EPuncker ? xD

Znote:
> i just made these changes and it didnt work, the creature gets pushed if i close the door while it is walking to the door sqm, but if its already standing there it doesnt work
> 
> https://gyazo.com/92b42853f791e6203d234663c4597965
> 
> https://gyazo.com/58b4849639fffa453f407dd6cdcffd8b
> 
> would u know how to help me @EPuncker ? xD

doors2

I am unable to reproduce your issue. It appears to be a problem with otx4/OtservBR and not TFS.

Juanzin420:
im using tfs but it seems to be a older version, any idea what it could be?

emil92b:
would be cool if gate of expertise aswell quest doors got some love aswell : ) talking about "StepOut" open doors, and their creature/item relocation

for example in black knight quest cave, there are 2 gate of expertise doors, if you stand in the left open door and put item under you and click on the right closed door = the creature and item should be moved to the right (same position), and if you step out of that one the item will be moved south coz there is a wall to the east, so im pretty sure it uses the same logic (east -> south -> west -> north)

if wana go further that that, look into relogging inside those doors and check what happens if blocking with objects on the free positions, i have my suspicions that the order would be something like starting from east and go anti clockwise, but i could be wrong.

regards

PR #2740, opened 2019-12-05
Use numeric indexes for spell vocations

infernumx:
Now we can loop voc table with ipairs!


Znote:
How do you use this?

```lua
player:sendTextMessage(MESSAGE_INFO_DESCR, dump(Spell("Light healing"):vocation()))
```

image

PR #2746, opened 2019-12-23
Better Revscripts Constructors

delusionx:
Can now initialize an entire object with a table definition, rather than tediously applying methods to the newly-created object. Has backwards compatibility, so you can switch between default constructor and the custom ones.

Example:
```lua
local talk = TalkAction {
    init = {"/pos"},
    separator = " "
}

    if player:getGroup():getAccess() and param ~= "" then
        player:teleportTo(Position(split[1], split[2], split[3]))
    else
        player:sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, "Your current position is: " .. position.x .. ", " .. position.y .. ", " .. position.z .. ".")
    end
    return false
end

talk:register()

```


CristoferMartins:
Yeah, i think you can setup parameters by a default table and also have the individual methods to set parameters.

MillhioreBT:
I was following you on Twitch and I saw how you created this, it's really beautiful how you can build a script now, in my opinion.

What I do not agree with is defining the register function, this should be the opposite, define register in case you wish not to register it, for example: 
image
I understand that to simulate this you can do the following: [https://ghostbin.co/paste/yr99t](url), But I don't like this syntax, unless it is implemented in the mother class, for example: 
image They're just my tastes, but they can't deny me that it would be more intuitive for people who don't know how to scripting.

The fact is that we don't have that yet, for now I think this would be an improvement for this builder:
image
image
Example of scripts/lib/helper_constructor.lua: [https://ghostbin.co/paste/eeekc](url)

delusionx:
I don't see the issue with register(), like I said, this still maintains backwards compatibility and only defines what you make it define in the table. If you don't want it to instantly register, you can either not define it or set it to false.
I also don't understand the point of having a defaultParam, you cannot change it after the fact so having a default parameter would completely destroy the configurability of the object once it's created if you don't pass an argument to it (which I don't understand why you wouldn't in the first place, if this was a real issue, there should also be defaults in XML, but there isn't).
Same with using table indexes after the fact, there's no use for it afterwards because you can use the previous method to alter the state of the object before its registration. Also the fact that it just defeats the entire purpose of writing this code, because you shouldn't have to be forced to re-type the same object name over and over to configure the object, which gets tedious.
In any case, like I said, defining objects this way is not forced by any means, the old way still works perfectly normal and you can even mix & match by not defining register, and manually doing it later.

MillhioreBT:
What you say is very good, if everything is about compatibility with previous systems, but my modification does not take away the fact that it is still compatible, and I apologize for always being stupid, I just like things to change a little more than normal and can still be compatible, perhaps for new future people who want to learn to write code in one or a thousand different ways with the same result

CristoferMartins:
Something i want to add to the subject, why backward compatibility is so important when this revscripts is so new and as a new thing its obviously that it will envolve and change till a "stable" version. No should use it in production yet.

delusionx:
> Something i want to add to the subject, why backward compatibility is so important when this revscripts is so new and as a new thing its obviously that it will envolve and change till a "stable" version. No should use it in production yet.

Because users should still be able to use the methods directly, this way users get to pick and choose how they want to create scripts.

PR #2747, opened 2019-12-27
ItemType constructor should return nil if nothing is passed to it.

### LINKED ISSUE #2742: ItemType(nil) creates ItemType(22617) userdata

```print(ItemType(nil):getId())```

### Expected behaviour
```0```

### Actual behaviour
```22617```

img

### Environment
appVeyor build 1.0.299 (x64 windows, from 8 dec 2019), also broken on 1.0.277

delusionx:
Also strict type checking for string as the `name` argument, rather than simply interpreting the argument as a string regardless.

Fixes #2742

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -11299,8 +11299,11 @@ int LuaScriptInterface::luaItemTypeCreate(lua_State* L)
 	uint32_t id;
 	if (isNumber(L, 2)) {
 		id = getNumber<uint32_t>(L, 2);
-	} else {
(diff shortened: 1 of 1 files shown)


MillhioreBT:
What do you think about adding this to the function: Items::items.getItemIdByName
image
There are cases in which there are articles without names, and it makes no sense to search for an item by name if you don't really have one.

delusionx:
I believe it should still match an item even if you make its name an empty string, because if it's configurable that way you should still be allowed to retrieve it as well regardless of how dumb it is.

PR #2750, opened 2019-12-29
Multiplier stage in float value. (value 0.1)

ramon-bernardo:
I had problems with the multiplier below 0 and made some changes. If it is incorrect, I ask you to change it.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4779,7 +4779,7 @@ void Game::loadPlayersRecord()
 	}
 }
 
-uint64_t Game::getExperienceStage(uint32_t level)
+float Game::getExperienceStage(uint32_t level)
 {
 	if (!stagesEnabled) {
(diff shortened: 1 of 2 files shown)


MillhioreBT:
It is not really a problem, it is simply that the code was intentionally written so that it only reads integers, with respect to floating point numbers, it was something of the old servers and is still useful for quite custom servers with respect to the experience multiplier, I in My server used it this way too, I guess if it was never published in the master branch it was because nobody has needed it urgently since there are cleaner alternatives such as the LUA system published in otland.

vankk:
This isn't a problem but it could support double values.. Simple question "why not?". 

Instead of using `float` that may bug the experience in Lua just replace to `double`, and the first map value could be uint32_t instead of float in your code 😋 

Also don't know if that will pass due #1642

PR #2751, opened 2019-12-31
Added large seashell

EPuncker:
Script by @zxmatzx on otland, I've just changed a few things to make it "compliant" to others  TFS scripts


TheSumm:
We should probably decide on a way to define constants for storage keys that are used within the core scripts before adding more scripts with arbitrary storage keys.

data/actions/scripts/other/largeseashell.lua
```diff
@@ -0,0 +1,24 @@
...
+			msg = "You found a beautiful pearl."
+		else
+			msg = "Nothing is inside."
+		end
+		player:say(msg, TALKTYPE_MONSTER_SAY, false, player, item:getPosition())
+		item:transform(7553)
+		player:setStorageValue(43000, os.time() + 72000)
+		item:getPosition():sendMagicEffect(CONST_ME_BUBBLES)
+		return true
+	else
```
TheSumm: `else` is not needed since you are returning in the if-block.

data/items/items.xml
```diff
@@ -13166,6 +13166,8 @@
 	</item>
 	<item id="7553" article="a" name="large seashell">
 		<attribute key="description" value="It is empty." />
+		<attribute key="decayTo" value="7552"/>
+		<attribute key="duration" value="900"/>
```
TheSumm: AFAIK it is only 5 minutes, but I am not exactly sure.

data/actions/scripts/other/largeseashell.lua
```diff
@@ -0,0 +1,24 @@
...
+		player:say("You have already opened a shell today.", TALKTYPE_MONSTER_SAY, false, player, item:getPosition())
+		return false
```
TheSumm: Should be `return true`

data/actions/scripts/other/largeseashell.lua
EPuncker: 2019-12-31 23_00_07-Window
this then?

data/global.lua
```diff
@@ -2,6 +2,7 @@ math.randomseed(os.time())
 dofile('data/lib/lib.lua')
 
 STORAGEVALUE_PROMOTION = 30018
+DELAY_LARGE_SEASHELL = 30019
```
DSpeichert: The constant name says nothing about the type, i.e. it should be named `STORAGEVALUE_DELAY_LARGE_SEASHELL`

PR #2755, opened 2020-01-02
Updated copyright year

EPuncker:
Updated copyright year to 2020

### DIFF
--- a/src/account.h
+++ b/src/account.h
@@ -1,6 +1,6 @@
 /**
  * The Forgotten Server - a free and open-source MMORPG server emulator
- * Copyright (C) 2019  Mark Samman <mark.samman@gmail.com>
+ * Copyright (C) 2020  Mark Samman <mark.samman@gmail.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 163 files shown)


soul4soul:
I feel the copyright year should not be updated. Unless you want to change it to 2019-2020, in the future only the second number would be updated each year... which I still prefer not doing.

[We should not do that every year](https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated)

> The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work.
>
> Therefore, if the claim is based on common law copyright (not formally registered), then the date should be the date of first publication.

_Originally posted by @ranisalt in https://github.com/otland/forgottenserver/pull/2590#issuecomment-491904930_

DSpeichert:
I do not see the need to update the date. I'll let Mark do that if he still wants to, since it's his name next to the copyright anyway.

ranisalt:
We should actually find a way to use range years for when files were edited, as to correctly represent copyright.

PR #2756, opened 2020-01-03
Argument handler

rookgaard:
Hello. I would like to introduce another thing what is missing in comparision to 0.X and in my opinion it was very useful - argument handler. The simpliest example could be someone who want to run two similar servers which will share between each other for example data folder, but with slightly different map and that person don't want to copy whole folder with binary and other files to a different location.

Solution is here - thanks to this feature (and TFS developers) - he will just run command (for example in 2 screen sessions) to tfs binary with `--config=world1.lua` and `--config=world2.lua`.

Original solution had more options, but I've copied only some of them to show how it might look like.


src/configmanager.cpp
```diff
@@ -106,7 +111,9 @@ bool ConfigManager::load()
 		boolean[BIND_ONLY_GLOBAL_ADDRESS] = getGlobalBoolean(L, "bindOnlyGlobalAddress", false);
 		boolean[OPTIMIZE_DATABASE] = getGlobalBoolean(L, "startupDatabaseOptimization", true);
 
-		string[IP] = getGlobalString(L, "ip", "127.0.0.1");
```
EPuncker: is this and port checks really needed?

rookgaard: Yes, because when you change IP (or game/login port) from argument it will be set here https://github.com/otland/forgottenserver/pull/2756/files#diff-b7fa32a8694c6de78c4454514026dba8R352 and then during config loading it will overwrite it when detects that value inside config lua file.

DSpeichert:
Isn't there some C++ config-handling library that can automatically provide file/env-var/CLI-flag option handling?

rookgaard:
Maybe there is, but I'm not that good in C++, so I offered only solution from 0.4.

src/otserv.cpp
```diff
@@ -307,3 +313,48 @@ void mainLoader(int, char*[], ServiceManager* services)
 	g_game.setGameState(GAME_STATE_NORMAL);
 	g_loaderSignal.notify_all();
 }
+
+bool argumentsHandler(StringVector args)
+{
+	StringVector tmp;
+
+	for(StringVector::iterator it = args.begin(); it != args.end(); ++it) {
```
djarek:
```suggestion
	for(const auto& arg : args) {
```

src/otserv.cpp
```diff
@@ -307,3 +313,48 @@ void mainLoader(int, char*[], ServiceManager* services)
...
+			std::cout << "x86" << std::endl;
+#elif defined(__arm__)
+			std::cout << "ARM" << std::endl;
+#else
+			std::cout << "unknown" << std::endl;
+#endif
+			return false;
+		}
+
+		tmp = explodeString((*it), "=");
```
djarek:
```suggestion
		StringVector tmp = explodeString((*it), "=");
```

No need to have the variable declared in the outside scope, since we always assign over the vector. This way the reader clearly knows it's only used within this iteration.

rookgaard:
Thanks @djarek for review. I've made changes you proposed.

src/configmanager.h
```diff
@@ -23,6 +23,9 @@
 class ConfigManager
 {
 	public:
+		ConfigManager();
+		virtual ~ConfigManager() {}
```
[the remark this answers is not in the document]
pawliczka: contructor also. This is broken basic rule of zero/five from cppCoreGuideLines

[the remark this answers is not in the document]
djarek: Rules of 5 and 0 are about destructor, move/copy constructors and move/copy assignment: https://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_five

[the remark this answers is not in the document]
pawliczka: Yes and here you have empty user defined destructor = broken rule of five. No move ctor or operator will be generated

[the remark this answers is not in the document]
pawliczka: Yes and here you have empty user defined destructor = broken rule of five. No move ctor or operator will be generated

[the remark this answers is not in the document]
pawliczka: In this case please remove ctor and dtor due to rule of zero. Compiler is smarter than you. Trust me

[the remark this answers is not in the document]
djarek: Not true: https://godbolt.org/z/P3f2xz
Here's a table that explains when and what is generated/deleted/not generated:
https://i.stack.imgur.com/b2VBV.png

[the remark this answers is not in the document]
pawliczka: I have said dtor. You are taking about ctor 😂 we are both right.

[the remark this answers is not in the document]
pawliczka: I though that ctor is also empty and because of that i said that this code broke rule of zero due to empty ctor and rule of five due to empty dtor

src/configmanager.cpp
```diff
@@ -95,7 +100,7 @@ bool ConfigManager::load()
 
 	luaL_openlibs(L);
 
-	if (luaL_dofile(L, "config.lua")) {
+	if (luaL_dofile(L, string[CONFIG_FILE].c_str())) {
```
nekiro: Why not getString method?

src/configmanager.cpp
```diff
@@ -106,7 +111,9 @@ bool ConfigManager::load()
 		boolean[BIND_ONLY_GLOBAL_ADDRESS] = getGlobalBoolean(L, "bindOnlyGlobalAddress", false);
 		boolean[OPTIMIZE_DATABASE] = getGlobalBoolean(L, "startupDatabaseOptimization", true);
 
-		string[IP] = getGlobalString(L, "ip", "127.0.0.1");
+		if (string[IP] == "")
+			string[IP] = getGlobalString(L, "ip", "127.0.0.1");
```
nekiro: curly brackets? same with the others

src/otserv.cpp
```diff
@@ -70,6 +71,10 @@ void mainLoader(int argc, char* argv[], ServiceManager* services);
 
 int main(int argc, char* argv[])
 {
+	StringVector args = StringVector(argv, argv + argc);
+	if(argc > 1 && !argumentsHandler(args))
```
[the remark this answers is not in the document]
djarek: `argv[0]` always contains the executable path, so if `argc` is 1, there are no arguments.

rookgaard:
@nekiro @djarek I've committed changes you requested, how this PR looks like now?

rookgaard:
@EPuncker @djarek @pawliczka @DSpeichert @nekiro is there anything I can do more to finish this feature?

nekiro (approved):

src/otserv.cpp
```diff
@@ -144,12 +150,13 @@ void mainLoader(int, char*[], ServiceManager* services)
 	std::cout << std::endl;
 
 	// check if config.lua or config.lua.dist exist
-	std::ifstream c_test("./config.lua");
+	const std::string& configFile = g_config.getString(ConfigManager::CONFIG_FILE);
+	std::ifstream c_test("./" + configFile);
 	if (!c_test.is_open()) {
-		std::ifstream config_lua_dist("./config.lua.dist");
+		std::ifstream config_lua_dist("./" + configFile + ".dist");
```
DSpeichert, line 156: This should not be variable. Since we know this file defaults to `config.lua.dist`, I think a more prudent thing is to leave it hardcoded.

src/otserv.cpp
```diff
@@ -324,3 +331,47 @@ void mainLoader(int, char*[], ServiceManager* services)
...
+
+#if defined(__amd64__) || defined(_M_X64)
+			std::cout << "x64" << std::endl;
+#elif defined(__i386__) || defined(_M_IX86) || defined(_X86_)
+			std::cout << "x86" << std::endl;
+#elif defined(__arm__)
+			std::cout << "ARM" << std::endl;
+#else
+			std::cout << "unknown" << std::endl;
+#endif
```
DSpeichert, line 360: This doesn't match what we have in otserv.cpp now. Maybe it shouldn't be copy-pasta'd, but a function somewhere.

DSpeichert (changes requested):

rookgaard:
@DSpeichert I've changed what you've requested.

src/otserv.cpp
```diff
@@ -324,3 +337,35 @@ void mainLoader(int, char*[], ServiceManager* services)
...
+		}
+
+		StringVector tmp = explodeString(arg, "=");
+
+		if (tmp[0] == "--config")
+			g_config.setString(ConfigManager::CONFIG_FILE, tmp[1]);
+		else if (tmp[0] == "--ip")
+			g_config.setString(ConfigManager::IP, tmp[1]);
+		else if (tmp[0] == "--login-port")
+			g_config.setNumber(ConfigManager::LOGIN_PORT, atoi(tmp[1].c_str()));
```
djarek, line 365:
```suggestion
			g_config.setNumber(ConfigManager::LOGIN_PORT, std::stoi(tmp[1]));
```

src/otserv.cpp
```diff
@@ -324,3 +337,35 @@ void mainLoader(int, char*[], ServiceManager* services)
...
+		else if (tmp[0] == "--game-port")
+			g_config.setNumber(ConfigManager::GAME_PORT, atoi(tmp[1].c_str()));
```
djarek, line 367:
```suggestion
			g_config.setNumber(ConfigManager::GAME_PORT, std::stoi(tmp[1]));
```

rookgaard:
Is there anything I can do more on this PR? @djarek @DSpeichert @nekiro ?

DSpeichert (approved):

PR #2757, opened 2020-01-04
sync game-time to real-time & update worldLight

Sundance:
Synced the in-game-time to real-time. Every 1h real-time = 24h cycle in-game (hh:30 is always 12:00 in-game) before this commit worldTime is set to 14:27 when tfs is started.
I personally found the code that calculates worldLight to be a pain to read (could be that I'm just more of a noob than I thought) so I rewrote it.
Added a setWorldLight(level, color) function for any ot owners that want to play around with custom lighting. The function is only usable by setting "defaultWorldLight" in config to false so the custom and default don't overwrite each other.

Let me know if anything doesn't make sense/can be improved
Cheers


src/luascript.cpp
```diff
@@ -1004,6 +1004,9 @@ void LuaScriptInterface::registerFunctions()
 	//getWorldLight()
 	lua_register(luaState, "getWorldLight", LuaScriptInterface::luaGetWorldLight);
 
+	//setWorldLight()
```
slavidodo: //setWorldLight(level, color)

config.lua.dist
```diff
@@ -76,6 +76,7 @@ classicEquipmentSlots = false
 classicAttackSpeed = false
 showScriptsLogInConsole = true
 showOnlineStatusInCharlist = false
+defaultWorldLight = true
```
EPuncker: maybe add a comment -- telling what it does since I don't know it either 😛

ranisalt:
Why not moving it straight to Lua? Just have a global interval event that calculates world time from system time and update it. Much cleaner and customizable :)

DSpeichert (approved):

MillhioreBT (approved):
Simply lovely. LGTM
tested and working...

PR #2758, opened 2020-01-06
Docker compose support

Humberd:
## Docker Compose

There is also a way of running the server + database at once using `docker-compose`.
By default it:
 * Creates a database named `forgottenserver`.
 * Executes `schema.sql` file, so that you don't have to manually import it.
 * Saves all the database files inside a `./db` local directory.

##### Preparation:

1. Update your `configu.lua.dist` with these values, so that server properly connects to the database:

```lua
-- MySQL
mysqlHost = "db"
mysqlUser = "forgottenserver"
mysqlPass = "<your_db_password>"
mysqlDatabase = "forgottenserver"
mysqlPort = 3306
mysqlSock = ""
```

2. Change your database user password by replacing
`<your_db_password>` with your password of choice.
**Remember** to also update it in `docker-compose.yml`.

3. Change your database root password by replacing
`<your_root_password>` in `docker-compose.yml`.

##### Start:

```bash
docker-compose up -d
```

##### Stop:
```bash
docker-compose down
```

##### Rebuild container after source code changes:

```bash
docker-compose up -d --build
```

#### Created services:

1. The Forgotten Server - `localhost:7171`, `localhost:7172`.
2. Database - direct connection hidden to the outside world.
3. Database explorer - `localhost:8080`. Allows viewing what's inside the database.

#### Docker-compose issues:

1. During startup the server container is created after database, however, the database might not be initially available.
This is why the server will restart a couple of times before it successfully establishes the connection.

2. If you are using Docker Toolbox for Windows and it uses VirtualBox,
 then the server and database explorer address host name will be `192.168.99.100` insead of `localhost`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ By default it:
  
 ##### Preparation:
 
-Update your `configu.lua.dist` with these values, so that server properly connects to the database:
+1. Update your `configu.lua.dist` with these values, so that server properly connects to the database:
 
 ```lua
 -- MySQL
@@ -37,6 +37,13 @@ mysqlDatabase = "forgottenserver"
 mysqlPort = 3306
 mysqlSock = ""
 ```
+
+2. Change your database user password by replacing 
+`<your_db_password>` with your password of choice. 
+**Remember** to also update it in `docker-compose.yml`.
+
+3. Change your database root password by replacing 
+`<your_root_password>` in `docker-compose.yml`.
 
 ##### Start:
 
@@ -59,7 +66,7 @@ docker-compose up -d --build
 
 1. The Forgotten Server - `localhost:7171`, `localhost:7172`.
 2. Database - direct connection hidden to the outside world.
-3. Database explorer - `localhost:8080`. It allows viewing what's inside the database.
+3. Database explorer - `localhost:8080`. Allows viewing what's inside the database.
 
 #### Docker-compose issues:
 
--- a/README.md
+++ b/README.md
@@ -55,9 +55,16 @@ docker-compose down
 docker-compose up -d --build
 ```
 
-Additionally, it also creates a database explorer available at `localhost:8080`.
+#### Created services:
 
-##### Issues on startup
+1. The Forgotten Server - `localhost:7171`, `localhost:7172`.
+2. Database - direct connection hidden to the outside world.
+3. Database explorer - `localhost:8080`. It allows viewing what's inside the database.
 
-During startup the server container is created after database, however, the database might not be initially available.
+#### Docker-compose issues:
+
+1. During startup the server container is created after database, however, the database might not be initially available.
 This is why the server will restart a couple of times before it successfully establishes the connection.
+
+2. If you are using Docker Toolbox for Windows and it uses VirtualBox,
+ then the server and database explorer address host name will be `192.168.99.100` insead of `localhost`.
--- a/README.md
+++ b/README.md
@@ -37,3 +37,8 @@ docker-compose down
 ```
 
 Additionally, it also creates a database explorer available at `localhost:8080`.
+
+##### Issues on startup
+
+During startup the server container is created after database, however, the database might not be initially available.
+This is why the server will restart a couple of times before it successfully establishes the connection.
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ If you need help, please visit the [support forum on OTLand](https://otland.net/
 
 We use the [issue tracker on GitHub](https://github.com/otland/forgottenserver/issues). Keep in mind that everyone who is watching the repository gets notified by e-mail when there is activity, so be thoughtful and avoid writing comments that aren't meaningful for an issue (e.g. "+1"). If you'd like for an issue to be fixed faster, you should either fix it yourself and submit a pull request, or place a bounty on the issue.
 
-### Docker Compose
+## Docker Compose
 
 There is also a way of running the server + database at once using `docker-compose`.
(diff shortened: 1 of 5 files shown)


EPuncker:
are those changes on config.lua.dist really necessary?

DSpeichert:
See #2687

Humberd:
@EPuncker yea, you are right. Will remove it later today

Humberd:
@DSpeichert my changes are simpler and dont require additional scripts

Humberd:
@EPuncker I reverted all the changes from `config.lua.dist`. User now has to configure database connection by himself. What is more I added detailed description what to edit in the README.

Humberd:
Bump

soul4soul:
This project don't use the repo for documentation. This project uses the wiki feature of github for documentation. Consider undoing the changes to the readme.md. Update the OP to include code blocks with the any documentation you have written. Promise you will update the wiki with the new docker documentation if these changes are merged.

I don't think anyone would object to you adding a 3rd bullet point under the `Getting Started` header which links directly to a new `docker` wiki page.

Humberd:
@soul4soul Removed documentation from README. To add new wiki page I would need some permissions.

Znote:
https://github.com/otland/forgottenserver/wiki/Docker-Compose

docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - "7172:7172"
+    depends_on:
+      - db
+    networks:
+      - db-network
+  db:
+    image: mysql
+    restart: always
+    environment:
+      MYSQL_ROOT_PASSWORD: <your_root_password>
```
DSpeichert: Try `MYSQL_RANDOM_ROOT_PASSWORD: '1'`

docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+      MYSQL_USER: forgottenserver
+      MYSQL_PASSWORD: <your_db_password>
```
DSpeichert: Instead of that, try running both containers in the same network namespace and run a script that allows no password from `localhost` for the user.

Alternatively, try sharing a unix socket and using https://mariadb.com/kb/en/authentication-plugin-unix-socket/

Cavitt: I'd recommend at least putting the variables in a separate file that can be in gitignore

```
    env_file:
      - db.env
```

README.md
```diff
@@ -7,6 +7,7 @@ The Forgotten Server is a free and open-source MMORPG server emulator written in
 
 * [Compiling](https://github.com/otland/forgottenserver/wiki/Compiling), alternatively download [AppVeyor builds for Windows](https://ci.appveyor.com/project/otland/forgottenserver)
 * [Scripting Reference](https://github.com/otland/forgottenserver/wiki/Script-Interface)
+* [Docker Compose](https://github.com/otland/forgottenserver/wiki/Docker-Compose)
```
Xaekai, line 10: I'd move this up a line, because compiling and docker are in a similar sphere related to service deployment, whereas scripting is in the sphere of dataset content creation.

docker-compose.yaml
ranisalt: I would use `MYSQL_RANDOM_ROOT_PASSWORD` instead. It does not need a root account at all, and the less privileges, the better.

docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+      # Automatically creates a database
+      MYSQL_DATABASE: forgottenserver
+    volumes:
+      # Saves all the data in ./db directory
+      - ./db:/var/lib/mysql
```
ranisalt, line 26: I would use a volume here, instead of relying on the host file system.

docker-compose.yaml
ranisalt: @DSpeichert that's not trivial with Docker, and I think he wants the password to be used in this "adminer" container below - I didn't even Google it, but it looks like phpMyAdmin.

ranisalt (changes requested):

docker-compose.yaml
DSpeichert: Sharing unix socket may not be easy due to uid mapping but sharing network namespaces is very easy with docker-compose:

```
...
services:
  db:
    ...
  pma:
    network_mode: "service:db"
    ...
...
```

docker-compose.yaml
DSpeichert: It's unusually hard to fish the files out of a docker-managed volume and the point here is for the data to persist, so I think this is the easy to understand option here.

ranisalt: Why would you want to extract raw files from a database storage? It's not like you can do anything more with them out of the container.

DSpeichert: You can back them up and run a server out of them, see mariabackup etc.

ranisalt: You can do the very same with volumes, just prepend `docker exec` to the same `mysqldump -u <user> -p <password> <db>` you would run.

gesior:
I think we should also change:
```
RUN cmake .. && make
```
to:
```
RUN cmake .. && make -j $(nproc)
```
in `Dockerfile` or make it use shared `build` folder, to do not recompile all files, after every little change in .cpp file.

DSpeichert:
Closing this as abandoned, but it could use a refresh and maybe make its way into the wiki instead.
There seems to be so many ways of running TFS, it may not be worth adding each method to the source repo.

Issue #2753, opened 2019-12-31
Removing unnecessary cast to uint32_t


Xawx:
I've been messing with my own death handling and I found very confusing part of code

https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L8125

    lua_pushnumber(L, static_cast<uint32_t>(player->getLostPercent() * 100));

I cannot understand why would percentage value like 3.99 (after being multiplied by 100)
was rounded to 3. 

Not only it can create significant differences in deathPenalty, in this example it's 25% (4 --> 3) difference, but what's even the reason?

I asked about it on otland and got told that it's due to "readability" or some shit, like why not go with double with precision like 2 numbers after dot instead of int?

### Expected behaviour

Returned value shouldnt be rounded to unit.

### Actual behaviour

Returned value should be double cuz you know, percentages aren't only ints.

Since I've no rights to push & create new branch, then I'll put here my suggestion

Change this line:

    lua_pushnumber(L, static_cast<uint32_t>(player->getLostPercent() * 100));

to:

    lua_pushnumber(L, player->getLostPercent() * 100);

Sundance:
> Since I've no rights to push & create new branch, then I'll put here my suggestion

For small changes like this you can go to the file you want to edit, click the pen icon at your top right, make your edit and at the bottom click "propose file change".

Otherwise the standard way to do it is:
Fork -> create a branch on your fork -> checkout to your new branch -> make your edit(s) -> commit -> pull request

PR #2759, opened 2020-01-07
 Removing unnecessary cast to uint32_t

Xawx:
https://github.com/otland/forgottenserver/issues/2753


DSpeichert:
`lua_pushnumber` by default takes _double_

https://pgl.yoyo.org/luai/i/lua_pushnumber
https://pgl.yoyo.org/luai/i/lua_Number

PR #2760, opened 2020-01-10
mappedPlayerGuids to improve player search by GUID

danyelvarejao1:


src/game.cpp
```diff
@@ -473,12 +473,11 @@ Player* Game::getPlayerByGUID(const uint32_t& guid)
 		return nullptr;
 	}
 
-	for (const auto& it : players) {
-		if (guid == it.second->getGUID()) {
-			return it.second;
-		}
+	auto it = mappedPlayerGuids.find(guid);
```
Humberd: What if `Game::getPlayerByNameWildcard` had not been called before. It would not find a player, wouldn't it?

kygov: It's not related, in 'Game::addPlayer(Player* player)' is inserted player to 'mappedPlayerGuids' map, exactly same way as 'getPlayerByName' or 'getCreatureByName' etc.

danyelvarejao1: Yes, I have no problem with that, as mappedPlayersName is also added.

Lyuzera: this is an extra cache for improve seach by GUID as the title says. Now `Player(guid)` is fast as `Player(name)` or `Player(cid)`.

Znote:
This is great, as player GUID is frequently used to index player data in locally, and sometimes complex structures of data in various script systems. The player userdata is also frequently reinitialized when passed through addEvent calls for delayed execution.

nekiro:
Did you do any benchmarks if that changes anything?

danyelvarejao1:
> Did you do any benchmarks if that changes anything?
Sure this is much faster.

nekiro:
> @nekiro, Sure this is much faster.

I actually asked about some type of proof, data or something.

kygov:
@nekiro 
I've made some tests, 'find' based function is almost **33x faster** than 'loop' on my pc.
Code I've used: https://pastebin.com/hUaMqRMq

> 1500 players size, 1500 mapped size
For Took 12446 ms
Find Took 366 ms

Issue #2764, opened 2020-01-21
Lua Monster (Revscripts)


karimadnan:
https://gyazo.com/6ead0e3927aa058732d7509d206b70b8

1- Monster don't spawn with skull.
2- Monster is hostile even tho it's false in flags.
3- How can i add custom mTypes to the config? ex. i'm using a monster level system.

EPuncker:
1 and 2 I can confirm, for 3 you need to ask to the person who made the monster level system, not here.

![2020-01-21 14_32_27-Window](https://user-images.githubusercontent.com/4684880/72828168-e383a480-3c5a-11ea-807d-176329af955b.png)
![2020-01-21 14_32_15-Window](https://user-images.githubusercontent.com/4684880/72828169-e41c3b00-3c5a-11ea-827a-f6e36411476e.png)

tagging @EvilHero90 just in case 😝

Znote:
I can faintly remember that skullTicks? skullDuration? inCombatDuration? had to be declared before giving someone a skull, else they would get a skull which would immediately dissapear. Making it appear as if they never got a skull.

karimadnan:
> I can faintly remember that skullTicks? skullDuration? inCombatDuration? had to be declared before giving someone a skull, else they would get a skull which would immediately dissapear. Making it appear as if they never got a skull.

not true i used creature:setSkull(3) it works fine.

EvilHero90:
Yea skull hasn't been added, will do a pr soon and about hostile I need to look a bit deeper why it's not working as intended.

karimadnan:
@EvilHero90  can you point me where can i add custom flags for lua monster?

EvilHero90:
This is weird, the hostile code just looks fine, did you actually try to leave it out completly to have it false as default? @karimadnan

EPuncker:
> This is weird, the hostile code just looks fine, did you actually try to leave it out completly to have it false as default? @karimadnan

I just tested letting it out completely and it still attacks me

ninjalulz:
Please correct me if I'm wrong (I haven't tested this "revscripts") but the script interface is being loaded before monsters, so couldn't it simple be a case of the monster actually being overwritten by XML settings (assuming it's still registered in there)? That would partially explain why it doesn't update `isHostile` properly.

EvilHero90:
> 
> 
[...]

PR #2766, opened 2020-01-24
MonsterType skull method & flags fix

EvilHero90:
MonsterType now supports ex: get :skull() and set :skull("red") : skull(SKULL_RED)
It supports strings and constants.
This also solves that the MonsterType flags where not correctly set if you set them to "false"
solves #2764

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -75,28 +75,28 @@ registerMonsterType.corpse = function(mtype, mask)
 end
 registerMonsterType.flags = function(mtype, mask)
 	if mask.flags then
-		if mask.flags.attackable then
+		if mask.flags.attackable ~= nil then
 			mtype:isAttackable(mask.flags.attackable)
 		end
-		if mask.flags.healthHidden then
+		if mask.flags.healthHidden ~= nil then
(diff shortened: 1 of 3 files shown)


Znote:
Hmm, wouldn't it be better for it to support our skull constants?

https://github.com/otland/forgottenserver/blob/380903887a92d0abcf00158f37f59094d6340bfa/src/const.h#L402-L410

https://github.com/otland/forgottenserver/blob/380903887a92d0abcf00158f37f59094d6340bfa/src/luascript.cpp#L1619-L1625

EvilHero90:
> 
> 
> Hmm, wouldn't it be better for it to support our skull constants?
> 
> https://github.com/otland/forgottenserver/blob/380903887a92d0abcf00158f37f59094d6340bfa/src/const.h#L402-L410
> 
> https://github.com/otland/forgottenserver/blob/380903887a92d0abcf00158f37f59094d6340bfa/src/luascript.cpp#L1619-L1625

some people are used to strings in this case but yea you where right I've added the possibility to either add through str or constant now.

Znote:
`monster.skull = "white"` works. 

image

`monster.skull = SKULL_WHITE` does not work. 

image

EvilHero90:
I've updated the entrance title and the initial post, this solves #2764 now entirely.

PR #2767, opened 2020-01-29
Updated 6.x monsters

EPuncker:
all according to https://tibia.fandom.com/wiki/Updates/

feel free to test and review


Wirox:
Scorpion with poison value 17? It doesn't sound right.

EPuncker:
> Scorpion with poison value 17? It doesn't sound right.

I need someone to explain this to me:
`https://github.com/otland/forgottenserver/blob/master/src/monsters.cpp#L240`

EPuncker:
> Scorpion with poison value 17? It doesn't sound right.

it is actually right tho, just tested

Wirox:
> it is actually right tho, just tested
No, that's incorrect. In real Tibia scorpion's poison starts hitting for 17, which is exactly how the old poison value (340) works. With poison value 17 it would start from 1 damage, I believe.
This also applies to other poison-inflicting creatures you've edited.

Wirox:
It's also worth noting that your defense values are incorrect (TibiaWiki doesn't have any information about defense of creatures).

EPuncker:
> It's also worth noting that your defense values are incorrect (TibiaWiki doesn't have any information about defense of creatures).

it was always been like that: armor = defense what is your suggestion to change it to then?

Wirox:
> it was always been like that: armor = defense what is your suggestion to change it to then?

I'd probably not touch it at all (there is not much point in changing values from incorrect to incorrect).

But if you want to do this properly, I'd probably use fist_fighting and defense stats from leaked 7.72 monster files, and based on some reliable defense formula (such as that from tibia-stats) calculate average damage reduction from defense for each monster, and use TFS defense values that have the closest average damage reduction.

raymondtfr:
There are equations you could use to get the defense value but that would depend entirely on the melee formula (real tibia's one) and actual tests (1000+) for each creature, so...

It also doesn't make that much of a difference since our random number generator is not even close as the one used by Tibia, since they seem to use a piecewise distribution somewhat or something else they coded from scratch (they could also use different distr. for let's say, damage reduction by defense, armor, distance damage, wand damage, rune damage, etc.).

That's why it doesn't feel like we are hunting "Tibia" creatures, but something else. And that lack of feeling is also due to the delay a character, monsters waits till it attacks again which is not similar to Tibia even though it seems to be 2 seconds, it does not feel the same as on Tibia.

All these things makes hunting creatures on OTS not that exciting, if you want to have the same feel as on the original game.

BahamutxD:
Even if the changes to armor/defense are not perfect they are much better than what we have right now - this is specially noticiable on weaker monsters where low level/low skilled characters have no chance killing  and Orc, Wasp or even a Cave Rat which is supposed to be killed by a level 1 or 2 at worst due to their high defense/armor.

Wirox:
> Even if the changes to armor/defense are not perfect they are much better than what we have right now - this is specially noticiable on weaker monsters where low level/low skilled characters have no chance killing and Orc, Wasp or even a Cave Rat which is supposed to be killed by a level 1 or 2 at worst due to their high defense/armor.

Armor values are correct because they are taken from real Tibia's Bestiary.
Defense values, on the other hand, while for weak creatures are indeed closer to the actual ones, for stronger monsters it's the opposite. So I wouldn't say these defense values are better than what we have at the moment.

Leo32onGIT:
If we're trying to reach parity with real tibia, then yes we should use leaked 772 and do math.
If we aren't, then its a question of are these changes better or worse then existing monsters?
https://github.com/otland/forgottenserver/pull/2767#issuecomment-582093698 - I feel this.

This is going to change the feel of monsters (loot/survive-ability) but we shouldn't be afraid to go ahead and push it through if the intent of the PR is to align monsters more accurately to vanilla monsters.

EPuncker:
updated

Leo32onGIT:
LGTM.
Needs more eyes due to the amount of monsters changed.

data/monster/Bio-Elementals/squidgy_slime.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
EvilHero90: encoding="ISO-8859-1"

EvilHero90: It's not that important but we should stick to one standard

EPuncker: yes I talked about it with @DSpeichert but we need to get to a standard one first

Znote:
With #2851 enabled:
```
>> Loading monsters
[Error - Monsters::loadMonster] Failed to load data/monster/Misc/Misc Mammals/badger.xml: File was not found
[Error - Monsters::loadMonster] Failed to load data/monster/Misc/Misc Mammals/ghoulish_hyaena.xml: File was not found
[Error - Monsters::loadMonster] Failed to load data/monster/Misc/Misc Mammals/hyaena.xml: File was not found
[Error - Monsters::loadMonster] Failed to load data/monster/Misc/Misc Mammals/skunk.xml: File was not found
[Error - Monsters::deserializeSpell] - dipthrah - Unknown spell name: dipthrah skill reducer
[Warning - Monsters::loadMonster] Cant load spell. data/monster/Pharaohs/dipthrah.xml
[Warning - Monsters::loadMonster] Unknown loot item "jalapeno pepper".
[Warning - Monsters::loadMonster] Cant load loot. data/monster/Sorcerers/fury.xml

>> Loading lua monsters
> #example.lua [disabled]
>> Loading outfits
>> Checking world type... PVP
>> Loading map
> Map size: 1000x1000.
> Map loading time: 1.098 seconds.
> Loaded house items in: 0.001 s
>> Initializing gamestate
>> Loaded all modules, server starting up...
>> Forgotten Server Online!

Znotegm has logged in.
Znotegm executed talkaction: !hello

=============================
=== Loaded monster types: ===
=============================
672
=============================
```

WibbenZ:
Wtf is up with the errors :D 

```
[Error - Monsters::deserializeSpell] - dipthrah - Unknown spell name: dipthrah skill reducer
[Warning - Monsters::loadMonster] Cant load spell. data/monster/Pharaohs/dipthrah.xml
[Warning - Monsters::loadMonster] Unknown loot item "jalapeno pepper".
[Warning - Monsters::loadMonster] Cant load loot. data/monster/Sorcerers/fury.xml
```

I didn't see these errors today when testing codes.
Dipthrah errors were added in #2777 
Fury errors were added in #2761

EPuncker:
merged master again, can you guys @Znote and @WibbenZ check it again?

Znote:
Better, 2 errors now:
```
>> Loading monsters
[Error - Monsters::deserializeSpell] - dipthrah - Unknown spell name: dipthrah skill reducer
[Warning - Monsters::loadMonster] Cant load spell. data/monster/Pharaohs/dipthrah.xml
>> Loading lua monsters
> #example.lua [disabled]
>> Loading outfits
>> Checking world type... PVP
>> Loading map
> Map size: 1000x1000.
> Map loading time: 0.862 seconds.
> Loaded house items in: 0 s
>> Initializing gamestate
>> Loaded all modules, server starting up...
>> Forgotten Server Online!

Znotegm has logged in.
Znotegm executed talkaction: !hello

=============================
=== Loaded monster types: ===
=============================
672
=============================
```

EPuncker:
I don't really understand why is it erroring, triple checked everything and it is correct

Znote:
Ohh, right! I might have forgotten to patch the spells folder with your PR, sorry. :p

DSpeichert:
Needs rebase.

EPuncker:
> Needs rebase.

done (I guess)

EPuncker:
once this is merged I'll finally be able to work on a new monster folder structure PR (1 folder for bosses, everything else out alphabetically)

PR #2768, opened 2020-01-29
Updated invalid looktype types list

EPuncker:
the list itself was updated to all outfits as of this PR date, and the talkaction itself was updated to check the from 903 instead of 700, that is the latest (?) 10.98 outfit id


data/talkactions/scripts/looktype.lua
```diff
@@ -14,7 +29,7 @@ function onSay(player, words, param)
 	end
 
 	local lookType = tonumber(param)
-	if lookType >= 0 and lookType < 700 and not table.contains(invalidTypes, lookType) then
+	if lookType >= 0 and lookType < 903 and not table.contains(invalidTypes, lookType) then
```
DSpeichert: If all lookTypes > 903 are invalid, then why are you adding a lot of >700 IDs to the list above?

Znote: He is back from the future. x)

PR #2779, opened 2020-02-05
Added wall mirror script

EPuncker:


WibbenZ:
Thanks! :)

data/actions/scripts/other/wallmirror.lua
```diff
@@ -0,0 +1,19 @@
...
+	"You rarely looked better.",
+	"Well, you can't look good every day.",
+	"You should think about a makeover.",
+	"Is that the indication of a potbelly looming under your clothes?",
+	"You look irresistible.",
+	"You look tired.",
+	"You look awesome!",
+	"You nearly don't recognize yourself.",
+	"You look fabulous.",
+	"Don't be so vain about your appearance.",
```
TheSumm: You can use wall mirrors once per day and a random message will appear, afterwards this message will show (`"Don't be so vain about your appearance."`)

EPuncker: yeah I just didn't wanted to add a new variable once again

Issue #2778, opened 2020-02-05
Defense attribute in monsters 


EPuncker:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

Where is it calculated when formulating the damage received? Is it used anywhere? I tried searching for it in source but no success, if any dev could enlighten this to me I'll be grateful in removing it from monster files if it is the case

Wirox:
https://github.com/otland/forgottenserver/blob/293099a8cc518eefe66276a90544ad8d45505b87/src/creature.cpp#L800-L808

EPuncker:
> https://github.com/otland/forgottenserver/blob/293099a8cc518eefe66276a90544ad8d45505b87/src/creature.cpp#L800-L808

isn't monster block hit code this one? https://github.com/otland/forgottenserver/blob/7eb5d3a975b2e76c61fe5add259bd2c3c8351559/src/monster.cpp#L585

MillhioreBT:
> > https://github.com/otland/forgottenserver/blob/293099a8cc518eefe66276a90544ad8d45505b87/src/creature.cpp#L800-L808
> 
> isn't monster block hit code this one?
> 
> https://github.com/otland/forgottenserver/blob/7eb5d3a975b2e76c61fe5add259bd2c3c8351559/src/monster.cpp#L585

It is also filtered through creature::blockhit
![image](https://user-images.githubusercontent.com/28090948/74048371-74bb8080-49a8-11ea-9aee-f165d7738749.png)

EPuncker:
oh I see, thanks

PR #2782, opened 2020-02-07
Updated 7.x monsters [part 2]

EPuncker:
- Updated a few missed monsters that the implementation date was unknown
- Created two new classes: bats and vampires and moved the files accordingly


soul4soul:
As with the previous monsters packs you are working on I too wonder about updating the defense stat from a wrong value to another wrong value. I agree with @Wirox comments from your 6.x PR.

If you going to reorganize monsters I suggest following the cipbia [bestiary classes](https://tibia.fandom.com/wiki/Bestiary/Classes). Boss, raid, and quest monsters would be an exception since cipbia doesn't put them in bestiary classes. For such creatures we would need to decide were to place those creatures.

+1 for the putting in the effort to improve the default monster pack provided with TFS.

EPuncker:
> As with the previous monsters packs you are working on I too wonder about updating the defense stat from a wrong value to another wrong value. I agree with @Wirox comments from your 6.x PR.
> 
> If you going to reorganize monsters I suggest following the cipbia [bestiary classes](https://tibia.fandom.com/wiki/Bestiary/Classes). Boss, raid, and quest monsters would be an exception since cipbia doesn't put them in bestiary classes. For such creatures we would need to decide where to place those creatures.
> 
> +1 for the putting in the effort to improve the default monster pack provided with TFS.

bestiary classes is not as useful as it sounds, look at the humanoid/human/magical class, there are too many files in one place, for example there is no "Orcs", "Dwarves" and "Elves" anymore, I think the way that we currently "handle" them is better for organization. As for the defense thing, I don't really see this as changing a wrong value to another, but rather changing it to the "pattern" we always used that the armor value was the same as defense and since now we know the armor value I just updated the defense value to mirror it, but it is up to devs to decide the way it should be done, I even opened #2778 to gather some info

soul4soul:
Moving monsters and making more folders in this PR is to match your preference. Maybe I prefer less folders. Some monster files are even hard to place as they fit into multiple folders. Organization of monsters is a subjective topic with lots of opinions. The same is true for many of the features in TFS and tibia. TFS can be customized to any behavior one prefers but commonly defaults to matching tibia. I am suggesting to follow the tibia classification of monsters by doing so it removes some subjectivity. By following cipbia bestiary It would make it easier to determine where to place monsters in the future.

I don't mind how monsters are organized. There are so many of monsters most times I end up searching for them by name in an editor or explorer. Lets keep from putting demons in the insects folder and I'll be happy.  Since your PR happened to include a reorganization it seemed like a good opportunity to mention a organizational system that can be defined. It's not even possible to define the current organization as it changes with each PR.

Leo32onGIT:
I always hated the monster folder structure.

Could always find them easier when they were flat with the monsters.xml main file, just sort by a-z and everything's in reach.

Coming back and now having to do a recursive subfolder search is annoying, monster categories aren't intuitive enough for them to stick in memory.

data/monster/Dworcs/dworc_voodoomaster.xml
```diff
@@ -8,7 +8,7 @@
 		<flag attackable="1" />
 		<flag hostile="1" />
 		<flag illusionable="1" />
-		<flag convinceable="1" />
+		<flag convinceable="0" />
```
Leo32onGIT: F

It's been a good run.

EPuncker:
> I always hated the monster folder structure.
> 
> Could always find them easier when they were flat with the monsters.xml main file, just sort by a-z and everything's in reach.
> 
> Coming back and now having to do a recursive subfolder search is annoying, monster categories aren't intuitive enough for them to stick in memory.
> 
> What did you use to base your value changes on these monsters btw?
> Appreciate the effort 👍

2020-03-11 20_21_59-Frazzlemaw _ TibiaWiki _ Fandom

it is true that we have way too many subfolders, we should discuss and get to an agreement to a better approach, but until them I'll keep this on halt

Wirox:
This pull request shouldn't have been merged because it contains significant errors and clearly hasn't been tested thoroughly. It also applies to all other monster-changing PRs made by EPuncker.

Even if we ignore the aforementioned defense values, there are a few other important issues:

1. Some monsters (such as hydra) have countmax attribute for gold coins above 100, but it doesn't work like some of you may think it does. It only increases the odds of the creature dropping exactly 100 items, but it can never make it drop more than 100.

2. Apparently the countmax values were modified to match the maximum drops from TibiaWiki's loot statistics, but this isn't a correct approach. They should be made to match, together with the chance attribute, the **average** drops.
Let's take a look at the orc rider. In real Tibia it drops 1–81 gold coins, but only 2.7 on average. With the proposed attributes (countmax=81 and chance=48000) it'll drop 19.68 gold coins on average, which is more than 7(!) times of the correct average. If we don't know the exact distribution, we should instead choose such chance/countmax that will result in an average very close to the one from real Tibia, for example countmax=10 and chance=48000.

3. Some creatures have incorrect loot data because they were blindly copied from TibiaWiki's loot statistics. For example Kongra had its gold drop changed to 1–61 because such value appears in loot statistics, but it's only because of an erroneous edit: https://tibia.fandom.com/index.php?title=Loot_Statistics%3AKongra&diff=750937&oldid=743243

4. There most likely are more issues than the ones above. A pull request cannot be merged just because it compiles and looks okay after a quick glance. All changes proposed by this pull request should be checked one by one, and then, if they all turn out to be correct, the PR should be merged.

DSpeichert:
@Wirox This PR was up for review for over a month, you're finding issues 2 days after it was merged. At this point, all you can do is open a PR with fixes.

EPuncker:
> @Wirox This PR was up for review for over a month, you're finding issues 2 days after it was merged. At this point, all you can do is open a PR with fixes.

his job is to nitpick all my PR instead of actually pointing the errors, as you can see [here](https://otland.net/threads/call-to-community-at-tfs-github.268947)

EPuncker:
https://github.com/otland/forgottenserver/pull/2860

PR #2787, opened 2020-02-10
Added costume bags script

EPuncker:
introduced in 8.1 update

https://tibia.fandom.com/wiki/Costume_Bag


EPuncker:
> We use tabs (4 spaces) as indentation method, so please change that.

I just checked every other script and we use tab in all of them, unless I missed the point in your comment (which I think so), and the script from this PR is using tab as well, so I really don't know

nekiro:
Most likely code in browser showed it different, looking from phone it is fun.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2789, opened 2020-02-10
Added Very Noble-Looking Watch

EPuncker:
Added Very Noble-Looking Watch to watch.lua


nekiro:
Hey @EPuncker just a small question, can you tell me why you spam pull requests instead of doing one big with all these changes?

PR #2791, opened 2020-02-16
Save storages in lib

marcomoa:
In my understanding I think it would be interesting for the repository to start using a lib to store the storages, because in a way it makes reading easier and there is also a great improvement in the organization of the files.


EPuncker:
@TheSumm will like this 😛

data/lib/core/storages.lua
```diff
@@ -0,0 +1,12 @@
+Storage = {
+    annihilator = {
+        Reward = 30015,
+    }
```
EPuncker: isn't it missing a comma (,) here?

Sickness-chan: yeah, it is

data/creaturescripts/scripts/login.lua
```diff
@@ -21,7 +21,7 @@ function onLogin(player)
 	if player:isPremium() then
 		local value = player:getStorageValue(STORAGEVALUE_PROMOTION)
```
kornholi: You missed quite a few keys

EvilHero90:
Ok well, the idea in general is good but we should talk about consistency here.
Either you go first letter always upper case or not but not mixed, this defeats the purpose of easily remembering such things.
Also in my opinion ```Storage``` should be renamed to ```StorageKey(s)``` as it represents just a list of key's and nothing more.

data/lib/core/storages.lua
DSpeichert: is there anything else that would go into `annihilator`? Why so much nesting?

nekiro:
Yeah, that's how ORTS was handling it since ages, very good pull request. Approved.
Just fix the small/big letters like @EvilHero90 mentioned.

data/lib/core/storages.lua
marcomoa: in my opinion because it is a quest it should be separated from the others, in case it is necessary to add more keys in the future.

data/lib/core/storages.lua
```diff
@@ -0,0 +1,12 @@
+StorageKeys = {
+    annihilator = {
+        reward = 30015,
```
DSpeichert: `annihilatorQuestReward` will be just fine, no nesting necessary

data/lib/core/storages.lua
```diff
@@ -0,0 +1,9 @@
+StorageKeys = {
```
DSpeichert: This could be called PlayerStorageKeys to make it more obvious

PR #2797, opened 2020-02-22
Added christmas bundle

EPuncker:
They were added on 7.9 update

script fixed by @joseluis2g 
tested ingame and working as intended


data/scripts/actions/christmas_bundle.lua
```diff
@@ -0,0 +1,43 @@
...
+		[6507] = { -- red christmas bundle
+			{6569, 15}, {2687, 20}, {2688, 10}, {2675, 10}, {2674, 5}, 6501, 6502, 6490, 6504, 6388
+		},
+		[6508] = { -- blue christmas bundle
+			{6569, 15}, {2687, 20}, {2688, 10}, {2675, 10}, {2674, 5}, 6501, 6502, 6490, 6505, 6388
+		},
+		[6509] = { -- green christmas bundle
+			{6569, 15}, {2687, 20}, {2688, 10}, {2675, 10}, {2674, 5}, 6501, 6502, 6490, 6503, 6388
+		}
+	}
```
gpedro: this table looks repeatable. I think this should be flat cuz you don't need key once you have ids registration on end of the script and the values are the same for all keys.
what are you think about that?

EPuncker: yes but the code will look even messier because of this I guess, [look](https://otland.net/threads/tfs-1-3-how-to-make-script-select-more-than-one-reward.268511/#post-2593711)

gpedro:
```lua
local rewards = {
  {6569, 15}, {2687, 20}, {2688, 10}, {2675, 10}, {2674, 5}, 6501, 6502, 6490, 6504, 6388
}
```
messier? lol... well, i just saying to make the table simple to this cuz all rewards is the same

ranisalt: The rewards are not all the same. Mind 6503 through 6505.

gpedro: that's right, sorry :P

data/scripts/actions/christmas_bundle.lua
```diff
@@ -0,0 +1,43 @@
...
+		local count = 1
+		local rand = math.random(#targetItem)
+		local gift = targetItem[rand]
+		if type(gift) == "table" then
+			count = gift[2]
+			gift = gift[1]
+		end
+		rewards[#rewards + 1] = {gift, count}
+		table.remove(targetItem, rand)
+	until #rewards == 7
```
ranisalt: I personally prefer `while #rewards < 7` over `repeat...until` in this case.

data/scripts/actions/christmas_bundle.lua
```diff
@@ -0,0 +1,43 @@
...
+
+	for i = 1, #rewards do
+		player:addItem(rewards[i][1], rewards[i][2])
```
ranisalt: Try `unpack(rewards[i])` instead.

data/scripts/actions/christmas_bundle.lua
```diff
@@ -0,0 +1,43 @@
...
+	end
+
+	local rewards = {}
+	repeat
+		local count = 1
+		local rand = math.random(#targetItem)
+		local gift = targetItem[rand]
+		if type(gift) == "table" then
+			count = gift[2]
+			gift = gift[1]
```
ranisalt: Try `gift, count = unpack(gift)` (not sure it works but probably).

EPuncker:
don't merge this yet, need to figure this out first #2803

EPuncker:
okay "fixed" it, since in this script the effect is displayed only on player, no matter where the item is, it can be merged now

data/scripts/actions/christmas_bundle.lua
```diff
@@ -0,0 +1,42 @@
+local christmasBundle = Action()
+
+function christmasBundle.onUse(player, item, fromPosition, target, toPosition, isHotkey)
+	local presents = {
+		[6507] = { -- red christmas bundle
```
DSpeichert: a comment should include the purpose of the mapping, eg.:
-- (christmas present item id) => ((reward item id, count), ...)

data/scripts/actions/christmas_bundle.lua
```diff
@@ -0,0 +1,42 @@
+local christmasBundle = Action()
+
+function christmasBundle.onUse(player, item, fromPosition, target, toPosition, isHotkey)
+	local presents = { -- [christmas bundle item id] = {{reward item id, count}, ...}
+		[6507] = { -- red bundle
+			{6569, 15}, {2687, 20}, {2688, 10}, {2675, 10}, {2674, 5}, 6501, 6502, 6490, 6504, 6388
```
DSpeichert: It would be just as beneficial do the reader to annotate every item here. Perhaps like this?

```lua
{6569, 15}, -- blue fork
{2687, 20}, -- tasty armor
6504, -- a lamp
```

EPuncker: yeah indeed, I was thinking the same

PR #2798, opened 2020-02-24
Game.getMonsterTypes()

EvilHero90:
This will add the function `Game.getMonsterTypes()`
It will return as the key the monster name and as the value the metatable.

This helps a lot if you want to register a certain event to all of your existing MonsterTypes.

Beware! 
This does not trigger unloaded MonsterTypes which are created through xml and not once spawned on map load, it'll only fetch MonsterTypes which are created through revscriptsys (monster/lua) or xml monsters which are load and spawned on startup.

### DIFF
--- a/src/monsters.h
+++ b/src/monsters.h
@@ -241,6 +241,7 @@ class Monsters
 		bool deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std::string& description = "");
 
 		std::unique_ptr<LuaScriptInterface> scriptInterface;
+		std::map<std::string, MonsterType> monsters;
 
 	private:
 		ConditionDamage* getDamageCondition(ConditionType_t conditionType,
@@ -252,7 +253,6 @@ class Monsters
 		void loadLootContainer(const pugi::xml_node& node, LootBlock&);
 		bool loadLootItem(const pugi::xml_node& node, LootBlock&);
 
-		std::map<std::string, MonsterType> monsters;
 		std::map<std::string, std::string> unloadedMonsters;
 
 		bool loaded = false;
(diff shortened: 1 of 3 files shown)


WibbenZ:
It should either return all monster types or change the function name so it's clear that it will only return rev monster types.

EvilHero90:
> 
> 
> It should either return all monster types or change the function name so it's clear that it will only return rev monster types.

Well it basicly returns all monster types (from lua and xml), the problem is the following:
if you have let's say the monster "test.xml" added to your server but you don't let it spawn somewhere on your map for once on server load, then this monster is marked as "unused" and is in a different map::list which doesn't even hold the data of the monster type itself as it has not been created yet (because it's not needed) this prevents the server to create monster types which the server primarly doesn't use.
Monster types through revscriptsys are immediatly load and ready to go that's why it doesn't apply to them.
We could either get rid of the entire "unused part" or we keep it as is.

EPuncker:
the unused part is okay IMO, since there are creatures that we only spawn during raids/events

EvilHero90:
LGTM

src/monsters.h
```diff
@@ -241,6 +241,7 @@ class Monsters
 		bool deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std::string& description = "");
 
 		std::unique_ptr<LuaScriptInterface> scriptInterface;
+		std::map<std::string, MonsterType> monsters;
```
nekiro: instead of moving that from being private, maybe we should add public method to acquire that?

EvilHero90: I didn't really feel the necessity to do so, as we just use this one time.
If we had several cases where we would need the public class, then yes it should be done but other than adding one more line it wont really gives us any kind of benefit (atleast that's what I think)

Znote:
https://github.com/otland/forgottenserver/wiki/Functions#Game.getMonsterTypes

soul4soul:
@Znote 
Nice use of wsl 👌

Znote:
yep, got the ultimate development & git reviewing environment set up now, where I use the best of both Windows and Linux. I have also made a script that lets me checkout, compile and run other forks with a simple shell command. :)

PR #2800, opened 2020-02-27
Added April fools items and a few more

EPuncker:
- Added April fools items implemented on 8.1 update
- Added mad sheep to ungulates
- Added elvenhair rope and light shovel to their respective rope and shovel scripts
- Organized a few files in script folder
- Added an old clock
- Added dracoyle statue to enchanting script
- Added missing grass and wild growth to machete script
- Added heavy machete to machete.lua
- Fixed wild growth wrong distance effect
- Added clay lump action script

everything tested in game so far


data/actions/scripts/tools/machete.lua
```diff
@@ -1,8 +1,18 @@
+local jungleGrass = { 2782, 3985, 19433 }
+local wildGrowth = { 1499, 11099 }
+
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
-	if target.itemid == 2782 then
-		target:transform(2781)
+	local targetId = target.itemid
+	if table.contains(jungleGrass, targetId) then
+		target:transform(targetId == 19433 and 19431 or targetId - 1)
```
DSpeichert: Are you sure about that?

data/monster/Ungulates/mad_sheep.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
DSpeichert: Wow, that should be UTF-8 header.

data/actions/scripts/tools/machete.lua
EPuncker: yes
asdasd

DSpeichert: You are right, this is correct but not very readable.
Can you simply use a table of mapping what item is transformed to what?
Would make it more noob-friendly.

EPuncker: okay, see it now

PR #2806, opened 2020-03-04
Updated flower pot entries in items.xml

EPuncker:
all credit to AGS at otland


data/items/items.xml
```diff
@@ -13347,15 +13347,21 @@
 	<item id="7670" article="a" name="flower pot">
 		<attribute key="description" value="The plant in there sprouted, but only gardeners might recognise what it will develop into." />
 		<attribute key="weight" value="150" />
+		<attribute key="decayTo" value="7678"/>
```
WibbenZ: Fixed it in your last PR, but the code style is " /> (note the space)

EPuncker: something always goes unnoticed in my eyes

PR #2808, opened 2020-03-04
Distance effect fix for weapons with last charge

### LINKED ISSUE #2799: Whirlwind Throw + Ice Rapier is bugged

**1.** Use the command "/i Ice Rapier".
**2.** Use the spell "exori hur".
**3.** No CONST_ANI will be played.

### Expected behaviour
**CONST_ANI_SWORD** should be played when using it with Ice Raper due to it being a Sword Type.

### Actual behaviour
It displays **CONST_ANI_NONE**, which is the default in the sources when no weapon type can be found.

### Extra information
Most likely due to the fact that Ice Rapier only has 1 charge of use.
So when you cast Exori Hur the weapon disappears before the spell has time to register what weapon type it was used with.

nekiro:
Fix #2799


Znote:
For some reason, the second ice rapier I used was able to be used twice? (1 melee attack + 1 exori hur)










Entire gif replay: (bit to large for github embedded attachment)
http://znote.eu/versions/exorihur.gif

nekiro:
@Znote looks like ice rapier wasnt consumed when you used it as melee, can you check if charges are removed when you hit the dragon with ice rapier but as melee not with spell?

Znote:
Turns out when I first summoned ice rapier, I summoned one with two charges. (In hopes of getting 2 ice rapiers). :P

PR #2809, opened 2020-03-04
Cautious monsters fix

### LINKED ISSUE #2793: "cautious monsters"
Hello Everyone.
I would like to report an error that is happening in the "pathing system".
For example; A demon should have this flag:
flag canwalkonenergy="1"
flag canwalkonpoison="1"
But the problem is : if there is other way to the monster walks to go to the player position, the monster will try to get the "best way" to not receive the damage from the energy wall.
But the demon should ignore the "caution", and just pass trough the energy wall.

Will post a video showing it better ..
https://www.youtube.com/watch?v=OKohlC-Bl64

Note: that's my first topic, so take easy with me :( also my english is not that good.. Im learning ;- )

nekiro:
Fix #2793


Lycanzito:
Good bless you, your soul, all your family, your childrens, your dog, your cat, your parrot.

nekiro:
@Znote it should move after you hit him.

PR #2810, opened 2020-03-05
Equip messages fix

### LINKED ISSUE #2773: Level check for items returning the wrong message

1. Try to equip any item that have level requirement while not meeting it

### Expected behaviour
- it should return RETURNVALUE_NOTENOUGHLEVEL
- it should not de-equip the equipped item if the dressing requirements are not met

### Actual behaviour
- it is returning RETURNVALUE_CANNOTBEDRESSED
- it is de-equipping the item even if you can't equip the new one

nekiro:
Fix #2773


src/movement.cpp
```diff
@@ -659,29 +659,33 @@ uint32_t MoveEvent::RemoveItemField(Item*, Item*, const Position&)
...
-		return 1;
+		return RETURNVALUE_NOERROR;
 	}
 
 	if (!player->hasFlag(PlayerFlag_IgnoreWeaponCheck) && moveEvent->getWieldInfo() != 0) {
+		if (player->getLevel() < moveEvent->getReqLevel()) {
+			return RETURNVALUE_NOTENOUGHLEVEL;
+		}
+
 		if (player->getLevel() < moveEvent->getReqLevel() || player->getMagicLevel() < moveEvent->getReqMagLv()) {
```
maattch: duplicate level check?

nekiro:
@Znote are you 100% sure about that? I really tested that and it wasnt deequip my item.

Znote:
leveleq

image

nekiro:
@Znote 
https://nekiro.s-ul.eu/xGISvQIJ.gif

danyelvarejao1:
I have a bug, add an item to the player when it doesn't have a slot to receive it, using player:addItem()
The item will not be sent to the tile.
There will be a userdata for that item, thus leaking "memory".

nekiro:
@danyelvarejao1 you are wrong and how is this even related? If item fails to add to player then it gets deleted. This pr does not change any of that functionality either.
https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L9011

Znote:
Still getting weird results. Looks fine on my sorcerer, but not on my knights. 
leveleq2

PR #2813, opened 2020-03-05
Better premium time handling

### LINKED ISSUE #2436: [Suggestion] premend instead of premdays and lastday

nekiro:
Getting rid of premium days and lastDay in favor of premiumEndsAt.
Fix #2436


Leo32onGIT:
> Correct me if I'm wrong, but I didn't found in your code any place that updates old accounts.

Looks like you're right here.
Need a couple extra lines on migrations, to convert existing lastdays.

nekiro:
@slawkens need more feedback about that, if we really care about that.

Leo32onGIT:
> @slawkens need more feedback about that, if we really care about that.

Should always be backwards compatible? isn't that the point of migrations?

nekiro:
@Leo32onGIT it is backwards compatible, this point slawkens made only apply to „running servers”, but I can add that conversion...

Znote:
Your migration script with SQL queries should also convert the value to fit the new format, etc multiply the value with 86400 in an update statement?

data/migrations/25.lua
```diff
@@ -1,3 +1,6 @@
 function onUpdateDatabase()
-	return false
-end
+	print("> Updating database to version 26 (Better premium time handling)")
+	db.query("ALTER TABLE `accounts` DROP COLUMN `lastday`")
+	db.query("ALTER TABLE `accounts` CHANGE COLUMN `premdays` `premend` int(11) NOT NULL DEFAULT 0");
```
EvilHero90: ```db.query("UPDATE `accounts` SET `premend`= premend * 86400")```
So it updates the values correctly and fixes what @slawkens and @Znote suggested.

schema.sql
```diff
@@ -4,8 +4,7 @@ CREATE TABLE IF NOT EXISTS `accounts` (
   `password` char(40) NOT NULL,
   `secret` char(16) DEFAULT NULL,
   `type` int(11) NOT NULL DEFAULT '1',
-  `premdays` int(11) NOT NULL DEFAULT '0',
-  `lastday` int(10) unsigned NOT NULL DEFAULT '0',
+  `premend` int(11) NOT NULL DEFAULT '0',
```
Znote: If you modify the schema, you also have to bump migration (from 24 to 25) at the bottom INSERT statement `server_config`

EvilHero90: no @Znote that will happen once the ```onUpdateDatabase``` was successfully called, this doesn't have to be applied manual.
I mean the ```db_version```

EvilHero90:
Why has this been tagged for 2.0? it contains everything necessary to be merged as of now and has a clean transition, so I don't really see a reason of why to hault this PR until 2.0

Znote:
This is is a database modification that breaks TFS 1.x compatibility.

nekiro:
> This is is a database modification that breaks TFS 1.x compatibility.

isn't that why we have migrations?

data/migrations/25.lua
```diff
@@ -1,3 +1,7 @@
...
+	db.query("UPDATE `accounts` SET `premend` = `premend` * 86400 + " .. os.time())
+	return true
+end
```
DSpeichert, line 7: missing EOL

schema.sql
DSpeichert, line 7: I'd favor a more verbose `premium_ends_at`, I don't know why we make it so short and vague.

src/account.h
```diff
@@ -26,9 +26,8 @@ struct Account {
 	std::vector<std::string> characters;
 	std::string name;
 	std::string key;
-	time_t lastDay = 0;
 	uint32_t id = 0;
-	uint16_t premiumDays = 0;
+	int32_t premEnd = 0;
```
DSpeichert, line 30:
```suggestion
	int32_t premiumEndsAt = 0;
```

DSpeichert (changes requested):
If the migrations work well and Lua old functions stay functional, I'm okay with this going into 1.4 as it's a nice simplification of an otherwise weird and complicated system.

@Znote I know it's probably going to require special handling for AAC, that's why we intend for this to be tagged at 1.4. 

My requested changes relate to just naming of vars, not the mechanism, which is fine.

DSpeichert (approved):

EPuncker:
yeah finally

nekiro:
Hey @Znote your request is blocking this PR from being merged.

Znote (approved):
~unblocking~

DSpeichert:
Actually - one more thing. I know we've always integer timestamp since epoch but how complex would it be to instead store `premium_ends_at` as `DATETIME` in MySQL? It just seems like the right thing to do.

nekiro:
@DSpeichert conversion in migration is kinda more complex now, but you cannot convert integer to datetime directly, because mysql fails to convert it.

slawkens:
This is kinda bad idea to use DATETIME as actually all columns that exist in TFS are timestamps.

Examples:
```
players.lastlogin, lastlogout
account_bans.banned_at, expires_at
player_namelocks.namelocked_at
guilds.creationdata
guildwar_kills.time
market_history.expires_at
```
and many more....

Also what is worth noticing is that MySQL differently saves TIMESTAMP and DATETIME (one with timezone adjustments, and other one without).

So I highly recomment to keeping some standards in the repository.

Either change ALL timestamps to DATETIME or don't change anything. Because one DATETIME vs rest TIMESTAMPS is kinda bad practice.

Second thing: The naming convention. Here again with change from premend to premium_ends_at you are breaking compatibility with OTServ. OTServ and TFS 0.3/0.4 both are using "premend" column. So why break standards and not keep as it was for years?

DSpeichert:
@slawkens You make good arguments.

I have a question about `premend` - you say it's used in 0.3/0.4. I see it's a column on `players` table. How was it used? I don't think it's for the same purpose we're using it.

https://github.com/otland/tfs-old-svn/blob/curated/r3884/schemas/mysql.sql#L100

slawkens:
> @slawkens You make good arguments.
> 
> I have a question about `premend` - you say it's used in 0.3/0.4. I see it's a column on `players` table. How was it used? I don't think it's for the same purpose we're using it.
> 
> https://github.com/otland/tfs-old-svn/blob/curated/r3884/schemas/mysql.sql#L100

Then I was wrong, it wasn't used in 0.3/0.4, just in OTServ, here:
https://github.com/opentibia/server/blob/33a81ef95a9b407533e0a7ce48aff12204b4f3b1/sql/schema.mysql#L42

nekiro:
@slawkens who cares about otserv?
I can revert datetime changes, because it's really problematic and I understand it, even from source look it's kinda bad to have to use another sql method to cast it to unix timestamp. The name is though okay and descriptive, all you have to prolly change in your AAC and others AAC rename one column in accounts query, this shouldn't be that bad. So I will leave the name as it is.

slawkens:
@nekiro 

50% of code in this repo is based on otserv, and the structure of the MySQL schema is another 50% of otserv.

So be careful when you say "who cares about otserv" ( ͡° ͜ʖ ͡°)

nekiro:
> @nekiro 
> 
> 
> 
> 50% of code in this repo is based on otserv, and the structure of the MySQL schema is another 50% of otserv.
> 
> 
> 
> So be careful when you say "who cares about otserv" ( ͡° ͜ʖ ͡°)

Doesn't mean anything. Anyway repo is ready to be merged.

src/luascript.cpp
```diff
@@ -9391,55 +9391,52 @@ int LuaScriptInterface::luaPlayerHasMount(lua_State* L) {
...
-		}
+	uint32_t seconds = getNumber<uint32_t>(L, 2);
+	int32_t oldTime = std::max<int32_t>(0, player->premiumEndsAt - time(nullptr));
+	if (oldTime < seconds) {
+		player->setPremiumTime(0);
+		IOLoginData::updatePremiumTime(player->getAccount(), 0);
+	} else {
+		player->setPremiumTime(time(nullptr) + (oldTime - seconds));
+		IOLoginData::updatePremiumTime(player->getAccount(), player->premiumEndsAt);
 	}
```
DSpeichert, line 9440: I think what you're doing here is that you're resetting the `premiumEndsAt` timestamp to 0, whenever the premium time is set to a time in the past. That does not sound like expected behavior.

Imagine a Lua script that removes 7 days of premium and then adds 7 days of premium for someone who has 6 days of premium left. With your logic, their premium would now expire 6 days after epoch (in the past) instead of 6 days from now.

```suggestion
	if (player->premiumEndsAt < seconds) {
		player->setPremiumTime(0);
		IOLoginData::updatePremiumTime(player->getAccount(), 0);
	} else {
		player->setPremiumTime(player->premiumEndsAt - seconds);
		IOLoginData::updatePremiumTime(player->getAccount(), player->premiumEndsAt);
	}
```

src/luascript.cpp
```diff
@@ -9391,55 +9391,52 @@ int LuaScriptInterface::luaPlayerHasMount(lua_State* L) {
...
-		int32_t addDays = std::min<int32_t>(0xFFFE - player->premiumDays, days);
-		if (addDays > 0) {
-			player->setPremiumDays(player->premiumDays + addDays);
-			IOLoginData::addPremiumDays(player->getAccount(), addDays);
-		}
-	}
+	uint32_t seconds = getNumber<uint32_t>(L, 2);
+	int32_t oldTime = std::max<int32_t>(0, player->premiumEndsAt - time(nullptr));
+	player->setPremiumTime(time(nullptr) + oldTime + seconds);
+	IOLoginData::updatePremiumTime(player->getAccount(), player->premiumEndsAt);
```
DSpeichert, line 9418: Why calculate time left from now when it's not used?

```suggestion
	uint32_t seconds = getNumber<uint32_t>(L, 2);
	player->setPremiumTime(player->premiumEndsAt + seconds);
	IOLoginData::updatePremiumTime(player->getAccount(), player->premiumEndsAt);
```

DSpeichert (changes requested):
I think it would be easier for Player:getPremiumTime to return a timestamp and be called Player:getPremiumEndsAt, as well as Player:setPremiumEndsAt. The remaining functions (add/remove as well as compat) can be all done in Lua.

src/luascript.cpp
```diff
@@ -9391,56 +9391,46 @@ int LuaScriptInterface::luaPlayerHasMount(lua_State* L) {
...
 
-int LuaScriptInterface::luaPlayerGetPremiumDays(lua_State* L)
+int LuaScriptInterface::luaPlayerGetPremiumTime(lua_State* L)
 {
-	// player:getPremiumDays()
+	// player:getPremiumTime()
 	Player* player = getUserdata<Player>(L, 1);
 	if (player) {
-		lua_pushnumber(L, player->premiumDays);
+		lua_pushnumber(L, std::max<int32_t>(0, player->premiumEndsAt - time(nullptr)));
```
DSpeichert, line 9399:
```suggestion
        #include <climits>
		lua_pushnumber(L, std::max<int32_t>(INT_MAX, player->premiumEndsAt - time(nullptr)));
```
Maybe cap at max int32 instead of resetting to zero?

nekiro: isn't that equivalent to std::numeric_limits<int32_t>::max()?

DSpeichert (approved):

EvilHero90 (approved):

PR #2815, opened 2020-03-05
Immune monsters condition fix

### LINKED ISSUE #2540: Monsters and immune damage effect
- [ x] This is a bug in the software that resides in this repository, and not a
- [x ] This issue is reproducible without changes to the code in this repository

### Steps to reproduce
1. Use a poison field on ground and wait for example a demon skeleton pass over field, some seconds after, the creature will receive the immune effect.

### Expected behaviour

When the monsters are immune to certain damage, only receive the effect when walk over field, the effect not continue until the damage is done.

### Actual behaviour

when the monsters are immune to certain damage, receive the effect until the damage is done.

You can see here the issue.
I tested with slime this time.
https://streamable.com/hgpp9

### LINKED ISSUE #2651: Stepping into firefield while having fire condition from soulfire does not inflict damage
Hey there,

recently I stumbled across the bug with conditions. Steps to reproduce:
1. Get hurt by soulfire rune.
2. Get hurt by firebomb and try to walk on  fire fields

In this scenario player does not receive any damage from fields, and after a while player gets damage until he dies. Any ideas on how to fix it :)? Tested on newest rev of tfs 1.3.

I also found gif of this bug:
69154412_2440797525966780_8664424821154643968_o

nekiro:
Fix #2540


nekiro:
This pr was intended to address soulfire rune issues, but it does not, but it does fix the other issue mentioned in description/

PR #2818, opened 2020-03-06
Small fixes following up PR #2800

EPuncker:
- removed duplicate registered heavy machete from actions.xml
- fixed a duplicated item ID
- updated crazed beggar loot to use ID, because now we have two items with same name

a big thanks to @BahamutxD for noticing them


WibbenZ:
Thanks!

PR #2821, opened 2020-03-10
Rewrite Spell constructor to avoid conflicts & properly return runes

infernumx:
* Can now access existing runes through Spell(runeId) calls, previously impossible because of the initial type check where 2nd arg is converted, resulting in an equivalent of a Spell(SPELL_INSTANT) call
* Removed Spell(spellId) calls because they are not unique identifiers and conflict with underlying SpellType_t ids
* Checks for existing spells now  and falls back to constructing a new spell with the determined type, using int32_t number instead of SpellType_t when retrieving 2nd argument to avoid undefined behavior when the underlying `static_cast` occurs to SpellType_t with an out-of-range value

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -13896,9 +13896,40 @@ int LuaScriptInterface::luaSpellCreate(lua_State* L)
 		return 1;
 	}
 
-	SpellType_t type = getNumber<SpellType_t>(L, 2);
+	SpellType_t type = SPELL_UNDEFINED;
 
-	if (isString(L, 2)) {
+	if (isNumber(L, 2)) {
+		int32_t id = getNumber<int32_t>(L, 2);
+		RuneSpell* rune = g_spells->getRuneSpell(id);
+
+		if (rune) {
+			pushUserdata<Spell>(L, rune);
+			setMetatable(L, -1, "Spell");
+			return 1;
+		}
+
+		type = static_cast<SpellType_t>(id);
+	} else if (isString(L, 2)) {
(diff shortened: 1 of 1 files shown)


src/luascript.cpp
```diff
@@ -13957,7 +13929,31 @@ int LuaScriptInterface::luaSpellCreate(lua_State* L)
 			setMetatable(L, -1, "Spell");
 			return 1;
 		}
+
+		std::string tmp = asLowerCaseString(getString(L, 2));
```
TheSumm: `arg` can be reused here: `asLowerCaseString(arg)`

I would also prefer a different name for the variable e.g. `spellType`

Znote:
We are missing wiki documentation on the Spell() metatable.

TheSumm:
I meant the variable `tmp`. Sorry for the misunderstanding.

Znote:
I failed to build the [Spell metatable](https://github.com/otland/forgottenserver/wiki/Metatable:Spell) in our wiki due to my bad perception of our source code. 

Here is my [gist](https://gist.github.com/Znote/3039a032fb468efb9e80ec8e5ea9e57b) and [raw](https://gist.githubusercontent.com/Znote/3039a032fb468efb9e80ec8e5ea9e57b/raw/f29a49cef4578bc61102006f78e05c0e5d32a279/MetatableSpells.md)

I am unable to properly review this until this is in place.

Sickness-chan:
What's the status on this? @nekiro @DSpeichert

Issue #2771, opened 2020-02-01
Low level yelling "new" behaviour


EPuncker:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

Tibia has changed how yelling was restricted on low levels and I think we need to do so too (why not? 😝 )

### Expected behaviour
You should not yell as long as you are bellow lvl 20 or your account has premium status
(the message says: "You may not yell unless you have reached level 20 or your account has premium status")

### Actual behaviour
Currently we only restrict yelling to lvl > 1

### My suggestions
- [x] add the level 20 or premium status needed
- [x] make level configurable in config.lua
- [x] make premium restriction configurable in config.lua

PS: can't test with a char level 1 since you start at level 2 now when creating a character no matter what
PS²: need someone with a premium account available to check what the message says when you are premium and level 2 and tries to yell

Znote:
What if your stuck in a cave without a rope? :P

EPuncker:
> What if your stuck in a cave without a rope in rook? :P

they remapped everything so you can't be stuck anymore, to enter any cave that need a rope, you need to have a rope in the first place ;P

WibbenZ:
In that case we should implement that at the same time, not a bad ide tbh

PR #2823, opened 2020-03-11
New configurable 'Yell' options

Leo32onGIT:
This is for: https://github.com/otland/forgottenserver/issues/2771

`yellAlwaysAllowPremium`
if `true` players less than `yellMinimumLevel` need to be premium in order to yell.
if `false` players less than `yellMinimumLevel` may not yell.

Tested in all all four instances and works exactly as expected.

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -76,6 +76,8 @@ classicEquipmentSlots = false
 classicAttackSpeed = false
 showScriptsLogInConsole = true
 showOnlineStatusInCharlist = false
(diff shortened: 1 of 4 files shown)


Leo32onGIT:
Original issue conversation mentioned level 20.

Is that was cip do now?
Do we want to replicate that and have 20 the default?

EPuncker:
they did that to prevent bots, but I think any value is good (2 or 20) as people tend to change most of the things in config.lua this shouldn't be an issue

PR #2832, opened 2020-03-13
Event - onHear

EvilHero90:
Creatures (Player/Monster/NPC) can now be responsive through other Creatures talking.
There's to much possibilities to even list up on what could be achieved with this.

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -247,7 +247,7 @@ ReturnValue Events::eventCreatureOnTargetCombat(Creature* creature, Creature* ta
 	return returnValue;
 }
 
-void Events::eventCreatureOnHear(Creature* creature, Creature* speaker, const std::string words, SpeakClasses type, Position pos)
+void Events::eventCreatureOnHear(Creature* creature, Creature* speaker, const std::string& words, SpeakClasses type, const Position& pos)
 {
 	// Creature:onHear(speaker, words, type, pos)
 	if (info.creatureOnHear == -1) {
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -247,9 +247,9 @@ ReturnValue Events::eventCreatureOnTargetCombat(Creature* creature, Creature* ta
 	return returnValue;
 }
 
-void Events::eventCreatureOnHear(Creature* creature, Creature* speaker, const std::string& words, SpeakClasses type, const Position& pos)
+void Events::eventCreatureOnHear(Creature* creature, Creature* speaker, const std::string& words, SpeakClasses type)
 {
-	// Creature:onHear(speaker, words, type, pos)
+	// Creature:onHear(speaker, words, type)
 	if (info.creatureOnHear == -1) {
 		return;
 	}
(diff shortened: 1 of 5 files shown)


src/events.cpp
```diff
@@ -245,6 +247,37 @@ ReturnValue Events::eventCreatureOnTargetCombat(Creature* creature, Creature* ta
 	return returnValue;
 }
 
+void Events::eventCreatureOnHear(Creature* creature, Creature* speaker, const std::string words, SpeakClasses type, Position pos)
```
TheSumm: `words` should be a reference.

src/events.cpp
```diff
@@ -245,6 +247,37 @@ ReturnValue Events::eventCreatureOnTargetCombat(Creature* creature, Creature* ta
...
+
+	LuaScriptInterface::pushUserdata<Creature>(L, creature);
+	LuaScriptInterface::setCreatureMetatable(L, -1, creature);
+
+	LuaScriptInterface::pushUserdata<Creature>(L, speaker);
+	LuaScriptInterface::setCreatureMetatable(L, -1, speaker);
+
+	LuaScriptInterface::pushString(L, words);
+	lua_pushnumber(L, type);
+	LuaScriptInterface::pushPosition(L, pos);
```
TheSumm: I don't see the purpose of the `pos` parameter. You can easily retrieve that information in the event by calling `speaker:getPosition()`.

src/events.h
```diff
@@ -72,6 +74,7 @@ class Events
 		bool eventCreatureOnChangeOutfit(Creature* creature, const Outfit_t& outfit);
 		ReturnValue eventCreatureOnAreaCombat(Creature* creature, Tile* tile, bool aggressive);
 		ReturnValue eventCreatureOnTargetCombat(Creature* creature, Creature* target);
+		void eventCreatureOnHear(Creature* creature, Creature* speaker, const std::string words, SpeakClasses type, Position pos);
```
nekiro: if you arent removing that position, it should be passed by reference and const like words (which isnt)

soul4soul:
Cool idea to add onHear event. I can imagine a quest/task where a creature needs to be lured to a NPC to scare them etc...

src/events.cpp
EvilHero90: I know that you can do it that way aswell but sometimes it's not bad to have the extra param

src/events.h
EvilHero90: done

src/events.cpp
EvilHero90: done

src/events.cpp
WibbenZ: I agree, there is no need for it.
All you are doing is sending speaker:getPosition() but from C++ as creature->getPosition()
If the function is needed for a code we can do it via Lua.

Leo32onGIT: I'd have to agree with @TheSumm here - the position stuff is unnecessary.
The concept of this is fantastic though.

PR #2833, opened 2020-03-14
Changed the heading of all monster files to UTF-8

EPuncker:
Instead of ISO-8859-1, to match file encoding and there is no need to use it as we don't use any latin character in any file


DSpeichert:
Technically you're just changing the header, not encoding.
In fact, I believe the original client only supports latin1 characters, not the full UTF-8 set.
While normally I would argue we should be using UTF-8, I'm not sure this change is actually meaningful in any way.

EPuncker:
@DSpeichert if that is okay I can change their encoding to utf-8 (w/o boom) if they are, but as I've looked, everyone is utf-8 already, and yes this PR is actually to change the heading not the encoding, just to make it consistent since there are files with both

PR #2835, opened 2020-03-15
Fix tile stacking when swapping items

### LINKED ISSUE #2441: NEED_EXCHANGE - not merging cumulatives on the ground when exchanged
The Forgotten Server 1.2

### Steps to reproduce
1. Put 5 gold coins in either slot of your inventory.
2. Put any quantity of gold coins that is lower than 100 on the ground.
3. Put any other item that is not cumulative above the gold coins on the ground.
4. Move the non cumulative item to your inventory so your gold coins are replaced with the cumulative items (gold coins in this case).

### Expected behaviour
The gold coins of your inventory, merge with the gold coins on the ground.

### Actual behaviour
The gold coins of your inventory are placed _above_ the gold coins on the ground, not merging them, resulting in capability to stack multiple items like this.

nekiro:
Fix #2441

### DIFF
--- a/src/tile.cpp
+++ b/src/tile.cpp
@@ -927,6 +927,28 @@ void Tile::addThing(int32_t, Thing* thing)
 						}
 					}
 				}
+			} else if (itemType.stackable) {
+				// find suitable item to stack with
+				for (ItemVector::const_iterator it = items->getBeginDownItem(), end = items->getEndDownItem(); it != end; ++it) {
+					Item* toItem = *it;
+					if (item->equals(toItem) && toItem->getItemCount() != 100) {
+						uint32_t n = std::min<uint32_t>(100 - (*it)->getItemCount(), item->getItemCount());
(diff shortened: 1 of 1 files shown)


TheSumm:
The expected behaviour described in #2441 is wrong. At least it is not what I expect or what actually happened in game.

The procedure described in the issue is exactly how you would create these [gold coin stacks](https://guildstats.eu/sow=76&lang=es&lang=en) back in the day. Nowadays it is no longer possible since you cannot put gold in any of your inventory slots.

nekiro:
@TheSumm this happens when you enable classicEquipmentSlots and as long as we have that in our config.lua, we should handle that bug that he described.

TheSumm:
I know that `classicEquipmentSlots` exists. However, as I said the behaviour we currently have is exactly how it has always worked.

nekiro:
@TheSumm I understand, what if I change the code to only look for the top tile item to stack with? This way we gonna have working stacking when exchange and that stuff you linked also gonna be possible.

PR #2836, opened 2020-03-15
NEEDEXCHANGE Changes

nekiro:
We should only be allowed to exchange items while they are in player or in depot. This pull request adds that functionality. This is based on cip servers behaviour and seems reasonable.


andersonfaaria:
I don't think this was supposed to be done. Now you can't pick an stack of arrows and put in your arrow slot if their summ is higher than 100 without putting in your bp first.

DSpeichert:
@andersonfaaria You can open a new bug report and describe the exact situation that is desirable, how it behaves in RL and how it is currently handled.

PR #2837, opened 2020-03-15
Round down required mana

### LINKED ISSUE #2613: Rounding when determining required mana to reach magic level is incorrect
I was supposed to open this ticket more than a month ago, but completely forgot about it. Well, bettter late than never, right?

A test I've made for the purpose of #2580 has shown that the following way of rounding required mana (i.e. to the nearest multiple of 20) is incorrect:
https://github.com/otland/forgottenserver/blob/a382f16cc341677b62b1308eef6db7717c48c94c/src/vocation.cpp#L207-L216

As a sorcerer you need exactly 2576 mana to go from magic level 5 to 6, (which is exactly `floor(1600 * 1.1^5)`), but according to the current formula it would be rounded to 2580. I'm not sure what exactly is the correct way to round it, but rounding it down to the nearest integer at least works in this one case and I wouldn't be surprised if it worked every time.

nekiro:
Resolve #2613

### DIFF
--- a/src/vocation.cpp
+++ b/src/vocation.cpp
@@ -204,14 +204,7 @@ uint64_t Vocation::getReqMana(uint32_t magLevel)
 		return it->second;
 	}
 
-	uint64_t reqMana = static_cast<uint64_t>(1600 * std::pow<double>(manaMultiplier, static_cast<int32_t>(magLevel) - 1));
-	uint32_t modResult = reqMana % 20;
-	if (modResult < 10) {
-		reqMana -= modResult;
-	} else {
-		reqMana -= modResult + 20;
-	}
-
+	uint64_t reqMana = std::floor<uint64_t>(1600 * std::pow<double>(manaMultiplier, static_cast<int32_t>(magLevel) - 1));
 	cacheMana[magLevel] = reqMana;
 	return reqMana;
 }


Leo32onGIT:
Simple fix?
Was there a reason we were rounding to the nearest 20 to begin with?

Znote:
> Was there a reason we were rounding to the nearest 20 to begin with?
> Anyone know?

Because this was actually how it worked on the olden days. Not sure which protocol they changed it, but back in the days cip would only record manaspent at an interval at 20 (like 7.5 or 7.6?). So you wouldn't gain magic level any faster by casting `exura` (25 mana) than utevo lux (20 mana).

soul4soul:
> 
> 
> > Was there a reason we were rounding to the nearest 20 to begin with?
> > Anyone know?
> 
> Because this was actually how it worked on the olden days. Not sure which protocol they changed it, but back in the days cip would only record manaspent at an interval at 20 (like 7.5 or 7.6?). So you wouldn't gain magic level any faster by casting `exura` (25 mana) than utevo lux (20 mana).

Good times. I remember making HMMs at 70 mana per rune. The trick was to wait until 140 mana and put a blank rune in each hand to avoid losing mana spent. 

A site perhaps tibia-status has an interesting research paper about mana training.

PR #2838, opened 2020-03-15
Map converted to 10.98 protocol

nekiro:
As our forgottenserver.otbm stopped at some 8.0/8.1 protocol. I coverted it using latest RME to be 10.98 like the protocol version.


nekiro:
Make sure you are loading correct map. Looks like you arent, cause I dont have 8.1 client either and RME loads the map for me. If you are sure you do, maybe I pushed the wrong map to this branch.

PR #2841, opened 2020-03-15
Notify monsters that summon appears

### LINKED ISSUE #2235: bug in new setMaster behavior

### Steps
1. Place any monster near your
2. Summon any monster when the monster placed is in idle state
3. Puff. Because the summon is placed first as a monster, theres no entry in the surrounding monsters target list, our old setMaster behavior as using the monster class conviceCreature, which is updating the target list.

### Expected behaviour
Surrounding monsters target list should be updated when in idle state and on placing a summon

### Actual behaviour
Its not

nekiro:
Fix #2235


WibbenZ:
Haven't tested it, but if you ask me it would be better to add this to C++ since we (and other datapacks) call the x:addSummon() / setMaster functions in other scripts.

nekiro:
@WibbenZ then we should add it to addSummon method (its in lua)

WibbenZ:
My mistake and in that case, yes we should add it there.
In these cases it's better to fix it at the source of a function rather than adding small fixes in all the codes we have(not to mention backwards compatibility).

If the rest disagrees with me you would have to fix it in these cases aswell: https://github.com/otland/forgottenserver/search?q=addSummon&unscoped_q=addSummon

Leo32onGIT:
The only situations that trigger this, is when users on GOD characters are summoning monsters and trying to mess with players.

If you're doing that, who cares if monster are idle or 'dance' near you, plus it means you've already messed with groups.xml - so maybe this is appropriate?

https://github.com/otland/forgottenserver/pull/2848#issuecomment-599689755

nekiro:
@Leo32onGIT because its annoying.

PR #2846, opened 2020-03-16
Added secret service tool gears

EPuncker:
They were added on 8.5 update
- moved main functions to actions lib file
- added crowbar and kitchen knife functions
- fixed the jammed tool item description
- script kinda based on the one from ORTS
- added jamming chance

and kinda tested them in game


data/actions/lib/actions.lua
```diff
@@ -1,3 +1,20 @@
+local wildGrowth = { 1499, 11099 } -- wild growth destroyable by machete
+local jungleGrass = { -- grass destroyable by machete
+	[2782] = 2781,
+	[3985] = 3984,
+	[19433] = 19431
+}
+local groundIds = {354, 355} -- pick usable ground
```
WibbenZ: Add spaces around the other curly brackets (looks better) or remove them from line 1

data/actions/lib/actions.lua
```diff
@@ -39,3 +56,175 @@ function destroyItem(player, target, toPosition)
...
+	if not table.contains({2566, 10511, 10515}, item.itemid) then
+		return false
+	end
+
+	if table.contains(fruits, target.itemid) and player:removeItem(6278, 1) then
+		target:remove(1)
+		player:addItem(6279, 1)
+		player:getPosition():sendMagicEffect(CONST_ME_MAGIC_GREEN)
+	else
+		return false
```
WibbenZ: You can return true here and false on line 229, that way you don't need the else statment.

EPuncker: like this?

PR #2847, opened 2020-03-16
Added achievements system

EPuncker:
Achievements Lib Created By Eduardo Montilva (Darkhaos)

added most of them until 10.98 update, still have lots missing tho 😝 

Most of them can be added, but I see no need in taking out the quest ones, because some may have an use for it in their server...


data/lib/core/achievements.lua
```diff
@@ -0,0 +1,659 @@
...
+	Player:getSecretAchievements()
+	Player:getPublicAchievements()
+	Player:getAchievements()
+	isAchievementSecret(achievement_id/name)
+	Player:getAchievementPoints()
+	Player:addAchievementProgress()
+]]
+
+-- base storage
+ACHIEVEMENTS_BASE = 300000
```
DSpeichert: data/lib/core/storages.lua ?

EPuncker: since it will increase by itself, how will it work?

nekiro:
I think it would ne better if we had all these achievement methods in one global table Achievements.getInfoById etc

data/lib/core/achievements.lua
```diff
@@ -0,0 +1,659 @@
...
+	[255] = {name = "Heartburn", grade = 1, points = 3, description = "Never-tiring, you attack the inner organs of the mighty hive. Your attacks on the hive's digestion system begin to cause some trouble."},
+	[256] = {name = "Hickup", grade = 1, points = 2, description = "You have grown accustomed to frequenting the hive's stomach system. Your actions have caused the hive some first digestion problems."},
+	[257] = {name = "Hive Blinder", grade = 2, points = 4, description = "You have put a lot of time and energy into keeping the hive unaware of what is happening on Quirefang. The hive learnt to fear your actions. It would surely crush you with all its might ... if it could only find you!"},	
+	[258] = {name = "Hive Fighter", grade = 1, points = 1, description = "You have participated that much in the hive war, that you are able to create some makeshift armor from the remains of dead hive born that can be found in the major hive, to show of your skill."},
+	[259] = {name = "Hive Infiltrator", grade = 1, points = 3, description = "The most powerful warriors of the hive were killed by you by the dozens. The hive is not safe anymore because of your acti
```
WibbenZ: Would look better if you don't have these as one-liners (same with the rest below)

WibbenZ: Update the variable names to make it more clear what we are storing

data/lib/core/achievements.lua
DSpeichert: Just the base should be there with proper comment that a lot of them from that starting point to X are used.
What's the point of central place to store storage IDs otherwise?

data/lib/core/achievements.lua
EPuncker: did you comment on the right line? I don't quite understand what are you saying

WibbenZ: Yes but github seem to be the issue 😂
Click changes at the top(what you changed) and scroll down, atleast I can see the correct lines I commented on there 👍

EPuncker: sorry but I don't know what they do and there are so many of them that it would be nice if someone could do a list of the changes

WibbenZ: Once again, github seems to be the issue, click changes and scroll down to line 472 (local t = ...)

Variables like those (in most loops) could have better names.
No an issue merging the PR but to keep the code clean and readable it should be changed.

EPuncker: yeah, but my point still, I don't know what each one is storing xD

WibbenZ: We are trying to get info from an achievement and we need 1 variable to do that (id)
Then we loop the achievements and if we find one with the same id we create a new table (t)
We add the id and the storage and the content from achievements[id]

So to make it simple you could name them achievementsToReturn / targetAchievement or something like that.
Almost all (if not all) can be named that since the tables are for achievements.

EPuncker: like this?

data/lib/core/achievements.lua
```diff
@@ -0,0 +1,668 @@
...
+	local storage = PlayerStorageKeys.achievementsCounter + achievement.id
+	local progress = self:getStorageValue(storage)
+	if progress < value then
+		self:setStorageValue(storage, math.max(1, progress) + 1)
+	elseif progress == value then
+		self:setStorageValue(storage, value + 1)
+		self:addAchievement(achievement.id)
+	end
+	return true
+end
```
soul4soul: Nice addition to the code base, I think a few changes can make it even better.

The total amount of times something needs to be completed should be defined in the `achievements` table. A good name for the variable might be `goal`. It doesn't feel right that the `goal` amount needs to be passed into this function for every call. All `goal`s should be defined in the `achievements` table instead of all over the lua scripts.

The last parameter for this function could then me a `count` which is how much progress to add for the achievement towards reaching the goal.

This change makes the system more flexible too. It would be possible to due things like have mana potions add 1 to the progress. strong mana potions add 2 to the progress. ultimate mana potions add 3 to the progress.

omarcopires:
You could be migrating the functions to compat.lua.

I also think it is unnecessary to keep the comment above explaining the functions and giving credit to the author.
Imagine if everyone who sends a pull request inserts their credit somewhere in the code. In my point of view, it is enough to give credit to the pull request, this is more than enough.

EPuncker:
> You could be migrating the functions to compat.lua.
> 
> I also think it is unnecessary to keep the comment above explaining the functions and giving credit to the author.
> Imagine if everyone who sends a pull request inserts their credit somewhere in the code. In my point of view, it is enough to give credit to the pull request, this is more than enough.

compat.lua is meant for compatibility between old function and new functions, not for adding new functions 😋  and yes I agree about the credit, will change it

EPuncker:
does anyone know why the appveyor build failed with the latest master branch merge?

Znote:
Can this be converted to a revscriptsys modification? So it can easily be removed for those who don't want it?

EPuncker:
> Can this be converted to a revscriptsys modification? So it can easily be removed for those who don't want it?

maybe a config.lua entry? enableAchievements = true/false 🤔

DSpeichert:
Given how this follow RL, I see no problem with merging this but it needs a rebase.

EPuncker:
> Given how this follow RL, I see no problem with merging this but it needs a rebase.

done

Zbizu:
I suggest replacing word Tibia in achievements.lua (lib folder)

PR #2848, opened 2020-03-16
Adjust group flags

nekiro:
Adjusted group flags like they were before these changes in groups.xml. GMs/Gods should be able to use combat and attack monster and generate loot.


Leo32onGIT:
This is what I personally use, plus:
`<flag ignoredbymonsters="0" />`

Which also resolves #2841 #2235.

nekiro:
@Leo32onGIT no, setting gods to be visible for monsters isnt the solution for idle monsters.

Leo32onGIT:
It solves the situation where idle monsters ignore the summons that gods create.

data/XML/groups.xml
```diff
@@ -3,9 +3,9 @@
 	<group id="1" name="player" access="0" maxdepotitems="0" maxvipentries="0" />
 	<group id="2" name="gamemaster" access="1" maxdepotitems="0" maxvipentries="200" >
 		<flags>
-			<flag cannotusecombat="1" />
+			<flag cannotusecombat="0" />
```
WibbenZ: GMs shouldn't be able to kill monsters / get loot.
https://github.com/otland/forgottenserver/commit/7f8002e1479b75e6be1349cec80089e192118072#diff-f3b5bad7afcfdf743ffeb317b0a09f3aL4
https://my-aac.org/flags/

nekiro: They were before, because I clearly remember I could do that, right now each time I get default tfs I'm unable to attack anything and get loot from anything,

WibbenZ: Not with 1.x in that case, just checked previous commits and no one has changed the GM flags.
And if you ask me a GM shouldn't be able to kill monsters, since it could be abused in quests etc

The god update is correct, they were able to loot and attack monsters etc

nekiro: I did literally, when writting the new group flags code, from the numeric flags.

WibbenZ: This commit was the one before you changed how we load the flags
https://github.com/otland/forgottenserver/commit/308abd117fb14ef3b47be4cfff8053ee08dd335e#diff-f3b5bad7afcfdf743ffeb317b0a09f3a
And with those flags (137438953471) you were unable to (as a GM) loot items or attack.

nekiro: I see, I missed that one then.

WibbenZ: @nekiro canbroadcast was also changed / removed, so GMs and gods can't broadcast 😄 
Might be good to check all the flags to make sure everything is like it should be.

nekiro: No idea why would gms/gods not be able to do that? Even if it was like that somewhere before cause someone commited that. It makes no sense. Gms/gods should be able to broadcast, to attack monsters and to use combat. No idea what you mean by „abuse quests” when gods/gms arent regular players and we shouldnt even consider that

DSpeichert:
Finally the god mode is useful in debugging :)

PR #2851, opened 2020-03-17
fix #2843

### LINKED ISSUE #2843: Loot from monsters does not check if string is correct

1. Ex in #2842, TFS does not check (/ report an error) when the item does not exist

### Expected behaviour

Throw an error that the item does not exist

### Actual behaviour

Nothing

### Environment

Does not matter AFAIK

EvilHero90:
Added to config.lua
- forceMonsterTypesOnLoad (default = true)

People can decide on their own if they want to load all MonsterTypes immediatly on server startup or not.
This adresses the issue in #2843

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -159,9 +159,12 @@ registerMonsterType.events = function(mtype, mask)
 end
 registerMonsterType.loot = function(mtype, mask)
 	if type(mask.loot) == "table" then
+		local lootError = false
 		for _, loot in pairs(mask.loot) do
 			local parent = Loot()
-			parent:setId(loot.id)
+			if not parent:setId(loot.id) then
+				lootError = true
+			end
 			if loot.chance then
 				parent:setChance(loot.chance)
 			end
@@ -180,7 +183,9 @@ registerMonsterType.loot = function(mtype, mask)
 			if loot.child then
 				for _, children in pairs(loot.child) do
 					local child = Loot()
-					child:setId(children.id)
+					if not child:setId(children.id) then
+						lootError = true
+					end
 					if children.chance then
 						child:setChance(children.chance)
 					end
@@ -200,6 +205,9 @@ registerMonsterType.loot = function(mtype, mask)
 				end
 			end
 			mtype:addLoot(parent)
+		end
+		if lootError then
+			print("[Warning - end] Monster: \"".. mtype:name() .. "\" loot could not correctly be load.")
 		end
 	end
 end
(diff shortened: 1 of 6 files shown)


Znote:
Does this also work together with #2798 to grab all registered xml monster types (even ones that arent used)?

EvilHero90:
Yes, Game.getMonsterTypes() will return in this case all monsters, from monsters.xml

nekiro:
Imo it should be false by default not true, why would we want to load all monster types into memory? If they arent used

EvilHero90:
It should be load as default in debug mode, so it shows errors immediatly as @WibbenZ already said in #2843 
I can add a console debug print, so people are aware it's enabled, so they can turn it off themselves.

EvilHero90:
Implemented revscriptsys checks for loot
tfs5

src/luascript.cpp
```diff
@@ -13111,13 +13111,28 @@ int LuaScriptInterface::luaLootSetId(lua_State* L)
...
+
+			uint32_t id = ids.first->second;
+
+			if (std::next(ids.first) != ids.second) {
+				std::cout << "[Warning - Loot:setId] Non-unique loot item \"" << name << "\". " << std::endl;
+				pushBoolean(L, false);
+				return 1;
+			}
+
+			loot->lootBlock.id = id;
```
WibbenZ: loot->lootBlock.id = ids.first->second
I don't see a reason to create the variable since you don't use it anywhere else

data/scripts/lib/register_monster_type.lua
```diff
@@ -159,9 +159,12 @@ registerMonsterType.events = function(mtype, mask)
 end
 registerMonsterType.loot = function(mtype, mask)
 	if type(mask.loot) == "table" then
+		local lootError = false
 		for _, loot in pairs(mask.loot) do
 			local parent = Loot()
-			parent:setId(loot.id)
+			if not parent:setId(loot.id) then
+				lootError = true
```
WibbenZ: Would be a good ide to print out the id / name + you can remove the lootError code and just print here insted
Same for line 187

EvilHero90: The error handling print is done in the c++ function, so we can differentiate the error.
If I moved the print inside here it would result in double the prints for actually no reason, it doesn't need to remind me after every line that this monster has an error, I'd rather have a summary of errors and at the end it shows which monster it is.

src/luascript.cpp
EvilHero90: removed it

PR #2854, opened 2020-03-18
Remove trailing spaces

WibbenZ:
Removed trailing spaces and added spaces to some XML tags
Removed else statment from sweetheart_ring.lua


DSpeichert:
Needs rebase

PR #2858, opened 2020-03-18
Remove spaces and unified (- _) in monster file names

WibbenZ:
Since it was mixed mon-ster and mon_ster (folder names) I changed so they all use - to make it more clean

Removed spaces from Misc * folders


Znote:
Conflicting files
data/monster/Event_Creatures/the_halloween_hare.xml

DSpeichert:
@WibbenZ a few conflicts but looking forward to it.

soul4soul:
This is obsoleted by #2974

WibbenZ:
Correct, the problem was having spaces in monster names etc, so this is a pointless commit if we are going back to the messy "everything in one folder".

DSpeichert:
I'm not sure if we are right now, that's a low of history pollution for questionable benefit, I'm hesitant.

EPuncker:
> I'm not sure if we are right now, that's a low of history pollution for questionable benefit, I'm hesitant.

I was hesitant at first glance (I even created new sub folders in my former monster update commits), I know how wibenz feels right now, but after looking at it with I noticed a few things:

- new "groups" of monsters will appear at every update, leading to a mess worse than everything on a single folder
- the called "groups" we have right now have no consistency since there is nothing we can base them on... who decides what belongs to where?
- there are groups that have only one creature on it (Cnidarians, Cryo-Elementals, Salamanders etc)
- there are monsters that doesn't have a defined group and are sub-sub-divided into 3 folders (Misc Mammals, Misc Mounts, Misc Reptiles)

while the approach of everything in a single folder may sound "polluted", it is still less confuse to work than having sub folders that will increase by each protocol update

WibbenZ:
Well I can agree on some of your points, but what I do when I wanna find a file (no matter if they are stacked in folders or nested in 1, I just use the "find file" ~ function.
Saves time compared to scrolling down or entering 50 folders to find them.

I guess the folder names are taken from wikia and if you ask me there really is no point to have 3-4 misc folders, would be better to make it into 1 misc without the ex misc-mounts folders.
Same with ex cyro-elementals, move that file into ex elementals or where we think it would fit.

When it comes to new monsters we can always decide where we think they are best placed, ex elementals, demons, bosses etc etc

DSpeichert:
> When it comes to new monsters we can always decide where we think they are best placed, ex elementals, demons, bosses etc etc
I think that's exactly what went wrong last time.

Anyway, so are we closing this in favor of #2974 ?

WibbenZ:
Up to you, I would prefer if they stay in diffrent folders just like actions and spells, but wont really matter for us who uses the search function to find a file.

Cant see any major reason that makes this or the other pr superior in any way, just diffrent ways of org. the files.

DSpeichert:
Well, we're gonna need third (and fourth and fifth) opinion here.

PR #2862, opened 2020-03-19
Rearranging actions.xml by itemid

pegason:
- It has been reorganized for a better search.


data/actions/actions.xml
```diff
@@ -97,143 +100,142 @@
 	<action itemid="24843" script="other/food.lua" />
 	<action itemid="26191" script="other/food.lua" />
 	<action itemid="26201" script="other/food.lua" />
-
+
```
DSpeichert: whitespace

PR #2864, opened 2020-03-20
Perfect blockHit and correct value result

MillhioreBT:
This small problem was present for a long time and apparently nobody cares, since no one else plans to fix it, I have done it for you to support this engine to be more perfect every time.
With this change there will no longer be a difference between an object that protects 50% or two objects that protect 25% + 25%
If you test the sources prior to this change, you will notice that the results are different and mathematically wrong.

TEST VISUAL
TestVisual

Lua Example:
local damage50 = 1000
damage50 = damage50 - (damage50 / 100 * 50)
print(damage50)

local damage2525 = 1000
damage2525 = damage2525 - (damage2525 / 100 * 25)
damage2525 = damage2525 - (damage2525 / 100 * 25)
print(damage2525)

-- Print 500 <-- 50%
-- Print 562.5 <--- 25% + 25%

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1846,9 +1846,10 @@ BlockType_t Player::blockHit(Creature* attacker, CombatType_t combatType, int32_
 			continue;
 		}
 
+		int32_t unmodifiedDamage = damage;
 		const int16_t& absorbPercent = it.abilities->absorbPercent[combatTypeToIndex(combatType)];
 		if (absorbPercent != 0) {
-			damage -= std::round(damage * (absorbPercent / 100.));
+			damage -= std::round(unmodifiedDamage * (absorbPercent / 100.));
 
 			uint16_t charges = item->getCharges();
 			if (charges != 0) {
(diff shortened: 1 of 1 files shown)


Leo32onGIT:
This is because the amount is multiplicative.
25% of 1000 = 750
25% of 750 = 187.5
= 750 - 187.5 = 562.5

This isn't incorrect, this is intentional.
If you want to use absolute percentages you should define that login in your lua script that is reducing the damage.

local sourceDamage = 1000
local reduce25 = sourceDamage / 100 * 25
local reduce30 = sourceDamage / 100 * 30

print(sourceDamage - reduce25 - reduce30) -- 55% absolute reduction

nekiro:
Despite @Leo32onGIT comment, seems like your code does nothing either? You are creating new integer that copies damage variable and then uses it to reduce 
Its value, so you broke the code, because that integer isnt used anywhere else. Did I miss something?

MillhioreBT:
Well the truth I see no reason for it to be intentional, 50 is 50, 25 is 25 without more.
It is not a script as @ Leo32onGIT mentions, it is the source code.
Be that as it may, when some conscious person sees this, they like it and they will agree with me.

Is it I think about it one more time and it makes me laugh XD, how can it be possible that 50 is different from 25 + 25? ajajajajja Greetings loves

nekiro:
This feature/code is working as intended, so there is no need to change it.

PR #2867, opened 2020-03-20
Event Callbacks

EvilHero90:
This will enable callbacks for events, so we can dynamically create callbacks in whichever script we want and it will automaticly pick them up, if the event is enabled in events.xml
example of usage:
```lua
local ec = EventCallback

ec.onTargetCombat = function(self, target)
	print(self:getName(), target:getName())
	return RETURNVALUE_NOERROR
end

ec.onAreaCombat = function(self, tile, isAggressive)
	print(self:getName(), isAggressive)
	return RETURNVALUE_YOUMAYNOTATTACKTHISCREATURE
end

ec.onHear = function(self, speaker, words, type)
	print(self:getName(), speaker:getName(), words, type)
end

ec.onBrowseField = function(self, position)
	print(self:getName(), position.x)
	return false
end

ec.onLook = function(self, thing, position, distance, description)
	print(self:getName(), thing:getName(), position, distance, description)
	return "You see a noob"
end

ec.onLookInBattleList = function(self, creature, distance, description)
	print(self:getName(), creature:getName(), distance, description)
	return "You see a noob too"
end

ec.onLookInShop = function(self, itemType, count, description)
	print(self:getName(), itemType, count, description)
	return "wtf is this"
end

ec.onMoveItem = function(self, item, count, fromPosition, toPosition, fromCylinder, toCylinder)
	print(self, item, count, fromPosition, toPosition, fromCylinder, toCylinder)
	return true
end

ec.onItemMoved = function(self, item, count, fromPosition, toPosition, fromCylinder, toCylinder)
	print("item got moved")
end

ec.onMoveCreature = function(self, creature, fromPosition, toPosition)
	print(self, creature, fromPosition, toPosition)
	return false
end

ec.onReportBug = function(self, message, position, category)
	print(self, message, position, category)
	return true
end

ec.onTurn = function(self, direction)
	print(self, direction)
	return false
end

ec.onGainExperience = function(self, source, exp, rawExp)
	print(self, source, exp, rawExp)
	return exp
end

ec.onWrapItem = function(self, item)
	print(self, item, position)
	return true
end

ec.onSpawn = function(self, position, startup, artificial)
	print(self, position, startup, artificial)
	return false
end
```
I just did basic tests so far, so it still needs a lot more testing before I remove the WIP

(edit by Znote):
Suggested tests:
- [x] onChangeOutfit
- [x] onAreaCombat
- [x] onTargetCombat
- [x] onHear
- [x] onJoin
- [x] onLeave
- [x] onDisband
- [x] onShareExperience
- [x] onBrowseField
- [x] onLook
- [x] onLookInBattleList
- [x] onLookInTrade
- [x] onLookInShop
- [x] onMoveItem
- [x] onItemMoved
- [x] onMoveCreature
- [x] onReportBug
- [x] onReportRuleViolation
- [x] onTurn
- [x] onTradeRequest
- [x] onTradeAccept
- [x] onGainExperience
- [x] onLoseExperience
- [x] onGainSkillTries
- [x] onWrapItem
- [x] onDropLoot
- [x] onSpawn


MillhioreBT:
I like the idea, it opens up possibilities at the highest level, I thought you would write it in the sources, but that's fine, if I find a way to improve it, I won't hesitate to comment on my ideas about it.

EvilHero90:
Feel free to do that, I'm completly open to suggestions.
This leads us to the final stage where revscriptsys can pretty much show it's full potential now.

Znote:
data/scripts/PR_quicktest.lua
```lua
local event = EventCallback

event.onChangeOutfit = function(self, outfit)
    tdump("outfit change:", outfit)
end
```
image

EvilHero90:
The ones which aren't checked should still be working just fine (I just couldn't test them alone right now)
I'd still like to have 1 or 2 more people to test and confirm so we can be on the safe side.

Leo32onGIT:
I'll test this today.
This is part of what I need to redo my attributes system in revscript

MillhioreBT:
I have tested all events and they work fine as long as the event is declared in the revscripts module.

Today I was testing something like this: https://gph.is/g/4AjWPnO
and the results were not what I expected, I don't know if this is exclusively for revscripts or should it really work anywhere?
Note: in the action module, it is often repeated even more.

EvilHero90:
Did you /reload actions a few times before testing it?
The problem is, if you reload the classic interface (actions by example) it wont clear the data table and it'll be added once more with every /reload you do.
feel free to hit me up on discord: Evil Hero#8530 @MillhioreBT

EvilHero90:
I found out the other problem why classic interface is not suited for it.
The classic interfaces get load twice on startup, once from the "Test Interface" and once from the real Interface (actions as an example)
I've added a restriction, so the EventCallback function can only be declared inside the "Scripts Interface"
eventcallbacks

EvilHero90:
LGTM

data/events/scripts/party.lua
```diff
@@ -1,18 +1,19 @@
 function Party:onJoin(player)
-	return true
+	if hasEventCallback(EVENT_CALLBACK_ONJON) then return EventCallback(EVENT_CALLBACK_ONJON, self, player) else return true end
```
Kamenuvol: who is jon? 😛

EvilHero90: will fix that, thanks for pointing it out!

EvilHero90:
Can we get a final review on this please.

data/events/scripts/creature.lua
```diff
@@ -1,14 +1,15 @@
 function Creature:onChangeOutfit(outfit)
-	return true
+	if hasEventCallback(EVENT_CALLBACK_ONCHANGEOUTFIT) then return EventCallback(EVENT_CALLBACK_ONCHANGEOUTFIT, self, outfit) else return true end
```
EPuncker, line 2: no one liners for better readability, here and in the two bellow

nekiro: +1 this look ugly as hell

EvilHero90: I actually did this intentional, so people don't start screwing around with it (because usually people don't touch one liners) but I can change it everywhere I don't mind.

Znote (dismissed):
We should probably set up a wiki page for this.

EPuncker:
2.0 only? 😢

EPuncker (dismissed):

Znote:
This is backwards compatible and tested by multiple users. I think its safe to merge this into 1.4, I don't think it will get further reviews/tests by being stalled as a PR, and is a nice feature addition.

data/events/scripts/player.lua
```diff
@@ -74,145 +79,86 @@ function Player:onLookInBattleList(creature, distance)
...
-	self:sendTextMessage(MESSAGE_INFO_DESCR, description)
+	local ret = hasEventCallback(EVENT_CALLBACK_ONLOOKINBATTLELIST) and EventCallback(EVENT_CALLBACK_ONLOOKINBATTLELIST, self, creature, distance, description) or description
+	self:sendTextMessage(MESSAGE_INFO_DESCR, ret)
 end
 
 function Player:onLookInTrade(partner, item, distance)
-	self:sendTextMessage(MESSAGE_INFO_DESCR, "You see " .. item:getDescription(distance))
+	local description = "You see " .. item:getDescription(distance)
+	local ret = hasEventCallback(EVENT_CALLBACK_ONLOOKINTRADE) and EventCallback(EVENT_CALLBACK_ONLOOKINTRADE, self, partner, item, distance, description) or description
+	self:sendTextMessage(MESSAGE_INFO_DESCR, EventCallback(EVENT_CALLBACK_ONLOOKINTRADE, self, partner, item, distance, description))
```
joseluis2g, line 89: looking at the previous example, seems like you should use ret instead you call the function again.
`self:sendTextMessage(MESSAGE_INFO_DESCR, ret)`

MillhioreBT (approved):
Tested and works correctly, I have been waiting for this for many months, I'm glad it now works great, more free for RevScript <3

DSpeichert (approved):

tartariarpg:
how Install this?

PR #2870, opened 2020-03-22
Use the new doTargetCombat instead of the old one

EPuncker:
follow up to https://github.com/otland/forgottenserver/commit/831adb2090df2b9e02e285fd820bb0e07f5a6d61

and added the correct line at large seashell

### DIFF
--- a/data/actions/scripts/other/largeseashell.lua
+++ b/data/actions/scripts/other/largeseashell.lua
@@ -3,7 +3,7 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 		local chance = math.random(100)
 		local msg = ""
 		if chance <= 16 then
-			doTargetCombat(0, player, COMBAT_PHYSICALDAMAGE, -100, -200, CONST_ME_NONE)
+			doTargetCombat(0, player, COMBAT_PHYSICALDAMAGE, -200, -200, CONST_ME_NONE)
 			msg = "Ouch! You squeezed your fingers."
 		elseif chance > 16 and chance <= 64 then
 			Game.createItem(math.random(7632,7633), 1, player:getPosition())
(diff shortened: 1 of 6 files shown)


data/actions/scripts/other/potions.lua
```diff
@@ -56,11 +56,11 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 		return true
 	else
 		if potion.health then
-			doTargetCombatHealth(0, target, COMBAT_HEALING, potion.health[1], potion.health[2])
+			doTargetCombat(0, target, COMBAT_HEALING, potion.health[1], potion.health[2])
 		end
 
 		if potion.mana then
-			doTargetCombatMana(0, target, potion.mana[1], potion.mana[2])
+			doTargetCombat(0, target, COMBAT_MANADRAIN, potion.mana[1], potion.mana[2])
```
EPuncker: @nekiro is that right? I just wanted to be sure

data/actions/scripts/other/largeseashell.lua
```diff
@@ -3,7 +3,7 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 		local chance = math.random(100)
 		local msg = ""
 		if chance <= 16 then
-			player:addHealth(-200)
```
nekiro: I dont think you should touch that one, its whole different method, so dont change it

EPuncker: actually the original script used doTargetCombat, I was the one who changed it to addHealth because I was noob 😝

nekiro: why is there -100 to -200 if the code before had flat -200?

EPuncker: because the damage is not flat, it is random and take armor into account, just like the hand of the noxius spawn

nekiro: https://tibia.fandom.com/wiki/Large_Seashell_(Closed)
so you wrong, min and max should be -200 not min -100 and max -200

EPuncker: i hope its right now

Issue #2859, opened 2020-03-19
Loot countmax "issue" when using values bigger than 100


EPuncker:
I'm opening it to gather opinions rather than reporting an issue, currently our loot countmax attribute doesn't create any count higher than 100 (which is okay), the only issue here is the hassle when doing monsters that drops lots of gold coins, we need to use:

```
<item name="gold coin" countmax="100" chance="34000" />
<item name="gold coin" countmax="100" chance="34000" />	
<item name="gold coin" countmax="46" chance="20000" />
```

when we could instead use:

`<item name="gold coin" countmax="246" chance="88000" />`

=======
the first one makes more sense if we want to have each "amount" of gold to have different chance like cipsoft does, but since we don't have access to these rates, what should we do?

every opinion is welcome!

MillhioreBT:
Here EPuncker:
I like the idea, and already implemented it thanks to you ;)
https://github.com/otland/forgottenserver/compare/master...MillhioreBT:patch-2

EPuncker:
ohhhh, now that you mention it, maybe it used to work the way I explained, maybe it was just not ported when @joseluis2g ported loot to lua, can you confirm this to me @joseluis2g ? and thanks @MillhioreBT

MillhioreBT:
@EPuncker  You're welcome, even though the system was passed to lua, it still didn't allow adding more than 100 stacked objects

Zbizu:
What if we allowed countmax higher than 100 and just loop through stacks? For example, we rolled 250 gold so it would be added as 100+100+50.

PR #2875, opened 2020-03-27
Make monster loot count accept count higher than 100

EPuncker:
As I've explained in #2859 , @MillhioreBT provided the code, I tested it, and it retains backward compatibility so, why not do it? :+1: 

test with:
`<item name="gold coin" countmax="700" chance="52500" />`

2020-03-27 18_31_28-Groove Música

fixes #2875 
Co-Authored-By: Sarah Wesker <millhiorebt@users.noreply.github.com>

### DIFF
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -18,30 +18,47 @@ function Container.createLootItem(self, item)
 	end
 
 	if itemCount > 0 then
-		local tmpItem = self:addItem(item.itemId, math.min(itemCount, 100))
-		if not tmpItem then
+		local tmpItem = {}
+		while itemCount > 100 do
+			local itm = self:addItem(item.itemId, 100)
+			if itm then
+				tmpItem[#tmpItem + 1] = itm
+			end
+			itemCount = itemCount - 100
+		end
+
+		if itemCount > 0 then
+			local itm = self:addItem(item.itemId, itemCount)
+			if itm then
+				tmpItem[#tmpItem + 1] = itm
+			end
+		end
+		
+		if #tmpItem == 0 then
 			return false
 		end
 
-		if tmpItem:isContainer() then
-			for i = 1, #item.childLoot do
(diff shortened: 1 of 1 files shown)


EvilHero90:
I'm not 100% happy with this, instead of having 1 roll with the entire maxcount, there should rather be a ```repeat``` attribute which let's you set how often this item can be random rolled.
Because we are forgetting a really important problem here, what happens when you give 1000 as maxcount but on this roll the chance was to low? yea exactly you wont have any gold inside then.
odds are completly different if we roll 6x 100 instead of 1x 600 if you get what I mean.

EPuncker:
> I'm not 100% happy with this, instead of having 1 roll with the entire maxcount, there should rather be a `repeat` attribute which let's you set how often this item can be random rolled.
> Because we are forgetting a really important problem here, what happens when you give 1000 as maxcount but on this roll the chance was to low? yea exactly you wont have any gold inside then.
> odds are completly different if we roll 6x 100 instead of 1x 600 if you get what I mean.

good point, I'll close this PR and let @MillhioreBT do the job in another PR so he can edit code if he want to 😛

EvilHero90:
There was no need to close this, should have rather let her do the changes and then merge them in here, opening same PR a few times is rather annoying when browsing through them (even if they're closed)

MillhioreBT:
> I'm not 100% happy with this, instead of having 1 roll with the entire maxcount, there should rather be a `repeat` attribute which let's you set how often this item can be random rolled.
> Because we are forgetting a really important problem here, what happens when you give 1000 as maxcount but on this roll the chance was to low? yea exactly you wont have any gold inside then.
> odds are completly different if we roll 6x 100 instead of 1x 600 if you get what I mean.

I made certain modifications to expand a new way of granting more objects on a single line, it was not with the intention of replacing the default use.

in the same way I will take into account everything you said and I will begin to correct this in search of possible problems, improvements... :)

PR #2882, opened 2020-03-31
Fix Item::equals

CristoferMartins:
Better attributes class content comparison between items.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -196,12 +196,10 @@ bool Item::equals(const Item* otherItem) const
 		return false;
 	}
 
-	if (!attributes) {
-		return otherItem->attributes ? (otherItem->attributes->attributeBits == 0) : true;
-	}
-
 	const auto& otherAttributes = otherItem->attributes;
-	if (!otherAttributes) {
+	if (!attributes) {
+		return !otherAttributes || (otherAttributes->attributeBits == 0);
+	} else if (!otherAttributes) {
 		return (attributes->attributeBits == 0);
 	}
 
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -197,11 +197,15 @@ bool Item::equals(const Item* otherItem) const
 	}
 
 	if (!attributes) {
-		return !otherItem->attributes;
+		return otherItem->attributes ? (otherItem->attributes->attributeBits == 0) : true;
 	}
 
 	const auto& otherAttributes = otherItem->attributes;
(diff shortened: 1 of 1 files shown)


src/item.cpp
```diff
@@ -197,11 +197,15 @@ bool Item::equals(const Item* otherItem) const
...
 
 	if (!attributes) {
-		return !otherItem->attributes;
+		return otherItem->attributes ? (otherItem->attributes->attributeBits == 0) : true;
 	}
 
 	const auto& otherAttributes = otherItem->attributes;
-	if (!otherAttributes || attributes->attributeBits != otherAttributes->attributeBits) {
+	if (!otherAttributes) {
+		return (attributes->attributeBits == 0);
```
andersonfaaria, line 205: this should be

src/item.cpp
```diff
@@ -197,11 +197,15 @@ bool Item::equals(const Item* otherItem) const
 	}
 
 	if (!attributes) {
-		return !otherItem->attributes;
+		return otherItem->attributes ? (otherItem->attributes->attributeBits == 0) : true;
 	}
 
 	const auto& otherAttributes = otherItem->attributes;
-	if (!otherAttributes || attributes->attributeBits != otherAttributes->attributeBits) {
+	if (!otherAttributes) {
```
andersonfaaria, line 204: You should use:
```
if (!otherAttributes && attributes->attributeBits != 0) {
		return false;
```
otherwise you'll return true and skip the checks below

src/item.cpp
```diff
@@ -197,11 +197,15 @@ bool Item::equals(const Item* otherItem) const
 	}
 
 	if (!attributes) {
-		return !otherItem->attributes;
+		return otherItem->attributes ? (otherItem->attributes->attributeBits == 0) : true;
```
rechandler12, line 200: maybe:
`return !otherItem->attributes || otherItem->attributes->attributeBits == 0`

CristoferMartins: True, both do the same but yours is better.

nekiro:
What was wrong with old comparison and what changes that give? You didn't explain anything

CristoferMartins:
@nekiro Old Item::equals only checked if the item had the Attributes class instantiated to know if it had no attributes.  If the item had attributes before and they were all removed(so Attributes is empty but instantiated)  Item::equals would return false even if the both items are equal(both are empty whatever they have the Attributes instantiated or attributeBits = 0).

PR #2883, opened 2020-04-01
ItemAttributes optimization

CristoferMartins:
Replaced std::forward_list with std::vector to avoid cache misses.


CristoferMartins:
@ranisalt i believe its only allow one type of attribute in the list(so in any case may only remove one Attribute at max). Attributes do not require any order.
Btw here "(*it) = attributes.back();" should it use std::move?

ranisalt:
No need to, attribute returns a copy so it is probably being moved anyway.

PR #2895, opened 2020-04-04
Fixes to mad sheep summoning

EPuncker:
- Added cooldown so you can only summon one every 12 hours
- Removed allowFarUse from spellwand, it can only be used in close range
- Increased mad sheep speed, she is pretty fast


omarcopires:
Is there any reason why otland does not work with hours instead of direct calculation, for example, why don't we use 12 * 60 * 60 instead of 43200?

EPuncker:
> Is there any reason why otland does not work with hours instead of direct calculation, for example, why don't we use 12 * 60 * 60 instead of 43200?

true, it is better for reading, I'll change it

PR #2897, opened 2020-04-05
Fix crash when reloading invalid CreatureEvents

MillhioreBT:
The server crashes when you reload the CreatureEvents module, this happens when you create an event, and then delete it, the server still tries to use the copy of this event, but the copy is invalid, because it was cleaned in the reload.

GIF/already fixed test:
creatureEventFixed


EvilHero90:
LGTM

PR #2904, opened 2020-04-06
Moved changesex cost to config.lua

EPuncker:
also made it possible to disable it by setting it to -1

that is my first "source edit" (just a copy of buyhouse code), please check if its right


soul4soul:
I don't see the purpose of moving `sexchangecost` to config.lua

- The required and most frequently changed configuration should be in config.lua
- How many people are changing the premium time cost of sex change? I suspect not many, at least in my OT travels I rarely see sex change being used.
- Changing it with the current (pre PR) method is still a one line change in an aptly named file
- Adding to config.lua values makes the most sense when the value is used by the engine where a feature otherwise can't be changed without recompiling.

EPuncker:
> I don't see the purpose of moving `sexchangecost` to config.lua
> 
> * The required and most frequently changed configuration should be in config.lua
> * How many people are changing the premium time cost of sex change? I suspect not many, at least in my OT travels I rarely see sex change being used.
> * Changing it with the current (pre PR) method is still a one line change in an aptly named file
> * Adding to config.lua values makes the most sense when the value is used by the engine where a feature otherwise can't be changed without recompiling.

the same could be said of the remove charge from potions config 🤔 I just thought it would be better to have all configs in one place instead of needing to have to modify several files, I was thinking about moving the buy premy cost to config lua as well but you discouraged me

omarcopires:
I believe it would be better to have a configuration file separate from config.lua, where all these configurations would be available.

But anyway the code is working.

DSpeichert:
Basically, you've added part of a script that is purely Lua to now have a dependency in the engine (C++) in the form of the config value.

I don't think that is necessary.

If anything, we could have a common config for scripts that the engine doesn't care about, just like @omarcopires suggested.

EPuncker:
yes I agree, I'll let @omarcopires in charge of it! it is a nice idea

PR #2906, opened 2020-04-07
Update Container.createLootItem ( item.maxCount increase )

MillhioreBT:
With these changes you can increase the maximum number of stackable objects, for example:
image

Other info: If the maximum number is greater than 100, take measures so that this does not affect the standard behavior, if for example the number was 200, the calculation is divided into two, so that it is treated as two items of 100 and 100, the same for the other major cases.

This is a request from the PR: https://github.com/otland/forgottenserver/pull/2875

### DIFF
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -23,7 +23,7 @@ function Container.createLootItem(self, item)
 			end
 			if stacks > 1 then
 				randvalue = getLootRandom()
-				if randvalue < item.chance then
+				if randvalue >= item.chance then
 					i = i + 1
 				end
 			end
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -2,9 +2,9 @@ function Container.isContainer(self)
 	return true
 end
 
-function Container:createLootItem(item, player, bonus)
+function Container.createLootItem(self, item)
 	if self:getEmptySlots() == 0 then
-        return true
+		return true
 	end
 
 	local itemCount = 0
@@ -43,7 +43,7 @@ function Container:createLootItem(item, player, bonus)
 			if item.actionId ~= -1 then
 				itm:setActionId(item.actionId)
 			end
-			
+
 			if item.text and item.text ~= "" then
 				itm:setAttribute(ITEM_ATTRIBUTE_TEXT, item.text)
 			end
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -2,71 +2,53 @@ function Container.isContainer(self)
 	return true
 end
 
-function Container.createLootItem(self, item)
+function Container:createLootItem(item, player, bonus)
 	if self:getEmptySlots() == 0 then
-		return true
+        return true
 	end
 
 	local itemCount = 0
-	local randvalue = getLootRandom()
-	if randvalue < item.chance then
-		local maxCount = math.min(item.maxCount, 100)
-		local isStackable = ItemType(item.itemId):isStackable()
+	local maxCount = math.min(item.maxCount, 100)
+	if ItemType(item.itemId):isStackable() then
 		local stacks = math.ceil(item.maxCount / 100)
 		for i = 1, stacks do
-			if isStackable then
+			local randvalue = getLootRandom()
+			if randvalue < item.chance then
 				itemCount = itemCount + (randvalue % maxCount + 1)
 				item.maxCount = item.maxCount - maxCount
 				maxCount = math.min(item.maxCount, 100)
-			else
-				itemCount = itemCount + 1
-			end
-			if stacks > 1 then
-				randvalue = getLootRandom()
-				if randvalue >= item.chance then
-					i = i + 1
-				end
 			end
 		end
+	elseif getLootRandom(player, bonus) < item.chance then
+		itemCount = 1
 	end
 
-	if itemCount > 0 then
-		local tmpItem = {}
-		repeat
-			local addCount = math.min(itemCount, 100)
-			local itm = self:addItem(item.itemId, addCount)
-			if itm then
-				tmpItem[#tmpItem + 1] = itm
-			end
-			itemCount = itemCount - addCount
-		until itemCount < 1
-		
-		if #tmpItem == 0 then
-			return false
-		end
-
-		for index, tmp in pairs(tmpItem) do
-			if tmp:isContainer() then
+	repeat
+		local addCount = math.min(itemCount, 100)
+		local itm = self:addItem(item.itemId, addCount)
+		if itm then
+			if itm:isContainer() then
 				for i = 1, #item.childLoot do
-					if not tmp:createLootItem(item.childLoot[i]) then
-						tmp:remove()
+					if not itm:createLootItem(item.childLoot[i]) then
+						itm:remove()
 						return false
 					end
 				end
 			end
 
 			if item.subType ~= -1 then
-				tmp:setAttribute(ITEM_ATTRIBUTE_CHARGES, item.subType)
+				itm:setAttribute(ITEM_ATTRIBUTE_CHARGES, item.subType)
 			end
 
 			if item.actionId ~= -1 then
-				tmp:setActionId(item.actionId)
+				itm:setActionId(item.actionId)
 			end
-
+			
 			if item.text and item.text ~= "" then
-				tmp:setText(item.text)
+				itm:setAttribute(ITEM_ATTRIBUTE_TEXT, item.text)
 			end
 		end
-	end
+		itemCount = itemCount - addCount
+	until itemCount < 1
 	return true
 end
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -10,38 +10,55 @@ function Container.createLootItem(self, item)
 	local itemCount = 0
 	local randvalue = getLootRandom()
 	if randvalue < item.chance then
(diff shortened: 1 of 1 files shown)


EPuncker:
> if for example the number was 200, the calculation is divided into two

wow, that is even better

EvilHero90:
You are still missing the random roll on each stack itself.
This works exactly as @MillhioreBT suggested with maxCount > 100
I've made this function and so far it works fine, feel free to test and adjust yours.
```lua
function Container.createLootItem(self, item)
		return true
	end

	local maxCount = item.maxCount
	local items = {}
	if ItemType(item.itemId):isStackable() then
		if maxCount > 100 then
			local stacks = math.ceil(maxCount / 100)
			for i = 1, stacks do
				randvalue = getLootRandom()
				if maxCount >= 100 then
					maxCount = maxCount - 100
						table.insert(items, randvalue % 100 + 1)
					end
				else
						table.insert(items, randvalue % maxCount + 1)
					end
				end
			end
		else
				table.insert(items, randvalue % maxCount + 1)
			end
		end
	else
			items[1] = 1
		end
	end

	for _, count in pairs(items) do
		local tmpItem = self:addItem(item.itemId, count)
		if not tmpItem then
			return false
		end

		if tmpItem:isContainer() then
			for i = 1, #item.childLoot do
				if not tmpItem:createLootItem(item.childLoot[i]) then
					tmpItem:remove()
					return false
				end
			end
		end

		if item.subType ~= -1 then
			tmpItem:setAttribute(ITEM_ATTRIBUTE_CHARGES, item.subType)
		end

		if item.actionId ~= -1 then
			tmpItem:setActionId(item.actionId)
		end

		if item.text and item.text ~= "" then
			tmpItem:setText(item.text)
		end
	end
	return true
end
```
sarah6

MillhioreBT:
Obviously I forgot, thanks for reminding me, I will modify it so now if that is perfect.
image

soul4soul:
I like the previous PR #2875  better. This code is more complicated for no perceivable value. This PR over the previous PR only effects the distribution model. The community doesn't have a good grasp on cipbia distribution models. I haven't seen any research to indicate if it's an even distribution model, a bell curve, or any other time of model. For that reason I prefer the previous PR which is simpler even if only slightly.

Which ever PR is merged +1 for the work. It is convenient to be able to set maxcount > 100.

EvilHero90:
> 
> 
> I like the previous PR #2875 better. This code is more complicated for no perceivable value. This PR over the previous PR only effects the distribution model. The community doesn't have a good grasp on cipbia distribution models. I haven't seen any research to indicate if it's an even distribution model, a bell curve, or any other time of model. For that reason I prefer the previous PR which is simpler even if only slightly.
> 
> Whichever PR is merged +1 for the work. It is convenient to be able to set maxcount > 100.

actually that is not correct, the other code had the problem that if you set maxcount to let's say 800 and by chance the roll was to low it wouldn't give any gold at all which is not correct imo, each stack should be random rolled on his own and that is what this code exactly does.

ranisalt:
Please note that multiple rolls of dice is not the same distribution as a single roll of a big die. That is, rolling a 10-sided die 10 times is not the same as rolling a 100-sided die once. Same applies for taking uniform random values.

MillhioreBT:
> Please note that multiple rolls of dice is not the same distribution as a single roll of a big die. That is, rolling a 10-sided die 10 times is not the same as rolling a 100-sided die once. Same applies for taking uniform random values.
> 
> Edit: now I see this was taken into account in the previous PR even.

Good way of thinking, it is also a good option, we should vote for the correct one.

EPuncker:
maybe a new config.lua setting so the user would chose if he want the chance to be divided by each stack of 100 or if he want the chance to be applied to the total value?

soul4soul:
> 
> 
> maybe a new config.lua setting so the user would chose if he want the chance to be divided by each stack of 100 or if he want the chance to be applied to the total value?

If a user wants multiple stacks with the same drop chance or even unique drop chances they should add multiple instances of the item to the creature. 

Right now It's already possible to do what this PR is doing by adding the item multiple times
```
<item name="gold coin" countmax="100" chance="98000" />
<item name="gold coin" countmax="100" chance="98000" />
<item name="gold coin" countmax="29" chance="98000" />
```

It's possible to add the same item multiple times with difference chances even.
```
<item name="gold coin" countmax="100" chance="78000" />
<item name="gold coin" countmax="50" chance="88000" />
<item name="gold coin" countmax="25" chance="98000" />
```

Right now I **can't** do this and this PR **doesn't** add support for this. 
```
<item name="gold coin" countmax="225" chance="88000" />
```

This PR converts the above example to the following. So it's basically a short hand for a preexisting feature.
```
<item name="gold coin" countmax="100" chance="88000" />
<item name="gold coin" countmax="100" chance="88000" />
<item name="gold coin" countmax="25" chance="88000" />
```

The original PR adds something new since the entire max count is rolled once. Which is why I prefer the initial PR over this PR. Overall I doubt anyone will care about or notice this difference in behavior. So at the end of the day I'd be happy with either PR getting merged.

DSpeichert:
I see your points and I think the most flexible and simple way to orchestrate this is:

One roll (400 gold or nothing):
```xml
<item name="gold coin" countmax="400" chance="88000" />
```

Two rolls (two independent chances of 200 gold):
```xml
<item name="gold coin" countmax="200" chance="88000" />
<item name="gold coin" countmax="200" chance="88000" />
```

Simply, each entry in the XML file is a separate lottery, and each support any arbitrary count.

If #2875 provided that, I would merge it. I see, however, that PR is closed and branch is deleted, so let's use this PR to converge on a solution.

I am AGAINST adding a config.lua option for this. It's unnecessary if it can be simply expressed by adding multiple lines.

EvilHero90:
How about another approach, we could add a xml tag like this:
```xml
<item name="gold coin" countmax="400" chance="88000" roll="once/stack"/>
```
depending on the option it does either of both and default would be "once" if nothing different is set.
Which means it would roll the entire stack once and not roll per stack (we probably need better names for this)

DSpeichert:
@EvilHero90 I'm not convinced that adding a custom flag for this makes sense. It still limits you to roll-per-stack, which by stack it means a count of 100. So you still can't do this in one line:
```xml
<item name="gold coin" countmax="200" chance="88000" />
<item name="gold coin" countmax="200" chance="88000" />
```

EvilHero90:
> 
> 
> @EvilHero90 I'm not convinced that adding a custom flag for this makes sense. It still limits you to roll-per-stack, which by stack it means a count of 100. So you still can't do this in one line:
> 
> ```
> <item name="gold coin" countmax="200" chance="88000" />
> <item name="gold coin" countmax="200" chance="88000" />
> ```

I'm not really sure if you exactly got the idea which I meant.
"once" means it would roll one time for all 400 countmax.
"stack" means that it would roll equal to 100 count per stack.
That would actually not limit us, but instead gives us both possibilites.

DSpeichert:
@EvilHero90 That's exactly what I meant. You'd still need to use this:
```xml
<item name="gold coin" countmax="200" chance="88000"  roll="once"/>
<item name="gold coin" countmax="200" chance="88000" roll="once"/>
```
Because with `stack`, you'd actually get this:
```xml
<item name="gold coin" countmax="100" chance="88000"/>
<item name="gold coin" countmax="100" chance="88000"/>
<item name="gold coin" countmax="100" chance="88000"/>
<item name="gold coin" countmax="100" chance="88000"/>
```

So my point is that this is no better than the following using my "version" of the idea:
```xml
<item name="gold coin" countmax="200" chance="88000" />
<item name="gold coin" countmax="200" chance="88000" />
```

To reiterate: It uses `once` by default and to do "stack" you would have to go back to splitting it into multiple lines but you are free to modify the count of each "stack/roll" by using different `countmax` per line.

From backwards and forward compatibility, I don't think introducing a new XML attribute here makes much sense.

EPuncker:
okay almost a year later I've come to the conclusion this will do more bad than good doesn't matter how we look into it, so I vote for leaving it as is and adding something like this:

```
	if ((attr = node.attribute("countmax"))) {
		int32_t lootCountmax = pugi::cast<int32_t>(attr.value());
		if (lootCountmax > 100) {
			std::cout << "[Warning - Monsters::loadMonster] Invalid \"countmax\" "<< lootCountmax <<" used for loot, the max allowed value is 100. " << std::endl;
		}
		lootBlock.countmax = std::max<int32_t>(1, pugi::cast<int32_t>(attr.value()));
	} else {
		lootBlock.countmax = 1;
	}
```

DSpeichert:
@EPuncker I've approved #3284 but i still think that raising the limit >100 is possible provided that it remains a single lottery.
My subjective opinion is that each XML entry is a single lottery, therefore it's reasonable to assume that having >100 item is a stack, as I explained above and I'd love that to be "fixed". Right now it's capped at 100 silently, which may be doing more harm than good.

ranisalt:
Here's ~a Python script~ pseudocode to split a stack of X items into ceil(X / 100) groups of up to 100 items each:

```python
countmax = 780
n_stacks = ceil(countmax / 100)

stacks = []
for n in range(n_stacks, 0, -1):
    diff = n * 100 - countmax
    r = randint(100 - diff, 100)
    stacks.append(r)

    countmax = countmax - r
```

Hope you find it of good use!

MillhioreBT:
I think there is really no need to add this so I will close it until new ideas come.

PR #2907, opened 2020-04-08
Added juice squeezer

EPuncker:
Added script from juice squeezer added on **7.8**

https://tibia.fandom.com/wiki/Juice_Squeezer

created by cyckotitan [I guess](https://otland.net/threads/juice-squeezer.110243/#post-1098564)


data/items/items.xml
```diff
@@ -4600,7 +4600,7 @@
 	<item id="2680" article="a" name="strawberry" plural="strawberries">
 		<attribute key="weight" value="20" />
 	</item>
-	<item id="2681" name="grapes" plural="grapes">
+	<item id="2681" name="grapes">
```
DSpeichert, line 4603: isn't TFS going to add `s` and make it `grapess` for plural?

EPuncker: true 🤦

PR #2912, opened 2020-04-08
Added noxious claw

EPuncker:
- added its missing article in items.xml

Added on 8.5 update
https://tibia.fandom.com/wiki/Claw_of_%27The_Noxious_Spawn%27

part of script found in ORTS and part on otland


M4g1ch:
When using the item while equipped, I'm getting the following error in console many times:

> [Error - MoveEvent::executeEquip] Call stack overflow

After that, item gets frozen in the ring slot and causes similar errors when actions are performed on it

EPuncker:
> When using the item while equipped, I'm getting the following error in console many times:
> 
> > [Error - MoveEvent::executeEquip] Call stack overflow
> 
> After that, item gets frozen in the ring slot and causes similar errors when actions are performed on it

make sure you have applied these changes to your items.xml:
https://github.com/otland/forgottenserver/pull/2912/commits/414c0c234fdb865eb9df722f27b9fce943aa054f

M4g1ch:
Those lines are the same as in my items.xml :hushed:

EPuncker:
> Those lines are the same as in my items.xml 😯

I mean, you need to remove them 😄  or it will be an infinite loop that is why it crashes

M4g1ch:
Yes, this is how it was looking when I tested:
>         <item id="10309" article="a" name="claw of 'The Noxious Spawn'">
>                 <attribute key="description" value="It feels tight and uncomfortable on your hand." />
>                 <attribute key="weight" value="500" />
>                 <attribute key="slotType" value="ring" />
>         </item>
>         <item id="10310" article="a" name="claw of 'The Noxious Spawn'">
>                 <attribute key="description" value="It's like a glove and you can't wear a ring on it. You are not sure if you should really put it on." />
>                 <attribute key="weight" value="500" />
>                 <attribute key="slotType" value="ring" />
>         </item>
I tested again now (just in case) and still the same

EPuncker:
> Yes, this is how it was looking when I tested:
> 
> > ```
> >     <item id="10309" article="a" name="claw of 'The Noxious Spawn'">
> >             <attribute key="description" value="It feels tight and uncomfortable on your hand." />
> >             <attribute key="weight" value="500" />
> >             <attribute key="slotType" value="ring" />
> >     </item>
> >     <item id="10310" article="a" name="claw of 'The Noxious Spawn'">
> >             <attribute key="description" value="It's like a glove and you can't wear a ring on it. You are not sure if you should really put it on." />
> >             <attribute key="weight" value="500" />
> >             <attribute key="slotType" value="ring" />
> >     </item>
> > ```
> 
> I tested again now (just in case) and still the same

thanks, I've made a few updates if you can test it again

PR #2915, opened 2020-04-08
fix #2909

### LINKED ISSUE #2909: Duplicated registered ID in revscriptsys
Duplicated registered ID in revscriptsys gives a C++ handler exception error instead of warning the user that he have two (scripts) registering the same ID

to reproduce just make a duplicated script of any sort, for example:

```
local test = Action()

function test.onUse(player, item, fromPosition, target, toPosition, isHotkey)
	return true
end

test:id(13941)
test:register()
```
```
local test2 = Action()

function test2.onUse(player, item, fromPosition, target, toPosition, isHotkey)
	return true
end

test2:id(13941)
test2:register()
```
I'd like to tag @EvilHero90 just in case 😋

EvilHero90:
Now the correct duplicate message appears.

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -197,7 +197,7 @@ bool Actions::registerLuaEvent(Action* event)
 	Action_ptr action{ event };
 	if (action->getItemIdRange().size() > 0) {
 		if (action->getItemIdRange().size() == 1) {
-			auto id = action->getItemIdRange().at(0);
+			auto id = action->getItemIdRange().front();
 			auto result = useItemMap.emplace(id, std::move(*action));
 			if (!result.second) {
 				std::cout << "[Warning - Actions::registerLuaEvent] Duplicate registered item with id: " << id << std::endl;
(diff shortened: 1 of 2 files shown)


EvilHero90:
It was trying to get a value from a unique_pointer which was already freed that's why it threw an c++ exception, this works correctly now.

src/actions.cpp
```diff
@@ -197,9 +197,10 @@ bool Actions::registerLuaEvent(Action* event)
 	Action_ptr action{ event };
 	if (action->getItemIdRange().size() > 0) {
 		if (action->getItemIdRange().size() == 1) {
-			auto result = useItemMap.emplace(action->getItemIdRange().at(0), std::move(*action));
+			auto id = action->getItemIdRange().at(0);
```
ranisalt, line 200: `front()`?

src/actions.cpp
```diff
@@ -215,9 +216,10 @@ bool Actions::registerLuaEvent(Action* event)
 		}
 	} else if (action->getUniqueIdRange().size() > 0) {
 		if (action->getUniqueIdRange().size() == 1) {
+			auto uid = action->getUniqueIdRange().at(0);
 			auto result = uniqueItemMap.emplace(action->getUniqueIdRange().at(0), std::move(*action));
```
ranisalt, line 220: Use `uid`

src/actions.cpp
```diff
@@ -233,9 +235,10 @@ bool Actions::registerLuaEvent(Action* event)
 		}
 	} else if (action->getActionIdRange().size() > 0) {
 		if (action->getActionIdRange().size() == 1) {
+			auto aid = action->getActionIdRange().at(0);
 			auto result = actionItemMap.emplace(action->getActionIdRange().at(0), std::move(*action));
```
ranisalt, line 239: Use `aid`

src/actions.cpp
ranisalt, line 220: Use `uid`

src/actions.cpp
ranisalt, line 239: Use `aid`

src/actions.cpp
EvilHero90: completly overlooked that, thanks :D

PR #2921, opened 2020-04-10
Added Ursagrodon taming script

EPuncker:
- Added the decay at the items.xml
- Added the tinder box script to create melting horn
- Added the taming script to unfrozen the creature

it was added on 10.37 version
scripts by me
tested and working


raymondtfr:
You could add named constants e.g. `ITEM_MELTING_HORN`, `ITEM_FIREPROOF_HORN`, etc, to `data/lib/core/constants` for each item you're using in these scripts, it would improve readability a lot and you wouldn't need to add comments for them.

luanluciano93:
https://ghostbin.co/paste/nyej7/raw

soul4soul:
Minor nit. Change the name of the file to `melting_horn.lua` or `taming_melting_horn.lua`. Many of the action scripts are named by the object that has the action. For example there is a script called `machete.lua` not `tall_grass.lua`. To match that example the script would be named `melting_horn.lua`. However there could be a bunch taming scripts (I saw your other PR 👍) so it might be nice for them to group together in the folder. In that case `taming_melting_horn.lua` would be the way to go.

EPuncker:
> Minor nit. Change the name of the file to `melting_horn.lua` or `taming_melting_horn.lua`. Many of the action scripts are named by the object that has the action. For example there is a script called `machete.lua` not `tall_grass.lua`. To match that example the script would be named `melting_horn.lua`. However there could be a bunch taming scripts (I saw your other PR 👍) so it might be nice for them to group together in the folder. In that case `taming_melting_horn.lua` would be the way to go.

I agree with the rename, but please no more sub folders, you opened my eyes a while ago about the useless amount of monsters sub folder, let not do the same with actions, A-Z is the way to go 😋

DSpeichert:
Instead of doing the constants as suggested by @raymondtfr (this would essentially create second items.xml), I'd rather have comments in Lua with the name of the item next to it.

PR #2930, opened 2020-04-10
PVP config values update

nekiro:
Follow up to #2920 
White skull ticks are actually in ms, but we don't need ms precision in our config.lua anyway, so we have to multiply it by 1000 here. 
I thought that we should have all these consistent so I changed pzLocked to seconds in config.lua too.


gpedro:
i disagree with that. will cause a breaking change.

EPuncker:
how on the earth is changing 60000 to 60 * 1000 a breaking change? 🤷

soul4soul:
> 
> 
> how on the earth is changing 60000 to 60 * 1000 a breaking change? 🤷

It's breaking because it requires existing users to update their `configu.lua` files. Anyone who fails to do so will have a huge pzlock time. I do like the change but I can see why the devs might reject it.

nekiro:
@soul4soul we had broken frag ticks since I dont even know when, config value was in ms instead of seconds, so I dont think there is a problem with that.

gpedro:
I'm thinking about all the people were are using this server as a base. cuz some ppl use this distribution, made your customizations but always keep synced with the TFS repository. this will cause a lot of boring problems.

well, it's my personal opinion as a developer and "user".

nekiro:
@gpedro no, you are getting too serious about a stupid config value, thats not something that backwards incompatible. You can literally change that in matter of seconds. Making another config value to work the same way is just redundant code and a mess and you should really read my message before yours. The best we can do is probably put a comment above these values to remind users its in seconds not any other format.

Leo32onGIT:
We've been pretty good with backwards compat up to now.
How about a new variable `pzLockedMs`, if that variables doesnt exist refer back to `pzLocked` / 1000

Xaekai:
Are you sure there will never be anything in config.lua that will require 500ms or better resolution? It seems like it would be easier to just **include commentary schema in the file**. Like every Unix /etc/conf for the past 40 years.

I'm actually curious about you are truncating the resolution to 1000ms in a game engine? Because if there is every a for something in the config that does require ms precision, you'll have created an inconsistency of time value types.

What problem does this solve?

DSpeichert:
Config should definitely have comments providing the units to use.

On the related change, maybe we should revert 81c54baa64c477b9e1b085d9c3ea1a974813f0b5, fix the bug (actually not multiply by 1000 for those values in engine) and make it clear everything is in miliseconds.

nekiro:
@DSpeichert I would prefer to have these in seconds precision, because we don’t need miliseconds anyway, but as you wish.

DSpeichert:
@nekiro I agree it's unlikely anybody would need milisecond precision.
However, backwards-compatibility is important and degrading precision here doesn't really provide much value to necessitate the change. I'd rather do the minimum amount of changes to keep it consistent. That looks to me like making it all miliseconds.

Riverlance:
### _The timeline_

**1**. `"Fix skull ticks overflow"` (which fixes #2323 but generated the issue mentioned in `step 3`): https://github.com/otland/forgottenserver/commit/a62389ae00b49c5de2664c596d434fb9e68b429a
**2**. `"Fix wrong variable format (time ms -> s)"`: https://github.com/otland/forgottenserver/commit/81c54baa64c477b9e1b085d9c3ea1a974813f0b5
**3**. I figured out this (_skulltime issue_):
image
**4**. Then, @nekiro made this actual `pull request`.

---

## Issue fix - Option A (I think this is the best one)
**1**. Consider the `step 3` of the mentioned _timeline_ above.

Why I think this is the best choice?
* Since these changes in skulltime started at `step 1` as mentioned at _timeline_ above ( https://github.com/otland/forgottenserver/commit/a62389ae00b49c5de2664c596d434fb9e68b429a ), I think you guys **don't need to accept** all the code of this pull request. **But**, by considering the fix in `step 3` of _timeline_ above (just **1 line** to fix), should solve everything **without needed changes in database or _config.lua_**.

---

## Issue fix - Option B
If you guys wants to revert the skulltime to milliseconds, you need to revert these commits (in following order):
**1**. `"Fix wrong variable format (time ms -> s)"`: https://github.com/otland/forgottenserver/commit/81c54baa64c477b9e1b085d9c3ea1a974813f0b5
**2**. `"Fix skull ticks overflow"` (which fixes #2323): https://github.com/otland/forgottenserver/commit/a62389ae00b49c5de2664c596d434fb9e68b429a

**Important**: It will need people to update their `skulltime` in `database` and update their `config.lua` with the reversion changes.

---

@nekiro @Xaekai @DSpeichert

EPuncker:
hey guys, consider [opening an issue](https://github.com/otland/forgottenserver/issues/new/choose) for a better discussion 😛  since this branch has been deleted and can't be reopened

PR #2932, opened 2020-04-11
Added Lua VoiceModule to NPCs

EPuncker:
the XML tag sucks and we all know it does, this module is great, by ORTS

also added an example on the runes npc

### DIFF
--- a/data/npc/lib/npcsystem/modules.lua
+++ b/data/npc/lib/npcsystem/modules.lua
@@ -1203,4 +1203,51 @@ if Modules == nil then
 		end
 		return true
 	end
+
+	VoiceModule = {
+		voices = nil,
+		voiceCount = 0,
+		lastVoice = 0,
+		timeout = nil,
+		chance = nil,
+		npcHandler = nil
+	}
+
+	-- VoiceModule: makes the NPC says/yell random lines from a table, with delay, chance and yell optional 
+	function VoiceModule:new(voices, timeout, chance)
+		local obj = {}
+		setmetatable(obj, self)
+		self.__index = self
+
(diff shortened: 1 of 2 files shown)


data/npc/lib/npcsystem/modules.lua
```diff
@@ -1203,4 +1203,51 @@ if Modules == nil then
...
 		return true
 	end
+
+	VoiceModule = {
+		voices = nil,
+		voiceCount = 0,
+		lastVoice = 0,
+		timeout = nil,
+		chance = nil,
+		npcHandler = nil
```
TheSumm, line 1213: This is not necessary unless you want to assign it in `init` and use the handler in `callbackOnThink` instead of the npc constructor `Npc()`

EPuncker: so should I remove entire 1207 to 1214? or just 1213?

TheSumm: I assume 1213 only, I think it got there due to copy-and-pasting.

data/npc/lib/npcsystem/modules.lua
```diff
@@ -1203,4 +1203,50 @@ if Modules == nil then
...
+
+	VoiceModule = {
+		voices = nil,
+		voiceCount = 0,
+		lastVoice = 0,
+		timeout = nil,
+		chance = nil
+	}
+
+	-- VoiceModule: makes the NPC says/yell random lines from a table, with delay, chance and yell optional
```
DSpeichert, line 1215: `says/yell` => `say/yell`

PR #2933, opened 2020-04-11
Add option CleanProtectedZones to config.lua.dist

MillhioreBT:
Add the functionality of activate or deactivate, if you want the map to be completely cleaned or to protect the articles in protected zones.

this can be very good for custom servers that use this engine as a base.


config.lua.dist
```diff
@@ -83,6 +83,7 @@ showOnlineStatusInCharlist = false
 yellMinimumLevel = 2
 yellAlwaysAllowPremium = false
 forceMonsterTypesOnLoad = true
+cleanProtectedZones = false
```
DSpeichert, line 86: it's called "protection zone", so this should probably be worded `cleanProtectionZones`.

Lyuzera:
Nice idea, but this clean can be improved.
So, why #2647 is not merged? I tested it with 200+ players online and worked very well.

config.lua.dist
DSpeichert: This is still not updated.

nekiro:
We have that config added in #2647 so this pull request is redundant.

PR #2936, opened 2020-04-12
Fix isPersistent condition check

danyelvarejao1:
Fixed - When a player dies, the condition muted is removed.
Fixed - When a player received conditions such as outfit, drunk, etc. All will be removed if he relogged.

It doesn't make sense for the function name to be isPersistent which to me means "persists" and after dead/logout the condition to be removed.

### DIFF
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -266,7 +266,7 @@ bool Condition::isPersistent() const
 		return true;
 	}
 
-	if (id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT) {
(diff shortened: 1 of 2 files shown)


DSpeichert:
This needs further review and testing.
I'm not convinced that conditions like "muted" and "drunk" should be removed when you relog.

ranisalt:
Muted, definitely not. Otherwise muting is useless as the player can just relog.

However you can't logout when drunk (IIRC) so whatever.

danyelvarejao1:
Well, for me a function given the name Persistent("infinite") would have to have the result of conditions that do not come out after death/relogin

Muted condition in my opinion was to be until finished, regardless of whether the player dies/logs out.

PR #2937, opened 2020-04-12
Added script to Gnomish Voucher items

EPuncker:
They were added on 9.6 version

- script by ORTS
- it was a single script for both outfits but it looked better to be split in two in case anyone want to disable one and keep the other one and vice-versa

https://tibia.fandom.com/wiki/Updates/9.6#New_Outfits


data/scripts/actions/others/crystal_warlord_outfit.lua
```diff
@@ -0,0 +1,43 @@
+local config = {
+	[18520] = {female = 513, male = 512, effect = CONST_ME_GIANTICE}, -- gnomish voucher type CB
+	[18521] = {female = 513, male = 512, addon = 1, effect = CONST_ME_GIANTICE}, -- gnomish voucher type CA1
+	[18522] = {female = 513, male = 512, addon = 2, effect = CONST_ME_GIANTICE} -- gnomish voucher type CA2
+}
+
+local crystalWarlord = Action()
```
Zbizu, line 7: This needs consistency. We don't want hidden scripts active in the datapack.

either do normal function onUse and register in actions.xml or drop it to data/scripts folder where other revscripts are

EPuncker: they are not hidden lol tfs 1.x users should expect scripts to be both in old locations and in revscript location, and I'm trying to be consistent since nearly everything I'm doing is being done in the scripts folder...

as for moving everything to scripts, #2663 

or I just missed your point

soul4soul:
These scripts are the same thanks to the use of the config array. I suggest merging to a single file named `gnomish_voucher_types.lua`. In the combined script the variables named `crystalWarlord` and `soilGuardian` can be replaced with `gnomishVoucher`. Lines 11 to 13 are not necessary. The valid itemids from the config array are directly registered for the action. It's not possible for this to be triggered with an invalid item id.

+1 EPuncker

Zbizu:
actions.xml -> script -> function onUse
scripts -> revscripts way

I'm not saying that the scripts are different. I'm just saying that scripts should belong in respective folders and mods should be mods. When the script is registered from lua we can easily call it a mod.

data/scripts/actions/others/crystal_warlord_outfit.lua
Zbizu: let's wait for official decision regarding ditching xml file in favour of revscriptsys

personally I'm against it because it's really good to keep track to which action/unique/item ids are tied to which scripts without having to open every script in your datapack to tell that

DSpeichert:
@EPuncker let me know if you want to change it how @soul4soul explained or leave it as-is.

EPuncker:
> @EPuncker let me know if you want to change it how @soul4soul explained or leave it as-is.

done

PR #2938, opened 2020-04-12
Added nail case script

EPuncker:
Item used to tame the monster "gravedigger", both added on 10.1 update

https://tibia.fandom.com/wiki/The_Hellgrip

script made by me with the help of k.k on otland

tested and working


BahamutxD:
Working just fine. Ignore the green artifact at the end - thats just an issue with compression.

https://gfycat.com/rewardingfrighteningeidolonhelvum

PR #2939, opened 2020-04-12
Added Slippery Northern Pike monster

EPuncker:
fixed northern pike flags, corpse, race, speed, also added the summoning of the slippery one

they were added on 9.4 version
https://tibia.fandom.com/wiki/Northern_Pike_(Creature)

### DIFF
--- a/data/scripts/creaturescripts/northern_pike_death.lua
+++ b/data/scripts/creaturescripts/northern_pike_death.lua
@@ -1,0 +1,10 @@
+local creatureevent = CreatureEvent("NorthernPikeDeath")
+
+function creatureevent.onDeath(creature, corpse, lasthitkiller, mostdamagekiller, lasthitunjustified, mostdamageunjustified)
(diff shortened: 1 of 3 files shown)


data/scripts/creaturescripts/northern_pike_death.lua
```diff
@@ -0,0 +1,10 @@
+local creatureevent = CreatureEvent("NorthernPikeDeath")
+
+function creatureevent.onDeath(creature, corpse, lasthitkiller, mostdamagekiller, lasthitunjustified, mostdamageunjustified)
+	if math.random(100) < 11 then
+		Game.createMonster("Slippery Northern Pike", creature:getPosition())
```
nekiro, line 5: I think you could put there force parameter so it always summons that monster in place.
so false, true after getPosition()

BahamutxD:
Working just fine.

https://gfycat.com/sparseverifiablehorsefly

PR #2942, opened 2020-04-15
Store player storages and learned spells in binary

danyelvarejao1:
Storages and learned spells are now blobbed in the database, improving performance.
Correction of the isPersistent condition function

### DIFF
--- a/data/migrations/25.lua
+++ b/data/migrations/25.lua
@@ -1,3 +1,8 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 26 (store player storages and learned spells in binary)")
+	db.query("DROP TABLE `player_storage`;")
+	db.query("DROP TABLE `player_spells`;")
+	db.query("ALTER TABLE `players` ADD `spells` blob DEFAULT NULL AFTER `conditions`;")
+	db.query("ALTER TABLE `players` ADD `storages` mediumblob DEFAULT NULL AFTER `spells`;")
+	return true
 end
--- a/data/migrations/26.lua
+++ b/data/migrations/26.lua
@@ -1,0 +1,3 @@
+function onUpdateDatabase()
+	return false
+end
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -266,7 +266,7 @@ bool Condition::isPersistent() const
 		return true;
 	}
 
-	if (id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT || conditionType == CONDITION_MUTED) {
+	if (conditionType == CONDITION_MUTED) {
 		return true;
 	}
 
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -266,7 +266,7 @@ bool Condition::isPersistent() const
 		return true;
 	}
 
-	if (id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT) {
+	if (id == CONDITIONID_DEFAULT || id == CONDITIONID_COMBAT || conditionType == CONDITION_MUTED) {
 		return true;
 	}
 
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -263,14 +263,14 @@ bool Condition::startCondition(Creature*)
 bool Condition::isPersistent() const
 {
 	if (ticks == -1) {
-		return false;
+		return true;
 	}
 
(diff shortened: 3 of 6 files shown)


EPuncker:
how do I query a storage value after this?
for example, right now I can do this:

`local resultId = db.storeQuery("SELECT `player_id` FROM `player_storage` WHERE `key` = " .. storage .. " AND `value` >= 3;")`

how would I do after your change? just a question 😛

Lyuzera:
> how do I query a storage value after this?
> for example, right now I can do this:
> 
> `local resultId = db.storeQuery("SELECT `player_idplayer_storagekeyvalue`FROM``WHERE``= " .. storage .. " AND`` >= 3;")`
> 
> how would I do after your change? just a question 😛

inside the tfs or website? at tfs you can get normally with `player:getStorageValue`. I prefer binary because it is much faster than the generic estructure, but to show storages on website is not cool.

EPuncker:
> > how do I query a storage value after this?
> > for example, right now I can do this:
> > `local resultId = db.storeQuery("SELECT `player_idplayer_storagekeyvalue`` FROM``WHERE``= " .. storage .. " AND`` >= 3;") ``
> > how would I do after your change? just a question 😛
> 
> inside the tfs or website? at tfs you can get normally with `player:getStorageValue`. I prefer binary because it is much faster than the generic estructure, but to show storages on website is not cool.

not website, but getting storage of an offline player

Lyuzera:
> not website, but getting storage of an offline player

good question, we lost that possibility, because unserialize is bad for get a simple information. But for me, that possibility can be deprecated.. so why?

```lua
-- we can store anything by player guid and get it from all datapack interfaces (thx lua state)
--  simpleCache[player:getGuid()] = {}

-- if you created a big system, you can use cache lua + database
-- storage is very well to quests, but we no need to get infomations when player is offline.
```

data/migrations/25.lua
```diff
@@ -1,3 +1,8 @@
 function onUpdateDatabase()
-	return false
-end
+	print("> Updating database to version 26 (store player storages and learned spells in binary)")
+	db.query("DROP TABLE `player_storage`;")
+	db.query("DROP TABLE `player_spells`;")
+	db.query("ALTER TABLE `players` ADD `spells` blob DEFAULT NULL AFTER `conditions`;")
+	db.query("ALTER TABLE `players` ADD `storages` mediumblob DEFAULT NULL AFTER `spells`;")
```
DSpeichert, line 6: So if someone upgrades to this version all their spells/storage is... gone.

EPuncker:
> > not website, but getting storage of an offline player
> 
> good question, we lost that possibility, because unserialize is bad for get a simple information. But for me, that possibility can be deprecated.. so why?
> 
> ```lua
> -- we can store anything by player guid and get it from all datapack interfaces (thx lua state)
> --  simpleCache[player:getGuid()] = {}
> 
> -- if you created a big system, you can use cache lua + database
> -- storage is very well to quests, but we no need to get infomations when player is offline.
> ```

okay, I was only thinking about small things, like achievements displays, hunted lists, custom points rankings (anything that deals with values of offline players) will need to have a sql table I think... 😢  sacrificing a functionality for performance 🤔 i hope devs find a way

nekiro:
I dont think we need that type of optimization on storages, they are fine and does not need to be optimized. The real optimization is needed in items storing.

CristoferMartins:
Storages never been a performance problem even in servers which uses a lot of them. Items storage(specially inbox) is the problem. And btw the ideal fix is to load/save players in another thread in an async fashion for the dispatcher.

danyelvarejao1:
@gesior optimized the saving of items/depot items/inbox items in binary and it is perfect, it has been 100% tested on production servers and the PR of it has never been approved.

Lyuzera:
The @gesior's binary PR was really 100% tested and approved. I've never seen anything as fast as binary items, 150k items and no lag during the login.. and it is faster than cache in RAM.

DSpeichert:
This PR causes data loss and as such has no change of being merged. Let me know if you intend to work on that anymore or if you'd rather close it.

danyelvarejao1:
I see no reason for so much "data loss", in most cases who is using the repository is who is creating/developing your server. Whoever has an online server and with many players does not care for the current repository, you sometimes do not give a damn to those who are updating/optimizing codes that are horrible in TFS just because there is a "data loss", I am referring to @gesior code.
You can actually close the pull request, I also agree that storages should not be changed to binary, but think about it, value people who want to contribute to the repository and improve various things.

PR #2948, opened 2020-04-19
Added editor suffix to locked doors

EPuncker:
Made it a little more clear on RME, to prevent more doubt like [these](https://otland.net/threads/problem-with-doors-in-house.269971/)


EPuncker:
> I didn't know such parameter even exists.

the weird thing is that TFS doesn't throw a warning for non existent items tag 🤣

Issue #1193, opened 2015-02-11
Request: getParameter for combat/condition


Codinablack:
Ok so I seen a post today from someone on otland about doing an action when the ticks of a condition have worn off, and yes there are rudimentary ways of doing this sure, but I would like to request a method for both combat and condition that will return the value of a specified parameter.  I am sure everyone understands what I mean but for clarification here is an example.

condition:getParameter (CONDITION_PARAM_TICKS)

Return value would of course be the amount of ticks left for that condition. Please note, I may have the variable for ticks wrong, nothing to reference right this second because I am posting from my phone.
##

Kamenuvol:
I've seen that post too, and I think doCombat should return true when it is finished and you check:
if doCombat.. then

marmichalski:
This can be closed. @nekiro?

nekiro:
@marmichalski yep, I forgot to mark it properly in #2951

PR #2951, opened 2020-04-21
condition:getParameter and combat:getParameter methods

nekiro:
Adding requested methods in #1193


src/condition.cpp
```diff
@@ -691,6 +712,128 @@ bool ConditionAttributes::setParam(ConditionParam_t param, int32_t value)
...
+			return stats[STAT_MAXHITPOINTS];
+		}
+
+		case CONDITION_PARAM_STAT_MAXMANAPOINTS: {
+			return stats[STAT_MAXMANAPOINTS];
+		}
+
+		case CONDITION_PARAM_STAT_MAGICPOINTS: {
+			stats[STAT_MAGICPOINTS] = value;
+			return true;
```
maattch, line 789: a little mistake here

nekiro: thanks!

DSpeichert:
Sample script would be useful for both testing and showoff/guide how to use it.

src/condition.cpp
```diff
@@ -691,6 +708,100 @@ bool ConditionAttributes::setParam(ConditionParam_t param, int32_t value)
...
+			return specialSkills[SPECIALSKILL_LIFELEECHAMOUNT];
+
+		case CONDITION_PARAM_SPECIALSKILL_MANALEECHCHANCE:
+			return specialSkills[SPECIALSKILL_MANALEECHCHANCE];
+
+		case CONDITION_PARAM_SPECIALSKILL_MANALEECHAMOUNT:
+			return specialSkills[SPECIALSKILL_MANALEECHAMOUNT];
+
+		default:
+			return value;
```
ranisalt, line 801: You should move the initial check here instead:

```
default:
    return ConditionGeneric::getParam(param);
```
So the branching is unnecessary and the result is the same.

ranisalt (changes requested):
My comment applies to all the functions.

src/condition.cpp
nekiro: True, not sure why I did it this way. It was few months ago xD

nekiro: Long time since we talked about this, but here we go. It should be more reasonable now.

src/condition.cpp
```diff
@@ -952,6 +1092,41 @@ bool ConditionDamage::setParam(ConditionParam_t param, int32_t value)
...
+			return tickInterval;
+
+		case CONDITION_PARAM_PERIODICDAMAGE:
+			return periodDamage;
+
+		case CONDITION_PARAM_FIELD:
+			return field ? 1 : 0;
+
+		default:
+			return Condition::getParam(param);;
```
ranisalt, line 1126: On an otherwise spotless pull request
```suggestion
			return Condition::getParam(param);
```

src/condition.cpp
```diff
@@ -1334,6 +1509,17 @@ bool ConditionSpeed::setParam(ConditionParam_t param, int32_t value)
...
 }
 
+int32_t ConditionSpeed::getParam(ConditionParam_t param)
+{
+	switch (param) {
+		case CONDITION_PARAM_SPEED:
+			return speedDelta;
+
+		default:
+			return Condition::getParam(param);;
```
ranisalt, line 1519:
```suggestion
			return Condition::getParam(param);
```

src/condition.cpp
```diff
@@ -1590,6 +1776,20 @@ bool ConditionLight::setParam(ConditionParam_t param, int32_t value)
...
+{
+	switch (param) {
+		case CONDITION_PARAM_LIGHT_LEVEL:
+			return lightInfo.level;
+
+		case CONDITION_PARAM_LIGHT_COLOR:
+			return lightInfo.color;
+
+		default:
+			return Condition::getParam(param);;
```
ranisalt, line 1789:
```suggestion
			return Condition::getParam(param);
```

ranisalt (changes requested):
I see you have been using Python too much @nekiro that you really miss writing semicolons :laughing:

ranisalt (approved):

PR #2952, opened 2020-04-21
tile:remove method

### LINKED ISSUE #1522: Add tile:remove()
It's necessary to create huge random generated maps.

https://otland.net/threads/tfs-1-0-how-do-i-remove-a-tile.235513/#post-2273523

nekiro:
Resolve #1522

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -174,7 +174,11 @@ void Map::removeTile(uint16_t x, uint16_t y, uint8_t z)
 	if (tile) {
 		if (const CreatureVector* creatures = tile->getCreatures()) {
 			for (int32_t i = creatures->size(); --i >= 0;) {
-				g_game.removeCreature((*creatures)[i]);
+				if (Player* player = (*creatures)[i]->getPlayer()) {
+					g_game.internalTeleport(player, player->getTown()->getTemplePosition(), false, FLAG_NOLIMIT);
+				} else {
+					g_game.removeCreature((*creatures)[i]);
+				}
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


Lyuzera:
nice, I think that with this method we can improve some things like loadMapChunk to replace all of the map and not duplicate.

EvilHero90:
Aren't you forgetting to check for Creatures and remove them before removing the tile? (just quikcly looked over it)

nekiro:
@EvilHero90 indeed, thanks!

EvilHero90 (dismissed):
code wise this looks good now, havn't tested tho.

ninjalulz:
You might want to set `flags` to `0` but other than that, it looks fine to me.

nekiro:
@ninjalulz tile class will take care of that when removing creatures and then items.

DSpeichert:
Could use a sample script to showcase how it's used (or useful).

src/map.cpp
```diff
@@ -154,6 +154,44 @@ void Map::setTile(uint16_t x, uint16_t y, uint8_t z, Tile* newTile)
...
+	const Floor* floor = leaf->getFloor(z);
+	if (!floor) {
+		return;
+	}
+
+	Tile* tile = floor->tiles[x & FLOOR_MASK][y & FLOOR_MASK];
+	if (tile) {
+		if (const CreatureVector* creatures = tile->getCreatures()) {
+			for (int32_t i = creatures->size(); --i >= 0;) {
+				g_game.removeCreature((*creatures)[i]);
```
WibbenZ, line 177: Tested the code and works as it should, but not sure about this part.
What happens now is you can still see the ingame screen but your char is removed (just like when you use /r on a player)
Might be a good ide to kick the player and maybe even teleport it to temple, just tested to remove all the tiles around my char and got a client debug when I tried to log back in.

soul4soul: I like the idea of teleporting them to the temple in this odd case

WibbenZ (approved):
Removing merge block (not tested or read code)

Znote:
Updated wiki
https://github.com/otland/forgottenserver/wiki/Metatable:Tile

PR #2953, opened 2020-04-21
Fix Container::queryMaxCount() bug

### LINKED ISSUE #1561: Container::queryMaxCount() failing in special cases.
How to reproduce
Take any character with capacity (so no GM/God characters), open it's main backpack, and fill it's main backpack with whatever you want as long as one slot has a single gold coin in it and the rest of the slots are occupied. Do whatever else you have to to make sure the character has very little capacity left, but enough to carry at least 1 more gold coin.

Simply, you can:
Create a paladin of level 1008, clear its inventory, and give it's main backpack these contents: http://i.imgur.com/cz6jX2o.png

When you try to add another gold coin to that container from another source (e.g, ground) you will get a "not enough room" error even though there is room to stack and you have enough capacity. The return happens here: https://github.com/otland/forgottenserver/blob/master/src/container.cpp#L385

I haven't had time to pinpoint the exact issue, but I'm assuming it happens in the Cylinder->queryAdd() call in Container::queryAdd().

nekiro:
Fix #1561 

Explanation: current code is trying to query remainder which is for example 90 thus it calculates cap for 90x of that item not the real count that we add to backpack.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2256,7 +2256,7 @@ ReturnValue Player::queryAdd(int32_t index, const Thing& thing, uint32_t count, 
 	if (childIsOwner) {
 		//a child container is querying the player, just check if enough capacity
 		bool skipLimit = hasBitSet(FLAG_NOLIMIT, flags);
-		if (skipLimit || hasCapacity(item, count)) {
+		if (skipLimit || hasCapacity(item, item->getItemCount())) {
 			return RETURNVALUE_NOERROR;
 		}
 		return RETURNVALUE_NOTENOUGHCAPACITY;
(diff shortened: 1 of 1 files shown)


EPuncker:
tested and working, fixes the bug of not letting put the item inside the container, but it made me notice another bug, but let me fill a new issue to it

nekiro:
@EPuncker if this pull request caused the new issue, you should give the info here not in new issue, cause its not master code.

EPuncker:
> @EPuncker if this pull request caused the new issue, you should give the info here not in new issue, cause its not master code.

no, it is not a new bug, it is that I just noticed it now, don't worry

PR #2955, opened 2020-04-21
Auto stack fix

### LINKED ISSUE #2954: Player::queryAdd always picking first slot
how to reproduce:
- have a full container of items:
2020-04-21 18_48_41-CN=Microsoft Windows, O=Microsoft Corporation, L=Redmond, S=Washington, C=US

- try to move a meat from the ground to the stack with 20 meat (or to any stack that is not the first)

result: it will move the meat to the first slot, no matter what slot you move it to

curious fact: it works fine moving meats from a stack to another tho, but from outside the container there is the bug

nekiro:
Fix #2954

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -453,6 +453,10 @@ Cylinder* Container::queryDestination(int32_t& index, const Thing& thing, Item**
 
 	bool autoStack = !hasBitSet(FLAG_IGNOREAUTOSTACK, flags);
 	if (autoStack && item->isStackable() && item->getParent() != this) {
+		if (*destItem && (*destItem)->equals(item) && (*destItem)->getItemCount() < 100) {
+			return this;
+		}
+
 		//try find a suitable item to stack with
 		uint32_t n = 0;
 		for (Item* listItem : itemlist) {


EPuncker:
tested and working like a charm, I can select any stack in the container to move the item to and it wont always chose the first slot

CristoferMartins:
For a reason i forget i did it differently:

```
if (index != INDEX_WHEREEVER) {
    Item* itemFromIndex = getItemByIndex(index);
    if (itemFromIndex) {
        if (itemFromIndex != item && itemFromIndex->equals(item) && itemFromIndex->getItemCount() < 100) {
            *destItem = itemFromIndex;
            return this;
        }
    }
}
```

nekiro:
@CristoferMartins your code does not consider if itemFromIndex is a cylinder (container) which happens here:
https://github.com/otland/forgottenserver/blob/5a8f50bc6f44d3076555e5b113c04f8e65c9790f/src/container.cpp#L446

CristoferMartins:
Yeah but no bug will happens since equals won't match the id and moving the item direct to the backpack so it will move inside the backpack stills works.

nekiro:
@CristoferMartins thats nice, also it does ignore the flag "ignoreautostack"

CristoferMartins:
The code i posted happens below the autostack flag is checked:
```
bool autoStack = !hasBitSet(FLAG_IGNOREAUTOSTACK, flags);
bool isStackable = item->isStackable();
if (autoStack && item->isStackable() && item->getParent() != this) {
    if (index != INDEX_WHEREEVER) {
        Item* itemFromIndex = getItemByIndex(index);
        if (itemFromIndex) {
            if (itemFromIndex != item && itemFromIndex->equals(item) && itemFromIndex->getItemCount() < 100) {
                *destItem = itemFromIndex;
                return this;
            }
        }
    }
```

nekiro:
@CristoferMartins oh I see, so you wrote a bunch of redundant code imo. You have all these things you wrote before that autoStack check and you can just check the destItem variable.

CristoferMartins:
Yeah i think its redundant as you said.

PR #2956, opened 2020-04-22
Removed Purse item

EPuncker:
It was removed on 10.8 update and it's now a blank item

Purse2

https://tibia.fandom.com/wiki/Purse


omarcopires:
I think a good idea would not be to remove the item, just the attributes and write as a "reserved sprite" as @dudantas did at the clientid convention.

EPuncker:
> I think a good idea would not be to remove the item, just the attributes and write as a "reserved sprite" as @dudantas did at the clientid convention.

but they were **deleted**, they wont be used for anything in the future (?) and doesn't have an image associated to it, so there is no point in "reserving" it, but deleting it to avoid confusion and "visual" file size increase, we have maybe (just a blind guess) more than 20k of blank item IDs, imagine how messy it would be if we reserved sprite for all of them 😝 we need to think in another solution

soul4soul:
I'd rather leave it alone or add a comment above the item that it was removed in 10.8 and above. I don't think there is much harm in keeping it. It makes things easier for those who want to downgrade TFS to support 10.8.

Zbizu:
what about this?
```
<!-- purse, removed in 10.8  -->
<item id ="17432"/>
```

WibbenZ:
@EPuncker nothing to decide here if you ask me, best way is to just add a comment, like @Zbizu wrote.

PR #2959, opened 2020-04-22
Prevent table.contains and player:getLossPercent being called multipl…

gudan-garam:
Prevent table.contains and player:getLossPercent being called multiple times on player death.

If the player isn't wearing an Amulet of Loss, `table.contains({SKULL_RED, SKULL_BLACK}, ...)` and `player:getLossPercent()` would be called once for every item on players slots.


data/creaturescripts/scripts/droploot.lua
```diff
@@ -21,20 +22,32 @@ function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unj
...
+			end
 		end
 	end
 
 	if not player:getSlotItem(CONST_SLOT_BACKPACK) then
 		player:addItem(ITEM_BAG, 1, false, CONST_SLOT_BACKPACK)
 	end
 	return true
-end
+end
```
EPuncker, line 53: don't remove new line at eof 😋

gudan-garam: done, ty

data/creaturescripts/scripts/droploot.lua
```diff
@@ -3,8 +3,9 @@ function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unj
 		return true
 	end
 
+	local isRedOrBlack = table.contains({SKULL_RED, SKULL_BLACK}, player:getSkull())
 	local amulet = player:getSlotItem(CONST_SLOT_NECKLACE)
-	if amulet and amulet.itemid == ITEM_AMULETOFLOSS and not table.contains({SKULL_RED, SKULL_BLACK}, player:getSkull()) then
+	if amulet and amulet.itemid == ITEM_AMULETOFLOSS and not isRedOrBlack then
```
DSpeichert, line 8: This would look better in-order (amulet is used first, define it first).

data/creaturescripts/scripts/droploot.lua
```diff
@@ -21,15 +22,27 @@ function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unj
...
 						item:remove()
 					end
 				end
 			end
+		else 
+			for i = CONST_SLOT_HEAD, CONST_SLOT_AMMO do
+				local item = player:getSlotItem(i)
+				local lossPercent = player:getLossPercent()
+				if item then
+					if math.random(item:isContainer() and 100 or 1000) <= lossPercent then
```
DSpeichert, line 39: This could be:
```suggestion
					if isRedOrBlack or math.random(item:isContainer() and 100 or 1000) <= lossPercent then
```
and then the previous `if isRedOrBlack then...` is unnecessary as it's effectively the same code duplicated.

By the way, this function doesn't seem entirely correct, when lossPercent is 0, the random value could still be 0 and player would lose the item despite supposedly having 0% chance. There's <1% chance of that but still.

gudan-garam: Done.

I guess I exagerated on not checking the same thing multiple times.

Also fixed the 0% lossPercent chance.

data/creaturescripts/scripts/droploot.lua
```diff
@@ -23,8 +24,9 @@ function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unj
 	else
 		for i = CONST_SLOT_HEAD, CONST_SLOT_AMMO do
 			local item = player:getSlotItem(i)
+			local lossPercent = player:getLossPercent()
 			if item then
-				if table.contains({SKULL_RED, SKULL_BLACK}, player:getSkull()) or math.random(item:isContainer() and 100 or 1000) <= player:getLossPercent() then
+				if isRedOrBlack or math.random(item:isContainer() and 100 or 1000) < lossPercent then
```
DSpeichert, line 29: This is still off by one, I'm afraid.
Imagine it's a container, your random number is 100 and your lossPercent is 100.
You don't lose an item but you should.

I think proper algorithm would be like this:
```suggestion
				if isRedOrBlack or math.random(item:isContainer() and 100 or 1000) <= lossPercent then
```
and also wrap the inner code in `if lossPercet ~= 0 then`...

data/creaturescripts/scripts/droploot.lua
```diff
@@ -23,8 +24,9 @@ function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unj
...
 					if not item:moveTo(corpse) then
```
DSpeichert, line 30:
```suggestion
					if lossPercent ~= 0 and not item:moveTo(corpse) then
```

data/creaturescripts/scripts/droploot.lua
```diff
@@ -23,9 +24,10 @@ function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unj
 	else
 		for i = CONST_SLOT_HEAD, CONST_SLOT_AMMO do
 			local item = player:getSlotItem(i)
+			local lossPercent = player:getLossPercent()
 			if item then
-				if table.contains({SKULL_RED, SKULL_BLACK}, player:getSkull()) or math.random(item:isContainer() and 100 or 1000) <= player:getLossPercent() then
-					if not item:moveTo(corpse) then
+				if isRedOrBlack or math.random(item:isContainer() and 100 or 1000) <= lossPercent then
+					if lossPercent ~= 0 and not item:moveTo(corpse) then
```
DSpeichert, line 30:
```suggestion
					if (isRedOrBlack or lossPercent ~= 0) and not item:moveTo(corpse) then
```

PR #2962, opened 2020-04-23
Cache AccessList's result of regexp evaluations

DSpeichert:
Apparently expensive regular expressions are a risk factor.

### DIFF
--- a/src/house.cpp
+++ b/src/house.cpp
@@ -411,6 +411,7 @@ void AccessList::parseList(const std::string& list)
 	guildRankList.clear();
 	expressionList.clear();
 	regExList.clear();
+	regExCache.clear();
 	this->list = list;
 	if (list.empty()) {
 		return;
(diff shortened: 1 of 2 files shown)


DSpeichert:
Closing this, as we've decided to go with #2965.

Xaekai:
I would make you guys a patch that remained performant regardless of what players tried to do as house owners, while retaining RegEx, but it would pull `libpcre` as a dep, and I'm not familiar with how well that's supported on the Windows side, so I nixed that idea. :disappointed: 

Part of what we do at CyTube is offer channels custom filters, so they have tools to combat trolls and spam and such. Some channels have literally thousands of filters, and possibly hundreds of users in a channel, and every single message inbound to the server backend goes through every single filter sequentially. So we've spent some time dealing with minutiae of RegEx performance and attack surfaces.

DSpeichert:
@Xaekai even using `boost::regex` is supposedly "fixing" the problem by making the performance much better.
However, weighing the usefulness of this feature, we decided to remove it.
What's the point of putting a wildcard into your access list to potentially allow someone to create a character name that will fit and rob your house. They only need to know the pattern.

We retained static `*` for the main purpose to putting it on doors.

Xaekai:
I guess one would question why expressions were added in the first place then? Because `*` is not an expression, it's a globbing.

The git blame tells me most of the regex code goes back 7 years to the 2013 init, so I'll have to look in my SVN clone and see what was said there.

I do agree however that RegEx seems out of place here, given it presents attack surfaces in **two** contexts this way. But I could imagine me using it, if the surfaced ACL controls required anything past a single command to add a whole list, and it would be in the form of `( a | long | list | machines | serve | humans | not | the | other | way | around )` And I distinctly remember doing just that very thing many years ago, which is why I queried about this change at all. :smile: 

I'd actually like to see much more powerful forms of ACL, or at least one more form: Extend the "all characters under x account" metaphor to be available to guests too, and surface a way for the house owner to use it. Useful for the "IRLs"

DSpeichert:
That's why there is a way to invite a guild or a specific guild rank, I think that covers the use case.

The functionality existed because that's exactly how it worked in RL.

Xaekai:
That makes two presumptions: Guilds are enabled && guilds are reasonably easy to acquire.

I see your point that it's doable, but I can imagine numerous corner cases that would reveal it's hackiness in that role. But I concede to your authority over the codebase.

DSpeichert:
There is no other construct of groups in the codebase that could be used to represent groups of characters, really. In-game, the concept of an "account" is private to the owner of the account. Note that no feature of the game tells you whether two characters are on the same account or not.

Xaekai:
If you mean the unintended leaking of information, when I envisioned that in my head, the surfaced presentation to the player would be `account of %charactername%`

And thus, it would be _implicit_ such determinations could be made by the virtue of that player choosing to enter the house under the banner of such an invitation line.

Issue #1545, opened 2015-08-27
Player String Storage and Creature Storage


gesior:
1. I understand that for some efficiency and 'script safety' reasons, you prefere 'number' storage, but 'string' storage would be really useful for many scripts.

I think that should be added as new type of storage with functions like 'getStringStorageValue' and 'setStringStorageValue'.
1. Maybe add storage to creatures as it was in TFS 0.4? It was really useful for 'action'/'creaturescripts' to save some information in creature.

ranisalt:
:+1: to storage on creatures, it's really good.

brunominervino:
@gesior look:
https://gist.github.com/ninjalulz/481ff43ab084cdff69ae

About storage on creatures, I think it's good.

PrinterLUA:
"Npcs and monsters are temporary, use global storages.
If you want to use strings then make some sort of tables for each numeric value."

I quote something @dalkon told me.

gesior:
@brunominervino 
Send SQL query everytime you want to set/get storage? If you use it on 'players online' loop, it will freez server!

@PrinterLUA 
What? Make table of string that map to numbers? I want to keep in database string values!
Like player say '!mydescription I am cool guy' and I want to store it somewhere.

Use globalstorages? What? How? I got monster with some weird IDs and I want to keep for it few values. Is it so hard to add 1 array/map to Creature?

WibbenZ:
I can actually agree on this, might be quicker to use numbers, but it's not that bad having the ability to use strings.
Insted of huge arrays to let the server know that Storages.WrathOfTheEmpror.Mission1.HasTalkedWIth* means 1001 etc

About adding a query with each get / set function is just crazy?
Thats the kind of stuff you find on OTLand IMO :p 

Best would be to store it in the memory till we restart / save etc and then save it to the database (like we do with integer storages).

But these things gets closed right away pretty much with the "wontfix" label.

djarek:
@gesior
Asynchronous queries could be used to get and set storage values (though that requires the lua programmer to restructure their code to be "sort of" functional in style), too bad they're basically broken ATM (#1533).

Also, what's wrong with using a global lua table for creature storage? Making it persistent doesn't make sense for monsters anyway.

gesior:
@djarek 
Memory leak? Monsters die, storages stay. J/K
Global storage would be nice, because you could keep tables and objects in it.

I spent too much time on 0.3.6/0.4 where each folder had own LUA environment.
[...]

PR #2963, opened 2020-04-23
Change storage values size to int64_t

infernumx:
Main point of this was to use `os.mtime()` as a storage value for more precision.

### DIFF
--- a/data/migrations/26.lua
+++ b/data/migrations/26.lua
@@ -1,0 +1,3 @@
+function onUpdateDatabase()
+	return false
+end
--- a/schema.sql
+++ b/schema.sql
@@ -312,7 +312,7 @@ CREATE TABLE IF NOT EXISTS `player_spells` (
 CREATE TABLE IF NOT EXISTS `player_storage` (
   `player_id` int(11) NOT NULL DEFAULT '0',
   `key` int(10) unsigned NOT NULL DEFAULT '0',
-  `value` int(11) NOT NULL DEFAULT '0',
+  `value` bigint(20) NOT NULL DEFAULT '0',
   PRIMARY KEY (`player_id`,`key`),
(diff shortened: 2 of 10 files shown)


andersonfaaria:
In order for my antibot work I had to store the os.mtime() in some specific situations, what I did to workaround this was to make some math in order reduce the size of the number and be able to still store miliseconds without having to perform such an aggressive change.

One could simply use modulo operator in order to achieve that, but another simply way is to collect the os.time() once and store it forever as a global storage value, it will take a couple decades to explode if you use sugarcoat a function to consider os.time() - globalStorage with os.time()

Besides this 'workarounds', your PR is looking really neat but I would suggest we change it for string instead and try to tackle #1545

data/migrations/26.lua
```diff
@@ -0,0 +1,3 @@
+function onUpdateDatabase()
+	return false
+end
```
DSpeichert, line 3: missing EOL

schema.sql
```diff
@@ -312,7 +312,7 @@ CREATE TABLE IF NOT EXISTS `player_spells` (
 CREATE TABLE IF NOT EXISTS `player_storage` (
   `player_id` int(11) NOT NULL DEFAULT '0',
   `key` int(10) unsigned NOT NULL DEFAULT '0',
-  `value` int(11) NOT NULL DEFAULT '0',
+  `value` bigint(20) NOT NULL DEFAULT '0',
```
DSpeichert, line 315: You make it inconsistent by having a schema snapshot with latest change but without actually upgrading the version number in the schema. You don't really need to edit this file.

Zbizu:
You can store those values in server memory(`antibotLastCheck[cid] = os.mtime()`). This isn't really something we need to save like we do with quest states.

DSpeichert:
It's been 10 days without any response, are you still interested @infernumx ?

PR #2964, opened 2020-04-23
Added delayed spawn feature

EPuncker:
it was added at 10.7 version

Once this get merged, I'll do a new PR updating the monster files that have the isblockabe flag attribute...

https://tibia.fandom.com/wiki/Updates/10.7#Non-Blockable_Creatures

code by slavi dodo and otservbr-global


EPuncker:
and if this is going to be implemented, I want to tag @EvilHero90 to make the revscript part of the monster flag 😋

src/spawn.cpp
```diff
@@ -268,6 +273,17 @@ void Spawn::checkSpawn()
 	}
 }
 
+void Spawn::scheduleSpawn(uint32_t spawnId, spawnBlock_t& sb, uint16_t interval)
+{
+	auto delayedSpawn = g_config.getBoolean(ConfigManager::NEW_DELAYED_SPAWN);
```
infernumx, line 278: Why the use of `auto` here?

DSpeichert:
Definitely needs a better name than `newDelayedSpawn`!

EPuncker:
> Definitely needs a better name than `newDelayedSpawn`!

lmao I was out of ideas, still am, scheduledSpawn? delayedSpawn? newSpawnTimer? 🤔

soul4soul:
How about `forceSpawn`? Others which I don't like as much are `noBlock`, `nonBlockable`, `spawnOnPlayer`, `alwaysSpawn`, `dynamicSpawn`, `ignorePlayerBlock`.

`forceSpawn` flag - When the `forceSpawn` flag is set players will be unable to block the spawn of the monster by standing near the spawn position.

This should be a monster flag instead of a config option

1. It would provide much more flexibility by being a per creature option
2. It's a better mirror of cipbia implementation.

EPuncker:
> This should be a monster flag instead of a config option

it is still a monster flag (blockable=true by default) and I'll set the non blockable creature from [this list](https://tibia.fandom.com/wiki/Updates/10.7#Non-Blockable_Creatures) in a following up commit @soul4soul

EPuncker:
well... actually the config.lua entry makes no sense since the monster flag does exactly what it should already, I'll be removing it

soul4soul:
oh got it! I missed the isblockable flag added to the creatures. I agree with your last comment about not adding anything to config.lua for this change.

data/scripts/lib/register_monster_type.lua
```diff
@@ -81,6 +81,9 @@ registerMonsterType.flags = function(mtype, mask)
 		if mask.flags.healthHidden ~= nil then
 			mtype:isHealthHidden(mask.flags.healthHidden)
 		end
+		if mask.flags.Blockable ~= nil then
+			mtype:isBlockable(mask.flags.Blockable)
```
EPuncker, line 85: @EvilHero90 is that right?

andersonfaaria:
looking neat, thanks for the contribution

data/scripts/lib/register_monster_type.lua
EvilHero90: lowercase please

src/spawn.cpp
```diff
@@ -268,22 +273,26 @@ void Spawn::checkSpawn()
...
 	auto it = spawnedMap.begin();
 	while (it != spawnedMap.end()) {
 		uint32_t spawnId = it->first;
 		Monster* monster = it->second;
 		if (monster->isRemoved()) {
-			if (spawnId != 0) {
-				spawnMap[spawnId].lastSpawn = OTSYS_TIME();
-			}
-
+			spawnMap[spawnId].lastSpawn = OTSYS_TIME();
```
nekiro, line 293: why did you remove that spawnId check?

src/spawn.cpp
```diff
@@ -268,22 +273,26 @@ void Spawn::checkSpawn()
...
 			monster->decrementReferenceCounter();
 			it = spawnedMap.erase(it);
-		} else if (!isInSpawnZone(monster->getPosition()) && spawnId != 0) {
```
nekiro, line 284: and this one, why is it removed?

src/spawn.cpp
EPuncker: it was the "fix" for infinite respawn bug if monster walked away from the spawn, or was lured away

https://github.com/opentibiabr/otservbr-global/commit/34a20be159d84fd35877a49e2ca6dca4e9f519e3#diff-eecbbd78a76c67f252961d084417115a

PR #2965, opened 2020-04-24
House AccessList: drop support for regular expressions

DSpeichert:
This change removes support for house access list lines using:
* regular expressions (? for single character, * for many characters)
* exclusions (lines starting with !)

Guild name and guild rank is still supported.
Access list is limited to 100 entries and each entry can be max 100 characters long.

Conflicts: #2962

### DIFF
--- a/src/house.cpp
+++ b/src/house.cpp
@@ -409,8 +409,6 @@ void AccessList::parseList(const std::string& list)
 {
 	playerList.clear();
 	guildRankList.clear();
-	expressionList.clear();
-	regExList.clear();
 	this->list = list;
 	if (list.empty()) {
 		return;
@@ -419,7 +417,12 @@ void AccessList::parseList(const std::string& list)
 	std::istringstream listStream(list);
 	std::string line;
 
+	int lineNo = 1;
 	while (getline(listStream, line)) {
+		if (++lineNo > 100) {
+			break;
+		}
+
 		trimString(line);
 		trim_left(line, '\t');
 		trim_right(line, '\t');
@@ -439,7 +442,7 @@ void AccessList::parseList(const std::string& list)
 				addGuildRank(line.substr(0, at_pos - 1), line.substr(at_pos + 1));
 			}
 		} else if (line.find("!") != std::string::npos || line.find("*") != std::string::npos || line.find("?") != std::string::npos) {
-			addExpression(line);
+			continue; // regexp no longer supported
 		} else {
 			addPlayer(line);
 		}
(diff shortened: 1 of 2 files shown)


gpedro:
@FakeShinoda told me about this error in the last night. someone used this payload to increase cpu usage on his server

```` 
*********************(some numbers and letters)********
*********************(some numbers and letters)********
*********************(some numbers and letters)********

*********************(some numbers and letters)********
*********************(some numbers and letters)********
*********************(some numbers and letters)********

*********************(some numbers and letters)********
*********************(some numbers and letters)********
*********************(some numbers and letters)********
````

File: src/house.cpp [hunk 1]
```diff
@@ -439,7 +442,7 @@ void AccessList::parseList(const std::string& list)
 				addGuildRank(line.substr(0, at_pos - 1), line.substr(at_pos + 1));
 			}
 		} else if (line.find("!") != std::string::npos || line.find("*") != std::string::npos || line.find("?") != std::string::npos) {
-			addExpression(line);
+			continue; // regexp no longer supported
 		} else {
 			addPlayer(line);
```
gpedro, line 447: what are you think about check if the line contains only valid characters? [A-Za-z- ']

CristoferMartins:
I don't agree with this PR. there is a fix for the lag issue and atleast one * should be support to allow everyone enters in the house(players will miss this i'm sure).

src/house.cpp
DSpeichert: I don't think it matters here, we only match it against the character name with an exact match.

Zbizu:
what about ignoring the line entirely if the expression gets too complex?

gpedro:
@DSpeichert can you test if use `boost::regex` instead ` std::regex` solves the problem?

DSpeichert:
@gpedro we know about this option and that it's used somewhat successfully.

However, regular expressions can still have dramatically varying complexity, regardless of the library in use. They don't really make much sense in the context of houses, hence the decision to remove them.

ranisalt:
It breaks due to catastrophic backtracking, and can be solved by replacing repetitions of * with a single. After all, `.*.*` is the same as `.*`, sans the catastrophe.

Another option is to move it to Lua and use [patterns](http://lua-users.org/wiki/PatternsTutorial) instead. It's much simpler and faster.

src/house.cpp
```diff
@@ -499,52 +502,11 @@ void AccessList::addGuildRank(const std::string& name, const std::string& rankNa
...
-				regExList.emplace_back(std::regex(outExp), true);
-			}
-		}
-	} catch (...) {}
-}
-
 bool AccessList::isInList(const Player* player)
 {
 	std::string name = asLowerCaseString(player->getName());
 	std::cmatch what;
```
Lyuzera, line 514: you can remove these variables `name` and `what`, they are no longer being used.

Lyuzera: ..and you need to reset the variable `allowEveryone`
```diff
void AccessList::parseList(const std::string& list)
{
	playerList.clear();
	guildRankList.clear();
+	allowEveryone = false;
```
Why? because even if you remove `*` from aleta sio/som, players continue to enter the house.

DSpeichert: good catch, thanks!

Sickness-chan:
Looks good to me so far, but if we keep the _allow everyone (*)_ functionality, we should also have exclusions (!) available.

src/house.cpp
```diff
@@ -419,7 +418,12 @@ void AccessList::parseList(const std::string& list)
 	std::istringstream listStream(list);
 	std::string line;
 
+	int lineNo = 1;
 	while (getline(listStream, line)) {
+		if (++lineNo > 100) {
```
EvilHero90, line 423: I think we should approach this with not more than 50, I'd honestly doubt that someone is really willing to put 100 different characters in there for a real reason.
I would also change ```line.length() > 100``` to something lower aswell, as we only support one character per line now and I honestly doubt anyone is having characters in his database which are probably longer than 25 letters in length.

DSpeichert: Rather than finding the lowest limit we can think would work, my goal is to establish an upper bound that is extremely unlikely to be problematic for normal usage.
Unless there is still concern for abuse with current limits, I think 100x100 is fine.

Sickness-chan: Why don't you just make it configurable through a config variable that goes along with the other house settings? In my eyes it wouldn't be much different from making max packets per second configurable.

PR #2966, opened 2020-04-24
Fix trade bug

mackerel225:
More information about the bug can be found here: https://otland.net/threads/trade-bug.270083/

Feel free to change the code if you feel there is a better approach as it's my first time looking into C++ properly. One of the things I wanted to change is to simply call slotType for traded item, like this: playerTradeItem->getSlotType(). But unfortunately, it does not seem possible as first you have to retrieve its type and then use another function to retrieve its slot type.

Anyway, with these changes, the bug goes away!

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -2722,19 +2722,43 @@ void Game::playerAcceptTrade(uint32_t playerId)
 		}
 
 		bool isSuccess = false;
+		
+		// assign item as itemType
+		const ItemType& partnerItemType = Item::items[partnerTradeItem->getID()];
+		const ItemType& playerItemType = Item::items[playerTradeItem->getID()];
 
-		ReturnValue tradePartnerRet = internalAddItem(tradePartner, playerTradeItem, INDEX_WHEREEVER, 0, true);
-		ReturnValue playerRet = internalAddItem(player, partnerTradeItem, INDEX_WHEREEVER, 0, true);
+		// retrieve slot type for given item
+		slots_t partnerSlotItem = getSlotType(partnerItemType);
+		slots_t playerSlotItem = getSlotType(playerItemType);
+
+		// initialise default variables
+		ReturnValue tradePartnerRet = RETURNVALUE_NOERROR;
+		ReturnValue playerRet = RETURNVALUE_NOERROR;
+
+		// if player is trying to trade its own backpack
+		if (tradePartner->getInventoryItem(CONST_SLOT_BACKPACK) == partnerTradeItem) {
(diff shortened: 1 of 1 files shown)


nekiro:
I think you should try using queryAdd to query if the item can be added to player before you remove item from anyone instead of that player:getInvetoryItem

src/game.cpp
```diff
@@ -2722,19 +2722,43 @@ void Game::playerAcceptTrade(uint32_t playerId)
 		}
 
 		bool isSuccess = false;
+		
+		// assign item as itemType
+		const ItemType& partnerItemType = Item::items[partnerTradeItem->getID()];
+		const ItemType& playerItemType = Item::items[playerTradeItem->getID()];
+
+		// retrieve slot type for given item
+		slots_t partnerSlotItem = getSlotType(partnerItemType);
```
nekiro, line 2731:
```
slots_t partnerSlotItem = getSlotType(Item::items[partnerTradeItem->getID()]);
slots_t playerSlotItem = getSlotType(Item::items[playerTradeItem->getID()]);
```
and remove these variables for itemtypes

mackerel225: This is what I was looking for! Didn't know that was possible. Thanks. Anyway, the code has been refactored.

mackerel225:
> 
> 
> This does not solve all different scenarios when trade can still fail right?
> If so, in which cases it still fails?

The code does not modify current behaviour of the trade system, it just provides a check against the bug that was happening when trading own backpack.

src/game.cpp
```diff
@@ -2722,19 +2722,35 @@ void Game::playerAcceptTrade(uint32_t playerId)
 		}
 
 		bool isSuccess = false;
+		
+		ReturnValue tradePartnerRet = RETURNVALUE_NOERROR;
+		ReturnValue playerRet = RETURNVALUE_NOERROR;
+
+		// if player is trying to trade its own backpack
+		if (tradePartner->getInventoryItem(CONST_SLOT_BACKPACK) == partnerTradeItem) {
+			slots_t playerSlotItem = getSlotType(Item::items[playerTradeItem->getID()]);
```
nekiro, line 2731: as far as I see this variable and the other one are used only once right? If you aren't checking them or anything I think you can just put the code as 
```
tradePartnerRet = (tradePartner->getInventoryItem(getSlotType(Item::items[playerTradeItem->getID()])) ? RETURNVALUE_NOTENOUGHROOM : RETURNVALUE_NOERROR);
```
What do you think?

PR #2967, opened 2020-04-24
Updated hidden nest and related items name

EPuncker:
- also fixed its item name and added missing dragon's nest entries in items.xml
- the surprise nest will be added in a following PR that have all random items into one file

it was added on 9.5 update
https://tibia.fandom.com/wiki/Hidden_Nest


DSpeichert:
So 16093 is NOT a surprise nest?

EPuncker:
> So 16093 is NOT a surprise nest?

I've checked in rl tibia and in ot: 16093 is unmovable object called hidden nest and 16102 is pick-upable item called surprise nest, both have same sprite

DSpeichert:
Okay, is this ready to review? I don't see any script right now.

EPuncker:
> Okay, is this ready to review? I don't see any script right now.

ready to merge, I cam to the conclusion that the script is a lot more complicated than I thought since it is not an unique time reward, I'll need to gather more information for it in the next event, so no script for now

PR #2968, opened 2020-04-25
Fixes using shovel on sand hole spots

gudan-garam:
When using a shovel on sand with action id 100 it should:
1) 20% of the time open the hole
2) 5% of the time spawn a scarab
3) 1% of the time spawn a scarab coin

This script was able to do 2 and 3 on a sand spots without action id 100.
This script was not able to do 3 on a sand spot with action id 100.

It is now fixed.

### DIFF
--- a/data/actions/lib/actions.lua
+++ b/data/actions/lib/actions.lua
@@ -170,14 +170,14 @@ function onUseShovel(player, item, fromPosition, target, toPosition, isHotkey)
 
 		toPosition.z = toPosition.z + 1
 		tile:relocateTo(toPosition)
-	elseif table.contains(sandIds, groundId) then
+	elseif table.contains(sandIds, groundId) and target.actionid == 100 then
 		local randomValue = math.random(1, 100)
-		if target.actionid == 100 and randomValue <= 20 then
+		if randomValue <= 20 then
 			ground:transform(489)
 			ground:decay()
-		elseif randomValue == 1 then
+		elseif randomValue >= 98 then
 			Game.createItem(2159, 1, toPosition)
-		elseif randomValue > 95 then
+		elseif randomValue >= 93 then
 			Game.createMonster("Scarab", toPosition)
 		end
 		toPosition:sendMagicEffect(CONST_ME_POFF)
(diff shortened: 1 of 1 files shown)


DSpeichert:
There should probably be a central list of all action IDs, like `data/lib/core/storages.lua` but for action IDs. How do we actually know it should be 100?

EPuncker:
> There should probably be a central list of all action IDs, like `data/lib/core/storages.lua` but for action IDs. How do we actually know it should be 100?

well, since there is not that many places in default datapack that uses AIDs/UIDs, maybe we should just do like the pick-able ground and use:

`and (ground:hasAttribute(ITEM_ATTRIBUTE_UNIQUEID) or ground:hasAttribute(ITEM_ATTRIBUTE_ACTIONID))`

instead of target.aid 100?

soul4soul:
Neither script it quite correct. The original is better because it won't cause holes to appear all over the map where they don't belong!!

Sand digging was my favorite past time for regaining soul points. 

There are only 7 spots in all of Ankrahmun that digging sand will turn into a hole. Those 7 spots are the entrances to tombs. Those 7 spots will never reveal a scarab or scarab coin. 20% chance to open the hole sounds about correct. Failure to open the hole will cause a poof effect on the hole. It is possible to dig on these spots when a player or creature is on the spot. If the hole opens the player or creature will fall down the hole. It is not possible to dig if an item (worm, gold coin, label, etc...) is on the spot.

There are about 150~ spots that reveal scarabs or scarab coins. These 150 spots will never turn into a hole. When "spawned" there is 100% chance to get a scarab or scarab coin, it will never be nothing. These spots that reveal scarabs and scarab coins have a 1 hour "respawn" before they can be dug again. The drop rate of a scarab coin is closer to 5% and the other 95% of the time a scarab will appear.

With the addition of the map cyclopdeia all sand spots which can be dug show up as gray dots on the minimap. Due to this change I feel it's possible cip changed the ID for sand which can be dug.

See reference picture. Notice all of the light gray dots.
image

gudan-garam:
As soul4soul pointed out, I don't think we should change the action id of just check if there is an action id, I have a datapack from 6 years ago that I don't even know where I downloaded and it already had this action id on the right spots. Almost the same with the doors and etc.

I didn't know about the number of spots able to spawn scarabs and scarab coins. I have hunted on ankrahmun thombs from 2010 ~ 2011 though and it would spawn scarabs and scarab coins on the thomb entrance, did that change?

gudan-garam:
If souls4soul's logic is right, this should be closed because #2971 is a better approach.

PR #2969, opened 2020-04-25
Disappearing item trade fix

nekiro:
This is alternative version of fix provided by @mackerel225 in #2966
also now player and partner variable are guaranteed to be the real trade requester and trade partner, does not matter who accepted the trade last.

PR #2970, opened 2020-04-25
Player:onTradeCompleted(...)

Zbizu:
new player event triggering after trade was successful

(onTradeAccept triggers whether the trade is successful or not so they're different)

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -100,8 +100,8 @@ bool Events::load()
 				info.playerOnTradeRequest = event;
 			} else if (methodName == "onTradeAccept") {
 				info.playerOnTradeAccept = event;
-			} else if (methodName == "onTradeSuccess") {
-				info.playerOnTradeSuccess = event;
+			} else if (methodName == "onTradeCompleted") {
+				info.playerOnTradeCompleted = event;
 			} else if (methodName == "onMoveItem") {
 				info.playerOnMoveItem = event;
 			} else if (methodName == "onItemMoved") {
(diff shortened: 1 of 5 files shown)


Zbizu:
personal thoughts:
most of `Events::eventPlayerOnTradeSuccess` function content is same as tradeRequest and tradeAccept

There is probably a way to optimize it, but I'm not aware of such.

nekiro:
shouldnt this event be more unified like OnTradeCompleted and push if success or not in parameter?

mackerel225:
> 
> 
> shouldnt this event be more unified like OnTradeCompleted and push success variable as parameter?

My initial thought, with this you could remove onTradeAccept, or keep them both.

Zbizu:
great idea

I will edit it in a few hours.

src/events.cpp
```diff
@@ -820,6 +822,41 @@ bool Events::eventPlayerOnTradeAccept(Player* player, Player* target, Item* item
 	return scriptInterface.callFunction(4);
 }
 
+void Events::eventPlayerOnTradeCompleted(Player* player, Player* target, Item* item, Item* targetItem, bool isSuccess)
+{
+	// Player:onTradeSuccess(target, item, targetItem)
```
nekiro, line 827: you are missing rename and parameter in that comment

Zbizu:
There we go

nekiro:
@Zbizu onTradeAccept is called before all that happens so you can prevent from trading some stuff etc, so no thats different event, used for different purposes.

Zbizu:
possible conflicts: #2966 #2969

PR #2971, opened 2020-04-25
Added ActionIds table

EPuncker:
as requested here
https://github.com/otland/forgottenserver/pull/2968#issuecomment-619307217

### DIFF
--- a/data/actions/lib/actions.lua
+++ b/data/actions/lib/actions.lua
@@ -107,7 +107,7 @@ function onUsePick(player, item, fromPosition, target, toPosition, isHotkey)
 		return false
 	end
 
-	if table.contains(groundIds, ground.itemid) and ground.actionid == aId.pickHole then
+	if table.contains(groundIds, ground.itemid) and ground.actionid == actionIds.pickHole then
 		ground:transform(392)
 		ground:decay()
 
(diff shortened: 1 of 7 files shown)


EPuncker:
@DSpeichert like this? 🤔

data/actions/lib/actions.lua
```diff
@@ -172,7 +172,7 @@ function onUseShovel(player, item, fromPosition, target, toPosition, isHotkey)
 		tile:relocateTo(toPosition)
 	elseif table.contains(sandIds, groundId) then
 		local randomValue = math.random(1, 100)
-		if target.actionid == 100 and randomValue <= 20 then
+		if target.actionid == aId.sandHole and randomValue <= 20 then
 			ground:transform(489)
 			ground:decay()
 		elseif randomValue == 1 then
```
gudan-garam, line 178: if [this](https://github.com/otland/forgottenserver/pull/2968#issuecomment-619383257) is right, we should check for another action id here and on the last else.

data/lib/core/actionids.lua
```diff
@@ -0,0 +1,7 @@
+aId = {
```
DSpeichert, line 1: this can be `actionIds`, no need to shorten in like that, it's a global variable so short names can introduce naming collisions.

PR #2506, opened 2018-10-02
Update FindLua.cmake and use CMAKE_CXX_STANDARD

ranisalt:
Copied `FindLua` from CMake 3.12, it is a bit more robust than our version. Also, set `CMAKE_CXX_STANDARD` variable instead of using a script file, but it requires the minimum version to be 3.1

PR #2973, opened 2020-04-25
cmake: tweaks for cmake, vcpkg & CI

DSpeichert:
Makes compilation using CLion, cmake and vcpkg possible on Windows/Linux/MacOS.

* cmake is upgraded and related configs modified (`cmake` minimum `3.10`)
   * For per-OS cmake version check [here](https://github.com/otland/forgottenserver/pull/2506#issuecomment-619726048)
   * `FindLua.cmake` removed because we can use the cmake-bundled version
   * `FindCXX11.cmake` removed for the same reason
   * `FindLTO.cmake` removed (https://stackoverflow.com/a/47370726/1447395)
   * Closes #2506
* build using cmake on Windows (with vcpkg) is now possible
* cotire is not used when redundant (cmake >= 3.16)
* Lua version shown on startup
* Release-like zipped artifacts can be used by anyone for the target platform: https://github.com/otland/forgottenserver/actions/runs/92588998

Provides Github Actions workflows, which use the overhauled cmake build.
[Software preinstalled on Github Runners - Ubuntu 18.04](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md)

There are multiple workflows:
* Lua check
   * Triggered on both master and PRs
   * Runs luac on all *.lua files for syntax errors ONLY (not luacheck)
* XML check
   * Triggered on both master and PRs
   * Runs xmllint on all *.xml files for syntax errors ONLY (no DTD checks)
* Coverity
    * Runs a scan once a week on schedule (not triggered by any push)
    * Scan based on masteer
    * Compilation on Ubuntu using system packages (not vcpkg)
    * Uses Ubuntu 18.04 currently (github's ubuntu-latest)
    * Only LuaJIT variant
* Ubuntu build
  * Triggered on PR and master
  * Similar to Coverity build (ubuntu-latest, luajit, no vcpkg)
  * Produces `tfs` binary as the only artifact, the binary is kinda portable
* Docker 
  * Triggered on master branch push only
  * Uses alpine and system packages
  * LuaJIT only
  * Pushes docker image (with datapack) to Github's package registry
* vcpkg-based build
  * Triggered on PR and master branch
  * Matrix of Windows/Linux/MacOS x Release/Debug x Lua/LuaJIT
  * Produces most release-like artifacts, ready to run on every platform with datapack
  * Caveat: vcpkg doesn't know how to build LuaJIT on MacOS and Linux, so system packages are used there
 

With these, we can remove Travis CI and AppVeyor as they provide nothing more and are frequently failing by themselves.
We technically lose two things though:
* Travis CI tested compilation with clang, proposed Github Actions don't do that currently
* AppVeyor used actual vc14 project files (not cmake-based), so now we won't be testing `vc14/*` project files. Should we?


.github/workflows/build-ubuntu.yml
```diff
@@ -0,0 +1,55 @@
...
+      - master
+      - v*
+
+    tags:
+      - v*
+
+    paths:
+      - cmake/**
+      - src/**
+      - CMakeLists.txt
```
ranisalt, line 15: What is being skipped here? Just let it run always...

.github/workflows/build-ubuntu.yml
```diff
@@ -0,0 +1,55 @@
+name: Build on Ubuntu
+
+on:
+  push:
+    branches:
+      - master
+      - v*
```
ranisalt, line 7: I don't think it needs to be specified twice. Just use `tags` and it will suffice.

CMakeLists.txt
```diff
@@ -1,56 +1,77 @@
...
+else ()
+    include(cotire)
+    set_target_properties(tfs PROPERTIES COTIRE_CXX_PREFIX_HEADER_INIT "src/otpch.h")
+    set_target_properties(tfs PROPERTIES COTIRE_ADD_UNITY_BUILD FALSE)
+    cotire(tfs)
+endif ()
 
-add_compile_options(-Wall -Werror -pipe -fvisibility=hidden)
+if (NOT WIN32)
+    add_compile_options(-Wall -Werror -pipe -fvisibility=hidden)
```
ranisalt, line 26: I suggest removing `-pipe` here and let the user specify via build flags instead.

cmake/FindLuaJIT.cmake
```diff
@@ -10,18 +10,19 @@
...
-  PATH_SUFFIXES include/luajit-2.0 include/luajit-2.1 include
+  PATH_SUFFIXES include/luajit-2.0 include/luajit-2.1 include luajit
   PATHS
   ~/Library/Frameworks
   /Library/Frameworks
   /sw # Fink
   /opt/local # DarwinPorts
   /opt/csw # Blastwave
   /opt
+  PATH_SUFFIXES luajit
```
ranisalt, line 21: Introduced duplicate

.github/workflows/build-ubuntu.yml
DSpeichert: Uh, `data/**` ? A lot of commits for datapack only. This artifact doesn't include datapack here.

DSpeichert:
> Personal preference on formatting CMake files, eh?

CLion -> Reformat code.
Sorry, habit.

CMakeLists.txt
DSpeichert: I didn't add it, so it has only been "reorganized". We can remove it but I'm trying to limit changes here.

https://stackoverflow.com/questions/1512933/when-should-i-use-gccs-pipe-option

.github/workflows/build-ubuntu.yml
DSpeichert: Yes and no. We don't currently tag stable-backports. But we also don't really merge into stable branches either. I agree it may not be necessary. Artifacts don't persist longer than 90 days, so we will need actual release builds for github releases anyway.

DSpeichert: Actually, I'll leave it because we may want to get a preview build of a stable branch before actually cutting out a tag from it.

cmake/FindLuaJIT.cmake
DSpeichert: Fixed!

src/luascript.h
```diff
@@ -20,11 +20,7 @@
 #ifndef FS_LUASCRIPT_H_5344B2BC907E46E3943EA78574A212D8
 #define FS_LUASCRIPT_H_5344B2BC907E46E3943EA78574A212D8
 
-#if __has_include("luajit/lua.hpp")
-#include <luajit/lua.hpp>
-#else
 #include <lua.hpp>
```
gpedro, line 23: thaaaaaaaaaaaaaanks for that!

soul4soul:
I really like this PR with all of its improvements to CI. What about dropping the vc14 files from the repo and sticking with cmake only? It's something I was thinking trying to get past a PR anyway. Then we don't have to worry about building the vc14 project. cmake is bundled with the visual studio installer nowadays. Visual studio even supports opening cmake projects. Visual studio users don't even have to run cmake manually before popping open visual studio anymore.

.github/workflows/build-ubuntu.yml
ranisalt: Ah, that makes sense, I didn't remember data :P it could be a separate repo as a submodule.

PR #2974, opened 2020-04-26
Organized monster folder alphabetically

EPuncker:
Now it is easier to browser through the files, we have only 3 folders now: Bosses, lua and Monsters... all sorted alphabetically, and the monsters.xml file has been divided into two parts, bosses and monsters, with the two lists being sorted alphabetically as well! Now it should be way more intuitive to find a monster instead of the mess of lots of sub-folders that we had before!

as discussed [here](https://otland.net/threads/brainstorm-boss-folder-organization.268447/)

- no other change has been done in the monsters files other than changing its location
- arena bosses were moved to bosses
- no errors/warnings in the console at all


soul4soul:
I find three reasonable options for organizing monsters. I have no preference towards any of these organizational system. I'm happy that we are moving to a defined system instead of a random system.

1. Cipsoft classifies monsters themselves in the bestiary. We could chose to follow them and still use folders
1. If we want a boss folder and all other monsters folder we should consider following tibiawiki. I'm not completely sure how tibiawiki editors decide but they use a isboss flag. All creatures which a gold star next to their name are consider boss monsters. https://tibia.fandom.com/wiki/List_of_Creatures/. It appears special raid monsters, yearly event monsters, and monsters that have a 20hr kill cooldown are all consider bosses.
1. We could get rid of all folders and have 1 huge alphabetically organized folder.


Additional reference image showing isboss flag used by tibiawiki.
image

slavidodo:
Why removing category comment from monsters.xml though?

DSpeichert:
Conflicts again.

EPuncker:
> Conflicts again.

fixed

EPuncker:
@DSpeichert solved everything

Xaekai:
Mmmkay.

I myself have no horse in this race, because I have no issues writing scripts to warp upstreams divergence into whatever shape I prefer locally. But....this seems like the sort of thing a bash script could fabricate by way of symlinks for anyone who desired to browse it in such a fashion, whereas the reversal does not have such a programmatic solution. (Because a reverse map of this would become maintenance overhead) And symlinks exist in NTFS as well, so such a solution is not Linux specific.

Znote:
> Mmmkay.
> 
> I myself have no horse in this race, because I have no issues writing scripts to warp upstreams divergence into whatever shape I prefer locally. But....this seems like the sort of thing a bash script could fabricate by way of symlinks for anyone who desired to browse it in such a fashion, whereas the reversal does not have such a programmatic solution. (Because a reverse map of this would become maintenance overhead) And symlinks exist in NTFS as well, so such a solution is not Linux specific.

You sure?
You can use `Game.getMonsterTypes()` to retrieve all loaded monsters, and each monsterType has a race. (But is race a unique id for the monster, or a common nominator for the group of monsters, etc "Orcs")? This can be exported to a flatfile you can do whatever you want with in bash on startup.

Sickness-chan (approved):

DSpeichert:
Conflicts... let's get that merged soon!

EPuncker:
> Conflicts... let's get that merged soon!
solved, finally

DSpeichert (approved):

luanluciano93 (approved):

PR #2978, opened 2020-04-28
Parse packet in lua feature

nekiro:
This pr allows you to parse packet in lua using creaturescripts interface.
I have included template creaturescript how to parse packet.
Registering a packet in lua means it will not be parsed in source code.


CristoferMartins:
There is a "major" mistake in this pr in the protocolgame code. The player methods and specially the creatureevents code should be called by the dispatcher thread, you are calling them from the main/asio thread = crash.

infister:
> There is a "major" mistake in this pr in the protocolgame code. The player methods and specially the creatureevents code should be called by the dispatcher thread, you are calling them from the main/asio thread = crash.

I guess this is not very hard to correct. However, I prefer separating this feature into modules, like in OTX.

nekiro:
@CristoferMartins interesting, can you tell me in which cases or why would it crash?

nekiro:
@infister to get module "feeling" you can use revscriptsys.

infister:
> @infister to get module "feeling" you can use revscriptsys.

try to reload this with syntax error and then receive such packet

SaiyansKing:
> @CristoferMartins interesting, can you tell me in which cases or why would it crash?
> cause I was testing and I didnt see any crashes, so I'm curious

You have a racing condition inside "getParsePacketEvent" if somehow new event will be registered from dispatcher thread and asio thread will be looping through eventsList it'll result in crash.

Also I'm seeing that many people don't understand what is passing object by reference because your function "executeParsePacket" might get different data when it actually get called by dispatcher thread(data can be changed by asio thread).

MillhioreBT:
I prefer to use the OTX method, and it is also compatible with Revscripts ;)

nekiro:
@MillhioreBT "otx" you meant @slavidodo method right? cause their code is stolen from slavi

nekiro:
@SaiyansKing can you enligthen me, how and why would be the data changed by asio thread?

MillhioreBT:
> @MillhioreBT "otx" you meant @slavidodo method right? cause their code is stolen from slavi
It's pretty good anyway, why switch to yours? is better?

nekiro:
@MillhioreBT ofc, you are free to use whatever you like to use. I'm not the one to judge which one is better and if you like their implentation better then use the one that suit you the best.
This is my implementation and that's all.

SaiyansKing:
> @SaiyansKing can you enligthen me, how and why would be the data changed by asio thread? I will agree thats something I dont know much so I want to learn, will enjoy a good explanation

Its just that when you had reference to "Connection::msg" asio thread would read new data to that buffer without waiting for your dispatcher task to do its work. It is rather rare racing condition but a possibility is a possibility so it was worth mentioning it.

nekiro:
@SaiyansKing thanks! Can I request your review on that pr? I changed the code a bit I guess it should be fine now?

CristoferMartins:
I also have a question(which is related to this pull request): why the packet parsing is not done by the dispatcher thread?

nekiro:
@CristoferMartins can you review the pull request, because I'm not sure what you mean now

SaiyansKing:
> @SaiyansKing thanks! Can I request your review on that pr? I changed the code a bit I guess it should be fine now?

I'm sorry, I can't do reviews on pr's here, it looks ok, however someone might question you why you passed player directly instead of player uid but looking at connection.cpp it shouldn't be possible to queue work before "Protocol::release" will be called so I don't know reason why they used player uid in first place.

> I also have a question(which is related to this pull request): why the packet parsing is not done by the dispatcher thread?

If you're asking me then I don't know performance-wise it doesn't matter because at most you read only few bytes and main job still do dispatcher, maybe they didn't want to make unnecessary copies of networkmessage buffer because as I posted before passing it as reference would have rare racing condition.

CristoferMartins:
Memory leak in the Game::playerExecuteParsePacketEvent, message(pointer to NetworkMessage) is not deleted.

nekiro:
@CristoferMartins message is deleted in lua

MillhioreBT:
@nekiro Will a situation be possible where it is necessary to keep the pointer alive for an extra time at the end of the event? otherwise, we can clean pointer from fonts to avoid writing an additional line, what do you think?

nekiro:
@MillhioreBT you shouldnt have to delete the message manually, garbage collector should delete it at the next run, when reference count hits 0.

kygov:
In my opinion we should implement code from otservbr as its much better. 
https://github.com/opentibiabr/otservbr-global/blob/develop/src/modules.cpp

nekiro:
@kygov if you want to openly say something is better than something else. You should give us arguments on why.

kygov:
As @infister said it has much better shape than yours code, they use separated 'module' to packets and you mixing it with creature events..

nekiro:
@kygov so extendedopcode event is also wrongly implemented, because its also in creaturescripts interface and not separate module?

kygov:
Parsing packets in lua is probably targeted in bigger systems like game store, daily reward, inspection, quick looting etc. and in my opinion its reason why should have separated folder/module. It makes sesne in bigger projects like real maps. 

Even if something could be done in creature events its not reason to do it in this way, we should choose best way.

MillhioreBT:
> In my opinion we should implement code from otservbr as its much better.
> https://github.com/opentibiabr/otservbr-global/blob/develop/src/modules.cpp

I had the same idea, but this module is a copy of **CreatureEvents.cpp** and suffers from an old problem that is currently solved (#2897) for the **CreatureEvents**, so honestly it is better to use it in this method, obviously it can be improved, such as adding the **delay** and it would be everything

The point is to improve, either by changing the way of thinking and leaving old customs, but you can still give your opinion.

MillhioreBT:
> Parsing packets in lua is probably targeted in bigger systems like game store, daily reward, inspection, quick looting etc. and in my opinion its reason why should have separated folder/module. It makes sesne in bigger projects like real maps.
> 
> Even if something could be done in creature events its not reason to do it in this way, we should choose best way.

Remember it is 2020, there is something new called "**Revscripts**" ;)

src/game.cpp
```diff
@@ -5582,6 +5582,29 @@ void Game::playerAnswerModalWindow(uint32_t playerId, uint32_t modalWindowId, ui
 	}
 }
 
+void Game::playerExecuteParsePacketEvent(uint32_t playerId, uint8_t recvbyte, NetworkMessage* message)
+{
+	Player* player = getPlayerByID(playerId);
+	if (!player) {
+		return;
```
CristoferMartins, line 5589: If player not online = memory leak.

src/game.cpp
```diff
@@ -5582,6 +5582,29 @@ void Game::playerAnswerModalWindow(uint32_t playerId, uint32_t modalWindowId, ui
...
+	}
+
+	if (!player->hasEventRegistered(CREATURE_EVENT_PARSE_PACKET)) {
+		return;
```
CristoferMartins, line 5593: If player does not have the event registered = memory leak.

src/game.cpp
```diff
@@ -5582,6 +5582,29 @@ void Game::playerAnswerModalWindow(uint32_t playerId, uint32_t modalWindowId, ui
...
+	}
+
```
CristoferMartins, line 5595: From now everything creates a memory leak, only exception to this if creatureEvent->executeParsePacket is executed correctly and the networkmessage is added to the lua enviroment, then in this case it will be deleted by the gc.

src/protocolgame.cpp
```diff
@@ -415,6 +415,8 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
 		}
 	}
 
+	addGameTask(&Game::playerExecuteParsePacketEvent, player->getID(), recvbyte, new NetworkMessage(msg));
```
gpedro, line 418: this should be on default-switch, uhm? else you will handle twice (c++ and lua). I'm just saying...

nekiro: yeah, we are copying msg here so we dont care. Lua gets his own copy and c++ has its own copy, so we it interfere with each other

nekiro:
@CristoferMartins thanks for pointing that out, is it better now?

nekiro:
I didn't test that pr using revscriptsys, just the normal way via registering it in xml. I would appreciate if someone could test it using revscriptsys.

nekiro:
@MillhioreBT thats not wrong, why would you parse one recvbyte with few events? You can register one event for one recvbyte and thats correct and I implemented it like that on purpose. What problem with lua garbage collector?

MillhioreBT:
> @MillhioreBT thats not wrong, why would you parse one recvbyte with few events? You can register one event for one recvbyte and thats correct and I implemented it like that on purpose. What problem with lua garbage collector?

It should be possible to log multiple events with the same 'byte', this will greatly expand the x1000 possibilities, for example I want to create several separate events that use the same byte, I am trying to do it as below, but it generates the famous __gc error which I have mentioned previously...

```c++
void Game::playerExecuteParsePacketEvent(uint32_t playerId, uint8_t recvbyte, NetworkMessage* message)
{
	bool deleteMsg = true;
	if (player && player->hasEventRegistered(CREATURE_EVENT_PARSE_PACKET)) {
		for (CreatureEvent* creatureEvent : player->eventsList) {
			if (!creatureEvent->isLoaded()) {
				continue;
			}

			if (creatureEvent->getEventType() == CREATURE_EVENT_PARSE_PACKET && creatureEvent->getRecvbyte() == recvbyte) {
				creatureEvent->executeParsePacket(player, recvbyte, message);
				deleteMsg = false;
			}
		}
	}

	if (deleteMsg) {
		delete message;
		message = nullptr;
	}
}
```

Screenshot_2

nekiro:
@MillhioreBT I wont change that cause imo it shouldnt be, but if you want it to work like that, you have to copy network message for every event you execute, you are probably getting problems with trying deleting deleted pointer.

EvilHero90:
What about making NetworkMessage* into a smart pointer, so we don't run into problems?

MillhioreBT:
> @MillhioreBT I wont change that cause imo it shouldnt be like that, but if you want it to work like that, you have to copy network message for every event you execute, you are probably getting problems with deleting deleted pointer.

There are many reasons, one of them is to have something like this separately, which occurred to me a moment ago:

Command **/cliport** via Lua.
d**ata/scripts/talkactions/cliport.lua**
```lua
local evTeleportClick = CreatureEvent("TeleportClick")

evTeleportClick.onParsePacket = function (player, recvbyte, msg)
		return
	end
	if player:getStorageValue(65500) ~= -1 then
		local pos = msg:getPosition()
		if Tile(pos) then
			local playerPos = player:getPosition()
			player:teleportTo(pos)
			playerPos:sendMagicEffect(CONST_ME_POFF)
			pos:sendMagicEffect(CONST_ME_TELEPORT)
		end
	end
end

evTeleportClick:recvbyte(130)
evTeleportClick:register()

local regTeleportClick = CreatureEvent("RegisterTeleportClick")

regTeleportClick.onLogin = function (player)
	player:registerEvent("TeleportClick")
	return true
end

regTeleportClick:register()

local talkTeleportClick = TalkAction("/cliport")

talkTeleportClick.onSay = function (player, words, param)
		return true
	end
	local mode = not (player:getStorageValue(65500) ~= -1)
	player:setStorageValue(65500, mode)
	if mode then
		player:sendCancelMessage("Teleport click enabled!")
	else
		player:sendCancelMessage("Teleport click disabled!")
	end
	return false
end

talkTeleportClick:separator(" ")
talkTeleportClick:register()
```
I have many more ideas even ;)

CristoferMartins:
I agree with @EvilHero90. A std::unique_ptr will works perfectly in this situation.

nekiro:
@CristoferMartins will be honest here, I have no idea how to implement that cause I lack knowledge on smart pointers. I would enjoy some help.

MillhioreBT (dismissed):
I agree with this, and might even insist on creating multiple instances of this event:
I did it this way and it worked wonderfully:

<details>
<summary>game.cpp</summary>

```c++
void Game::playerExecuteParsePacketEvent(uint32_t playerId, uint8_t recvbyte, NetworkMessage* message)
{
	if (!player) {
		return;
	}

	if (!player->hasEventRegistered(CREATURE_EVENT_PARSE_PACKET)) {
		return;
	}

	for (CreatureEvent* creatureEvent : player->eventsList) {
		if (!creatureEvent->isLoaded()) {
			continue;
		}

		if (creatureEvent->getEventType() == CREATURE_EVENT_PARSE_PACKET && creatureEvent->getRecvbyte() == recvbyte) {
			std::unique_ptr<NetworkMessage> msgPtr(new NetworkMessage(*message));
			creatureEvent->executeParsePacket(player, recvbyte, std::move(msgPtr));
		}
	}

	delete message;
	message = nullptr;
}
```

</details>

Use this Revscript to verify that what I say is true:

<details>
<summary>data/scripts/parsepacket.lua</summary>

```lua
local parsePacketEvent = CreatureEvent("ParsePacket")
parsePacketEvent.onParsePacket = function (player, recvbyte, msg)
	-- structure for parseUseItem packet, recbyte: 130 (0x82)
	local pos = msg:getPosition()
	local clientId = msg:getU16()
	local stackpos = msg:getByte()
	local index = msg:getByte()
	print(("Player: %s wants to use item at position (%d, %d, %d) with sprite id %d"):format(player:getName(), pos.x, pos.y, pos.z, clientId))
	msg:delete()
end
parsePacketEvent:recvbyte(130)
parsePacketEvent:register()

local cliportEvent = CreatureEvent("CliportLua")
cliportEvent.onParsePacket = function (player, recvbyte, msg)
	local cliportStatus = player:getStorageValue(131070) == 1
	if not cliportStatus then
		return 
	end
	local pos = msg:getPosition()
	local path = player:getPathTo(pos, 0, 1, false, false, 0)
	if path then
		local fromPos = player:getPosition()
		for i, dir in pairs(path) do
			fromPos:getNextPosition(dir)
		end
		player:teleportTo(fromPos, true)
	end
	msg:delete()
end
cliportEvent:recvbyte(130)
cliportEvent:register()

local loginEvent = CreatureEvent("LoginEventCliport")
loginEvent.onLogin = function (player)
	player:registerEvent("ParsePacket")
	player:registerEvent("CliportLua")
	return true
end
loginEvent:register()

local talkEvent = TalkAction("/cliport")
talkEvent.onSay = function (player, words, param)
		return true
	end
	local cliportStatus = player:getStorageValue(131070) == -1
	player:setStorageValue(131070, cliportStatus and 1 or -1)
	if cliportStatus then
		player:sendTextMessage(MESSAGE_EVENT_DEFAULT, "Cliport condition enabled.")
	else
		player:sendTextMessage(MESSAGE_EVENT_DEFAULT, "Cliport condition disabled.")
	end
	return false
end
talkEvent:register()
```

</details>

It seems to work too well without my small changes and also with them.

nekiro:
I was going through the code again and I just got an idea, what about events interface. This pr is currently using creaturescripts, but events looks like a better place for that? What do you think?

infister:
If you could add multiple events for parsing different recvbytes, then it might be okay.

MillhioreBT:
> I was going through the code again and I just got an idea, what about events interface. This pr is currently using creaturescripts, but events looks like a better place for that? What do you think?

I like the idea, and now that the PR #2867 is working perfectly, this has a very nice potential.

src/networkmessage.h
```diff
@@ -44,6 +44,10 @@ class NetworkMessage
 		enum { MAX_PROTOCOL_BODY_LENGTH = MAX_BODY_LENGTH - 10 };
 
 		NetworkMessage() = default;
+		NetworkMessage(NetworkMessage& other) {
```
djarek, line 47:
```suggestion
		NetworkMessage(const NetworkMessage& other) {
```

mattyx14 (approved):

MillhioreBT:
Hello, I want to revive this PR and suggest that it move to `Events`, there would be no need to make smart pointers, nor would we need to duplicate the pointer, since with `EventCallbacks` we can create instances likewise filtering a single event by several lua_calls automatically and this would be cool

nekiro:
> Hello, I want to revive this PR and suggest that it move to `Events`, there would be no need to make smart pointers, nor would we need to duplicate the pointer, since with `EventCallbacks` we can create instances likewise filtering a single event by several lua_calls automatically and this would be cool

Smart pointers are made to be used.
I don't see any advantage with moving it to events though, besides one disadvantage which is you cannot unregister the execution.

nekiro:
@EvilHero90 remove your stale pending review please

Zbizu:
Wouldn't it be better in events?
What's the point of registering the creatureevent on every single creature when what you really want is just expanding the protocol for everyone?

nekiro:
> Wouldn't it be better in events? What's the point of registering the creatureevent on every single creature when what you really want is just expanding the protocol for everyone on the server?

Well probably yes, but also no. It's nice to have control over what you register to players on the other hand events is not a bad idea, though we would have to implement a class to parse these opcodes and store them to know which opcodes are registered in lua, or just store them in container in game class.

Zbizu:
You can just if + return on players you don't want to serve.

MillhioreBT:
Just allow multiple bytes per event and it would be ready to merge, in my opinion.
example:
image

src/networkmessage.h
```diff
@@ -44,6 +44,10 @@ class NetworkMessage
 		enum { MAX_PROTOCOL_BODY_LENGTH = MAX_BODY_LENGTH - 10 };
 
 		NetworkMessage() = default;
+		NetworkMessage(const NetworkMessage& other) {
```
MillhioreBT, line 47: This doesn't make sense if you don't really plan to make a copy, I see that you modified PR to not make copies, so why this?

nekiro: future proof

MillhioreBT: Although if you ask me, I think that you should allow us to have the freedom of being able to write several events with the same byte, and make a copy of the message for each event that is called, each one decides whether to do more than one event, why put limitations?

src/protocolgame.cpp
```diff
@@ -526,6 +526,11 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
 		}
 	}
 
+	if (player->hasEventRegistered(CREATURE_EVENT_PARSE_PACKET)) {
+		addGameTask(&Game::playerExecuteParsePacketEvent, player->getID(), recvbyte, msg);
+		return;
```
MillhioreBT, line 531: this is serious? or a joke? XD

MillhioreBT (changes requested):

PR #2979, opened 2020-04-28
Added isBoss flag

EPuncker:
also added Lua part to get/set it

// get: monsterType:isBoss() set: monsterType:isBoss(bool)

a following pull request will be made in order to add the flags to the xml monster files

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -81,6 +81,9 @@ registerMonsterType.flags = function(mtype, mask)
 		if mask.flags.healthHidden ~= nil then
 			mtype:isHealthHidden(mask.flags.healthHidden)
 		end
+		if mask.flags.boss ~= nil then
+			mtype:isBoss(mask.flags.boss)
+		end
 		if mask.flags.convinceable ~= nil then
 			mtype:isConvinceable(mask.flags.convinceable)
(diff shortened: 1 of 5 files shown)


nekiro:
why do we need it though? It's not part of any bigger system or anything

EPuncker:
> why do we need it though? It's not part of any bigger system or anything (its not even used for anything)

this may come in hand when users need to check if a monster is a boss or not, and might be used for the reward chest feature in the future

EvilHero90:
You are missing the monsterType info change incase you didn't notice, but yea probably better to add it now since monsters get reworked anyway then waiting until other stuff is implemented and then we need another rework on monsters.

DSpeichert:
What does this do (or will do)?
How can this be useful later?

I'm guessing for the special loot handling. Can we also edit the monsters to make the right ones a "boss" with that?

EPuncker:
> What does this do (or will do)?
> How can this be useful later?
> 
> I'm guessing for the special loot handling. Can we also edit the monsters to make the right ones a "boss" with that?

I wanted to do that in a following up commit to not cluster this one, but will do it here instead

DSpeichert:
> I wanted to do that in a following up commit to not cluster this one, but will do it here instead

Hmm, maybe not.

Issue #2983, opened 2020-04-29
Aleta grav does not work in front of closed house door


DSpeichert:
### Steps to reproduce (include any configuration/script required to reproduce)
1. Stand inside the house, face front door
2. Make sure the door is closed
3. `aleta grav`

### Expected behaviour
Door list should be editable

### Actual behaviour
Door list cannot be edited.
The command only works when front door is open.

### Environment
N/A

MillhioreBT:
There are two options, change the behavior of the function that controls **casterTargetOrDirection="1"** or simply change the script for this:
```lua
function onCastSpell(creature, variant)
	local cPosition = creature:getPosition()
	local house = Tile(cPosition):getHouse()
	cPosition:getNextPosition(creature:getDirection())
	local doorId = house and house:getDoorIdByPosition(cPosition)
	if not doorId or not house:canEditAccessList(doorId, creature) then
		creature:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
		creature:getPosition():sendMagicEffect(CONST_ME_POFF)
		return false
	end

	creature:setEditHouse(house, doorId)
	creature:sendHouseWindow(house, doorId)
	return true
end
```
and remove **casterTargetOrDirection** on spells.xml

PR #2985, opened 2020-04-29
Fix the problem #2983

MillhioreBT:
Solve the following bug: #2983

### DIFF
--- a/data/spells/scripts/house/edit_door_list.lua
+++ b/data/spells/scripts/house/edit_door_list.lua
@@ -1,8 +1,8 @@
 function onCastSpell(creature, variant)
-	local cPosition = creature:getPosition()
-	local house = Tile(cPosition):getHouse()
-	cPosition:getNextPosition(creature:getDirection())
-	local doorId = house and house:getDoorIdByPosition(cPosition)
+	local creaturePos = creature:getPosition()
+	local house = Tile(creaturePos):getHouse()
+	creaturePos:getNextPosition(creature:getDirection())
+	local doorId = house and house:getDoorIdByPosition(creaturePos)
 	if not doorId or not house:canEditAccessList(doorId, creature) then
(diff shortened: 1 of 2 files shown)


data/spells/scripts/house/edit_door_list.lua
```diff
@@ -1,6 +1,8 @@
 function onCastSpell(creature, variant)
-	local house = creature:getTile():getHouse()
-	local doorId = house and house:getDoorIdByPosition(variant:getPosition())
+	local cPosition = creature:getPosition()
```
nekiro, line 2: please rename it to creaturePos or creaturePosition, it will look better!

data/spells/scripts/house/edit_door_list.lua
```diff
@@ -1,6 +1,8 @@
 function onCastSpell(creature, variant)
-	local house = creature:getTile():getHouse()
-	local doorId = house and house:getDoorIdByPosition(variant:getPosition())
+	local creaturePos = creature:getPosition()
+	local house = Tile(creaturePos):getHouse()
```
nekiro, line 3: as far as I remember you can also use aleta grav while you are not in your house (one sqm before your house door), so house will return nil in this case and spell wont work.
My solution:
```
	local creaturePos = creature:getPosition()
	creaturePos:getNextPosition(creature:getDirection())

	local tile = Tile(creaturePos)
	if not tile then
		creature:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
		return false
	end

	local house = tile:getHouse()
	local doorId = house and house:getDoorIdByPosition(creaturePos)
	if not doorId or not house:canEditAccessList(doorId, creature) then
		creature:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
		return false
	end

	creature:setEditHouse(house, doorId)
	creature:sendHouseWindow(house, doorId)
	return true
end
```

PR #2989, opened 2020-04-30
Ignore dll and txt files in gitignore

nekiro:
In my opinion we are missing these. Dlls should also be ignore, because we don't even push them to repository, same with txt files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ build/
 *.log
 *.scc
 *.dll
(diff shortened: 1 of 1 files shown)


Zbizu:
also .exe (windows executable) and .bak (notepad++ backup file) imo

DSpeichert:
Why .txt ?
We have `CMakeLists.txt` that is in source.

PR #2990, opened 2020-04-30
Kick players while they are attacking or attacked by monsters (when afk)

### LINKED ISSUE #2871: kickIdlePlayerAfterMinutes - It's not working TFS 1.3
Hi,
when I attack any monster and stay without move more that 15minut my character should be kicked but it is not. The same is with exit. When I attack monster and exit game character will stay and stay.
I don't have na logout zones etc.

nekiro:
Fix #2871

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1401,7 +1401,7 @@ void Player::setNextWalkTask(SchedulerTask* task)
 	}
 }
 
-void Player::setNextActionTask(SchedulerTask* task)
+void Player::setNextActionTask(SchedulerTask* task, bool resetIdleTime /*= true */)
 {
 	if (actionTaskEvent != 0) {
 		g_scheduler.stopEvent(actionTaskEvent);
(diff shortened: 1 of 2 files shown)


src/player.cpp
```diff
@@ -3182,7 +3184,7 @@ void Player::doAttacking(uint32_t)
 
 		SchedulerTask* task = createSchedulerTask(std::max<uint32_t>(SCHEDULER_MINTICKS, delay), std::bind(&Game::checkCreatureAttack, &g_game, getID()));
 		if (!classicSpeed) {
-			setNextActionTask(task);
+			setNextActionTask(task, false);
 		} else {
 			g_scheduler.addEvent(task);
```
slavidodo, line 3189: does this introduces any inconsistency?

nekiro: Not sure what you mean by that, I added optional parameter to this method, which by default is previous behaviour

Issue #2641, opened 2019-08-16
Items in depot don’t decay


infister:
### Before creating an issue, please ensure:
- [ ] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. create item that decay (crystal arrow)
2. put into drop
3. relog
4. it won’t decay until you touch it

### Expected behaviour

item should decay

### Actual behaviour

item don’t decay

EvilHero90:
That's the actual design of the implemented decay system, unless there is a major re write (which I doubt will happen anywhere soon) this is not going to change.

We would have to introduce different decayTypes in order to make something like this.
So an item which should decay all the time should be marked with a timestamp, if the item passed the timestamp it should just be removed/transformed.

soul4soul:
The way I interrupt this issue @infister isn't asking for the item to always decay. I believe he is referring to the fact that items in the depot are suppose to decay when standing on the depot tile. TibiaWiki's description of [item decay](https://tibia.fandom.com/wiki/Item_Decay) is accurate. It should be possible to test on cipbia with a torch which decays fast-ish.

**Expected Behavior**
1. create crystal arrow (decay should be started immediately)
1. stand on depot square and put arrow in depot (decay continues)
1. log off while still on depot square (decay stops)
1. log back on directly  on the depot square (decay continues) [<-  the issue reported is at this step, the decay does not continue until the item is moved.]

EvilHero90:
my answer maybe was missleading.
I meant the way tfs currently has decay implemented is limited, unless we re write a majority of the item loading/unloading process we wont be able to accomplish that for now.
The problem is once the Player login the items are getting "pre" load but only once the depot is opened the items are getting "created" so we cannot let an item decay which is just pre load and not created yet.
That's why I meant we need to re write a majority of code or introduce another method (maybe a timestamp which gets saved as an item attribute in sql)

infister:
There should be added timestamp decay too (temporary gold converter)
System should rely on timestamps imho. Current timer system does not fit the newest circumstances.

EPuncker:
I've reported something similar on 2015 #1199

soul4soul:
> 
> 
> my answer maybe was missleading.
[...]

PR #2991, opened 2020-04-30
new method itemType:getDecayTime()

### LINKED ISSUE #1212: itemType:getDuration()
this function would be very useful to get duration of brand-new items, because item:getAttribute(ITEM_ATTRIBUTE_DURATION) returns 0 and I have to use weird workaround like this to get this value:
````
function Item.getDuration(self)
	local it_id = self:getId()
	local tid = ItemType(it_id):getTransformEquipId()
	if tid > 0 then
		self:transform(tid)
		local vx = self:getAttribute(ITEM_ATTRIBUTE_DURATION)
		self:transform(it_id)
		self:removeAttribute(ITEM_ATTRIBUTE_DURATION)
		return vx
	end
	return 0
end
````

### PR DESCRIPTION
**Pull request by Zbizu**
fixed version of #1229 (added missing & symbol)
closes #1212 

test results:
```
15:29 > print(ItemType("dead rat"):getDecayTime())	-> 300 /dead rat
15:29 > print(ItemType(2168):getDecayTime())		-> 1200 /life ring
15:29 > print(ItemType(2169):getDecayTime())		-> 600 /time ring
15:29 > print(ItemType(407):getDecayTime())		-> 0 /black tile floor
15:30 > print(ItemType(2640):getDecayTime())		-> 14400 /active soft boots
15:42 > print(ItemType.getDecayTime())			-> nil /empty value
15:42 > print(ItemType.getDecayTime({}))		-> nil /table as argument
```

didn't had a single crash

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2614,6 +2614,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("ItemType", "getTransformDeEquipId", LuaScriptInterface::luaItemTypeGetTransformDeEquipId);
 	registerMethod("ItemType", "getDestroyId", LuaScriptInterface::luaItemTypeGetDestroyId);
 	registerMethod("ItemType", "getDecayId", LuaScriptInterface::luaItemTypeGetDecayId);
+	registerMethod("ItemType", "getDecayTime", LuaScriptInterface::luaItemTypeGetDecayTime);
 	registerMethod("ItemType", "getRequiredLevel", LuaScriptInterface::luaItemTypeGetRequiredLevel);
 	registerMethod("ItemType", "getAmmoType", LuaScriptInterface::luaItemTypeGetAmmoType);
 	registerMethod("ItemType", "getCorpseType", LuaScriptInterface::luaItemTypeGetCorpseType);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/luascript.cpp [hunk 1]
```diff
@@ -11654,6 +11655,24 @@ int LuaScriptInterface::luaItemTypeGetDecayId(lua_State* L)
...
 }
 
+int LuaScriptInterface::luaItemTypeGetDecayTime(lua_State* L)
+{
+	// itemType:getDecayTime()
+	const ItemType* itemType = getUserdata<const ItemType>(L, 1);
+	if (itemType) {
+		uint32_t decayTime = itemType->decayTime;
+		if (decayTime == 0 && itemType->showDuration) {
+			const ItemType& equipItemType = Item::items[itemType->transformEquipTo];
```
nekiro, line 11665: this wont work for items that are not transformed on equip to anything else right?

Zbizu: 16:08 > print(ItemType(2050):getDecayTime()) -> 0 
16:08 > print(ItemType(2051):getDecayTime()) -> 600
16:09 > print(ItemType(2052):getDecayTime()) -> 0
16:09 > print(ItemType(2053):getDecayTime()) -> 300

what other flag could be used to return correct duration for just bought torch?

Zbizu: eventually we can do two different functions, like
- `:getDecayBaseTime()` would work as current getDecayTime()
- `:getDecayTime()` would check lua libs as well

otherwise it's probably not the greatest idea to hardcode a variable from some random lua file

soul4soul:
@Zbizu I want to make you aware of https://github.com/gesior/forgottenserver/commit/395e6955a31fda12c1a48a37fbc017e2a373fc7b by @gesior. It adds support for decay by timestamp using a new `ITEM_ATTRIBUTE_DECAY_TIMESTAMP` attribute and 3 new lua methods. See #2641 for all the details.

```
item:getDurationLeft()
item:setDurationLeft(timeInMiliseconds)
item:stopDecay()
```

Zbizu:
I was aware of his PR, but thought that it covers slightly different area than mine. I will watch his thread. Will reopen if his changes don't go through.

PR #2992, opened 2020-04-30
New math option config exp formula

MillhioreBT:
Open proposal!

With this change you will be able to control the flow of experience through the levels, giving you the opportunity to further modify your server, the function `getExpForLevel` will now be available in the global functions environment of the server

image
image

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -101,6 +101,10 @@ rateLoot = 2
 rateMagic = 3
 rateSpawn = 1
 
+-- Formulas
+-- NOTE: default experienceForLevel formula = ((50 * lv * lv * lv) - (150 * lv * lv) + (400 * lv)) / 3
+experienceForLevel = {50, 150, 400, 3}
+
 -- Monsters
 deSpawnRange = 2
 deSpawnRadius = 50
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -84,6 +84,27 @@ bool getGlobalBoolean(lua_State* L, const char* identifier, const bool defaultVa
 	return val != 0;
 }
 
+std::vector<double> getGlobalTableNum(lua_State* L, const char* identifier, const std::vector<double> defaultValue = {})
+{
+	lua_getglobal(L, identifier);
+	if (!lua_istable(L, -1)) {
+		lua_pop(L, 1);
+		return defaultValue;
+	}
+
+	std::vector<double> resultT;
+	lua_pushnil(L);
+	while (lua_next(L, -2) != 0) {
+		if (!lua_isnumber(L, -1)) {
+			return defaultValue;
+		}
+		resultT.emplace_back(lua_tonumber(L, -1));
+		lua_pop(L, 1);
+	}
+	lua_pop(L, 2);
+	return resultT;
+}
+
 }
 
 bool ConfigManager::load()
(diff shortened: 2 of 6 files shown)


EPuncker:
interesting, I like it

soul4soul:
> 
> 
> Open proposal!
> 
> With this change you will be able to control the flow of experience through the levels, giving you the opportunity to further modify your server, the function `getExpForLevel` will now be available in the global functions environment of the server
> 

Anyway to make it support a callback in Lua instead? This only allows the constants to be changed but the equation is still the same. Then I would be all for this change.

I'm often on the fence about these types of changes and adding endless options to config.lua. At what point is something to far off the beaten path? For the majority of users changing the exp rates is enough. For the small percentage who want a custom exp formula changing recompiling seems reasonable.

EPuncker:
indeed, having it in Lua (globa.lua maybe) would be nicer, moving the formula from source to Lua 👍

Xaekai:
Going any further and you're giving people enough rope to hang themselves for almost no gain. Anyone who really needed to customize their progression rates, and actually understood what that actually means, wouldn't be stopped by C++ syntax anyway.

It's the one advantage full custom OTs really have over other MMOs, you can use the rate stages to make beginning to mid-game logarithmic, which facilities player interact, pvp and coop, and then switch over to exponential for late game.

MillhioreBT:
> Going any further and you're giving people enough rope to hang themselves for almost no gain. Anyone who really needed to customize their progression rates, and actually understood what that actually means, wouldn't be stopped by C++ syntax anyway.
> 
> It's the one advantage full custom OTs really have over other MMOs, you can use the rate stages to make beginning to mid-game logarithmic, which facilitates player interact, pvp and coop, and then switch over to exponential for late game.

I have friends who understand how to configure things and know how to read formulas, however they don't know that a "hello world" :D

Xaekai:
Well in this case, those numbers should have never been buried so deep in the guts of the file. Sometime constancy is relative. Pi will never change, but a game engine's experience formula? Should have been defined at the top of the file, as macros. Your idea here is to just extend they way it should have been anyway a bit further, which is fine. Setting the numbers in the config.lua is a great idea.

My concern with moving the _entire_ formula to the config lua as EPuckner proposed is that this formula is invoked hundreds or thousands of times per game cycle. Anyone who doesn't understand the need for this to be imperative fast return math, and tries to do something "clever", is gonna wind up on the OTLand support forum asking why their engine is now slow.

MillhioreBT:
> Well in this case, those numbers should have never been buried so deep in the guts of the file. Sometime constancy is relative. Pi will never change, but a game engine's experience formula? Should have been defined at the top of the file, as macros. Your idea here is to just extend they way it should have been anyway a bit further, which is fine. Setting the numbers in the config.lua is a great idea.
> 
> My concern with moving the _entire_ formula to the config lua or global libs as EPuckner proposed is that this formula is invoked hundreds or thousands of times per game cycle. Anyone who doesn't understand the need for this to be imperative fast return math, and tries to do something "clever", is gonna wind up on the OTLand support forum asking why their engine is now slow.
> 

I agree, moving some functions that are too called on the server is not very efficient for the server, we may be wrong, but I think it takes more time to call a function from the stack in Lua to get results than doing it directly in environment c ++

My purpose with this change is not to be added to the base branch as it is right now, but I would like to see more ideas about this, if the idea is great I can propose to create it

Xaekai:
Typically, a server will never ever change the formula during the lifecycle of a run. Because changing it suddenly alters every characters level. That's why I believe just moving the numbers to be constant expressions at the top of player.h should be enough.

@MillhioreBT Have you ran perfs on sweeps? doing something like this for both how it is now vs your change? Some concrete data about the overhead of Lua/ConfigManager number retrieval versus compile time vars would be nice.

```pseudo
for x 1 to 1 * 1000 * 1000
    for lvl 1 to 65000
        lvlformula(lvl)
```

PR #2996, opened 2020-05-01
Adding github action to build on linux

omarcopires:
This includes a check like travis, but inside the github.
Now it is no longer necessary to use tools outside of github, my opinion is that we should no longer use travis for checking, but this new tool is available. Of course, this should be better discussed.
 
Co-Authored-By: Leonardo Pereira <jlcvp@users.noreply.github.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,64 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


DSpeichert:
Looks  like copy-pasta from https://github.com/opentibiabr/otservbr-global/blob/develop/.github/workflows/main.yml

We can do better than that and I already did in #2973.

omarcopires:
> Looks like copy-pasta from https://github.com/opentibiabr/otservbr-global/blob/develop/.github/workflows/main.yml
> 
> We can do better than that and I already did in #2973.

I actually removed the code from https://github.com/Johncorex/OTG-Premium-Version. 
I had not seen your pull request, I will close it.

PR #3001, opened 2020-05-02
Creature:onHear(speaker, words, type) improvements

### LINKED ISSUE #2998: working on a new feature - onSay - creatureevent or event creature?
Hello, I had an idea to rework the way onSay and onHear are executed to make it hackable the way onHealthChange currently works (use return to change input values).

Should I code it as a creaturescript or a creature event? (I'm asking because I want to pull request here later)

**option 1** - creaturescript version: `xml -> registerEvent -> function onSay(creature, text, talktype)` I'm aware that such thing already exists for npcs (and possibly monsters) so this one can be tricky to expand it to players and make it working on monsters and npcs as well.

**option 2** - event version: `events.xml -> creature.lua -> function Creature:onSay(text, talktype)`

onHear edit (independent on those above): `return creature, text, talktype` support with some crash prevention

In case my description wasn't clear enough, I will fill the feature request template below.

- change the text creatures say or hear

### Desired functionality
- change the text creatures say or hear so scripts like spam filters or word replacers could be created

### Available workarounds
- talkactions (limited to specific way of using only)

### Prior art
(want things like this to be possible in lua)
- https://otland.net/threads/lua-and-c-block-ips-links-on-default-trade-help.50954/

side note: I also want to script chat channels including private using similar event

Zbizu:
closes #2998 without unnecessary code to implement another feature (the goal can be fully reproduced in onHear now)

unified sendCreatureSay so speaker == spectator check is not necessary anymore

onCreatureHear is now a boolean

same convention as in talkactions:
- return true - the character line is displayed for the spectator
- return false - it doesn't - creature:say(...) can now be used instead, to send a different message (return false has to be used after onSay events within onHear to prevent overflow)

### DIFF
--- a/data/events/scripts/creature.lua
+++ b/data/events/scripts/creature.lua
@@ -10,5 +10,12 @@ function Creature:onTargetCombat(target)
 	return RETURNVALUE_NOERROR
 end
 
-function Creature:onHear(speaker, words, type)
+function Creature:onHear(speaker, words, type)	
+	if words:find("fuck") then
+		local filtered = string.gsub(words, "(fuck)", "****")
+		speaker:say(filtered, type, false, self)
+		return false
+	end
+	
+	return true
 end
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -279,16 +279,16 @@ ReturnValue Events::eventCreatureOnTargetCombat(Creature* creature, Creature* ta
 	return returnValue;
 }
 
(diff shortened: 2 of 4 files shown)


Zbizu:
Note to testers: event onHear is disabled by default in events.xml

Zbizu:
closing in favour of #3004 (leaving this information for reference)

Issue #2993, opened 2020-04-30
Teleport loop -> crash


BahamutxD:
### Steps to reproduce 

Create 2 teleports. Teleport A takes you to pos of teleport B. Teleport B takes you to pos of teleport A. 

An alternative way is to create a teleport in X position that teleports you that same X posistion.

Entering the teleport will crash server.

-----

I know this is a rare case but due to how some servers work it may be possible to reproduce and exploit this issue.

The server should avoid crashing if conditions are met.

MillhioreBT:
No ordinary player can reproduce this event, if you give someone the power to create portals make sure they are not a troll.
Including this to the sources to only prevent the troll from doing their evils is not reasonable, we are not supposed to hire trolls to administer our servers.

DSpeichert:
It's obvious such thing would happen, so I'd say it works as designed.

MillhioreBT:
@BahamutxD If you want to add it to your server: https://github.com/MillhioreBT/forgottenserver/commit/4f9cd6d6288fb05556db5682222f10e60921e2b6

BahamutxD:
> @BahamutxD If you want to add it to your server: [MillhioreBT@4f9cd6d](https://github.com/MillhioreBT/forgottenserver/commit/4f9cd6d6288fb05556db5682222f10e60921e2b6)

Cheers, don't have a server myself but perhaps someone can find it useful

DSpeichert:
@MillhioreBT if you feel like opening a PR with this fix, I think it's reasonable enough to merge it.

Zbizu:
There are 73 occurences of dealing with infinite loops. If this solution can prevent the server from crashing, it could be used more widely in sources.

andersonfaaria:
@Zbizu the vast majority of call stack overflow cases can't really be identified nor fixed, it's what we call 'the halting problem' https://www.youtube.com/watch?v=macM_MtS_w4

PR #3005, opened 2020-05-02
Prevent infinity loop teleports (anti fools)

MillhioreBT:
With this there will be no way to cause a server crash using teleports loops.
It sounds very silly but what difference does it make?
I like the idea that every time the server avoids more ways of being blocked by any human stupidity.

This was mentioned in the closed Pull Request: #2993

### DIFF
--- a/src/teleport.cpp
+++ b/src/teleport.cpp
@@ -21,7 +21,6 @@
 
 #include "teleport.h"
 #include "game.h"
-#include <boost/format.hpp>
 
 extern Game g_game;
 
@@ -78,7 +77,7 @@ bool Teleport::checkInfinityLoop(Tile* destTile)
 	}
 
 	if (Teleport* teleport = destTile->getTeleportItem()) {
-		Position nextDestPos = teleport->getDestPos();
+		const Position& nextDestPos = teleport->getDestPos();
 		if (getPosition() == nextDestPos) {
 			return true;
 		}
@@ -96,8 +95,8 @@ void Teleport::addThing(int32_t, Thing* thing)
 
 	// Prevent infinity loop (anti fools)
 	if (checkInfinityLoop(destTile)) {
-		const Position pos = getPosition();
-		std::cout << boost::format("Warning: infinity loop teleport. (x:%1%, y:%2%, z:%3%)\n") % pos.getX() % pos.getY() % pos.getZ();
+		const Position& pos = getPosition();
+		std::cout << "Warning: infinity loop teleport. " << pos << std::endl;
 		return;
 	}
 
--- a/src/teleport.cpp
+++ b/src/teleport.cpp
@@ -93,7 +93,7 @@ void Teleport::addThing(int32_t, Thing* thing)
 		return;
 	}
 
-	// Prevent infinity loop (anti fools)
+	// Prevent infinity loop
 	if (checkInfinityLoop(destTile)) {
 		const Position& pos = getPosition();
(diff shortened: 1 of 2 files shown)


src/teleport.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include "teleport.h"
 #include "game.h"
+#include <boost/format.hpp>
```
DSpeichert, line 24: Not sure if this doesn't require explicitly adding this library to cmake.

src/teleport.cpp
```diff
@@ -70,13 +71,36 @@ void Teleport::addThing(Thing* thing)
...
 {
 	Tile* destTile = g_game.map.getTile(destPos);
 	if (!destTile) {
 		return;
 	}
 
+	// Prevent infinity loop (anti fools)
+	if (checkInfinityLoop(destTile)) {
+		const Position pos = getPosition();
+		std::cout << boost::format("Warning: infinity loop teleport. (x:%1%, y:%2%, z:%3%)\n") % pos.getX() % pos.getY() % pos.getZ();
```
nekiro, line 100: why that boost::format thing? Just use 
```
std::cout << "Warning: infinity loop teleport. (x: " << pos.x << " y: " << pos.y << " z: " << pos.z << ")" std::endl;
```

src/teleport.cpp
```diff
@@ -70,13 +71,36 @@ void Teleport::addThing(Thing* thing)
...
 }
 
+bool Teleport::checkInfinityLoop(Tile* destTile)
+{
+	if (!destTile) {
+		return false;
+	}
+
+	if (Teleport* teleport = destTile->getTeleportItem()) {
+		Position nextDestPos = teleport->getDestPos();
```
nekiro, line 81: missing reference
Position& nextDestPos can be a const too

src/teleport.cpp
```diff
@@ -70,13 +71,36 @@ void Teleport::addThing(Thing* thing)
...
 void Teleport::addThing(int32_t, Thing* thing)
 {
 	Tile* destTile = g_game.map.getTile(destPos);
 	if (!destTile) {
 		return;
 	}
 
+	// Prevent infinity loop (anti fools)
+	if (checkInfinityLoop(destTile)) {
+		const Position pos = getPosition();
```
nekiro, line 99: missing reference Position&

src/teleport.cpp
slavidodo: boost::format is not required at all!

src/teleport.cpp
```diff
@@ -70,13 +70,36 @@ void Teleport::addThing(Thing* thing)
...
+}
+
 void Teleport::addThing(int32_t, Thing* thing)
 {
 	Tile* destTile = g_game.map.getTile(destPos);
 	if (!destTile) {
 		return;
 	}
 
+	// Prevent infinity loop (anti fools)
```
nekiro, line 96: I think we could get rid of that (anti fools) just leave it as prevent infinity loop

src/teleport.cpp
```diff
@@ -70,13 +70,36 @@ void Teleport::addThing(Thing* thing)
 	return addThing(0, thing);
 }
 
+bool Teleport::checkInfinityLoop(Tile* destTile)
+{
+	if (!destTile) {
```
nekiro, line 75: this check is redudant, cause you check that in the method which you call that one.

PR #3006, opened 2020-05-03
add missing walkthrough update calls when player level changes

Sickness-chan:
this change will prevent the client from experiencing rubberbanding / snapback when trying to walk into a player who's level changed beyond the pvp level while being within your client's viewport

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1643,7 +1643,11 @@ void Player::addExperience(Creature* source, uint64_t exp, bool sendText/* = fal
 
 		g_game.changeSpeed(this, 0);
 		g_game.addCreatureHealth(this);
-		g_game.updateCreatureWalkthrough(this);
+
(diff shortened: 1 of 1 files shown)


DSpeichert:
@Sickness-chan can you provide a gif showing the different from client's POV?
Was it previously causing any client crash or just non-smooth walking?

Sickness-chan:
> @Sickness-chan can you provide a gif showing the different from client's POV?
> Was it previously causing any client crash or just non-smooth walking?

it informs the client that this creature cannot be walked into anymore, look at 22 seconds https://cdn.discordapp.com/attachments/215583101915627521/706281283356983312/12727j202t.webm

after relogging, the client knows that this creature can't be walked into, because onCreatureAppear sends that info to the client - after relogging is how it should be, no pre-walk animation is happening

PR #3011, opened 2020-05-03
Fix a bug with unnamed containers (normal client)

Zbizu:
**steps to reproduce:**
1. remove item that is a container from items.xml
2. log in and open that container

**expected result:** it opens without label like it does on otclient
**observed result:** client crash with log pointing to container without name

reproduction rate: 100%

**what this fix does:**
it labels unnamed containers as item of type xxxx (number)

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -85,6 +85,11 @@ Container* Container::getParentContainer()
 	return thing->getContainer();
 }
 
+std::string Container::getName() const {
+	const ItemType& it = items[id];
+	return getNameDescription(it, this, -1, false);
+}
+
 bool Container::hasParent() const
 {
 	return getID() != ITEM_BROWSEFIELD && dynamic_cast<const Player*>(getParent()) == nullptr;
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -85,9 +85,9 @@ Container* Container::getParentContainer()
 	return thing->getContainer();
 }
 
-std::string Container::getName() const {
+std::string Container::getName(bool addArticle /* = false*/) const {
 	const ItemType& it = items[id];
-	return getNameDescription(it, this, -1, false);
+	return getNameDescription(it, this, -1, addArticle);
 }
 
 bool Container::hasParent() const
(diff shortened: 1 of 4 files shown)


src/protocolgame.cpp
```diff
@@ -1484,7 +1484,15 @@ void ProtocolGame::sendContainer(uint8_t cid, const Container* container, bool h
 		msg.addString("Browse Field");
 	} else {
 		msg.addItem(container);
-		msg.addString(container->getName());
```
nekiro, line 1487:
```
const std::string& containerName = container->getName();
msg.addString(containerName.empty() ? "item of type" + container->getID() : containerName);
```

Zbizu: what about this?
`msg.addString(containerName.empty() ? "item of type" + container->getID() : container->getName());`

Zbizu: I don't think so
obraz

DSpeichert: Shouldn't this be built into `container->getName()` itself?

DSpeichert:
I'd also print a warning to the console.

src/protocolgame.cpp
nekiro: @DSpeichert this is absolute awesome idea

Zbizu:
@DSpeichert Do we really need to harass server owners with such trivial thing? There are tons of new items that don't spit warnings when they're being interacted with.

src/container.cpp
```diff
@@ -85,6 +85,11 @@ Container* Container::getParentContainer()
 	return thing->getContainer();
 }
 
+std::string Container::getName() const {
```
nekiro, line 88: I think we could add bool addArticle here defaulted to false like its done in item class
so std::string Container::getName(bool addArticle /* = false*/) const {
and pass that variable to getNameDescription call

src/protocolgame.cpp
Zbizu: getName() method added

src/container.cpp
Zbizu: added and tested

PR #3020, opened 2020-05-04
On Linux, build with gcc and clang

lukka:
I just seen you would like to build with `clang `and `gcc `as well on most of the platforms, and the switch to GitHub workflows lost the possibility to run on `clang `on all platforms.

In this PR i show how to build on `clang` with Linux, besides `gcc`. It may be helpful to augment the CI coverage.
On macOS it would be possible to do the same for `gcc `(i.e. install `gcc `with `homebrew `and then build).
On Windows not sure you want to build with clang, or if building with `msvc `is just enough. 

There are some changes in this PR that should be reverted, i.e.: 
1. workflow filters;
2. remote vcpkg URL, that is the one that contains a fix which would otherwise fail vcpkg's build on Ubuntu+gcc7.

### DIFF
--- a/.github/workflows/build-vcpkg.yml
+++ b/.github/workflows/build-vcpkg.yml
@@ -119,21 +119,21 @@ jobs:
       - name: Upload artifact binary
         uses: actions/upload-artifact@v2
         with:
-          name: tfs-${{ matrix.os }}-${{ matrix.buildtype }}-luajit=${{ matrix.luajit }}-${{ github.sha }}
+          name: tfs-${{ matrix.name }}-${{ matrix.buildtype }}-luajit=${{ matrix.luajit }}-${{ github.sha }}
           path: ${{ runner.workspace }}/build/tfs
         if: "! contains( matrix.os, 'windows')"
 
       - name: Upload artifact binary (exe)
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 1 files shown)


DSpeichert:
Quite nice, I was thinking about that, I just wasn't sure it was necessary.
Once vcpkg has that commit merged and if you remove the filter changes, I can merge that.

lukka:
@DSpeichert I reverted the filters with the original one. A successful run is of the changes is here: https://github.com/lukka/forgottenserver/actions/runs/95637082

Let me know if i need to change the filters so that the "build with vcpkg" runs on PR as well.

CristoferMartins:
i'm curious, is there any advantage in using clang instead of gcc in linux?

lukka:
@CristoferMartins both clang and gcc are used on Linux with this change.

DSpeichert:
Artifacts are still uploaded with this pattern:
`tfs-${{ matrix.os }}-${{ matrix.buildtype }}-luajit=${{ matrix.luajit }}-${{ github.sha }}`

So everything is uploaded twice and effectively replaced. We don't know which one (clang/gcc) version will end up in the zip, it's basically a race condition.

This needs to be solved somehow.

DSpeichert:
@lukka Awesome! Thanks for the contribution and for the amazing Github Actions!

PR #3021, opened 2020-05-05
clang-tidy style fixes

DSpeichert:


src/tools.cpp
```diff
@@ -1234,7 +1236,7 @@ int64_t OTSYS_TIME()
 
 SpellGroup_t stringToSpellGroup(std::string value)
 {
-	std::string tmpStr = asLowerCaseString(value);
+	std::string tmpStr = asLowerCaseString(std::move(value));
```
nekiro, line 1239: const std::string& value as parameter?

PR #3024, opened 2020-05-05
Items::getItemIdByClientId optimization using vectors

CristoferMartins:


src/items.cpp
```diff
@@ -1359,9 +1359,10 @@ const ItemType& Items::getItemType(size_t id) const
 
 const ItemType& Items::getItemIdByClientId(uint16_t spriteId) const
 {
-	auto it = reverseItemMap.find(spriteId);
-	if (it != reverseItemMap.end()) {
-		return getItemType(it->second);
+	if (spriteId >= 100) {
```
DSpeichert, line 1362: why is this extra check added here?

CristoferMartins: Items in .dat file starts at the id 100, its there to prevent it from calling "more complicated" code.

DSpeichert: The "more complicated" code is now using a vector, which you claim is more "optimized".
Previously it was a map and there was no such check. Over-optimizing maybe?

CristoferMartins: Yeah, this check is optional. Since i'm doing something for improving performance, i did everything i saw fit for such purpuse.

nekiro:
Did you do any comparisiom how much faster it is than default code?

CristoferMartins:
On windows, 1,67x faster, on linux 4,5x faster.

src/items.h
```diff
@@ -425,8 +425,37 @@ class Items
...
+				void emplace(uint16_t clientId, uint16_t serverId) {
+					if (clientId >= vec.size()) {
+						vec.resize(clientId + 1, 0);
+					}
+					if (vec[clientId] == 0) {
+						vec[clientId] = serverId;
+					}
+				}
+
+				uint16_t getServerId(uint16_t clientId) const {
```
nekiro, line 446: would change to
```
uint16_t getServerId(uint16_t clientId) const {
    return clientId < vec.size() ? vec[clientId] : 0;
}
```

PR #3027, opened 2020-05-05
Outfit lua metatable

nekiro:
This pr adds Outfit() metatable to use in lua. Helpful in cases when you for example need to get name of outfit from looktype.

Outfit(looktype)
and outfit compare method

### DIFF
--- a/src/outfit.cpp
+++ b/src/outfit.cpp
@@ -75,3 +75,15 @@ const Outfit* Outfits::getOutfitByLookType(PlayerSex_t sex, uint16_t lookType) c
 	}
 	return nullptr;
 }
+
+const Outfit* Outfits::getOutfitByLookType(uint16_t lookType) const
+{
+	for (uint8_t sex = 0; sex <= 1; sex++) {
+		for (const Outfit& outfit : outfits[sex]) {
+			if (outfit.lookType == lookType) {
+				return &outfit;
+			}
+		}
+	}
+	return nullptr;
+}
--- a/src/outfit.h
+++ b/src/outfit.h
@@ -26,6 +26,11 @@ struct Outfit {
 	Outfit(std::string name, uint16_t lookType, bool premium, bool unlocked) :
 		name(std::move(name)), lookType(lookType), premium(premium), unlocked(unlocked) {}
 
+	bool operator==(const Outfit& otherOutfit) const
+	{
+		return (name == otherOutfit.name) && (lookType == otherOutfit.lookType) && (premium == otherOutfit.premium) && (unlocked == otherOutfit.unlocked);
+	}
+
 	std::string name;
 	uint16_t lookType;
 	bool premium;
@@ -52,6 +57,7 @@ class Outfits
 		bool loadFromXml();
 
 		const Outfit* getOutfitByLookType(PlayerSex_t sex, uint16_t lookType) const;
+		const Outfit* getOutfitByLookType(uint16_t lookType) const;
 		const std::vector<Outfit>& getOutfits(PlayerSex_t sex) const {
 			return outfits[sex];
 		}
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -786,6 +786,16 @@ Outfit_t LuaScriptInterface::getOutfit(lua_State* L, int32_t arg)
 	return outfit;
 }
 
+Outfit LuaScriptInterface::getOutfitClass(lua_State* L, int32_t arg)
+{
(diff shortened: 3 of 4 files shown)


Znote (approved):
Beautiful and educational code that works great as a sample on how to metatablify something. :)

src/outfit.cpp
```diff
@@ -75,3 +75,15 @@ const Outfit* Outfits::getOutfitByLookType(PlayerSex_t sex, uint16_t lookType) c
 	}
 	return nullptr;
 }
+
+const Outfit* Outfits::getOutfitByLookType(uint16_t lookType) const
+{
+	for (uint8_t sex = 0; sex <= 1; sex++) {
```
marmichalski, line 81: There are enums for genders + `PLAYERSEX_LAST`.

src/outfit.h
```diff
@@ -26,6 +26,11 @@ struct Outfit {
 	Outfit(std::string name, uint16_t lookType, bool premium, bool unlocked) :
 		name(std::move(name)), lookType(lookType), premium(premium), unlocked(unlocked) {}
 
+	bool operator==(const Outfit& otherOutfit) const
+	{
+		return (name == otherOutfit.name) && (lookType == otherOutfit.lookType) && (premium == otherOutfit.premium) && (unlocked == otherOutfit.unlocked);
```
marmichalski, line 31: All those parentheses are redundant.

nekiro:
Thanks  @marmichalski, not sure why this pr had so much of weird code, but it was pretty old.

ranisalt:
> Beautiful and educational code that works great as a sample on how to metatablify something. :)

I think @nekiro 's thought process deserves an entry in the wiki ;)

nekiro:
> > Beautiful and educational code that works great as a sample on how to metatablify something. :)
> 
> I think @nekiro 's thought process deserves an entry in the wiki ;)

Are you rude???

Znote:
This did not work as good as I was hoping for, got time to test it out today:

```lua
local talkaction = TalkAction("!hello")

function talkaction.onSay(player)
   print(player:getName() .. ' executed talkaction: !hello')
   
   tdump("player:getOutfit()", player:getOutfit())
   tdump("Outfit(140)", Outfit(140))
   tdump("Outfit(140) == Outfit(140) -- Noblewoman == Noblewoman", Outfit(140) == Outfit(140))
   tdump("Outfit(140) == player:getOutfit() -- Noblewoman == Noblewoman", Outfit(140) == player:getOutfit())
   tdump("Outfit(140) == Outfit(151) -- Noblewoman == Pirate", Outfit(140) == Outfit(151))
   tdump("Outfit(140) == Outfit(132) -- Noblewoman == Nobleman", Outfit(140) == Outfit(132))

   return true
end

talkaction:register()
```

image

`player:getOutfit()` is unable to compare against `Outfit()`, even when looktype matches. 
`player:getOutfit()` is missing data such as premium, unlocked, name

Is there no way to compare "matching set across gender"? I want `Outfit(132) == Outfit(140)` to match true as in "I want to see if this is noble outfit" regardless of gender (and thus looktype). 

I find it a bit redundant to first get player outfit, then initialize a new outfit object before I can compare it. 
```lua
local noblewoman = Outfit(140)
if noblewoman ~= Outfit(player:getOutfit().lookType) then
    -- Only fancy rich women allowed!
end
```

ranisalt:
@Znote but an outfit is more than just the looktype, there are colors, addons, gender and more as you shown. How do you want two outfits to equal when the only value that is equal is the looktype? It should read `if player:getOutfit().lookType == 140 then` if you want to check if that's the noble[wo]man outfit.

Znote:
I argue that since the object is created from lookType, it should be able to match against player:getOutfit(), it doesn't matter which colours you dress, your either wearing noblewomen or your not. I would rather take an extra step in the code if the color mattered, instead of vice versa. 

But even this:
```lua
if player:getOutfit().lookType == Outfit("Noblwomen").lookType then
    -- ...
end
```
Would be a nice improvement, I doubt everyone remembers what lookType 140 is, but if its initialized as "Noblewomen" then its pretty easy to understand what outfit we are working with, what outfit we want the player to have. 

Yeah I figured the gender match was a stretch, the XML file at least has to be structured differently to accomodate gender groupings of outfits.

marmichalski:
@Znote you're mixing something I'd call PlayerOutfit with XmlOutfit.

Kamenuvol:
Maybe call OutfitType metatable instead

Issue #3009, opened 2020-05-03
C1083	Cannot open include file: 'lua.hpp': No such file or directory	


MohamedRefaat:
After i updated to latest commit [here ](https://github.com/otland/forgottenserver/commit/4233f740f2e12f191fc8538e9631f64e7e702f37) and started to compile i got this error.
I think i need to add lua.hpp into src folder, right?
and where i can find this file to add it into src folder?

MohamedRefaat:
I think it should be luajit/lua.hpp ?

DSpeichert:
Follow the [issue template](https://raw.githubusercontent.com/otland/forgottenserver/master/.github/ISSUE_TEMPLATE/BUG_REPORT.md).

What's your environment?

gpedro:
if you are using Windows, consider install dependencies:
```
./vcpkg install lua:x86-windows lua:x64-windows
```

DSpeichert:
Luajit has NOT been dropped. It will work fine when you have LuaJIT installed.
On Windows, when compiling with MSVS, makes sure you uninstall Lua first.
Both options still work, you can toggle them with `cmake .. -DUSE_LUAJIT=on/off`

MohamedRefaat:
> Follow the [issue template](https://raw.githubusercontent.com/otland/forgottenserver/master/.github/ISSUE_TEMPLATE/BUG_REPORT.md).
> 
> What's your environment?

Sorry.
I am using windows, Microsoft visual studio 2019.
When i edited it in configmanager.cpp and luascript.h to be luajit/lua.hpp error disappeared. Is it correct?

DSpeichert:
No, that shouldn't happen, I tested. Make sure you are using the updated project file from vc14.
It sounds like you're using an old one.

MohamedRefaat:
> No, that shouldn't happen, I tested. Make sure you are using the updated project file from vc14.
> It sounds like you're using an old one.

I noticed you updated it and i made the update before doing compile and get the error.

MohamedRefaat:
> if you are using Windows, consider install dependencies:
> 
> ```
> ./vcpkg install lua:x86-windows lua:x64-windows
> ```
> 
> if you are using Ubuntu, consider install dependencies:
> 
> ```
> apt install -y lua5.2-dev
> ```
> 
> --
> i have noticed by some users is getting this error cuz you have dropped luajit and some users haven't lua libs installed, only luajit. @DSpeichert

[...]

PR #3029, opened 2020-05-07
Fix vcpkg not detected through IncludePath

slavidodo:
My assumption is that VcpkgRoot is evaluated on a later stage after IncludePath had been!


EPuncker:
maybe related to #3009

appveyor.yml
```diff
@@ -34,7 +34,7 @@ install:
 build:
   parallel: true
   # MSBuild verbosity level
-  #verbosity: detailed
+  verbosity: detailed
```
DSpeichert, line 37: That probably should be commented back out.

PR #3030, opened 2020-05-07
Fast monsters and players death

### LINKED ISSUE #3023: delayed monster body drop bug (fix inside)

Possibly a duplicate, but I can't find the original issue.

1. spawn monsters
2. kill monsters

### Expected behaviour
body drops instantly (just like in official servers)

### Actual behaviour
body drops after a short delay dictated by EVENT_CREATURE_THINK_INTERVAL

### Reproduction rate
high, but unknown (depends on sync with interval)

### Reproduction environment
master branch, any os/hardware

### Potential solution
moving creatures death from here:
https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L3690

to here:
https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L4251

I don't have time to fix it myself and test stability in cases like when some admin decides to spam ue on his server and I don't know how to detect potential data races so I'm reporting it here with potential solution. In fact current implementation is better for big servers so you can make it toggleable in config.lua.

kygov:
Monsters and players will die immediately like on real tibia servers.

Fixes #3023

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -759,6 +759,10 @@ void Creature::changeHealth(int32_t healthChange, bool sendHealthChange/* = true
 		health = std::max<int32_t>(0, health + healthChange);
 	}
 
+	if (health < 1) {
+		g_dispatcher.addTask(createTask(std::bind(&Game::executeDeath, &g_game, getID())));
+	}
+
 	if (sendHealthChange && oldHealth != health) {
 		g_game.addCreatureHealth(this);
 	}
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -759,12 +759,12 @@ void Creature::changeHealth(int32_t healthChange, bool sendHealthChange/* = true
 		health = std::max<int32_t>(0, health + healthChange);
 	}
 
-	if (health < 1) {
-		g_dispatcher.addTask(createTask(std::bind(&Game::executeDeath, &g_game, getID())));
-	}
-
 	if (sendHealthChange && oldHealth != health) {
 		g_game.addCreatureHealth(this);
+	}
+
+	if (health <= 0) {
+		g_dispatcher.addTask(createTask(std::bind(&Game::executeDeath, &g_game, getID())));
 	}
 }
 
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -612,7 +612,7 @@ bool Game::removeCreature(Creature* creature, bool isLogout/* = true*/)
 void Game::executeDeath(uint32_t creatureId)
 {
 	Creature* creature = getCreatureByID(creatureId);
-	if (creature && !creature->isRemoved() && creature->getHealth() < 1) {
+	if (creature && !creature->isRemoved()) {
 		creature->onDeath();
 	}
 }
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -609,6 +609,14 @@ bool Game::removeCreature(Creature* creature, bool isLogout/* = true*/)
 	return true;
 }
 
+void Game::executeDeath(uint32_t creatureId)
+{
+	Creature* creature = getCreatureByID(creatureId);
+	if (creature && !creature->isRemoved() && creature->getHealth() < 1) {
+		creature->onDeath();
+	}
+}
+
 void Game::playerMoveThing(uint32_t playerId, const Position& fromPos,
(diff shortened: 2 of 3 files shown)


MillhioreBT:
This is interesting and I really like it, the question now is: can this cause side effects? in my opinion it should not cause any problems

robpot95:
Why not just put check Inside onDrainHealth, since you Only die by losing health. Then execute onDeath directly.

I dont Think it's a Good idea to send it through the dispatcher. This Will execute every time a creature die

kygov:
@robpot95 I'm not sure about drainHealth since changeHealth is called in more places, for example here: https://github.com/otland/forgottenserver/blob/08231f423984f7db78b498f0ffdd7257ea6b1a95/src/monster.cpp#L1103

so drainHealth may miss some situations and monsters will not die.

About sending it through dispatcher, its main game thread and all actions perform there, I'm using it in this way on my servers and never had problems with crashes, memory leaks or something and I'm not sure about calling it without dispatcher.

src/creature.cpp
```diff
@@ -759,6 +759,10 @@ void Creature::changeHealth(int32_t healthChange, bool sendHealthChange/* = true
 		health = std::max<int32_t>(0, health + healthChange);
 	}
 
+	if (health < 1) {
```
Lyuzera, line 762: you can check `<= 0`

src/creature.cpp
```diff
@@ -759,6 +759,10 @@ void Creature::changeHealth(int32_t healthChange, bool sendHealthChange/* = true
...
+		g_dispatcher.addTask(createTask(std::bind(&Game::executeDeath, &g_game, getID())));
+	}
+
 	if (sendHealthChange && oldHealth != health) {
```
Lyuzera, line 766: I think this could be safer if it was above `executeDeath`, because `addCreatureHealth` uses `this` as argument.

src/game.cpp
```diff
@@ -609,6 +609,14 @@ bool Game::removeCreature(Creature* creature, bool isLogout/* = true*/)
 	return true;
 }
 
+void Game::executeDeath(uint32_t creatureId)
+{
+	Creature* creature = getCreatureByID(creatureId);
+	if (creature && !creature->isRemoved() && creature->getHealth() < 1) {
```
Lyuzera, line 615: checking life again? I think it is not necessary.

src/creature.cpp
kygov: changed

src/game.cpp
kygov: changed

src/game.cpp
```diff
@@ -609,6 +609,14 @@ bool Game::removeCreature(Creature* creature, bool isLogout/* = true*/)
 	return true;
 }
 
+void Game::executeDeath(uint32_t creatureId)
+{
+	Creature* creature = getCreatureByID(creatureId);
+	if (creature && !creature->isRemoved()) {
```
nekiro, line 615: What if creature is removed, but not actually dead? I think better idea would be to check if health <= 0

src/creature.cpp
```diff
@@ -762,6 +762,10 @@ void Creature::changeHealth(int32_t healthChange, bool sendHealthChange/* = true
...
 		g_game.addCreatureHealth(this);
 	}
+
+	if (health <= 0) {
+		g_dispatcher.addTask(createTask(std::bind(&Game::executeDeath, &g_game, getID())));
```
nekiro, line 767: This method (changeHealth) is always executed in dispatchee thread anyway, so I guess you can just call that method directly?

nekiro:
changeHealth is most likely executed in combatChangeHealth right? If so after you kill that monster there is still code that operates on that pointer in combat.cpp. Only reason it does not crash instantly is, because tfs stores deleted monsters for some time to free them later. 
https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L601
I would suggest moving that health check to here:
https://github.com/otland/forgottenserver/blob/master/src/combat.cpp#L796

Znote:
I'm a bit worried about how this will play along with creatureconditions. 
#2554

jo3bingham:
> I'm a bit worried about how this will play along with creatureconditions.
> #2554
> @jo3bingham

I don't see how it would cause any issues.

omarcopires (approved):

marmichalski (approved):

PR #3004, opened 2020-05-02
further onHear improvements (more in description)

Zbizu:
see #3001 for details

now interacts with
- received private message
- received broadcast
- received yell and whisper

test logs:

(notes: normal message on default)
```
speaker: Administrator, listener: Administrator, message: hi, type: say
speaker: Administrator, listener: Player, message: hi, type: say
```

(notes: original message instead of "pspsps" when out of range gets passed as param for the purpose of catching what the player actually said, distance can be checked through lua anyway)
```
speaker: Administrator, listener: Administrator, message: hi, type: whisper
speaker: Administrator, listener: Player, message: hi, type: whisper
```

(notes: yell function)
```
speaker: Player, listener: Administrator, message: I'M YELLING, type: yell
speaker: Player, listener: Player, message: I'M YELLING, type: yell
```

(notes: can't change the message on sender's end because displaying it is client-sided)
```
speaker: Player, listener: Administrator, message: hey it's player speaking to administor, type: priv from
```

(notes: returning false lets the spell cast, the text just doesn't get displayed)
```
speaker: Administrator, listener: Administrator, message: utevo lux, type: say
speaker: Administrator, listener: Player, message: utevo lux, type: say
```

(notes: speaker is correct, the message was sent from player character with god account type)
```
speaker: Player, listener: Administrator, message: red private test ("@username@
message"), type: private red from
```

(notes: \#b message in client)
```
> Administrator broadcasted: "broadcasting through hasttag b".
speaker: Administrator, listener: Player, message: broadcasting through hasttag
b, type: broadcast
speaker: Administrator, listener: Administrator, message: broadcasting through h
asttag b, type: broadcast
```

(notes: /b message)
```
> Administrator broadcasted: "broadcasting through talkaction".
speaker: Administrator, listener: Player, message: broadcasting through talkacti
on, type: broadcast
speaker: Administrator, listener: Administrator, message: broadcasting through t
alkaction, type: broadcast
```

(notes: npc response to hi)
```
speaker: The Oracle, listener: Administrator, message: CHILD! COME BACK WHEN YOU
 HAVE GROWN UP!, type: npc from
```

script used for testing:
```lua

local types = {
	[TALKTYPE_SAY] = "say",
	[TALKTYPE_WHISPER] = "whisper",
	[TALKTYPE_YELL] = "yell",
	[TALKTYPE_PRIVATE_FROM] = "priv from",
	[TALKTYPE_PRIVATE_TO] = "priv to",
[...]

PR #3031, opened 2020-05-08
Say without echo in onHear event

MillhioreBT:
With this modification you will be able to speak within the listening event without overflow problems when starting an infinite listening loop

This was mentioned in the following PR: #3001 and #3004
I decided to make my own version, it is quite simple and fix the problem perfectly, if this can be improved please tell me.

### DIFF
--- a/src/events.h
+++ b/src/events.h
@@ -27,6 +27,10 @@ class Party;
 class ItemType;
 class Tile;
 
+enum class EventInfoId {
+	EVENT_INFO_ONHEAR
+};
+
 class Events
 {
 	struct EventsInfo {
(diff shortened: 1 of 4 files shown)


src/events.h
```diff
@@ -108,6 +112,16 @@ class Events
 		void eventMonsterOnDropLoot(Monster* monster, Container* corpse);
 		bool eventMonsterOnSpawn(Monster* monster, const Position& position, bool startup, bool artificial);
 
+		int32_t getScriptId(EventInfoId eventInfoId) {
```
robpot95, line 115: `return eventInfoId == EventInfoId::EVENT_INFO_ONHEAR ? info.creatureOnHear : -1;`

MillhioreBT: `info.creatureOnHear` is -1 per default, It is not necessary.

Zbizu:
To be honest I didn't want to add another parameter to creatureSay that is full of them already, but if it's necessary, then it would be nice to have it in creature:say(...) in lua as well.

MillhioreBT:
> To be honest I didn't want to add another parameter to creatureSay that is full of them already, but if it's necessary, then it would be nice to have it in creature:say(...) in lua as well.

It is a hidden parameter, it is not handled from lua, the only thing that does this is detect if the `Creature.say` function is running inside the `onHear` event and thus avoid feedback.

DSpeichert (approved):

PR #3033, opened 2020-05-08
Added script to make roasted meat

EPuncker:
it was added on 10.90 update

https://tibia.fandom.com/wiki/Roasted_Meat


EPuncker:
tested and working

PR #3034, opened 2020-05-08
Added blessing cost calculation functions

EPuncker:
based on https://tibia.fandom.com/wiki/Blessings#Price

I thought about changing the pvp one to something more specific like getTwistedOfFateCost, but what do you think?


joseluis2g:
what about just one function with a pvp param so you dont need two functions

EPuncker:
> what about just one function with a pvp param so you dont need two functions

teach me senpai 😄

soul4soul:
I don't see why this necessarily needs to be a single function. How about adding a small description for each function so their intended purpose is more clear. Then the name is not such a big deal.

`getBlessingsCost` - The function `getBlessingsCost` returns the cost of a single blessing. The cost of the blessing is a function of the players level. 

`getPvpBlessingCost` - The function `getPvpBlessingCost` returns the cost of the twist of fate blessing. The twist of fate blessing only applies to PVP related deaths. The cost of the blessing is a function of the players level.

Cipbia made blessing so damn confusing. A blessing protects other blessings which only is in effect when you are killed in PVP...

PR #3037, opened 2020-05-08
Added ice flower harvesting

EPuncker:
it was added on 9.4 update

https://tibia.fandom.com/wiki/Ice_Flower

- added its achievement
- after harvesting it will return to its original state after 4 hours

script based on [this video](https://www.youtube.com/watch?v=j2aPzj4Z8-Y)


BahamutxD:
Works just fine

PR #3039, opened 2020-05-09
Magicwall shouldn't disappear on forced step-in

Sundance:
Test done on a retro world

img
img
img
https://www.tibia.com/community/?subtopic=characters&name=Merlins+Pain

Code changes not tested

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -1445,15 +1445,6 @@ void AreaCombat::setupExtArea(const std::list<uint32_t>& list, uint32_t rows)
 
 void MagicField::onStepInField(Creature* creature)
 {
-	//remove magic walls/wild growth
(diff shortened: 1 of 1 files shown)


EPuncker:
wild growth have same behavior?

Sundance:
Unconfirmed, but those are also a type of magic wall so they most likely have the same behavior

MillhioreBT:
I do not understand this PR very well, but I cannot reproduce the one of its GIF
GIF 10-05-2020 01-20-02 a  m
Neither with wild growth.

Zbizu:
Imo it's okay. If somebody wants it to be kept, this can be done in lua and since it's one of bigger pvp things, it can be put in config.lua

emil92b:
back in the days you would get stuck in the door ;p so make sense they fixed that

EPuncker:
#3836

PR #3046, opened 2020-05-12
Define coins ( items.xml ) <attribute key="worth" value=""/>

MillhioreBT:
With this changes in the sources, we can now add new coins in the game without much problem, this feature was on the old servers, but I don't know why it was removed.

The `Game::addMoney` function is not mine, I tried it and it works very well, who agrees with this change?

I know that tfs is based on emulating at all costs Tibia RL, but this change does not affect that and gives the opportunity to people who use this base for their custom projects.

Example of new coin:
```xml
<item id="9971" article="a" name="gold ingot" plural="gold ingots">
	<attribute key="worth" value="1000000" />
	<attribute key="weight" value="1800" />
</item>
```


nekiro:
I didn’t look at the code yet, but this change is great. Add a lot of flexibility to how we handle money in-game. Awesome, great job.

yrpen:
Maybe we should move it to LUA?

Xaekai:
> tested and working


Sweet summer child, please tell me you aren't always this careless.

MillhioreBT:
With the latest changes it already covers everything that was needed, thanks @Xaekai
Now it is no longer necessary to configure the currencies in the lua files, just adding a new currency in items.xml will be enough, it will have the conversion functionality just like the others and as always.

EPuncker:
we need a change related to this commit now, right?
https://github.com/otland/forgottenserver/commit/3cb540590ee953d1b3f4a83090168ca1d228f897

there are some new functions dealing with hard coded money that will need change like you did with change gold.

and about the function name since nobody suggested anything, maybe lets stay with Game.getCoinsList

Znote:
Iterates through the 3 default currency items and probably needs a patch in this PR to auto-consider all coin items. This function ensures a player not only have enough cap to carry it, but available item slots. 
https://github.com/otland/forgottenserver/blob/3cb540590ee953d1b3f4a83090168ca1d228f897/data/lib/core/player.lua#L138-L173

And this which only cares about cap. 
https://github.com/otland/forgottenserver/blob/3cb540590ee953d1b3f4a83090168ca1d228f897/data/npc/lib/npc.lua#L149-L156

MillhioreBT:
> we need a change related to this commit now, right?
> [3cb5405](https://github.com/otland/forgottenserver/commit/3cb540590ee953d1b3f4a83090168ca1d228f897)
> 
> there are some new functions dealing with hard coded money that will need change like you did with change gold.
> 
> and about the function name since nobody suggested anything, maybe lets stay with Game.getCoinsList

I like `Game.getCoinList`

EPuncker:
this need a rebase, and a nice function name xikini suggested: getCurrencyItems

MillhioreBT:
Closed due to conflicts files and I also want to open a cleaner PR

PR #3049, opened 2020-05-16
Moved experience stages to Lua

Lyuzera:


nekiro:
This was also suggested here #1642 
No idea why it never got merged, probably backwards incompatible? I like that change though.

data/stages.lua
```diff
@@ -0,0 +1,33 @@
+-- set to true to enable experience stages
+stagesEnabled = false
```
nekiro, line 2: I guess this is already defined in config.lua and should still be

data/stages.lua
```diff
@@ -0,0 +1,33 @@
...
+	{minlevel = 21, maxlevel = 50, multiplier = 5},
+	{minlevel = 51, maxlevel = 100, multiplier = 4},
+	{minlevel = 101, multiplier = 3}
+}
+
+stagesMap = {}
+useLastStageLevel = false
+lastStageLevel = 0
+
+for _, stage in ipairs(experienceStages) do
```
nekiro, line 16: I actually dont understand all of this code, why not just search thru table directly?

DSpeichert:
Here we go again.
Not going to be merged if you drop XML support, new functionality may be added but it must fall back to XML.

data/stages.lua
Lyuzera: so that no iteration is necessary every time a player gains experience. So with this cache, we can get the multiplier using the player level as index. this Lua code does exactly the same cache as the old C ++ code does.

```lua
-- stagesMap dump
{
    [1] = 7,
    [2] = 7,
    [3] = 7,
    [4] = 7,
    [5] = 7,
    [6] = 7,
    [7] = 7,
    [8] = 7,
    [9] = 6,
    [10] = 6,
    [11] = 6,
    [12] = 6,
    [13] = 6,
    [14] = 6,
    [15] = 6,
    [16] = 6,
    [17] = 6,
    [18] = 6,
    [19] = 6,
    [20] = 6,
    [21] = 5,
    [22] = 5,
    [23] = 5,
    [24] = 5,
    [25] = 5,
    [26] = 5,
    [27] = 5,
    [28] = 5,
    [29] = 5,
    [30] = 5,
    [31] = 5,
    [32] = 5,
    [33] = 5,
    [34] = 5,
    [35] = 5,
    [36] = 5,
    [37] = 5,
    [38] = 5,
    [39] = 5,
    [40] = 5,
    [41] = 5,
    [42] = 5,
    [43] = 5,
    [44] = 5,
    [45] = 5,
    [46] = 5,
    [47] = 5,
    [48] = 5,
    [49] = 5,
    [50] = 5,
    [51] = 4,
    [52] = 4,
    [53] = 4,
    [54] = 4,
    [55] = 4,
    [56] = 4,
    [57] = 4,
    [58] = 4,
    [59] = 4,
    [60] = 4,
    [61] = 4,
    [62] = 4,
    [63] = 4,
    [64] = 4,
    [65] = 4,
    [66] = 4,
    [67] = 4,
    [68] = 4,
    [69] = 4,
    [70] = 4,
    [71] = 4,
    [72] = 4,
    [73] = 4,
    [74] = 4,
    [75] = 4,
    [76] = 4,
    [77] = 4,
    [78] = 4,
    [79] = 4,
    [80] = 4,
    [81] = 4,
    [82] = 4,
    [83] = 4,
    [84] = 4,
    [85] = 4,
    [86] = 4,
    [87] = 4,
    [88] = 4,
    [89] = 4,
    [90] = 4,
    [91] = 4,
    [92] = 4,
    [93] = 4,
    [94] = 4,
    [95] = 4,
    [96] = 4,
    [97] = 4,
    [98] = 4,
    [99] = 4,
    [100] = 4,
    [101] = 3
}
```

data/stages.lua
Lyuzera: really, i forgot it!

Lyuzera:
> Here we go again.
> Not going to be merged if you drop XML support, new functionality may be added but it must fall back to XML.

Why not discard the xml? I no longer see the need to use it, just as there is no need to maintain stages in C++, because onGainExperience/onLossExperience are in Lua. Maybe you say that it works for websites, in this case, we can easily insert the cache into the database.

data/stages.lua
```diff
@@ -0,0 +1,30 @@
...
+	end
+
+	if not useLastStageLevel then
+		for i = stage.minlevel, stage.maxlevel do
+			stagesMap[i] = stage.multiplier
+		end
+	else
+		stagesMap[lastStageLevel] = multiplier
+	end
+end
```
EPuncker, line 30: add new line at end of the file

data/lib/core/game.lua
```diff
@@ -66,3 +66,15 @@ end
...
+	if not configManager.getBoolean(configKeys.EXPERIENCE_STAGES) then
+		return configManager.getNumber(configKeys.RATE_EXPERIENCE)
+	end
+
+	if useLastStageLevel and level >= lastStageLevel then
+		return stagesMap[lastStageLevel]
+	end
+
+	return stagesMap[level]
+end
```
EPuncker, line 80: add new line at end of the file

data/stages.lua
```diff
@@ -0,0 +1,30 @@
+experienceStages = {
+	{minlevel = 1, maxlevel = 8, multiplier = 7},
+	{minlevel = 9, maxlevel = 20, multiplier = 6},
+	{minlevel = 21, maxlevel = 50, multiplier = 5},
+	{minlevel = 51, maxlevel = 100, multiplier = 4},
+	{minlevel = 101, multiplier = 3}
```
lgrossi, line 6: shouldn't variables name follow camel case? am I missing anything?

data/stages.lua
```diff
@@ -0,0 +1,30 @@
...
+		useLastStageLevel = true
+	end
+
+	if not useLastStageLevel then
+		for i = stage.minlevel, stage.maxlevel do
+			stagesMap[i] = stage.multiplier
+		end
+	else
+		stagesMap[lastStageLevel] = multiplier
+	end
```
lgrossi, line 29: You can also simplify this whole for:

```
local multiplier = stage.multiplier or 1.0
if not stage.maxlevel then
  lastStageLevel = stage.minlevel or 1
  stagesMap[lastStageLevel] = multiplier
else
  for i = stage.minlevel, stage.maxlevel do
    stageMap[i] = multiplier
  end
end
```

data/lib/core/game.lua
```diff
@@ -66,3 +66,15 @@ end
 function Game.setStorageValue(key, value)
 	globalStorageTable[key] = value
 end
+
+function Game.getExperienceStage(level)
+	if not configManager.getBoolean(configKeys.EXPERIENCE_STAGES) then
+		return configManager.getNumber(configKeys.RATE_EXPERIENCE)
+	end
+
+	if useLastStageLevel and level >= lastStageLevel then
```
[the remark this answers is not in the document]
Lyuzera: I liked the second option, very good, thanks: D

data/stages.lua
```diff
@@ -0,0 +1,30 @@
+experienceStages = {
+	{minlevel = 1, maxlevel = 8, multiplier = 7},
+	{minlevel = 9, maxlevel = 20, multiplier = 6},
+	{minlevel = 21, maxlevel = 50, multiplier = 5},
+	{minlevel = 51, maxlevel = 100, multiplier = 4},
+	{minlevel = 101, multiplier = 3}
+}
+
+stagesMap = {}
+useLastStageLevel = false
```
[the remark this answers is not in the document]
Lyuzera: yeah, bye bye `useLastStageLevel`!

data/stages.lua
Lyuzera: do you talk about the `maxlevel` variable  on the last stage that doesn't exist?

data/stages.lua
Lyuzera: very good bro, so simple and compact! Thanks again!

data/stages.lua
lgrossi: Im talking about minlevel, maxlevel in the table, specifically. Not sure what pattern are we following for table 'properties' like table.minlevel. Im used to camelcase here as well (from a javascript world /json patter I'd guess a.minLevel instead). Not sure if the same would apply here. It's more like a general doubt.

Lyuzera: I don't usually use camelcase when they are short variables, but if the pattern is camelcase, I can change it.

lgrossi: I don't really know, it was just a question. Im not the right person to answer that :sweat_smile:

Xaekai: Because the var names are both short and the particular letter breaks involved, `nl` `xl`, you can freely get away with using all lowercase here. I'd argue it's even desirable because of the presence of a long var name like multiplier would make capitals in the other two jarring to the eyes.

config.lua.dist
```diff
@@ -94,7 +94,7 @@ serverSaveClose = false
 serverSaveShutdown = true
 
 -- Rates
--- NOTE: rateExp is not used if you have enabled stages in data/XML/stages.xml
+experienceStages = false
```
Xaekai, line 97: If no one objects, I think `useStages` would look nicer here. (The all caps enum on the cpp side is fine though)

Lyuzera: really better!

Lyuzera:
> Is it really necessary to change the name of this variable? I think the new name does not aim to the stages of experience specifically, there may be a system of stages that needs a name in the future and this will seem very general.

until we add another stages system, we can use `useStages` as it is more simplified.

Underewarrr (approved):
Well looks this code have some knowlege and can work.
Dont forget update Lua config with new bolean.

PR #3050, opened 2020-05-16
Added Rift Runner mount

EPuncker:
- added missing sparkion monster
- moved three mounts from taming script to other one, since they aren't available by targeting a creature but by using an item

PS: need a better name for script file and **TEST NEEDED**


BahamutxD:
Vibrant Egg (blue) is also a usable item. You don't need the monster to tame it.

Other than that the script is working just fine. 

As for the name no suggestions, there's only one more mount that will fit here https://tibia.fandom.com/wiki/Haze but its from version 12.30

EPuncker:
@BahamutxD okay, try now xD

BahamutxD:
The achievements ain't working or I am messing up something.

Using one of the 4 items (doesn't matter which one) will grant the "Natural Born Cowboy" achievement. The other achievement won't trigger.

Also, PlayerStorageKeys.vortexTamer will conflict in storages.lua with the mutatedPumpkin one.

data/scripts/actions/others/usable_mount_items.lua
```diff
@@ -0,0 +1,56 @@
...
+		return false
+	end
+	
+	if table.contains({26194, 26340, 26341}, useItem) then
+		local storage = player:getStorageValue(PlayerStorageKeys.vortexTamer)
+		if storage == 3 then
+			player:addAchievement("Vortex Tamer")
+		else
+			player:setStorageValue(PlayerStorageKeys.vortexTamer, storage + 1)
+		end
```
M4g1ch, line 43: The way you get the achievement is by taming all three monsters from what I understood, now it would be enough to tame one of them three times

data/lib/core/storages.lua
```diff
@@ -11,6 +11,7 @@ PlayerStorageKeys = {
 	afflictedPlagueBell = 30028,
 	nailCaseUseCount = 30031,
 	swampDigging = 30032,
+	vortexTamer = 30033,
```
M4g1ch, line 14: Do you always use the same number for storage values? :hushed:

EPuncker:
@M4g1ch and @BahamutxD everything should be working 100% now, I've tested, can you guys test it once again please

BahamutxD:
Works good now.

https://gfycat.com/madeupdistortedheron

PR #3054, opened 2020-05-17
Feature/refactor viewports

lgrossi:
I did this change on my personal project, might be useful for some of you.
Centralizing the viewport usage and removing the hardcode allows you to perform some more complex client/server view operations, like customize the view size with server/client sync (e.g. zooming in the client requests more tiles to the server and updates the viewport).

While those more complex usages demands otc, this simple refactor does not.

### DIFF
--- a/src/creature.h
+++ b/src/creature.h
@@ -172,6 +172,13 @@ class Creature : virtual public Thing
 			hiddenHealth = b;
 		}
 
+		bool isMoveLocked() const {
+			return moveLocked;
+		}
+		void setMoveLocked(bool locked) {
+			moveLocked = locked;
+		}
+
 		int32_t getThrowRange() const override final {
 			return 1;
 		}
@@ -525,6 +532,7 @@ class Creature : virtual public Thing
 		bool forceUpdateFollowPath = false;
 		bool hiddenHealth = false;
 		bool canUseDefense = true;
+		bool moveLocked = false;
 
 		//creature script events
 		bool hasEventRegistered(CreatureEventType_t event) const {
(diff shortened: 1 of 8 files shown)


andersonfaaria:
@lgrossi you also added the 'moveLocked' feature along with it

Xaekai:
I think you're doing the math on some of this wrong... at least if the existing patches I saw floating around OTLand I based [this](https://gist.github.com/Xaekai/373876d7b4bd1a1c2f48304add23ea57) change I made to my 8.60 fork on are anything to go by.

It would be nice to know a definitive answer on that. Though I suppose it really doesn't matter what the canonical usage of the params is for the real client since it will most likely never support the feature this is supposed to make easier.

CristoferMartins:
Besides the extra feature that has nothing to do with the pr, this pr fix an issue magic numbers. Those numbers are just there with little to none explanation.

EPuncker:
apart from the comments above, wouldn't rangeX and rangeY be better names than rangex and rangey?

lgrossi:
Sorry I did it in a much cleaner way, will post a new PR soon removing the unnecessary feature and adding the better approach. thanks

Znote:
> Sorry I did it in a much cleaner way, will post a new PR soon removing the unnecessary feature and adding the better approach. thanks

This PR interests me, did you create a new one? I imagine this would be super useful for anyone wanting to change the "tile resolution" on their clients.

lgrossi:
Hey, so yes I have it done properly but I have no PR yet. Actually Im aiming to change the whole tile dimension definition to the server that will inform the client to adapt to the server resolution. So far I did it the otherway around, what I decided that is not ideal. That way the server can check if its ot client, if not it uses default resolution, if yes it tell the client the required one.

Issue #2218, opened 2017-03-26
Investigate the use of ASIO timers instead of the Scheduler tasks


djarek:
The current solution of implementing timers (Scheduler tasks) with thread sleeps is likely to be [inefficient](https://gha.st/short-sleeps/) in terms of OS<->Userspace context switches and thread rescheduling for short sleep times (1 ms and less). We should investigate using the ASIO timer library (which uses the ```timerfd``` API) instead.

## Requirements:
- use one of ASIO's timer implementations instead of the Scheduler
- the timers should post a functor to the Dispatcher's execution queue
- should be safe when #2217 is resolved

CristoferMartins:
@djarek My bad, i have said dispatcher but i what mean was the replacement of scheduler thread with asio timers. Also why we need multiple asio threads to start using asio timers?

djarek:
There's no need for a separate thread explicitly - there's need for a separarte `io_context` because the networking code is not thread-safe enough for more than 1 network/scheduling thread to operate in.

CristoferMartins:
@djarek do you got any idea where its not thread safe? the network code(thanks to you too) is quit "simple", the only thing i can think that is issue is how the player object is used inside protocolgame.cpp.

PR #3056, opened 2020-05-18
Implement scheduler timers using asio deadline timers

yamaken93:
As a bonus someone can use scheduler thread to run tasks using the io_service.post function.

### DIFF
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -23,113 +23,86 @@
 
 void Scheduler::threadMain()
 {
-	std::unique_lock<std::mutex> eventLockUnique(eventLock, std::defer_lock);
-	while (getState() != THREAD_STATE_TERMINATED) {
-		std::cv_status ret = std::cv_status::no_timeout;
-
-		eventLockUnique.lock();
-		if (eventList.empty()) {
-			eventSignal.wait(eventLockUnique);
-		} else {
-			ret = eventSignal.wait_until(eventLockUnique, eventList.top()->getCycle());
-		}
-
-		// the mutex is locked again now...
-		if (ret == std::cv_status::timeout && !eventList.empty()) {
-			// ok we had a timeout, so there has to be an event we have to execute...
-			SchedulerTask* task = eventList.top();
-			eventList.pop();
-
-			// check if the event was stopped
-			auto it = eventIds.find(task->getEventId());
-			if (it == eventIds.end()) {
-				eventLockUnique.unlock();
-				delete task;
-				continue;
-			}
-			eventIds.erase(it);
-			eventLockUnique.unlock();
-
-			task->setDontExpire();
-			g_dispatcher.addTask(task, true);
-		} else {
-			eventLockUnique.unlock();
-		}
-	}
+	io_service.run();
 }
 
 uint32_t Scheduler::addEvent(SchedulerTask* task)
 {
-	eventLock.lock();
-
-	if (getState() != THREAD_STATE_RUNNING) {
-		eventLock.unlock();
-		delete task;
-		return 0;
-	}
-
 	// check if the event has a valid id
 	if (task->getEventId() == 0) {
+		uint32_t tmpEventId = ++lastEventId;
 		// if not generate one
-		if (++lastEventId == 0) {
-			lastEventId = 1;
+		if (tmpEventId == 0) {
+			tmpEventId = ++lastEventId;
 		}
-
-		task->setEventId(lastEventId);
+		task->setEventId(tmpEventId);
 	}
 
 	// insert the event id in the list of active events
-	uint32_t eventId = task->getEventId();
-	eventIds.insert(eventId);
+	io_service.post([this, task]() {
+		boost::asio::deadline_timer* timer;
+		if (timerCacheVec.empty()) {
+			timer = new boost::asio::deadline_timer(io_service);
+		} else {
+			timer = timerCacheVec.back();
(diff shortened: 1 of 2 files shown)


MillhioreBT:
@CristoferMartins some people do not understand the reason for some changes, can you explain what benefits does this change bring? Or is it just a more modern way of writing the same thing?

CristoferMartins:
@MillhioreBT just another way to implement timers. But also #2218

MillhioreBT:
> @MillhioreBT just another way to implement timers. But also #2218

I already understand thank you very much, it is important for better precision in all sub-processes, I will definitely test it

CristoferMartins:
As a note, in my tests, there is min 4 max 74 microseconds delay to io_service execute the addevent/stopevent lambda in linux.

src/scheduler.h
```diff
@@ -38,44 +38,36 @@ class SchedulerTask : public Task
...
 		std::thread thread;
-		std::mutex eventLock;
-		std::condition_variable eventSignal;
-
-		uint32_t lastEventId {0};
-		std::priority_queue<SchedulerTask*, std::deque<SchedulerTask*>, TaskComparator> eventList;
-		std::unordered_set<uint32_t> eventIds;
+		std::atomic<uint32_t> lastEventId {0};
+		std::unordered_map<uint32_t, boost::asio::deadline_timer*> eventIdTimerMap;
+		std::vector<boost::asio::deadline_timer*> timerCacheVec;
```
ranisalt, line 68: Does it require pointers? I would store instances otherwise. Don't like the sight of two levels of indirections.

src/scheduler.cpp
```diff
@@ -23,113 +23,86 @@
...
-		return 0;
-	}
-
 	// check if the event has a valid id
 	if (task->getEventId() == 0) {
+		uint32_t tmpEventId = ++lastEventId;
 		// if not generate one
-		if (++lastEventId == 0) {
-			lastEventId = 1;
+		if (tmpEventId == 0) {
```
ranisalt, line 35: How would it ever be 0 if `lastEventId` is initialized to 0 and incremented before testing?

src/scheduler.cpp
```diff
@@ -23,113 +23,86 @@
...
-
-	eventIds.erase(it);
-	return true;
+	io_service.post([this, eventId]() {
+		// search the event id..
+		auto it = eventIdTimerMap.find(eventId);
+		if (it == eventIdTimerMap.end()) {
+			return;
+		}
+		it->second->cancel();
```
ranisalt, line 82: You can invert the condition (`it != end`) to save a line here.

src/scheduler.cpp
```diff
@@ -23,113 +23,86 @@
...
-		delete eventList.top();
-		eventList.pop();
-	}
+		// thanks to cancel we will have all the timers in timerCacheVec
+		// deleted them to prevent memory leak
+		// this must be a new task since timer->cancel() adds a new task for each cancel so we want to delete the timers only after all cancel callbacks were called
+		io_service.post([this]() {
+			for (auto* timer : timerCacheVec) {
+				delete timer;
+			}
```
ranisalt, line 101: If you end up not using pointers, you can probably get off by replacing the loop with `timerCacheVec = {}`

nekiro:
@CristoferMartins any update on this? This is one of the latest prs marked for 1.4

src/scheduler.h
CristoferMartins: I believe if i don't use pointers i can't reuse the deadline_timer object. Got ideas? "Don't like the sight of two levels of indirections" elaborate please.

CristoferMartins (commented):
@nekiro i don't even know of tfs oficial devs are interested in this pr to be merged in the project.

src/scheduler.h
ranisalt: Vector is allocated on the heap, and if you use pointers for `deadline_timers` they will also be allocated in the heap, possibly incurring in performance issues if these are accessed  often and out of order. If you use the objects instead of pointers, they are stored contiguouly, so you avoid one round trip to the slow RAM.

Why/when would you reuse the object instead?

src/scheduler.cpp
```diff
@@ -23,113 +23,86 @@
...
-	eventList.push(task);
+	io_service.post([this, task]() {
+		boost::asio::deadline_timer* timer;
+		if (timerCacheVec.empty()) {
+			timer = new boost::asio::deadline_timer(io_service);
+		} else {
+			timer = timerCacheVec.back();
+			timerCacheVec.pop_back();
+		}
+		eventIdTimerMap[task->getEventId()] = timer;
```
ranisalt, line 50: I get when you are reusing the same timer objects here. Is the construction of a timer that heavy that you can't create a new one every time, and delete it when done?

ranisalt:
@CristoferMartins I like the cleanup, consistency of

ranisalt:
Sorry, misclick

Znote:
> @nekiro i don't even know of tfs oficial devs are interested in this pr to be merged in the project.

We do, this is one out of 4 remaining PR tagged for TFS 1.4. We are waiting for you to address and resolve the review feedback. 
And do you have any suggestions on how to test this PR?

nekiro:
Just wanted to add that boost::asio::io_service is deprecated, more info https://stackoverflow.com/questions/59753391/boost-asio-io-service-vs-io-context

PR #3057, opened 2020-05-18
Cleaner and faster XTEA implementation

ranisalt:
Out with all that optimal block size width, let the compiler figure it out. Reduces the amount of code by half and increases performance by ~20%.

Shameless plug: [I wrote about the optimization of this algorithm](https://dev.to/rsa/making-an-algorithm-15x-faster-without-parallelism-38k5), please read it :stuck_out_tongue:

### DIFF
--- a/src/xtea.cpp
+++ b/src/xtea.cpp
@@ -1,6 +1,6 @@
 /**
  * The Forgotten Server - a free and open-source MMORPG server emulator
- * Copyright (C) 2019  Mark Samman <mark.samman@gmail.com>
+ * Copyright (C) 2020  Mark Samman <mark.samman@gmail.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


jo3bingham:
Nice. This is the kind of code/blog I like to see/read. FWIW, the code LGTM. :shipit:

src/xtea.cpp
```diff
@@ -30,111 +30,46 @@ namespace {
...
-constexpr auto encrypt_v = XTEA<true, InitialBlockSize>();
-constexpr auto decrypt_v = XTEA<false, InitialBlockSize>();
-
-} // anonymous namespace
-
-void encrypt(uint8_t* data, size_t length, const key& k) { encrypt_v(data, length, k); }
-void decrypt(uint8_t* data, size_t length, const key& k) { decrypt_v(data, length, k); }
+void decrypt(uint8_t* data, size_t length, const key& k)
+{
+    for (auto i = 0u, sum = delta * 32, next_sum = sum - delta; i < 32u; ++i, sum = next_sum, next_sum -= delta) {
```
MillhioreBT, line 67: Why not use `uint64_t` instead of `auto` ? this removes the `C26450` warning.

Xaekai:
I quite enjoyed reading the write up. Thank you.

src/xtea.cpp
CristoferMartins: Compiling in 32 bits there is this warning in both visual studio 2015 and 2019: "warning C4307: '*': integral constant overflow". long long fix the warning.

ranisalt: @MillhioreBT it should actually infer `uint32_t`. Using `auto` is personal taste, I will change it due to the warning.

ranisalt: @CristoferMartins how do I tell the VS compiler the overflow it is intended?

ranisalt: Did it fix for you both?

MillhioreBT: > @MillhioreBT it should actually infer `uint32_t`. Using `auto` is personal taste, I will change it due to the warning.

@ranisalt yeah it is correct the use of `uint32_t`

Issue #2879, opened 2020-03-30
[Suggestion] New house settings


EPuncker:
It seems that currently we have no limitations to buying houses other than one per character, and since recently tibia added the possibility to own 3 houses on the same character (or account/world), here is my list of suggestions regarding our current house configs:

- [ ] House per account limit instead of per character
- [x] Add option to show house price on look at door = true/false
- [x] Level requirement
- [x] Premium requirement
- [x] a setting to make all characters on account invited on the house by default = true/false

MillhioreBT:
i like your idea, i safe will work with this

MillhioreBT:
> and a bonus if you are really feeling good today: a setting to make all characters on account invited on the house by default = true/false

check!
https://github.com/otland/forgottenserver/pull/2935

EPuncker:
closing as the last feature is more likely to be added AAC side than server's

PR #3060, opened 2020-05-21
Additional Configs

EPuncker:
- Added !buyhouse level and premium requirement config (#2879 )
- Added setting to enable/disable change sex and buy premium command
- Moved a few talkaction settings config.lua file


MillhioreBT:
I don't like the idea of ​​the `script config.lua` file but I do like the reason.

DSpeichert:
The second config is quite confusing.
Can't we just add additional options to the "main" config.lua that the C++ code will simply ignore?

soul4soul:
I'm don't like the idea of combining all of the scripts with a common config file. I do like some of the enhancements made to the individual scripts and would like to see the specific configs stay in those files. EG. keep the new config added for dice on depot in the dice.lua file.

EPuncker:
> The second config is quite confusing.
> Can't we just add additional options to the "main" config.lua that the C++ code will simply ignore?

I didn't knew it was possible 😄 gonna do it

DSpeichert:
So config.lua is implicitly available throughout Lua scope? I don't see the libs/ executing it anywhere.

EPuncker:
> So config.lua is implicitly available throughout Lua scope? I don't see the libs/ executing it anywhere.

yeah its kinda weird, there are some that works and some that don't, will investigate further

config.lua.dist
```diff
@@ -92,6 +100,12 @@ yellMinimumLevel = 2
 yellAlwaysAllowPremium = false
 forceMonsterTypesOnLoad = true
 cleanProtectionZones = false
+changeSexEnabled = true
+changeSexDaysCost = 3
+premiumToOfflineTraining = true
+deathListEnabled = true
+maxDeathRecords = 5
+useWormsToFishing = true
```
DSpeichert, line 108:
```suggestion
useWormsForFishing = true
```

DSpeichert (changes requested):

nekiro:
Hoenstly I find these redundant and config bloating, why creating config values for easily editable lua scripts?

EPuncker:
> Honestly I find these redundant and config bloating, why creating config values for easily editable lua scripts?
> Some of them like death related are okay, but "useWormsForFishing" is kind of redundant to me.

I find it better to have all of them in a single place, but yes that is kinda of a matter of opinion

peonso:
Bloating. Config should only deal with server setup (map name/ip/database) and hardcoded stuff. If you add everysingle script option to config.lua because someone want it there we can reach thousand lines really fast.

DSpeichert:
I have reconsidered my earlier opinion and decided that config.lua is not the place to gather all config options from the datapack.
IMHO it would be clearer to the user when config.lua focuses of the built-in engine features, which generally means fewer and fewer things as things move to Lua.

What we need instead is a good datapack README, showing which scripts are responsible for which behavior, so that a user can go and reconfigure them.

I'm primarily thinking about TFS as an engine that can be used with the widest range of datapacks, therefore comingling the core config and datapack config is confusing.

Issue #2561, opened 2019-04-09
Crash on eventPlayerOnLook


infister:
### Before creating an issue, please ensure:
- [ ] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [X] This issue is reproducible without changes to the code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. no idea, happens rarely and i cannot reproduce (dont know a way)

### Actual behaviour
crash

### Environment

```
#0  0x00007f4b37b1918c in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f4b383c652b in std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007f4b383b758e in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00005627f5f7e105 in Player::getDescription[abi:cxx11](int) const ()
#4  0x00005627f5eeff77 in LuaScriptInterface::luaCreatureGetDescription(lua_State*) ()
#5  0x00007f4b39660c75 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#6  0x00007f4b3966c825 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#7  0x00007f4b39660f9e in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#8  0x00007f4b396605cf in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#9  0x00007f4b39661201 in ?? () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#10 0x00007f4b3965d0a1 in lua_pcallk () from /usr/lib/x86_64-linux-gnu/liblua5.2.so.0
#11 0x00005627f5e7ba24 in LuaScriptInterface::protectedCall(lua_State*, int, int) ()
#12 0x00005627f5e7ca0d in LuaScriptInterface::callVoidFunction(int) ()
#13 0x00005627f5da7ecd in Events::eventPlayerOnLook(Player*, Position const&, Thing*, unsigned char, int) ()
[...]

PR #3062, opened 2020-05-23
Use shared_ptr for GuildRank

ninjalulz:
This should resolve the crash bug that occurs in `Player::getDescription`.

Related issue: #2561

### DIFF
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -498,7 +498,7 @@ bool Chat::talkToChannel(const Player& player, SpeakClasses type, const std::str
 	}
 
 	if (channelId == CHANNEL_GUILD) {
-		const GuildRank* rank = player.getGuildRank();
+		GuildRank_ptr rank = player.getGuildRank();
 		if (rank && rank->level > 1) {
 			type = TALKTYPE_CHANNEL_O;
 		} else if (type != TALKTYPE_CHANNEL_Y) {
--- a/src/guild.cpp
+++ b/src/guild.cpp
@@ -47,7 +47,7 @@ void Guild::removeMember(Player* player)
 	}
 }
 
-GuildRank* Guild::getRankById(uint32_t rankId)
+GuildRank_ptr Guild::getRankById(uint32_t rankId)
 {
 	for (auto& rank : ranks) {
 		if (rank.id == rankId) {
@@ -57,7 +57,7 @@ GuildRank* Guild::getRankById(uint32_t rankId)
 	return nullptr;
 }
 
-const GuildRank* Guild::getRankByName(const std::string& name) const
+GuildRank_ptr Guild::getRankByName(const std::string& name) const
 {
 	for (const auto& rank : ranks) {
 		if (rank.name == name) {
@@ -67,7 +67,7 @@ const GuildRank* Guild::getRankByName(const std::string& name) const
 	return nullptr;
 }
 
-const GuildRank* Guild::getRankByLevel(uint8_t level) const
+GuildRank_ptr Guild::getRankByLevel(uint8_t level) const
 {
 	for (const auto& rank : ranks) {
 		if (rank.level == level) {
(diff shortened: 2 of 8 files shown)


lollers420:
Bad fix. breaks changing rank names completley they will not be updated till after a server restart!

Lyuzera:
> Bad fix. breaks changing rank names completley they will not be updated till after a server restart!

can you explained it? how to reproduce it? I use that fix for a long time and it solves my problem.

lollers420:
> > Bad fix. breaks changing rank names completley they will not be updated till after a server restart!
> 
> can you explained it? how to reproduce it? I use that fix for a long time and it solves my problem.

Explain what??? with this change rank names will no longer be updated in the game. So if a player changes his guild rank with name "a member" to "a gamer" it will stay "a member" till after a server restart. This was not the case before, then it was properly updated to "a gamer" wihtout any restarts and it should be fixed!

Lyuzera:
> > > Bad fix. breaks changing rank names completley they will not be updated till after a server restart!
> > 
> > 
> > can you explained it? how to reproduce it? I use that fix for a long time and it solves my problem.
> 
> Explain what??? with this change rank names will no longer be updated in the game. So if a player changes his guild rank with name "a member" to "a gamer" it will stay "a member" till after a server restart. This was not the case before, then it was properly updated to "a gamer" wihtout any restarts and it should be fixed!

if it really happens, doesn't can merged. So we have to found where are the memory leak, because probably is what crashes the server.

lollers420:
> > > > Bad fix. breaks changing rank names completley they will not be updated till after a server restart!
> > > 
> > > 
> > > can you explained it? how to reproduce it? I use that fix for a long time and it solves my problem.
> > 
> > 
> > Explain what??? with this change rank names will no longer be updated in the game. So if a player changes his guild rank with name "a member" to "a gamer" it will stay "a member" till after a server restart. This was not the case before, then it was properly updated to "a gamer" wihtout any restarts and it should be fixed!
> 
> if it really happens, doesn't can merged. So we have to found where are the memory leak, because probably is what crashes the server.

Yes ofcourse its beter than crash, but Still it removes 1 error and adds another one.

ninjalulz:
> Bad fix. breaks changing rank names completley they will not be updated till after a server restart!

I don't see how it's connected since you can neither modify or delete a rank in-game.

lollers420:
> Even if you alter the `name` in your database, it won't update either way because it checks whether a rank exists or not by looking up its ID.

But AACs change the name? Before in gesior when u changed rank name, it would also be update ingame (if a player relogged) now only after a server reboot. I think its same on znote right?? Same would be if u do it ingame thorugh some guild managment

To me it feels very wrong that we need a server reboot to change a player ranks now. And especially since this wasnt the case before this Fix.

Xaekai:
I think that's the problem.

The game itself should be the only thing modifying the rank names, and should propagate those changes both to logged in players and the database.

If the webportal wants to surface controls for that, it should merely be a relay to an API surfaced by gameworld engine. Top down, full stop. Countless bugs are a result of not enforcing such an architecture.

infister:
> I think that's the problem.
> 
> The game itself should be the only thing modifying the rank names, and should propagate those changes both to logged in players and the database.
> 
> If the webportal wants to surface controls for that, it should merely be a relay to an API surfaced by gameworld engine. Top down, full stop. Countless bugs are a result of not enforcing such an architecture.

There is missing proper API in tfs for that architecture, for now.

LukSrT:
> > I think that's the problem.
> > The game itself should be the only thing modifying the rank names, and should propagate those changes both to logged in players and the database.
> > If the webportal wants to surface controls for that, it should merely be a relay to an API surfaced by gameworld engine. Top down, full stop. Countless bugs are a result of not enforcing such an architecture.
> 
> There is missing proper API in tfs for that architecture, for now.

I believe with this PR, it would already solve the problem of data manipulation on the site, no?

https://github.com/otland/forgottenserver/pull/2010/

Xaekai:
> I believe with this PR, it would already solve the problem of data manipulation on the site, no?

I'm completely aware of that PR. That PR is the only reason I haven't already made a PR doing the same thing, though I'd use libuv instead of Beast. I don't like Beast at all.

PR #3064, opened 2020-05-23
Fix container queryAdd issues

### LINKED ISSUE #1561: Container::queryMaxCount() failing in special cases.

### LINKED ISSUE #3063: Move item to container bug

### Steps to reproduce
1. use a normal character with regular free cap (like 200)
2. make 100 demonic candy ball (id: 12543) and put them on the ground 
3. turn off classic control (in case you use it)
4. hold shift to move one candy ball to your backpack

### Expected behaviour
it should move it to your backpack since it weights only 9.50 oz

### Actual behaviour
it doesn't success move it because it acts like you were trying to move the whole stack that weights 950 oz (using shift and moving to other sqm in the ground works tho)

nekiro:
This reverts #2953 which fixed #1561, but introduced #3063. 
This pull request fixes #3063 and #1561 at once.

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -352,18 +352,16 @@ ReturnValue Container::queryMaxCount(int32_t index, const Thing& thing, uint32_t
 			uint32_t slotIndex = 0;
 			for (Item* containerItem : itemlist) {
 				if (containerItem != item && containerItem->equals(item) && containerItem->getItemCount() < 100) {
-					uint32_t remainder = (100 - containerItem->getItemCount());
-					if (queryAdd(slotIndex++, *item, remainder, flags) == RETURNVALUE_NOERROR) {
-						n += remainder;
+					if (queryAdd(slotIndex++, *item, count, flags) == RETURNVALUE_NOERROR) {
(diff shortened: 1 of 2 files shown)


nekiro:
We need to test every scenario here, like top parent of container could  also be tile and depot, inbox etc.

PR #3066, opened 2020-05-25
Added alias keyword function

EPuncker:
Should be used if you have to answer the same thing to several keywords.

- added sample on runes script, tested and working:

code from ORTS

### DIFF
--- a/data/npc/lib/npcsystem/keywordhandler.lua
+++ b/data/npc/lib/npcsystem/keywordhandler.lua
@@ -64,6 +64,21 @@ if not KeywordHandler then
 		return self:addChildKeywordNode(new)
 	end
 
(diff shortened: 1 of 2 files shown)


data/npc/lib/npcsystem/keywordhandler.lua
```diff
@@ -64,6 +64,21 @@ if not KeywordHandler then
 		return self:addChildKeywordNode(new)
 	end
 
+	-- Adds an alias keyword. Should be used if you have to answer the same thing to several keywords.
+	function KeywordNode:addAliasKeyword(keywords)
+		if #self.children == 0 then
+			print('KeywordNode:addAliasKeyword no previous node found')
```
DSpeichert, line 70: Isn't this going to spam in console?

PR #3067, opened 2020-05-27
Fixed some inconsistency on OnDeath parameters

EPuncker:
also updated the changes to [wiki](https://github.com/otland/forgottenserver/wiki/Revscriptsys#creatureevent)

### DIFF
--- a/data/creaturescripts/scripts/droploot.lua
+++ b/data/creaturescripts/scripts/droploot.lua
@@ -1,5 +1,4 @@
-function onDeath(creature, corpse, killer, mostDamageKiller, lastHitUnjustified, mostDamageUnjustified)
-	local player = creature:getPlayer()
+function onDeath(player, corpse, killer, mostDamageKiller, lastHitUnjustified, mostDamageUnjustified)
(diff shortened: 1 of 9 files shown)


data/creaturescripts/scripts/playerdeath.lua
```diff
@@ -1,8 +1,9 @@
 local deathListEnabled = true
 local maxDeathRecords = 5
 
-function onDeath(player, corpse, killer, mostDamageKiller, unjustified, mostDamageUnjustified)
```
nekiro, line 4: if its registered only for player then this parameter will be always player, so why changing it?

data/creaturescripts/scripts/droploot.lua
```diff
@@ -1,4 +1,5 @@
-function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unjustified)
+function onDeath(creature, corpse, killer, mostDamageKiller, lastHitUnjustified, mostDamageUnjustified)
+	local player = creature:getPlayer()
```
nekiro, line 2: Same as my other comment

PR #3071, opened 2020-05-30
Lua-Generated Item descriptions

infernumx:
* Includes config key `luaItemDesc` to swap between C++/Lua versions of each (can be swapped with /reload config -> global), `false` uses the C++ version, `true` uses the Lua version.

By default every single description should be the exact same (tested with default items.xml and compared both descriptions with each other with [this script](https://pastebin.com/HT6vFXcr)). I encourage those with custom servers with custom items to test this script out as well and check the desc_output.txt written to your server's root directory to ensure this works properly before any merges occur.


data/lib/compat/compat.lua
```diff
@@ -1312,3 +1312,65 @@ end
...
+		[COMBAT_DROWNDAMAGE] = 'drown',
+		[COMBAT_ICEDAMAGE] = 'ice',
+		[COMBAT_HOLYDAMAGE] = 'holy',
+		[COMBAT_DEATHDAMAGE] = 'death'
+	}
+
+	function getCombatName(combat)
+		return combats[combat]
+	end
+end
```
ranisalt, line 1335: Why wrapping with `do...end`?

CristoferMartins:
Just a random thought: Some stuff could be cached right? Maybe even whole items descriptions could be cached(which would be very nice performance wise).

data/lib/compat/compat.lua
infernumx: Just preference to enclose it with the function that's going to use the table. There's no need for that table to be used anywhere else.

infernumx:
> Just a random thought: Some stuff could be cached right? Maybe even whole items descriptions could be cached(which would be very nice performance wise).

I thought about this as well, but I'm not sure how to go about it. I think the only actual caching that's possible are items that don't have any attributes rewritten, just the raw spawned/mapped in item with no changes. Once a change gets made to one of the attributes that should be displayed, the cached description is no longer valid and the item becomes 'un-cachable' because we have to check for the custom attributes each time or the cached version may be wrong if the value of a certain attribute is changed multiple times.
Correct me if I'm wrong, but I don't see a way around that. I'm open to ideas & improvements, maybe I can write some code to cache raw items and see how it goes.

Lyuzera:
what do you think?
```lua
StringStream.str = StringStream.concat
```

```lua
local ss = StringStream()
ss:append('Just a test message')

print(ss:str())
```

Erza (dismissed):

Sickness-chan:
What is the status on this? @DSpeichert @ranisalt

DSpeichert:
1. PR needs rebase.
2. `vc14/.vs/theforgottenserver/v16/Browse.VC.db*` shouldn't be added.

Sickness-chan:
> 2. `vc14/.vs/theforgottenserver/v16/Browse.VC.db*` shouldn't be added.

They were removed again in this commit https://github.com/otland/forgottenserver/pull/3071/commits/a0bb87344b82f1cfd4a87eecf55bd9140cd25707

DSpeichert (approved):

EPuncker (approved):
been using it on my test server for quite some time already, and so far so good

PR #3072, opened 2020-05-31
Print duplicate storages

omarcopires:
I believe that this will not be added to the repository, but it can be useful for customs servers.
Function taken from: https://github.com/opentibiabr/otservbr-global/pull/1158/commits

- Credits to the author: Anderson Faria <andercfaria@hotmail.com>

### DIFF
--- a/data/lib/core/storages.lua
+++ b/data/lib/core/storages.lua
@@ -20,12 +20,8 @@ GlobalStorageKeys = {
 
 -- Checking for duplicate storages:
 local function extractValues(tab, ret)
-	if type(tab) == "number" then
-		table.insert(ret, tab)
-	else
-		for _, v in pairs(tab) do
-			extractValues(v, ret)
-		end
+	for _, v in pairs(tab) do
+ 		table.insert(ret, tab)
 	end
 end
 
--- a/data/lib/core/storages.lua
+++ b/data/lib/core/storages.lua
@@ -17,3 +17,31 @@ PlayerStorageKeys = {
 
 GlobalStorageKeys = {
 }
+
+-- Values extraction function
+local function extractValues(tab, ret)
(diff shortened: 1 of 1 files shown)


EPuncker:
this is kinda interesting, but the benchmark part is unnecessary (this code takes almost zero resources to run), and this should include the GlobalStorageKeys table as well

omarcopires:
> this is kinda interesting, but the benchmark part is unnecessary (this code takes almost no resources to run), and this should include the GlobalStorageKeys table as well

There is no global storage to check, this can easily be introduced later.

data/lib/core/storages.lua
```diff
@@ -17,3 +17,26 @@ PlayerStorageKeys = {
...
+-- Checking for duplicate storages:
+local function extractValues(tab, ret)
+	if type(tab) == "number" then
+		table.insert(ret, tab)
+	else
+		for _, v in pairs(tab) do
+			extractValues(v, ret)
+		end
+	end
+end
```
DSpeichert, line 30: This function is unnecessarily complicated, it gives false impression of being recursive.
There's no reason it has to call itself. It's only used by passing two tables, so for readability, this would be much better:
```suggestion
local function extractValues(tab, ret)
	for _, v in pairs(tab) do
 		table.insert(ret, tab)
	end
end
```
In fact, this function does really need to be a function at all if there is no intention to reuse it.
Don't overcomplicate the code.

lgrossi: Thats what happened when someone copy a code that does something in one base and move to another where it doesn't make sense without realizing what exactly was he doing. @DSpeitchert the function makes sense when you have nested storage tables, then the recursion would be needed, what is not tfs base case. For simple tables your code is more than enough.

PR #3079, opened 2020-06-04
Correction in crate usable

omarcopires:
According to the Brazilian wiki when used by a player, the box is unusable for 1 real hour. This means that no player will receive any items from the box during this period. Likewise, the player who used the box will not be able to use any other box for 1 real hour.

https://www.tibiawiki.com.br/wiki/Crate_(Usable)

### DIFF
--- a/data/lib/core/storages.lua
+++ b/data/lib/core/storages.lua
@@ -5,7 +5,6 @@ PlayerStorageKeys = {
 	firstRod = 30020,
 	delayWallMirror = 30021,
 	madSheepSummon = 30023,
-	crateUsable = 30024,
 	afflictedOutfit = 30026,
 	afflictedPlagueMask = 30027,
 	afflictedPlagueBell = 30028,
@@ -16,4 +15,5 @@ PlayerStorageKeys = {
 }
 
 GlobalStorageKeys = {
+	crateUsable = 30000,
 }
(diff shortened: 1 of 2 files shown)


luanluciano93:
Friend I believe you got it wrong, the script is correct, the **global storage** issue would have to be implemented by a particular **actionId** for each crate implemented on the map.

omarcopires:
> Friend I believe you got it wrong, the script is correct, the **global storage** issue would have to be implemented by a particular **actionId** for each crate implemented on the map.
> 
> something like this, for example: https://hastebin.com/xebugigomi.rb

But my changes don't affect that, it just corrects the break time and leaves it as global storage, as it should be.

if the user wants to use the boxes in different locations this can also be easily implemented.

luanluciano93:
But that way with **global Storage** is wrong, because this should not be taken into account in all crates. The only correct thing is time.

EPuncker:
i'll test on real tibia right now

EPuncker:
same server, two rusty legs, br wiki is wrong 😛

EPuncker:
but yes, the interval should be one hour and not 20

omarcopires:
> But this way with **Global Storage** is wrong, because it should not be taken into account in all crates. The only correct thing is time.

I'm trying to understand your logic, the original script already treated it as a single box, the only difference here is that the range will be applied to all players instead of one!

luanluciano93:
The only change will be in the time that the player can perform the action again.

DSpeichert:
@EPuncker Can you comment on whether any part of this PR is correct?

EPuncker:
> @EPuncker Can you comment on whether any part of this PR is correct?

the only correct part is the use delay from 20 hours to 1 hour 😄

PR #3082, opened 2020-06-07
Action grouping

rookgaard:
Hello.

I would like to introduce another thing what is missing in comparision to 0.X and in my opinion it was very useful - actions (as well movements) grouping. The simpliest example - you have in actions.xml 2 entries for rope (ID 2120 and 7731) which you can merge into one separating IDs by defined char (like "," as most languages in array constructing). Solution is easy and is based on movements' position split - vectorAtoi. If this PR would get accepted during review process I will expand it to movements also.

Of course it can be made using revscripts, but as long we won't throw away XML from actions completely, that solution could be very usefull.


MillhioreBT:
It is not a bad idea! Good!

nekiro:
It won't hurt to have this, but I would change "," to ";" as it already is in other part of sources for xml format for consistency.

rookgaard:
> It won't hurt to have this, but I would change "," to ";" as it already is in other part of sources for xml format for consistency.

@nekiro May I ask for an example? In my opinion it should remain as `,` as that's a way to separate the array elements.

nekiro:
@rookgaard the thing is xml is not scripting language so using array delimeter missed the point. I would just do it as 2222;4444, because this looks more accruate for xml file, if you wanted to see where we use it like that you can change npc xml files.

rookgaard:
@nekiro I've changed `,` to `;` as you requested. Also I've extended feature also for movements (with one example) as it uses almost the same solution. Can you check on it now?

src/actions.cpp
```diff
@@ -87,13 +87,18 @@ bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node)
...
 	pugi::xml_attribute attr;
 	if ((attr = node.attribute("itemid"))) {
-		uint16_t id = pugi::cast<uint16_t>(attr.value());
+		std::vector<int32_t> idList = vectorAtoi(explodeString(attr.as_string(), ";"));
+		bool success = true;
 
-		auto result = useItemMap.emplace(id, std::move(*action));
-		if (!result.second) {
-			std::cout << "[Warning - Actions::registerEvent] Duplicate registered item with id: " << id << std::endl;
+		for (size_t i = 0; i < idList.size(); i++) {
```
nekiro, line 93: why not use range based iteration?
```for (const auto& id : idList)```

**Comment by DSpeichert** [comment]
I'm against merging this change.
It is unnecessary and causes certain ambiguities. What should happen when certain IDs are invalid? What happens when you put in malformed "array" (e.g. `;;;;`).
Readability suffers because IDs cannot really be in order anymore. Static analysis tools (if any) will also have a more complicated job to account for this functionality.

This is not a feature that provides anything you can't already do.

**Comment by rookgaard** [comment]
> I'm against merging this change.
> It is unnecessary and causes certain ambiguities. What should happen when certain IDs are invalid? What happens when you put in malformed "array" (e.g. `;;;;`).
> Readability suffers because IDs cannot really be in order anymore. Static analysis tools (if any) will also have a more complicated job to account for this functionality.
> 
> This is not a feature that provides anything you can't already do.

Well @DSpeichert:
- if someone will put malformed string like `;;;`, it will happen the same as he would put `<action itemid="" script="tools/rope.lua" />`
- if somebody use analysis tool he won't group ID using this solution
- if someone care about ID order, he won't group them using this solution (beside there are many exampled in current code when ID are not in right order - https://github.com/otland/forgottenserver/blob/master/data/weapons/weapons.xml#L45)

This feature is made to make something simplier - it won't have to do things you can't already do. The same goes with "fromid" and "toid" - someone could also say in the past `This is not a feature that provides anything you can't already do.` when looking at `<action fromid="1873" toid="1876" script="other/decayto.lua" />` instead of
```
<action itemid="1873" script="other/decayto.lua" />
<action itemid="1874" script="other/decayto.lua" />
<action itemid="1875" script="other/decayto.lua" />
<action itemid="1876" script="other/decayto.lua" />
```

**Comment by rookgaard** [comment]
Is there anything I can do more on this PR @DSpeichert @nekiro ?

File: data/actions/actions.xml [hunk 2]
```diff
@@ -26,10 +26,9 @@
 	<action itemid="2580" script="tools/fishing.lua" allowfaruse="1" />
 	<action itemid="2554" script="tools/shovel.lua" />
 	<action itemid="5710" script="tools/shovel.lua" />
-	<action itemid="7731" script="tools/rope.lua" />
-	<action itemid="10511" script="tools/tool_gear.lua" />
-	<action itemid="10513" script="tools/tool_gear.lua" />
-	<action itemid="10515" script="tools/tool_gear.lua" />
+	<action itemid="10511" script="tools/toolgear.lua" />
+	<action itemid="10513" script="tools/toolgear.lua" />
+	<action itemid="10515" script="tools/toolgear.lua" />
```
DSpeichert, line 31: which one is correct, `tool_gear.lua` or `toolgear.lua`?

data/actions/actions.xml
```diff
@@ -16,7 +16,7 @@
 	<action uniqueid="30015" script="quests/annihilator.lua" />
 
 	<!-- Tools -->
-	<action itemid="2120" script="tools/rope.lua" />
+	<action itemid="2120;7731" script="tools/rope.lua" />
```
DSpeichert, line 19: Let's not modify the default datapack to use this.

DSpeichert (commented):
I'm inclined to accept this change in efforts to preserve backward compatibility but I ask that you do not modify the actions.xml to use it.
It's not in XML spirit IMO to encode arrays like that and I don't want to make it look like this is the new default. It's a valuable feature to make some old datapacks work better but it's not the way I would design it today.

data/actions/actions.xml
rookgaard: I've reverted this change.

rookgaard: Right, `tool_gear.lua`

data/actions/actions.xml
rookgaard: I've reverted this change.

data/movements/movements.xml
```diff
@@ -29,8 +29,7 @@
 	<movevent event="StepIn" itemid="1387" script="citizen.lua" />
 
 	<!-- Walkback when walking on quest chest -->
-	<movevent event="StepIn" itemid="1738" script="walkback.lua" />
-	<movevent event="StepIn" itemid="1740" script="walkback.lua" />
+	<movevent event="StepIn" itemid="1738;1740" script="walkback.lua" />
```
DSpeichert, line 32: Missed it before - that's also the kind of datapack change I wanted to avoid. I assume you've tested using that one.

rookgaard: Of course :) It was just proof of concept.

DSpeichert (approved):

MillhioreBT:
It looks like this is going to be merged, There is little left for TSF 1.4 you can start to review the PR's and verify if they are mergeables.

@nekiro 
@DSpeichert

peonso:
Looks good

rookgaard:
@DSpeichert  It means we are going to release 1.4 soon?

PR #3085, opened 2020-06-10
Added several functions mainly related to banker NPC

EPuncker:
These functions were added in order for the upcoming sample banker NPC and in the future for the guild balance feature, I'd like to ask for a meticulous evaluation to avoid any error or exploit of the bank system. Please review function usefulness, names and possible enhancements

- function getPlayerDatabaseInfo(name_or_guid)
TFS 1.3 function that queries for player information from the database, returns a table of useful information if success, false if player doesn't exist.

- function isNumber(str)
used in several distros but wasn't there on compat.lua yet
- function Player.transferMoneyTo(self, target, amount)
- function Player.withdrawMoney(self, amount)
- function Player.depositMoney(self, amount)
- function isValidMoney(money)
- function getMoneyCount(string)
- function getMoneyWeight(money)

### DIFF
--- a/data/global.lua
+++ b/data/global.lua
@@ -69,27 +69,98 @@ if not nextUseStaminaTime then
 	nextUseStaminaTime = {}
 end
 
-function capitalize(str)
-	return str:gsub("%a+", function(c) return c:sub(1, 1):upper() .. c:sub(2) end)
-end
+function getPlayerDatabaseInfo(name_or_guid)
+	local sql_where = ""
 
--- check if offline player exist
-function playerExists(name)
-	local query = db.storeQuery("SELECT `name` FROM `players` WHERE `name` = " .. db.escapeString(name))
-	if query then
-		result.free(query)
-		return true
+	if type(name_or_guid) == 'string' then
+		sql_where = "WHERE `p`.`name`=" .. db.escapeString(name_or_guid) .. ""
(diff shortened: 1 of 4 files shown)


Znote:
Could it be useful to have a general purpose "load player info from db" function? In order to reduce the amount of separate query requests to the SQL server?

Something along the likes of this:
https://gist.github.com/Znote/252757b013bd390e940cad65e8019fde

It returns player info (such as vocation, correct way to spell the name) upon success, false if player doesn't exist.

luanluciano93:
Couldn't you directly use "return tonumber (money) ~ = nil and money> 0" instead of using the isValidMoney (money) function?

RomeroBA:
> Couldn't you directly use "return tonumber (money) ~ = nil and money> 0" instead of using the isValidMoney (money) function?

Yup, and he could also code everything into a single big ugly block of code, but creating functions for stuff that might be used later is a good idea.

EPuncker:
> Couldn't you directly use "return tonumber (money) ~ = nil and money> 0" instead of using the isValidMoney (money) function?

one check instead of two is fine 😋  i also thought about it but someone else said to keep it and so I did

luanluciano93:
I don't think two checks are good for something that can be done in one. To be honest, it shouldn't. use the compat.lua for nothing.

PR #3088, opened 2020-06-14
Fix waiting list

### LINKED ISSUE #3055: Waiting list bug

> The list stay stack when player abort or get timeOut, the next players can´t login the list add next count ++.
> 
> e.g: Config: maxPlayers = 3, the first three players can login without problems, and the next players logins add in waitlist, the problem comes when player in tha list abort or get timeOut the player stay stack in the list and not cleared from the waitlist.
> So next players cant login.
> 
> After logout all players... Players online 0, and waitlist stay count.

there seems to be a bug in the waiting list after the change  https://github.com/otland/forgottenserver/commit/3f83f9b0bea22bfa7ac484be7161b84e6c8fb1ed by @ranisalt 

there are two threads about this problem [here](https://otland.net/threads/bug-in-wait-list-tfs-1-3-downgraded-8-6.270717/) and [here](https://otland.net/threads/possible-bug-maxplayers-wait-list-slot-stack.270291/)

and a fix by fabian766 [here](https://github.com/SaiyansKing/optimized_forgottenserver/commit/2224d9df1ecc76b131449820012fc77175b01867)

Znote:
Fix #3055

cherry-pick from https://github.com/SaiyansKing/optimized_forgottenserver

[Fix Waitlist - Fixes non-progressive Waitlist.](https://github.com/SaiyansKing/optimized_forgottenserver/commit/2224d9df1ecc76b131449820012fc77175b01867)
[Fix Waitlist compilation on linux](https://github.com/SaiyansKing/optimized_forgottenserver/commit/8c3081bcad2009dc9fd5be76da1b334fa348f7ec)
[Fix waitlist compilation on older gcc versions](https://github.com/SaiyansKing/optimized_forgottenserver/commit/960c547dacd334f8b81d547859561db0a4c461cb)

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -108,10 +108,9 @@ void ProtocolGame::login(const std::string& name, uint32_t accountId, OperatingS
 			}
 		}
 
-		WaitingList& waitingList = WaitingList::getInstance();
-		if (!waitingList.clientLogin(player)) {
-			uint32_t currentSlot = waitingList.getClientSlot(player);
-			uint32_t retryTime = WaitingList::getTime(currentSlot);
+		std::size_t currentSlot;
+		if (!WaitingList::getInstance().clientLogin(player, currentSlot)) {
+			int64_t retryTime = WaitingList::getTime(currentSlot);
 			std::ostringstream ss;
 
 			ss << "Too many players online.\nYou are at place "
@@ -120,7 +119,7 @@ void ProtocolGame::login(const std::string& name, uint32_t accountId, OperatingS
 			auto output = OutputMessagePool::getOutputMessage();
 			output->addByte(0x16);
 			output->addString(ss.str());
-			output->addByte(retryTime);
+			output->addByte(static_cast<uint8_t>(retryTime));
 			send(output);
 			disconnect();
 			return;
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -110,7 +110,7 @@ void ProtocolGame::login(const std::string& name, uint32_t accountId, OperatingS
 
 		std::size_t currentSlot;
 		if (!WaitingList::getInstance().clientLogin(player, currentSlot)) {
-			int64_t retryTime = WaitingList::getTime(currentSlot);
+			uint8_t retryTime = WaitingList::getTime(currentSlot);
 			std::ostringstream ss;
 
 			ss << "Too many players online.\nYou are at place "
@@ -119,7 +119,7 @@ void ProtocolGame::login(const std::string& name, uint32_t accountId, OperatingS
 			auto output = OutputMessagePool::getOutputMessage();
 			output->addByte(0x16);
 			output->addString(ss.str());
-			output->addByte(static_cast<uint8_t>(retryTime));
+			output->addByte(retryTime);
 			send(output);
 			disconnect();
 			return;
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -108,8 +108,8 @@ void ProtocolGame::login(const std::string& name, uint32_t accountId, OperatingS
 			}
 		}
 
-		std::size_t currentSlot;
-		if (!WaitingList::getInstance().clientLogin(player, currentSlot)) {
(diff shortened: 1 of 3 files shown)


src/protocolgame.cpp
```diff
@@ -120,7 +119,7 @@ void ProtocolGame::login(const std::string& name, uint32_t accountId, OperatingS
 			auto output = OutputMessagePool::getOutputMessage();
 			output->addByte(0x16);
 			output->addString(ss.str());
-			output->addByte(retryTime);
+			output->addByte(static_cast<uint8_t>(retryTime));
```
ranisalt, line 122: Why is this cast needed? Also, why doesn't `getTime` return `uint8_t` straight away?

Znote: The client probably expects 1 byte, not 8. (int64). 
But yeah, why not just return it correctly in the first placed. Fixed in my latest commit.

src/waitlist.cpp
```diff
@@ -105,54 +106,42 @@ std::size_t WaitingList::getTime(std::size_t slot)
 	}
 }
 
-bool WaitingList::clientLogin(const Player* player)
+bool WaitingList::clientLogin(const Player* player, std::size_t& currentSlot)
```
ranisalt, line 109: I personally really dislike this C-style function, I would make it return `std::optional<std::size_t>` or rework it so that `0` is an invalid return value that indicates failure. Not something to do right now though.

Znote: This is a very educational experience for me (Im not a c/c++ dev, this PR is cherry picked), could you collaborate on the difference? It would be awesome if you could add a commit to this PR by pushing to the `fix-waiting-list` branch on `Znote/forgottenserver` so I can see how you would change this.

jo3bingham: What @ranisalt means by "C-style function" is that `WaitingList::clientLogin` returns a value, but also takes a value by reference as one of its parameters so that it can "return" more than one value. C doesn't have the concept of multiple return values (e.g., tuples), so devs would have to pass variables as arguments as sort of a hack.

With this C-style function, the caller **always** has to define a variable for `currentSlot` and pass it to `clientLogin`:
```cpp
std::size_t currentSlot;
bool canLogin = WaitingList::getInstance().clientLogin(player, currentSlot);
if (!canLogin) {
    // use the value set in currentSlot to get the retry time
    // currentSlot is only used if canLogin is false
}
```
A better way to do this would be to have `clientLogin` return a `std::optional<std::size_t>>` that you would then check if a value exists, or switch it from a `bool` return type to a `size_t` and there would be no need for the `currentSlot` parameter.

`std::optional`:
```cpp
std::optional<std::size_t> WaitingList::clientLogin(const Player* player)
{
    // do whatever...
    if (slot > 0) {
        return slot;
    }
    return std::nullopt;
}

{
    // there may be a prettier way to do this, but you get the idea
    auto ret = WaitingList::getInstance().clientLogin(player);
    if (ret) {
        auto currentSlot = ret.value();
        ....
    }
}
```
However, the code above checks if slot is 0 anyway, so you could just return slot and not use `std::optional`:
```cpp
size_t WaitingList::clientLogin(const Player* player)
{
    // do whatever...
    return slot;
}

{
    // there may be a prettier way to do this, but you get the idea
    auto currentSlot = WaitingList::getInstance().clientLogin(player);
    if (currentSlot != 0) {
        ....
    }
}
```

ranisalt: Excellent explanation, @jo3bingham

Znote: @jo3bingham Totally AWESOME! I think I'm starting to understand. Let me know what you think of my latest commit. I figured, why return a bool when you can just return currentSlot, and in scenarios where you skip current slot, return 0 as in (your not waiting for anyone, walk right in). That way, the function can be executed as a value to currentSlot in protocolgame.cpp, thus removing the need to pass anything through scetchy param references.

Znote:
> LGTM

@ranisalt You gotta flag your review as "Approved" instead of "Comment" else the merging remaings blocked. (or any other otland dev has to approve, except me because I made this PR and cannot approve it).

ranisalt:
@Znote the gods blessed this pull request so I merged it straight away. Thanks!

PR #3089, opened 2020-06-15
Revert "libs: warn user abpit duplicate storage value (#3072)"

Znote:
This reverts commit d26ee0b80c1a58918714faf28c7b809bc0392566.

It doesn't seem like this code was ever tested, it crashes Lua loading.


lgrossi:
As I said before, that happens when leechers copy code that they don't know about. :sweat_smile:

PR #3092, opened 2020-06-16
Made so buy house and buy premy commands use removeTotalMoney function

EPuncker:
Instead of using only the inventory money it will use both inventory and bank money, just like everything else: npcs, market etc


Znote:
Please clarify the difference between `player:removeMoney` and `player:removeTotalMoney`.

EPuncker:
> Please clarify the difference between `player:removeMoney` and `player:removeTotalMoney`.

done

PR #3093, opened 2020-06-16
Added a config to enable/disable dice usage on depot

EPuncker:
It is not possible any longer to use Dies on Depot Lockers and the Counter between the them. [10.36 update]

### DIFF
--- a/data/actions/scripts/other/die.lua
+++ b/data/actions/scripts/other/die.lua
@@ -1,7 +1,17 @@
+local depotTiles = {11062, 11063}
+local diceEnabledOnDepot = true
+
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 	local position = item:getPosition()
 	local value = math.random(1, 6)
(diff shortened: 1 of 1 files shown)


nekiro:
use tile:hasFlag(TILESTATE_DEPOT)

EPuncker:
> tile:hasFlag(TILESTATE_DEPOT)

that way I would still be able to use the dice on myself, I would need a function to check the 9 SQM around me to look for a depot for it to be perfect

EPuncker:
@nekiro maybe creating a Lua function for this?
`https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L759-L775`

DSpeichert:
Is this ready or a draft? I'm unsure based on your last comment.

EPuncker:
> Is this ready or a draft? I'm unsure based on your last comment.

ready

DSpeichert (approved):

PR #3095, opened 2020-06-16
Fixed Lua monsters walk through issue, fixes #3091

### LINKED ISSUE #3091: RevScriptSys Monsters Stacking Themselves

### Steps to reproduce
1. enable the example monster on data/monster/lua
2. start server, go on GM and do "/m example" as many as possible

### Expected behaviour
everything working as expected, a box of monster is created and then the "there is not enough room" message is sent

### Actual behaviour
They over stack, and walk thru each other, and you can create "infinite" amount of monsters

lol

### PS
Despite my gif being a modified client/tfs, I've talked to several people on discord and they were able to reproduce the issue on a fresh TFS master, including @Znote

EPuncker:
thanks to fabian677 for pointing it out!

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -102,6 +102,11 @@ registerMonsterType.flags = function(mtype, mask)
 		if mask.flags.canPushCreatures ~= nil then
 			mtype:canPushCreatures(mask.flags.canPushCreatures)
 		end
+		-- if a monster can push creatures,
+		-- it should not be pushable
+		if mask.flags.canPushCreatures then
+			mtype:isPushable(false)
+		end
 		if mask.flags.targetDistance then
 			mtype:targetDistance(mask.flags.targetDistance)
 		end


soul4soul:
This fix doesn't feel right. I think it's wrong that pushable is dependent on canpushcreatures. There are separate flags which allow fine grain control over monster behavior. The server shouldn't be overriding that behavior. The developer must be allowed complete control over how they chose to set the flags on each monster they creature.

It would be better to remove the following code and then figure out what is really wrong. I'm positive I have seen monsters on OT servers that can push other monsters and be pushed themselves. It's funny looking the first time you see it. I don't envision any logical issues that would prevent supporting such behavior.

https://github.com/otland/forgottenserver/blob/3612ae8fa10dfee3430bfc692b9d34621ff0282f/src/monsters.cpp#L935-L940

For reference cipbia has both flags. To be fair I haven't ever try to reverse engineer how they parse they flags. Regardless of the cipbia server works TFS should support this behavior.

TFS
```
	<flags>
		<flag summonable="0" />
		<flag attackable="1" />
		<flag hostile="1" />
		<flag illusionable="0" />
		<flag convinceable="0" />
		<flag pushable="0" />
		<flag canpushitems="1" />
		<flag canpushcreatures="1" />
...snip
```

Real cipbia monster files use separate flags too.
```
Flags         = {KickBoxes,
                 KickCreatures,
                 SeeInvisible,
                 Unpushable,
...snip
```

nekiro:
@soul4soul I doubt this is conversion for this pr. If you wish to discuss this, it would be better to create new issue.

PR #3098, opened 2020-06-17
updated items.xml with initdamage

EPuncker:
follow up of https://github.com/otland/forgottenserver/commit/f58ea81e3a26a809cfed46bb2205882e473406cf

fields were not dealing damage on stepIn:
rme

### DIFF
--- a/data/items/items.xml
+++ b/data/items/items.xml
@@ -522,7 +522,7 @@
 	<item fromid="1421" toid="1422" article="a" name="campfire" />
 	<item fromid="1423" toid="1425" article="a" name="campfire">
 		<attribute key="field" value="fire">
-			<attribute key="damage" value="20" />
+			<attribute key="initdamage" value="20" />
 			<attribute key="ticks" value="4000" />
 			<attribute key="count" value="2" />
 			<attribute key="damage" value="10" />
@@ -599,7 +599,7 @@
 		<attribute key="decayTo" value="1488" />
 		<attribute key="duration" value="120" />
 		<attribute key="field" value="fire">
-			<attribute key="damage" value="20" />
+			<attribute key="initdamage" value="20" />
 			<attribute key="ticks" value="10000" />
 			<attribute key="count" value="7" />
 			<attribute key="damage" value="10" />
@@ -610,6 +610,7 @@
 		<attribute key="decayTo" value="1489" />
 		<attribute key="duration" value="120" />
 		<attribute key="field" value="fire">
+			<attribute key="initdamage" value="10" />
 			<attribute key="ticks" value="10000" />
 			<attribute key="count" value="7" />
 			<attribute key="damage" value="10" />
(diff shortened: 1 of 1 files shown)


Znote:
The previous commit shouldn't have been merged into 1.4, it breaks datapack compatibility. 
We might want to consider "damage" as "initdamage" for fields that inflict damageconditions.

nekiro:
@Znote do we really need to be so strict with backwards compatibility? This is simple xml change it is not used in any script or anything.

DSpeichert:
Yes

nekiro:
Well, so right now. There is backwards incompatible field fix in this repo, revert it or merge this.

Znote:
> Well, so right now. There is backwards incompatible field fix in this repo, revert it or merge this.

Can you just add a commit to fix backwards compatibility? 
If this is a field/condition, see if there is an initdamage, if not, see if there is damage and use that as initdamage. 
If the initdamage is configured, the above check isn't neccesary and you can go ahead with initdamage. 

I might try to look into it later, but since you made this it would be great if you could fix this.

nekiro:
> > Well, so right now. There is backwards incompatible field fix in this repo, revert it or merge this.
> 
> Can you commit a fix for backwards compatibility?
> If this is a field/condition, see if there is an initdamage, if not, see if there is damage and use that as initdamage.
> If the initdamage is configured, the above check isn't neccesary and you can go ahead with initdamage.
> 
> I might try to look into it later, but since you made this it would be great if you could fix this. I do like your commit, we just gotta keep it versatile. :)

There is a reason I named it initdamage and not damage. Damage is reserved and used for field damage which is issued by condition. Initdamage is not pushed into condition damage list, so to make it backwards compatible you would have to do some dirty hacks. If you find a way you can tell me.
https://github.com/otland/forgottenserver/blob/f58ea81e3a26a809cfed46bb2205882e473406cf/src/items.cpp#L1211

Znote:
@nekiro 
https://gist.github.com/Znote/ebc3d2602d6e12b9333af2d782c93cdd/revisions

Znote:
#3128 should be merged before this, and this PR can then omit any additions where initdamage is identical to ordinary damage. (and only include changes where initDamage is different to damage).

EPuncker:
> #3128 should be merged before this, and this PR can then remove any additions where initdamage is identical to ordinary damage. (and only include changes where initDamage is different to damage).

done

waqmaz93:
initDamage doesnt work properly. When I step on the field one more time i deals normal damage instead of initDamage.
`<item id="1491" article="an" name="energy field">
	<attribute key="type" value="magicfield"/>
	<attribute key="decayTo" value="0"/>
	<attribute key="duration" value="120"/>
	<attribute key="field" value="energy">
		<attribute key="initDamage" value="30"/>
		<attribute key="ticks" value="10000"/>
		<attribute key="count" value="7"/>
		<attribute key="damage" value="25"/>
	</attribute>
</item>`

What is more if I use double damage (like the first one instead of initDamage:
`<item id="1491" article="an" name="energy field">
	<attribute key="type" value="magicfield"/>
	<attribute key="decayTo" value="0"/>
	<attribute key="duration" value="120"/>
	<attribute key="field" value="energy">
		<attribute key="damage" value="30"/>
		<attribute key="ticks" value="10000"/>
		<attribute key="count" value="7"/>
		<attribute key="damage" value="25"/>
	</attribute>
</item>`

Then when I step first time on the field it deals 25 dmg and 30 dmg  in the same time.

PR #3100, opened 2020-06-18
Dont allocate new attributes when calling getCustomAttribute

nekiro:
item:getCustomAttribute method was recreating attributes class each time it was called, which in case when they don't exist shouldn't be created.
This was also breaking stackables after you call item:getCustomAttribute they would stop stacking.

How to reproduce?
call thing:getCustomAttribute("")
in events Player:onLook on one stackable item and try to stack it with other of the same type.

### DIFF
--- a/src/item.h
+++ b/src/item.h
@@ -647,14 +647,14 @@ class Item : virtual public Thing
 
 		bool removeCustomAttribute(int64_t key) {
 			if (!attributes) {
-				return true;
+				return false;
 			}
 			return getAttributes()->removeCustomAttribute(key);
 		}
 
 		bool removeCustomAttribute(const std::string& key) {
 			if (!attributes) {
-				return true;
+				return false;
 			}
 			return getAttributes()->removeCustomAttribute(key);
 		}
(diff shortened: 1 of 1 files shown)


ranisalt:
In that case, just skip `getAttributes` altogether, AFAIK the purpose of that is to allocate and prevent null dereference but you are checking it twice.

nekiro:
@ranisalt what? If attributes is nullptr and we try to getCustomAttribute out of it then it will always create new attributes instance just for that purpose which is pointless. Not sure if you noticed but these calls are in getCustomAttribute method.

ranisalt:
@nekiro I was talking about removing the method `getAttribute`, now I see it is used elsewhere so nevermind.

CristoferMartins:
I think its missing the same check in the functions removeCustomAttribute.

nekiro:
> I think its missing the same check in the functions removeCustomAttribute.

Agree on this one. The same thing will happen if you try to remove attribute which does not exist.

nekiro:
This is ready to be merged.

CristoferMartins:
Its inconsistent that removeCustomAttribute returns true when there is no attributes as getAttributes()->removeCustomAttribute() return false if the attribute doesn't exist.

Issue #3113, opened 2020-07-01
Client will act weird if healthNow is bigger than heathMax


EPuncker:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. go to any monster file and make its health now value bigger than the health max

PS: don't forget to check NPC and players health too if its the case

### Expected behaviour
That shouldn't be allowed at all

### Actual behaviour
![2020-07-01 06_16_08](https://user-images.githubusercontent.com/4684880/86226949-c5539600-bb62-11ea-820d-b96dce23ec8b.png)

and sometimes it debugs the client

EPuncker:
fixed on https://github.com/otland/forgottenserver/commit/01d4556d5271e931f30e33a4be678c5c635d4102

PR #3116, opened 2020-07-02
Fixed if healthNow is bigger than healthMax!

MillhioreBT:
This fixes the problem mentioned here: #3113

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -36,11 +36,17 @@ end
 registerMonsterType.maxHealth = function(mtype, mask)
 	if mask.maxHealth then
 		mtype:maxHealth(mask.maxHealth)
+		if mask.maxHealth < mtype:health() then
+			mtype:health(mask.maxHealth)
+		end
 	end
 end
 registerMonsterType.health = function(mtype, mask)
 	if mask.health then
 		mtype:health(mask.health)
+		if mask.health > mtype:maxHealth() then
+			mtype:maxHealth(mask.health)
+		end
 	end
 end
 registerMonsterType.runHealth = function(mtype, mask)
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -50,7 +50,7 @@ Monster::Monster(MonsterType* mType) :
 	defaultOutfit = mType->info.outfit;
 	currentOutfit = mType->info.outfit;
 	skull = mType->info.skull;
-	health = mType->info.health;
+	health = std::min(mType->info.health, mType->info.healthMax);
 	healthMax = mType->info.healthMax;
 	baseSpeed = mType->info.baseSpeed;
 	internalLight = mType->info.light;
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -50,7 +50,7 @@ Monster::Monster(MonsterType* mType) :
 	defaultOutfit = mType->info.outfit;
 	currentOutfit = mType->info.outfit;
 	skull = mType->info.skull;
-	health = mType->info.health
+	health = mType->info.health;
 	healthMax = mType->info.healthMax;
 	baseSpeed = mType->info.baseSpeed;
 	internalLight = mType->info.light;
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -50,7 +50,7 @@ Monster::Monster(MonsterType* mType) :
 	defaultOutfit = mType->info.outfit;
 	currentOutfit = mType->info.outfit;
 	skull = mType->info.skull;
-	health = std::min(mType->info.health, mType->info.healthMax);
+	health = mType->info.health
 	healthMax = mType->info.healthMax;
 	baseSpeed = mType->info.baseSpeed;
 	internalLight = mType->info.light;
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2289,7 +2289,8 @@ void ProtocolGame::sendCreatureHealth(const Creature* creature)
 	if (creature->isHealthHidden()) {
 		msg.addByte(0x00);
 	} else {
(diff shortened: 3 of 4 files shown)


nekiro:
What if player has this issue?

DSpeichert:
Please use meaningful PR titles.

MillhioreBT:
> What if player or npc has this issue?

I don't know how Npc's works in OTC, but in the TibiaClient official the health bar is not visible, so I'm not sure if we should do `std::min` with the nowHealth of the Npc!

nekiro:
I think same thing will happen for players, so you might want to check it.

src/protocolgame.cpp
```diff
@@ -2289,7 +2289,8 @@ void ProtocolGame::sendCreatureHealth(const Creature* creature)
 	if (creature->isHealthHidden()) {
 		msg.addByte(0x00);
 	} else {
-		msg.addByte(std::ceil((static_cast<double>(creature->getHealth()) / std::max<int32_t>(creature->getMaxHealth(), 1)) * 100));
+		int32_t maxHealth = std::max(creature->getMaxHealth(), 1);
+		msg.addByte(std::ceil((static_cast<double>(std::min(creature->getHealth(), maxHealth)) / maxHealth) * 100));
```
nekiro, line 2293: This will only fix client side (so showing it in-game) server still gonna hold invalid values I guess?

MillhioreBT: I am not 100% sure, but as I understand it, this information comes from the server to the client, and this calculation should not be done at any time from the client side, so if everything is fine from the server, it should not be wrong in the tibia-client, if I'm wrong please notify me.

nekiro: yeah, but value in server gonna still be broken like max health being for example 100 while health is 200. I don't think this should happen server side, so the fix should be in server sources not at the time when they are sent to client.

EPuncker: and/or at least add a console warning about the mismatching values

MillhioreBT: > yeah, but value in server gonna still be broken like max health being for example 100 while health is 200. I don't think this should happen server side, so the fix should be in server sources not at the time when its sent to client.

I understand what you are saying, just add this in case something weird happens (example: force current health from sources.), But I have not been able to find anything that can affect amounts of current health and maximum health, cannot be broken with luascripts , and neither with equipment methods.

renatorib: I will leave my opinion about it

You shouldn't want to suppress the side effects of something wrong. You will be hiding that there is something wrong. If something is wrong, it must be visible for someone to see and correct it.

You want to correct what is wrong. In this case, the best thing to do is to alert the user (with warning) that healthNow is greater than healthMax (validation of monsters.xml?) (when the monster/creature is instantiated?)

Fix the cause, not the symptom!

src/monster.cpp
```diff
@@ -50,7 +50,7 @@ Monster::Monster(MonsterType* mType) :
 	defaultOutfit = mType->info.outfit;
 	currentOutfit = mType->info.outfit;
 	skull = mType->info.skull;
-	health = std::min(mType->info.health, mType->info.healthMax);
+	health = mType->info.health
```
EPuncker: missing ;

MillhioreBT: `std::min` is no longer needed because the MonsterType on loading fixes the healthNow overflow, so there will never be `mType->info.health` greater than `mType->info.healthMax`
[monsters.cpp](https://github.com/otland/forgottenserver/pull/3116/commits/8160012a428efad21427528e309f7dd280e1307e#diff-721c64f80744c3e002360fff5827e2c8R883) and [register_monster_type.lua](https://github.com/otland/forgottenserver/pull/3116/files#diff-77b54c98309c519e7f79ba6b583c1826R39)

renatorib: He fixed the missing ; in this commit https://github.com/otland/forgottenserver/pull/3116/commits/04558bcd14b2cbfd876abdd6d96166a4d8b9b921

data/scripts/lib/register_monster_type.lua
```diff
@@ -36,11 +36,17 @@ end
 registerMonsterType.maxHealth = function(mtype, mask)
 	if mask.maxHealth then
 		mtype:maxHealth(mask.maxHealth)
+		if mask.maxHealth < mtype:health() then
+			mtype:health(mask.maxHealth)
```
nekiro, line 40: just math.min(mask.maxHealth, mtype:health()) same with this second change

Znote:
I think you can avoid the extra if conditions by just replacing
`mtype:health(mask.health)`
with
`mtype:health(std::min(mask.health, mtype:maxHealth()))`
as std::min will pick the lowest of the two values.

EPuncker:
could you also add a warning in the npc.cpp file about npc health bars? I forgot about them hehe

EPuncker (changes requested):
ok now the only missing things is to also add the warning to npc.cpp file and the math.min thing nekiro suggested

MillhioreBT:
I can't recover this repository, so whoever dares to propose these changes in another PR is fine.
If they want to close it, that's fine too!

EPuncker:
> I can't recover this repository, so whoever dares to propose these changes in another PR is fine.
> If they want to close it, that's fine too!

why not open a new PR with the changes? 😄

DSpeichert (approved):

DSpeichert:
@MillhioreBT do you intend to open another PR to replace this and implements suggestions or this is final version?

MillhioreBT:
> @MillhioreBT do you intend to open another PR to replace this and implements suggestions or this is final version?

I did not plan to do it, but since you say so, of course ;) I will open another PR with these changes and implement a suggestion that I could not add because my repository was deleted and is unrecoverable

MillhioreBT:
Closed, because now it will be treated in another PR #3312

PR #3117, opened 2020-07-02
Adding back sendAnimatedText if client version is lower than 9

EPuncker:
this is debatable, this should be accepted or should we also remove protocolold code?!

code from @SaiyansKing

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -606,7 +606,7 @@ function doPlayerPopupFYI(cid, message) local p = Player(cid) return p and p:pop
 function doSendTutorial(cid, tutorialId) local p = Player(cid) return p and p:sendTutorial(tutorialId) or false end
 function doAddMapMark(cid, pos, type, description) local p = Player(cid) return p and p:addMapMark(pos, type, description or "") or false end
 function doPlayerSendTextMessage(cid, type, text, ...) local p = Player(cid) return p and p:sendTextMessage(type, text, ...) or false end
-if Game.getClientVersion < 900 then
+if Game.getClientVersion().min < 900 then
 	function doSendAnimatedText(pos, text, color, player) Game.sendAnimatedText(text, pos, color, player) return true end
 else
 	function doSendAnimatedText() debugPrint("Deprecated function.") return true end
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -606,7 +606,7 @@ function doPlayerPopupFYI(cid, message) local p = Player(cid) return p and p:pop
 function doSendTutorial(cid, tutorialId) local p = Player(cid) return p and p:sendTutorial(tutorialId) or false end
 function doAddMapMark(cid, pos, type, description) local p = Player(cid) return p and p:addMapMark(pos, type, description or "") or false end
 function doPlayerSendTextMessage(cid, type, text, ...) local p = Player(cid) return p and p:sendTextMessage(type, text, ...) or false end
-if(CLIENT_VERSION_MIN < 900) then
+if Game.getClientVersion < 900 then
 	function doSendAnimatedText(pos, text, color, player) Game.sendAnimatedText(text, pos, color, player) return true end
 else
 	function doSendAnimatedText() debugPrint("Deprecated function.") return true end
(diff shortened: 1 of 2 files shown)


MillhioreBT:
I don't think it is necessary to have this function, since it could never be used, because the tibia client will not have this exact functionality anymore, but possibly in the most current 12+ versions there is already some animated text that I don't know

EPuncker:
> I don't think it is necessary to have this function, since it could never be used, because the tibia client will not have this exact functionality anymore, but possibly in the most current 12+ versions there is already some animated text that I don't know

this can be still used with OTClient and still need confirmation if it works with @SaiyansKing and @slavidodo clients, but I think it does

MillhioreBT:
> > I don't think it is necessary to have this function, since it could never be used, because the tibia client will not have this exact functionality anymore, but possibly in the most current 12+ versions there is already some animated text that I don't know
> 
> this can be still used with OTClient and still need confirmation if it works with @SaiyansKing and @slavidodo clients, but I think it does

I did not remember that there is such a thing, in this case, if it should be included.

data/lib/core/game.lua
```diff
@@ -66,3 +66,25 @@ end
 function Game.setStorageValue(key, value)
 	globalStorageTable[key] = value
 end
+
+if(CLIENT_VERSION_MIN < 900) then
```
nekiro, line 70: if CLIENT_VERSION_MIN < 900 then

data/lib/compat/compat.lua
```diff
@@ -606,7 +606,11 @@ function doPlayerPopupFYI(cid, message) local p = Player(cid) return p and p:pop
 function doSendTutorial(cid, tutorialId) local p = Player(cid) return p and p:sendTutorial(tutorialId) or false end
 function doAddMapMark(cid, pos, type, description) local p = Player(cid) return p and p:addMapMark(pos, type, description or "") or false end
 function doPlayerSendTextMessage(cid, type, text, ...) local p = Player(cid) return p and p:sendTextMessage(type, text, ...) or false end
-function doSendAnimatedText() debugPrint("Deprecated function.") return true end
+if(CLIENT_VERSION_MIN < 900) then
```
nekiro, line 609: if CLIENT_VERSION_MIN < 900 then

nekiro:
Also I don't think CLIENT_VERSION_MIN is exposed to lua and you didn't add it to be in luascript.cpp

EPuncker:
> Also I don't think CLIENT_VERSION_MIN is exposed to lua and you didn't add it to be in luascript.cpp
> Also I don't think we have Game.sendAnimatedText so you missed this too.

I added it on data/lib/core/game.lua or am I missing something?

data/lib/core/game.lua
```diff
@@ -66,3 +66,25 @@ end
...
+		msg:addByte(0x84)
+		msg:addPosition(pos)
+		msg:addByte(color)
+		msg:addString(message)
+
+		local spectators = nil
+		if player ~= nil then
+			spectators = {player}
+		else
+			spectators = Game.getSpectators(pos)
```
nekiro, line 82: populate parameters in Game.getSpectators. I don't think you want to send animated text to monsters or npcs right?

data/lib/compat/compat.lua
```diff
@@ -606,7 +606,11 @@ function doPlayerPopupFYI(cid, message) local p = Player(cid) return p and p:pop
 function doSendTutorial(cid, tutorialId) local p = Player(cid) return p and p:sendTutorial(tutorialId) or false end
 function doAddMapMark(cid, pos, type, description) local p = Player(cid) return p and p:addMapMark(pos, type, description or "") or false end
 function doPlayerSendTextMessage(cid, type, text, ...) local p = Player(cid) return p and p:sendTextMessage(type, text, ...) or false end
-function doSendAnimatedText() debugPrint("Deprecated function.") return true end
+if Game.getClientVersion < 900 then
```
nekiro, line 609: this wont work.
Game.getClientVersion().min < 900

data/lib/core/game.lua
```diff
@@ -66,3 +66,25 @@ end
 function Game.setStorageValue(key, value)
 	globalStorageTable[key] = value
 end
+
+if Game.getClientVersion < 900 then
```
nekiro, line 70: this wont work.
Game.getClientVersion().min < 900

DSpeichert:
This version of the engine never intended to support multiple versions (or "old protocol"), so things like that should just be turned into compat with either implementation that "converts" it to something that currently supported client (10.98) knows, or a warning message.

EPuncker:
> This version of the engine never intended to support multiple versions (or "old protocol"), so things like that should just be turned into compat with either implementation that "converts" it to something that currently supported client (10.98) knows, or a warning message.

but OTC still supports it, so maybe should I add a new config.lua entry for enabling/disabling it, warning that it will only work on OTC or client 8.72 and lower?

DSpeichert:
I wouldn't add that, it's simply not in 10.98, regardless of what OTC does.

EPuncker:
ok 🤷

PR #3120, opened 2020-07-05
Fixed variable name on playerdeath.lua

### LINKED ISSUE #3119: Question in playerdeath
**forgottenserver/data/creaturescripts/scripts/playerdeath.lua**
> db.query("INSERT INTO `player_deaths` (`player_id`, `time`, `level`, `killed_by`, `is_player`, `mostdamage_by`, `mostdamage_is_player`, `unjustified`, `mostdamage_unjustified`) VALUES (" .. playerGuid .. ", " .. os.time() .. ", " .. player:getLevel() .. ", " .. db.escapeString(killerName) .. ", " .. byPlayer .. ", " .. db.escapeString(mostDamageName) .. ", " .. byPlayerMostDamage .. ", " .. (unjustified and 1 or 0) .. ", " .. (mostDamageUnjustified and 1 or 0) .. ")")

Is the variable correct In line "unjustified and 1 or 0"?

> db.asyncQuery("INSERT INTO `guildwar_kills` (`killer`, `target`, `killerguild`, `targetguild`, `time`, `warid`) VALUES (" .. db.escapeString(killerName) .. ", " .. db.escapeString(player:getName()) .. ", " .. killerGuild .. ", " .. targetGuild .. ", " .. os.time() .. ", " .. warId .. ")")

Wouldn't it be better to put the player:getName() function outside db.asyncQuery?
> local playerName = player:getName()
> db.asyncQuery("INSERT INTO `guildwar_kills` (`killer`, `target`, `killerguild`, `targetguild`, `time`, `warid`) VALUES (" .. db.escapeString(killerName) .. ", " .. db.escapeString(playerName) .. ", " .. killerGuild .. ", " .. targetGuild .. ", " .. os.time() .. ", " .. warId .. ")")

EPuncker:
closes #3119

### DIFF
--- a/data/creaturescripts/scripts/playerdeath.lua
+++ b/data/creaturescripts/scripts/playerdeath.lua
@@ -47,7 +47,7 @@ function onDeath(player, corpse, killer, mostDamageKiller, lastHitUnjustified, m
 	end
 
 	local playerGuid = player:getGuid()
-	db.query("INSERT INTO `player_deaths` (`player_id`, `time`, `level`, `killed_by`, `is_player`, `mostdamage_by`, `mostdamage_is_player`, `unjustified`, `mostdamage_unjustified`) VALUES (" .. playerGuid .. ", " .. os.time() .. ", " .. player:getLevel() .. ", " .. db.escapeString(killerName) .. ", " .. byPlayer .. ", " .. db.escapeString(mostDamageName) .. ", " .. byPlayerMostDamage .. ", " .. (unjustified and 1 or 0) .. ", " .. (mostDamageUnjustified and 1 or 0) .. ")")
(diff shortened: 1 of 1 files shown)


data/creaturescripts/scripts/playerdeath.lua
```diff
@@ -81,7 +81,8 @@ function onDeath(player, corpse, killer, mostDamageKiller, lastHitUnjustified, m
 				end
 
 				if warId ~= false then
-					db.asyncQuery("INSERT INTO `guildwar_kills` (`killer`, `target`, `killerguild`, `targetguild`, `time`, `warid`) VALUES (" .. db.escapeString(killerName) .. ", " .. db.escapeString(player:getName()) .. ", " .. killerGuild .. ", " .. targetGuild .. ", " .. os.time() .. ", " .. warId .. ")")
+					local playerName = player:getName()
```
MillhioreBT, line 84: All is fine, but why should you define a variable with the player name to use in formatting the string passed to the `db.asyncQuery` function? I think it does not affect anything to leave it as it was, if there is a reason for this, I really want to know.

EPuncker: @MillhioreBT db.asyncQuery is not executed instantly, so the player might not "exist" anymore, the same issue from addEvent I guess

MillhioreBT: There is no doubt that the function does not execute immediately, but the parameter yes it does, you can enter an print("hello") and check that the string is formatted immediately when the function is called.

nekiro: you are not passing player userdata, but string, string won't get nulled anywhere. Using player:getName() is perfectly fine here.

PR #3127, opened 2020-07-08
Fixed monsters paralyzing themselves

### LINKED ISSUE #3070: Monsters paralyze themselves
https://github.com/nekiro/forgottenserver/issues/44 [Monsters paralyze themselves]

Yes, this from downgrade but Nekiro suggest that combat and spells are not modified so I decided report bug here.

If monster hit in area, not single point, eg with paralyze, and will be in range then paralyze themselves.

EPuncker:
fixes #3070
fix by @joseluis2g
tested by me

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -714,7 +714,7 @@ void Combat::doCombat(Creature* caster, const Position& position) const
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


EPuncker:
antes2
depois
before and after comparison

PR #3128, opened 2020-07-09
Use damage instead of initDamage if not defined

Znote:
If a field item is missing initdamage property, load damage as initdamage.
This is to retain backward compatibility after #2815 which was addressed in #3098

### DIFF
--- a/src/items.cpp
+++ b/src/items.cpp
@@ -1188,8 +1188,7 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 						uint32_t ticks = 0;
 						int32_t start = 0;
 						int32_t count = 1;
-						int32_t initDamage = 0;
-						bool initDamageConfigured = false;
+						int32_t initDamage = -1;
 						int32_t damage = 0;
 						for (auto subAttributeNode : attributeNode.children()) {
 							pugi::xml_attribute subKeyAttribute = subAttributeNode.attribute("key");
@@ -1204,8 +1203,7 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 
 							tmpStrValue = asLowerCaseString(subKeyAttribute.as_string());
 							if (tmpStrValue == "initdamage") {
-								initDamage = -pugi::cast<int32_t>(subValueAttribute.value());
-								initDamageConfigured = true;
+								initDamage = pugi::cast<int32_t>(subValueAttribute.value());
 							} else if (tmpStrValue == "ticks") {
 								ticks = pugi::cast<uint32_t>(subValueAttribute.value());
 							} else if (tmpStrValue == "count") {
@@ -1230,10 +1228,10 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 
 						// datapack compatibility, presume damage to be initialdamage if initialdamage is not declared.
 						// To avoid any initialdamage, add initialdamage xml property and set it to 0.
-						if (!initDamageConfigured && damage != 0) {
+						if (initDamage > 0) {
+							conditionDamage->setInitDamage(-initDamage);
+						} else if (initDamage == -1 && damage != 0) {
 							conditionDamage->setInitDamage(damage);
-						} else if (initDamageConfigured && initDamage != 0) {
-							conditionDamage->setInitDamage(initDamage);
 						}
 
 						conditionDamage->setParam(CONDITION_PARAM_FIELD, 1);
--- a/src/items.cpp
+++ b/src/items.cpp
@@ -1227,8 +1227,9 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 						}
 
 						// datapack compatibility, presume damage to be initialdamage if initialdamage is not declared.
-						// To avoid any initialdamage, add initialdamage xml property and set it to 0.
-						if (initDamage > 0) {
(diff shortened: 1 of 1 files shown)


src/items.cpp
```diff
@@ -1188,6 +1188,9 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 						uint32_t ticks = 0;
 						int32_t start = 0;
 						int32_t count = 1;
+						int32_t initdamage = 0;
+						bool initdamageConfigured = false;
```
nekiro, line 1192: this bool seeems redundant? You can simply check initDamage variable if its not 0
also rename initdamage to initDamage

src/items.cpp
```diff
@@ -1224,6 +1228,14 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 							}
 						}
 
+						// datapack compatibility, presume damage to be initialdamage if initialdamage is not declared.
+						// To avoid any initialdamage, add initialdamage xml property and set it to 0.
+						if (!initdamageConfigured && damage != 0) {
```
nekiro, line 1233: here simply check if initDamage != 0 and set initDamage else set damage, damage is mandatory for field to work, so it should never be 0 in this case

Znote: No, because initdamage can be configured to 0 in order to avoid any initial damage. (else damage will be initialdamage). 
Example if you want a field to intentionally begin a burn condition, but don't want it to initially deal any damage. 

The bool figures out if the 0 value is from code initialization, or loaded from ItemNode. This is critical to determine if initialdamage should be overriden with damage. I originally intended to initialize initdamage to null, but I don't think int32_t support null initializations, so I supplemented it with a bool.

src/items.cpp
Znote: No because initDamage 0 does have a purpose. (to let us know that we don't want damage to override initdamage). 
Yes I can rename it.

src/items.cpp
nekiro: @Znote before my changes, you didn't have this option, so why do you have it now?

nekiro: You can just set initDamage to -1 to check if it was parsed from xml file instead of 0 and this bool is redundant then

Znote: Ohh, I though you meant initDamage. damage is **not** mandatory as you say, there are fields (like small fire fields) that are completely without any damage. Thus a check for damage != 0 is neccesary. (I think)

Znote: > You can just set initDamage to -1 to check if it was parsed from xml file instead of 0 and this bool is redundant then

No because `<attribute key="initdamage" value="1" />` will translate to -1 in the source. :P (-1 reduction to hit points). 
Theoretically, the opposite is true for a field that heals you. (healing field?). So I shouldn't twist the logic either.

nekiro: no it wont convert, just save the value raw and don't cast it to negative later when you have the code to use initDamage variable just use -initDamage

Znote: Aight, done.

Znote:
initDamage

src/items.cpp
```diff
@@ -1224,6 +1226,15 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 							}
 						}
 
+						// datapack compatibility, presume damage to be initialdamage if initialdamage is not declared.
+						// initDamage = 0 (dont override initDamage with damage, dont set any initDamage)
+						// initDamage = -1 (undefined, override initDamage with damage)
+						if (initDamage > 0 || initDamage < -1) {
```
nekiro, line 1232: < -1? in this case you gonna set opposite, someone gonna write -200 in xml file and you gonna cast it like -initDamage so it will end up in 200, so it will heal not damage.

Znote: Yes, this is intentional. If you look at my first gif in the above post, with this snippet:
```xml
<attribute key="field" value="fire">
	<attribute key="initdamage" value="-2000" />
	<attribute key="ticks" value="2000" />
	<attribute key="count" value="5" />
	<attribute key="damage" value="400" />
</attribute>
```

Instead of dealing damage, it will heal you 2000 hit points, but then over 2x5 seconds you will loose the obtained 2000 hitpoints. 

Or the opposite:
```xml
<attribute key="field" value="fire">
	<attribute key="initdamage" value="2000" />
	<attribute key="ticks" value="2000" />
	<attribute key="count" value="5" />
	<attribute key="damage" value="-400" />
</attribute>
```
Initdamage will deal 2000 damage, and then it will heal you back over 2x5 seconds. 
This is flexibility. :) You could make custom healing fields.

nekiro: well, you can easily achieve this with simple movement script and condition_regeneration, so I don't know if that's something we should support in condition_damage which says DAMAGE, makes no sense to make it support healing.

Znote: > well, you can easily achieve this with simple movement script and condition_regeneration, so I don't know if that's something we should support in condition_damage which says DAMAGE, makes no sense to make it support healing.

The damage property already supports this (without this PR). This PR just makes initialdamage consistent with how damage works. (And I think its cool). The opposite of damage is healing, which correlates to a negative damage value. 

If you do not think damage should have the posibility to do healing, it should be considered an issue separately from this PR?

nekiro: well my point is, you can by accident break your server, when you put -200 instead of 200, cause someone might think damage is negative all the time. Imagine having fields that will heal for absurd amounts of hp.

Issue #3126, opened 2020-07-08
Add alias words to revscriptsys TalkActions


EPuncker:
### Explanation of what you want to do that is currently impossible
Currently talkactions support only one **word**, if we want to add more we need to duplicate the code

### Desired functionality
It would be nice if we could add several words when registering it, or even an additional setting like:

`talkaction:alias("a", "b", "c")`

### Prior art
```
local talkaction = TalkAction("!hello", "!hi", "/hello")

function talkaction.onSay(player)
	player:say("Hello!", TALKTYPE_YELL)
	return true
end

talkaction:register()
```

PR #3132, opened 2020-07-13
Multiple words to talkaction RevScript

MillhioreBT:
I liked this implementation a lot and I always wondered if someone else would think the same, this was mentioned by @EPuncker in the following Issue (#3126)

Example:
`data/scripts/talkactions/position.lua`
```lua
local talk = TalkAction("/pos", "/position", "/mypos")

	if player:getGroup():getAccess() and param ~= "" then
		player:teleportTo(Position(split[1], split[2], split[3]))
	else
		player:sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, "Your current position is: " .. position.x .. ", " .. position.y .. ", " .. position.z .. ".")
	end
	return false
end

talk:separator(" ")
talk:register()
```

_For a couple of days now, the compilations have not happened correctly, no matter if the code is ok, this is a github error?_

### DIFF
--- a/src/talkaction.cpp
+++ b/src/talkaction.cpp
@@ -144,7 +144,7 @@ std::string TalkAction::getScriptEventName() const
 	return "onSay";
 }
 
-bool TalkAction::executeSay(Player* player, const std::string& param, SpeakClasses type) const
+bool TalkAction::executeSay(Player* player, const std::string& word, const std::string& param, SpeakClasses type) const
 {
 	//onSay(player, words, param, type)
 	if (!scriptInterface->reserveScriptEnv()) {
@@ -162,7 +162,7 @@ bool TalkAction::executeSay(Player* player, const std::string& param, SpeakClass
 	LuaScriptInterface::pushUserdata<Player>(L, player);
 	LuaScriptInterface::setMetatable(L, -1, "Player");
 
-	LuaScriptInterface::pushString(L, words);
+	LuaScriptInterface::pushString(L, word);
 	LuaScriptInterface::pushString(L, param);
 	lua_pushnumber(L, type);
 
--- a/src/talkaction.cpp
+++ b/src/talkaction.cpp
@@ -83,40 +83,39 @@ TalkActionResult_t TalkActions::playerSaySpell(Player* player, SpeakClasses type
 {
 	size_t wordsLength = words.length();
 	for (auto it = talkActions.begin(); it != talkActions.end(); ) {
-		const std::string& talkactionWords = it->first;
-		size_t talkactionLength = talkactionWords.length();
-		if (wordsLength < talkactionLength || strncasecmp(words.c_str(), talkactionWords.c_str(), talkactionLength) != 0) {
-			++it;
-			continue;
-		}
-
-		std::string param;
-		if (wordsLength != talkactionLength) {
-			param = words.substr(talkactionLength);
-			if (param.front() != ' ') {
-				++it;
+		for (auto& talkactionWords : it->second.getWothers()) {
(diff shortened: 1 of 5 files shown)


EPuncker:
nice, I'm going to test it soon!

DSpeichert:
Is this because 0.4 allowed using multiple words separated by `;` ?

MillhioreBT:
> Is this because 0.4 allowed using multiple words separated by `;` ?

That's where the idea comes from and it's a good one.
but it would also be good to customize the separator to your liking
for example:
```xml
<talkactions>
	<separator words=";" />
```

> tested and working, my only concern is if there is a limit of how many aliases we can use? and if its good for we to put a limit just for precaution, like 16 words or so

It would not be wrong to have a limit and thus avoid errors caused intentionally

src/talkaction.cpp
```diff
@@ -83,40 +83,39 @@ TalkActionResult_t TalkActions::playerSaySpell(Player* player, SpeakClasses type
...
-			++it;
-			continue;
-		}
-
-		std::string param;
-		if (wordsLength != talkactionLength) {
-			param = words.substr(talkactionLength);
-			if (param.front() != ' ') {
-				++it;
+		for (auto& talkactionWords : it->second.getWothers()) {
```
EvilHero90, line 86: Please re structure the function ```getWords``` instead of using a newly created function which doesn't give us a clue what it should even mean.

src/talkaction.h
```diff
@@ -26,6 +26,7 @@
 
 class TalkAction;
 using TalkAction_ptr = std::unique_ptr<TalkAction>;
+using TalkAction_words = std::vector<std::string>;
```
EvilHero90, line 29: Instead of doing a ```using``` here you should rather convert ```std::string words``` into a vector directly so we are not leaving dead stuff

src/talkaction.h
```diff
@@ -45,13 +46,17 @@ class TalkAction : public Event
 		}
 		void setWords(std::string word) {
 			words = word;
```
EvilHero90, line 48: same here convert it so you only use a vector

src/talkaction.h
```diff
@@ -45,13 +46,17 @@ class TalkAction : public Event
...
+			wothers.push_back(word);
 		}
 		std::string getSeparator() const {
 			return separator;
 		}
 		void setSeparator(std::string sep) {
 			separator = sep;
 		}
+		const TalkAction_words getWothers() const {
```
EvilHero90, line 57: this could be erased and converted to ```getWords``` then

src/talkaction.h
```diff
@@ -62,6 +67,7 @@ class TalkAction : public Event
 
 		std::string words;
```
EvilHero90, line 68: ```std::vector<std::string> words;```

src/talkaction.h
```diff
@@ -62,6 +67,7 @@ class TalkAction : public Event
...
 		std::string separator = "\"";
+		TalkAction_words wothers = {};
```
EvilHero90, line 70: could be erased then

MillhioreBT:
@EvilHero90 I have taken into account everything you mentioned to me, I will do something better in all aspects ;) thank you <3

EvilHero90 (approved):

DSpeichert (approved):

MillhioreBT:
Good evening, I did not realize when this PR was merged, I was not sure at all if it was working 100%, thank you very much for reviewing it, I will try to do it from another approach in which it is simpler and more functional <3

PR #3134, opened 2020-07-13
Added possibility to control spawn interval rate on config.lua

EPuncker:
- Added a comment of what rateSpawn does, because it was confusing and did not do what the user expected it to do
- Added spawnIntervalRate that control the rate of monster spawning interval:
100 = default
50 = half spawntime
200 = double spawntime
and so on...


config.lua.dist
```diff
@@ -101,11 +101,13 @@ rateExp = 5
 rateSkill = 3
 rateLoot = 2
 rateMagic = 3
-rateSpawn = 1
 
 -- Monsters
+-- spawnIntervalRate control the rate of monster spawning interval
+-- 100 = default, 50 = half spawntime, 200 = double spawntime
```
nekiro, line 107: its percentage so add comment, so people know. 100 is 100%, 50 is 50% in this case.

nekiro:
What do you think about making this like the old good spawnRate (which didnt work) 1 is 1x, 2x, 3x, 4x etc, so respectively 100%, 200%, 300%, 400% seems easier and simplier to me

EPuncker:
> What do you think about making this like the old good spawnRate (which didnt work) 1 is 1x, 2x, 3x, 4x etc, so respectively 100%, 200%, 300%, 400% seems easier and simplier to me

yeah I thought about it too, but then it wouldn't support slower spawn rates, just faster :( unless the PR with accepting float numbers get accepted

nekiro:
> > What do you think about making this like the old good spawnRate (which didnt work) 1 is 1x, 2x, 3x, 4x etc, so respectively 100%, 200%, 300%, 400% seems easier and simplier to me
> 
> 
> 
> yeah I thought about it too, but then it wouldn't support slower spawn rates, just faster :( unless a PR that adds the possibility of accepting float values get implemented

I really doubt someone gonna want to slower their respawn rate xD

config.lua.dist
```diff
@@ -101,11 +101,13 @@ rateExp = 5
...
 rateLoot = 2
 rateMagic = 3
-rateSpawn = 1
 
 -- Monsters
+-- spawnIntervalRate controls the rate of monster spawning interval
+-- (spawntime divided by spawnIntervalRate) 1 = 100%, 2 = 50%, etc
 deSpawnRange = 2
 deSpawnRadius = 50
+spawnIntervalRate = 1
```
nekiro, line 110: If you make it 0, server will crash. Force value as 1 using std:max(1, configVal)

EPuncker:
@DSpeichert I hope it looks better now

omarcopires:
LGTM

DSpeichert (dismissed):

DSpeichert:
Needs rebase

EPuncker:
> Needs rebase

done

Znote (dismissed):

src/spawn.cpp
```diff
@@ -100,7 +100,8 @@ bool Spawns::loadFromXml(const std::string& filename)
 					centerPos.y + pugi::cast<uint16_t>(childNode.attribute("y").value()),
 					centerPos.z
 				);
-				int32_t interval = pugi::cast<int32_t>(childNode.attribute("spawntime").value()) * 1000;
+
+				int32_t interval = pugi::cast<int32_t>(childNode.attribute("spawntime").value()) * 1000 * (g_config.getNumber(ConfigManager::SPAWN_INTERVAL_RATE) / 100);
 				if (interval >= MINSPAWN_INTERVAL && interval <= MAXSPAWN_INTERVAL) {
```
DSpeichert, line 105: There is a min and a max, should we document those in config.lua to make user aware?

config.lua.dist
```diff
@@ -98,15 +98,19 @@ serverSaveShutdown = true
...
+-- rateSpawn limits the number of monsters that will spawn within the same spawn object
 rateExp = 5
 rateSkill = 3
 rateLoot = 2
 rateMagic = 3
 rateSpawn = 1
 
 -- Monsters
+-- spawnIntervalRate controls the rate of monster spawning interval
+-- its in percentage so 100 = 100%, 50 = 50% and so on
```
DSpeichert, line 110: Could be worth noting that higher number = less frequent spawn

config.lua.dist
```diff
@@ -98,15 +98,21 @@ serverSaveShutdown = true
...
 rateExp = 5
 rateSkill = 3
 rateLoot = 2
 rateMagic = 3
 rateSpawn = 1
 
 -- Monsters
+-- spawnIntervalRate controls the rate of monster spawning interval
+-- its in percentage so 100 = 100% (normal spawn), 50 = 50% (half spawn time) and so on
+-- the higher the number, the spawn will be less frequent
```
DSpeichert, line 111:
```suggestion
-- the higher the number, the less frequent spawn will be
```

config.lua.dist
```diff
@@ -98,15 +98,21 @@ serverSaveShutdown = true
...
+-- NOTE: minimum spawn interval is 10 seconds while maximum is 1 day for the map spawn file
```
DSpeichert, line 112: User is setting the intervalRate in % but the limits are in seconds/days (time). This seems confusing and if the resulting interval falls out of bounds, the monster will not spawn at all.
I think that most users will not know all the intervals in `spawns.xml` and a better default behavior would be to cap the interval and spawn the monster with the lowest/highest permissible interval instead to avoid confusion with the monster simply not spawning at all.

peonso: What you would suggest, adding or subtracting absolute values?

I would suggest to just ditch this feature, doesn't make much sense, nor has much use. It's just to compensate the fact people thought they had it when they actually didn't, but never truly missed it anyway. Since it's blocking 1.4 I would only add the ```-- rateSpawn limits the number of monsters that will spawn within the same spawn object``` line to config.lua and scrap the idea.

Codinablack:
All checks have passed, only waiting for one review, yet this is still stalled... can we get a reviewer?

Znote:
> All checks have passed, only waiting for one review, yet this is still stalled... can we get a reviewer?

Would be great if you could help review this. :)

config.lua.dist
Codinablack: Using this 100 == 1 and 200 == 2, so it would be extremely difficult to go outside of bounds here. Also, as the current implementation goes, there is nothing stopping a player from being able to "go outside of bounds" in the spawns.xml anyways, pretty much same thing, just relocated and / 100. I see no reason this should stall 1.4

Codinablack (dismissed):
I believe everything here is in order, yes this does take the spawns and multiplies it by the number in the configuration, this allows for increasing/decreasing rates of all spawns simultaneously, there isn't any additional checks to make sure that the resulting "time" is within certain "bounds", but as far as I can tell the original code doesn't check for this either. Everything works, and no merge conflicts, I see no reason for this to be stalling 1.4, and so I approve.

DSpeichert:
@EPuncker needs rebase

config.lua.dist
```diff
@@ -141,9 +142,14 @@ rateSpawn = 1
 -- despawnRange is the amount of floors a monster can be from its spawn position
 -- despawnRadius is how many tiles away it can be from its spawn position
 -- removeOnDespawn will remove the monster if true or teleport it back to its spawn position if false
+-- spawnIntervalRate controls the rate of monster spawning interval
+-- its in percentage so 100 = 100% (normal spawn), 50 = 50% (half spawn time) and so on
```
DSpeichert, line 146:
```suggestion
-- it's in percentage so 100 = 100% (normal spawn), 50 = 50% (half spawn time) and so on
```

floki21uu:
> I believe everything here is in order, yes this does take the spawns and multiplies it by the number in the configuration, this allows for increasing/decreasing rates of all spawns simultaneously, there isn't any additional checks to make sure that the resulting "time" is within certain "bounds", but as far as I can tell the original code doesn't check for this either. Everything works, and no merge conflicts, I see no reason for this to be stalling 1.4, and so I approve.

"/Reload config" doesnt reload set spawn interval in config.lua

PR #3137, opened 2020-07-20
Many words in talkactions

rookgaard:
Hello.

I would like to introduce another thing that is missing compared to 0.X and it was very useful in my opinion - many words in one talkaction entry (that's why it's called `words` not `word`). The simpliest example - you can have 2 entries in talkactions.xml to display players online - by typing `!online` and `/online`, but with this code you can combine them into one entry - `!online;/online`.

@DSpeichert  I hope this PR won't receive comment `This is not a feature that provides anything you can't already do.` - because if something can be done easier, why not have it?

Duplicate of #3132


joseluis2g:
I never tried this but wasnt this already possible? if not im sorry and great job

korejwo:
@joseluis2g No, because engine register it as string https://github.com/otland/forgottenserver/blob/master/src/talkaction.cpp#L71  without exploding by delimiter https://github.com/otland/forgottenserver/blob/master/src/talkaction.cpp#L137.

rookgaard:
@joseluis2g No, because engine register it as string https://github.com/otland/forgottenserver/blob/master/src/talkaction.cpp#L71  without exploding by delimiter https://github.com/otland/forgottenserver/blob/master/src/talkaction.cpp#L137.

kygov:
Why not just use `playerSaySpell` and iterate through words map? In this way its creating new Talkaction object for every words, or I'm wrong?

rookgaard:
@kygov Talkaction object is created once (https://github.com/otland/forgottenserver/blob/master/src/talkaction.cpp#L77, https://github.com/rookgaard/forgottenserver/blob/feature/talkactions-words/src/talkaction.cpp#L82) and then just it's address is copied for every word. It's made the same way as actions where object is created once (https://github.com/otland/forgottenserver/blob/master/src/actions.cpp#L197) and then addres copied for every ID (https://github.com/otland/forgottenserver/blob/master/src/actions.cpp#L201).

gpedro:
what's the difference between #3132 and yours?

EPuncker:
> what's the difference between #3132 and yours?

one for xml and other for revscript

MillhioreBT:
> > what's the difference between #3132 and yours?
> 
> one for xml and other for revscript

My RP now supports XML too, if you want to give it a try and give me recommendations, here I am ;)

rookgaard:
My one also support revscripts too from now and requires less code changes :)

@DSpeichert what means `decisions` label? It means PR waits for review?

MillhioreBT:
> My one also support revscripts too from now and requires less code changes :)
> 
> @DSpeichert what means `decisions` label? It means PR waits for review?

Hello Goodnight!
I will not say which modification is better or more efficient, the decision will be made by all together!
What I did not like was that you put that my PR was duplicate, when in reality the duplicate is yours, just look at the date and the PR number

```lua
print(3132 < 3137)
-- true
```

regards!!

Znote (approved):
Tested, both XML and Revscriptsys works great. :) 

image

PR #3143, opened 2020-07-30
Make combat for "no vocation" configurable

kairion:
Removed outdated check that disables combat with no vocation characters.

There are enough better ways to deal with it today like using non-PVP tiles on map or even using protectionLevel option on config.lua.

### DIFF
--- a/data/events/scripts/creature.lua
+++ b/data/events/scripts/creature.lua
@@ -7,6 +7,20 @@ function Creature:onAreaCombat(tile, isAggressive)
 end
 
 function Creature:onTargetCombat(target)
+	-- No Vocation Combat Config --
+	disableNoVocationCombat = true
+	-- No Vocation Combat Config --
+	
+	if disableNoVocationCombat then
+		if self:isPlayer() and target:isPlayer() then
+			if self:getVocation():getId() == VOCATION_NONE or target:getVocation():getId() == VOCATION_NONE then
+				if not self:getGroup():getAccess() then
+					return RETURNVALUE_YOUMAYNOTATTACKTHISPLAYER
+				end
+			end
+		end
+	end
+	
 	return RETURNVALUE_NOERROR
 end
 
--- a/data/events/events.xml
+++ b/data/events/events.xml
@@ -3,7 +3,7 @@
 	<!-- Creature methods -->
 	<event class="Creature" method="onChangeOutfit" enabled="0" />
 	<event class="Creature" method="onAreaCombat" enabled="0" />
-	<event class="Creature" method="onTargetCombat" enabled="0" />
+	<event class="Creature" method="onTargetCombat" enabled="1" />
 	<event class="Creature" method="onHear" enabled="0" />
 
 	<!-- Party methods -->
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -273,10 +273,6 @@ bool Combat::isProtected(const Player* attacker, const Player* target)
 		return true;
 	}
 
-	if (attacker->getVocationId() == VOCATION_NONE || target->getVocationId() == VOCATION_NONE) {
-		return true;
-	}
-
 	if (attacker->getSkull() == SKULL_BLACK && attacker->getSkullClient(target) == SKULL_NONE) {
 		return true;
 	}


kairion:
> you need to add the alternative then because right now you are just allowing combat between no vocations... and in MY opinion a 3 line check is better than having to having to no-pvp brush the entire rookgaard map (for example)
> 
> perhaps move the check to the onTargetCombat event

Why would I have to if the source already has it?
On `config.lua`:
```lua
protectionLevel = 1
```

This option makes it possible to limit combat between lower levels (or any given level) and even allow Rookers to fight on Rookgaard (there are servers who allow it).

Having this hard-coded on the source breaks player flag `CannotAttackPlayer` as pointed by user Demnish on OTLand [here](https://otland.net/threads/bug-report-cannotattackplayer-flag-is-bugged.268705/).

Also, as far as I could get from TFS forum and answers on issues here, TFS is meant to be a base-source (not a data pack) for any kind of server, not strictly to RL map servers, so there's no point in imposing on source a configuration related to map (as in: server owner decides to have a starter map where players can't fight each others) or even related to level (as in: server owner decides to do not allow combat between players with X level).

kairion:
This code is a legacy from when non-PVP tiles or protectionLevel doesn't exist, it was meant to be used as a way to prevent PVP on Rookgaard, so since it's not needed anymore I honestly don't see a reason to keep it. If there's a valid reason to keep it as it is, I'd be glad if someone could enlighten me.

Having this code on source, if you could check the post I linked from Demnish on OTLand, could break `CannotAttackPlayer` grou flag.

And finally: even if, for some unknown reasons, someone really wants to limit a vocation combat, it can be achieved with creaturescripts and being way more configurable than hard-coded on source.

DSpeichert:
I agree with your arguments, just not the way you introduce the change. If someone updates the engine to a version with your change, their behavior will change unexpectedly. They wouldn't know that now they must edit the map, create a creaturescript or whatever else you point out as a possible remedy for this implicit behavior change. We keep things backwards compatible.

If you provide at least a creaturescript that will retain this behavior and also make it configurable (because you can disable a creaturescript), then this change can be merged.

kairion:
Oh sure, I can do that. I just thought we were not allowed to add to TFS these kind of scripts as they seem to be more suitable on specific distributions (like OTX, OTServBR-Global etc.).

I'll make a configurable creaturescript for it and send a new commit to this PR as soon as I got time.

DSpeichert:
There no such things that can't be added, especially when things are configurable.

kairion:
Ok, I did it with events, but since Lua isn't exactly my cup of tea I really advise whoever will check it to do it very carefully and any kind of feedback will be appreciated.

What the new commit does?
1. Enable `onTargetCombat` on `Creature` class from \data\events\events.xml
2. Adds a check on `\data\events\scripts\creature.lua` to see if attacker (`self`) or `target` do not have a vocation and if that's the case forbiddens combat.
3. Adds a check for group access so we don't break CannotAttackPlayer (if it's set to 0) if a GM/God is a no vocation
4. Adds a boolean to enable/disable this block of code.

Why did I add a boolean if events already have an `enabled` option?
I did it because I thought some people may want to enable this event for other purposes, so this way they can add extra code for their own purposes and just set this boolean to false if they want to allow combat between players with no vocation.

data/events/scripts/creature.lua
```diff
@@ -7,6 +7,20 @@ function Creature:onAreaCombat(tile, isAggressive)
 end
 
 function Creature:onTargetCombat(target)
+	-- No Vocation Combat Config --
+	disableNoVocationCombat = true
```
nekiro, line 11: I don't think that's valid place to put this global variable in. Try moving to to config.lua and rename it to noVocationPvp = false, by default it should be false cause it was like that since ages.

data/events/scripts/creature.lua
```diff
@@ -7,6 +7,20 @@ function Creature:onAreaCombat(tile, isAggressive)
...
+	-- No Vocation Combat Config --
+	
+	if disableNoVocationCombat then
```
nekiro, line 14: if not noVocationPvp then

nekiro (changes requested):

data/events/scripts/creature.lua
peonso: I like it within the script. It would just clutter config.lua, I don't think it is desireable that every possible customization is added to config.lua.

DSpeichert (approved):

DSpeichert:
@kairion needs conflict resolution

PR #3144, opened 2020-07-30
Removed unused and outdated player flag and talktype

### LINKED ISSUE #3142: Group flag CanTalkRedChannelAnonymous doesn't work
1. Download TFS stable (1.2) or TFS 1.3 (latest commit), it's up to you;
2. Check if "cantalkredchannelanonymous" on \data\XML\groups.xml is enabled (GM and God come with it enabled by default);
3. Try to talk in a chat channel using the talking syntax `#d` like that: `#d testing talking red anonymous`;
4. See that nothing happens.

### Expected behaviour
The message should be sent in red without GM/God name.

### Actual behaviour
The message will be sent in orange (default behavior for all messages from GM/God groups) with character name and level.

kairion:
Removed unused player flag `CanTalkRedChannelAnonymous` and talktype `TALKTYPE_CHANNEL_R2` since both of them are deprecated.

It should close #3142.


kairion:
@EPuncker you're totally right, I changed on my local files but forgot to change on my PR.
Done.

src/const.h
```diff
@@ -542,7 +540,6 @@ enum PlayerFlags : uint64_t {
 	PlayerFlag_SetMaxSpeed = 1 << 29,
 	PlayerFlag_SpecialVIP = 1 << 30,
 	PlayerFlag_NotGenerateLoot = static_cast<uint64_t>(1) << 31,
-	PlayerFlag_CanTalkRedChannelAnonymous = static_cast<uint64_t>(1) << 32,
```
DSpeichert, line 545: Good place to leave a comment here that such a thing existed at `<< 32` and was removed.

kairion: I agree. I've sent a new commit adding a comment about it on a new line right after `PlayerFlag_NotGenerateLoot = static_cast<uint64_t>(1) << 31,`.

DSpeichert: I think the proper word would be "exponent 32", not "exponential" (adjective)

kairion: Oops, my bad. Fixed.

PR #3145, opened 2020-07-30
fix paralyze rune correct

Johncorex:
will work correctly as needed


EPuncker:
finally got to test this!

while it seems fine with lower level players, for example changing the speed of a level 189 from 318 to 60

but at higher level it doesn't seem to be okay, it changed the speed of a level 565 from 674 to only 416

EPuncker:
closing this as no information/source was given at all and no feedback

PR #3151, opened 2020-08-04
Added debuff removal to swimming and achievements

EPuncker:
Since update 8.1 when swimming you will get all your negative conditions removed.

https://tibia.fandom.com/wiki/Swimming

- Added Waverider, Snowbunny and Deep Sea Diver achievements
- Removed bubbles effect


EPuncker:
@DSpeichert I tested and its working, just forgot to take screenies

PR #3152, opened 2020-08-05
Removed protocolold

EPuncker:
its not used at all


EPuncker:
don't hate me guys

DSpeichert:
Well, it is used to tell old clients that they are old instead of not being able to decipher their messages on the wire.

If the current protocol gets upgraded from 11.98 to future versions and login method changes, we'd use similar concept for current protocol.

Zbizu:
and this is a reason good enough to keep it

DSpeichert:
Nothing to gain here.

PR #3154, opened 2020-08-06
Add some sanity check to spawns

### LINKED ISSUE #3133: Function Spawns::loadFromXml is lacking a few "sanity" checks
Currently our Spawns::loadFromXml is missing a few checks and warnings, since most of the available datapacks contain lots of wrong values and so on, but since we have no warnings, users never know what they are doing wrong, here are just a few examples of bad values:

- [x] warn user if monster spawn is set inside a PZ tile (currently it just doesn't spawn it, but user won't ever know)
- [x] negative spawntime `<monster name="Killer Caiman" x="-1" y="0" z="7" spawntime="-1488876288" />`
- [x] absurd spawntime values `<monster name="Lancer Beetle" x="0" y="1" z="7" spawntime="104185600" />`
- [x] spawn radius bigger than 30 `<spawn centerx="529" centery="294" centerz="8" radius="100">` (RME max is 30)
- [x] we should consider rising our MINSPAWN_INTERVAL to 10 seconds instead of 1, to match RME
- [x] should we add a max spawn interval as well (at least a warning, but still allow it)? (RME caps it to 3600 seconds)
- [x] empty spawns (this one is optional but wouldn't hurt, there are maps that have a few, but there are maps that have HUNDREDS) ([image](https://i.imgur.com/9OKmG15.png))

They may have been caused by bad map editors or even lack of knowledge from the user, should we blame them? **no**.
Should we warn them? **yes**, that is why I'm opening this issue

EPuncker:
- Inform owner when monsters can't be spawned at startup (for example: monsters in PZ, or inside walls)
- Inform owner about negative spawntimes
- Inform owner about spawn radius bigger than 30
- Inform owner about empty spawns on map
- Raised our MINSPAWN_INTERVAL to 10 seconds to match RME
- Add maximum spawn interval to inform owner about absurd spawntimes

code by @SaiyansKing at https://github.com/SaiyansKing/optimized_forgottenserver/commit/51be9cf60552f6084c1f2a7f82d670ee4243da54
fixes #3133


EPuncker:
tested and working

src/spawn.cpp
```diff
@@ -33,7 +33,8 @@ extern Monsters g_monsters;
 extern Game g_game;
 extern Events* g_events;
 
-static constexpr int32_t MINSPAWN_INTERVAL = 1000;
+static constexpr int32_t MINSPAWN_INTERVAL = 10000; // 10 seconds to match RME
+static constexpr int32_t MAXSPAWN_INTERVAL = 86400000; // 1 day
```
ranisalt, line 37: What about using `std::chrono::seconds` here?

src/spawn.cpp
```diff
@@ -90,11 +100,15 @@ bool Spawns::loadFromXml(const std::string& filename)
 					centerPos.y + pugi::cast<uint16_t>(childNode.attribute("y").value()),
 					centerPos.z
 				);
-				uint32_t interval = pugi::cast<uint32_t>(childNode.attribute("spawntime").value()) * 1000;
-				if (interval > MINSPAWN_INTERVAL) {
-					spawn.addMonster(nameAttribute.as_string(), pos, dir, interval);
+				int64_t interval = pugi::cast<int64_t>(childNode.attribute("spawntime").value()) * 1000;
```
ranisalt, line 103: Inconsistent variable types. Either make everything int32_t, or int64_t (preferably the former... I don't think anyone will ever make a spawn with interval over 136 years)

src/spawn.cpp
Znote: Can you elaborate?
found some interesting docs though:
https://en.cppreference.com/w/cpp/chrono/duration
chrono durations might be a better way generally to deal interval and timed stuff

Do you want the type declaration to be changed to std::chrono::seconds? Or do you want this entire thing to be chrono-fyed?

ranisalt: I suggest the whole thing to use chrono::seconds.

src/spawn.cpp
```diff
@@ -33,7 +33,8 @@ extern Monsters g_monsters;
 extern Game g_game;
 extern Events* g_events;
 
-static constexpr int32_t MINSPAWN_INTERVAL = 1000;
+static constexpr int32_t MINSPAWN_INTERVAL = 10000; // 10 seconds to match RME
```
nekiro, line 36: what if we used here 10 * 1000 and 24 * 60 * 60 * 1000 for better readability?

nekiro (approved):
LGTM

PR #3157, opened 2020-08-06
Added float support to config.lua

EPuncker:
code by @infernumx


src/configmanager.cpp
```diff
@@ -188,6 +201,8 @@ bool ConfigManager::load()
 	integer[SERVER_SAVE_NOTIFY_DURATION] = getGlobalNumber(L, "serverSaveNotifyDuration", 5);
 	integer[YELL_MINIMUM_LEVEL] = getGlobalNumber(L, "yellMinimumLevel", 2);
 
+	floating[SHARED_EXPERIENCE_MULTIPLIER] = getGlobalFloat(L, "sharedExperienceMultiplier", 1.20);
```
Shawak, line 204: 1.2f

EPuncker: thanks!

DSpeichert (dismissed):

peonso:
While I like it a lot, I'm bothered by consistency, why exp, loot and skill rates aren't floats either? They should and this change might lead people to believe they are.

danyelvarejao (dismissed):

nekiro:
> While I like it a lot, I'm bothered by consistency, why exp, loot and skill rates aren't floats either? They should and this change might lead people to believe they are.

That's great observation. We can update them later in next prs. This one is fine as it is, to me at least.

nekiro (dismissed):

DSpeichert (approved):

DSpeichert:
I like Float support but I'm not sure if we want to move everything to config.lua.

EPuncker:
> I like Float support but I'm not sure if we want to move everything to config.lua.

that entry was added just as example, should I remove it?

DSpeichert:
> that entry was added just as example, should I remove it?

Hard to say, it's a nice representation of what can be done now, but it could be unnecessary to move things to config.lua and also confusing if other values can't be floats. Maybe remove it and we keep it as a "hidden feature" for now?

EPuncker:
> > that entry was added just as example, should I remove it?
> 
> Hard to say, it's a nice representation of what can be done now, but it could be unnecessary to move things to config.lua and also confusing if other values can't be floats. Maybe remove it and we keep it as a "hidden feature" for now?

done

DSpeichert (approved):

src/configmanager.h
```diff
@@ -137,12 +137,17 @@ class ConfigManager
 			LAST_INTEGER_CONFIG /* this must be the last one */
 		};
 
+		enum floating_config_t {
+			LAST_FLOATING_CONFIG
```
nekiro, line 141: arrays cant be zero sized, you have to mark this 1 or add at least one config enum

nekiro (changes requested):

EPuncker:
what about we add it when it is useful at all? :D

PR #3160, opened 2020-08-12
New container method container:getItems(recursive)

nekiro:
New method which will return all container items, optional bool recursive to search in sub containers of container.

Usage:
```
local items = container:getItems(true)
for _, item in ipairs(items) do
	print(item:getName())
end
```

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -634,7 +634,7 @@ std::map<uint32_t, uint32_t>& Container::getAllItemTypeCount(std::map<uint32_t, 
 	return countMap;
 }
 
-std::vector<Item*> Container::getItems(bool recursive /*= false*/)
+ItemVector Container::getItems(bool recursive /*= false*/)
 {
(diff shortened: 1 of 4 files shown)


MillhioreBT:
Hello good afternoon
I don't want to question your changes, I just wonder if that code is more efficient than this:

```lua
function Container:getItems(recursive, array)
	local array = array or {}
	for slot = 0, self:getCapacity() - 1 do
		local item = self:getItem(slot)
		if recursive and item and item:isContainer() then
			array = item:getItems(recursive, array)
		elseif item then
			array[#array +1] = item
		end
	end
	return array
end
```
If efficiency shouldn't be a big issue, why not implement this function in lua in the file `lib/core/container.lua`

If I'm confused about something, please ignore my question, it's just curious!

EPuncker (approved):

DSpeichert (approved):

PR #3161, opened 2020-08-14
Monsters updates

EPuncker:
- Added a few missing bosses and creatures (credit: my datapack, ORTS and otland forum)
- Update a few monsters outfit to match rl (based on staticdata.xml) (because why not)
- Some random fixes here and there
- Removed default lightcolor/lightlevel
- Fixed a lot of monsters with runonhealth="1", with correct value
- Added look type of wild magics monsters

nothing tested in game by me, but fixed all console errors and these files have been around for quite long time


EPuncker:
@DSpeichert this is ready!

DSpeichert (changes requested):
One more request from me: let's not use `(`/`)` in file names.

Like `amarie_(creature).xml` - we know it's a create, not NPC. It's in monsters/ folder after all. There are a bunch of those.

Next, it will be up to @Znote to merge that.

EPuncker:
> One more request from me: let's not use `(`/`)` in file names.
> 
> Like `amarie_(creature).xml` - we know it's a create, not NPC. It's in monsters/ folder after all. There are a bunch of those.
> 
> Next, it will be up to @Znote to merge that.

it is like that to help differentiate the spawn file entry, I have see very often RME changing between NPC/monster (azerus, pythus the rotten etc) because the npc file and monster file had the same name and thus breaking the game part, but okay I've removed it.

> There are a bunch of those.

there were only 2 😛

DSpeichert (approved):

EPuncker:
I need this merged so I can work on a new PR in order to fix the errors from #3239 😛  @Znote @DSpeichert

Issue #3148, opened 2020-08-01
Magic Wall and Wild Growth missing changes


EPuncker:
- [x] magic wall should have a random duration between 14 and 20 seconds when shot
- [x] wild growth should have a random duration between 38 and 45 seconds when shot
- [x] ITEM_WILDGROWTH_NOPVP and ITEM_MAGICWALL_NOPVP should be destroyed on stepIn by any player (only on NOPVP world types or NOPVP tiles)

[since 10.01 update](https://tibia.fandom.com/wiki/Updates/10.01)

soul4soul:
I have never heard of the random duration nor is it documented on tibiawiki. Do you have any youtube videos from RL tibia which show the random time? I have no characters that can perform these tests so I can't check myself.

EPuncker:
> I have never heard of the random duration nor is it documented on tibiawiki. Do you have any youtube videos from RL tibia which show the random time? I have no characters that can perform these tests so I can't check myself.

https://www.tibia.com/forum/?action=thread&postid=35059085#post35059085
https://www.tibiaqa.com/11236/are-the-duration-the-magic-wall-and-the-wild-growth-random-there-factors

PR #3168, opened 2020-08-19
new config for attribute "duration" ( MagicWall - Wild Growth duration random )

MillhioreBT:
With these changes, it is now possible to configure a range of durations for temporary articles, thus complying with what is mentioned in the PR #3148

> - [x] magic wall should have a random duration between 14 and 20 seconds when shot
> - [x]  wild growth should have a random duration between 38 and 45 seconds when shot


EPuncker:
maybe add a warning if the first value is bigger than the second one

MillhioreBT:
> maybe add a warning if the first value is bigger than the second one

`random_number` function reverses the order from least to greatest always, so it's not a problem, the warning would only be something for messy people, and in the base branch we will never include messy things, or at least that's what I think, if what I say not true let me know, i am prone to errors :(

data/items/items.xml
```diff
@@ -686,15 +686,15 @@
 	<item id="1497" article="a" name="magic wall">
 		<attribute key="type" value="magicfield" />
 		<attribute key="decayTo" value="0" />
-		<attribute key="duration" value="20" />
+		<attribute key="duration" value="14;20" />
```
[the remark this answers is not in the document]
MillhioreBT: <3

src/items.cpp
```diff
@@ -819,8 +821,29 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
...
+					auto split = explodeString(valueAttribute.as_string(), ";");
+					if (split.size() == 2) {
+						it.decayTime = pugi::cast<uint32_t>(split[0].c_str());
+						uint32_t value = pugi::cast<uint32_t>(split[1].c_str());
+						if (it.decayTimeMax) {
+							std::cout << "[Warning - Items::parseItemNode] decayTimeMax is being overwritten to: " << value << std::endl;
+						}
+						it.decayTimeMax = value;
+					}
+					else {
```
nekiro, line 835: this whole code is not needed as we are not using ; serparator for min and max
just
it.decayTime = pugi::cast<uint32_t>(valueAttribute.value());

nekiro (changes requested):

src/items.cpp
MillhioreBT: I think you hate the separator ahahahaha, but since you're the only one paying attention to me, I'm going to listen to you <3

lucasromagnoli (dismissed):

nekiro:
Hey @MillhioreBT care to resolve requested changes?

src/item.h
```diff
@@ -967,6 +967,9 @@ class Item : virtual public Thing
 			}
 		}
 		uint32_t getDefaultDuration() const {
+			if (items[id].decayTimeMax) {
```
nekiro, line 970: This method gonna return random values now, how we gonna get durationmin and durationmax in this case? I think getDefaultDuration shouldn't be changed and getDefaultDurationMax should be added.

nekiro (changes requested):
Also I think you missed luascript changes, we have no possibility to getDurationMax from itemtype now.
Edit: ah we didn't have them before too, but wont hurt to add these.

MillhioreBT:
I have a problem and inadvertently i delete the repositories that had the changes, now that I want to complete the suggestions, it does not let me update this PR, I suppose I will have to open a totally new one

Znote:
> I have a problem and inadvertently i delete the repositories that had the changes, now that I want to complete the suggestions, it does not let me update this PR, I suppose I will have to open a totally new one

If you need to recover this, I think you can pull/checkout otland/0178a3c7a3678c4b6b594db651c23bdd1943b7fc

PR #3173, opened 2020-09-10
Item abilities updated

penerka:
testing


EPuncker:
what is this PR 😛

MillhioreBT:
First of all, there is no explanation of what it is and why they should include it
I know my opinion is worth nothing, but I think this should not be included since it is outside the classic concept of what the Tibia RL is like, this would be rather a deep skill system for modified servers
If there is a good reason for this, I would like to know since I probably do not know how Tibia RL is working right now

EPuncker:
> First of all, there is no explanation of what it is and why they should include it
> I know my opinion is worth nothing, but I think this should not be included since it is outside the classic concept of what the Tibia RL is like, this would be rather a deep skill system for modified servers
> If there is a good reason for this, I would like to know since I probably do not know how Tibia RL is working right now

I think he just made a PR by mistake while trying to implement this on his TFS, I'm closing this just in case, since he didn't replied me!

PR #3176, opened 2020-10-01
Optimize IOLoginData::saveItems by reducing number of allocations

marksamman:
Note that this is untested, just an idea I got by quickly looking at the function. std::list would allocate for each push, std::vector tends to have a growth strategy where it only allocates when it reaches capacity

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -591,7 +591,8 @@ bool IOLoginData::saveItems(const Player* player, const ItemBlockList& itemList,
 	std::ostringstream ss;
 
 	using ContainerBlock = std::pair<Container*, int32_t>;
-	std::list<ContainerBlock> queue;
+	std::vector<ContainerBlock> containers;
+	containers.reserve(150);
 
 	int32_t runningId = 100;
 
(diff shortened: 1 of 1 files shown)


src/iologindata.cpp
```diff
@@ -613,22 +614,21 @@ bool IOLoginData::saveItems(const Player* player, const ItemBlockList& itemList,
...
 		if (Container* container = item->getContainer()) {
-			queue.emplace_back(container, runningId);
+			containers.emplace_back(container, runningId);
 		}
 	}
 
-	while (!queue.empty()) {
-		const ContainerBlock& cb = queue.front();
+	for (size_t i = 0; i < containers.size(); i++) {
+		const ContainerBlock& cb = containers[i];
```
djarek, line 622:
```suggestion
	for (const ContainerBlock& cb : containers) {
```

marksamman: That wouldn't work because we append to the containers vector in this loop.

nekiro:
This is interesting. I saw a lot of places in the source where we use std::list or pop_front, if I understand correctly if you pop_front all items in list have to be relocated to index - 1, which does not sound like a good performance thing, so this code should be faster right?

marksamman:
pop_front on std::list is fine, a similar operation on std::vector would
have to relocate all items. This is faster because std::list allocates
memory for each push while std::vector allocates in bulk. Iterating over a
std::vector and accessing the items is also faster than over a std::list
because the items will be in CPU cache, but most of the gain in this PR
should be from the reduced amount of memory allocation calls

ranisalt:
I think the biggest gains are moving to a container with contiguous allocation, it's just faster to access. The STL implementation of vector doubles in size when it hits the capacity, so it only avoids 5 small reallocations at best.

Another good point is that a list is just wastes a lot of memory with pointers.

This might be the perfect use case of a [`std::deque`](https://en.cppreference.com/w/cpp/container/deque), which is a compromise between a list and a vector.

>  if I understand correctly if you pop_front, all items in list have to be relocated to index - 1

Wrong, a `std::list` is a linked list and insertions in any position are constant (provided that you have the iterator, and you always have for front and back).

DSpeichert (approved):
Looks like this started an avalanche of converting things to use vectors, so we should probably revisit this. LGTM, since there are no better proposals.

PR #3177, opened 2020-10-01
Fix #2941

### LINKED ISSUE #2941: getPluralName function missing a few checks
I stumbled into a double grapes entry when doing a change in items.xml file and noticed that our getPluralName code:

https://github.com/otland/forgottenserver/blob/b7ef46dfc7ac4fad1bcea5e3604c9ef8c4350864/src/items.h#L275-L289

is kinda "incomplete" and missing a few checks, it only adds "s" to the end of the word if it doesn't have a plural tag, that is why we need to have:

`<item id="2681" name="grapes" plural="grapes">`

or it would show "**grapess**" in game.

my suggestion is to add a check to see if the word ends in "s" it shouldn't add anything, and also if possible (I don't really know english gramar rules since its not my main language), but maybe we could also add other rules... in fact I've just researched and noticed that there is a JS lib for it, that we could borrow at least a few rules:

https://github.com/blakeembrey/pluralize/blob/master/pluralize.js#L329-L356

or maybe I'm being too deep into a simple issue? xDDD

marksamman:


src/items.h
```diff
@@ -283,6 +283,10 @@ class ItemType
 				return name;
 			}
 
+			if (name.empty() || name.back() == 's') {
```
EvilHero90, line 286: Wouldn't it make more sense if the console would be showing information that an item has no name set but a plural name set, other than that it wouldn't make much sense to check for ```name.empty()``` or am I misstaken?

marksamman: You can't call name.back() on an empty string, so empty string case must be handled and I think returning an empty string for plural name makes more sense than returning "s" which is the current behavior.

marksamman: I'm not opposed to logging if there is an item with an empty name though, not sure how common that is and if it would be of use.

DSpeichert (approved):

PR #3178, opened 2020-10-01
Converting talkactions to revscript

omarcopires:


DSpeichert:
I'm against converting existing systems to talkactions just for the sake of doing it. XML support cannot be dropped anyway, so all this is going to do is cause confusion for people who are more used to XML-style talkactions or merging such datapacks.

RafaelVidaurre:
I'm in favor, revscript makes way more sense as a system overall. Especially since it fixes the issue with the XML approach where logic for a particular feature is spread across multiple different files in different parts of the project. It also allows for way better portability which enables people to share functionality more easily.

I'm not sure what the roadmap is, but I'd be definitely in favor of moving everything there. I do partially agree with @DSpeichert  in the sense that we either move everything or nothing, otherwise yeah, that's pretty confusing.

I'm not in favor of backwards compatibility in general though, I projects degrade over time if they take backwards compatibility into account. A better solution is to have migration guidelines. This is assigned to the 2.0 Milestone, which even though we are not using Semver still sounds like the best moment version-wise to perform huge breaking changes

PR #2975, opened 2020-04-26
Convert all monsters to rev script system

soul4soul:
A long time ago I was working on a program to converter monsters from TFS XML format to pyOT format. Last fall I updated that program to support the TFS revscriptsys monster format. I finally got around to adding a UI and command line interface to the program. Most of the logic is completed for converting from TFS XML format to TFS revscriptsys format. The program is open source and can be found at https://github.com/soul4soul/ot-monster-converter.

For anyone curious to try it themselves they can find binaries hosted on the bitbucket repo. Be aware the only working input format at this time is "TFS XML Format". As for output formats both "TFS Rev Script System Format" and "PyOT" are working. The executable is a whopping 82MB because I made it a self contained exe by packing in the entire .NET core 3.1 SDK. I made that decision since it's not yet common for non devs to have .NET Core SDKs installed on their machines.

Screenshot of the application running. The entire conversion process takes 20 seconds or so.
image

There are still a few issues to work out but I opened this PR so the community would be aware of this work.

- [ ] Discuss with community changing revscriptsys monsters loot drop chance to be a percentage from 0 to 1. Where 1 means 100% drop rate.
- [ ] Discuss with community changing element resistance to a decimal percentage between 0 and "1". 1 = 100% damage no immunity. 0.85 = 15% immunity. 1.15 = 15% weakness.
- [ ] Discuss with community replacing the monsters attacks and defenses for revscriptsys  with the player spell system. Due to the revscriptsys, monsters and spells can be defined in the same file. All creature spells would be defined with the spell system in the monster file and registered directly to the creature. The player spell system is already more powerful then the creature attacks and defenses. 

With this program the process of updating to rev script is seamless and easy for everyone. We should be able to remove the entire XML monster system in a future version of TFS.

PR #3179, opened 2020-10-01
Converting globalevents to revscript

omarcopires:


EvilHero90:
There was a quite lengthy discussion about this already, there is no plan to convert classic scripts to revscriptsys before the release of 1.4, same with #3178 #2653 #2975

PR #3180, opened 2020-10-06
Moved check for leech

Mrpox:
**Move the check for mana and life leech to make it not possible to leech life when just casting spells in the air without any target.
**Added new bool to check if already leeched for areacombat so it only leeches once when hitting multiple targets.

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -860,7 +860,7 @@ void Combat::doAreaCombat(Creature* caster, const Position& position, const Area
 	g_game.map.getSpectators(spectators, position, true, true, rangeX, rangeX, rangeY, rangeY);
 
 	postCombatEffects(caster, position, params);
-
+	CombatDamage leechAmount; //Setup new CombatDamage to keep highest damage
 	for (Tile* tile : tileList) {
 		if (canDoCombat(caster, tile, params.aggressive) != RETURNVALUE_NOERROR) {
 			continue;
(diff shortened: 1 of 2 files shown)


src/combat.cpp
```diff
@@ -776,7 +775,9 @@ void Combat::doTargetCombat(Creature* caster, Creature* target, CombatDamage& da
 		if (g_game.combatBlockHit(damage, caster, target, params.blockedByShield, params.blockedByArmor, params.itemId != 0)) {
 			return;
 		}
-
+		if (casterPlayer) {
```
nekiro, line 778: why did you move the code here?

src/combat.cpp
```diff
@@ -905,6 +905,10 @@ void Combat::doAreaCombat(Creature* caster, const Position& position, const Area
 						if (g_game.combatBlockHit(damageCopy, caster, creature, params.blockedByShield, params.blockedByArmor, params.itemId != 0)) {
 							continue;
 						}
+						if (casterPlayer && !damage.leeched) {
```
nekiro, line 908: same, why move there?
I agree with that bool though.

src/combat.cpp
Mrpox: I moved it here to be after block check.

src/combat.cpp
Mrpox: You're completely right, should check for highest damage and then leech.

src/combat.cpp
Mrpox: > why did you move the code here I think you can just check the target variable if it leechs even on no target?

I moved it here to be after block check.

src/combat.cpp
Mrpox: > For area spells this is actually still wrong. Cip calculates the highest damage from all the creatures that were hit in area spell and then calculates leech amount out of it. Although it should fix leeching for every target.

This has been altered, it now stores the highest damage done in areaattack and leeches off that.

nekiro:
All the issues listen in this pr were fixed with https://github.com/otland/forgottenserver/pull/3267.

PR #3182, opened 2020-10-13
Fix crash related to map cleaning

TheSumm:
* Fixes a crash scenario and a potential crash scenario related to map cleaning code
* Simplifies a conditional expression


nekiro (approved):
LGTM

PR #3183, opened 2020-10-14
Removing unused corpses and tab adjustments

omarcopires:


EPuncker (approved):

nekiro:
I kinda don't know what are these corpses. We aren't using them anywhere?

EPuncker:
> I kinda don't know what are these corpses. We aren't using them anywhere?

nope, we don't, maybe they were used in the past by the animate dead rune, but since the implementation of the isCorpse, this table is not used anywhere else

nekiro (approved):

soul4soul:
I'd rather just keep it but won't be upset should this PR be merged. The corpse table is hurting anything and makes it easier to bring back animate dead for those who liked playing with that spell.

omarcopires:
> Prefiro ficar com ele, mas não ficarei chateado se esta relações públicas for mesclada. A mesa do cadáver está machucando qualquer coisa e torna mais fácil trazer de volta animado morto para aqueles que gostavam de brincar com esse feitiço.

The rune remains in the game and functioning as it should, except that there is now a function to check if it is a corpse.

PR #3184, opened 2020-10-14
Walkthrough everything Ctrl + Arrow keys

omarcopires:
Adding an old functionality present in tfs 0.4 that was later removed.
I did tests and it seems all right, I didn't find any problem when using it.

### DIFF
--- a/data/events/scripts/player.lua
+++ b/data/events/scripts/player.lua
@@ -200,6 +200,12 @@ function Player:onReportBug(message, position, category)
 end
 
 function Player:onTurn(direction)
+	if self:getGroup():getAccess() and self:getDirection() == direction then
+		local nextPosition = self:getPosition()
+		nextPosition:getNextPosition(direction)
+
+		self:teleportTo(nextPosition, true)
+    end
 	return true
 end
 
--- a/data/events/events.xml
+++ b/data/events/events.xml
@@ -23,7 +23,7 @@
 	<event class="Player" method="onMoveCreature" enabled="0" />
 	<event class="Player" method="onReportBug" enabled="1" />
 	<event class="Player" method="onReportRuleViolation" enabled="1" />
-	<event class="Player" method="onTurn" enabled="0" />
+	<event class="Player" method="onTurn" enabled="1" />
 	<event class="Player" method="onTradeRequest" enabled="0" />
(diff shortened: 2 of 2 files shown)


data/events/scripts/player.lua
```diff
@@ -200,6 +200,12 @@ function Player:onReportBug(message, position, category)
 end
 
 function Player:onTurn(direction)
+	if self:getGroup():getAccess() and self:getDirection() == direction then
```
lgrossi, line 203: That is not a good approach. You're embedding extra functionality in something that was just not designed to that.
"onTurn" is intended to make a turn not a walk. It comes from the client call that is also designed for turn, not walk. You're embedding logic and creating complexity in a function that was not designed to that, without the proper walking validations what could lead to issues. 
You should aim a solution that provides a proper benefit, not that just "pack" everything you think that is necessary in a single function because you want it to do tons of things for different scenarios, that is not functional and, definitely, not scalable. It's hell to give maintenance.

EPuncker:
doesn't this exceed packet per seconds limit? I think that is why we didn't brought it from 0.4

nekiro:
> doesn't this exceed packet per seconds limit? I think that is why we didn't brought it from 0.4

yup, spams packets like mad

MillhioreBT:
This should not be in the master branch, it is only optional for those who need it, an administrator does not need this functionality at all, it is just for the convenience of some.

nekiro:
> This should not be in the master branch, it is only optional for those who need it, an administrator does not need this functionality at all, it is just for the convenience of some.

I agree, kinda implementing functionality in event that isn't mandatory in any case. Plus it spams the server very bad, so it's really a bad practice to use it.

omarcopires:
> doesn't this exceed packet per seconds limit? I think that is why we didn't brought it from 0.4

You are right, when using "uncontrolled" the packet limit is exceeded.

Zbizu:
I agree with Nekiro and I propose to do that as onLook event. It could allow the user to get anywhere and wouldn't spam packets that much. Travelling in straight line is already covered by /a talkaction.

DSpeichert:
This method won't be merged but an alternative such as suggested point-click-teleport method can be proposed.

Issue #2156, opened 2017-02-05
Question about Store Inbox


brunominervino:
@ranisalt what you think about create new class called 'StoreInbox' for implement Store System? Suggested by @ninjalulz [here](https://github.com/otland/forgottenserver/pull/2143#discussion_r99316904) and supported by @Mkalo.

If they agree, is it best to do a new pull request?

ranisalt:
Very similar to #1641 on this issue. I don't know how the store system works, does it really have a separate inbox for bought items?

WibbenZ:
Yes its prety much the code from the purse except that you can only get items in it, you can't put the old premium scrolls in it.
So tbh #531 should be closed.

But... If you buy supplies let's say GMPs the server counts them as if you had them in your backpack, so it's a way to not lose supplies if you are hunting.
Not 100% on this but im pretty sure about it xD

brunominervino:
Basically how @WibbenZ told.

WibbenZ:
Agreed, some might use Znote / Gesior on the side of this and might want ppl to be able to store those items in there.
Would be alot easier to accept certain itemids insted of recompiling the source code.

brunominervino:
@WibbenZ exactly.

ninjalulz:
I'm slightly confused as to what you two are talking about. Not losing supplies? Accepting certain item ids?

WibbenZ:
Well in Tibia you get all the things you buy in the store inbox.
If you die you don't lose those items, so if you keep the GMPs you bought via the store in there you won't lose them if you die.

You can move bought items back, that should be OK with this system, but if you bought items via a website you can't put them in the store inbox, that is why it would be better to handle the store inbox in Lua so we can add an exception for those donation items not bought via the ingame store.

brunominervino:
@WibbenZ you can use, example GMPs, inside store inbox?

WibbenZ:
Aa I said not 100 on that but I think so.
Bubba did smth the store inbox when a new server started, but im not sure if he moved a couple to his bp or not.

Im sure you can't have pz when buying atleast since he had to stop hunting to buy more.

ninjalulz:
@WibbenZ Yes, I do understand that, however I believe you are forgetting one thing. The ingame store and AACs (e.g. Znote & Gesior) relies on Lua to add things to players, so it shouldn't be an issue. It's not like we must have this feature, do we? Having a locked store inbox, and only allowing items to be added with the no limit flag would be sufficient.

[...]

PR #3186, opened 2020-10-17
New store inbox implementation

### LINKED ISSUE #3185: player:addItem and player:addItemEx does not drop the item to the ground if there's no space

1. Download/compile/run the latest version of TFS without performing any changes and create a God
2. Drop your backpack, put a helmet on and say `/i dark helmet` (or try to spawn any other item)

### Expected behaviour
The item should drop to the ground when there is no **space**, note that without capacity this works properly

### Actual behaviour
The item is not dropped on the map

### Environment
Used Docker image provided in this repository

### Reproduction environment
Simply use the Docker image provided in this repository

## Notes

* This is an issue with the `player:addItem` method. It doesn't matter if you pass `dropOnMap` as `true` which is the default anyways
* Does work with insufficient cap
* Commit I tested this on is `24508912`

# UPDATE: Temporary workaround

Comment these lines in `iologindata.cpp`

```cpp
if (!player->inventory[CONST_SLOT_STORE_INBOX]) {
        player->internalAddThing(CONST_SLOT_STORE_INBOX, Item::CreateItem(ITEM_STORE_INBOX));
}
```

If you have already created players you will need to delete item `26052` from all off them. Also, not sure if this might cause issues in modern clients, but yeah...

nekiro:
Fix #3185 

New class StoreInbox inherits from Container
Every player now has its own StoreInbox.
To get player store inbox use player:getStoreInbox()
Store inbox has its own dedicated table in database.
To add item to store inbox it must be store item:
```
<item id="2382" article="a" name="club">
	<attribute key="weight" value="2500" />
	<attribute key="defense" value="7" />
	<attribute key="attack" value="7" />
	<attribute key="weaponType" value="club" />
	<attribute key="storeitem" value="true" />
</item>
```
or
item:setStoreItem(true)

New item class methods:
item:isStoreItem()
item:setStoreItem(bool storeItem)

New ItemType class methods:
itemType:isStoreItem()

inventory[CONST_SLOT_STORE_INBOX] is empty, it acts like defined const only.

Feel free to test it!


EPuncker:
I'm unable to open it anymore after your changes 😄 always get "Sorry, not possible."

> store inbox has no functionality
the items there shouldn't be dropped on death and that is a pretty important functionality (IMO)

> We shouldn't consider store inbox slot as normal inventory slot
as long as scripts could still count items from inside it, its ok

> these checks are adding unnecessary complexity.
well, that is just how the slot works, the only "missing part" we have is the shop system and reward system that add items to the inbox

nekiro:
> I'm unable to open it anymore after your changes 😄 always get "Sorry, not possible."
> 
> > store inbox has no functionality
> 
> the items there shouldn't be dropped on death and that is a pretty important functionality (IMO)
> 
> > We shouldn't consider store inbox slot as normal inventory slot
> 
> as long as scripts could still count items from inside it, its ok
> 
> > these checks are adding unnecessary complexity.
> 
> well, that is just how the slot works, the only "missing part" we have is the shop system and reward system that add items to the inbox
> 
> PS: there seems to be a warning at iologindata.cpp

You can't drop store inbox, because it's not even considered in droploot.lua anyway.
Store inbox is rather problematic I think we shouldn't consider it as CONST_SLOT across the sources, but anyway the main problem with your changes in container::queryAdd was that server considered this slot as valid slot to add items to (cause its a container), but it couldn't without FLAG_NOLIMIT, so it failed and deleted items like it was mentioned in #3185

nekiro:
Update repository with whole new logic for store inbox, test it please :)

src/tools.cpp
```diff
@@ -1222,6 +1222,9 @@ const char* getReturnMessage(ReturnValue value)
 		case RETURNVALUE_YOUDONTHAVEREQUIREDPROFESSION:
 			return "You don't have the required profession.";
 
+		case RETURNVALUE_CANNOTMOVEITEMISNOTSTOREITEM:
```
EPuncker, line 1225: we also need the "This item cannot be moved there." when trying to move items from store inbox to anywhere that is not a depot box (or house if its a wrapable item)

data/migrations/25.lua
```diff
@@ -1,3 +1,16 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 25 (Store inbox changes)")
+	db.query([[
+		CREATE TABLE IF NOT EXISTS `player_storeinboxitems` (
+		  `player_id` int(11) NOT NULL,
+		  `sid` int(11) NOT NULL,
```
DSpeichert, line 6: `slot_id` ?

data/migrations/25.lua
```diff
@@ -1,3 +1,16 @@
...
+		  `pid` int(11) NOT NULL DEFAULT '0',
```
DSpeichert, line 7: `parent_id`?

data/migrations/25.lua
```diff
@@ -1,3 +1,16 @@
...
+		  `itemtype` smallint(6) NOT NULL,
```
DSpeichert, line 8: `item_type`?

data/migrations/25.lua
```diff
@@ -1,3 +1,16 @@
...
+		  `count` smallint(5) NOT NULL DEFAULT '0',
+		  `attributes` blob NOT NULL,
+		  UNIQUE KEY `player_id_2` (`player_id`, `sid`),
+		  FOREIGN KEY (`player_id`) REFERENCES `players`(`id`) ON DELETE CASCADE
+		) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
```
DSpeichert, line 13: Shouldn't there be foreign key on `parent_id`?

nekiro: Foreign key to what? pid - parent_id is id of another container in the same table, this table is cloned inboxitems table and few others

nekiro: I can change these, but then it will be inconsistent with rest of the tables that got the same structure like player_depotitems, player_inboxitems etc.

DSpeichert: Foreign key to the same table. You can make parent_id to have a foreign key constraint on `id`, although it would have to default to NULL, as 0 is likely not a valid record.

I believe we were talking about changing all the other tables as well, but it was tabled until 2.0 because it's a change that could be breaking. Here, you're creating a new table, so it's easy to make the column names sound from the get-go.

nekiro: Kinda don't get it also don't get it why would we need that.
Did you mean?
```
FOREIGN KEY (`parent_id`) REFERENCES `player_storeinboxitems`(`item_type`) ON DELETE CASCADE
```

DSpeichert (approved):

EPuncker:
just leaving this here for reference:
https://github.com/otland/forgottenserver/issues/2156

MillhioreBT:
i already know what this PR is merged and aproved, but i want resaltar algo importante, y saber si es intencional o un error

i have a question, is strange but i want send a backpack with various item into this to the store inbox, but is impossible, only work for single items, per example:

That work:
```lua
local item = Game.createItem(xxxx, x)
if item then
	item:storeItem(true)
	storeInbox:addItemEx(item)
	item:storeItem(false)
end
```

But that not:
```lua
local item = Game.createItem(xxxx, x)
if item then
	for i in pairs(items) do
		local itm = Game.createItem(i.id, i.count)
		if itm then
			item:addItemEx(itm)
		end
	end
	item:storeItem(true)
	storeInbox:addItemEx(item)
	item:storeItem(false)
end
```

It also doesn't work like this:
```lua
local item = Game.createItem(xxxx, x)
if item then
	local itmList = {}
	for i in pairs(items) do
		local itm = Game.createItem(i.id, i.count)
		if itm then
			itm:storeItem(true)
			item:addItemEx(itm)
			table.insert(itmList, itm)
		end
	end
	item:storeItem(true)
	storeInbox:addItemEx(item)
	item:storeItem(false)
	for i in pairs(itmList) do
		i:storeItem(false)
	end
end
```

PR #3189, opened 2020-10-17
Add backwards compatibility for pre 1.x loot containers

nekiro:
In old versions you used to write loot containers like:
``` 
<item id="1987" chance="100000"> 
        <inside>
            <item id="2597" chance="8300" />
            <item id="2553" chance="10750" /> 
            <item id="2530" chance="10750" />
            <item id="2213" chance="150" /> 
        </inside>
    </item>
```
now you use it like:
```
  <item id="1987" chance="100000"> 
            <item id="2597" chance="8300" />
            <item id="2553" chance="10750" /> 
            <item id="2530" chance="10750" />
            <item id="2213" chance="150" /> 
    </item>
```

This pr brings back the optional "inside" tag, so it does parse both and is backwards compatible in result.

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -1362,6 +1362,8 @@ bool Monsters::loadLootItem(const pugi::xml_node& node, LootBlock& lootBlock)
 
 void Monsters::loadLootContainer(const pugi::xml_node& node, LootBlock& lBlock)
 {
+	// NOTE: <inside> attribute was left for backwards compatibility with pre 1.x TFS versions.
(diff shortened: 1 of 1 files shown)


EPuncker (approved):
yeah backward compatibility is always nice:
https://i.imgur.com/vDO1H4x.jpg

DSpeichert (commented):
I'd add a comment that this was brought back for compatibility, so it doesn't get axed in the future when someone decides that "it's not used in the datapack anyway"

DSpeichert (approved):

PR #3190, opened 2020-10-17
WIP Fix sight line algorithm

### LINKED ISSUE #2727: Wrong behavior of monsters with distance attacks.
**This issue is reproducible on Tfs 1.2**

Steps to reproduce (include any configuration/script required to reproduce)
Look for some monster that needs distance from you to attack.
Example: Warlocks

**Expected behaviour** -  I will upload a movie here, this will make understanding easier
https://streamable.com/h2uxk

In this case, the monster tries to go to the player to attack him.

**Actual behaviour** [Wrong] -  the movie here.
https://streamable.com/3jlj3

the monster is standing attacking the target, when he should go to the player and come back.

Environment
Using Windows 10, 64 bits.

Does anyone have any suggestions to fix this?
what is the function responsible for this behavior?

nekiro:
Attempt to fix #2727 

Not sure if this is the correct behaviour, but I tried. Feel free to test it deeply, cause I did only briefly.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -511,12 +511,10 @@ bool Map::checkSightLine(const Position& fromPos, const Position& toPos) const
 		int32_t move_ver = std::abs(A * (start.x) + B * (start.y + my) + C);
 		int32_t move_cross = std::abs(A * (start.x + mx) + B * (start.y + my) + C);
 
-		if (start.y != destination.y && (start.x == destination.x || move_hor > move_ver || move_hor > move_cross)) {
-			start.y += my;
-		}
-
 		if (start.x != destination.x && (start.y == destination.y || move_ver > move_hor || move_ver > move_cross)) {
 			start.x += mx;
+		} else if (start.y != destination.y && (start.x == destination.x || move_hor > move_ver || move_hor > move_cross)) {
+			start.y += my;
 		}
 
 		const Tile* tile = getTile(start.x, start.y, start.z);


EPuncker (approved):
depois
yeah that worked for me, dunno if it have any side effect tho, but don't think so

EvilHero90:
I'm not sure how it is correctly working in Tibia but logic wise both of the left orcs should also move towards the target to be able to shoot or am I wrong?

EPuncker:
> I'm not sure how it is correctly working in Tibia but logic wise both of the left orcs should also move towards the target to be able to shoot or am I wrong?

they are on target distance "4" so I didn't bother, but yeah, maybe they shouldn't stand still, but I think that is related to the flag **staticattack** (that I see no behavior change btw changing it to any number)

EvilHero90:
> 
> 
> > I'm not sure how it is correctly working in Tibia but logic wise both of the left orcs should also move towards the target to be able to shoot or am I wrong?
> 
> they are on target distance "4" so I didn't bother, but yeah, maybe they shouldn't stand still, but I think that is related to the flag **staticattack** (that I see no behavior change btw changing it to any number)

No I mean more like because of line of sight, basicly where those monsters are standing they shouldn't be able to attack the player, the upper left orc would have to walk to the edge same like the orc above the player to see him and the left bottom orc would have to walk 2sqm to be able to see the player, correct me if I'm wrong.

EPuncker:
> > > I'm not sure how it is correctly working in Tibia but logic wise both of the left orcs should also move towards the target to be able to shoot or am I wrong?
> > 
> > 
> > they are on target distance "4" so I didn't bother, but yeah, maybe they shouldn't stand still, but I think that is related to the flag **staticattack** (that I see no behavior change btw changing it to any number)
> 
> No I mean more like because of line of sight, basicly where those monsters are standing they shouldn't be able to attack the player, the upper left orc would have to walk to the edge same like the orc above the player to see him and the left bottom orc would have to walk 2sqm to be able to see the player, correct me if I'm wrong.

Okay I got confused too and loged in to rl tibia to test this, and no, they don't walk to, they can shot just like in the gif above, I tried to record some footage but cipsoft is doing a great job at blocking screencapturers (but not the bots lol), so if you guys want to check this behavior its very easy, just go to thais, the king castle and down to the orcs

joseluis2g:
@EPuncker have you tried with obs?

EPuncker:
> @EPuncker have you tried with obs?

just did, too complicated for me (first time using it) and as I can see only provide high size files, there is no option to record in low resolution/fps

Sundance:
This is the behaviour on real tibia
https://streamable.com/2mc5ua
It's probably a bug that players get an extra sqm on the right side as you can see with the crate

Zbizu:
Why not just use [an algorithm that just draws lines pixel by pixel](https://en.wikipedia.org/wiki/Line_drawing_algorithm) to check tiles? That's probably what they're using.

I did some testing and it seems to be working:
```
bool Map::checkSightLine(const Position& fromPos, const Position& toPos) const
{
	if (fromPos == toPos) {
		return true;
	}

	Position start(fromPos.z > toPos.z ? toPos : fromPos);
	Position destination(fromPos.z > toPos.z ? fromPos : toPos);

	double ax = (double)start.x;
	double bx = (double)destination.x;
	double ay = (double)start.y;
	double by = (double)destination.y;

	double dx = bx - ax;
	double dy = by - ay;



	for (int x = ax; x <= bx; x++) {
		double y = round(ay + dy * (x - ax) / dx);

		const Tile* tile = getTile(x, (int)y, start.z);

		//print the results
		//std::cout << (tile ? "true" : "false") << " " << x << ":" << y << std::endl;

		if (tile && tile->hasProperty(CONST_PROP_BLOCKPROJECTILE)) {
			return false;
		}
	}

	// now we need to perform a jump between floors to see if everything is clear (literally)
	while (start.z != destination.z) {
		const Tile* tile = getTile(start.x, start.y, start.z);
		if (tile && tile->getThingCount() > 0) {
			return false;
		}

		start.z++;
	}

	return true;
}
```

However, there is a `-nan(ind)` output for `y` value sometimes and there is way more elegant way to write that, therefore I leave it up to you.

nekiro:
> Why not just use [an algorithm that just draws lines pixel by pixel](https://en.wikipedia.org/wiki/Line_drawing_algorithm) to check tiles? That's probably what they're using.
> 
> I did some testing and this seems to be working:
> 
> ```
> bool Map::checkSightLine(const Position& fromPos, const Position& toPos) const
> {
> 	if (fromPos == toPos) {
> 		return true;
> 	}
> 
> 	Position start(fromPos.z > toPos.z ? toPos : fromPos);
> 	Position destination(fromPos.z > toPos.z ? fromPos : toPos);
> 
> 	double ax = (double)start.x;
> 	double bx = (double)destination.x;
> 	double ay = (double)start.y;
> 	double by = (double)destination.y;
> 
> 	double dx = bx - ax;
> 	double dy = by - ay;
> 
> 
> 
> 	for (int x = ax; x < bx; x++) {
> 		double y = round(ay + dy * (x - ax) / dx);
> 
> 		const Tile* tile = getTile(x, (int)y, start.z);
> 
> 		//print the results
> 		//std::cout << (tile ? "true" : "false") << " " << x << ":" << y << std::endl;
> 
> 		if (tile && tile->hasProperty(CONST_PROP_BLOCKPROJECTILE)) {
> 			return false;
> 		}
> 	}
> 
> 	// now we need to perform a jump between floors to see if everything is clear (literally)
> 	while (start.z != destination.z) {
> 		const Tile* tile = getTile(start.x, start.y, start.z);
> 		if (tile && tile->getThingCount() > 0) {
> 			return false;
> 		}
> 
> 		start.z++;
> 	}
> 
> 	return true;
> }
> ```
> 
> However, there is way more elegant way to write that, therefore I leave it up to you. Remember to round values properly. Leaving that to compiler may mess the line drawing behaviour.
> 
> I'm also not sure how will the code behave if both frompos and topos are on same Y axis, forgot to test that, probably needs an adjustment.

Nope, that's not their sight line algorithm.

Zbizu:
further real tibia analysis:
obraz

now I'm even more confident that they're using the algorithm I mentioned earlier (uses y axis instead if the slope is above 45 degrees)
obraz

lua implementation I used if anyone is curious (still glitchy, but it gives almost identical results):
```
function math.round(n)
	return n % 1 >= 0.5 and math.ceil(n) or math.floor(n)
end

function getAngle(peak, leftArm, rightArm) -- A B C
	local AB = {x = peak.x - leftArm.x, y = peak.y - leftArm.y}
	local BC = {x = rightArm.x - leftArm.x, y = rightArm.y - leftArm.y}
	
	local vecMult = AB.x * BC.x + AB.y * BC.y
	local vecLen = math.sqrt(AB.x^2 + AB.y^2) * math.sqrt(BC.x^2 + BC.y^2)
	if vecLen == 0 then
		return 0
	end
	
	local theta = math.asin(vecMult / vecLen)
	local result = theta * 180 / math.pi
	return result
end

function draw_line_test(fromPos, toPos, itemId)
	if fromPos == toPos then
		return
	end
	
	local ax = fromPos.x
	local bx = toPos.x
	local ay = fromPos.y
	local by = toPos.y
	
	local dx = bx - ax
	local dy = by - ay
	
	local axisPos = Position(toPos.x, fromPos.y, fromPos.z)
	
	local angle = getAngle(fromPos, toPos, axisPos)
	if angle < 45 then
		if bx < ax then
			local t = bx
			bx = ax
			ax = t
		end
		
		for x = ax, bx do
			local y
			
			if x < bx then
				y = math.floor(ay + dy * (x - ax) / dx) -- round(ay + dy * (x - ax) / dx); w c++
			else
				y = toPos.y
			end
			
			local checkPos = Position(x, y, toPos.z)
			local tile = Tile(checkPos)

			if tile then
				Game.createItem(itemId, 1, checkPos)
			end

			if tile:hasProperty(CONST_PROP_BLOCKPROJECTILE) then
				return false
			end
		end
	else
		if by < ay then
			local t = by
			by = ay
			ay = t
		end

		for y = ay, by do
			local x
			
			if y < by then
				x = math.floor(ax + dx * (y - ay) / dy) -- round
			else
				x = toPos.x
			end
			
			local checkPos = Position(x, y, toPos.z)
			local tile = Tile(checkPos)

			if tile then
				Game.createItem(itemId, 1, checkPos)
			end

			if tile:hasProperty(CONST_PROP_BLOCKPROJECTILE) then
				return false
			end
		end
	end
end

		return true
	end

		return false
	end

	if not (STARTPOS and ENDPOS) then
		player:sendCancelMessage("Points not set.")
	end
	
	local groundId = tonumber(param)
	if not ItemType(groundId) then
		player:sendCancelMessage("Item id does not exist.")
	end
	
	draw_line_test(STARTPOS, ENDPOS, groundId)
	return false
end
```

Zbizu:
ok I managed to reproduce that in lua:
```
function draw_line_test(fromPos, toPos, itemId)
	if fromPos == toPos then
		return
	end

	local x0 = fromPos.x
	local x1 = toPos.x
	local y0 = fromPos.y
	local y1 = toPos.y

	-- checks if line slope is above 45 degrees
	local steep = math.abs(y1 - y0) > math.abs(x1 - x0)

	if steep then
		-- in c++ it will be:
		-- std::swap(x0, y0);
		-- std::swap(x1, y1);
		x0, y0 = y0, x0
		x1, y1 = y1, x1
	end

	if x0 > x1 then
		-- also swap
		x0, x1 = x1, x0
		y0, y1 = y1, y0
	end

	local dx = x1 - x0
	local dy = y1 - y0
	local grad = 1

	if dx ~= 0 then
		grad = dy / dx
	end

	-- handle first endpoint
	local endX = math.floor(x0)
	local endY = y0 + grad * (endX - x0)
	local xpxl1 = endX
	local ypxl1 = math.floor(endY)

	if steep then
		local posA = Position(ypxl1, xpxl1, fromPos.z)
		Game.createItem(itemId, 1, posA)
	else
		local posA = Position(xpxl1, ypxl1, fromPos.z)
		Game.createItem(itemId, 1, posA)
	end

	local interY = endY + grad -- first y-intersection for the main loop

	-- handle second endpoint
	endX = math.floor(x1)
	endY = y1 + grad * (endX - x1)
	local xpxl2 = endX
	local ypxl2 = math.floor(endY)

	if steep then
		local posA = Position(ypxl2,   xpxl2, fromPos.z)
		Game.createItem(itemId, 1, posA)
	else
		local posA = Position(xpxl2, ypxl2, fromPos.z)
		Game.createItem(itemId, 1, posA)
	end

	-- main loop
	if steep then
		for x = xpxl1+1, xpxl2-1 do
			local posA = Position(math.floor(interY), x, fromPos.z)
			Game.createItem(itemId, 1, posA)
			interY = interY + grad
		end
	else
		for x = xpxl1+1, xpxl2-1 do
			local posA = Position(x, math.floor(interY), fromPos.z)
			Game.createItem(itemId, 1, posA)
			interY = interY + grad
		end
	end
end

		return true
	end

		return false
	end

	if not (STARTPOS and ENDPOS) then
		player:sendCancelMessage("Points not set.")
	end

	local groundId = tonumber(param)
	if not ItemType(groundId) then
		player:sendCancelMessage("Item id does not exist.")
	end

	draw_line_test(STARTPOS, ENDPOS, groundId)
	return false
end
```

some item with use with action:
```
LINE_MODE = false
STARTPOS = nil
ENDPOS = nil

	local pos = target:getPosition()
	if LINE_MODE then
		LINE_MODE = false
		ENDPOS = pos
		player:sendTextMessage(MESSAGE_INFO_DESCR, "End point set.")
	else
		LINE_MODE = true
		STARTPOS = pos
		player:sendTextMessage(MESSAGE_INFO_DESCR, "Start point set.")
	end

	return true
end
```

talkaction (eg. set two points with item then run `/draw 231`)
```
<talkaction words="/draw" separator=" " script="drawLine.lua" />
```

nekiro:
@Zbizu please create new PR, thanks.

Codinablack:
So looking over the files changed, was that not a proper working fix or was it just merge issues, or what? I don't understand why it is now marked closed.

Znote:
> So looking over the files changed, was that not a proper working fix or was it just merge issues, or what? I don't understand why it is now marked closed.

The problem is tricky, the new PR is:
https://github.com/otland/forgottenserver/pull/3390

PR #3191, opened 2020-10-18
Adding attribute talkaction

omarcopires:


data/scripts/talkactions/commands/attributes.lua
```diff
@@ -0,0 +1,322 @@
...
+
+local itemFunctions = {
+    ["actionid"] = {
+        isActive = true,
+        targetFunction = function(item, target)
+            return item:setActionId(target)
+        end
+    },
+
+    ["uniqueid"] = {
```
MillhioreBT, line 11: It is not allowed to set a `uniqueId`, so why add this option?

data/scripts/talkactions/commands/attributes.lua
```diff
@@ -0,0 +1,322 @@
...
+    },
+
+    ["name"] = {
+        isActive = true,
+        targetFunction = function(item, target)
+            return item:setAttribute(ITEM_ATTRIBUTE_NAME, target)
+        end
+    },
+
+    ["remove"] = {
```
MillhioreBT, line 32: We already have `/r` and it works better because we can set the amount we want to erase, also everyone already knows this command and I doubt they want to use this method instead

data/scripts/talkactions/commands/attributes.lua
```diff
@@ -0,0 +1,322 @@
...
+    },
+
+    ["transform"] = {
+        isActive = true,
+        targetFunction = function(item, target)
+            return item:transform(target)
+        end
+    },
+
+    ["clone"] = {
```
MillhioreBT, line 53: Why clone if the result obtained is not stored anywhere and we cannot use the cloned article, this code must be improved if we really want this clone option, as it currently  is useless

data/scripts/talkactions/commands/attributes.lua
```diff
@@ -0,0 +1,322 @@
...
+    },
+
+    ["group"] = {
+        isActive = true,
+        targetFunction = function(player, target)
+            return player:setGroup(Group(target))
+        end
+    },
+
+    ["vocation"] = {
```
MillhioreBT, line 176: This has a problem, it happens that when trying something like this: `/attr vocation,1` the parameter `1` is sent to function `Vocation` as a string and not as a number and then the result is: `[Warning - Vocations::getVocation] Vocation 65535 not found.`

This is not your fault, what happens is that in the sources the function: `isNumber` from the lua interface only checks if the value on the stack is really number and not anything convertible to a number.

Just modify the code and handle this case correctly ;)

data/scripts/talkactions/commands/attributes.lua
```diff
@@ -0,0 +1,322 @@
...
+    },
+
+    ["stamina"] = {
+        isActive = true,
+        targetFunction = function(player, target)
+            return player:setStamina(target)
+        end
+    },
+
+    ["town"] = {
```
MillhioreBT, line 190: The same thing happens how `vocation`

data/scripts/talkactions/commands/attributes.lua
```diff
@@ -0,0 +1,322 @@
...
+    },
+
+    ["maxHealth"] = {
+        isActive = true,
+        targetFunction = function(player, target)
+            return player:setMaxHealth(target)
+        end
+    },
+
+    ["addItem"] = {
```
MillhioreBT, line 239: The same thing happens how `vocation`

data/scripts/talkactions/commands/attributes.lua
```diff
@@ -0,0 +1,322 @@
...
+        isActive = true,
+        targetFunction = function(player, target)
+            return player:addItem(target, 1)
+        end
+    },
+
+    ["removeItem"] = {
```
MillhioreBT, line 246: The same thing happens how `vocation`

data/scripts/talkactions/commands/attributes.lua
omarcopires: Could you submit a change suggestion?

PR #3192, opened 2020-10-18
Server save script improvements

omarcopires:
Now the map will not be cleared before restarting the server.

### DIFF
--- a/data/globalevents/scripts/serversave.lua
+++ b/data/globalevents/scripts/serversave.lua
@@ -1,14 +1,19 @@
 local function ServerSave()
-	if configManager.getBoolean(configKeys.SERVER_SAVE_CLEAN_MAP) then
-		cleanMap()
-	end
-
-	if configManager.getBoolean(configKeys.SERVER_SAVE_CLOSE) then
-		Game.setGameState(GAME_STATE_CLOSED)
-	end
-
 	if configManager.getBoolean(configKeys.SERVER_SAVE_SHUTDOWN) then
 		Game.setGameState(GAME_STATE_SHUTDOWN)
+	else
+		local closeAtServerSave = configManager.getBoolean(configKeys.SERVER_SAVE_CLOSE)
+		if closeAtServerSave then
+			Game.setGameState(GAME_STATE_CLOSED)
+		end
+
+		if configManager.getBoolean(configKeys.SERVER_SAVE_CLEAN_MAP) then
+			cleanMap()
+		end
+
+		if closeAtServerSave then
+			Game.setGameState(GAME_STATE_NORMAL)
+		end
 	end
 end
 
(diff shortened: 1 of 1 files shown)


DSpeichert (commented):
I don't see the previous version opening the server back. Did it make sense? Was there a reason for it?

omarcopires:
> Não vejo a versão anterior abrindo o servidor de volta. Fez sentido? Havia alguma razão para isso?

There is no reason to be able to clean if you will shutdown to restart it.

Zbizu:
how is this script supposed to save server when it doesn't have `saveServer()` anywhere?

also while we're at it, why there is no /save talkaction?

data/globalevents/scripts/serversave.lua
```diff
@@ -1,14 +1,19 @@
 local function ServerSave()
```
Zbizu, line 1: the function that actually saves the server is `saveServer()`
it doesn't appear anywhere in this function, therefore the server doesn't really get saved

Zbizu (changes requested):

MillhioreBT:
> I don't see the previous version opening the server back. Did it make sense? Was there a reason for it?

I agree with @DSpeichert
the only situations in which the server is actually saved is when it is closed or when it is shut down, why is this event called `ServerSave`? if your main purpose is not really to save the server every so often, the name `LastServerSave` makes more sense :D

I always thought this was a mistake, I had to create my own custom event for this purpose, since the current script is useless, it is not useful at all

MillhioreBT (dismissed):

DSpeichert:
It still doesn't actually call saveServer as @Zbizu pointed out.
I understand that by closing the server, it kicks people out, which in turns saves all those players that were kicked out - but not those that are immune to server closing via special permissions.

Do we want to keep that behavior with closing the server or use the saveServer function instead?
I'd think that if you kick all the players (by closing server) you might as well restart it.

Zbizu:
kicking players is ok - they get saved before everything else so no rollback will happen

however, `saveServer` has to be called in order to save houses

Clean can be executed last, after everything was saved.

EPuncker (approved):

data/globalevents/scripts/serversave.lua
```diff
@@ -1,15 +1,22 @@
...
+		if configManager.getBoolean(configKeys.SERVER_SAVE_CLEAN_MAP) then
+			cleanMap()
+		end
+
+		if closeAtServerSave then
+			Game.setGameState(GAME_STATE_NORMAL)
+		end
 	end
+
+	saveServer()
```
DSpeichert, line 19: I don't think this makes sense in `GAME_STATE_SHUTDOWN` state so probably belongs to the `else` case

Zbizu: I'd also move it above cleanMap. It's better to save before you start removing items bevause the server may crash when something tries to reference a removed item.

omarcopires: I don't know if I understand correctly, but I believe that it is now working well.

Zbizu (approved):

EPuncker (approved):

PR #3196, opened 2020-10-19
Add additional prints for invalid data in npc lib

### LINKED ISSUE #3149: Add invalid itemID warning to shop NPC item lists
https://github.com/otland/forgottenserver/blob/master/data/npc/lib/npcsystem/modules.lua#L568-L715

Currently we have warnings for missing subtype, cost and whatsoever on our modules.lua, but not a single warning to invalid itemID, so if we mistype or use an itemID that is not present in our version (and the NPC has a very long list of items, like Yasir), we will just get a debug when getting the shop window!

PS: a warning for duplicated item IDs would be nice too

and also maybe we should add a check on the ProtocolGame::sendShop/AddShopItem/sendSaleItemList if the item exist or not

nekiro:
Fix #3149

### DIFF
--- a/data/npc/lib/npcsystem/modules.lua
+++ b/data/npc/lib/npcsystem/modules.lua
@@ -566,6 +566,7 @@ if Modules == nil then
 
 	-- Parse a string contaning a set of buyable items.
 	function ShopModule:parseBuyable(data)
+		local alreadyParsedIds = {}
 		for item in string.gmatch(data, "[^;]+") do
 			local i = 1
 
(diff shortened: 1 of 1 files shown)


DSpeichert (dismissed):

DSpeichert (approved):

omarcopires (approved):

Znote:
Does `alreadyParsedIds = {}` has to be re-initialized everytime? Wouldn't it be more efficient to parse valid ids once globally?

nekiro:
> Does `alreadyParsedIds = {}` has to be re-initialized everytime? Wouldn't it be more efficient to parse valid ids once globally?
> 
> How often does `function ShopModule:parseBuyable(data)` executes? Once per NPC, once per "trade"? Once per interaction?

Once per npc load, so it doesnt matter.

PR #3197, opened 2020-10-23
Additional checks for unsigned integers to avoid underflow and overflow

nekiro:
Additional checks in getNumber<T>(L, arg)

Prints a handy warning if unsigned integer is negative or exceeds the max number for defined type.

image

### DIFF
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -280,12 +280,37 @@ class LuaScriptInterface
 		{
 			return static_cast<T>(static_cast<int64_t>(lua_tonumber(L, arg)));
 		}
+
 		template<typename T>
-		static typename std::enable_if<std::is_integral<T>::value || std::is_floating_point<T>::value, T>::type
+		static typename std::enable_if<std::is_integral<T>::value && std::is_unsigned<T>::value, T>::type
 			getNumber(lua_State* L, int32_t arg)
 		{
-			return static_cast<T>(lua_tonumber(L, arg));
+			LUA_NUMBER luaNum = lua_tonumber(L, arg);
+			if (luaNum < 0 || luaNum > std::numeric_limits<T>::max()) {
+				std::ostringstream ss;
+				ss << "Passed argument " << arg << " has invalid value" << std::endl;
+				reportErrorFunc(ss.str());
+				return 0;
+			}
+
+			return static_cast<T>(luaNum);
 		}
+
+		template<typename T>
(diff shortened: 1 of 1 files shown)


DSpeichert (dismissed):

DSpeichert (approved):

Znote (approved):

DSpeichert:
Note the potential performance impact as noted by Mark.

MillhioreBT:
This implementation causes performance to be compromised, in fact it drops by about 70%~

You can try the following code and you can compare the time, you will notice that with this implementation increases the time dramatically.

```c++
time_t now = OTSYS_TIME();
for (int i = 0; i < 1000000000; i++) {
    getNumber<time_t>(L, 2);
}
std::cout << "Test getNumber of LuaInterface: " << std::setprecision(15) << (double)(OTSYS_TIME() - now) << std::endl;
```

I know that there are never cases in which `getNumber` is called with this frequency, I don't know if it is worth it, Someone else should do tests, I don't know if the tests I did were enough or well done

yamaken93:
@MillhioreBT are you testing with release version all optimizations enabled?

MillhioreBT:
> @MillhioreBT are you testing with release version all optimizations enabled?

Of course, the code is clear to the naked eye, obviously it takes more time, but it wouldn't be a big problem really, since at small scales there doesn't seem to be a much difference.
If calling `getNumber(L, 2);` 1000000000 times requires 3 seconds, with these modifications it will require approximately 5 seconds, but let's face it, when can this situation occur?

MillhioreBT:
It would be amazing to have the opinion of @marksamman about this

nekiro:
> It would be amazing to have the opinion of @marksamman about this

Is it dropping performance in general or is it with invalid value only?

DSpeichert (approved):

Zbizu:
imo not worth losing 70% of the performance for that. Should be optional in compiler settings.

Zbizu (changes requested):
imo not worth losing 70% of the performance for that. Should be optional in compiler settings.

DSpeichert:
The test conducted by @MillhioreBT is synthetic, not representative of what datapacks in the wild do.
In the absence of other comments over the time, I'm willing to take it as a feature unless real-world performance problems are reported.

dex-89:
image

PR #3199, opened 2020-10-25
Avoid copying Position when not needed

nekiro:


ranisalt:
Note that none of these calls will trigger a copy due to [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision).

tl;dr the Position object is constructed directly in the stack of the callee, avoiding a copy anyway.

PR #3201, opened 2020-10-26
Remove redundant parentheses and use enums

nekiro:
Follow up to #3027

### DIFF
--- a/src/outfit.cpp
+++ b/src/outfit.cpp
@@ -78,7 +78,7 @@ const Outfit* Outfits::getOutfitByLookType(PlayerSex_t sex, uint16_t lookType) c
 
 const Outfit* Outfits::getOutfitByLookType(uint16_t lookType) const
 {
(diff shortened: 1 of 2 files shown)


Znote (approved):

src/outfit.cpp
```diff
@@ -78,7 +78,7 @@ const Outfit* Outfits::getOutfitByLookType(PlayerSex_t sex, uint16_t lookType) c
 
 const Outfit* Outfits::getOutfitByLookType(uint16_t lookType) const
 {
-	for (uint8_t sex = 0; sex <= 1; sex++) {
+	for (uint8_t sex = PLAYERSEX_FEMALE; sex <= PLAYERSEX_LAST; sex++) {
```
marmichalski, line 81: I'd argue that adding `PLAYERSEX_FIRST` enum would be a better choice and would be done as other enums. If you decide to add it, do not forget to register it in lua (along with `PLAYERSEX_LAST` which is missing for some reason 🤔)

PR #3202, opened 2020-10-28
Changing group name

omarcopires:


Zbizu:
I see no reason to name it community manager. A developer or an administrator would be more a suitable name if there's anything to change.

DSpeichert:
I'm going to close this because it's such an arbitrary change with no reason and explanation.

PR #3204, opened 2020-11-03
Fix: rate spawn in config.lua don't works

omarcopires:


Kamenuvol:
Rate spawn is the number of monsters that spawn per cycle. e.g: spawn with 5 goblins and 30sec spawn time. you kill them all and only after 150sec without blocking respawn that they will be all respawned.

EPuncker:
take a look:
https://github.com/otland/forgottenserver/pull/3134/files

PR #3209, opened 2020-11-07
Fix isSummon on compat.lua and refactoring player:sendChannelMessage

LukSrT:


nekiro (approved):

Znote:
I like the changes but I think the title of this PR could be a bit more declarative beside "code improvements". 
What would be a better explanation? 
`Minor fix on compat.lua and refactoring player:sendChannelMessage` ?

PR #3211, opened 2020-11-10
Decay time random

MillhioreBT:
I know that I had already opened a PR for this issue, but the truth was I had some problems on my computer and I accidentally deleted everything, so here it is...

With these changes, it is now possible to configure a range of durations for temporary articles, thus complying with what is mentioned in the PR #3148

> - [x] magic wall should have a random duration between 14 and 20 seconds when shot
> - [x]  wild growth should have a random duration between 38 and 45 seconds when shot


data/items/items.xml
```diff
@@ -686,15 +686,17 @@
 	<item id="1497" article="a" name="magic wall">
 		<attribute key="type" value="magicfield" />
 		<attribute key="decayTo" value="0" />
-		<attribute key="duration" value="20" />
```
Zbizu, line 689: it's nice to have this feature but please don't modify the duration of existing items unless they get official change in rl servers

Zbizu (changes requested):

data/items/items.xml
MillhioreBT: > it's nice to have this feature but please don't modify the duration of existing items unless they get official change in rl servers

Thanks for reviewing <3, I made the changes because it mentions it @EPuncker on the PR 3148

Zbizu: Alright, then looks good to me. Sorry, I didn't knew that. Checked on the wiki and nothing was there either.

EPuncker: > Sorry, I didn't knew that. Checked on the wiki and nothing was there either.

https://github.com/otland/forgottenserver/issues/3148#issuecomment-667573270

MillhioreBT: perfect, I'll fix it later, thanks <3

src/items.h
```diff
@@ -315,6 +316,7 @@ class ItemType
 		uint32_t weight = 0;
 		uint32_t levelDoor = 0;
 		uint32_t decayTime = 0;
+		uint32_t decayTimeMax = 0;
```
ericlancebrown, line 319: Please check indentation.

src/items.cpp
```diff
@@ -66,6 +66,8 @@ const std::unordered_map<std::string, ItemParseAttributes_t> ItemParseAttributes
 	{"transformequipto", ITEM_PARSE_TRANSFORMEQUIPTO},
 	{"transformdeequipto", ITEM_PARSE_TRANSFORMDEEQUIPTO},
 	{"duration", ITEM_PARSE_DURATION},
+	{"durationmin", ITEM_PARSE_DURATION},
```
ericlancebrown, line 69: Shouldn't this be ITEM_PARSE_DURATION_MIN?

ericlancebrown (changes requested):
Some style/indentation things as well as a possible misspelled variable.

src/items.cpp
EPuncker: yes

MillhioreBT: What happens is that I want to reuse the same variable, for this case, since `duration` and `durationmin` are the same, only to make it look more beautiful in the xml file, add a derivative that can be called `durationmin` so that it matches `min` <-> `max`

EPuncker: oh true, I forgot that they are the same, disregard my previous comment then

src/luascript.h
```diff
@@ -1182,6 +1182,9 @@ class LuaScriptInterface
 
 		static int luaItemTypeIsStoreItem(lua_State* L);
 
+		static int luaItemTypeGetDefaultDurationMin(lua_State* L);
```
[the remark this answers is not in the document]
MillhioreBT: Since the repository where I made these changes was deleted, I cannot fix these details, but in case of being merged, of course I will fix the indentation and tabs correctly, even you could do it yourself in any case.

data/items/items.xml
MillhioreBT: > Sorry, I didn't knew that. Checked on the wiki and nothing was there either.

I would like to make the changes, but the repository i deleted it by accident, I suppose I will have to redo everything again, or I don't know what to do about it. thanks!

EPuncker: @MillhioreBT If you are recovering a personal repository that is not part of an organization, start by opening “Settings” in the top right dropdown menu. Click on the “Repositories” menu item. From that screen, you should see a section for deleted repositories in the top bar, click on that.

MillhioreBT: @EPuncker I will try, although I have doubts to achieve it, it can be said that I am the baddest woman using git and github 😭

MillhioreBT:
I will close this PR, and open another with these same changes, since I could not recover the previous repository,
I will take into account the revisions that were made here.

PR #3215, opened 2020-11-12
Update Dockerfile

xitobuh:
luajit is not in edge/testing (anymore)
adressing error:
-- Could NOT find LuaJIT (missing: LUA_INCLUDE_DIR)
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find Lua (missing: LUA_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindLua.cmake:113 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:43 (find_package)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,12 @@ RUN apk add --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/test
   crypto++-dev \
   gcc \
   gmp-dev \
-  luajit-dev \
   make \
   mariadb-connector-c-dev \
   pugixml-dev
+# luajit-dev is in edge/main
+RUN apk add --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/main/ \
+	  luajit-dev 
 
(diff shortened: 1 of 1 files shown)


marmichalski:
Maybe depending on `alpine:edge` is not needed anymore? Can't this be bound to some version already?

xitobuh:
> Maybe depending on `alpine:edge` is not needed anymore? Can't this be bound to some version already?

I'm new at this, hopefully i created this PR the right way!
I'll have a go with alpine:3.12

DSpeichert:
Current Dockerfile works for me with no changes needed.

ranisalt:
In fact, one can use the latest stable release (3.13) to install everything, except `crypto++`, which is only available in edge/testing

DSpeichert:
Builds with #3291

Issue #3217, opened 2020-11-13
incorrect return values in vocation:getRequiredSkillTries(skill, level)


Zbizu:
tl;dr: It's a rounding error that seems to be happening in vocation.cpp only

### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. print(Vocation(3):getRequiredSkillTries(SKILL_AXE, 100))
2. put actual formula instead of c++ one

**lua:** `print(50 * 1.2^(100 - 11))`
**wolframalpha:** type `50 * 1.2^(100 - 11)` in formula box (wolframalpha.com)
**mysql:** type `select FLOOR(50 * POW(1.2,(100 - 11))) as '' ` in phpmyadmin

### Expected behaviour
return value should be 557315218 or 557315218.7056

### Actual behaviour
return value in tfs is 557317189

### Environment
Windows 7, compilation mode 64-bit release

Shawak:
The `float skillMultipliers[SKILL_LAST + 1] = {1.5f, 2.0f, 2.0f, 2.0f, 2.0f, 1.5f, 1.1f};` differ, you used 1.2 but tfs uses 2.0 fox axe.

MillhioreBT:
I have been doing tests by compiling the server several times but the returned value is always wrong.

I did some tests outside the compiler and the result was correct, as shown in the following image:
![image](https://user-images.githubusercontent.com/28090948/99085551-b6640200-259e-11eb-8bc1-8bee734b9d29.png)

I think the problem lies in some configuration present in the compiler regarding the function: `std::pow`

Zbizu:
I think it's more of a rounding/cast error

it returns correct value on doubles only:
```
#include <iostream>
#include<math.h>
using namespace std;

int main() {
	double a = 50.0;
	double b = 1.2;
	double c = 89.0;

	double result = a * pow(b, c);
	cout << result;
	cin.get();

	return 0;
}
```
environment: windows 7, codeblocks 16.01, c++11, gnu gcc

Zbizu:
I tested it on double only and it returned correct values. However, when I reproduced tfs code, it showed incorrect value again so the mistake must be in the code itself.

```
#include <iostream>
#include<math.h>
#include<map>

using namespace std;

[...]

PR #3219, opened 2020-11-14
Correct return values in get required skill tries

MillhioreBT:
This correction was at the request of PR #3217, the code was inspired directly from the repository of SaiyansKing/optimized_forgottenserver@a5f525a


MillhioreBT:
wtf! xd

Issue #3222, opened 2020-11-17
player:addItem(...) freezes the server when the item is a wall


Zbizu:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce
1. go to place where a wall is (tested on tfs rook temple)
2. /r that wall
3. /i a new wall

more details:
what is freezing the server: `Player("playername"):addItem(1026, 1)`
what doesn't: `Game.createItem(1026, 1, pos)`

### Environment
windows 7
compilation mode release 64-bit
tfs sources from Nov 10 2020 (freepremium fix commit)
visual studio 2017, using vcpkg setup
v141, sdk version 10.0.18362.0
boost version 1_71 x64

edit: in fact /i of any wall freezes the server
edit2: this issue also happens in my older builds

on 1.2 it works fine

Zbizu:
Closing because bug was caused for a different reason.

PR #3224, opened 2020-11-17
prevent lag in /i talkaction

Zbizu:
workaround for #3222

this is NOT a bugfix to the issue itself

### DIFF
--- a/data/talkactions/scripts/create_item.lua
+++ b/data/talkactions/scripts/create_item.lua
@@ -43,7 +43,7 @@ function onSay(player, words, param)
 		end
 	end
 
-	local result = player:addItem(itemType:getId(), count)
+	local result
 	if itemType:isPickupable() then
 		result = player:addItem(itemType:getId(), count)
 	else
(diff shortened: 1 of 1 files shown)


MillhioreBT:
I have been testing what you say and it is not true, my server does not cause any delay, if you can give us some other way to reproduce the performance problem it would be great, to help you confirm the problem.

It goes without saying that I tested it with my TFS fork without any changes, totally vanilla as is the master branch

Zbizu:
@MillhioreBT see #3222 

I downloaded the zip of master branch now and compiled it again without any changes (64 bit release same setup as in description) and the freeze still occurs.

EPuncker:
I tested here as well and there is no lag/freeze at all, maybe there is another variable necessary to reproduce this

Zbizu:
@EPuncker I tested it on appveyor build and it also froze

EPuncker:
> @EPuncker I tested it on appveyor build now and it also froze

do you perhaps, have access to a second pc where you could test this?

Zbizu:
ok for some reason it works fine on my laptop so I have no idea what is causing it on my pc

PR #3226, opened 2020-11-17
Removing useless thread variable

rmalizia44:
ThreadHolder already have a thread member, i think this variable is useless.


ericlancebrown (approved):
I see no reason not to remove it. It's used nowhere at all.

EPuncker (approved):

PR #3227, opened 2020-11-20
Vocation cache optimization

### LINKED ISSUE #3217: incorrect return values in vocation:getRequiredSkillTries(skill, level)

EPuncker:
- Change cache complexity from O(log n) to O(1)
- Use faster pow function
- Prevent skill try calculation from underflowing

also fixes #3217

all code by @SaiyansKing and @Erza

### DIFF
--- a/src/vocation.cpp
+++ b/src/vocation.cpp
@@ -87,7 +87,7 @@ bool Vocations::loadFromXml()
 		}
 
 		if ((attr = vocationNode.attribute("manamultiplier"))) {
-			voc.manaMultiplier = pugi::cast<float>(attr.value());
+			voc.manaMultiplier = pugi::cast<double>(attr.value());
 		}
 
 		if ((attr = vocationNode.attribute("attackspeed"))) {
@@ -116,7 +116,7 @@ bool Vocations::loadFromXml()
 				if (skillIdAttribute) {
 					uint16_t skill_id = pugi::cast<uint16_t>(skillIdAttribute.value());
 					if (skill_id <= SKILL_LAST) {
-						voc.skillMultipliers[skill_id] = pugi::cast<float>(childNode.attribute("multiplier").value());
+						voc.skillMultipliers[skill_id] = pugi::cast<double>(childNode.attribute("multiplier").value());
 					} else {
 						std::cout << "[Notice - Vocations::loadFromXml] No valid skill id: " << skill_id << " for vocation: " << voc.id << std::endl;
 					}
@@ -181,19 +181,22 @@ uint16_t Vocations::getPromotedVocation(uint16_t vocationId) const
 
 uint32_t Vocation::skillBase[SKILL_LAST + 1] = {50, 50, 50, 50, 30, 100, 20};
 
-uint64_t Vocation::getReqSkillTries(uint8_t skill, uint16_t level)
+uint64_t Vocation::getReqSkillTries(uint8_t skill, uint32_t level)
 {
 	if (skill > SKILL_LAST || level <= 10) {
 		return 0;
 	}
+	level -= 10;
 
-	auto it = cacheSkill[skill].find(level);
-	if (it != cacheSkill[skill].end()) {
-		return it->second;
+	std::vector<uint64_t>& skillCache = cacheSkill[skill];
+	if (level >= skillCache.size()) {
+		skillCache.resize(static_cast<size_t>(level) + 1, 0);
 	}
 
-	uint64_t tries = static_cast<uint64_t>(skillBase[skill] * std::pow(static_cast<double>(skillMultipliers[skill]), level - 11));
-	cacheSkill[skill][level] = tries;
+	uint64_t& tries = skillCache[level];
+	if (tries == 0) {
+		tries = std::max<uint64_t>(1, static_cast<uint64_t>(skillBase[skill] * fast_pow(skillMultipliers[skill], level - 1)));
+	}
 	return tries;
 }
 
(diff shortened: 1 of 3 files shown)


src/vocation.cpp
```diff
@@ -181,30 +181,39 @@ uint16_t Vocations::getPromotedVocation(uint16_t vocationId) const
...
 uint64_t Vocation::getReqMana(uint32_t magLevel)
 {
-	auto it = cacheMana.find(magLevel);
-	if (it != cacheMana.end()) {
-		return it->second;
+	if (magLevel == 0) {
+		return 0;
+	}
+
+	std::vector<uint64_t>& manaCache = cacheMana;
```
ranisalt, line 209: Why making a reference here and not using `cacheMana` straight away?

src/vocation.cpp
```diff
@@ -181,30 +181,39 @@ uint16_t Vocations::getPromotedVocation(uint16_t vocationId) const
...
+	if (magLevel >= manaCache.size()) {
+		manaCache.resize(static_cast<size_t>(magLevel) + 1, 0);
 	}
 
-	uint64_t reqMana = std::floor<uint64_t>(1600 * std::pow<double>(manaMultiplier, static_cast<int32_t>(magLevel) - 1));
-	cacheMana[magLevel] = reqMana;
-	return reqMana;
+	uint64_t& tries = manaCache[magLevel];
```
ranisalt, line 214: Don't take references from primitives. Use plain `uint64_t`.

ranisalt:
> Change cache complexity from O(log n) to O(1)

Nice!

> Use faster pow function

What makes you think that STL implementation is not better than yours? Did you benchmark?

That said, did someone ever test if caching so little data is worth the overhead? I'm skeptical that accessing the heap and pointer arithmetic is cheaper than a simple math equation.

Znote:
```cpp
constexpr double fast_pow(double base, uint32_t exp) {
	double result = 1.0;
	while (exp) {
		if (exp & 1) {
			result *= base;
		}
		exp >>= 1;
		base *= base;
	}
	return result;
}
```
This does not seem to be C++11 compliant, is there a way to make this body logic into a one liner return expression?

image

Erza:
> ```c++
> constexpr double fast_pow(double base, uint32_t exp) {
> 	double result = 1.0;
> 	while (exp) {
> 		if (exp & 1) {
> 			result *= base;
> 		}
> 		exp >>= 1;
> 		base *= base;
> 	}
> 	return result;
> }
> ```
> 
> This does not seem to be C++11 compliant, is there a way to make this body logic into a one liner return expression?
> 
> image

On Windows, the project targets C++17 along with Visual Studio 2019, the C++ standard in the `CMakeLists.txt` file should be updated to target C++17.

https://github.com/otland/forgottenserver/blob/master/CMakeLists.txt#L13

Znote:
> On Windows, the project targets C++17 along with Visual Studio 2019, the C++ standard in the `CMakeLists.txt` file should be updated to target C++17.
> 
> https://github.com/otland/forgottenserver/blob/master/CMakeLists.txt#L13

This should then be fixed in a separate pull request before we test this one further. I tried a quick and dirty bump to C++17 and getting other errors that should be addressed. (Ubuntu 20.04).

joseluis2g:
change constexpr to static inline

ranisalt:
[Your hand-rolled implementation of power is far slower than `std::pow`.](https://gist.github.com/ranisalt/19332b42d945b729f29a3030ef96c6e3)

image

Erza:
> [Your hand-rolled implementation of power is far slower than `std::pow`.](https://gist.github.com/ranisalt/19332b42d945b729f29a3030ef96c6e3)
> 
> image

Nice, thanks for benching it. In that case the fast_pow function should be dropped.

cc @SaiyansKing (author)

SaiyansKing:
> > [Your hand-rolled implementation of power is far slower than `std::pow`.](https://gist.github.com/ranisalt/19332b42d945b729f29a3030ef96c6e3)
> > image
> 
> Nice, thanks for benching it. In that case the fast_pow function should be dropped.
> 
> cc @SaiyansKing (author)

Do you even test it by yourself? I've compiled the code he posted and boom:
Windows: image
Linux: image

interesting isn't it?
ps. it isn't function that I wrote but it is a well known algorithm [https://en.wikipedia.org/wiki/Exponentiation_by_squaring](url) and yes it is faster than STL implementation because STL implementation needs to guarantee that it'll work in all cases(negative exponent etc).

ranisalt:
We are talking about a 25% variance for a snippet that is running in about 25 to 30ns for both our systems. I pointed out the performance because you named it "fast" pow, but in reality, this optimization is not significant for this use case, beyond depending on the system. When uncertain, I would always choose the built-in implementation.

With that said, [I wrote another code to test whether the cache is significantly helpful](https://gist.github.com/ranisalt/19332b42d945b729f29a3030ef96c6e3) and the conclusion is, again, that it is not. I expect that is because accessing the heap and going through the processor cache when you can just do quick math with what you have in memory is costlier. The difference is within 10%, with no caching being consistently faster.

I'm not 100% sure on the code, though, and the difference might be larger in an actual server with processor cache being dirty more often.

SaiyansKing:
> We are talking about a 25% variance for a snippet that is running in about 25 to 30ns for both our systems. I pointed out the performance because you named it "fast" pow, but in reality, this optimization is not significant for this use case, beyond depending on the system. When uncertain, I would always choose the built-in implementation.

The thing is that I'm 100% sure it is "always" faster because I tested in on many VPS'es(and even online compilators that let you run the code) and noticed 30-40% performance increase. In my case it is 50%, on really old dual core E8500 you can get 60% increase. Of course you can say it isn't fast if you like to I don't care. I didn't make this pull request and I'll not make any pull request so this is my last post here.


> With that said, [I wrote another code to test whether the cache is significantly helpful](https://gist.github.com/ranisalt/19332b42d945b729f29a3030ef96c6e3) and the conclusion is, again, that it is not. I expect that is because accessing the heap and going through the processor cache when you can just do quick math with what you have in memory is costlier. The difference is within 10%, with no caching being consistently faster.

And again for me(and some other older systems) it is 20% better for cache. However I agree that this test shouldn't be taken seriously because it is mostly due to cache being warm.

ps. keep in mind that TFS developers make skill tries cache for some reason and that reason is probably that std::pow perform horrible on older systems. You need to remember that TFS most likely will be run on VPS with some old intel xeon that will benefit from cache not on modern system where things might be different.

MillhioreBT:
The changes work very well, even using the default built-in `std::pow` since `fast_pow` doesn't make much sense.
the result obtained is as expected, and I will definitely use this in my sources, of course without the `fast_pow`

EPuncker:
closed in favor of https://github.com/otland/forgottenserver/pull/3275

Issue #3225, opened 2020-11-17
/i freezes when holding a container with more items than it can hold


Zbizu:
tl;dr: a bag with 10/8 items inside will freeze the server for a few seconds when you try to /i something that can't be picked up (eg. a wall or a ground)

- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce
1. go to items.xml, set a bag volume to 10
2. have this bag in your eq
3. put 10 items in it
4. save server and shut it down
5. set bag volume back to 8
6. start your server again
7. try to /i a ground tile or a wall
8. wait a few seconds to log in again once it unfreezes
9. throw 2 items out of this bag
10. /i works normally

### Expected behaviour
created wall/ground is on the floor without freezing

### Actual behaviour
server freezes

### Environment
windows 7
compilation mode release 64-bit
tfs sources from Nov 10 2020 (freepremium fix commit)
visual studio 2017, using vcpkg setup
v141, sdk version 10.0.18362.0
boost version 1_71 x64

This bug happened in both - tfs compiled by myself and appveyor build.

Zbizu:
also leaving a reference to #3222 for clarity

kygov:
I'm not sure, it may be because of uint32_t overflow here.
https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L2644

EPuncker:
could you test these https://github.com/SaiyansKing/optimized_forgottenserver/commit/a205b5f033e88fe217721116dd623b6ef253bb43

and also open a PR if its the case

PR #3228, opened 2020-11-21
Container overflow fix (freeze on /i)

Zbizu:
regarding #3225

credits [SaiyansKing/optimized_forgottenserver@a205b5f](https://github.com/SaiyansKing/optimized_forgottenserver/commit/a205b5f033e88fe217721116dd623b6ef253bb43)

tests performed:

- picking up items
- /i pickupable item
- /i non-pickupable item

in these situations:
- player has no eq
- player has no eq but has a container
- player has eq but no container
- player has both eq and a container
- player has a overloaded container
- player has multiple overloaded containers in his eq

result:
- items cannot be picked up if there is no room in container
- created items get dropped on the floor with no freeze

Items that are beyond container limit are still inside that container. They can be regained through removing other items from the container.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2641,15 +2641,16 @@ Cylinder* Player::queryDestination(int32_t& index, const Thing& thing, Item** de
 			Container* tmpContainer = containers[i++];
 			if (!autoStack || !isStackable) {
 				//we need to find first empty container as fast as we can for non-stackable items
-				uint32_t n = tmpContainer->capacity() - tmpContainer->size();
-				while (n) {
-					if (tmpContainer->queryAdd(tmpContainer->capacity() - n, *item, item->getItemCount(), flags) == RETURNVALUE_NOERROR) {
-						index = tmpContainer->capacity() - n;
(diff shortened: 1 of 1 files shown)


ericlancebrown (approved):
Code looks fine syntax wise.

Was able to build perfectly fine as well on Ubuntu 20.04 as well as 18.04.

EPuncker (approved):

Dip-Set1:
This will create a few problems.. if player buys item from an NPC and has multiple containers (bp inside bp), even if 2nd bp has empty slots, the items will just delete

EPuncker:
> This will create a few problems.. if player buys item from an NPC and has multiple containers (bp inside bp), even if 2nd bp has empty slots, the items will just delete

I confirm this, items are getting deleted, to reproduce:

1. create two backpacks
2. fill the first one with 19 brass armor (or any other non stackable item)
3. open trade with an npc and buy any item
4. it will be deleted instead of being added to the empty backpack

EPuncker (changes requested):
the issue above need to be fixed

PR #3231, opened 2020-11-27
Performance Boost and Code Cleanup

ericlancebrown:


nekiro:
Describe your changes please.

ericlancebrown:
> Describe your changes please.

Will update asap with a description.

I'm moving today so I am not able.

ericlancebrown:
Changes:

I had included the const std::string& instead of std::string in many of the vars.

Passing as a const reference instead of a value is faster and recommended.

Cleaned up some unnecessary code. Just unused functions and variables that were giving me warnings to fix those.

Nothing too drastic but hopefully will increase performance and readability of the code.

ericlancebrown:
meme

src/iologindata.cpp
```diff
@@ -44,13 +44,6 @@ Account IOLoginData::loadAccount(uint32_t accno)
 	return account;
 }
 
-bool IOLoginData::saveAccount(const Account& acc)
```
nekiro, line 47: why are you removing it?

src/protocolgame.h
```diff
@@ -43,11 +43,11 @@ struct TextMessage
 	MessageClasses type = MESSAGE_STATUS_DEFAULT;
 	std::string text;
 	Position position;
-	uint16_t channelId;
+	uint16_t channelId{};
```
nekiro, line 46: ?

src/script.cpp
```diff
@@ -37,7 +37,7 @@ Scripts::~Scripts()
 	scriptInterface.reInitState();
 }
 
-bool Scripts::loadScripts(std::string folderName, bool isLib, bool reload)
+bool Scripts::loadScripts(const std::string folderName, bool isLib, bool reload)
```
nekiro, line 40: did you miss &?

src/spells.cpp
```diff
@@ -264,16 +264,6 @@ InstantSpell* Spells::getInstantSpell(const std::string& words)
 	return nullptr;
 }
 
-InstantSpell* Spells::getInstantSpellById(uint32_t spellId)
```
nekiro, line 267: why removing it?

src/protocolgame.h
```diff
@@ -43,11 +43,11 @@ struct TextMessage
...
 	struct {
 		int32_t value = 0;
 		TextColor_t color;
-	} primary, secondary;
+	} primary{}, secondary{};
```
ranisalt, line 50: Not needed.

src/iomarket.cpp
```diff
@@ -50,6 +50,7 @@ MarketOfferList IOMarket::getActiveOffers(MarketAction_t action, uint16_t itemId
 		offer.price = result->getNumber<uint32_t>("price");
 		offer.timestamp = result->getNumber<uint32_t>("created") + marketOfferDuration;
 		offer.counter = result->getNumber<uint32_t>("id") & 0xFFFF;
+		offer.itemId = result->getNumber<uint16_t>("itemtype");
```
ranisalt, line 53: Unrelated?

src/iologindata.cpp
ranisalt: Probably, it's not used in the code, both this and the other function he removed.

ranisalt (changes requested):

ranisalt:
> Passing as a const reference instead of a value is faster and recommended.

Note that passing a value instead of a reference does not necessarily imply in an extra copy. Compilers are smart and avoid copying as much as possible, so if you have a value that is not changed in the function body, it most likely generates code for a reference or move anyway.

To have improvements, you should use perfect forwarding (also called universal reference), [see this post](https://stackoverflow.com/a/17644402).

DSpeichert:
@ericlancebrown do you intend to update this PR per reviews?

ericlancebrown:
Nah I have gotten too busy. Sorry.

PR #3237, opened 2020-12-02
Expanded outfit limit size

### LINKED ISSUE #3236: Outfit number limitation

(to reproduce this you will need to have more outfits than the default value, I'll teach with downgraded cip files)

1. transform your original/latest/updated tibia files into spr/dat using this small tutorial [here](https://otland.net/threads/pure-latest-items-otb.268296/) and replace the original ones from your 10.98 client
2. use [this](https://pastebin.com/YMAi0qc6) outfits.xml file
3. start your server, login with your GOD male character
4. open outfit windows and start browsing from citizen outfit onward

### Expected behaviour
Show all available outfits

### Actual behaviour
It will only show up to 99 outfits, anything past that will not be available

### Environment
It happens on both cip client and OTC

EPuncker:
fixes #3236

it was capped to 100, but it has been raised since then.

"they changed it to uint16 in 11.85 version, it was uint8 always before that version"

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2692,7 +2692,7 @@ void ProtocolGame::sendOutfitWindow()
 		}
 
 		protocolOutfits.emplace_back(outfit.name, outfit.lookType, addons);
-		if (protocolOutfits.size() == 255) { // FIXME: Figure out the current limit
+		if (protocolOutfits.size() == 255) { // FIXME: Game client currently doesn't allow more than 255 outfits
 			break;
 		}
 	}
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2692,7 +2692,7 @@ void ProtocolGame::sendOutfitWindow()
 		}
 
 		protocolOutfits.emplace_back(outfit.name, outfit.lookType, addons);
-		if (protocolOutfits.size() == 255) { // FIXME: Game client currently doesn't allow more than 255 outfits
+		if (protocolOutfits.size() == std::numeric_limits<uint8_t>::max()) { // Game client currently doesn't allow more than 255 outfits
 			break;
 		}
 	}
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2692,7 +2692,7 @@ void ProtocolGame::sendOutfitWindow()
 		}
 
 		protocolOutfits.emplace_back(outfit.name, outfit.lookType, addons);
-		if (protocolOutfits.size() == 100) { // Game client doesn't allow more than 100 outfits
+		if (protocolOutfits.size() == 255) { // FIXME: Figure out the current limit
 			break;
 		}
 	}


EPuncker:
I've tested it with 234 outfits and it worked just fine with client 10.98 and OTC

DSpeichert (dismissed):

Zbizu:
Can you check what happens if we go beyond 255 looktypes in set outfit?

EPuncker:
> Can you check what happens if we go beyond 255 looktypes in set outfit? If it doesn't crash the client, we could allow more.
> 
> Also where is that quote from?

gonna test later if possible, the quote is from fabian/sayiansking at his discord

SaiyansKing:
> Can you check what happens if we go beyond 255 looktypes in set outfit? If it doesn't crash the client, we could allow more.
> 
> Also where is that quote from?

You can't fit more than 255 looktypes in Uint8 because it'll overflow and start from 0 again, on otclient it is easily editable to Uint16 but since it isn't a standard the answer is "you can't".

src/protocolgame.cpp
```diff
@@ -2692,7 +2692,7 @@ void ProtocolGame::sendOutfitWindow()
 		}
 
 		protocolOutfits.emplace_back(outfit.name, outfit.lookType, addons);
-		if (protocolOutfits.size() == 100) { // Game client doesn't allow more than 100 outfits
+		if (protocolOutfits.size() == 255) { // FIXME: Figure out the current limit
```
Zbizu, line 2695: So if it's 255, just change the comment back to what it was saying originally, write 255 in it instead and it should be ready to merge.

Zbizu (changes requested):

src/protocolgame.cpp
EPuncker: done

Zbizu:
Looks good to me

Zbizu (approved):

src/protocolgame.cpp
```diff
@@ -2692,7 +2692,7 @@ void ProtocolGame::sendOutfitWindow()
 		}
 
 		protocolOutfits.emplace_back(outfit.name, outfit.lookType, addons);
-		if (protocolOutfits.size() == 100) { // Game client doesn't allow more than 100 outfits
+		if (protocolOutfits.size() == 255) { // FIXME: Game client currently doesn't allow more than 255 outfits
```
DSpeichert, line 2695: I'm not sure what is the action item for the `FIXME` comment.
If the client doesn't support it, there's nothing we can do, right?

EPuncker: actually I left it for future use, because once we support a higher client version it will need to be raised

nekiro: if (protocolOutfits.size() == std::numeric_limits<uint8_t>::max())
more descriptive why the limit is there :)

DSpeichert (dismissed):

nekiro (approved):

Issue #3112, opened 2020-07-01
Friendly warnings on monsters files


EPuncker:
First of all: its the duty of the datapack maker to make sure that their files are fine, but it won't hurt to have a few warnings about wrong values, even if the code already cap them.

so my simple request is: throw a console warning if "chance" is higher than 100 on monster xml file just like the staticattack flag does.

- [ ] basically the 5 occurrences of (uniform_random(1, 100)) **in the monster.cpp file**
- [ ] Monsters::deserializeSpell - chance node **on monsters.cpp**
- [ ] mType->info.changeTargetChance **on monsters.cpp**
- [ ] mType->info.yellChance **on monsters.cpp**
- [ ] summonNode.attribute("chance") **on monsters.cpp**

part 2
- [ ] warn user if monster doesn't have a look type set
- [ ] warn user if monster have the same element on both <elements> and <immunities> tags
- [ ] warn user if an <attack> is missing a set chance (unless it is a melee)

part 3
- [ ] warn user if convinceable and/or summonable is set to 1 but there is no manacost set or its set to 0
- [ ] warn user about duplicate element on immunity and element tags
- [ ] warn user if <'flag targetdistance="0" /'>, minimum should be 1

MillhioreBT:
I agree with you, this does nothing wrong, in fact it encourages them to write the values ​​correctly for everything.

nkzou:
Hi there, I haven't worked on this project before but this seemed like a good first issue. I plan to implement most of these validation checks as I familiarize myself with the codebase.
After reading through some of the docs and code, I'm not sure the 5 occurrences of `(uniform_random(1, 100))` in the `monster.cpp` file need warnings, as all of those values are loaded within `Monsters::loadMonster()` and will be checked already. Seems like duplication of the out of bound checks. Thoughts? 
Also, unsure what the part 3 "warn user if , minimum should be 1" means in the issue description. Does this refer to minimum chance value in all the chance fields? @EPuncker any insight would be great, thanks!

EPuncker:
> Also, unsure what the part 3 "warn user if , minimum should be 1" means in the issue description. Does this refer to minimum chance value in all the chance fields?

**warn user if <'flag targetdistance="0" /'>minimum should be 1** sorry I'm still pretty new to this markdown thing

PR #3239, opened 2020-12-14
warnings on invalid monster data

nkzou:
addresses #3112

I did not implement warnings for the 5 occurrences of `(uniform_random(1, 100))` in the `monster.cpp` because I believe those values get loaded and checked in `monsters.cpp` beforehand, and would be duplicated validation checks. Other than that, I added checks for each item in the list.

I noticed this change results in many warnings raised for the default included datapack. From what I can tell, all the warnings are valid according to the issue linked above.

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -917,7 +917,7 @@ MonsterType* Monsters::loadMonster(const std::string& file, const std::string& m
 				mType->info.light.color = pugi::cast<uint16_t>(attr.value());
 			} else if (strcasecmp(attrName, "targetdistance") == 0) {
 				int32_t targetDistance = pugi::cast<int32_t>(attr.value());
-				if(targetDistance < 1) {
+				if (targetDistance < 1) {
 					targetDistance = 1;
 					std::cout << "[Warning - Monsters::loadMonster] targetdistance less than 1. " << file << std::endl;
 				}
@@ -943,7 +943,7 @@ MonsterType* Monsters::loadMonster(const std::string& file, const std::string& m
 			mType->info.pushable = false;
 		}
 	}
-	if(mType->info.manaCost == 0 && (mType->info.isSummonable || mType->info.isConvinceable)) {
+	if (mType->info.manaCost == 0 && (mType->info.isSummonable || mType->info.isConvinceable)) {
 		std::cout << "[Warning - Monsters::loadMonster] manaCost missing or zero on monster with summonable and/or convinceable flags: " << file << std::endl;
 	}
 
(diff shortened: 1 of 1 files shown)


src/monsters.cpp
```diff
@@ -1183,24 +1204,54 @@ MonsterType* Monsters::loadMonster(const std::string& file, const std::string& m
...
+					std::cout << "[Warning - Monsters::loadMonster] Same element \"drown\" on immunity and element tags. " << file << std::endl;
+				}
 			} else if ((attr = elementNode.attribute("lifedrainPercent"))) {
 				mType->info.elementMap[COMBAT_LIFEDRAIN] = pugi::cast<int32_t>(attr.value());
+				if (mType->info.damageImmunities & COMBAT_LIFEDRAIN){
+					std::cout << "[Warning - Monsters::loadMonster] Same element \"lifedrain\" on immunity and element tags. " << file << std::endl;
+				}
 			} else if ((attr = elementNode.attribute("manadrainPercent"))) {
 				mType->info.elementMap[COMBAT_MANADRAIN] = pugi::cast<int32_t>(attr.value());
+				if (mType->info.damageImmunities & COMBAT_MANADRAIN){
```
DSpeichert, line 1252: style: please keep space between `) {` in conditions

EPuncker:
> I noticed this change results in many warnings raised for the default included datapack. From what I can tell, all the warnings are valid according to the issue linked above.

don't worry I'll make a PR to be merged together with this one

src/monsters.cpp
nkzou: Sure thing. Also, Is there a specific style guide this project follows? I didn't see anything in the wiki but I might've missed it.

src/monsters.cpp
```diff
@@ -943,7 +943,7 @@ MonsterType* Monsters::loadMonster(const std::string& file, const std::string& m
 			mType->info.pushable = false;
 		}
 	}
-	if(mType->info.manaCost == 0 && (mType->info.isSummonable || mType->info.isConvinceable)){
+	if(mType->info.manaCost == 0 && (mType->info.isSummonable || mType->info.isConvinceable)) {
```
EPuncker, line 946: space between **if** and **(**

EPuncker (changes requested):

src/monsters.cpp
DSpeichert: The only style guide is to keep existing style :)

EPuncker (approved):

soul4soul:
I'm not sure about adding a check for `warn user if convinceable and/or summonable is set to 1 but there is no manacost set or its set to 0`. Its possible that 0 manacost could be a desired behavior.  All other warnings that were added are real issues that a developer would want to address. I guess its not a huge deal since it would only amount to a small amount of messages in the console during loading.

EPuncker:
> Its possible that 0 manacost could be a desired behavior.

unless an official monster needs that behavior, I see no point in that. summoning/convincing a monster at no cost should not be allowed unless you have a custom ot, and then you could just edit your source

DSpeichert (approved):

MillhioreBT (approved):
LGTM

PR #3240, opened 2020-12-19
MonsterType Inconsistencies and compatibility

MillhioreBT:
Currently, when creating monsters with "revscripts" there was a problem with invocations, because nowhere was the maximum of them specified, so by default the value was zero and therefore invocations were never generated, now with these changes if they can spawn invocations, as long as a specified maximum amount of the invocation or a global maximum amount of the monster as such is specified.

Change fixed a crazy thing that was in the function that gives lighting to the creature type: D

Also fixed a backward compatibility issue when setting the melee attack type, currently if you don't specify a skill level and attack damage constant then the monster just doesn't deal damage, no matter if you specify an absolute and constant damage, just like it is done in XML files

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -117,8 +117,11 @@ registerMonsterType.flags = function(mtype, mask)
 end
 registerMonsterType.light = function(mtype, mask)
 	if mask.light then
-		if mask.light.color and mask.light.level then
-			mtype:light(mask.light.color, mask.light.level)
+		if mask.light.color then
+			local color = mask.light.color
+		end
+		if mask.light.level then
+			mtype:light(color, mask.light.level)
 		end
 	end
 end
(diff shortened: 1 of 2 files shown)


MillhioreBT:
I apologize for the *reverts* I did, sometimes I get a bit confused with all this unintentionally uwu

data/scripts/lib/register_monster_type.lua
```diff
@@ -247,8 +244,12 @@ registerMonsterType.attacks = function(mtype, mask)
 			if attack.name then
 				if attack.name == "melee" then
 					spell:setType("melee")
-					if attack.attack and attack.skill then
-						spell:setAttackValue(attack.attack, attack.skill)
+					if attack.attack and attack.skill then
```
ranisalt, line 247: Does this check make sense? I'm thinking just the contents of it already do the job. Also, you aren't checking if `attack.minDamage` or `attack.maxDamage` are nil too, so why check for `attack.attack` and `attack.skill`?

MillhioreBT: they can be nil in some cases, more than anything so that when converting xml monsters to lua, they do not suffer from a little known problem, which causes them not to have the same damage that was configured in xml, if no one else provides any proof that this may be of help, then I'll close it in a moment.

PR #3242, opened 2020-12-22
Fixes to invalid monsters data

EPuncker:
follow-up of https://github.com/otland/forgottenserver/commit/4fcb51fb86d6e77150e2279d56c66bcb1c48cb36

now the console should be error free!

also a few small updates

### DIFF
--- a/data/monster/monsters/starving_wolf.xml
+++ b/data/monster/monsters/starving_wolf.xml
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
-<monster name="starving wolf" nameDescription="a starving wolf" race="blood" experience="65" speed="170">
+<monster name="Starving Wolf" nameDescription="a starving wolf" race="blood" experience="65" speed="188">
 	<health now="85" max="85" />
-	<look type="27" head="20" body="30" legs="40" feet="50" corpse="5968" />
+	<look type="27" corpse="5968" />
 	<targetchange interval="4000" chance="0" />
 	<flags>
-		<flag summonable="1" />
+		<flag summonable="0" />
 		<flag attackable="1" />
 		<flag hostile="1" />
 		<flag illusionable="1" />
-		<flag convinceable="1" />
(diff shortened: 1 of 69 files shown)


data/monster/monsters/starving_wolf.xml
```diff
@@ -36,7 +38,7 @@
 		<voice sentence="Yoooohhuuuu!" />
 	</voices>
 	<loot>
-		<item id="220" chance="64730" /><!-- dirty fur -->
```
EPuncker, line 39: @DSpeichert @EvilHero90  do you have any idea on why this didn't popped an error btw? item id 220 doesn't exist (it was a typo), but server didn't noticed this even with forceMonsterTypesOnLoad enabled

WibbenZ: https://github.com/otland/forgottenserver/blob/4fcb51fb86d6e77150e2279d56c66bcb1c48cb36/src/monsters.cpp#L1354

There is no check if an item with that id exists compared to one with a string / name.

https://github.com/otland/forgottenserver/blob/4fcb51fb86d6e77150e2279d56c66bcb1c48cb36/src/monsters.cpp#L1359-L1362

EPuncker: oh makes sense, hope someone could add it

MillhioreBT: @EPuncker you can add the warning as follows:
```c++
if ((attr = node.attribute("id"))) {
		int32_t id = pugi::cast<int32_t>(attr.value());
		const ItemType& it = Item::items.getItemType(id);
		if (it.name.empty()) {
			std::cout << "[Warning - Monsters::loadMonster] Unknown loot item id \"" << id << "\". " << std::endl;
			return false;
		}

		lootBlock.id = id;

	} else if ((attr = node.attribute("name"))) {
```
I already tried it and it seems to work perfectly.

DSpeichert (approved):

PR #3250, opened 2020-12-25
Monsters shouldn't see floors above 8 

### LINKED ISSUE #3246: Monsters not being idle when they should

1. Open two clients
2. GM character floor 8 (monsters idle)
3. Then login normal character on the floor above

### Expected behaviour
Monsters on floor 8 should not stay active (non-idle) while there are players on the above floor
(higher floors should remain as they are, current issue is only between floor 7 and 8)

### Actual behaviour
Monsters become active when a player out of their sight is detected

### Disclaimer
Yes, I know that this is not true in other cases, but specially on floor 8 (-1) monsters should not have their idle status changed because of what happens on ground floor (0), I've discussed it with other devs and even tested it on RL tibia. Floor 7 (ground) is special and has no interaction with lower floors.

nekiro:
Fix #3246 

If monster is on floor 8 and player is on floor 7, then monster shouldnt be able to see him and should go idle instead.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -61,6 +61,11 @@ bool Creature::canSee(const Position& myPos, const Position& pos, int32_t viewRa
 		}
 	} else if (myPos.z >= 8) {
 		//we are underground (8 -> 15)
+		// we can't see floors above 8
+		if (pos.z < 8) {
+			return false;
+		}
+
 		//view is +/- 2 from the floor we stand on
 		if (Position::getDistanceZ(myPos, pos) > 2) {
 			return false;


EPuncker (approved):

MillhioreBT (approved):

Zbizu:
doesn't that make roping far easier on retro servers?

EPuncker:
> doesn't that make roping far too easy on retro servers?

only on floor 7 to 8, I don't think this is a "big issue" since generally speaking there is no strong monster on the first underground floor, and the weak ones you could just diagonal parcel trap them to rope them

marmichalski (approved):

soul4soul:
I'm fairly certain this isn't how cipbia behaves monster should move on floor 8 when a player is on floor 7. Lure a bunch of monsters around a hole and go up, they will start moving and running away. They don't stand waiting around the ladder/rope spot.  It can be tested easily on a low pop world in rook since there won't be other players around to mess up the experiment by being on a different level.

The excuse that only weak monsters are on floor 8 is a poor one and not relevant to the actual behavior. Draconia has dragons on floor 8. OF has orc warlords on floor 8. Plus there are many places where monsters can be roped multiple levels (or at least was true in the past) ie. POH dragon lords.

nekiro:
> I'm fairly certain this isn't how cipbia monsters behave. Monster should move on floor 8 when a player is on floor 7. Lure a bunch of monsters around a hole and go up, they will start moving and running away. They don't stand waiting around the ladder/rope spot. It can be tested easily on a low pop world in rook since there won't be other players around to mess up the experiment by being on a different level.
> 
> The excuse that only weak monsters are on floor 8 is a poor one and not relevant to the actual behavior. Draconia has dragons on floor 8. OF has orc warlords on floor 8. Plus there are many places where monsters can be roped multiple levels (or at least was true in the past) ie. POH dragon lords.

Epuncker tested this behaviour on cip tibia and from his tests it appeared that creatures aren't moving.
@EPuncker can we get more info about this?

EPuncker:
> I'm fairly certain this isn't how cipbia monsters behave. Monster should move on floor 8 when a player is on floor 7. Lure a bunch of monsters around a hole and go up, they will start moving and running away. They don't stand waiting around the ladder/rope spot. It can be tested easily on a low pop world in rook since there won't be other players around to mess up the experiment by being on a different level.
> 
> The excuse that only weak monsters are on floor 8 is a poor one and not relevant to the actual behavior. Draconia has dragons on floor 8. OF has orc warlords on floor 8. Plus there are many places where monsters can be roped multiple levels (or at least was true in the past) ie. POH dragon lords.

of course they will keep moving because they will move back to their master position since they were lured as you stated, I did test on tarantulas south of thais, put lots of parcel at the cave and no, they didn't move a single one until I went down the hole, if you really have any proof feel free to provide

PR #3251, opened 2020-12-26
Dont logout players on death

nekiro:
Players shouldn't get logged out on death.

Would be great if this could get thoroughly tested.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -682,7 +682,7 @@ void Creature::onDeath()
 		setMaster(nullptr);
 	}
 
-	if (droppedCorpse) {
+	if (droppedCorpse && !getPlayer()) {
 		g_game.removeCreature(this, false);
 	}
 }
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -563,10 +563,18 @@ bool Game::placeCreature(Creature* creature, const Position& pos, bool extendedP
 	return true;
 }
 
-bool Game::removeCreature(Creature* creature, bool isLogout/* = true*/)
+bool Game::removeCreature(Creature* creature, bool isLogout/* = true*/, bool onlyRelease /* = false */)
 {
 	if (creature->isRemoved()) {
 		return false;
+	}
+
+	if (onlyRelease) {
+		creature->removeList();
+		creature->setRemoved();
+		ReleaseCreature(creature);
+		removeCreatureCheck(creature);
+		return true;
 	}
 
 	Tile* tile = creature->getTile();
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2081,6 +2081,88 @@ void Player::death(Creature* lastHitCreature)
 		onIdleStatus();
 		sendStats();
 	}
+
+	despawn();
+}
+
+bool Player::spawn()
+{
+	setDead(false);
+
+	if (!g_game.map.placeCreature(getLoginPosition(), this, false, true)) {
+		return false;
+	}
+
+	const Position& pos = getPosition();
+
+	SpectatorVec spectators;
+	g_game.map.getSpectators(spectators, pos, true);
+	for (Creature* spectator : spectators) {
+		if (Player* tmpPlayer = spectator->getPlayer()) {
+			tmpPlayer->sendCreatureAppear(this, pos, false);
+		}
+	}
+
+	for (Creature* spectator : spectators) {
+		spectator->onCreatureAppear(this, false);
+	}
+
+	getParent()->postAddNotification(this, nullptr, 0);
+	g_game.addCreatureCheck(this);
+
+	for (const auto& it : g_game.getPlayers()) {
+		it.second->notifyStatusChange(this, VIPSTATUS_ONLINE);
+	}
+	return true;
+}
+
+void Player::despawn()
+{
+	if (isDead()) {
+		return;
+	}
+
+	// remove check
+	g_game.removeCreatureCheck(this);
+
+	// remove from map
+	Tile* tile = getTile();
+
+	std::vector<int32_t> oldStackPosVector;
+
+	SpectatorVec spectators;
+	g_game.map.getSpectators(spectators, tile->getPosition(), true);
+	for (Creature* spectator : spectators) {
+		if (Player* player = spectator->getPlayer()) {
+			oldStackPosVector.push_back(player->canSeeCreature(this) ? tile->getStackposOfCreature(player, this) : -1);
+		}
+	}
+
+	tile->removeCreature(this);
+
+	const Position& tilePosition = tile->getPosition();
+
+	//send to client
+	size_t i = 0;
+	for (Creature* spectator : spectators) {
+		if (Player* player = spectator->getPlayer()) {
+			player->sendRemoveTileThing(tilePosition, oldStackPosVector[i++]);
+		}
+	}
+
+	//event method
+	for (Creature* spectator : spectators) {
+		spectator->onRemoveCreature(this, false);
+	}
+
+	getParent()->postRemoveNotification(this, nullptr, 0);
+
+	// show player as pending
+	for (const auto& it : g_game.getPlayers()) {
+		it.second->notifyStatusChange(this, VIPSTATUS_PENDING);
+	}
+
+	setDead(true);
 }
 
 bool Player::dropCorpse(Creature* lastHitCreature, Creature* mostDamageCreature, bool lastHitUnjustified, bool mostDamageUnjustified)
--- a/src/player.h
+++ b/src/player.h
@@ -1148,6 +1148,13 @@ class Player final : public Creature, public Cylinder
 		void forgetInstantSpell(const std::string& spellName);
 		bool hasLearnedInstantSpell(const std::string& spellName) const;
 
+		void setDead(bool isDead) {
+			dead = isDead;
+		}
+		bool isDead() const {
+			return dead;
+		}
+
 	private:
 		std::forward_list<Condition*> getMuteConditions() const;
 
@@ -1165,6 +1172,8 @@ class Player final : public Creature, public Cylinder
 		void setNextActionTask(SchedulerTask* task, bool resetIdleTime = true);
 
 		void death(Creature* lastHitCreature) override;
+		bool spawn();
+		void despawn();
 		bool dropCorpse(Creature* lastHitCreature, Creature* mostDamageCreature, bool lastHitUnjustified, bool mostDamageUnjustified) override;
 		Item* getCorpse(Creature* lastHitCreature, Creature* mostDamageCreature) override;
 
@@ -1303,6 +1312,7 @@ class Player final : public Creature, public Cylinder
 		fightMode_t fightMode = FIGHTMODE_ATTACK;
 		AccountType_t accountType = ACCOUNT_TYPE_NORMAL;
 
+		bool dead = false;
 		bool chaseMode = false;
 		bool secureMode = false;
 		bool inMarket = false;
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -394,7 +394,7 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
 
 	uint8_t recvbyte = msg.getByte();
 
-	if (!player) {
+	if (!player || player->isRemoved()) {
 		if (recvbyte == 0x0F) {
 			disconnect();
 		}
@@ -403,13 +403,25 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
 	}
 
 	//a dead player can not performs actions
-	if (player->isRemoved() || player->getHealth() <= 0) {
-		if (recvbyte == 0x0F) {
+	if (player->isDead() || player->getHealth() <= 0) {
+		if (recvbyte == 0x14) {
 			disconnect();
+			g_game.removeCreature(player, false, true);
 			return;
 		}
 
-		if (recvbyte != 0x14) {
+		if (recvbyte == 0x0F) {
+			if (!player->spawn()) {
+				disconnect();
+				g_game.removeCreature(player, false, true);
+				return;
+			}
+
+			sendAddCreature(player, player->getPosition(), 0, false);
+		}
+
+		if (recvbyte != 0x1D && recvbyte != 0x1E) {
+			// keep connection alive
 			return;
 		}
 	}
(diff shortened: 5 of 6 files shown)


MillhioreBT:
I like it, I'm going to try it

MillhioreBT:
Dont work and produce the following error:
image
this happens if you decide to log out instead of continuing, and if you continue, another violation occurs in any call to lua call protection function

nekiro:
> Dont work and produce the following error:
> image
> this happens if you decide to log out instead of continuing, and if you continue, another violation occurs in any call to lua call protection function:
> image

Didn't happen to me on linux, didn't test on windows.
But I see one possible reason for this. Check if it's better with new commit.

MillhioreBT:
The first error no longer occurs, but the second continues to appear every time I die with the character, it seems as if from lua it was trying to read a dead pointer or something like that, I will try to compile it with the base files instead of those from my sources to rule out possible errors in my sources

MillhioreBT:
I give up, I did more than 50 tests, in half the cases it seemed to work fine, but when reappearing, you would get guild shields and sometimes some npc icons, some actions such as seeing objects cause the crash, it is quite strange, and in the other half the server it just crashed when trying to move the character from its spawn location

nekiro:
> I give up, I did more than 50 tests, in half the cases it seemed to work fine, but when reappearing, you would get guild shields and sometimes some npc icons, some actions such as seeing objects cause the crash, it is quite strange, and in the other half the server it just crashed when trying to move the character from its spawn location.
> 
> We hope someone else can check this, the truth is I feel sad that I can not have this functionality at first sight

thats very odd. I tested with moving and generally still playing after dying and it didn't crash even once...

MillhioreBT:
> > I give up, I did more than 50 tests, in half the cases it seemed to work fine, but when reappearing, you would get guild shields and sometimes some npc icons, some actions such as seeing objects cause the crash, it is quite strange, and in the other half the server it just crashed when trying to move the character from its spawn location.
> > We hope someone else can check this, the truth is I feel sad that I can not have this functionality at first sight
> 
> That's very odd. I tested with moving and generally still playing after dying and it didn't crash even once...
> It must be your custom source.

If it is surely that, since my sources support versions higher than 10.98, I should probably add something else to the changes you made so that then it can also work for me, in the meantime I will wait, I really don't have much idea how to make it work for me

MillhioreBT:
This would be the first PR that does not work for me in my custom fonts, it is quite rare, I hope that other people can try this and confirm if everything is fine or if they are also doing just as badly as me

my problem is related to the light stack, in case someone else gives them an error and they relate it to the lua stack!

nekiro:
> This would be the first PR that does not work for me in my custom fonts, it is quite rare, I hope that other people can try this and confirm if everything is fine or if they are also doing just as badly as me
> 
> my problem is related to the lua stack, in case someone else gives them an error and they relate it to the lua stack!

It's probably related to your onDeath or onPrepareDeath events or maybe onThink events?

MillhioreBT:
> > This would be the first PR that does not work for me in my custom fonts, it is quite rare, I hope that other people can try this and confirm if everything is fine or if they are also doing just as badly as me
> > my problem is related to the lua stack, in case someone else gives them an error and they relate it to the lua stack!
> 
> It's probably related to your onDeath or onPrepareDeath events or maybe onThink events?
> If you could debug this and find the issue, it would be great. This might be issue with something you use (in lua) while tfs does not.

i opened my server in debug mode, and put a breakpoint in the function here:
image
Then I started doing the tests, and indeed at this point the process was interrupted twice, it was enough to give it twice to continue and then everything happened according to your changes, everything perfect, then it disable the breakpoint and then the error happened , and it was always related to the lua stack, it seems like some lua script is running on a different thread and thus not syncing.

I'll try to keep checking this when I can, because I really want this, and I feel sad that I can't add this PR to my custom sources now

src/protocolgame.cpp
```diff
@@ -403,13 +403,31 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
...
-	if (player->isRemoved() || player->getHealth() <= 0) {
+	if (player->isDead() || player->getHealth() <= 0) {
+		if (recvbyte == 0x14) {
+			disconnect();
+			addGameTask(&Game::removeCreature, player, false, true);
+			return;
+		}
+
 		if (recvbyte == 0x0F) {
+			if (!player->spawn()) {
```
kygov, line 414: Should not this method be executed on Dispatcher thread? like all other methods which interact with game?
I would move whole logic to Game and then execute with dispatcher maybe.

slawkens:
I can't test but I have a question.

Will it fire onLogout after player click logout in client after death? Or how does it work? Because AFAIK before it didn't fire onLogout after death.

src/protocolgame.cpp
nekiro: yeah, it should probably, hence the problems with lua stack from @MillhioreBT.
I will make new method for this in game class.

nekiro:
> I can't test but I have a question.
> 
> Will it fire onLogout after player click logout in client after death? Or how does it work? Because AFAIK before it didn't fire onLogout after death.

No it wont, because you aren't logging out. It will fire onLogout when you really logout, so when you are getting kicked for inactivity or when you decide to click cancel and not "OK" in the prompt.

MillhioreBT:
With the new modifications you made, the problem was solved by moving the events to the dispatcher thread
**_LGTM_**

MillhioreBT (dismissed):
With the new modifications you made, the problem was solved by moving the events to the dispatcher thread
**_LGTM_**

Diego-OT:
Try it:
Die -> click to open Store (it will dont open nothing) -> make some right-clicks in the map or move with your keyboard
I can't reproduce that 100% but if I keep trying, it will crash after few attempts

nekiro:
> Try it:
> Die -> click to open Store (it will dont open nothing in case) -> make some right-clicks in the map or move with your keyboard
> I can't reproduce that 100% but if I keep trying, it will crash after few attempts

Clicking "store" should probably kick out or respawn, cause it's not used atm, so clicking might have some unintended behaviour, good catch

nekiro:
So I checked it and clicking store actually acts like clicking OK. So not sure what are you talking about.

MillhioreBT:
The store button works for me, in my case I use client 10
to this day it still works wonderfully

Diego-OT:
I reviewed the changes and it's working fine (with store), I was confused the isRemoved() function with isDead() (I was  sleepy) - sorry for the bad feedback

src/player.cpp
```diff
@@ -2081,6 +2081,90 @@ void Player::death(Creature* lastHitCreature)
...
+	return true;
+}
+
+void Player::despawn()
+{
+	if (isDead()) {
+		return;
+	}
+
+	sendCancelWalk();
```
MillhioreBT, line 2123: With `sendCancelWalk()` this no longer happens:
GIF 09-01-2021 03-09-38 p  m
...
BUT

you need to add this:
```c++
onWalkComplete();
stopEventWalk();
```

otherwise the character will not be able to move when he reappears.

MillhioreBT (changes requested):

src/player.cpp
nekiro: This was tested and is working without any additional code. When you die while walking, your last step is automatically cancelled by server, sendCancelWalk is just there to sync client with current position.

MillhioreBT: > This was tested and is working without any additional code. When you die while walking, your last step is automatically cancelled by server, sendCancelWalk is just there to sync client with current position.

Have you tried it on windows?
I got an error even with `sendCancelWalk` once the player appeared, it could not move, I did the tests by sending the player to a far away place with a click, and in the middle of the way it made him die, and then that happened

nekiro: You may be right, calling the method to clean auto walk directions shouldn't change much and the dir vector should be cleared, thanks.

src/player.cpp
```diff
@@ -2120,7 +2120,10 @@ void Player::despawn()
 		return;
 	}
 
-	sendCancelWalk();
+	if (!listWalkDir.empty()) {
```
MillhioreBT, line 2123: This does not change anything, the same situation continues to occur in which the character becomes immobile due to the desynchronization of the client with the server.

I just realized that you only need to add the function: `stopEventWalk();` below `sendCancelWalk();` and this question is solved.

the problem lies here:
```c++
if (eventWalk != 0) {
		return;
	}
```
By basic rules of death, this event must be executed again as if it were your first step, do not forget that even if the player is not eliminated, he is actually dying.

MillhioreBT (changes requested):

src/player.cpp
nekiro: https://github.com/otland/forgottenserver/blob/a540e72a10f68cbd0a2e04e489652db4f9ce7902/src/creature.cpp#L294

MillhioreBT: > https://github.com/otland/forgottenserver/blob/a540e72a10f68cbd0a2e04e489652db4f9ce7902/src/creature.cpp#L292-L298
> 
> Honestly I have no idea what code snippet you posted.

sorry, in the function: `void Creature::addEventWalk(bool firstStep)`

nekiro: So what does addEventWalk has to do with the problem here?

MillhioreBT: > So what does addEventWalk has to do with the problem here?

<details>
<summary>Image</summary>

image

</details>

nekiro: Sorry, but I dont see the connection here. addEventWalk is called **each step** when player is auto walking. Player is auto walking basing on the listWalkDir container, so clearing listWalkDir and aborting the walk, cancels the auto walk.

https://github.com/otland/forgottenserver/blob/a540e72a10f68cbd0a2e04e489652db4f9ce7902/src/creature.cpp#L194

MillhioreBT: :v for this:
image

I'm not really sure, but apparently the reference is duplicated, or it may even lack it

nekiro: Yeah, calling stopEventWalk directly fixes it, but honestly I have no idea why xD

MillhioreBT: > Yeah, calling stopEventWalk directly fixes it, but honestly I have no idea why xD
> Maybe the one additional eventWalk execution is breaking the client.
> 
> It's not redundant code, cause this method is only called when creature is following or auto walking. I bet it's kind of bug fix, but if you sure and you looked up all the traces, then you can push the change :D

You are right, this function is sometimes executed in the dispatcher thread, and it may happen that this list is filled with directions from the main thread or others...

MillhioreBT:
Works great, I'll be testing all night, then give my aprove 😍

MillhioreBT (approved):
I don't know if I have tried all the possible situations in which this could cause an error, but I have not found any, everything seems to work fine for me, LGTM

EPuncker:
@Diego-OT could you check again after the latest changes please?

MillhioreBT:
@nekiro new problem to deal with, when you die your ID is deleted from the table: `nextUseStaminaTime` and therefore it will produce an error when receiving experience with the stamina option activated.

src/protocolgame.cpp
```diff
@@ -196,6 +196,36 @@ void ProtocolGame::connect(uint32_t playerId, OperatingSystem_t operatingSystem)
...
+
+void ProtocolGame::despawn()
+{
+	//dispatcher thread
+	if (!player) {
+		return;
+	}
+
+	//fire onLogout
+	g_creatureEvents->playerLogout(player);
```
Zbizu, line 223: onLogout shouldn't be executed when dying

this is how current tfs behaves

Zbizu (changes requested):

Zbizu:
needs rebase

Any progress on this?

src/creature.cpp
```diff
@@ -79,6 +79,10 @@ bool Creature::canSeeCreature(const Creature* creature) const
 	if (!canSeeInvisibility() && creature->isInvisible()) {
 		return false;
 	}
+
+	if (creature->isDead()) {
+		return false;
+	}
```
ramon-bernardo, line 85: Move it to first, avoiding call unnecessary functions

EPuncker (approved):

src/game.cpp
```diff
@@ -623,6 +631,82 @@ void Game::executeDeath(uint32_t creatureId)
...
+}
+
+void Game::spawnPlayer(uint32_t creatureId)
+{
+	Player* player = getPlayerByID(creatureId);
+	if (!player || !player->isDead()) {
+		return;
+	}
+
+	const Position& pos = player->getLoginPosition();
```
ramon-bernardo, line 687: position instead of pos

nekiro: even better, i'll name it loginPos

src/game.cpp
```diff
@@ -623,6 +631,82 @@ void Game::executeDeath(uint32_t creatureId)
 	}
 }
 
+void Game::despawnPlayer(uint32_t creatureId)
```
ramon-bernardo, line 634: Maybe playerId? same in spawn

EPuncker (approved):

src/player.h
```diff
@@ -1166,6 +1166,10 @@ class Player final : public Creature, public Cylinder
 		void forgetInstantSpell(const std::string& spellName);
 		bool hasLearnedInstantSpell(const std::string& spellName) const;
 
+		uint64_t getLastPong() const {
+			return lastPong;
+		}
```
ramon-bernardo, line 1171:
```
player.h
int64_t lastPong;
```

nekiro: lastPing and lastPong should be stored in u64, so this should be changed, but thats another pr

ranisalt: > ~u64~

`std::chrono::time_point`

ramon-bernardo (approved):

Diego-OT:
I tested and: 

`1- Global when player dies, the player goes to the temple instantly, so when we exiva, the exiva will tell us that he is in the temple. Currently the exiva points to the location where the player died
2- Crash: Die with a normal player and do nothing. With GM use the command /c player
3- Player never logout if no button is pressed (not even after 60 seconds) due to sent/receive ping. Also, the byte 0x14 is only called at the time of death and therefore impossible to check after 60 seconds in the case. The bytes that keep sending are 0x1D 0x1C 0x1E`

Felipe21:
is this buggy or does works?

Issue #3084, opened 2020-06-09
Refactor current fluid system


EPuncker:
Currently both TFS and OTC still lays into using the old fluid system where the type were simply repeated every xpatterns*ypatterns, ([code here](https://github.com/otland/otclient/blob/master/src/client/item.cpp#L307-L379)), but since tibia is adding new fluid types (and so we can add custom ones as well), we need to re-do our code, because the client expects fluid types to be in order from 0 to 17/18(ink)

![2020-06-09 11_54_24](https://user-images.githubusercontent.com/4684880/84163567-036c0700-aa48-11ea-9c83-f17632033d6f.png)

Zbizu:
I think that we also need to get rid of fluids from items.xml.

nekiro:
@EPuncker can you write a list of current tibia fluids that are relevant to our current tfs in order, with correct numbers for each fluid type that match client fluid type?

nekiro:
I don't really understand all this pattern part, how should it be or how does vanilla handle it then? Fluids are ordered from 0 to 17

Zbizu:
Why are fluids tied to items.xml anyway? imo they should be in separate file or even moved to lua.

nekiro:
> Why are fluids tied to items.xml anyway? imo they should be in separate file or even moved to lua.
> 
> @nekiro fluid holders have subid, it's similar to old rune charges and item count. We don't really know what is being sent. It's probably determined by spr/dat or in the client code itself.

they arent, server is only taking the name from items.xml.

EPuncker:
this was a mistake, our code is "fine"

PR #3252, opened 2020-12-26
Use ids under 100 for fluid names

nekiro:
There is no point to use ids after 30000 (which are hardcoded) if we can use numbers from 0 to 100, like vanilla is doing it :)

This is partial fix to
#3084

### DIFF
--- a/src/items.cpp
+++ b/src/items.cpp
@@ -541,12 +541,7 @@ void Items::buildInventoryList()
 
 void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 {
-	if (id > 30000 && id < 30100) {
-		id -= 30000;
-
-		if (id >= items.size()) {
-			items.resize(id + 1);
-		}
+	if (id > 0 && id < 100) {
 		ItemType& iType = items[id];
 		iType.id = id;
 	}
--- a/data/items/items.xml
+++ b/data/items/items.xml
@@ -1,5 +1,25 @@
 <?xml version="1.0" encoding="iso-8859-1"?>
 <items>
+	<item id="1" name="water" />
+	<item id="2" name="blood" />
+	<item id="3" name="beer" />
+	<item id="4" name="slime" />
+	<item id="5" name="lemonade" />
+	<item id="6" name="milk" />
+	<item id="7" name="manafluid" />
+	<item id="10" name="lifefluid" />
+	<item id="11" name="oil" />
+	<item id="13" name="urine" />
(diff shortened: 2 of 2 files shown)


src/items.cpp
```diff
@@ -541,12 +541,7 @@ void Items::buildInventoryList()
 
 void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 {
-	if (id > 30000 && id < 30100) {
-		id -= 30000;
-
-		if (id >= items.size()) {
-			items.resize(id + 1);
-		}
+	if (id > 0 && id < 100) {
```
marmichalski, line 544: id is unsigned 🤔

nekiro: so what?

MillhioreBT (approved):
Tested and works correctly for me!

src/items.cpp
marmichalski: Sorry, had an impression that id 0 is considered invalid and filtered. This happens for  it.id just few lines below. Anyway, I wonder if moving it just after `it.id == 0` block wouldn't be correct? Then the `id > 0` would actually become redundant.

marmichalski (approved):

MillhioreBT:
@nekiro these changes take into account this function:
```c++
const ItemType& Items::getItemIdByClientId(uint16_t spriteId) const
{
	if (spriteId >= 100) {
		if (uint16_t serverId = clientIdToServerIdMap.getServerId(spriteId)) {
			return getItemType(serverId);
		}
	}
	return items.front();
}
```
or should it not affect at all?
I haven't noticed anything unusual, but I don't know exactly how to test this either.

nekiro:
> @nekiro these changes take into account this function:
> 
> ```c++
> 
> const ItemType& Items::getItemIdByClientId(uint16_t spriteId) const
> 
> {
> 
> 	if (spriteId >= 100) {
> 
> 		if (uint16_t serverId = clientIdToServerIdMap.getServerId(spriteId)) {
> 
> 			return getItemType(serverId);
> 
> 		}
> 
> 	}
> 
> 	return items.front();
> 
> }
> 
> ```
> 
> or should it not affect at all?
> 
> I haven't noticed anything unusual, but I don't know exactly how to test this either.

SpriteId is clientId not serverId and either way this is only used in description and taken directly from Items static container.

MillhioreBT:
@nekiro then perfectly good!

EvilHero90 (approved):
looks good to me

PR #3253, opened 2020-12-26
Fixed column sizes to match their variable counterpart

EPuncker:
To match its variable that is uint64_t, and to prevent **Out of range value** errors


MillhioreBT (approved):

DSpeichert:
That should be a migration - is there one already?

EPuncker:
> That should be a migration - is there one already?

added

nekiro (approved):

MillhioreBT (dismissed):

MillhioreBT (approved):
Work for me!

nekiro (approved):

WibbenZ (approved):

PR #3256, opened 2020-12-31
Updated snow.lua check

EPuncker:
It should work only on players now.

https://www.youtube.com/watch?v=4FwX-Gbqsbk

### DIFF
--- a/data/movements/scripts/snow.lua
+++ b/data/movements/scripts/snow.lua
@@ -1,6 +1,6 @@
 function onStepOut(creature, item, position, fromPosition)
 	if not creature:isPlayer() or creature:isInGhostMode() then
-		return false
+		return true
 	end
 
 	if item:getId() == 670 then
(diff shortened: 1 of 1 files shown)


nekiro (approved):

data/movements/scripts/snow.lua
```diff
@@ -1,6 +1,6 @@
 function onStepOut(creature, item, position, fromPosition)
-	if creature:isPlayer() and creature:isInGhostMode() then
-		return true
+	if not creature:isPlayer() or creature:isInGhostMode() then
+		return false
```
marmichalski, line 3: It seems like the return value of onStep* functions does not matter, but when it would, then you would be blocking moving out of a tile by returning `false` here. 🤔

IMHO it should still be returning `true` or not return anything at all. Just something to think about.

peonso: I agree, there is no point to return false even if it "works".

DSpeichert (changes requested):
Like others pointed out - is there a reason it returns false?

marmichalski (approved):

peonso:
LGTM

ranisalt (approved):

DSpeichert (approved):

PR #3259, opened 2021-01-03
fixed EventCallback, gain/lose_exp and skilltries

MillhioreBT:
Some events require that one of their parameters be kept up to date with the previous results to work correctly, here is an example:

Current:
```lua
local ev = EventCallback

function ev.onGainExperience(player, source, exp, rawExp)
	exp = exp + 1000
	return exp
end

function ev.onGainExperience(player, source, exp, rawExp)
	exp = exp + 1000
	return exp
end

-- RESULT: exp + 1000
```

Now with this changes:
```lua
local ev = EventCallback

function ev.onGainExperience(player, source, exp, rawExp)
	exp = exp + 1000
	return exp
end

function ev.onGainExperience(player, source, exp, rawExp)
	exp = exp + 1000
	return exp
end

-- RESULT: exp + 2000
```

@EvilHero90 check pls

The code is a bit generalized with tables that only contain a single value, I did it with that intention in case in the future there are new events that require updating several parameters, I do not know if it is right

### DIFF
--- a/data/scripts/lib/event_callbacks.lua
+++ b/data/scripts/lib/event_callbacks.lua
@@ -70,6 +70,12 @@ local callbacks = {
 	["onSpawn"] = EVENT_CALLBACK_ONSPAWN
 }
 
+local feedbackParams = {
+	[EVENT_CALLBACK_ONGAINEXPERIENCE] = {3},
+	[EVENT_CALLBACK_ONLOSEEXPERIENCE] = {2},
+	[EVENT_CALLBACK_ONGAINSKILLTRIES] = {3}
+}
+
 -- can't be overwritten on /reload global/libs now
 if not EventCallbackData then
 	EventCallbackData = {}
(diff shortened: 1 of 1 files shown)


nekiro:
Why would we need this?
Events aren't meant to be duplicated like creaturescripts interface.

MillhioreBT:
> Why would we need this?
> Events aren't meant to be duplicated like creaturescripts interface.

I see the potential of being able to create everything in a single file, absolutely everything, so that we can create complete events in a single file, probably this will not be submerged in this repository, since it is designed to emulate Tibia RL.

If the events should not be duplicated then why does the interface allow it? currently gives the feeling that something like this can be done

An example of use would be the following:
- A person comes to the forum asking how to make his server every certain hour make the experience gained double, so instead of explaining that he must implement changes in some files and in a correct order, we can simply tell him to paste a file in the `scripts` folder and then everything will be working.

MillhioreBT:
I am going to close this compromise due to what this functionality you need the version of lua 5.3+ in order to have a consistent behavior in all possible cases, this is due to a problem with the `{}` operator or to be more specific in the function `lua_objlen`, so some extra changes are required, which I don't think they are willing to accept.

PR #3260, opened 2021-01-03
Handle exception when map file does not exist

### LINKED ISSUE #3254: Server closes on non-existing map name

1. Change default mapName in config.lua to something that doesn't exist

### Expected behaviour
Console should give at least some info what have gone wrong. For example: couldn't find map file.

### Actual behaviour
The server closes unexpectedly without giving any info. Screenshot included:
image

nekiro:
Fix #3254

### DIFF
--- a/src/iomap.cpp
+++ b/src/iomap.cpp
@@ -69,85 +69,90 @@ Tile* IOMap::createTile(Item*& ground, Item* item, uint16_t x, uint16_t y, uint8
 bool IOMap::loadMap(Map* map, const std::string& fileName)
 {
 	int64_t start = OTSYS_TIME();
-	OTB::Loader loader{fileName, OTB::Identifier{{'O', 'T', 'B', 'M'}}};
-	auto& root = loader.parseTree();
+	try {
+		OTB::Loader loader{fileName, OTB::Identifier{{'O', 'T', 'B', 'M'}}};
+		auto& root = loader.parseTree();
 
-	PropStream propStream;
-	if (!loader.getProps(root, propStream)) {
-		setLastErrorString("Could not read root property.");
-		return false;
-	}
-
-	OTBM_root_header root_header;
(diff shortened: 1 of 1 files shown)


src/iomap.cpp
```diff
@@ -69,85 +69,90 @@ Tile* IOMap::createTile(Item*& ground, Item* item, uint16_t x, uint16_t y, uint8
 bool IOMap::loadMap(Map* map, const std::string& fileName)
 {
 	int64_t start = OTSYS_TIME();
-	OTB::Loader loader{fileName, OTB::Identifier{{'O', 'T', 'B', 'M'}}};
-	auto& root = loader.parseTree();
+	try {
```
djarek, line 72: I don't like the scope of this `try` block, it catches a lot more errors than it should, e.g. if something throws during tile parsing, we may end up in an inconsistent state.

nekiro: Well I wanted to catch only the constructor, but I couldn't because of the variable being enclosed in try catch block, thus rest of the code wont proceed. I could probably use pointer for this though.

djarek: Hmm, this is actually a tough one, because parsing props may still cause us to end up in a partially-loaded state. I usually prefer to make such operations "transactional" (either succeeds completely or has no effect), but I guess it's way out of scope of this PR. 

I think it's sufficient to just change the type you accept in `catch` to `catch (const InvalidOTBFormat& err)`

nekiro: Honestly gonna sound like a noob here, but I tried but couldn't make it to work. No matter what I did it wouldn't find the custom exception type. I included fileloader.h on the top and still wasn't able to find it somehow...

djarek: `const OTB::LoadError&` should work just fine.

djarek (approved):
LGTM

PR #3263, opened 2021-01-05
Quiver systems

Underewarrr:
Hello. Inspired @xydius solution I've created possibility to make use for quivers for both TFS
 (1.X based on otland/forgottenserver (https://github.com/otland/forgottenserver)
 and 0.X on Fir3element/3777 (https://github.com/Fir3element/3777)).

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -1238,7 +1238,7 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
 		uint32_t volume = 0;
 		if (!item || !item->hasAttribute(ITEM_ATTRIBUTE_UNIQUEID)) {
 			if (it.isContainer()) {
-				volume = it.maxItems;
+				volume = item->getName() == "Quiver" ? 4 : it.maxItems;
 			} else {
 				volume = item->getContainer()->capacity();
 			}
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -288,6 +288,10 @@ ReturnValue Container::queryAdd(int32_t index, const Thing& thing, uint32_t coun
 		return RETURNVALUE_THISISIMPOSSIBLE;
 	}
 
+	if (getName() == "Quiver" && !(item->getSlotPosition() & SLOTP_AMMO)) {
+		return RETURNVALUE_NOTENOUGHROOM;
+	}
+
 	const Cylinder* cylinder = getParent();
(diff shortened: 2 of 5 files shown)


MillhioreBT:
This still exists in TibiaRL today? if that's the case that's fine with me, although it's better to compare the itemId instead of using the name, I think it takes a few nanoseconds less to compare a number instead of a string, to any other item that they modify the name with Quiver, then it could cause some weird problem, maybe no know ;)

joseluis2g:
Quiver is a new feature in Cipbia, like 12.61 feature.

src/item.cpp
```diff
@@ -1238,7 +1238,7 @@ std::string Item::getDescription(const ItemType& it, int32_t lookDistance,
 		uint32_t volume = 0;
 		if (!item || !item->hasAttribute(ITEM_ATTRIBUTE_UNIQUEID)) {
 			if (it.isContainer()) {
-				volume = it.maxItems;
+				volume = item->getName() == "Quiver" ? 4 : it.maxItems;
```
kygov, line 1241: I think Item can be nullptr.

nekiro:
Hardcoding names is never a good solution, this is very basic implementation, thus I wouldn't accept this code.

src/item.cpp
Underewarrr: This working fine. Just need to be attempted to slot system don't broke.

nekiro: This code will crash the server when you look at the item in npc trade.

Underewarrr: True.... I already start a new "new class" for "quivers"

Underewarrr: Okay, already test this script and are very obsolete, to come useful to version TFS 1.3 + Ammo type Verify, and some another things are completely wrong.
If anyone going to use that keep in mind this is a VERY RAW SCRIPT for you guys have some knowledge base, this is not usefully script C++.
This is complete unstable and can be very dangerous as these people don't have knowledge about what they are doing on the source...

EPuncker:
closing this based on the last comment of the PR author and also this can't be accepted in the current state

Zbizu:
I would wait with this feature until we begi working on protocol 12

Underewarrr:
@Zbizu  we have now !

Underewarrr:
@nekiro 
#4105 This sould go to tests.

PR #3223, opened 2020-11-17
throwing items and distance attacks rework

Zbizu:
fixes the behaviour of isSightClear (#2727)

things to test:
- throwing items
- throwing items between floors
- throwing distance attacks

the github comparing broke for some reason

edit: potentially closes #2891

Issue #2891, opened 2020-04-03
Directional spells vs. walls behaviour


Leo32onGIT:
### Notes
Original thread:
https://otland.net/threads/spells-versus-walls.268213/

His issue was caused by using OTX repo:
https://github.com/mattyx14/otxserver/commit/42c9cda0558aa994613d548338a463603fbddd69

But he brings up the point that rl tibia has different behaviour.
https://otland.net/threads/spells-versus-walls.268213/page-3#post-2592654

### Steps to reproduce (include any configuration/script required to reproduce)
1. Login to game.
2. Face wall.
3. Use a directional wave or strike spell.

### Expected behaviour
The spell goes off and cost mana regardless;
But the spell only hit actually hits a SQM if `isSightClear() == true` and `CONST_PROP_BLOCKSOLID == false`.

### Actual behaviour
You get `RETURNVALUE_NOTENOUGHROOM` instead.

Here:
https://github.com/otland/forgottenserver/blob/master/src/combat.cpp#L229

Leo32onGIT:
I think this is also tied to:
https://github.com/otland/forgottenserver/issues/2727#issuecomment-598104392

The fix would require fixing the isSightClear() logic, as it doesn't quite match rl tibia either.

infister:
I'd rather suggest checking this position:
https://github.com/otland/forgottenserver/blob/master/src/spells.cpp#L699

Zbizu:
Looks like every tile of area spell needs to have isSightClear checked but something fails after first tile.

MillhioreBT:
In theory eliminating the first 3 `if`, could spells be executed as in tibia rl, correct?

TinkzBot:
The actual line of sight of throwing, reaching, and using is the same as where the spells reach. I did some research on line of sight and found out that spells use the same line of sight as to where the person can throw / "see" / reach. 

These are some images from RealOTS Cipsoft files. I can also confirm that they are exactly the same as Tibia now. 

Corners: 
https://imgur.com/a/eDSSR

1Sqm From Corners:
https://imgur.com/a/3CusA

The end result, what the actual line of sight looks like if the wall is the player position. 
Watch carefully and see how the coins change whenever 1 position is changed, it has a value somehow.
https://imgur.com/a/fmWR91i

Zbizu:
Checking every tile separately would be wasting server resources though.

[...]

PR #3265, opened 2021-01-06
Fix Sight Line for players/monsters

gigastar:
Fixed issues with #3223 by zbizu.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -559,9 +559,8 @@ std::vector< std::pair<int, int> > get2DLine(int x0, int y0, int x1, int y1)
 
 	if (dx != 0) {
 		grad = dy * 1.0 / dx;
-		if (grad > 0) { // We need to offset the line for isometric layout
-			grad = grad + .03;
-		}
+		// We need to offset the line for isometric layout
+		grad = grad + .03;
 	}
 
 	float interY = y0 + grad; //first y - intersection for the main loop
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -625,8 +625,16 @@ bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floo
 		return true;
 	}
 
-	// Force false if same floor and sightline is not clear
+	// Force false if same floor and sightline is not clear and 1 floor above is not clear.
 	if (sameFloor && !sightLineClear) {
+		if (fromPos.z > 0 && fromPos.z <= 7) {
+			Position upperPathFrom = Position(fromPos.x, fromPos.y, fromPos.z - 1);
+			Position upperPathTo = Position(toPos.x, toPos.y, toPos.z - 1);
+
+			if (checkSightLine(upperPathFrom, upperPathTo)) {
+				return true;
+			}
+		}
 		return false;
 	}
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -535,6 +535,7 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool 
 // get a line between two points
 // based on Xiaolin Wu algorithm
 // https://rosettacode.org/wiki/Xiaolin_Wu%27s_line_algorithm#C.2B.2B
+
 std::vector<std::pair<int, int>> get2DLine(const Position& fromPos, const Position& toPos)
 {
 	std::vector<std::pair<int, int>> line;
@@ -561,14 +562,12 @@ std::vector<std::pair<int, int>> get2DLine(const Position& fromPos, const Positi
 
 	int sizeOfLine = x1 - x0;
 	line.reserve(sizeOfLine);
+
 	float dx = x1 - x0;
 	float dy = y1 - y0;
-	float grad = 1.0;
-
-	if (dx != 0) {
-		grad = dy * 1.0 / dx;
-		// We need to offset the line for isometric layout
-		grad = grad + .03;
+	float grad = (dy / dx) + 0.01;
+	if (dx == 0.0) {
+		grad = 1.0;
 	}
 
 	float interY = y0 + grad; //first y - intersection for the main loop
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -559,6 +559,8 @@ std::vector<std::pair<int, int>> get2DLine(const Position& fromPos, const Positi
 		std::swap(y0, y1);
 	}
 
+	int sizeOfLine = x1 - x0;
+	line.reserve(sizeOfLine);
 	float dx = x1 - x0;
 	float dy = y1 - y0;
 	float grad = 1.0;
@@ -572,17 +574,15 @@ std::vector<std::pair<int, int>> get2DLine(const Position& fromPos, const Positi
 	float interY = y0 + grad; //first y - intersection for the main loop
 
 	if (isSteep) {
-		for (int y = x0 + 1; y < x1; y++) {
+		for (int y = x0 + 1; y < x1; ++y) {
 			int newX = std::floor(interY);
-			int newY = y;
-			line.emplace_back(newX, newY);
+			line.emplace_back(newX, y);
 			interY += grad;
 		}
 	} else {
-		for (int x = x0 + 1; x < x1; x++) {
-			int newX = x;
+		for (int x = x0 + 1; x < x1; ++x) {
 			int newY = std::floor(interY);
-			line.emplace_back(newX, newY);
+			line.emplace_back(x, newY);
 			interY += grad;
 		}
 	}
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -532,60 +532,55 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool 
 	return isSightClear(fromPos, toPos, false);
 }
 
-// get a line between two points
-// based on Xiaolin Wu algorithm
-// https://rosettacode.org/wiki/Xiaolin_Wu%27s_line_algorithm#C.2B.2B
-
+//get a line between two points
+//based on Xiaolin Wu algorithm
+//https://rosettacode.org/wiki/Xiaolin_Wu%27s_line_algorithm#C.2B.2B
 std::vector<std::pair<int, int>> get2DLine(const Position& fromPos, const Position& toPos)
 {
 	std::vector<std::pair<int, int>> line;
-	int x0 = fromPos.x;
-	int x1 = toPos.x;
-	int y0 = fromPos.y;
-	int y1 = toPos.y;
-
-	if (x0 == x1 && y0 == y1) {
+	Position& newFromPos = const_cast<Position&>(fromPos);
+	Position& newToPos = const_cast<Position&>(toPos);
+	if (newFromPos.x == newToPos.x && newFromPos.y == newToPos.y) {
 		return line;
 	}
 
-	bool isSteep = std::abs(y1 - y0) > std::abs(x1 - x0);
+	bool isSteep = std::abs(newToPos.y - newFromPos.y) > std::abs(newToPos.x - newFromPos.x);
 
 	if (isSteep) {
-		std::swap(x0, y0);
-		std::swap(x1, y1);
+		std::swap(newFromPos.x, newFromPos.y);
+		std::swap(newToPos.x, newToPos.y);
 	}
 
-	if (x0 > x1) {
-		std::swap(x0, x1);
-		std::swap(y0, y1);
+	if (newFromPos.x > newToPos.x) {
+		std::swap(newFromPos.x, newToPos.x);
+		std::swap(newFromPos.y, newToPos.y);
 	}
 
-	int sizeOfLine = x1 - x0;
+	int sizeOfLine = newToPos.x - newFromPos.x;
 	line.reserve(sizeOfLine);
 
-	float dx = x1 - x0;
-	float dy = y1 - y0;
+	float dx = newToPos.x - newFromPos.x;
+	float dy = newToPos.y - newFromPos.y;
 	float grad = (dy / dx) + 0.01;
 	if (dx == 0.0) {
 		grad = 1.0;
 	}
 
-	float interY = y0 + grad; //first y - intersection for the main loop
+	float interY = newFromPos.y + grad; //first y - intersection for the main loop
 
 	if (isSteep) {
-		for (int y = x0 + 1; y < x1; ++y) {
+		for (int y = newFromPos.x + 1; y < newToPos.x; ++y) {
 			int newX = std::floor(interY);
 			line.emplace_back(newX, y);
 			interY += grad;
 		}
 	} else {
-		for (int x = x0 + 1; x < x1; ++x) {
+		for (int x = newFromPos.x + 1; x < newToPos.x; ++x) {
 			int newY = std::floor(interY);
 			line.emplace_back(x, newY);
 			interY += grad;
 		}
 	}
-	// now we need to perform a jump between floors to see if everything is clear (literally)
 	return line;
 }
 
@@ -615,7 +610,6 @@ bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floo
 		return false;
 	}
 
-	// Cast two converging rays and see if either yields a result.
 	//check if 2d sight line is clear
 	bool sightLineClear = checkSightLine(fromPos, toPos);
 
@@ -627,7 +621,6 @@ bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floo
 
 	//floorCheck is off and clear sight line was not found so we attempt to find 3D path now
 	//(used for throwing items)
-
 	//can't throw higher than one floor above us
 	if ((toPos.z < fromPos.z) && (fromPos.z - toPos.z > 1)) {
 		return false;
@@ -666,7 +659,7 @@ bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floo
 	}
 
 	//check toPos shaft
-	for (int z = currentZ; z != toPos.z; z++) {
+	for (int z = currentZ; z != toPos.z; ++z) {
 		const Tile* tile = getTile(toPos.x, toPos.y, currentZ);
 		bool obstacleFound = tile && (tile->getGround() || tile->hasProperty(CONST_PROP_BLOCKPROJECTILE));
 		if (obstacleFound) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -649,8 +649,7 @@ bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floo
 	if (getSightLineBool && (fromPos.z < toPos.z || sameFloor)) {
 		//path was found
 		currentZ = fromPos.z;
-	}
-	else {
+	} else {
 		//try throwing above the obstacle
 		//check path and fromPos shaft
 		const Tile* tile = getTile(fromPos.x, fromPos.y, fromPos.z - 1);
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -585,7 +585,7 @@ std::vector<std::pair<int, int>> get2DLine(const Position& fromPos, const Positi
 	return line;
 }
 
-bool Map::checkSightLine(const Position& fromPos, const Position& toPos) const
+int Map::checkSightLine(const Position& fromPos, const Position& toPos) const
 {
 	std::vector<std::pair<int, int>> line = get2DLine(fromPos, toPos);
 
@@ -594,11 +594,16 @@ bool Map::checkSightLine(const Position& fromPos, const Position& toPos) const
 		uint16_t y = static_cast<uint16_t>(it.second);
 
 		const Tile* tile = getTile(x, y, fromPos.z);
+
 		if (tile && tile->hasProperty(CONST_PROP_BLOCKPROJECTILE)) {
-			return false;
+			MagicField* field = tile->getFieldItem();
+			if (field) {
+				return 2;
+			}
+				return 0;
 		}
 	}
-	return true;
+	return 1;
 }
 
 bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floorCheck) const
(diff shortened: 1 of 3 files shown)


MillhioreBT:
What is this about? I was reading the references and such an error cannot be reproduced in my sources? Is it an improvement of the code or are you really fixing something essential?

gigastar:
It's possible the problems I found were linked to my set-up.

Pretty much the sight line worked correctly (if you look at the picture where gold coins were being placed) #3223.

It was working correctly like that, however you couldn't do the same thing to the north, and it was completely breaking if you tried it to the left rather than the right.

With the changes I made, in my set up, made it so the sight line is the same in all directions. Which it should be. There may be something I have overseen so it does need to be tested more.

I would suggest you do some testing to figure out if this has changed anything.

MillhioreBT:
> It's possible the problems I found were linked to my set-up.
> 
> Pretty much the sight line worked correctly (if you look at the picture where gold coins were being placed) #3223.
> 
> It was working correctly like that, however you couldn't do the same thing to the north, and it was completely breaking if you tried it to the left rather than the right.
> 
> With the changes I made, in my set up, made it so the sight line is the same in all directions. Which it should be. There may be something I have overseen so it does need to be tested more.
> 
> I would suggest you do some testing to figure out if this has changed anything.

Horizontally it works fine, however vertically it cannot be verified with the official TibiaClient, due to limited vision, however I verified it using the internalMoveItem function manually and it also works correctly.

I apologize if I still do not understand your point, am I doing something wrong?

image
image
image
image

gigastar:
Unless i'm mistaken you should be able to throw in the first 2 spots after the walls corner (if you are 3sqm away from the corner)
Which is represented in your first picture.

gigastar:
So this is the problem. There is also a problem in this code I posted now that I have run some more tests..

The gold coins represent here you can throw
The platinum represent where you cannot (but should be able to)

This is the outcome of Zbizu's code
show2


This is the outcome of the code I have modified. (I am working on fixing it now)
show1

gigastar:
This is the results of changes I just made. The only issue right now is that the top left isn't working 100%.
show1

MillhioreBT:
The question is ... this is so in TibiaRL?

because in this image you can see how the red lines collide with the wall
<details>
  <summary>Images:</summary>

image
image

</details>

gigastar:
True and I was assuming the information from zbizu and whoever else talking about that issue was correct about how it should be to mirror rl tibia. I am pretty sure that was the correct way. Are you using stock TFS sources?

MillhioreBT:
> True and I was assuming the information from zbizu and whoever else talking about that issue was correct about how it should be to mirror rl tibia. I am pretty sure that was the correct way. Are you using stock TFS sources?

I have checked TFS is wrong, but its configuration is also wrong, in TibiaRL you can only do the following:

<details>
<summary>Images:</summary>

image
image

</details>

gigastar:
Are you saying you shouldn't be able to throw the same to the north and to the west?

MillhioreBT:
> Are you saying you shouldn't be able to throw the same to the north and to the west?

Correct, I confirmed it in TibiaRL
You can also try to enter TibiaRL and check it

gigastar:
Well that doesn't make sense but okay.

MillhioreBT:
> Well that doesn't make sense but okay.

In an isometric world it makes a lot of sense, although it does not seem so for a flat or three-dimensional world, I also think the same as you, but this repository is governed by the mechanics of TibiaRL

gigastar:
This is the end result. I was able to reproduce the bug with the extra spot being thrown on the right side. An offset was added to the grade of a line which was needed to reproduce the isometric way the line should be drawn. I did test this in a vertical layout as-well.
show

MillhioreBT:
> This is the end result. I was able to reproduce the bug with the extra spot being thrown on the right side. An offset was added to the grade of a line which was needed to reproduce the isometric way the line should be drawn. I did test this in a vertical layout as-well.
> show

changes are 99% similar to TibiaRL
only this tile that I'm pointing with the green arrow must be unreachable
image

I'm sorry if I confused you with my images
great job fix what i told you and hopefully someone else approves it really works just like TibiaRL

gigastar:
That tile is unreachable on my side, is it reachable on your set up?

MillhioreBT:
> That tile is unreachable on my side, is it reachable on your set up?

yes :/ 
image

MillhioreBT:
> That tile is unreachable on my side, is it reachable on your set up?

yes :/

gigastar:
That wall isn't completed. You need the last end part (the pillar)

nekiro:
What was the tested scenarious?
I believe this also affects monsters targetung right? Was it thoroughly tested?

MillhioreBT:
> What was the tested scenarios?
> I believe this also affects monsters targeting and rune throwing (also other spells possibly?) right? Was it thoroughly tested?

I have already tried all the scenarios, and the monsters comply with the proposed rules, here I will leave the image provided by @EPuncker from Tibia RL

As you can see, the orc that is to the north is unreachable and therefore his strange walk to get closer


What still seems strange to me is that in my sources, these changes seem to err a bit, because I can move objects to the following place:


If anyone else can check that that would be great

gigastar:
Okay, now I am seeing the problem there. Thank you for slapping me a couple times. Ill get that fixed.

gigastar:
Try now please

MillhioreBT:
Meets expectations 😍
GIF 08-01-2021 11-09-46 p  m

Now if it perfectly emulates the behavior of Tibia RL

MillhioreBT (approved):

gigastar:
To be clear all credits to Zbizu he did great implementing the Xiaolin Wu algorithm to it. I just added an offset that was needed. my dumb a** almost made it so it works in all directions but we can forget about that.

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
 	return isSightClear(fromPos, toPos, false);
 }
 
-bool Map::checkSightLine(const Position& fromPos, const Position& toPos) const
+//get a line between two points
+//based on Xiaolin Wu algorithm
```
ranisalt, line 536: What algorithm is this? Please link to it in the code itself.

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+std::vector< std::pair<int, int> > get2DLine(int x0, int y0, int x1, int y1)
 {
-	if (fromPos == toPos) {
-		return true;
+	std::vector< std::pair<int, int> > line;
```
ranisalt, line 539: Remove the spaces after < and before > on both lines

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		std::swap(x1, y1);
+	}
 
-	int32_t A = Position::getOffsetY(destination, start);
-	int32_t B = Position::getOffsetX(start, destination);
-	int32_t C = -(A * destination.x + B * destination.y);
+	if (x0 > x1) {
+		std::swap(x0, x1);
+		std::swap(y0, y1);
+	}
```
ranisalt, line 554: Tip: you can use [std::minmax](https://en.cppreference.com/w/cpp/algorithm/minmax) to achieve the same result without the explicit branch:

```cpp
std::tie(x0, x1) = std::minmax(x0, x1);
std::tie(y0, y1) = std::minmax(y0, y1);
```
And then `x0` will always be less or equal than `x1`, ditto for `y0` and `y1`.

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+	float dy = y1 - y0;
+	float grad = 1.0;
 
-		if (start.y != destination.y && (start.x == destination.x || move_hor > move_ver || move_hor > move_cross)) {
-			start.y += my;
-		}
+	if (dx != 0) {
+		grad = dy * 1.0 / dx;
+		// We need to offset the line for isometric layout
+		grad = grad + .03;
```
ranisalt, line 563: Where does `0.03` come from?

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+			int newY = y;
+			line.push_back({ newX, newY });
+			interY += grad;
+		}
+	} else {
+		for (int x = x0 + 1; x < x1; x++) {
+			int newX = x;
+			int newY = std::floor(interY);
+			line.push_back({ newX, newY });
+			interY += grad;
```
ranisalt, line 580: There will be speed benefits if you `reserve` space in `line` before `push_back`s. It will always be `x1 - x0` long. Also, use `emplace_back` instead.

By the way, `++x` is preferred over `x++`.

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
 	// now we need to perform a jump between floors to see if everything is clear (literally)
-	while (start.z != destination.z) {
-		const Tile* tile = getTile(start.x, start.y, start.z);
-		if (tile && tile->getThingCount() > 0) {
+	return line;
+}
+
+bool Map::checkSightLine(const Position& fromPos, const Position& toPos) const
+{
+	std::vector< std::pair<int, int> > line = get2DLine(static_cast<int>(fromPos.x), static_cast<int>(fromPos.y), static_cast<int>(toPos.x), static_cast<int>(toPos.y));
```
ranisalt, line 590: Instead of this `static_cast` cluster, make your `get2DLine` function take the correct types.

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+	//alternatively we can end the calculations if clear sight line was found and the target is on the same floor
+	if (floorCheck || (sameFloor && sightLineClear)) {
+		return sightLineClear;
+	}
+
+	//floorCheck is off and clear sight line was not found so we attempt to find 3D path now
+	//(used for throwing items)
+
+	//can't throw higher than one floor above us
+	if ( (toPos.z < fromPos.z) && (fromPos.z - toPos.z > 1) ) {
```
ranisalt, line 629: Spacing

src/map.cpp
```diff
@@ -532,62 +532,150 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+	while (currentZ < toPos.z) {
+		const Tile* tile = getTile(toPos.x, toPos.y, currentZ);
+		bool obstacleFound = tile && (tile->getGround() || tile->hasProperty(CONST_PROP_BLOCKPROJECTILE));
+
+		if (obstacleFound) {
+			return false;
+		}
+
+		++currentZ;
+	}
```
ranisalt, line 676: This is a `for` loop, don't use `while`.

ranisalt (changes requested):
Most of your mistakes are related to not following the code convention. Please use as much as possible of the current code style.

src/map.cpp
Zbizu: I was basing that code on this:
https://rosettacode.org/wiki/Xiaolin_Wu%27s_line_algorithm#C.2B.2B

src/map.cpp
Zbizu: agreed. I used this to avoid issues with converting data types (like #3217), but it should definitely be rewritten.

src/map.cpp
nekiro: Probably from testing the code and finding the right value, but it's just my wild guess.

gigastar (commented):
Yes, it was after a lot of researching the different grad values that were being used in each spot. What I gathered was the issue was taking place if the grad was: 0.3333 repeating. If the grad was anything else it was fine. That value is just enough that it won't interfere with the way the line is reading in any other scenario. It is hard to explain without you seeing all grad values I printed out after trying to throw a coin in every spot open.

It's definitely not optimal. A better way would be to add the offset only in the specific instance its needed. This is what I was able to come up with as an alternative. Not that its detrimental to anything.

src/map.cpp
gigastar: > Probably from testing the code and finding the right value, but it's just my wild guess.

Yes, it was after a lot of researching the different grad values that were being used in each spot. What I gathered was the issue was taking place if the grad was: 0.3333 repeating. If the grad was anything else it was fine. That value is just enough that it won't interfere with the way the line is reading in any other scenario. It is hard to explain without you seeing all grad values I printed out after trying to throw a coin in every spot open.

It's definitely not optimal. A better way would be to add the offset only in the specific instance its needed. This is what I was able to come up with as an alternative. Not that its detrimental to anything.

gigastar: It would be good for someone to check it with otclient and see if the line remains the same past the limits of the tibia client but that's not what TFS is for.

src/map.cpp
gigastar: I tried to implement this, it could be because i'm in over my head with it, but it wouldn't produce the same outcome.

src/map.cpp
ranisalt: I think the ideal solution would be to mirror the vector to always be in the same direction, then calculate only one variation i.e. normalize every distance to the positive vector (top-left to bortom-right) and check that without magic numbers.

Zbizu: I recommend testing this on both Linux and Windows anyway because they handle calculations a bit differently (for example math.random gets different results on same seed).

src/map.cpp
```diff
@@ -532,62 +532,148 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		}
+	} else {
+		for (int x = x0 + 1; x < x1; ++x) {
+			int newY = std::floor(interY);
+			line.emplace_back(x, newY);
+			interY += grad;
 		}
 	}
-
 	// now we need to perform a jump between floors to see if everything is clear (literally)
```
Zbizu, line 588: outdated comment referencing removed lines of code

src/map.cpp
```diff
@@ -532,62 +532,148 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
-	if (floorCheck && fromPos.z != toPos.z) {
+	bool sameFloor = fromPos.z == toPos.z;
+
+	//if floorCheck is enabled, same floor clear sight line is required
+	//(used for distance attacks)
+	if (floorCheck && !sameFloor) {
 		return false;
 	}
 
 	// Cast two converging rays and see if either yields a result.
```
Zbizu, line 618: Another outdated comment. New algorithm doesn't cast two rays.

Zbizu (changes requested):

Zbizu:
> 
> 
> To be clear all credits to Zbizu he did great implementing the Xiaolin Wu algorithm to it. I just added an offset that was needed. my dumb a** almost made it so it works in all directions but we can forget about that.

Both you and ranisalt did a great job getting the code to tfs standards. I can have great algorhytmical thinking but I'm still a noob at c++.

gigastar:
@ranisalt do I HAVE to change anything else?

ranisalt (dismissed):
The code is now conforming to the standard, though I can't test the behavior. I'll trust @MillhioreBT on that.

nekiro:
I'll give this a test soon, need to find some free time.

nekiro:
Did anyone check performance of this code comparing to the old code?

EPuncker:
ok, so I tested these changes on my testing tfs (not really clean code, but close to master) and the monsters started behaving weirdly, so I'd like to ask you guys to test how are monsters behaving (with normal characters)

gigastar:
Tested on base files and everything seems to be normal. I tried all 3 scenarios that are in the original issue. Also some other ones just to see if anything weird was going on. I can't see any problem other than the slight differences in monster's AI behavior compared to Tibia.

gigastar:
After doing some more testing there seems to be a problem with the code. If someone can test it to make sure. For some reason items are not able to be thrown correctly. Which is weird because it was working fine before.

gigastar:
After doing more testing I found a problem with throwing items. It was due to trying to use const_cast and modifying the Position& created. I had to revert to x0,x1, ect. After those changes I tested all scenarios again and everything seems to be correct.

emil92b:
try using spells nearby or trough magic wall

gigastar:
Not sure how that got changed

emil92b:
should only be able to throw items up 1 floor, and down 7 floors

see image: https://i.imgur.com/oyrjmls.png (no limit)

currently you can only throw down 2 floors

gigastar:
The issue with how many floors down you can throw isn't related to this commit. I will look into being able to throw through magic walls.

gigastar:
Anyone have any ideas on the magic wall issue? I tried using Field* field = tile->getFieldItem() and checking if its a blocksolid, ect, ect. Nothing I have tried seems to work. I did fix the item throw range down floors but that will be in a different commit.

Zbizu:
Do magic walls have blockProjectile in otb?

nekiro:
> Do temporary magic walls have blockProjectile in otb?

They are immovable blocksolids by default.

EPuncker:
> > Do temporary magic walls have blockProjectile in otb?
> 
> They are immovable blocksolids by default.

I think @Zbizu means if magic wall have <attribute key="blocking" value="1" /> on otb

currently our items.xml is like that:

```
		<attribute key="duration" value="20" />
	</item>
	<item id="1499" article="a" name="rush wood">
		<attribute key="duration" value="45" />
		<attribute key="blocking" value="1" />
	</item>
```
But I don't know if that match 10.98 data as I don't have it to test

Zbizu:
`blocksolid` and `blockprojectile` are different flags. First tells that the object is something like a wall, or some rock (they get replaced by default in map editor unless you stack them on top of each other through moving), second tells that you can't shoot through it. If you have a window it's an object you can't walk through, but you can shoot through it.

https://github.com/otland/forgottenserver/blob/master/src/items.cpp#L447

gigastar:
I know I was just saying I tried looking for all flags and none of them changed the outcome.

Zbizu:
have you tried obtaining the magic wall by id? tile->getFieldItem might be bugged

I remember that it was possible to get mw by id back in tfs 1.1:
https://otland.net/threads/10-76-tfs-1-1-old-mw-sprite-timer.230810/

gigastar:
So you can throw through any walls with runes, ect. The problem is bigger than just mwalls.

gigastar:
Last commit should fix the problem.

emil92b:
> Last commit should fix the problem.

it breaks item throwing like this tho, maby something else too
https://i.imgur.com/k6ee1rt.png

gigastar:
Couldn't reproduce that.

nekiro:
> Couldn't reproduce that.

He means throwing over the roof.

gigastar:
I don't understand.

Zbizu:
I coded it to ignore borders:
`bool obstacleFound = tile && (tile->getGround() || tile->hasProperty(CONST_PROP_BLOCKPROJECTILE));`

Current tfs implementation for that is:
https://github.com/otland/forgottenserver/blob/master/src/map.cpp#L565

reason for my change: I tried to throw coins from tfs rook temple roof the the floor and I wasn't able to throw them next to temple wall tiles because the shaft was covered by roof border. If I remember correctly this doesn't happen in rl.

gigastar:
So is there a problem or? Can someone please explain a bit better im a little slow ;)

Zbizu:
You stand on a roof of a building. If the roof has a border, tfs prevents you to throw through it so ground floor tiles near building walls become unreachable.

gigastar:
You can throw passed roof boarders on rl tibia, yes.

gigastar:
It seems you can throw over walls/roofs if its 1 floor also which can't be done on TFS

gigastar:
You can throw past borders on RL tibia.

I have made it so we can determine what is blocking the sightline in the path. If it is a magic wall then we do not allow throwing over it. This should fix the last couple issues brought up. Now we need to test more.

Zbizu:
oops, linked wrong line earlier, here's the one responsible for checking if you can throw through floors in current tfs:
https://github.com/otland/forgottenserver/blob/master/src/map.cpp#L573

gigastar:
TFS and RL tibia are both able to move items through borders. I fixed the problem mentioned by emil92b, I believe.

src/map.cpp
```diff
@@ -532,62 +531,163 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		return true;
+	}
+
+	uint8_t currentZ;
+
+	//check if the path is clear or we have to attempt throwing above the obstacle
+	if (getSightLineBool && (fromPos.z < toPos.z || sameFloor)) {
+		//path was found
+		currentZ = fromPos.z;
+	}
```
EPuncker, line 652: } else {

EPuncker (changes requested):
this need a rebase tho

src/map.cpp
```diff
@@ -532,62 +531,162 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
 
-	return true;
+		const Tile* tile = getTile(x, y, fromPos.z);
+
+		if (tile && tile->hasProperty(CONST_PROP_BLOCKPROJECTILE)) {
+			MagicField* field = tile->getFieldItem();
+			if (field) {
+				return 2;
+			}
+				return 0;
```
Zbizu, line 603: one tab too much in this line

gigastar:
How does the code now deviate from that?

nekiro:
@TinkzBot that's definitely great information, can you please post your research in #2891?

Zbizu:
I see this was closed and the repository having these changes isn't reachable anymore. Are you going to continue working on that code @gigastar ?

What needs to be done to finish it? Only interaction with magic wall fix?

gigastar:
I am not sure if anything needs to be added for it to be finished. I fixed all the problems that were mentioned. The code wasn't beautiful but it wasn't terrible.

1) You can throw over 1 floor
2) Problem throwing over magic walls was fixed (You can not now.)
3) Magic walls blocked throwing if found anywhere in the path
4) The sightline seemed to mirror tibias sightline by all testing I did
5) Just waiting for reviews on code changes (to mimic TFS style) and any testing anyone wants to do the make sure the sightline is correct and everything previously mentioned is correct.

MillhioreBT:
@gigastar your commitment not only improves behavior, but emulates the real behavior of Tibia and fixes other existing problems, if your commitment was not accepted quickly it is due to the fact that this function is very delicate and is present in almost all the mechanics of the game and it should be reviewed very well before merging it, I hope you are not discouraged and at some point re-open this PR, you do not need to close it because we do not accept it immediately

PR #3267, opened 2021-01-07
Fix mana and life leech bugs

### LINKED ISSUE #3266: Leech System have bug, work without targets

1. create item with specialSkill lifeleechchance and amount or mana
2. equip this item and use a spell of area or direction

### Expected behaviour
You should not absorb anything, because you are not affecting any target with your spells

### Actual behaviour
The total amount absorbed must be calculated according to the total damage you inflict and divided by the targets that are in the affected area.

GIF 07-01-2021 01-06-20 p  m

nekiro:
Fix #3266 
also now the formula is correct and it also adds some graphical improvements, like correct message order in server log.

Credits to https://otland.net/members/unknown-soldier.101063/
for pretty good research https://otland.net/threads/buggy-life-and-mana-leech-in-tfs-1-3.274192/page-2#post-2641180

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -894,7 +894,8 @@ void Combat::doAreaCombat(Creature* caster, const Position& position, const Area
 
 	postCombatEffects(caster, position, params);
 
-	std::vector<int32_t> damageMap;
+	std::vector<Creature*> toDamageCreatures;
+	toDamageCreatures.reserve(100);
 
 	for (Tile* tile : tileList) {
 		if (canDoCombat(caster, tile, params.aggressive) != RETURNVALUE_NOERROR) {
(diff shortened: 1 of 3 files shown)


MillhioreBT:
The problem persists, remember that if you don't hurt anyone, where is it supposed to absorb life or mana uwu?
You must include some way to verify that the damage was effective and filtered correctly and then absorb part of the damage inflicted, this also includes the reduction by the target, if the target dodges this attack or reduces it, the absorption will also be affected

MillhioreBT (approved):
It works perfectly, I have no idea if absorption is affected by internal protections on monsters or players, but if that's not the case, these changes are perfect!

MillhioreBT:
Regardless of whether an expansive or single-target spell, the absorption is the same for both cases, so this RP is totally correct.

MillhioreBT (dismissed):

MillhioreBT:
Thanks to you I was inspired and learned to understand these functions a little! <3
I was watching the tests that the boy from Otland did and although his changes are almost perfect, they lack some details that are not very relevant but are important for detailed people.

I have made my own version inspired by your changes, you should take a look: [unrealtfs PR12](https://github.com/Unreal-TFS/unrealtfs/pull/12)

nekiro:
> Thanks to you I was inspired and learned to understand these functions a little! <3
> I was watching the tests that the boy from Otland did and although his changes are almost perfect, they lack some details that are not very relevant but are important for detailed people.
> 
> I have made my own version inspired by your changes, you should take a look: [unrealtfs PR12](https://github.com/Unreal-TFS/unrealtfs/pull/12)
> 
> this is how it should work ...
> Absorption messages must be separately and after damage msg in the log, and I think that if a monster ignores the attack for whatever reason, it is ignored for the formula, in LeagueOfLegends and other games this behavior is true, and Tibia I do not think it is the exception

I'm still investigating, because I didn't yet find proper formula for AOE spells. I'll need to lend some character to go around and test stuff. If anyone has anything I would appreciate.

MillhioreBT:
> > Thanks to you I was inspired and learned to understand these functions a little! <3
> > I was watching the tests that the boy from Otland did and although his changes are almost perfect, they lack some details that are not very relevant but are important for detailed people.
> > I have made my own version inspired by your changes, you should take a look: [unrealtfs PR12](https://github.com/Unreal-TFS/unrealtfs/pull/12)
> > this is how it should work ...
> > Absorption messages must be separately and after damage msg in the log, and I think that if a monster ignores the attack for whatever reason, it is ignored for the formula, in LeagueOfLegends and other games this behavior is true, and Tibia I do not think it is the exception
> 
> I'm still investigating, because I didn't yet find proper formula for AOE spells. I'll need to lend some character to go around and test stuff. If anyone has anything I would appreciate.
> It's not the final version of this pr, it has some issues, like it only gives leech for one creature, which is thing to change too, but at this time I focus on trying to get the formula first.

I love the way you think, always trying to innovate and make everything a little more efficient, but in the console logs in Tibia RL they show each message separately and not a total message of the absorption obtained, this means that it does not follow this order:
image

nekiro:
> > > Thanks to you I was inspired and learned to understand these functions a little! <3
> > > I was watching the tests that the boy from Otland did and although his changes are almost perfect, they lack some details that are not very relevant but are important for detailed people.
> > > I have made my own version inspired by your changes, you should take a look: [unrealtfs PR12](https://github.com/Unreal-TFS/unrealtfs/pull/12)
> > > this is how it should work ...
> > > Absorption messages must be separately and after damage msg in the log, and I think that if a monster ignores the attack for whatever reason, it is ignored for the formula, in LeagueOfLegends and other games this behavior is true, and Tibia I do not think it is the exception
> > 
> > 
> > I'm still investigating, because I didn't yet find proper formula for AOE spells. I'll need to lend some character to go around and test stuff. If anyone has anything I would appreciate.
> > It's not the final version of this pr, it has some issues, like it only gives leech for one creature, which is thing to change too, but at this time I focus on trying to get the formula first.
> 
> I love the way you think, always trying to innovate and make everything a little more efficient, but in the console logs in Tibia RL they show each message separately and not a total message of the absorption obtained, this means that it does not follow this order:
> image

Yeah, my last commit should have fixed this to mimic the right message order.

MillhioreBT:
I would like you to share the leech and the critical in separate functions, so that the same thing should not be repeated twice in different parts of the code, It is only something aesthetic, but I think it reduces lines of code and it looks more beautiful, for now is what I think, later I will give another review

nekiro:
> I would like you to share the leech and the critical in separate functions, so that the same thing should not be repeated twice in different parts of the code, It is only something aesthetic, but I think it reduces lines of code and it looks more beautiful, for now is what I think, later I will give another review

Making these in separate functions is bloating to me. These two arent repeated code, onTargetCombat uses different code (different formula) and onAreaCombat is also using different formula also it has some other things that would need to be passed to another function for it to work, so imo it's redundant to move it to separate method (it will be only used there anyway)

MillhioreBT:
> > I would like you to share the leech and the critical in separate functions, so that the same thing should not be repeated twice in different parts of the code, It is only something aesthetic, but I think it reduces lines of code and it looks more beautiful, for now is what I think, later I will give another review
> 
> Making these in separate functions is bloating to me. These two arent repeated code, onTargetCombat uses different code (different formula) and onAreaCombat is also using different formula also it has some other things that would need to be passed to another function for it to work, so imo it's redundant to move it to separate method (it will be only used there anyway)

Well I don't say anything sir: v
Great job, hope that someone with writing power and therefore has a relevant review can verify this and give their opinion on it. <3

EPuncker (approved):

PR #3269, opened 2021-01-09
Drop sudo instructions from Travis CI config

MillhioreBT:
* Build config validation
* deprecated key sudo (The key `sudo` has no effect anymore.)

REF: https://config.travis-ci.com/ref/sudo

and anothers warnings...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: required
 dist: bionic
 language: cpp
 compiler:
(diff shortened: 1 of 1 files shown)


MillhioreBT:
Oops. sorry XDD

ranisalt:
Are we even using Travis still? It seems that GitHub Actions perform the same checks and more, so both Travis and Appveyor can be removed for now.

DSpeichert:
@ranisalt I'm keeping them for now because the compiled Artifacts cannot be downloaded from github by guests, while Travis and AppVeyor don't require logging in to download. That's the only reason.

DSpeichert (approved):

PR #3272, opened 2021-01-16
Don't update monster idle status if it cannot see the creature

nekiro:
onCreatureFound is called for every single creature around, even if monster cannot see the creature, later it will call canSee on targets to delete them from target list.
This pr adds check to onCreatureFound, so it will check canSee before checking if its friend or opponent..

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -369,9 +369,7 @@ void Monster::updateTargetList()
 	g_game.map.getSpectators(spectators, position, true);
 	spectators.erase(this);
 	for (Creature* spectator : spectators) {
-		if (canSee(spectator->getPosition())) {
-			onCreatureFound(spectator);
-		}
+		onCreatureFound(spectator);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

ranisalt:
What is the difference?

nekiro:
> What is the difference?
When you login and monsters are under you, they will go active for a second to go idle right after.

EPuncker:
> What is the difference?

before:
depois1

after:
depois2

MillhioreBT (commented):
This is perfect 😍

ranisalt (approved):

emil92b:
check if your summons is moving after changing floor

MillhioreBT (commented):
I found a problem with this:
image

EPuncker:
> check if your summons is moving after changing floor

can you elaborate please?

emil92b:
> > check if your summons is moving after changing floor
> 
> can you elaborate please?

on real tibia they walk if your not on the same floor as them, just make sure its the same behaviour is all i was saying : )

PR #3274, opened 2021-01-16
Change storage values size to int64_t

EPuncker:


ranisalt:
I wouldn't do this unless very necessary, as it has the potential to waste a lot of memory and storage for very little usage.

EPuncker:
hey it was not meant to be here

PR #3275, opened 2021-01-17
Drop skill caches & fix skill tries precision

### LINKED ISSUE #3217: incorrect return values in vocation:getRequiredSkillTries(skill, level)

ranisalt:
Essentially what I proposed in #3227, just drop the caches since calculating the values is not slow enough to be worth it to cache & add code complexity.

[More details on the speed of std::pow](https://baptiste-wicht.com/posts/2017/09/cpp11-performance-tip-when-to-use-std-pow.html).

Also, moved `skillBase` to the cpp file as it does not need to be declared in the header.

Closes #3217

### DIFF
--- a/src/vocation.cpp
+++ b/src/vocation.cpp
@@ -179,7 +179,7 @@ uint16_t Vocations::getPromotedVocation(uint16_t vocationId) const
 	return VOCATION_NONE;
 }
 
-static uint32_t skillBase[SKILL_LAST + 1] = {50, 50, 50, 50, 30, 100, 20};
+static const uint32_t skillBase[SKILL_LAST + 1] = {50, 50, 50, 50, 30, 100, 20};
 
 uint64_t Vocation::getReqSkillTries(uint8_t skill, uint16_t level)
 {
--- a/src/vocation.cpp
+++ b/src/vocation.cpp
@@ -186,13 +186,10 @@ uint64_t Vocation::getReqSkillTries(uint8_t skill, uint16_t level)
 	if (skill > SKILL_LAST) {
 		return 0;
 	}
-
-	uint64_t tries = static_cast<uint64_t>(skillBase[skill] * std::pow<double>(skillMultipliers[skill], level - 11));
-	return tries;
+	return static_cast<uint64_t>(skillBase[skill] * std::pow<double>(skillMultipliers[skill], level - 11));
 }
 
 uint64_t Vocation::getReqMana(uint32_t magLevel)
 {
-	uint64_t reqMana = std::floor<uint64_t>(1600 * std::pow<double>(manaMultiplier, static_cast<int32_t>(magLevel) - 1));
-	return reqMana;
+	return std::floor<uint64_t>(1600 * std::pow<double>(manaMultiplier, static_cast<int32_t>(magLevel) - 1));
 }
--- a/src/vocation.cpp
+++ b/src/vocation.cpp
@@ -116,7 +116,7 @@ bool Vocations::loadFromXml()
 				if (skillIdAttribute) {
 					uint16_t skill_id = pugi::cast<uint16_t>(skillIdAttribute.value());
 					if (skill_id <= SKILL_LAST) {
-						voc.skillMultipliers[skill_id] = pugi::cast<float>(childNode.attribute("multiplier").value());
+						voc.skillMultipliers[skill_id] = pugi::cast<double>(childNode.attribute("multiplier").value());
 					} else {
 						std::cout << "[Notice - Vocations::loadFromXml] No valid skill id: " << skill_id << " for vocation: " << voc.id << std::endl;
 					}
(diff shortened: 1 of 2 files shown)


DSpeichert (approved):

nekiro (approved):

MillhioreBT (commented):
Test and work!

I have tried these changes, and they work very well, however it is not more efficient than the previous one. This does not mean that it is wrong, just the tests I did were somewhat exaggerated.

this was my test:
```lua
local talkTest = TalkAction("/tt")
function talkTest.onSay(player, words, param)
		return true
	end
	print(Vocation(3):getRequiredSkillTries(SKILL_AXE, 100))
	local now = os.mtime()
	for i = 1, 10000000 do
		player:getVocation():getRequiredSkillTries(SKILL_SWORD, 100)
	end
	print((os.mtime() - now) / 1000)
	return false
end

talkTest:separator(" ")
talkTest:register()
```

the results vary between a (60~70)% longer wait time for this function, however I think this function, it is not called so many times and in such short periods of time, as for me I think I'll take some ideas from these changes and use them to generate some merged version of cache + this for my server.

nekiro:
> Test and work!
> 
> I have tried these changes, and they work very well, however it is not more efficient than the previous one. This does not mean that it is wrong, just the tests I did were somewhat exaggerated.
> 
> this was my test:
> 
> ```lua
> local talkTest = TalkAction("/tt")
> function talkTest.onSay(player, words, param)
> 	if not player:getGroup():getAccess() then
> 		return true
> 	end
> 	print(Vocation(3):getRequiredSkillTries(SKILL_AXE, 100))
> 	local now = os.mtime()
> 	for i = 1, 10000000 do
> 		player:getVocation():getRequiredSkillTries(SKILL_SWORD, 100)
> 	end
> 	print((os.mtime() - now) / 1000)
> 	return false
> end
> 
> talkTest:separator(" ")
> talkTest:register()
> ```
> 
> the results vary between a (60~70)% longer wait time for this function, however I think this function, it is not called so many times and in such short periods of time, as for me I think I'll take some ideas from these changes and use them to generate some merged version of cache + this for my server.

It's cause he removed the caching. If you tested it this way before then you thats why you has better results.

ranisalt:
@MillhioreBT yeah, calling it with the same input is where the cache shines. It would be better to call with random parameters (skill, skill level and magic level) to test random access, where I think this difference will be smaller.

Czopeq:
With random input cache will be worst because of massive relocation.

If you think that test who will be never exist in server.. is good.. then something there is wrong. xd

ranisalt:
I made it to run with random values instead:

<details>
```lua
local talk = TalkAction("/test")

		return true
	end

	local vocations = {}
	for i = 1, 8 do
		vocations[i] = Vocation(i)
	end

	local randoms = {}
	local random = math.random
	for i = 1, 10000000 do
		randoms[i] = {random(1, 8), random(0, 6), random(10, 100)}
	end

	local now = os.mtime()
	for i, v in ipairs(randoms) do
		vocations[v[1]]:getRequiredSkillTries(v[2], v[3])
	end
	print(os.mtime() - now))
	return false
end

talk:separator(" ")
talk:register()
```
</details>

And it takes an average of 3500ms with cache (between 3100 and 3900) and 2030ms without (consistently between 2000 and 2100). YMMV.

EPuncker (approved):

CristoferMartins:
What if we generate the whole cache from startup? Usually skills are in a reasonable range(like 0-200). That would be faster? And there is like two more questions about this change. 1: Changing it to unordered_map would make it better? 2: In a real server with many players after not much time the cache will answer for all values so its pay off? @ranisalt

ranisalt:
Even without a cache, each call takes about 0.2us - that's *micro*, not even *milli*. The cache may pay off, but what do we gain from optimizing 0.2 millionths of a second?

MillhioreBT:
> @MillhioreBT yeah, calling it with the same input is where the cache shines. It would be better to call with random parameters (skill, skill level and magic level) to test random access, where I think this difference will be smaller.
> 
> Edit: your test takes 6848ms for me. However, if I move the `player:getVocation()` out of the loop (it's an invariant), it goes down to 1366ms, so that's the bulk of the time:
> 
> ```lua
> 	local voc = player:getVocation()
> 	local now = os.mtime()
> 	for i = 1, 10000000 do
> 		voc:getRequiredSkillTries(SKILL_SWORD, 100)
> 	end
> ```

You are absolutely right gentleman, I apologize for reviewing this at dawn when I'm tired 🤦‍♀️ ahaha

I did new tests with your suggestions, and the difference was reduced to a constant 30%, the difference is so minimal that it should not be detectable or cause slowness problems for the scripts, I do not see why keep the cache, it is already a matter of preferences, but it doesn't contribute more than a few extra µs ☺

<details>
<summary>Test Script</summary>

```lua
local talkTest = TalkAction("/tt")
function talkTest.onSay(player, words, param)
		return true
	end
	local vocation = Vocation(3)
	print(vocation:getRequiredSkillTries(SKILL_AXE, 100))
	local now = os.mtime()
	for i = 1, 10000000 do
		vocation:getRequiredSkillTries(SKILL_AXE, math.random(1, 200))
	end
	print((os.mtime() - now) / 1000)
	return false
end

talkTest:separator(" ")
talkTest:register()
```

</details>

MillhioreBT (approved):

DSpeichert (approved):

PR #3278, opened 2021-01-20
Implement scheduler timers using asio deadline timers

### LINKED ISSUE #2218: Investigate the use of ASIO timers instead of the Scheduler tasks

ranisalt:
As it's been a while that #3056 was last updated, and there is usage of deprecated facilities. This is a bit cleaner using smart pointers to avoid manual deletion and skip the cache as constructing a timer seems very lightweight.

Closes #2218
Closes #3056

### DIFF
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -23,113 +23,86 @@
 
 void Scheduler::threadMain()
 {
-	std::unique_lock<std::mutex> eventLockUnique(eventLock, std::defer_lock);
-	while (getState() != THREAD_STATE_TERMINATED) {
-		std::cv_status ret = std::cv_status::no_timeout;
-
-		eventLockUnique.lock();
-		if (eventList.empty()) {
-			eventSignal.wait(eventLockUnique);
-		} else {
-			ret = eventSignal.wait_until(eventLockUnique, eventList.top()->getCycle());
-		}
-
-		// the mutex is locked again now...
-		if (ret == std::cv_status::timeout && !eventList.empty()) {
-			// ok we had a timeout, so there has to be an event we have to execute...
-			SchedulerTask* task = eventList.top();
-			eventList.pop();
-
-			// check if the event was stopped
-			auto it = eventIds.find(task->getEventId());
-			if (it == eventIds.end()) {
-				eventLockUnique.unlock();
-				delete task;
-				continue;
-			}
-			eventIds.erase(it);
-			eventLockUnique.unlock();
-
-			task->setDontExpire();
-			g_dispatcher.addTask(task, true);
-		} else {
-			eventLockUnique.unlock();
-		}
-	}
+	io_service.run();
 }
 
 uint32_t Scheduler::addEvent(SchedulerTask* task)
 {
-	eventLock.lock();
-
-	if (getState() != THREAD_STATE_RUNNING) {
-		eventLock.unlock();
-		delete task;
-		return 0;
-	}
-
 	// check if the event has a valid id
 	if (task->getEventId() == 0) {
+		uint32_t tmpEventId = ++lastEventId;
 		// if not generate one
-		if (++lastEventId == 0) {
-			lastEventId = 1;
+		if (tmpEventId == 0) {
+			tmpEventId = ++lastEventId;
 		}
-
-		task->setEventId(lastEventId);
+		task->setEventId(tmpEventId);
 	}
 
 	// insert the event id in the list of active events
-	uint32_t eventId = task->getEventId();
-	eventIds.insert(eventId);
+	io_service.post([this, task]() {
+		boost::asio::deadline_timer* timer;
+		if (timerCacheVec.empty()) {
+			timer = new boost::asio::deadline_timer(io_service);
+		} else {
+			timer = timerCacheVec.back();
+			timerCacheVec.pop_back();
+		}
+		eventIdTimerMap[task->getEventId()] = timer;
 
-	// add the event to the queue
-	eventList.push(task);
+		timer->expires_from_now(boost::posix_time::milliseconds(task->getDelay()));
+		timer->async_wait([this, task, timer](const boost::system::error_code& error) {
+			timerCacheVec.push_back(timer);
+			eventIdTimerMap.erase(task->getEventId());
 
-	// if the list was empty or this event is the top in the list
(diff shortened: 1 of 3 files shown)


src/scheduler.cpp
```diff
@@ -20,116 +20,69 @@
...
+		task->setEventId(++lastEventId);
 	}
 
 	// insert the event id in the list of active events
-	uint32_t eventId = task->getEventId();
-	eventIds.insert(eventId);
-
-	// add the event to the queue
-	eventList.push(task);
+	boost::asio::post(io_context, [this, task]() {
```
djarek, line 39: Do we require C++14? If yes, `task` should be owned by a `unique_ptr` in the capture.

src/scheduler.cpp
```diff
@@ -20,116 +20,69 @@
...
+		auto& timer = eventIdTimerMap[task->getEventId()];
+		timer.reset(new boost::asio::deadline_timer(io_context));
```
djarek, line 41: I don't think there's a reason for the indirection in the mapped type, is there?

ranisalt: @djarek at first I didn't think that a task and an entry in the timer map were 1:1, but I just noticed they are and I can cleanup this a bit.

src/scheduler.cpp
ranisalt: We don't :(

yamaken93:
@ranisalt looks like a clean implementation. Is there is any tests to make sure it works like the previous one?

ranisalt:
> Is there is any tests to make sure it works like the previous one?

I don't know for sure, but a couple of `addEvent`s to a player logging in should be enough to see.

MillhioreBT (approved):
Tested and it works, it also increases performance!

<details>
<summary>Test Code</summary>

```lua
GLOBAL_TEST_EV = {}

local talkTest = TalkAction("/tt")
function talkTest.onSay(player, words, param)
		return true
	end

	if param == "test" then
		table.sort(GLOBAL_TEST_EV, function(a, b) return a < b end)
		print((GLOBAL_TEST_EV[#GLOBAL_TEST_EV] - GLOBAL_TEST_EV[1]) / 1000)
		GLOBAL_TEST_EV = {}
		return false
	end
	
	for i = 1, 10000 do
		addEvent(function(i)
			local s = ""
			for i = 1, 1000 do
				s = s .. 'i'
			end
			GLOBAL_TEST_EV[i] = os.mtime()
		end, 1000, i)
	end

	return false
end

talkTest:separator(" ")
talkTest:register()
```

</details>

After:
Screenshot_5
Before:
Screenshot_6

The improvement is evident, it takes less time for the processor to work on all scheduled tasks with these changes.

ranisalt:
> The improvement is evident, it takes less time for the processor to work on all scheduled tasks with these changes.

:pray: hopefully this helps with high intensity guild wars on large servers.

MillhioreBT:
Hi @SaiyansKing what is the correct test to verify that these changes are better than the previous code?
btw I would like to see your implementation in your optimized TFS

DSpeichert (dismissed):

DSpeichert (approved):

MillhioreBT (approved):

SaiyansKing:
> Hi @SaiyansKing what is the correct test to verify that these changes are better than the previous code?
I don't know what you mean by "better" but the thing is you expect it to give you free performance and that isn't the point of using asio deadline timers, the performance difference you get is because someone decided that it is better to put std::deque to std::priority_queue than std:;vector but std::vector performs much better(even boost asio use vector as underlying container for priority_queue) so if you would compare those changes on older tfs versions that used std::vector for std::priority_queue you wouldn't get a bit of performance gain.

> btw I would like to see your implementation in your optimized TFS
I don't use scheduler class at all since it is useless so it'll be difficult xD

ranisalt:
Yes, this change is not a performance improvement, it's a code cleanup by leveraging a library with dedicated code that we already include.

As @SaiyansKing said, for future improvements, a lot of bottlenecks are related to poor choice of data structures. In general, we should avoid a list when a deque could be used, and avoid deques and forward_list when a vector could be used.

gesior:
@ranisalt 
2 hosters reported to me random crashes on newest TFS. One sent backtrace, but without debug info:
https://otland.net/threads/tfs-service-main-process-exited-code-killed-status-11-segv.274811/#post-2646548
Can it be related to these changes?

ranisalt:
@gesior it could be, it appears on the stack trace, but this change by itself does not mess with pointers. Most likely they used addEvent with a pointer.

yamaken93:
@ranisalt https://github.com/otland/forgottenserver/blob/master/src/scheduler.cpp#L34 there is a data race here with multiple threads. That is why in my original code(which you used as base for this pr) i did the timer creation inside a lambda that was only run by the scheduler thread and i made sure that both the containers that handled timers were only accessed by scheduler thread so it was thread safe.

ranisalt:
@yamaken93 thanks for the heads-up, I didn't notice `addEvent` could be called from other threads too.

MillhioreBT:
> > Hi @SaiyansKing what is the correct test to verify that these changes are better than the previous code?
> 
> I don't know what you mean by "better" but the thing is you expect it to give you free performance and that isn't the point of using asio deadline timers, the performance difference you get is because someone decided that it is better to put std::deque to std::priority_queue than std:;vector but std::vector performs much better(even boost asio use vector as underlying container for priority_queue) so if you would compare those changes on older tfs versions that used std::vector for std::priority_queue you wouldn't get a bit of performance gain.
> 
> > btw I would like to see your implementation in your optimized TFS
> 
> I don't use scheduler class at all since it is useless so it'll be difficult xD

You are absolutely right!
With the fixes proposed in #3325 you no longer get the performance I mentioned above, it continues to work with the same performance of a lifetime

PR #3279, opened 2021-01-20
Missing reference to damageEntry in Creature:getDamageMap

nekiro:
Avoid copying the damage entry when iterating.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -7717,7 +7717,7 @@ int LuaScriptInterface::luaCreatureGetDamageMap(lua_State* L)
 	}
 
 	lua_createtable(L, creature->damageMap.size(), 0);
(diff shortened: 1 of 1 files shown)


ranisalt:
`const auto&`

src/luascript.cpp
```diff
@@ -7717,7 +7717,7 @@ int LuaScriptInterface::luaCreatureGetDamageMap(lua_State* L)
 	}
 
 	lua_createtable(L, creature->damageMap.size(), 0);
-	for (auto damageEntry : creature->damageMap) {
+	for (auto& damageEntry : creature->damageMap) {
```
ranisalt, line 7720: `const auto&`

ranisalt (changes requested):

ranisalt (approved):

PR #3281, opened 2021-01-20
Several monsters fixes

EPuncker:
- Mostly fixed gold coin **countmax** to work as intended with values higher than 100 (monsters were dropping only 100)
- Infected Weeper summons only 1 parasite now instead of 6
- Changed a few loot to use ID instead of name to prevent duplicate name conflict in the future (amphora, inkwell, closed trap, dirty cape and crystal rubbish)
- Removed two redundant max summon

### DIFF
--- a/data/monster/monsters/ancient_scarab.xml
+++ b/data/monster/monsters/ancient_scarab.xml
@@ -52,7 +52,8 @@
 		<summon name="Larva" interval="2000" chance="10" />
 	</summons>
 	<loot>
-		<item name="gold coin" countmax="187" chance="100000" />
+		<item name="gold coin" countmax="100" chance="50000" />
+		<item name="gold coin" countmax="87" chance="50000" />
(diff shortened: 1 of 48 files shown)


nekiro:
Honestly why are we updating it? We should have some template monsters (like npc) and that's it imo.

EPuncker:
> Honestly why are we updating it? We should have some template monsters (like npc) and that's it imo.

Because currently there is no monster pack released in the forum so I want tfs to provide a solid base and because I don't mind doing it as don't have anything else to do :D

MillhioreBT:
> > Honestly why are we updating it? We should have some template monsters (like npc) and that's it imo.
> 
> Because currently there is no monster pack released in the forum so I want tfs to provide a solid base and because I don't mind doing it as don't have anything else to do :D

Your answer is quite precise, doing this for anyone is very tedious, but since you take the trouble, it is very good that the package is improving in this regard thanks you 🤩

DSpeichert (approved):

PR #3284, opened 2021-01-21
Added warning to invalid countmax in Monsters::loadLootItem

EPuncker:


src/monsters.cpp
```diff
@@ -1385,6 +1385,11 @@ bool Monsters::loadLootItem(const pugi::xml_node& node, LootBlock& lootBlock)
 	}
 
 	if ((attr = node.attribute("countmax"))) {
+		int32_t lootCountmax = pugi::cast<int32_t>(attr.value());
+		if (lootCountmax > 100) {
```
marmichalski, line 1389: what about negative value? 🌞

src/monsters.cpp
```diff
@@ -1385,6 +1385,11 @@ bool Monsters::loadLootItem(const pugi::xml_node& node, LootBlock& lootBlock)
...
+			std::cout << "[Warning - Monsters::loadMonster] Invalid \"countmax\" "<< lootCountmax <<" used for loot, the max allowed value is 100. " << std::endl;
+			return false;
+		}
 		lootBlock.countmax = std::max<int32_t>(1, pugi::cast<int32_t>(attr.value()));
```
marmichalski, line 1393: you can reuse the variable `lootCountmax` here, which should prolly be written as `lootCountMax`

src/monsters.cpp
MillhioreBT: There is no need to handle negative values ​​if the value is extracted and passed through the function:
```c++
std::max<int32_t>(1, pugi::cast<int32_t>(attr.value()));
```
first.

marmichalski: Then why not use `std::clamp` and do not warn at all? I don't see a reason to warn about too high value, but not warn about too low value. It should be either both or none at all. 🤷

DSpeichert: A warning could be useful although it's quite unlikely that someone will try to put in a negative value for loot.

DSpeichert (dismissed):

src/monsters.cpp
EPuncker: something like this?
```
		if (lootCountMax > 100 || lootCountMax < 1) {
			std::cout << "[Warning - Monsters::loadMonster] Value for \"countmax\" "<< lootCountMax <<" out of bounds. " << std::endl;
			return false;
		}
```

DSpeichert (approved):

soul4soul:
How about we merge #2875 instead?

EPuncker:
> How about we merge #2875 instead?

don't worry, this is just a temporary measure while #2906 doesn't get in

PR #3288, opened 2021-01-25
Add config information for despawning monsters

### LINKED ISSUE #3247: deSpawn config needs a description and more
Currently we have **deSpawnRange** and **deSpawnRadius** at config.lua, but they have no description at all, so new users (even old users like me) will have a hard time figuring out what they do...

Also we need to add a new config so the user can chose if he wants the creature to disappear (poff) or to be teleported back to its spawn when it reaches deSpawnRadius value, currently we only teleport it back...

EPuncker:
Co-Authored-By: gigastar <26754137+gigastar@users.noreply.github.com>

### Pull Request Prelude

### Changes Proposed

- Added config.lua description to despawning settings
- Added setting to chose if monster should be removed or teleported back to its spawn

**Issues addressed:** closes #3247


src/monster.cpp
```diff
@@ -724,8 +726,13 @@ void Monster::onThink(uint32_t interval)
 	}
 
 	if (!isInSpawnRange(position)) {
-		g_game.internalTeleport(this, masterPos);
-		setIdle(true);
+		bool removeMonster = g_config.getBoolean(ConfigManager::REMOVE_ON_DESPAWN);
```
nekiro, line 729: you can use this g_config.getBoolean before defining this variable

nekiro (changes requested):

config.lua.dist
```diff
@@ -122,9 +122,13 @@ rateLoot = 2
...
 rateSpawn = 1
 
--- Monsters
+-- Monster Despawn Config
+-- despawnRange is the amount of floors a monster can be from its spawn position
+-- despawnRadius is how many tiles away it can be from its spawn position
+-- removeOnDespawn will remove the monster if true or teleport it back to its spawn position if false
 deSpawnRange = 2
 deSpawnRadius = 50
+removeOnDespawn = true
```
marmichalski, line 131: IMHO this should default to false to keep previous mechanism.

EPuncker: but previous mechanism was "wrong" and also with deSpawnRadius set to 50 I bet nobody would ever reach this anyway unless intentionally trying to lure a monster, I had to change mine to 20 because unless you are in a real map your players would hardly notice a change

EPuncker: leaving it as true because it is the new default setting

EPuncker:
depois2
actually tested

config.lua.dist
marmichalski: Wrong according to what? This is an MMORPG server emulator, not a cipbia clone 🤔

src/monster.cpp
```diff
@@ -724,8 +726,13 @@ void Monster::onThink(uint32_t interval)
 	}
 
 	if (!isInSpawnRange(position)) {
-		g_game.internalTeleport(this, masterPos);
-		setIdle(true);
+		if (g_config.getBoolean(ConfigManager::REMOVE_ON_DESPAWN)) {
+			g_game.addMagicEffect(this->getPosition(), CONST_ME_POFF);
+			g_game.removeCreature(this, false);
+		} else {
+			g_game.internalTeleport(this, masterPos);
```
ranisalt, line 733: I think the magic effect should be added no matter the config. The removal is transparent from the player, so visually it should be the same.

ranisalt (changes requested):

src/monster.cpp
```diff
@@ -727,12 +727,14 @@ void Monster::onThink(uint32_t interval)
...
 		if (g_config.getBoolean(ConfigManager::REMOVE_ON_DESPAWN)) {
-			g_game.addMagicEffect(this->getPosition(), CONST_ME_POFF);
 			g_game.removeCreature(this, false);
 		} else {
 			g_game.internalTeleport(this, masterPos);
 			setIdle(true);
 		}
+
+		g_game.addMagicEffect(this->getPosition(), CONST_ME_POFF);
+
```
ranisalt, line 737: Extra line

ranisalt (changes requested):

ranisalt (approved):

DSpeichert (approved):

PR #3292, opened 2021-01-30
Replace hand-rolled algorithms in spectator container

ranisalt:
### Pull Request Prelude

### Changes Proposed

Instead of writing algorithms by hand, this replaces with functions already present in the standard library. Additionally, fixes a potential duplication issue when concatenating with the cache.

If there is a specific reason to use `insert` instead of the usual `addSpectators`, then I have a workaround in mind, too.

**Issues addressed:** 
None

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -419,7 +419,7 @@ void Map::getSpectators(SpectatorVec& spectators, const Position& centerPos, boo
 			if (it != playersSpectatorCache.end()) {
 				if (!spectators.empty()) {
 					const SpectatorVec& cachedSpectators = it->second;
-					spectators.insert(spectators.end(), cachedSpectators.begin(), cachedSpectators.end());
+					spectators.addSpectators(cachedSpectators);
 				} else {
 					spectators = it->second;
 				}
@@ -434,7 +434,7 @@ void Map::getSpectators(SpectatorVec& spectators, const Position& centerPos, boo
 				if (!onlyPlayers) {
 					if (!spectators.empty()) {
 						const SpectatorVec& cachedSpectators = it->second;
-						spectators.insert(spectators.end(), cachedSpectators.begin(), cachedSpectators.end());
+						spectators.addSpectators(cachedSpectators);
 					} else {
 						spectators = it->second;
 					}
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -418,8 +418,7 @@ void Map::getSpectators(SpectatorVec& spectators, const Position& centerPos, boo
 			auto it = playersSpectatorCache.find(centerPos);
 			if (it != playersSpectatorCache.end()) {
 				if (!spectators.empty()) {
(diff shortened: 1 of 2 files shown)


nekiro:
Looks way cleaner I like it.

MillhioreBT:
Very nice changes, for a future path towards C++17 ;)

DSpeichert (approved):

src/map.cpp
```diff
@@ -419,7 +419,7 @@ void Map::getSpectators(SpectatorVec& spectators, const Position& centerPos, boo
 			if (it != playersSpectatorCache.end()) {
 				if (!spectators.empty()) {
 					const SpectatorVec& cachedSpectators = it->second;
-					spectators.insert(spectators.end(), cachedSpectators.begin(), cachedSpectators.end());
+					spectators.addSpectators(cachedSpectators);
```
MillhioreBT, line 422: I'm not sure, but I think you could even save an extra line, if you do it like this:
```cpp
spectators.addSpectators(static_cast<const SpectatorVec&>(it->second));
```
it is right?

MillhioreBT: I am not using my suggestions!
I am using exactly the changes that you post, and I have a problem with the function: `canSee`, apparently the spectators  they don't stay alive on the memory, causing a problem when obtaining for example the position:

Here I give a small explanation, at first I thought it was because of other changes, but I definitely checked 100% that the culprit is these changes:

https://github.com/otland/forgottenserver/pull/3272#pullrequestreview-582063937

ranisalt: They don't, `SpectatorVec` just holds (non-owning) pointers, your creatures are being deleted elsewhere.

src/spectators.h
```diff
@@ -35,46 +35,31 @@ class SpectatorVec
...
-				vec[len - 1] = vec[i];
-				vec[i] = tmp;
-				vec.pop_back();
-				break;
-			}
+		auto it = std::find(vec.begin(), vec.end(), spectator);
+		if (it == end()) {
+			return;
 		}
+		std::iter_swap(it, end());
```
MillhioreBT, line 52: change
`std::iter_swap(it, end());`
for
`std::iter_swap(it, end() - 1);`

This is the problem that I had and I had not noticed it.
remember that the function exchanges two iterators in valid ranges, `end()` is not a valid range, therefore the vector does not change at all, instead if you do `end() - 1` it will return the last valid position

;)

MillhioreBT (changes requested):

src/map.cpp
MillhioreBT: Just change to: `spectators.addSpectators(it->second);` since the same function converts the value into a read-only reference automatically, I have tried it and if it works, the conversion actually happens.

MillhioreBT (approved):
I give it my approval, although it is worthless, maybe I can encourage others to try this and check!

EPuncker (approved):

PR #3299, opened 2021-02-04
Update doors.lua

Xikini:
Locked Doors in rl tibia allow you to lock them while they are open using the key.. provided that the door has no items or creatures inside of it.
This alteration allows for this to happen.

A separate but more complicated issue is that doors in rl tibia have a toggle state, of locked or not locked.
When in the unlocked state the door can be open and closed normally, without using a key.

I can't think of an easy way to add this functionality.
If someone has an idea.. Let me know.

### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 


nekiro:
There is an very easy way. Hold door state in new attribute or custom attribute. I would prefer new ITEM_ATTRIBUTE_DOORSTATE

marmichalski:
```
I can't think of an easy way to add this functionality.
```

Doesn't they use different item id for locked/unlocked door? IIRC this is how it was also done in TFS 0.4.

data/scripts/actions/others/doors.lua
```diff
@@ -79,9 +79,14 @@ function door.onUse(player, item, fromPosition, target, toPosition, isHotkey)
 			return false
 		end
 		target = tile:getTopVisibleThing()
-		if target.actionid > 0 then
+		if target.actionid > 0 and not table.contains(keys, target.itemid) then
 			if item.actionid == target.actionid then
+				local DOORSTATE = target:getCustomAttribute(ITEM_ATTRIBUTE_DOORSTATE) == DOOR_STATE_UNLOCKED and DOOR_STATE_LOCKED or DOOR_STATE_UNLOCKED
```
nekiro, line 84: this is very ugly, please use normal camelCase coding style here

nekiro (changes requested):

Xikini:
> ```
> I can't think of an easy way to add this functionality.
> ```
> 
> Doesn't they use different item id for locked/unlocked door? IIRC this is how it was also done in TFS 0.4.

I think they changed the door sprite for locked vs unlocked doors.
It seems to have a giant padlock on it, and I can't seem to find a third door of this type.

So unless that third door exists somewhere, this is no longer a viable solution.

data/scripts/actions/others/doors.lua
Xikini: Hopefully this is what you meant?

nekiro: yeah, great

EPuncker:
actually most (if not all) of the doors have 3 types:

```
	<item id="5278" article="a" name="closed door" editorsuffix=" (Locked)">
		<attribute key="type" value="door" />
		<attribute key="blockprojectile" value="1" />
		<attribute key="description" value="It is locked." />
	</item>
	<item id="5279" article="a" name="closed door">
		<attribute key="type" value="door" />
		<attribute key="blockprojectile" value="1" />
	</item>
	<item id="5280" article="an" name="open door">
		<attribute key="type" value="door" />
	</item>
```
s

Xikini:
> 5278

Ah, I see.
Normal doors have a 3rd one, but 'Locked' doors (aka padlock doors) only have 2 variations.

With my proposed changes, these doors are also locked/unlocked correctly.

Untitled

soul4soul:
Maybe cip only uses keys on doors which have 3 item ids? Do you know any locations in cipbia where a key can be used on a padlock door?

Xikini:
> Maybe cip only uses keys on doors which have 3 item ids? Do you know any locations in cipbia where a key can be used on a padlock door?

If there is actually no padlock doors that a key can be used on, it would make quite a bit more sense the way the doors are structured.
From memory, I think padlock doors are only used for houses?

I don't really have a cipbia character to use to check though.

EPuncker:
as far as I am concerned, padlock doors are exclusive to houses, only "normal" doors have key

marmichalski:
> as far as I am concerned, padlock doors are exclusive to houses, only "normal" doors have key

I'm also quite certain that's how it is.

Xikini:
Does anyone know / can anyone test if when you unlock a door in rl tibia.. Does it stay closed but unlock, or unlock + open when the key is used?

I can technically test it myself.. but I'll need to make an account.. xD

Xikini:
Was able to test and confirm.

Use key on locked door -> opens door

Xikini:
Alright, had to make giant changes to the script to allow for the new functionality.

Technically House and Extra doors could be merged, but would make updates more difficult.

open Quest and Level doors could also be merged, but figured keeping them apart made more sense.
open Quest and Level Doors technically are not being used inside of doors.lua, so don't need to be a global.. either.

Doors functions, explained.

openDoors -> When used transforms into a closedDoors. If ActionId is present, and a key with the same ActionId is used on it, transforms into a lockedDoors.
closedDoors -> When used transforms into a openDoors. If ActionId is present, and a key with the same ActionId is used on it, transforms into a lockedDoors.
lockedDoors -> When used sends a green message "It is locked.". If ActionId is present, and a key with the same ActionId is used on it, transforms into a openDoors.

-- Doors that do not have a third, lockable door.
openExtraDoors -> When used transforms into a closedExtraDoors. ActionId has no effect on it's function.
closedExtraDoors -> When used transforms into a openExtraDoors. ActionId has no effect on it's function.

-- Can only be used if you are the houses owner.
openHouseDoors -> When used transforms into a closedHouseDoors. ActionId has no effect on it's function.
closedHouseDoors -> When used transforms into a openHouseDoors. ActionId has no effect on it's function.

openQuestDoors -> Unable to be used. Closes automatically when a player leaves the tile.
closedQuestDoors -> When used sends a green message "The door seems to be sealed against unwanted intruders.". If ActionId is present, checks storage of player. If not -1, then allow access.

openLevelDoors -> Unable to be used. Closes automatically when a player leaves the tile.
closedLevelDoors -> When used sends a green message "Only the worthy may pass.". ActionId has no effect on it's function. If ActionId is present, checks level of player. If playerLevel >= (actionid - 1000) then allow access.

PR #3300, opened 2021-02-07
Updating Doors.

Xikini:
I'm a noob at github and broke my old repository by deleting items.xml.. then re-uploading the file, which then showed 36,000+ line changes. Couldn't figure out how to undo the change.. so I deleted the repository and re-forked.. but then github started showing unknown repository.. even though it's the same named one...

So this is a duplicate pull request from https://github.com/otland/forgottenserver/pull/3299
xD

----
Locked Doors in rl tibia allow you to lock them while they are open using the key.. provided that the door has no items or creatures inside of it.

A separate but more complicated issue is that doors in rl tibia have a toggle state, of locked or not locked.
When in the unlocked state the door can be open and closed normally, without using a key.

---
The changes to doors.lua addresses both of these issues.
global.lua needed to be changed so that we could differentiate normal / extra / house doors from each other.
items.xml changes are a minor change to a few doors, to show the (locked) status on correct doors.

Generally, it was realised that house doors (padlock) doors should never be locked.
Extra doors, that do not currently have a 3rd lockable door cannot be locked either.

Doors with 2 closed 1 open door, now correctly allow 1/2 of the closed doors to be 'locked'.

Additional changes to quest & level doors in global.lua, to bring it in line with the rest of the updated global variable names.

While testing, there were quite a few doors in wrong categories or duplicated entries, so those were fixed as well.

### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 


Xikini:
http://s000.tinyupload.com/?file_id=06035785087675847094

Door testing ground in remere's.
Position(250, 250, 7)

MillhioreBT:
This check is pointless, in fact it is more time consuming.
```lua
if doorsSet[d] == nil then
doorsSet[d] = true
end
```
you can just do without `if`.
```lua
doorsSet[d] = true
```
Anyway, this script only runs when the scripts are load or reload and there shouldn't be much of a problem, it will serve as an advice for programming in lua

Xikini:
> This check is pointless, in fact it is more time consuming.
> 
> ```lua
> if doorsSet[d] == nil then
> doorsSet[d] = true
> end
> ```
> 
> you can just do without `if`.
> 
> ```lua
> doorsSet[d] = true
> ```
> 
> Anyway, this script only runs when the scripts are load or reload and there shouldn't be much of a problem, it will serve as an advice for programming in lua

I tried to just copy-paste that part, so that it stayed in line with how it was previously... but since you mentioned it... xD
I changed it to fit my normal coding style.

data/items/items.xml
```diff
@@ -2452,7 +2456,7 @@
 		<attribute key="slotType" value="ring" />
 		<attribute key="decayTo" value="0" />
 		<attribute key="transformDeEquipTo" value="2167" />
-		<attribute key="duration" value="600" />
+		<attribute key="duration" value="600000" />
```
EPuncker, line 2459: what is this? 😛

Xikini: a testing ring I forgot to remove. xD
Sorry >.<

soul4soul:
I haven't tested but I appreciate the effort. I always found the way locked doors worked in OTs to be frustrating. PS. You can attach small files directly to github messages instead of using a 3rd party hoster.

Xikini:
> I haven't tested but I appreciate the effort. I always found the way locked doors worked in OTs to be frustrating. PS. You can attach small files directly to github messages instead of using a 3rd party file host.

I tried. xD
It didn't like my file types. xP

EPuncker:
I have two suggestions tho while you are at it:

- a new indentation to tables at global.lua (like [this](https://github.com/orts/server/blob/master/data/global.lua#L5))
- adding some missing doors, these are the ids I've gathered so far:

```
12703 - 12701
12695 - 12692
22814 - 22831
25283 - 25296
25158 - 25165
20273 - 20290
18208 - 18209
14633 - 14635
```

Xikini:
Updated the indentation.

The new door id's is a large project by itself and should go in a new pull request.
This pull is only meant to update the locking feature of doors to match real tibia's locking mechanism.

EPuncker (approved):

PR #3301, opened 2021-02-07
Fix pagination in sendAddContainerItem

ninjalulz:
This should fix an issue with pagination upon adding items.

It was vaguely reported [here](https://otland.net/threads/the-reason-why-tfs-is-delayed-we-let-the-great-programmers-goes-away.274114/page-3#post-2645334) "first".


EPuncker (approved):

MillhioreBT:
I had already seen this in another repository, I'm going to test it, it sure works fine!

PR #3305, opened 2021-02-08
Replaced number with constant in offline training code

EPuncker:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -58,15 +58,15 @@ extern Scripts* g_scripts;
 
 Game::Game()
 {
+	offlineTrainingWindow.defaultEnterButton = 1;
+	offlineTrainingWindow.defaultEscapeButton = 0;
 	offlineTrainingWindow.choices.emplace_back("Sword Fighting and Shielding", SKILL_SWORD);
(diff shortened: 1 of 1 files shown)


soul4soul (approved):

Erza (approved):

marmichalski:
how is that a constant? 🤔

MillhioreBT:
The correct term would be, change the constant for a variable XD
a number is a constant, since its value cannot be changed, 1 will always be 1, while a variable can change its value, in case the variable is not declared as a constant `const`

Anyway the changes are fine, and do not cause any problems

soul4soul:
> 
> 
> The correct term would be, change the constant for a variable XD
> a number is a constant, since its value cannot be changed, 1 will always be 1, while a variable can change its value, in case the variable is not declared as a constant `const`
> 
> Anyway the changes are fine, and do not cause any problems

I would use "magic number" to describe the initial implementation instead of "constant" and it's being replaced with a variable which is properly named.

DSpeichert (approved):
LGTM

MillhioreBT (approved):

DSpeichert:
Next time make sure the commit message makes sense

Issue #3025, opened 2020-05-05
Reloading spells - crash


infister:
### Before creating an issue, please ensure:
- [X] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [X] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. add spell with addEvent (example below)
2. cast it
3. /reload spells before spell executes combat:execute()

### Actual behaviour

Segfault

```
#0  0x000056098a1f2106 in AreaCombat::getArea(Position const&, Position const&) const ()
#1  0x000056098a1efa8e in AreaCombat::getList(Position const&, Position const&, std::forward_list<Tile*, std::allocator<Tile*> >&) const ()
#2  0x000056098a1ebb9c in Combat::getCombatArea(Position const&, Position const&, AreaCombat const*, std::forward_list<Tile*, std::allocator<Tile*> >&) ()
```

### Reproduction environment

Spell:
```
local combat = Combat()
combat:setParameter(COMBAT_PARAM_TYPE, COMBAT_PHYSICALDAMAGE)
combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_FIREAREA)
combat:setArea(createCombatArea({{1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,3,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
                                {1,1,1,1,1,1,1,1,1,1,1,1,1,1,1},
[...]

PR #3309, opened 2021-02-11
Fix spells addevent crash after reload

MillhioreBT:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** #3025

The author of these changes is: @SaiyansKing all credits to this person.

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -77,7 +77,7 @@ CombatDamage Combat::getCombatDamage(Creature* creature, Creature* target) const
 	return damage;
 }
 
-void Combat::getCombatArea(const Position& centerPos, const Position& targetPos, const AreaCombat* area, std::forward_list<Tile*>& list)
+void Combat::getCombatArea(const Position& centerPos, const Position& targetPos, const AreaCombat* area, std::vector<Tile*>& list)
 {
 	if (targetPos.z >= MAP_MAX_LAYERS) {
 		return;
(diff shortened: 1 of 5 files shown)


nekiro:
It's possible to avoid combat crashes just by using the table instead of the object itself, like this:
Combat.execute(combat, creature, variant)

MillhioreBT:
> It's possible to avoid combat crashes just by using the table instead of the object itself, like this:
> Combat.execute(combat, creature, variant)

I have proven what you say, and if you are right about one thing, it does not produce the crash!
however it does not have the expected behavior, the combat is not supposed to be lost, much less to execute a different combat!

The following code has a strange behavior when reload spells before executions.
```lua
addEvent(Combat.execute, 5000, combat2, creature:getId(), variant)
addEvent(Combat.execute, 7000, combat, creature:getId(), variant)
```
when I say strange, I mean that any combat is executed and not the combat that it really is.

**Important: all this happens with master sources!**

MillhioreBT:
These changes allow this:
https://user-images.githubusercontent.com/28090948/126395353-b02dec76-016d-49fe-9620-72060b1f543a.mp4

For now there are some merge conflicts, but I can do another PR if necessary, I just need to know if this really they interests

MillhioreBT:
Closed, because #3503 seems like a better option, but it has yet to be decided

PR #3310, opened 2021-02-12
replace std::foward_list with std::vector in creature walking

yamaken93:
### Pull Request Prelude

### Changes Proposed

Now creature walking is using std::vector which is way faster than std::foward_list. This improves TFS since its a bad example/usage of data structures + its a highly used code since in a game server creatures are walking all the time.

Creature::startAutoWalk() has been added so we don't need to copy the walk dir vector again since te code was already using the Creature instance listWalkDir.
std::move has been used in some places where its binded by std::bind since the walk dir vector was a temp that would be wasted for no reason.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -915,7 +915,7 @@ void Creature::goToFollowCreature()
 
 			if (dir != DIRECTION_NONE) {
 				listWalkDir.clear();
-				listWalkDir.emplace_back(dir);
+				listWalkDir.push_back(dir);
 
 				hasFollowPath = true;
 				startAutoWalk();
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -260,7 +260,20 @@ void Creature::startAutoWalk()
 		return;
 	}
 
-	addEventWalk(listWalkDir.size() > 0);
+	addEventWalk(listWalkDir.size() == 1);
+}
+
+void Creature::startAutoWalk(Direction direction)
+{
+	Player* player = getPlayer();
+	if (player && player->isMovementBlocked()) {
+		player->sendCancelWalk();
+		return;
+	}
+
+	listWalkDir.clear();
+	listWalkDir.push_back(direction);
+	addEventWalk(true);
 }
 
 void Creature::startAutoWalk(const std::vector<Direction>& listDir)
@@ -271,8 +284,8 @@ void Creature::startAutoWalk(const std::vector<Direction>& listDir)
 		return;
 	}
 
-	listWalkDir.assign(listDir.begin(), listDir.end());
-	addEventWalk(listWalkDir.size() > 0);
+	listWalkDir = listDir;
+	addEventWalk(listWalkDir.size() == 1);
 }
 
 void Creature::addEventWalk(bool firstStep)
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -246,13 +246,13 @@ bool Creature::getNextStep(Direction& dir, uint32_t&)
 		return false;
 	}
 
-	dir = listWalkDir.front();
-	listWalkDir.pop_front();
+	dir = listWalkDir.back();
+	listWalkDir.pop_back();
 	onWalk(dir);
 	return true;
 }
 
-void Creature::startAutoWalk(const std::forward_list<Direction>& listDir)
+void Creature::startAutoWalk()
 {
 	Player* player = getPlayer();
 	if (player && player->isMovementBlocked()) {
@@ -260,13 +260,19 @@ void Creature::startAutoWalk(const std::forward_list<Direction>& listDir)
 		return;
 	}
 
-	listWalkDir = listDir;
+	addEventWalk(listWalkDir.size() > 0);
+}
 
-	size_t size = 0;
-	for (auto it = listDir.begin(); it != listDir.end() && size <= 1; ++it) {
-		size++;
+void Creature::startAutoWalk(const std::vector<Direction>& listDir)
+{
+	Player* player = getPlayer();
+	if (player && player->isMovementBlocked()) {
+		player->sendCancelWalk();
+		return;
 	}
-	addEventWalk(size == 1);
+
+	listWalkDir.assign(listDir.begin(), listDir.end());
+	addEventWalk(listWalkDir.size() > 0);
 }
 
 void Creature::addEventWalk(bool firstStep)
@@ -899,7 +905,7 @@ void Creature::goToFollowCreature()
 					listWalkDir.clear();
 					if (getPathTo(followCreature->getPosition(), listWalkDir, fpp)) {
 						hasFollowPath = true;
-						startAutoWalk(listWalkDir);
+						startAutoWalk();
 					} else {
 						hasFollowPath = false;
 					}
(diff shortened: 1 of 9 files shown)


src/creature.cpp
```diff
@@ -246,27 +246,33 @@ bool Creature::getNextStep(Direction& dir, uint32_t&)
...
+{
+	Player* player = getPlayer();
+	if (player && player->isMovementBlocked()) {
+		player->sendCancelWalk();
+		return;
 	}
-	addEventWalk(size == 1);
+
+	listWalkDir.assign(listDir.begin(), listDir.end());
+	addEventWalk(listWalkDir.size() > 0);
```
ranisalt, line 275:
```suggestion
	addEventWalk(!listWalkDir.empty());
```

src/creature.cpp
```diff
@@ -246,27 +246,33 @@ bool Creature::getNextStep(Direction& dir, uint32_t&)
...
+void Creature::startAutoWalk(const std::vector<Direction>& listDir)
+{
+	Player* player = getPlayer();
+	if (player && player->isMovementBlocked()) {
+		player->sendCancelWalk();
+		return;
 	}
-	addEventWalk(size == 1);
+
+	listWalkDir.assign(listDir.begin(), listDir.end());
```
ranisalt, line 274:
```suggestion
	listWalkDir = listDir;
```

src/creature.cpp
```diff
@@ -909,16 +915,16 @@ void Creature::goToFollowCreature()
 
 			if (dir != DIRECTION_NONE) {
 				listWalkDir.clear();
-				listWalkDir.push_front(dir);
+				listWalkDir.emplace_back(dir);
```
ranisalt, line 918:
```suggestion
				listWalkDir.push_back(dir);
```

`emplace*` only makes sense when constructing in-place (i.e. passing constructor parameters).

src/game.cpp
```diff
@@ -1823,7 +1823,7 @@ void Game::playerMove(uint32_t playerId, Direction direction)
 	player->resetIdleTime();
 	player->setNextWalkActionTask(nullptr);
 
-	player->startAutoWalk(std::forward_list<Direction> { direction });
+	player->startAutoWalk(std::vector<Direction> { direction });
```
ranisalt, line 1826:
```suggestion
	player->startAutoWalk({ direction });
```

Should work as well

src/luascript.cpp
```diff
@@ -7823,12 +7823,12 @@ int LuaScriptInterface::luaCreatureGetPathTo(lua_State* L)
...
 
-	std::forward_list<Direction> dirList;
+	std::vector<Direction> dirList;
 	if (creature->getPathTo(position, dirList, fpp)) {
 		lua_newtable(L);
 
 		int index = 0;
-		for (Direction dir : dirList) {
+		for (Direction dir : boost::adaptors::reverse(dirList)) {
 			lua_pushnumber(L, dir);
```
ranisalt, line 7832: I would rather use
```suggestion
		for (auto it = dirList.rbegin(); it != dirList.rend(); ++it) {
			lua_pushnumber(L, *it);
```
```

File: src/map.cpp [hunk 6]
```diff
@@ -761,21 +761,21 @@ bool Map::getPathMatching(const Creature& creature, std::forward_list<Direction>
...
+			dirList.emplace_back(DIRECTION_WEST);
 		} else if (dx == -1) {
-			dirList.push_front(DIRECTION_EAST);
+			dirList.emplace_back(DIRECTION_EAST);
 		} else if (dy == 1) {
-			dirList.push_front(DIRECTION_NORTH);
+			dirList.emplace_back(DIRECTION_NORTH);
 		} else if (dy == -1) {
-			dirList.push_front(DIRECTION_SOUTH);
+			dirList.emplace_back(DIRECTION_SOUTH);
```
ranisalt, line 778: `push` them all

src/map.h
```diff
@@ -260,7 +260,7 @@ class Map
 
 		const Tile* canWalkTo(const Creature& creature, const Position& pos) const;
 
-		bool getPathMatching(const Creature& creature, std::forward_list<Direction>& dirList,
+		bool getPathMatching(const Creature& creature, std::vector<Direction>& dirList,
```
ranisalt, line 263: I hate this function signature as much as possible. If you are willing, please make it return `std::pair<std::vector<Direction> dirList, bool>` and remove this in-place argument.

ranisalt (changes requested):
There are some more lines with the same issues too, but I love this change

src/map.h
yamaken93: I don't believe i can change this for performance reasons. It should use a reference to the vector so we could re-use already allocated vectors.

src/creature.cpp
yamaken93: Fixed. To keep the old behavior(before this pr) it should be addEventWalk(listWalkDir.size() == 1);

src/creature.cpp
yamaken93: Fixed.

src/creature.cpp
yamaken93: Fixed.

src/game.cpp
yamaken93: Fixed. Added a version which takes Direction as parameter instead of requiring to create a new vector just for one direction(copied from @SaiyansKing).

src/luascript.cpp
yamaken93: Fixed.

src/map.cpp
yamaken93: Fixed.

src/map.h
ranisalt: But in all usages of these functions, you need to either instantiate a new vector specifically to hold this, or you clear a previously allocated vector which would have the same performance impact. Maybe I missed something here?

yamaken93: In few places we directly use the Creature::listWalkDir so no reallocation needed.

DSpeichert (approved):

src/map.h
ranisalt: I see, but the move assignment for vector, which would be triggered when you run `listWalkDir = getPathTo(...)` is exactly the same as running `assign`: it only swaps the pointers.

ranisalt: I just checked here:

1) `bool Creature::getPathTo(const Position& targetPos, std::forward_list<Direction>& dirList, int32_t minTargetDist, int32_t maxTargetDist, bool fullPathSearch /*= true*/, bool clearSight /*= true*/, int32_t maxSearchDist /*= 0*/) const` is *always* called with a brand new container, so it could return a new vector with the same performance.
2) `bool Creature::getPathTo(const Position& targetPos, std::forward_list<Direction>& dirList, const FindPathParams& fpp) const` sometimes reuse a previous vector.
So maybe you can make 1) work in place, and 2) return a new vector? Not sure if it's worth the hassle though.

yamaken93: I got few doubts about this. What is faster? moving the vector or copying a 10-15 byte(steps) memory buffer? This move/copy case mostly happens for player walking and actions code.
My hope for this optimization was to keep the same vector allocated(Creature::listWalkDir) specially for monsters since for every 100 players online there may be 500 or 1000 monsters calculating a new path. That could prevent millions of calls to the allocator which can be expensive(new and and free are in the top 5 tfs most used functions in perf top).

ranisalt: Copy elision would kick in here. When you return a vector, the compiler omits copying the returned vector into the assigned vector, performing a move instead (it calls `operator=(const vector&&)`). This is possible because a return value that is not taken as a reference would be immediately destroyed after the return.

ranisalt (approved):

PR #3319, opened 2021-02-14
replace deadline_timer with steady_timer

yamaken93:
### Pull Request Prelude

### Changes Proposed

Replace deadline_timer with steady_timer so we use std::chrono::steady_clock. Its make senses for tfs to use steady_clock since all the intervals are delays from now independently of the system time.
Why not high_resolution_timer? Seens like chrono::high_resolution_clock is not reliable across platforms.

Links:
https://stackoverflow.com/a/16364002/470739
https://www.boost.org/doc/libs/1_75_0/doc/html/boost_asio/reference/steady_timer.html
https://en.cppreference.com/w/cpp/chrono/steady_clock
https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

**Issues addressed:** 

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -116,7 +116,7 @@ void Connection::accept()
 {
 	std::lock_guard<std::recursive_mutex> lockClass(connectionLock);
 	try {
-		readTimer.expires_from_now(boost::posix_time::seconds(CONNECTION_READ_TIMEOUT));
+		readTimer.expires_from_now(std::chrono::seconds(CONNECTION_READ_TIMEOUT));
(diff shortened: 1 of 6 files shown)


MillhioreBT:
LGTM
if the clock undergoes changes the timer is no longer affected.

MillhioreBT (commented):
LGTM
if the clock undergoes changes the timer is no longer affected.

ranisalt:
Makes a lot of sense. I also like using chrono instead of boost for duration.

ranisalt (approved):

DSpeichert (approved):

Issue #3322, opened 2021-02-16
Yelling should not be sent from floor 7+ to lower floors


EPuncker:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
![depois2](https://user-images.githubusercontent.com/4684880/108012781-0660c380-6fe9-11eb-901f-be006cb7f44c.gif)

### Expected behaviour
the player from the floor bellow ground should not hear anything from above

### Actual behaviour
he hears it, same with npc yell

### Considerations
I know we are not trying to mimic cibia or anything, but this is just weird after all these years... after all its just a simple check just like https://github.com/otland/forgottenserver/commit/b52a2f799528da37b77ae0bca2148291324b5c85

MillhioreBT:
Interesting, I didn't know about this

marmichalski:
There is also this part, according to wiki: `If you yell underground, only the players on the same ground level will hear you.`

https://tibia.fandom.com/wiki/Yelling

PR #3323, opened 2021-02-16
Yelling should not be sent from floor 7+ to lower floors

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
**Issues addressed:** #3322

[REF](https://tibia.fandom.com/wiki/Yelling)

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3631,7 +3631,7 @@ bool Game::internalCreatureSay(Creature* creature, SpeakClasses type, const std:
 			              Map::maxClientViewportX, Map::maxClientViewportX,
 			              Map::maxClientViewportY, Map::maxClientViewportY);
 		} else {
-			map.getSpectators(spectators, *pos, true, false, 18, 18, 14, 14);
+			map.getSpectators(spectators, *pos, (pos->z <= 7), false, 18, 18, 14, 14);
 		}
 	} else {
 		spectators = (*spectatorsPtr);


EPuncker:
it is still showing the yelling for me after these changes

MillhioreBT:
> it is still showing the yelling for me after these changes
I guess the wiki is out of date right?

so that the changes are 100% as you propose in your Issues, it is necessary to modify the function: `Map::getSpectators` I don't have good ideas yet to modify this function, after all it is one of the most expensive functions and I would not like to make this function have more overhead.

I'll be thinking of a good idea for this

marmichalski:
I wonder if the getSpectators should even be getting creatures from the underground when you are not there. IIRC the protocol is not sending the underground tiles if you're above, then why would you ever need spectators from there? 🤔

MillhioreBT:
> I wonder if the getSpectators should even be getting creatures from the underground when you are not there. IIRC the protocol is not sending the underground tiles if you're above, then why would you ever need spectators from there? 🤔

Hi @marmichalski, how are you? I hope it's ok!

What you say is fine, but the point is that the getSpectator function only has two ways of working:
* Get a list with all the creatures on only one floor
* Get a list with all creatures from all floors

Then you understand why have to modify the function, if what you want is to make those on floor 7- listen and not those on 8+

*You're probably thinking of doing a function that iterates over the required floors and using getSpectator to get only the creatures from the desired floors, but would it really be a good idea to call this function so many times?*
*In my opinion, I think there is a better way to handle this, only now at the moment I do not have the perfect idea, and therefore I have not done anything about it, if you have any other suggestions, or any example please tell me*

marmichalski:
What I meant is to inline the behaviour of getSpectators with the one of getMapDescription from ProtocolGame.

MillhioreBT:
Now the behavior is as expected!
The parameter `multifloor` from `boolean` was modified to be `uint8_t` which is well handled as a `boolean`, this was done to allow a new search mode in the `getSpectators` function, if you are making use of this function in the future you can continue to pass it as a `boolean` parameter so they don't lose the habit ;)

ranisalt:
> The parameter `multifloor` from `boolean` was modified to be `uint8_t`

If there are options, this should at least be an enum

src/map.cpp
```diff
@@ -399,7 +399,7 @@ void Map::getSpectatorsInternal(SpectatorVec& spectators, const Position& center
 	}
 }
 
-void Map::getSpectators(SpectatorVec& spectators, const Position& centerPos, bool multifloor /*= false*/, bool onlyPlayers /*= false*/, int32_t minRangeX /*= 0*/, int32_t maxRangeX /*= 0*/, int32_t minRangeY /*= 0*/, int32_t maxRangeY /*= 0*/)
+void Map::getSpectators(SpectatorVec& spectators, const Position& centerPos, uint8_t multifloorType /*= 0*/, bool onlyPlayers /*= false*/, int32_t minRangeX /*= 0*/, int32_t maxRangeX /*= 0*/, int32_t minRangeY /*= 0*/, int32_t maxRangeY /*= 0*/)
```
DSpeichert, line 402: Like @ranisalt pointed out suggesting enums, this int-type introduces head-scratching - what is 0? what does 2 mean? What does 1 stand for?

DSpeichert (changes requested):

src/map.cpp
MillhioreBT: I like your idea, before I go to sleep, I'll make some changes and they tell me how you think :v

MillhioreBT:
I'm sure the enum names are not too good, please if you have any suggestions for the 3 names it would be very good if you could help me with that ;)

Zbizu:
Please avoid copyrighted words whenever possible.

marmichalski:
> Please avoid copyrighted words whenever possible.

lol

EPuncker:
I've been using this for quite some time already, no issues so far, just need rebase

PR #3325, opened 2021-02-16
Fix data race on addEvent

ranisalt:
### Pull Request Prelude

### Changes Proposed

[As pointed out by](https://github.com/otland/forgottenserver/pull/3278#issuecomment-779525741) @yamaken93, there is a data race when manipulating the event timer map. I didn't notice that `addEvent` could be added by several threads at once, so this wraps in a scheduler post, so the map manipulation can only happen inside the scheduler thread now.

**Issues addressed:** 
#3278 

### DIFF
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -30,7 +30,7 @@ uint32_t Scheduler::addEvent(SchedulerTask* task)
 		task->setEventId(++lastEventId);
 	}
 
-	io_context.post([this, task]() {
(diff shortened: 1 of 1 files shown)


nekiro (approved):

nekiro (approved):

MillhioreBT (approved):
Unfortunately this takes away the extra performance that it had in the previous changes, but it is necessary to avoid collisions in the data race. F in the chat :(

DSpeichert (approved):

PR #3326, opened 2021-02-16
Dispatcher optimizations

yamaken93:
### Pull Request Prelude

### Changes Proposed

Use perfect fowarding(&& together with std::forward<>) for all tasks in TFS. As far i know, we are telling the compiler to forward the temp creation of the std::function till it reach Task::func.
Also now we are using std::vector in dispatcher. Its a major core change:
It will reduce allocations ALOT. Since this is the core of TFS, preventing it from allocating a slot(pointer, 4 bytes in 32 bits and 8 bytes in 64 bits?) like it was done for std::list every time we add a task is very good.
It *can* reduce lock contention since now we copy the whole vector of pointers to a tmp vec that will execute a list of tasks without the lock being locked.
push_front parameter in Dispatcher::addTask has been removed. It can't be supported by a single vector.
Also it may have a subtle bug. Its add a timed(scheduler) task at the front(its the only place that set this parameter to true) and make the task to be executed before all the other tasks(that kinda makes sense since you want the timed task to be executed asap) that already have been added to be dispatched, even other timed tasks that are older(this is the subtle bug). 
As far i can say there was no harm removing it.


src/tasks.h
```diff
@@ -81,7 +81,7 @@ class Dispatcher : public ThreadHolder<Dispatcher> {
 		std::mutex taskLock;
 		std::condition_variable taskSignal;
 
-		std::list<Task*> taskList;
+		std::vector<Task*> taskList;
```
kygov, line 84: Maybe we could use taskList.reserve in Dispatcher constructor?

yamaken93: We could, i did not bothered about reserving in both taskList and tmpTaskList since(atleast in visual studio) the vector keeps the capacity after copying another vector and call to clear(so its means if we add 1500 tasks to taskList in one time like in server startup it will keep a 1800~ capacity for taskList and 1500 for tmpTaskList).

ranisalt:
[An objectively better solution is to use lambdas instead of `bind`](https://stackoverflow.com/questions/17363003/why-use-stdbind-over-lambdas-in-c14). You should still forward the parameters, though.

yamaken93:
> [An objectively better solution is to use lambdas instead of `bind`](https://stackoverflow.com/questions/17363003/why-use-stdbind-over-lambdas-in-c14). You should still forward the parameters, though.

Replacing std::bind should be done in another pr(since i won't add more scope to this pr than it already has). the perfect forwarding here was just a small improvement together with the major change which is dispatcher using vector.

src/tasks.cpp
```diff
@@ -48,38 +49,30 @@ void Dispatcher::threadMain()
...
 		}
 
-		if (!taskList.empty()) {
-			// take the first task
-			Task* task = taskList.front();
-			taskList.pop_front();
-			taskLockUnique.unlock();
+		tmpTaskList = taskList;
+		taskList.clear();
+		taskLockUnique.unlock();
```
ranisalt, line 54: `std::vector<Task*> tmpTaskList = std::move(taskList);`

yamaken93: taskList will be left in a unknown state(?) That also kills the optimization which both taskList and tmpTaskList keeping the reserved buffer avoiding allocations when adding tasks. Moving would prevent a single copy between vectors, but at what price?

ranisalt: Hmm, that's fair, but since you are locking, what is the point of making a copy?

yamaken93: To reducer lock contention. Since we got a list of tasks, it will spend n time to execute those tasks. If they are executed inside the lock, the lock will be locked for that n time delaying any other any other thread that wants the lock and even worst as executing tasks will call Dispatcher:addTask there will be a deadlock.

MillhioreBT:
This is interesting because the simultaneously accessed is not locked until the tasks are finished, I still do not have something very good to say about this, I will have to do performance tests with millions of simultaneous tasks and compare with the current code

src/tasks.cpp
ranisalt: It makes sense, but just declare the copy array here instead of at the top of the function. I would do something along the lines of

```
std::vector<Task*> tasksToCheck;
std::swap(taskList, tasksToCheck);
taskLockUnique.unlock();
```

Solves the unknown state and still saves the copy :P if you want to keep the reserved capacity, you can use `reserve`, but I honestly wouldn't bother.

marcinho1994 (approved):

src/tasks.cpp
yamaken93: That probably will fix, but i don't agree. I had two aims with this PR: reduce allocation calls and lock contention.

danyelsalym (approved):

yamaken93:
> This is interesting because the simultaneously accessed is not locked until the tasks are finished, I still do not have something very good to say about this, I will have to do performance tests with millions of simultaneous tasks and compare with the current code

I believe the lock contention part can be tested if you add one or few tasks to dispatcher that lags/expends few milliseconds and right after that you try to add some thousand/hundred tasks. The thread which adds the tasks must be not dispatcher so we have true concurrency.

MillhioreBT:
> > This is interesting because the simultaneously accessed is not locked until the tasks are finished, I still do not have something very good to say about this, I will have to do performance tests with millions of simultaneous tasks and compare with the current code
> 
> I believe the lock contention part can be tested if you add one or few tasks to dispatcher that lags/expends few milliseconds and right after that you try to add some thousand/hundred tasks. The thread which adds the tasks must be not dispatcher so we have true concurrency.

I haven't noticed any difference in terms of performance, but clearly the code looks cleaner and careful
I am not confirming that this does not bring extra performance, only that on my computer there seems to be no difference

I will keep trying this code for several more days, and if I find a problem I will not hesitate for a second to come out crying to complain 😭

yamaken93:
I changed the code as @ranisalt asked + a small change and now its all good for me.

ranisalt (approved):

PR #3329, opened 2021-02-17
* New EventCallback style like Revscript *

MillhioreBT:
### Pull Request Prelude

### Changes Proposed

I propose a change to the callback system, I have already heard that the system was not created with the intention of having multiple instances, but for me this should have the ability to handle multiple instances correctly and in order.

If you wonder why I introduced a new method to the `table` - `table.pack`, it is simply because in lua 5.1 the parameter concatenator does not obtain parameters beyond the first nil it finds, and therefore it is necessary to make use of this function so that these changes to work properly.

Now the whole system is encapsulated in its own table `EventCallback`
* added method `register` for eventcallback instances
*this means that now you need to register the event, this makes it familiar with Revscript, because they live in the same environment*
*You can pass a parameter to the `register` method, this parameter must be a number, this number will define the order of execution, the following example can help you visualize this:*
<details>
<summary> Code Example 1 </summary>

```lua
local ec = EventCallback
function ec.onGainExperience(player, source, exp, rawExp)
	print(1)
	exp = exp * 2
	return exp
end
ec:register(2)

local ec = EventCallback
function ec.onGainExperience(player, source, exp, rawExp)
	print(2)
	exp = exp + 100
	return exp
end
ec:register(1)
```

</details>

*Even though it was declared first, and recorded first, its execution will take place after the second event, since its execution order has a higher index than the second event which has a index of 1*
*It is worth mentioning that any number is valid, this means that `-math.huge` will always be executed first, and if two events have this same index, lua will choose which one to execute first and then continue with the corresponding one, by default the index of the events is `math.huge` this means that it will always be at the end of the table*

* Each type of event has its own order of execution, which is obvious

These changes increase the flexibility of the *Revscript*, since you can really create complete systems in a single file
if you have suggestions, you are welcome to say something about it.

This was an idea that I had some time ago, but with the help of @EvilHero90 I got to complete the idea completely and with their help I created this.

Here another example with negative index: 🗡️
<details>
<summary> Code Example 2 </summary>

```lua
local ec = EventCallback
function ec.onGainExperience(self, source, exp, rawExp)
	print(1)
	exp = exp + 200
	return exp
end
ec:register(-5)

local ec = EventCallback
function ec.onGainExperience(self, source, exp, rawExp)
	print(2)
	exp = exp * 3
	return exp
end
ec:register(-10)
```

</details>

### DIFF
--- a/data/events/scripts/creature.lua
+++ b/data/events/scripts/creature.lua
@@ -1,34 +1,34 @@
 function Creature:onChangeOutfit(outfit)
-	if hasEventCallback(EVENT_CALLBACK_ONCHANGEMOUNT) then
-		if not EventCallback(EVENT_CALLBACK_ONCHANGEMOUNT, self, outfit.lookMount) then
+	if EventCallback:has(EVENT_CALLBACK_ONCHANGEMOUNT) then
+		if not EventCallback:get(EVENT_CALLBACK_ONCHANGEMOUNT, self, outfit.lookMount) then
 			return false
 		end
 	end
-	if hasEventCallback(EVENT_CALLBACK_ONCHANGEOUTFIT) then
-		return EventCallback(EVENT_CALLBACK_ONCHANGEOUTFIT, self, outfit)
+	if EventCallback:has(EVENT_CALLBACK_ONCHANGEOUTFIT) then
+		return EventCallback:get(EVENT_CALLBACK_ONCHANGEOUTFIT, self, outfit)
 	else
 		return true
 	end
 end
 
 function Creature:onAreaCombat(tile, isAggressive)
-	if hasEventCallback(EVENT_CALLBACK_ONAREACOMBAT) then
-		return EventCallback(EVENT_CALLBACK_ONAREACOMBAT, self, tile, isAggressive)
+	if EventCallback:has(EVENT_CALLBACK_ONAREACOMBAT) then
+		return EventCallback:get(EVENT_CALLBACK_ONAREACOMBAT, self, tile, isAggressive)
 	else
 		return RETURNVALUE_NOERROR
 	end
 end
 
 function Creature:onTargetCombat(target)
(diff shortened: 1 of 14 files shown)


EvilHero90:
I'm heavily in favour of this, I'll do a full review once I'm fully done testing

MillhioreBT:
* Compatibility with the previous version is now maintained, however the event must still be registered.
* The possibility of managing the order of execution of each eventcallback is maintained, the default index is `-math.huge` so that all the events you register will be competing to be executed first before the others, you can configure the order of execution by passing a parameter to the `register` method, you can pass it a numeric value or that is convertible into a number, here is an example:
```lua
ec:register(1)
ec:register("4")
ec:register(-5)
ec:register(-1.5)
ec:register(233)
ec:register(math.huge)
ec:register(-math.huge) -- default index is same -> ec:register()
```
the event with the lowest index will be executed first, and continues to the highest index!

* Added debug messages to inform the console if someone is configured or registering a callback incorrectly

EvilHero90 (approved):
Been testing this a while now, works like intended, now we can set an order of execution if needed and it passes the parameters onwards correctly.

3Magic (commented):
The line 4079 in luascript.cpp had too much indentation on the "if" statement.
It wouldn't compile in ubuntu, windows compile worked fine.

Should be:

int LuaScriptInterface::luaTablePack(lua_State* L)
{
	// table.pack(...)
	int i;
	int n = lua_gettop(L);  /* number of elements to pack */
	lua_createtable(L, n, 1);  /* create result table */
	lua_insert(L, 1);  /* put it at index 1 */
	for (i = n; i >= 1; i--)  /* assign elements */
		lua_rawseti(L, 1, i);
	if (luaL_callmeta(L, -1, "__index") != 0) {
			lua_replace(L, -2);
		}
	lua_pushinteger(L, n);
	lua_setfield(L, 1, "n");  /* t.n = number of elements */
	return 1;  /* return table */
}

PR #3330, opened 2021-02-17
Fix for bad value for ReqMana for 0 magic level

Herwi:
### Changes Proposed

** That update fixes disability to train magic level. In player.cpp there is if which check if actual required mana isnt higher or equal to next level required mana. If You invoke Vocation::getReqMana it returns extremly high value because it tried to invoke std::pow(manaMultiplier, static_cast<int32_t>(-1)).

### DIFF
--- a/src/vocation.cpp
+++ b/src/vocation.cpp
@@ -187,6 +187,8 @@ uint64_t Vocation::getReqSkillTries(uint8_t skill, uint16_t level)
 
 uint64_t Vocation::getReqMana(uint32_t magLevel)
 {
-	if (magLevel == 0) return 0;
+	if (magLevel == 0) {
+		return 0;
+	}
 	return 1600 * std::pow(manaMultiplier, static_cast<int32_t>(magLevel - 1));
 }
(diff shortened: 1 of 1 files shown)


nekiro:
I can't really understand what you fixed, but follow tfs coding style, we don't use one liners and we use curly brackets.
If you pass 0 to getReqMana it will do std::pow(manaMultiplier , -1) which should give very small value not big, but anyway I agree, it should return 0 if passed maglevel is 0.

```
if (maglevel == 0) {
    return 0;
}
```

yamaken93:
> I can't really understand what you fixed, but follow tfs coding style, we don't use one liners and we use curly brackets.
> If you pass 0 to getReqMana it will do std::pow(manaMultiplier , -1) which should give very small value not big, but anyway I agree, it should return 0 if passed maglevel is 0.
> 
> ```
> if (maglevel == 0) {
>     return 0;
> }
> ```

The formula  was only made to be used with maglevel  >= 1. So yeah the fix is needed.

EPuncker (approved):

nekiro (approved):

ranisalt:
IMHO it would be cleaner to use `std::max<int32_t>(magLevel - 1, 1)`

yamaken93:
@ranisalt the function should return 0 since the needed mana spent for mag level 0 is 0.

Issue #3335, opened 2021-02-20
VIP List review


EPuncker:
So I was taking a look at our source code and stumbled on the VIP List code, so I would like to propose a few things:

- [x] lower the default max value to 100 ([source](https://tibia.fandom.com/wiki/VIP)) [I don't think it has ever been 200 on rl)
- [x] make max value configurable in config.lua with pre-defined values for premium (100) and free (20) account

EPuncker:
closed by @idontreallywolf thank you

PR #3337, opened 2021-02-20
Prevent players from adding themselves to the viplist

idontreallywolf:
### Pull Request Prelude

### Changes Proposed
Added a check in `Game.playerRequestAddVip` to fix point 3 in #3335 

**Issues addressed:** #3335 

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3229,6 +3229,11 @@ void Game::playerRequestAddVip(uint32_t playerId, const std::string& name)
 		return;
 	}
 
+    if (player->getName() == name) {
+        player->sendTextMessage(MESSAGE_STATUS_SMALL, "You can not add yourself.");
+        return;
+    }
+
 	Player* vipPlayer = getPlayerByName(name);
 	if (!vipPlayer) {
 		uint32_t guid;
@@ -5757,4 +5762,3 @@ bool Game::reload(ReloadTypes_t reloadType)
 	}
 	return true;
 }
-
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3229,11 +3229,6 @@ void Game::playerRequestAddVip(uint32_t playerId, const std::string& name)
 		return;
 	}
 
-	if (player->getName() == name) {
-		player->sendTextMessage(MESSAGE_STATUS_SMALL, "You can not add yourself.");
-		return;
-	}
-
 	Player* vipPlayer = getPlayerByName(name);
 	if (!vipPlayer) {
 		uint32_t guid;
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -5757,3 +5757,4 @@ bool Game::reload(ReloadTypes_t reloadType)
 	}
 	return true;
 }
+
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3229,10 +3229,10 @@ void Game::playerRequestAddVip(uint32_t playerId, const std::string& name)
 		return;
 	}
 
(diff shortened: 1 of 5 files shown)


MillhioreBT:
I think it is better here:
https://github.com/otland/forgottenserver/blob/ad6865b79e585a2404cc58ca844a751fd267e70f/src/player.cpp#L2189-L2194

image

src/player.cpp
```diff
@@ -2188,6 +2188,11 @@ bool Player::removeVIP(uint32_t vipGuid)
 
 bool Player::addVIP(uint32_t vipGuid, const std::string& vipName, VipStatus_t status)
 {
+	if (getGUID() == vipGuid) {
```
MillhioreBT, line 2191: I think you can just use `guid` since this variable is available from `this`
sorry for bothering you so much 🤗

Someone else can come and say that something else is better so don't scold me later ;)

MillhioreBT (changes requested):

src/player.cpp
idontreallywolf: > I think you can just use `guid` since this variable is available from `this`

That makes sense xD. I'll make the changes and commit in a moment.

> sorry for bothering you so much

Not bothering at all, I appreciate the feedback from experienced developers.

src/game.cpp
```diff
@@ -5757,4 +5757,3 @@ bool Game::reload(ReloadTypes_t reloadType)
 	}
 	return true;
 }
-
```
ranisalt, line 5760: Undo this modification

ranisalt (changes requested):

MillhioreBT (dismissed):
I already had this in my sources and I assumed that TFS had it too, I don't know if this was removed at some point or just never implemented.

idontreallywolf:
> I already had this in my sources and I assumed that TFS had it too, I don't know if this was removed at some point or just never implemented.

🤔 

https://github.com/otland/forgottenserver/pull/2611/files
https://github.com/otland/forgottenserver/pull/2612/files

DSpeichert:
1. Why is this an important feature? What makes it that bad in adding yourself to VIP list that you have to tell the player "you can not..."?
2. If so, can you make it configurable (config.lua)?

idontreallywolf:
>     1. Why is this an important feature? What makes it that bad in adding yourself to VIP list that you have to tell the player "you can not..."?
> 
>     2. If so, can you make it configurable (config.lua)?

1. It probably isn't an *important* feature but what's the point of adding yourself to viplist xD?
2. done - it can now be configured from config.lua whether players should be able to add themselves or not.

EPuncker:
Okay I'm really really sorry for starting all this, but just tested on rl tibia and it still allow us to add ourselves to our own vip list (wonder why), I've removed this part from my issue, but thank you for fixing the other one

you may close this one if you want to

MillhioreBT:
> Okay I'm really really sorry for starting all this, but just tested on rl tibia and it still allow us to add ourselves to our own vip list (wonder why), I've removed this part from my issue, but thank you for fixing the other one
> 
> you may close this one if you want to

Thanks for the info! that's why i always say i trust the guys at tfs ;)

DSpeichert (commented):
The code looks good and I would approve it if the cipbia behavior has changed. However, as @EPuncker just commented, it has not changed. Therefore, I'm thinking that this is probably better left untouched. However useless adding yourself to the VIP list is, there's no harm in that and I can't see a reason why someone would want to block that.

PR #3338, opened 2021-02-20
make viplist size configurable

idontreallywolf:
### Pull Request Prelude

### Changes Proposed
Modified code to read viplist limit from config.lua based on premium status

**Issues addressed:** #3335 


MillhioreBT:
`freemium` It doesn't seem like a correct name, why not just use `free`

This looks like code in the style of TFS 0.4:
https://github.com/otland/forgottenserver/blob/fd2ce37bee352eec063bc78c0c94a1e1c871b542/src/player.cpp#L4522-L4524

idontreallywolf:
> freemium It doesn't seem like a correct name

would you say `vipFreeLimit` is ok to use?

> This looks like code in the style of TFS 0.4:

My mistake, will change that now

DSpeichert (approved):

PR #3341, opened 2021-02-20
Move signals internal functions to cpp

ranisalt:
### Pull Request Prelude

### Changes Proposed

Another one of these, reduces the exposed amount of functions for faster compilation and better compile optimization opportunities. Saves 72 bytes on the generated executable for me, too.

**Issues addressed:** 

### DIFF
--- a/src/signals.cpp
+++ b/src/signals.cpp
@@ -57,25 +57,6 @@ extern Events* g_events;
 extern Chat* g_chat;
 extern LuaEnvironment g_luaEnvironment;
 
-Signals::Signals(boost::asio::io_service& service) :
-	set(service)
-{
-	set.add(SIGINT);
-	set.add(SIGTERM);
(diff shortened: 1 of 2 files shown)


src/signals.cpp
```diff
@@ -57,8 +57,6 @@ extern Events* g_events;
 extern Chat* g_chat;
 extern LuaEnvironment g_luaEnvironment;
 
-using ErrorCode = boost::system::error_code;
-
 Signals::Signals(boost::asio::io_service& service) :
```
MillhioreBT, line 60: move this function down the `namespace`

MillhioreBT:
It is worth mentioning that in my case saves 25 bytes, this will depend on the compile, but in general you can see that if it works

MillhioreBT (approved):

DSpeichert (approved):

PR #3342, opened 2021-02-21
Change experience formula

omarcopires:
### Pull Request Prelude

### Changes Proposed

-Ciptibia formula have less multiplication which should make it a little faster.
-Because it have less multiplication it also don't overflow as often, with Ciptibia formula it is possible to reach 1034406 level instead of 717217 with Open-Tibia formula.

Co-Authored-By: Jakub[Madao] <38609240+SaiyansKing@users.noreply.github.com>

### DIFF
--- a/data/global.lua
+++ b/data/global.lua
@@ -167,6 +167,5 @@ function getPlayerDatabaseInfo(name_or_guid)
 end
 
 function getExperienceForLevel(level)
-	local level = level - 1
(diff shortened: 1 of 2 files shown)


ranisalt:
Interesting. This function should not be repeated, though, so `getExpForLevel` in Lua should be bound to C++.

ranisalt (approved):
Values are correct for the `uint64_t` range.

nekiro:
GetExpForLevel could be made static right?

PR #3344, opened 2021-02-23
add methods for revscriptsys talkactions

idontreallywolf:
### Pull Request Prelude

### Changes Proposed

This PR is an attempt to allow setting command permission (access & account type) from revscriptsys in order to avoid the following lines in most script files:
```lua
	return true
end

	return false
end
```

The above code can be done in the following way:
```lua
-- talkaction:permission(needAccess = false, AccountType_t = ACCOUNT_TYPE_NORMAL)
talkaction:permission(true, ACCOUNT_TYPE_GOD)
```

Example:
```lua
local talk = TalkAction("/abc")

	player:sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, 'Henlo! get bonked')
	return false
end

talk:permission(true, ACCOUNT_TYPE_GOD)
talk:separator(" ")
talk:register()
```

**Issues addressed:** none

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2941,6 +2941,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("TalkAction", "onSay", LuaScriptInterface::luaTalkactionOnSay);
 	registerMethod("TalkAction", "register", LuaScriptInterface::luaTalkactionRegister);
 	registerMethod("TalkAction", "separator", LuaScriptInterface::luaTalkactionSeparator);
+	registerMethod("TalkAction", "permission", LuaScriptInterface::luaTalkactionPermission);
 
 	// CreatureEvent
 	registerClass("CreatureEvent", "", LuaScriptInterface::luaCreateCreatureEvent);
@@ -15342,6 +15343,28 @@ int LuaScriptInterface::luaTalkactionSeparator(lua_State* L)
 	TalkAction* talk = getUserdata<TalkAction>(L, 1);
 	if (talk) {
 		talk->setSeparator(getString(L, 2).c_str());
+		pushBoolean(L, true);
+	} else {
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+int LuaScriptInterface::luaTalkactionPermission(lua_State* L)
+{
+	// talkAction:permission(needAccess = false[, AccountType_t = ACCOUNT_TYPE_NORMAL])
+	TalkAction* talk = getUserdata<TalkAction>(L, 1);
+	if (talk) {
+		if (lua_gettop(L) == 1) {
+			pushBoolean(L, true);
+			return 1;
+		}
+
+		if (lua_gettop(L) == 3) {
+			talk->setRequiredAccountType(getNumber<AccountType_t>(L, 3));
+		}
+
+		talk->setNeedAccess(getBoolean(L, 2));
 		pushBoolean(L, true);
 	} else {
 		lua_pushnil(L);
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2941,7 +2941,8 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("TalkAction", "onSay", LuaScriptInterface::luaTalkactionOnSay);
 	registerMethod("TalkAction", "register", LuaScriptInterface::luaTalkactionRegister);
 	registerMethod("TalkAction", "separator", LuaScriptInterface::luaTalkactionSeparator);
-	registerMethod("TalkAction", "permission", LuaScriptInterface::luaTalkactionPermission);
+	registerMethod("TalkAction", "access", LuaScriptInterface::luaTalkactionAccess);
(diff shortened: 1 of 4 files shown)


EvilHero90:
I like the idea in general but I'd rather have this in 2 seperate methods
`access(bool)` and `accountType(type)` having it as one method only, feels kinda clunky

idontreallywolf:
> I like the idea in general but I'd rather have this in 2 seperate methods
> `access(bool)` and `accountType(type)` having it as one method only, feels kinda clunky

Initially I did that but then changed my mind 😁 
Will make the changes in a moment.

src/luascript.cpp
```diff
@@ -15349,6 +15351,32 @@ int LuaScriptInterface::luaTalkactionSeparator(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaTalkactionAccess(lua_State* L)
+{
+	// talkAction:access(needAccess = false)
```
MillhioreBT, line 15356: to be really part of *Revscripts*, you should be able to get and set a value with the same function, also the parameters should have a cleaner name

example:
```c++
int LuaScriptInterface::luaTalkactionAccess(lua_State* L)
{
	// get: talkAction:access() or set: talkAction:access(bool)
	TalkAction* talk = getUserdata<TalkAction>(L, 1);
	if (talk) {
		if (lua_gettop(L) == 1) {
			pushBoolean(L, talk->getAccess());
		} else {
			talk->setAccess(getBoolean(L, 2));
			pushBoolean(L, true);
		}
	}
	else {
		lua_pushnil(L);
	}
	return 1;
}
```

MillhioreBT: I think evil hero was very clear with it 
image

idontreallywolf: Evil Hero mentioend that there shouldn't be `get` methods because revscriptsys wasn't designed for that purpose but only to register scripts. 🤔

src/talkaction.h
```diff
@@ -59,14 +59,33 @@ class TalkAction : public Event
...
+			needAccess = b;
+		}
 
 	private:
 		std::string getScriptEventName() const override;
 
 		std::string words;
 		std::vector<std::string> wordsMap;
 		std::string separator = "\"";
+
```
EvilHero90, line 85: remove the space

src/talkaction.h
```diff
@@ -59,14 +59,33 @@ class TalkAction : public Event
...
+		bool needAccess = false;
+
```
EvilHero90, line 87: remove this one aswell

src/talkaction.cpp
```diff
@@ -130,6 +130,16 @@ TalkActionResult_t TalkActions::playerSaySpell(Player* player, SpeakClasses type
 			}
 		}
 
+		if (it->second.fromLua) {
+			if (it->second.getNeedAccess() && !player->getGroup()->access) {
+				return TALKACTION_CONTINUE;
```
EvilHero90, line 135: shouldn't this be `return TALKACTION_BREAK;` as the player does not meet the required group access?

EvilHero90 (changes requested):
We can merge this once you have addressed the remaining issues.

src/talkaction.cpp
idontreallywolf: I was thinking that as well but after looking at the other scripts I thought I should keep the same behaviour; 
`player has no access -> return true`
`player has wrong accType -> return false`

src/luascript.cpp
MillhioreBT: You're right, this is only a feature for custom servers.
the small example of use would be the following:
```lua
local talk = TalkAction("/comandos", "!comandos", "/commands", "!commands")

	local description = ""
	for words, talk in pairs(Game.getTalkActions()) do
		if not talk:access() or player:getGroup():getAccess() then
			if description == "" then
				description = description .. '\n'
			end
			description = description .. words
		end
	end
	player:showTextDialog(26654, description)
	return false
end

talk:register()
```

src/talkaction.cpp
EvilHero90: Yea ok, I can see now what you mean, rest of code looks fine to me now

EvilHero90 (approved):
Looks good to me, well done!

EPuncker:
we need to update https://github.com/otland/forgottenserver/wiki/Revscriptsys#talkaction ASAP 😛

MillhioreBT:
btw why not implement the configuration for the xml version?

idontreallywolf:
> btw why not implement the configuration for the xml version?

Do you mean `access=""` and `accountType=""` ?

I thought TFS was moving away from xml 🤔

MillhioreBT:
> > btw why not implement the configuration for the xml version?
> 
> Do you mean `access=""` and `accountType=""` ?
> 
> I thought TFS was moving away from xml 🤔

Yes, in fact, it does it little by little, although it takes a long time to be completely independent of xml
Anyway, it was just a sudden idea I had, don't listen to me, if necessary people would ask for it, so that's fine

PR #3349, opened 2021-02-26
Moved place_monster talkAction to revscript

EPuncker:
### Pull Request Prelude

### Changes Proposed
Moved it to revscriptsys so users can have an example of the newly implement **access** and **accountType** methods added on https://github.com/otland/forgottenserver/commit/435cc83c9e2c905daaaba6d87da308181a2dbcb4

### DIFF
--- a/data/scripts/talkactions/place_monster.lua
+++ b/data/scripts/talkactions/place_monster.lua
@@ -1,12 +1,6 @@
-function onSay(player, words, param)
-	if not player:getGroup():getAccess() then
-		return true
-	end
+local talk = TalkAction("/m")
 
-	if player:getAccountType() < ACCOUNT_TYPE_GOD then
-		return false
-	end
-
+function talk.onSay(player, words, param)
 	local position = player:getPosition()
 	local monster = Game.createMonster(param, position)
 	if monster then
@@ -18,3 +12,8 @@ function onSay(player, words, param)
 	end
 	return false
 end
+
+talk:access(true)
+talk:accountType(ACCOUNT_TYPE_GOD)
+talk:separator(" ")
+talk:register()
--- a/data/talkactions/talkactions.xml
+++ b/data/talkactions/talkactions.xml
@@ -18,7 +18,6 @@
 	<talkaction words="/openserver" script="openserver.lua" />
 	<talkaction words="/closeserver" separator=" " script="closeserver.lua" />
 	<talkaction words="/B" separator=" " script="broadcast.lua" />
-	<talkaction words="/m" separator=" " script="place_monster.lua" />
 	<talkaction words="/i" separator=" " script="create_item.lua" />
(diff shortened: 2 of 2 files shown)


EvilHero90 (approved):
I don't mind moving one talkaction as showcase, however I'm not sure if others will be happy with it, as others wanted to leave old code untouched in xml

strndi:
I really don't understand the point behind splitting up actions, TFS looks like such a mess now when we have seperate folders for actions, movements, talkactions and everything else.

Revscripts is a really powerful system, but I don't really get the point of converting single actions into revscripts, it's better suited for systems that require several actions.

EPuncker:
> I really don't understand the point behind splitting up actions, TFS looks like such a mess now when we have two seperate folders for actions, movements, talkactions and everything else.
> 
> Revscripts is a really powerful system, but I don't really get the point of converting single actions into revscripts, it's better suited for systems that require several actions.

as I stated, "so users can have an example of usage of such feature", or what is the point of the feature added on https://github.com/otland/forgottenserver/commit/435cc83c9e2c905daaaba6d87da308181a2dbcb4 by @idontreallywolf if users wouldn't have any example of its usage? I doubt any non advanced user would go to luascript.cpp file to search for features they don't know that exist...

MillhioreBT:
> I really don't understand the point behind splitting up actions, TFS looks like such a mess now when we have two seperate folders for actions, movements, talkactions and everything else.
> 
> Revscripts is a really powerful system, but I don't really get the point of converting single actions into revscripts, it's better suited for systems that require several actions.

I agree, currently there is a little fight between revscript or xml, but that fight will end when it is implemented, items, npc, chat, and maybe some other additional module to revscript, and of course when Remeres have support for this system, but for now the engine is in an innovation stage and is a bit far from being a stable and complete release version, If you are looking for something stable and that any change in the sources does not directly affect the way you work on your server you can use some of the stable versions, such as TFS 1.0, 1.1 or 1.2

MillhioreBT:
> > I really don't understand the point behind splitting up actions, TFS looks like such a mess now when we have two seperate folders for actions, movements, talkactions and everything else.
> > Revscripts is a really powerful system, but I don't really get the point of converting single actions into revscripts, it's better suited for systems that require several actions.
> 
> as I stated, "so users can have an example of usage of such feature", or what is the point of the feature added on [435cc83](https://github.com/otland/forgottenserver/commit/435cc83c9e2c905daaaba6d87da308181a2dbcb4) by @idontreallywolf if users wouldn't have any example of its usage? I doubt any non advanced user would go to luascript.cpp file to search for features they didn't knew that exist...

your example idea is good, but it could also be implemented directly to the example file that comes by default in revscript for talkactions

[data/scripts/talkactions/position.lua](https://github.com/otland/forgottenserver/blob/master/data/scripts/talkactions/position.lua)

PR #3350, opened 2021-02-26
Use C++17

ranisalt:
### Pull Request Prelude

### Changes Proposed

By sticking to an older standard, we miss the new features of the language and the standard library, which can make our code more expressive. Notable new features are:
- [binary literals and digit separators](https://en.cppreference.com/w/cpp/language/integer_literal) for better writing large constants
- [generic lambdas](https://en.cppreference.com/w/cpp/language/lambda) - can use `auto` in parameter types
- relaxed `constexpr` restrictions, we can make more of the code precalculated on compile time without template shenanigans
- [return type deduction for functions](https://en.cppreference.com/w/cpp/language/function#Return_type_deduction_.28since_C.2B.2B14.29)
- simplified nested namespace syntax
- [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding), analogous to destructuring or multiple assignment in other languages
- [`nodiscard`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) for improved function usage safety
- [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp), [`std::gcd`](https://en.cppreference.com/w/cpp/numeric/gcd), [`std::lcm`](https://en.cppreference.com/w/cpp/numeric/lcm), and [`std::reduce`](https://en.cppreference.com/w/cpp/algorithm/reduce) utilities
- [`std::filesystem`](https://en.cppreference.com/w/cpp/filesystem)
- [`std::make_unique`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) - this one should actually be on C++11 but was forgotten)
- [`insert_or_assign`](https://en.cppreference.com/w/cpp/container/map/insert_or_assign) to upsert to map without implicit allocation
- [`merge`](https://en.cppreference.com/w/cpp/container/map/merge) for maps and sets
- [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional)
- array support for `std::shared_ptr`
- [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view)
- [`std::variant`](https://en.cppreference.com/w/cpp/utility/variant)
And many more

## Why not C++14 only?
Comparing to C++17, C++14 feels just like a small patch to the language. Most of the changes above are from C++17 and C++14 only does not provide much.

## Why not C++20?
This was mentioned by @djarek in a group conversation, however, I personally don't think it's a good idea to use it now while compiler support is still lacking and we support old versions of Ubuntu. C++17 is fully implemented in the compilers that ship with Ubuntu 18.04 (16.04 will EOL in 2 months)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ add_executable(tfs ${tfs_SRC})
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
-set_target_properties(tfs PROPERTIES CXX_STANDARD 11)
+set_target_properties(tfs PROPERTIES CXX_STANDARD 17)
 set_target_properties(tfs PROPERTIES CXX_STANDARD_REQUIRED ON)
 
 if (${CMAKE_VERSION} VERSION_GREATER "3.16.0")


nekiro (approved):

MillhioreBT (approved):

EPuncker (approved):

yamaken93:
Signed. C++ 14 should be ignored in favor of C++ 17 since C++ 20 is not ready yet.

DSpeichert (approved):
I see no issues with compatibility now.

PR #3352, opened 2021-02-27
add !commands for revscriptsys

idontreallywolf:
### Pull Request Prelude

### Changes Proposed
- Added `Game.getTalkactions()` method to retrieve revscriptsys talkactions
- Added a script in globalevent-scripts to cache talkactions on startup
- Added `!commands` to read cached commands.

Result while on God char:
god

Result while on player char:
player

**Issues addressed:** none


MillhioreBT:
I think that physical examples are not necessary, also the code may have much fewer lines, with the methods get/set are used to obtain access and accountType from the TalkAction object itself, as you propose right now it will not be accepted

idontreallywolf:
> I think that physical examples are not necessary, also the code may have much fewer lines, with the methods get/set are used to obtain access and accountType from the TalkAction object itself, as you propose right now it will not be accepted
> 
> The idea of ​​using cache is bad when it comes to something so small, making several files for cache is unnecessary, you can have a local table with the cache in any case, but I do not recommend using cache either, it does not bring any benefit in this case
> 
> If at some point you add this to tfs I would like it to be the way I have it on my server or maybe in a better way if possible
> Real Example Code

I see what you mean, it makes sense that caching would be best for large amounts of data where talkactions are not that heavy. I sometimes overthink on problems 😨

EPuncker (approved):

DSpeichert:
@idontreallywolf resolve conflict pls

src/talkaction.h
```diff
@@ -101,6 +101,10 @@ class TalkActions final : public BaseEvents
 		bool registerLuaEvent(TalkAction* event);
 		void clear(bool fromLua) override final;
 
+		std::map<std::string, TalkAction> getTalkactions() {
+			return talkActions;
+		}
+
```
ranisalt, line 107: I find it very confusing that a class named `TalkActions` has a method called `getTalkactions()`. How do you *get* talkactions out of... talkactions?

I would rather add `begin`/`end` to navigate through the internal map instead. At the very least, make this `const&`.

ranisalt (changes requested):

Issue #3010, opened 2020-05-03
create itemType through revscriptsys


Zbizu:
### Explanation of what you want to do that is currently impossible
create new itemType through lua and set flags/attributes to it (preferable onStartup)

### Desired functionality
Increase itemId variable capacity (uint16_t -> 32) in sources.
Create new items on the fly through lua scripts
Set clientId
Set/get all data that would be loaded from xml/otb otherwise

### Available workarounds
xml/otb edit, though I don't even know where to find a working otb editor.

### Prior art
TFS 0.4 could do it partially through mods.

### Problems to deal with
Loading the values the way they're seen as normal items in sources.

Dealing with items that don't get initialized anymore (set client id to avoid debug).

Dealing with container contents of items that aren't containers yet or anymore.

Deciding which house items should be loaded/saved. Making sure that they won't disappear.

(optional) NPC trade window different items with same clientId bug

EvilHero90:
This has already been planned and wont make it into 1.4 but it'll probably be a part of 1.5 along with NPC's and a few other things. I'll add this to the 2.0 roadmap tho, so we don't forget about it.

Zbizu:
No problem, it can even wait to 2.0.
A solid solution is better than rushed one.

Zbizu:
Possible problems:

- change of attributes - would have to update for all online users - including requirements to equip and weight changes. This would also have to update in houses.
- loading before registration - currently incorrect items get outright discarded. This could be avoided by giving the new items a reserved id range.
- xml/otb update - the items could overlap with new updates unless their ids were special.
- market - heavily client sided like npc trade.

Possible alternative:
- loading the items from xml file with all xml/otb attributes.

Zbizu:
If anyone is interested, I wrote a converter that could free us from items.xml.
![unknown-4](https://user-images.githubusercontent.com/5205079/126864619-99c093a4-922c-4265-9997-098d7234d13d.png)

What do you think about moving itemType registration entirely to lua?

PR #3354, opened 2021-03-02
createItemType Via Lua

MillhioreBT:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** Inspired by this PR #3010

### Some peculiarities:
* Items can be created through lua if you are in the Rev Scripts environment
* However its charge and recharge are separate
* Can't reload with `/reload scripts` instead use `/reload items` as commonly done with xml

### Note:
* I have no reason to say that this is the final version, improvements in the code are probably needed
* Although I have been testing these changes for almost 2 months, it has not been peer-reviewed yet, so if you notice something weird and want to help me improve this, your suggestions are welcome.
* Note that the `items.lua` file is by convention, but you can create all the luas files you want in the `data/items` folder so that you can sort sets of items by file as you wish.

How it looks:
```lua
Game.createItemType(26378):register({
   name = "gold converter",
   article = "a",
   weight = 1800,
   duration = 604800,
   charges = 500,
   showcharges = true,
   showduration = true
})
```


src/items.h
```diff
@@ -312,6 +322,8 @@ class ItemType
 		std::unique_ptr<Abilities> abilities;
 		std::unique_ptr<ConditionDamage> conditionDamage;
 
+		std::map<std::string, bool> hasEdited;
```
MillhioreBT, line 325: Here I will use a `unordered_map`, I am hoping to find more things to improve the code

data/lib/compat/compat.lua
```diff
@@ -1398,14 +1399,16 @@ function createFunctions(class)
...
-			if not (rawget(class, get) and rawget(class, set)) then
-				table.insert(temp, {set, setFunc, get, getFunc})
+			local is = "is" .. str
+			if not (rawget(class, get) and rawget(class, set) and rawget(class, is)) then
+				table.insert(temp, {set, setFunc, get, getFunc, is, getFunc})
 			end
 		end
 	end
-	for _, func in ipairs(temp) do
+	for _,func in ipairs(temp) do
```
EPuncker, line 1408:
```suggestion
	for _, func in ipairs(temp) do
```

DSpeichert:
If you're constraining this to just `data/items/` then isn't it going against modularity of Revscripts as mentioned somewhere else by @EvilHero90 ?

EvilHero90:
For now we can't really get around having it extra like we did with monsters because of the loading order (might throw errors and such if we load items from lua later than we do at xml), at current state as we approach further into lua handling stuff this is already a good enhancement and if we every going to rely on lua only we can start re arranging loading order and such so we have everything in one place.

Resolve the conflicts and then we should finaly merge it.

src/luascript.cpp
```diff
@@ -11431,764 +11652,1901 @@ int LuaScriptInterface::luaItemTypeCreate(lua_State* L)
...
+				abilities.absorbPercent[combatTypeToIndex(COMBAT_EARTHDAMAGE)] += value;
+				abilities.absorbPercent[combatTypeToIndex(COMBAT_ICEDAMAGE)] += value;
+				abilities.absorbPercent[combatTypeToIndex(COMBAT_HOLYDAMAGE)] += value;
+				abilities.absorbPercent[combatTypeToIndex(COMBAT_DEATHDAMAGE)] += value;
+				pushBoolean(L, true);
+				return 1;
+			} else if (abilitieName == "absorbpercentall") {
+				int16_t value = getNumber<int16_t>(L, 3);
+				for (uint8_t i = COMBAT_NONE; i <= COMBAT_COUNT; i++) {
+					abilities.absorbPercent[i] += value;
```
EvilHero90, line 13461:
```suggestion
					abilities.absorbPercent[indexToCombatType(i)] += value;
```

src/luascript.cpp
```diff
@@ -11431,764 +11652,1901 @@ int LuaScriptInterface::luaItemTypeCreate(lua_State* L)
...
+			// FieldAbsorbPercents
+			auto fieldAbsorbPercentVal = AbilitieFieldAbsorbPercentMap.find(abilitieName);
+			if (fieldAbsorbPercentVal != AbilitieFieldAbsorbPercentMap.end()) {
+				abilities.fieldAbsorbPercent[combatTypeToIndex(fieldAbsorbPercentVal->second)] += getNumber<int16_t>(L, 3);
+				pushBoolean(L, true);
+				return 1;
+			} else if (abilitieName == "fieldabsorbpercentall") {
+				int16_t value = getNumber<int16_t>(L, 3);
+				for (uint8_t i = COMBAT_NONE; i <= COMBAT_COUNT; i++) {
+					abilities.fieldAbsorbPercent[i] += value;
```
EvilHero90, line 13476:
```suggestion
					abilities.fieldAbsorbPercent[indexToCombatType(i)] += value;
```

EvilHero90 (changes requested):
tons of formatting issues in `luascript.cpp` which need to be resolved but other then that it looks fine.

PR #3355, opened 2021-03-02
Use C++17 in MSVC

EPuncker:
### Pull Request Prelude

### Changes Proposed
follow-up of https://github.com/otland/forgottenserver/commit/501ece8476a8644ca71bc791de1eaf236db120d1

sd

### DIFF
--- a/vc14/theforgottenserver.vcxproj
+++ b/vc14/theforgottenserver.vcxproj
@@ -93,6 +93,7 @@
       <EnableParallelCodeGeneration>true</EnableParallelCodeGeneration>
       <EnableEnhancedInstructionSet>AdvancedVectorExtensions</EnableEnhancedInstructionSet>
       <AdditionalIncludeDirectories>$(VcpkgRoot)include\luajit;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <LanguageStandard>stdcpp17</LanguageStandard>
     </ClCompile>
     <Link>
       <TargetMachine>MachineX86</TargetMachine>


MillhioreBT (approved):
I use it this way since the new src changes were implemented, no compatibility warnings/errors since then

yamaken93:
Signed.

ranisalt (approved):

ranisalt:
Seems like a no-brainer to merge. I would merge it straight away, but I will wait until the reviewers you request have responded

DSpeichert (approved):

soul4soul:
This PR only changes the setting for debug|win32. It's missing debug|x64, release|win32, and release|x64

EPuncker:
> This PR only changes the setting for debug|win32. It's missing debug|x64, release|win32, and release|x64. Better yet this change should be applied to settings.prop so it applies to all build configurations by default.
> 
> The vxcproj and *.props files need some TLC. Conflicting settings are used in some cases. From what I remember configuration in vxcproj overrides imported config from *.props files. For example `settings.prop` sets the warning level to 3 but that setting is overriden for release|x64 builds in the vxcproj file which sets the warning level to 4.
> 
> Now that TFS uses vcpkg it would be worth while to experiment with using cmake on windows with visual studio so we can drop the vcxproj files from the repo. Recent versions of VS studio can open folders with cmake files. VS will generate and maintain the visual studio specific files behind the scene.

https://github.com/otland/forgottenserver/commit/ac72c745a0aaedf143b298f14bdf710dd1f2845d

PR #3356, opened 2021-03-04
Correct return access talkaction

danyelvarejao:


lyuz1n (approved):

MillhioreBT:
> _No description provided._

Why should it be correct?
It has always worked this with BREAK

```lua
	return false
end
```

danyelvarejao:
> > _No description provided._
> 
> Why should it be correct?
> It has always worked this with BREAK
> 
> ```lua
> if player:getAccountType() < ACCOUNT_TYPE_GOD then
> 	return false
> end
> ```

This is an unnecessary check, as the source will already check this.
What happens before my modification is:
A player who doesn't have access to perform a talkaction that needs access, the message will not default.

After my modification:
When a player who doesn't have access executes talkaction, it is as if he is sending a message in the default.

EPuncker (approved):

MillhioreBT:
> > > _No description provided._
> > 
> > 
> > Why should it be correct?
> > It has always worked this with BREAK
> > ```lua
> > if player:getAccountType() < ACCOUNT_TYPE_GOD then
> > 	return false
> > end
> > ```
> 
> This is an unnecessary check, as the source will already check this.
> What happens before my modification is:
> A player who doesn't have access to perform a talkaction that needs access, the message will not default.
> 
> After my modification:
> When a player who doesn't have access executes talkaction, it is as if he is sending a message in the default.

Assuming we will return false, that is correct, but if we don't then it is no longer the same
The truth is I do not know if it is the best way to treat it, but if a change is surely required.

nekiro:
Is this to fix "/commands"? If so, that's a good change. Talkactions aren't even executed for normal chars if you prefix them with /

danyelvarejao:
Yes, it is to correct any command that needs access executed by a player.
If a player of account type = ACCOUNT_TYPE_NORMAL executes a command, this should appear as a normal message in the default, currently it is not so, only after this change.

MillhioreBT:
This change only affects those commands that only require verification of account type, if they do it in conjunction with access then the behavior is what you are expecting, as it is done in the commands in xml

what you propose is to change this behavior, which would affect the commands executed by administrators or characters with privileges, since if someone has access but their account type does not match the command is shown as if it were a simple player, I don't like this, at least for the default behavior no, whoever wants this can simply declare it within the same script

In addition, the changes were designed to be used together, access and account type, or only access, the scripts  no requires verification of account type only, it also verifies if the player has access, so I do not understand this change yet

danyelvarejao:
Take a player without access and try to execute a command that requires access, you will see that the message will not go out in the default channel, this is incorrect behavior (in my opinion), so that any player will know the commands of a god, for example.
This change is for the message to leave the default channel normally, in case a player without access tries to execute a command that requires access.

MillhioreBT:
> Take a player without access and try to execute a command that requires access, you will see that the message will not go out in the default channel, this is incorrect behavior (in my opinion), so that any player will know the commands of a god, for example.
> This change is for the message to leave the default channel normally, in case a player without access tries to execute a command that requires access.

As long as you have this configuration, the behavior is correct:
```lua
-- Example with access
local talk = TalkAction("/pos")
talk:access(true)
talk:accountType(ACCOUNT_TYPE_GOD)
talk:separator(" ")
talk:register()
```

The incorrect way would be:
```lua
-- Example without access
local talk = TalkAction("/pos")
talk:accountType(ACCOUNT_TYPE_GOD)
talk:separator(" ")
talk:register()
```

So far there is no command that only requires verification of the account type only.

DSpeichert (approved):

MillhioreBT (approved):
I thought about it a bit and it seems to me that the idea is not bad, I support this change, anyway we can still handle the requirements manually

PR #3358, opened 2021-03-04
Load outfits from lua

idontreallywolf:
### Pull Request Prelude

### Changes Proposed
This PR is an attempt to load outfits from lua instead of xml.

- Added lua function `loadOutfits(table)`
- Added config option to load outfits from lua or xml

## Example
```lua
loadOutfits({
	{sex = PLAYERSEX_MALE, lookType = 128, name = "Citizen", premium = false, unlocked = true, enabled = true},
	{sex = PLAYERSEX_MALE, lookType = 129, name = "Hunter", premium = false, unlocked = true, enabled = true},
	{sex = PLAYERSEX_MALE, lookType = 130, name = "Mage", premium = false, unlocked = true, enabled = true},
	{sex = PLAYERSEX_MALE, lookType = 131, name = "Knight", premium = false, unlocked = true, enabled = true},
	. . .
}
```

**Issues addressed:** none


MillhioreBT:
I don't like how the code is structured, and I don't like the way it's used

I don't like that this code loads as if it were any script, as it is not guaranteed to load before any other script that requires know something about the outfits

later if I have any suggestions I will let you know

idontreallywolf:
> I don't like that this code loads as if it were any script, as it is not guaranteed to load before any other script that requires know something about the outfits

I'm not sure the order of loading oufits have mattered so far. XML outfits are currently loaded after the scripts. Whether it loads with the scripts or after the scripts seems to not make a lot of difference. What kind of script do you assume will require the outfits to be loaded before anything/that specific script?

```cpp
std::cout << ">> Loading lua scripts" << std::endl;
if (!g_scripts->loadScripts("scripts", false, false)) {
    startupErrorMessage("Failed to load lua scripts");
    return;
}

// ... xml monsters loading here
// ... lua monsters loading here

std::cout << ">> Loading outfits" << std::endl;
if (!Outfits::getInstance().loadFromXml()) {
    startupErrorMessage("Unable to load outfits!");
    return;
}
```

MillhioreBT:
I was just thinking about the situation in which a script required to know some information about an outfit and since the outfits have not been loaded yet, then the behavior would not be as expected, anyway, currently it is not necessary to load the outfits from lua, the current code is not annoying, but of course it may be that later if we need a feature like this, but I would like to see some more elegance in the code and that be more versatile

It's just my opinion, someone else might think something different

soul4soul:
Due to the introduction of the revscriptsys I can see a far away future of TFS where xml is no longer used.  In the meantime I'm not so sure about this change. There is limited necessity in making a feature like this configurable.

From what I understand the dev team accepts TFS 2.0 would have some breaking changes. I think it would be appropriate on holding off on a change like this until TFS 2.0 and just make the change without it being configurable. Personally I'd think about moving the outfit information to the DB instead of lua if we were to change how outfit data is stored.

PR #3361, opened 2021-03-06
Fix crash when calling addEvent

yamaken93:
### Pull Request Prelude

### Changes Proposed

`addEvent(function() end)`
This is enough to crash TFS. Why? luajit api is sensible(which mean it will crash if misused) so the call to  
`lua_pop(globalState, 1);` here  https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L3678 will cause a crash since the code expects there is at least two parameters(callback and delay). You must pass a valid function as the first parameter thanks to the check here https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L3592.
I fixed that with two extra checks: Check if there is at least two parameters and if the second parameter is a lua number.
As a small optimization, we are not using std::list anymore because there was 0 reasons to use it instead of a std::vector.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -3684,7 +3684,7 @@ int LuaScriptInterface::luaAddEvent(lua_State* L)
 
 	LuaTimerEventDesc eventDesc;
 	eventDesc.parameters.reserve(parameters - 2); // safe to use -2 since we garanteed that there is at least two parameters
-	for (int i = 0; i < parameters - 2; ++i) { // safe to use -2 since we garanteed that there is at least two parameters
+	for (int i = 0; i < parameters - 2; ++i) {
 		eventDesc.parameters.push_back(luaL_ref(globalState, LUA_REGISTRYINDEX));
 	}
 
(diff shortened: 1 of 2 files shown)


DSpeichert (dismissed):

danyelvarejao (approved):

src/luascript.cpp
```diff
@@ -3670,7 +3683,8 @@ int LuaScriptInterface::luaAddEvent(lua_State* L)
 	}
 
 	LuaTimerEventDesc eventDesc;
-	for (int i = 0; i < parameters - 2; ++i) { //-2 because addEvent needs at least two parameters
+	eventDesc.parameters.reserve(parameters - 2); // safe to use -2 since we garanteed that there is at least two parameters
+	for (int i = 0; i < parameters - 2; ++i) { // safe to use -2 since we garanteed that there is at least two parameters
```
ranisalt, line 3687: Duplicate comment
```suggestion
	for (int i = 0; i < parameters - 2; ++i) {
```

MillhioreBT (approved):
Tested and work, LGTM
I was thinking of a warning for the minimum interval, and maybe it will accept any value convertible to number

ranisalt (approved):

EPuncker (approved):

Issue #3156, opened 2020-08-06
Add a console warning for invalid (non walkable) next positions


EPuncker:
I think we all already went thru a similar situation were players get stuck in some part of the map... a hole that leads you nowhere... a teleport that leads you to somewhere you can't move and so on... and its hard to figure it out as a owner if they don't report it, or don't even know how they got there in the first place, so here is my suggestion (**NO MECHANIC CHANGES**) : add a console warning with the position so we can be aware of such situations.

for example in these functions:

https://github.com/otland/forgottenserver/blob/master/data/lib/core/position.lua#L12-L48
https://github.com/otland/forgottenserver/blob/master/data/lib/core/tile.lua#L17-L38
stairs and holes

to prevent this:
![2020-08-06 13_36_30](https://user-images.githubusercontent.com/4684880/89563237-8663d080-d7f1-11ea-9523-1e07afc378cb.png)
![depois](https://user-images.githubusercontent.com/4684880/89563248-8a8fee00-d7f1-11ea-9f5c-13a5e1c69600.gif)

BahamutxD:
I think this would need to be optional to avoid spam on console as there are maps that deliberately use stairs/holes/blank teleports that lead nowhere or to non-pathable tiles as a way to trick/tease players.

EPuncker:
> I think this would need to be optional to avoid spam on console as there are maps that deliberately use stairs/holes/blank teleports that lead nowhere or to non-pathable tiles as a way to trick/tease players.

but those will only trigger if the player actually walk over them

BahamutxD:
> > I think this would need to be optional to avoid spam on console as there are maps that deliberately use stairs/holes/blank teleports that lead nowhere or to non-pathable tiles as a way to trick/tease players.
> 
> but those will only trigger if the player actually walk over them

Missunderstood you. Thought it was something that would be checked upon server start/map load. 

Nevermind me!

Zbizu:
What about checking if the destination is walkable before moving player there?

Zbizu:
make a walkback or prevent a movement command instead

if you remove the stairs/teleport, you might open the way for some treasure behind it

PR #3364, opened 2021-03-08
detect invalid stairs

idontreallywolf:
### Pull Request Prelude

### Changes Proposed
An attempt to fix stairs problem in #3156

- Replace stairs leading down to spots where players may get trapped.
- Remove ramps leading to a dead end (e.g no way to get back down or walk away)
- Print warning msg into console.
```
[Warning] Invalid stairs at position ( x / y / z )
``` 

#### Preview
img

**Issues addressed:** #3156 


Shawak:
I feel like this should be done on map loading and not in lua.

EPuncker:
> I feel like this should be done on map loading and not in lua.

as mentioned in the issue, there are maps that might have places like this that were made intentionally and are not reachable by players, so yes, the stepIn check is the best approach

now about the PR itself, wouldn't be easier to just edit the [Position:getNextPosition](https://github.com/otland/forgottenserver/blob/master/data/lib/core/position.lua#L12-L21) and [Position:moveUpstairs](https://github.com/otland/forgottenserver/blob/master/data/lib/core/position.lua#L12-L21) to check if the next position is walkable?

MillhioreBT:
> > I feel like this should be done on map loading and not in lua.
> 
> as mentioned in the issue, there are maps that might have places like this that were made intentionally and are not reachable by players, so yes, the stepIn check is the best approach
> 
> now about the PR itself, wouldn't be easier to just edit the [Position:getNextPosition](https://github.com/otland/forgottenserver/blob/master/data/lib/core/position.lua#L12-L21) and [Position:moveUpstairs](https://github.com/otland/forgottenserver/blob/master/data/lib/core/position.lua#L12-L21) to check if the next position is walkable?

I agree with you, there are places and stairs intentionally created this way, if you want a clean map just fix it from your map editor

Shawak:
> > I feel like this should be done on map loading and not in lua.
> 
> as mentioned in the issue, there are maps that might have places like this that were made intentionally and are not reachable by players, so yes, the stepIn check is the best approach

I still think it's bad behaviour to have two locations (items.xml and a lua file) where you will have to specify stairs. There should only be one location where you have to add a new stairs to make it work completly (including invalid checks).

And since TFS lets you define stairs in items.xml, I think you should either find a way to check it during server load or add a way of checking for this without having to speficy the IDs somewhere outside the items.xml file (e.g. moving the stair floorchange logic to lua and check it there, or check it in the cpp floor change function).

data/scripts/movements/stairscheck.lua
```diff
@@ -0,0 +1,70 @@
...
+	462,475,476,479,480,489,924,3135,3136,3137,3138,3219,3220,4834,4835,4836,4837,
+	5081,5731,6127,6128,6129,6130,6173,6917,6918,6919,6920,6921,6922,6923,6924,
+	8276,8277,8279,8280,8281,8282,8283,8284,8285,8286,8559,8560,8561,8562,8563,
+	8564,8565,8566,9574,9606,9846
+}
+
+local ramps = {
+	1385, 1388, 1390, 1394, 1392, 1396, 1398, 1400, 1402, 1404, 3687, 3688, 5259,
+	5258, 5260, 6909, 6911, 6913, 6915, 7924, 7925, 8376, 8372, 8374, 8378, 9573
+}
```
DSpeichert, line 12: This is not elegant and probably will get out of sync very soon. Can we do it more dynamically?

idontreallywolf: Am currently trying to do this in sources without requiring the IDs, will make a commit when it's (hopefully xd) ready.

data/global.lua
```diff
@@ -46,6 +46,13 @@ table.contains = function(array, value)
 	return false
 end
 
+table.merge = function(t1,t2)
```
MillhioreBT, line 49: `function(t1,t2)`
you should make a separation after the comma
example:
`function(t1, t2)`

data/scripts/movements/stairscheck.lua
```diff
@@ -0,0 +1,70 @@
...
+		p:sendMagicEffect(CONST_ME_POFF)
+		print("[Warning] Invalid stairs at position " .. positionToReadable(p))
+	end
+	return true
+end
+
+stairscheck:id(unpack(table.merge(stairs, ramps)))
+stairscheck:register()
+
+function findReplacementItem(position)
```
MillhioreBT, line 49: This function must be local since it will not be used anywhere else and if that were the case, this would not be the right place to declare functions in the global environment

data/scripts/movements/stairscheck.lua
```diff
@@ -0,0 +1,70 @@
...
+						return groundId
+					end
+				end
+			end
+		end
+	end
+	return defaultReplacementId
+end
+
+function positionToReadable(p)
```
MillhioreBT, line 68: ^ the same

data/lib/core/position.lua
```diff
@@ -18,7 +26,7 @@ function Position:getNextPosition(direction, steps)
 	end
 end
 
-function Position:moveUpstairs()
+function Position:moveUpstairs(cancelStep)
```
MillhioreBT, line 29: It is probably not necessary to modify this function, when we can only compare the positions and if they are the same, then it means that no tile was found

data/global.lua
```diff
@@ -46,6 +46,13 @@ table.contains = function(array, value)
 	return false
 end
 
+table.merge = function(t1, t2)
```
MillhioreBT, line 49: Why keep this function if it is no longer used?

idontreallywolf: I forgot to remove that one but don't you think it'd be useful in the future where someone might need to combine elements of two tables o.0?

If you(& maybe others) still believe that there's no point in keeping this function, I'll remove it.

DSpeichert: It should be in a separate PR in that case, so as not to create confusion on whether this is related.

Shawak:
Also why remove the item at all? Can't you just make it non walkable? Because maybe you want to make it clickable by giving it an action id or w/e.

idontreallywolf:
> Also why remove the item at all? Can't you just make it non walkable? Because maybe you want to make it clickable by giving it an action id or w/e.

I'm not sure I understand what you mean;
Removing the stairs is optional `patchInvalidStairs = false`.
What do you mean by making it clickable? usable? 🤔

data/lib/core/position.lua
```diff
@@ -42,6 +42,7 @@ function Position:moveUpstairs()
 				return self
 			end
 		end
+
```
EPuncker, line 45:
```suggestion
```

src/teleport.cpp
```diff
@@ -93,6 +95,42 @@ void Teleport::addThing(int32_t, Thing* thing)
...
+			if (creature) {
+				const Position creaturePos = creature->getPosition();
+				std::cout << fmt::format(
+					"[WARNING] Invalid teleport at position ({:d} / {:d} / {:d}) with destination ({:d} / {:d} / {:d})",
+					creaturePos.x, creaturePos.y, creaturePos.z,
+					destPos.x, destPos.y, destPos.z
+				) << std::endl;
+			} else {
+				std::cout << fmt::format(
+					"[WARNING] Invalid teleport with destionation ({:d} / {:d} / {:d})", destPos.x, destPos.y, destPos.z
```
EPuncker, line 127:
```suggestion
					"[Warning - Teleport::addThing] Invalid teleport with destination ({:d} / {:d} / {:d})", destPos.x, destPos.y, destPos.z
```

src/teleport.cpp
```diff
@@ -93,6 +95,42 @@ void Teleport::addThing(int32_t, Thing* thing)
...
+			}
+			if (pathExists) { break; }
+		}
+
+		if (!pathExists) {
+			Creature* creature = thing->getCreature();
+			if (creature) {
+				const Position creaturePos = creature->getPosition();
+				std::cout << fmt::format(
+					"[WARNING] Invalid teleport at position ({:d} / {:d} / {:d}) with destination ({:d} / {:d} / {:d})",
```
EPuncker, line 121:
```suggestion
					"[Warning - Teleport::addThing] Invalid teleport at position ({:d} / {:d} / {:d}) with destination ({:d} / {:d} / {:d})",
```

src/player.cpp
```diff
@@ -1280,6 +1283,76 @@ void Player::onCreatureMove(Creature* creature, const Tile* newTile, const Posit
...
+		const TileItemVector* tileItems = newTile->getItemList();
+		for (Item* item : *tileItems) {
+			if (item->hasProperty(CONST_PROP_IMMOVABLEBLOCKPATH)) {
+				g_game.internalRemoveItem(item);
+			}
+		}
+	}
+
+	std::cout << fmt::format(
+		"[WARNING] Invalid stairs at position ({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
```
EPuncker, line 1347:
```suggestion
		"[Warning - Player::onCreatureMove] Invalid stairs at position ({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
```

src/tile.cpp
```diff
@@ -236,6 +236,17 @@ BedItem* Tile::getBedItem() const
...
 
+Position Tile::getFloorchangeOffset() const
+{
+	if (hasFlag(TILESTATE_FLOORCHANGE_DOWN))  { return Position( 0,  0,  0); }
+	if (hasFlag(TILESTATE_FLOORCHANGE_NORTH)) { return Position( 0, -1,  0); }
+	if (hasFlag(TILESTATE_FLOORCHANGE_WEST))  { return Position(-1,  0,  0); }
+	if (hasFlag(TILESTATE_FLOORCHANGE_SOUTH)) { return Position( 0,  1,  0); }
+	if (hasFlag(TILESTATE_FLOORCHANGE_EAST))  { return Position( 1,  0,  0); }
+	
+	return Position( 0,  0,  0);
```
EPuncker, line 247:
```suggestion
	return Position(0,  0,  0);
```

EPuncker (changes requested):

src/player.cpp
```diff
@@ -1348,8 +1348,8 @@ void Player::checkInvalidStair(const Tile* newTile, const Position& newPos, cons
 	) << std::endl;
 
 	g_game.internalTeleport(this, oldPos);
-	sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, fmt::format(
-		"Invalid stairs found. Notify staff members ({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
+	sendTextMessage(MESSAGE_STATUS_SMALL, fmt::format(
+		"Invalid stairs found. Notify staff members Position({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
```
EPuncker, line 1352:
```suggestion
		"Invalid stairs found at Position({:d} / {:d} / {:d}). Please notify staff members.", newPos.x, newPos.y, newPos.z
```

src/player.cpp
```diff
@@ -1301,8 +1301,8 @@ void Player::checkInvalidStair(const Tile* newTile, const Position& newPos, cons
 			) << std::endl;
 
 			g_game.internalTeleport(this, oldPos);
-			sendTextMessage(MESSAGE_STATUS_CONSOLE_BLUE, fmt::format(
-				"Invalid stairs found. Notify staff members ({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
+			sendTextMessage(MESSAGE_STATUS_SMALL, fmt::format(
+				"Invalid stairs found. Notify staff members Position({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
```
EPuncker, line 1305:
```suggestion
				"Invalid stairs found at Position({:d} / {:d} / {:d}). Please notify staff members.", newPos.x, newPos.y, newPos.z
```

EPuncker (changes requested):

EPuncker (approved):

EPuncker (approved):

src/player.cpp
```diff
@@ -1280,6 +1283,76 @@ void Player::onCreatureMove(Creature* creature, const Tile* newTile, const Posit
...
+				newTile->patch();
+			}
+
+			std::cout << fmt::format(
+				"[Warning - Player::onCreatureMove] Invalid stairs at position ({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
+			) << std::endl;
+
+			g_game.internalTeleport(this, oldPos);
+			sendTextMessage(MESSAGE_STATUS_SMALL, fmt::format(
+				"Invalid stairs found at Position({:d} / {:d} / {:d}). Please notify staff members.", newPos.x, newPos.y, newPos.z
```
ranisalt, line 1305: Wait, is this supposed to send the information to the player?

src/player.cpp
```diff
@@ -1280,6 +1283,76 @@ void Player::onCreatureMove(Creature* creature, const Tile* newTile, const Posit
...
+	if (newTile->hasFlag(TILESTATE_FLOORCHANGE_DOWN)) {
+		Tile* tileBelow = g_game.map.getTile(newPos.x, newPos.y, newPos.z + 1);
+		if (!tileBelow || !tileBelow->isWalkable()) {
+			if (g_config.getBoolean(ConfigManager::PATCH_INVALID_STAIRS)) {
+				newTile->patch();
+			}
+
+			std::cout << fmt::format(
+				"[Warning - Player::onCreatureMove] Invalid stairs at position ({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
+			) << std::endl;
```
ranisalt, line 1301: Use `fmt::print`

src/player.cpp
```diff
@@ -1280,6 +1283,76 @@ void Player::onCreatureMove(Creature* creature, const Tile* newTile, const Posit
...
+				"Invalid stairs found at Position({:d} / {:d} / {:d}). Please notify staff members.", newPos.x, newPos.y, newPos.z
+			));
+		}
+		return;
+	}
+
+	bool walkableTileFound = false;
+
+	Position tileNextPos = newTile->getFloorchangeOffset();
+	tileNextPos.z--;
```
ranisalt, line 1314:
```suggestion
	--tileNextPos.z;
```

src/player.cpp
```diff
@@ -1280,6 +1283,76 @@ void Player::onCreatureMove(Creature* creature, const Tile* newTile, const Posit
...
+		for (Item* item : *tileItems) {
+			if (item->hasProperty(CONST_PROP_IMMOVABLEBLOCKPATH)) {
+				g_game.internalRemoveItem(item);
+			}
+		}
+	}
+
+	std::cout << fmt::format(
+		"[Warning - Player::onCreatureMove] Invalid stairs at position ({:d} / {:d} / {:d})", newPos.x, newPos.y, newPos.z
+	) << std::endl;
```
ranisalt, line 1348: Same, `fmt::print`. Mind the newline.

src/player.h
```diff
@@ -1155,6 +1155,8 @@ class Player final : public Creature, public Cylinder
 	private:
 		std::forward_list<Condition*> getMuteConditions() const;
 
+		void checkInvalidStair(const Tile* newTile, const Position& newPos, const Position& oldPos);
```
ranisalt, line 1158: I would call it `checkStairs` instead, because 1) you check all of them, not only the invalid 2) "stair" is a single step, "stairs" is the correct term

src/teleport.cpp
```diff
@@ -93,6 +95,47 @@ void Teleport::addThing(int32_t, Thing* thing)
 		return;
 	}
 
+	// Destination may be non-walkable on purpose ..
+	if (!destTile->isWalkable()) {
+		// .. but make sure players won't get stuck
+		bool pathExists = false;
+		for (uint16_t x = destPos.x - 1; x <= destPos.x + 1; x++) {
+			for (uint16_t y = destPos.y - 1; y <= destPos.y + 1; y++) {
```
ranisalt, line 103: Always use `++var` format unless you specifically need the previous value.

src/teleport.cpp
```diff
@@ -93,6 +95,47 @@ void Teleport::addThing(int32_t, Thing* thing)
...
+				if (x == destPos.x && y == destPos.y) {
+					continue;
+				}
+
+				if (g_game.map.getTile(x, y, destPos.z)->isWalkable()) {
+					pathExists = true;
+					break;
+				}
+			}
+			if (pathExists) { break; }
```
ranisalt, line 113: Either break lines around `break` or remove the brackets

src/teleport.cpp
```diff
@@ -93,6 +95,47 @@ void Teleport::addThing(int32_t, Thing* thing)
...
+
+		if (!pathExists) {
+			Creature* creature = thing->getCreature();
+			if (creature) {
+				const Position creaturePos = creature->getPosition();
+				std::cout << fmt::format(
+					"[Warning - Teleport::addThing] Invalid teleport at position ({:d} / {:d} / {:d}) with destination ({:d} / {:d} / {:d})",
+					creaturePos.x, creaturePos.y, creaturePos.z,
+					destPos.x, destPos.y, destPos.z
+				) << std::endl;
```
ranisalt, line 124: `fmt::print`

src/teleport.cpp
```diff
@@ -93,6 +95,47 @@ void Teleport::addThing(int32_t, Thing* thing)
...
+
+				Player* player = creature->getPlayer();
+				player->sendTextMessage(MESSAGE_STATUS_SMALL, fmt::format(
+					"Invalid Destination at Position({:d} / {:d} / {:d}). Please notify staff members.", destPos.x, destPos.y, destPos.z
+				));
+			} else {
+				std::cout << fmt::format(
+					"[Warning - Teleport::addThing] Invalid teleport with destination ({:d} / {:d} / {:d})", destPos.x, destPos.y, destPos.z
+				) << std::endl;
```
ranisalt, line 133: `fmt::print`

src/tile.cpp
```diff
@@ -236,6 +236,17 @@ BedItem* Tile::getBedItem() const
 	return nullptr;
 }
 
+Position Tile::getFloorchangeOffset() const
+{
+	const Position p = getPosition();
+	if (hasFlag(TILESTATE_FLOORCHANGE_DOWN))  { return Position(0, 0, 0); }
```
ranisalt, line 242: This is the default, so you can omit this branch.

src/tile.cpp
```diff
@@ -236,6 +236,17 @@ BedItem* Tile::getBedItem() const
 	return nullptr;
 }
 
+Position Tile::getFloorchangeOffset() const
+{
+	const Position p = getPosition();
```
ranisalt, line 241:
```suggestion
	const Position& p = getPosition();
```

src/tile.cpp
```diff
@@ -852,6 +863,32 @@ Tile* Tile::queryDestination(int32_t&, const Thing&, Item** destItem, uint32_t&
 	return destTile;
 }
 
+void Tile::patch() const
+{
+	uint16_t patchTileId = 0;
+
+	const Position& p = getPosition();
+	for(uint16_t x = p.x-1; x <= p.x+1; x++) {
+		for(uint16_t y = p.y-1; y <= p.y+1; y++) {
```
ranisalt, line 872: Missing space after `for`

src/tile.cpp
```diff
@@ -852,6 +863,32 @@ Tile* Tile::queryDestination(int32_t&, const Thing&, Item** destItem, uint32_t&
...
+				continue;
+			}
+
+			Tile* t = g_game.map.getTile(x, y, p.z);
+			if (t && t->isWalkable()) {
+				patchTileId = t->getGround()->getID();
+				break;
+			}
+		}
+		if(patchTileId > 0) {
```
ranisalt, line 883:
```suggestion
		if (patchTileId > 0) {
```

src/tile.cpp
```diff
@@ -852,6 +863,32 @@ Tile* Tile::queryDestination(int32_t&, const Thing&, Item** destItem, uint32_t&
...
+			break;
+		}
+	}
+
+	// default patch id : 100 (void)
+	g_game.transformItem(ground, patchTileId > 0 ? patchTileId : 100);
```
ranisalt, line 889: Why not defaulting it to 100 and skipping this ternary?

src/tile.cpp
```diff
@@ -1614,6 +1651,27 @@ bool Tile::isMoveableBlocking() const
...
+	if (isMoveableBlocking()) {
+		return false;
+	}
+
+	const TileItemVector* tileItems = getItemList();
+	if (!tileItems) {
+		return true;
+	}
+
+	for(const Item* item : *tileItems) {
```
ranisalt, line 1665:
```suggestion
	for (const Item* item : *tileItems) {
```

src/tile.h
```diff
@@ -174,6 +174,8 @@ class Tile : public Cylinder
 			return false;
 		}
 
+		Position getFloorchangeOffset() const;
```
ranisalt, line 177: It should be `getFloorChangeOffset`, "floor change" is separate

ranisalt (changes requested):
Note that you have not followed proper The Forgotten Server code styling.

That said, is there a reason not to perform this check after loading the map and fixing stairs before players have the opportunity to use them?

idontreallywolf:
> is there a reason not to perform this check after loading the map and fixing stairs before players have the opportunity to use them?
 
@EPuncker has previously replied to @Shawak about the same question & I agreed with what he said:
> > I feel like this should be done on map loading and not in lua.
>
> as mentioned in the issue, there are maps that might have places like this that were made intentionally and are not reachable by players, so yes, the stepIn check is the best approach

src/tile.cpp
idontreallywolf: Thought it might look odd to have a void tile and therefore decided to copy one of the available tiles around (if any).

ranisalt:
> as mentioned in the issue, there are maps that might have places like this that were made intentionally and are not reachable by players, so yes, the stepIn check is the best approach

It makes a lot of sense indeed.

src/tile.cpp
ranisalt: I mean, just set `patchTileId = 100` from the start. So, if it is not set in the logic, it's already on a valid value and you can write only

```suggestion
	g_game.transformItem(ground, patchTileId);
```

src/tile.cpp
```diff
@@ -852,6 +862,32 @@ Tile* Tile::queryDestination(int32_t&, const Thing&, Item** destItem, uint32_t&
...
+				break;
+			}
+		}
+
+		if (patchTileId > 100) {
+			break;
+		}
+	}
+
+	g_game.transformItem(ground, patchTileId);
```
ranisalt, line 888: I'm not a huge fan of this style and would rather do this:

```
	g_game.transformItem(ground, [&](const Position& p) {
		for (uint16_t x = p.x - 1; x <= p.x + 1; ++x) {
			for (uint16_t y = p.y - 1; y <= p.y + 1; ++y) {
				if (x == p.x && y == p.y) {
					continue;
				}
				Tile* t = g_game.map.getTile(x, y, p.z);
				if (t && t->isWalkable()) {
					return t->getGround()->getID();
				}
			}
		}
		return 100;
	}(getPosition()));
```

This is called *immediately invoked function expression* (IIFE for short) and I borrow the idea from Javascript. However, since our codebase in general does not use IIFEs, I'm just leaving it here as food for though, feel free to dismiss it.

ranisalt (dismissed):

TheSumm:
I like this a lot and have something similar in mind for teleports once. They way I would have solved it, would be to make the engine warn you if such stairs are detected. There were concerns that some of these cases might be intended. Therefore, I would have provided a method for the user to suppress those warnings (probably by assigning a reserved action id to the stairs).

src/tile.cpp
MillhioreBT: I love your style of using lambdas ;)

EPuncker (dismissed):

src/player.cpp
```diff
@@ -1291,20 +1291,42 @@ void Player::checkStairs(const Tile* newTile, const Position& newPos, const Posi
...
 
-		dir = (isNorthWest || isNorthEast) ? DIRECTION_NORTH : dir;
-		dir = (isSouthWest || isSouthEast) ? DIRECTION_SOUTH : dir;
+/*
+	1. find stairs on sides
+		- if found:
+			- set offSet X & Y based on opposite floorchange
+		- else
+			- check bottom; 0 offset.
+
```
EPuncker:
```suggestion
```

EPuncker (approved):
okay, now I've tested several cases and they seem to be fine now

DSpeichert:
@idontreallywolf needs rebase

Zbizu:
make a walkback or prevent a movement command instead (or update the post if you already did)

if you remove the stairs/teleport, you might open the way for some treasure behind it

PR #3365, opened 2021-03-10
Random decay time

### LINKED ISSUE #3148: Magic Wall and Wild Growth missing changes

EPuncker:
Co-Authored-By: Sarah Wesker <sarahelizabetwesker@gmail.com>

### Pull Request Prelude

### Changes Proposed
Added the possibility to items to have random decay time

**Issues addressed:** 
closes #3148

### DIFF
--- a/data/items/items.xml
+++ b/data/items/items.xml
@@ -706,7 +706,8 @@
 	<item id="1497" article="a" name="magic wall">
 		<attribute key="type" value="magicfield" />
 		<attribute key="decayTo" value="0" />
-		<attribute key="duration" value="20" />
+		<attribute key="durationMin" value="14" />
+		<attribute key="durationMax" value="20" />
 	</item>
 	<item id="1498" article="a" name="magic wall">
 		<attribute key="type" value="magicfield" />
@@ -714,7 +715,8 @@
 	<item id="1499" article="a" name="rush wood">
 		<attribute key="type" value="magicfield" />
 		<attribute key="decayTo" value="0" />
-		<attribute key="duration" value="45" />
+		<attribute key="durationMin" value="38" />
+		<attribute key="durationMax" value="45" />
 		<attribute key="blocking" value="1" />
 	</item>
 	<item id="1500" article="a" name="fire field">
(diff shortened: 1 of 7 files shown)


data/items/items.xml
```diff
@@ -706,15 +706,17 @@
 	<item id="1497" article="a" name="magic wall">
 		<attribute key="type" value="magicfield" />
 		<attribute key="decayTo" value="0" />
-		<attribute key="duration" value="20" />
+		<attribute key="durationMin" value="14" />
+		<attribute key="durationMax" value="20" />
```
ranisalt, line 710: What about `<attribute key="duration" minvalue="14" maxvalue="20" />`, or even `min` and `max`?

MillhioreBT: It doesn't seem like a bad idea, I like the way you think

EPuncker: but how would it work with the remaining items that use duration? 3 parameters?

ranisalt: It's either `value`, or `min` and `max`. If the item specifies `value`, don't even try to read the others.

yamaken93: Maybe value, minValue and maxValue. Somehow i don't like how its called value then it just min or max.

Zbizu: I suggest duration and durationMax - would be consistent with count and countMax in monsters loot

DSpeichert (commented):
I'm technically in favor of proper XML, copying @ranisalt's comment:
> It's either value, or min and max. If the item specifies value, don't even try to read the others.

EPuncker:
closed because no interest from anyone in helping

PR #3366, opened 2021-03-11
Fix luascript lua state inconsistency

yamaken93:
### Pull Request Prelude

### Changes Proposed

Remove legacy code from addEvent/stopEvent. Its 2013 code from a time that tfs was making a transition from multiple lua states to a single one and is useless now.
The only reason to keep the lua_xmove inside addEvent is if someone wants to use addevent from config.lua lua state or the database migration lua state (tfs has 3 lua states, one created and destructed for reading config.lua content, one that is created and destructed for database migration and the lua state that all run all the scripting code and is never destructed). I don't believe it was ever intended to be used there and as far i tested its not working, a call to lua_settop(globalState, 0) is needed before lua_xmove otherwise it won't grab the parameters and addEvent call will fail. Also previously to this code there is all this mess when printing the backtrace where the code only expects to exists one lua state(the scripting one).

Now addEvent accepts to be called from inside a coroutine (before even with lua_xmove it could not get the parameters, lack of lua_settop call) and when there is bad parameter when calling from a coroutine(or any lua state) we will print the correct backtrace.

About coroutines: they are a new lua state(lua_newthread) that is children of the main lua state and they have their own call stack but the whole global state is shared with the father lua state. the "newthread" is only to confuse us, its not a new OS thread.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -99,7 +99,7 @@ bool ScriptEnvironment::setCallbackId(int32_t callbackId, LuaScriptInterface* sc
 	if (this->callbackId != 0) {
 		//nested callbacks are not allowed
 		if (interface) {
-			interface->reportErrorFunc("Nested callbacks!");
+			reportErrorFunc(interface->getLuaState(), "Nested callbacks!");
 		}
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


ranisalt (approved):
LGTM. Always good to have some code consistency improvements.

DSpeichert (approved):

PR #3372, opened 2021-03-16
Removed unused check in onDropLoot

EPuncker:
### Pull Request Prelude

### Changes Proposed
https://github.com/otland/forgottenserver/blob/master/src/monsters.cpp#L1360-L1395

since now we have several checks to ensure that the monster loot is valid, this check wouldn't ever run at all

**Issues addressed:** 


nekiro:
No point in removing it if you ask me. It's the "just in case" situation.

PR #3375, opened 2021-03-18
Unnecessary whitespaces

rookgaard:
### Pull Request Prelude

### Changes Proposed

I know it's not super important, but it's nice to have a little cleaner code (although I still don't like how long lines are divided, like in [this bad example][bad] with a lot of spaces, just to align to line above).

[bad]: https://github.com/otland/forgottenserver/blob/master/src/actions.cpp#L451

### DIFF
--- a/data/lib/core/item.lua
+++ b/data/lib/core/item.lua
@@ -147,7 +147,7 @@ do
 			if attack ~= 0 then
 				begin = false
 				ss:append(' (Atk:%d', attack)
-				
+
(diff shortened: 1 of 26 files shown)


MillhioreBT:
For me some things are already fine, In fact, the example shown in the link is not bad, it was done that way intentionally, and that will depend on whether the guys from tf decide to change this style

EPuncker (approved):
my eyes says thank you

rookgaard:
@EPuncker what are conditions when PR is actually accepted and merged? How can I request for review for example from @Znote or @DSpeichert ?

PR #3383, opened 2021-03-24
Fix removeTileCreature client debug when spectator can't see the pos

### LINKED ISSUE #3381: Client crash on remove creature

1. Log in 2 characters in the same location.
2. Let a monster target character A while character B is within view range.
3. Character B walks off screen.
4. Character A kills the monster.
5. Client of character B crashes.

Video demonstration: https://youtu.be/IeefUzWqG4Y
OTLand discussion: https://otland.net/threads/random-server-and-client-crashes.274828
### Expected behaviour

Client should not crash, server should not send RemoveTileThing when the creature is not in view range.

### Actual behaviour

The server sends RemoveTileThing when the creature is outside view range, causing the client to crash.

### Environment

Not environment dependent.
Tested on 29d484ba0554c6d985244bedda9313d9bb8b9c68

### Possible solution
Add ```
if (!canSee(pos)) {
    return;
}```
In [ProtocolGame::sendRemoveTileCreature](https://github.com/otland/forgottenserver/blob/master/src/protocolgame.cpp#L2440)

nekiro:
Fix #3381

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2439,8 +2439,21 @@ void ProtocolGame::sendRemoveTileThing(const Position& pos, uint32_t stackpos)
 
 void ProtocolGame::sendRemoveTileCreature(const Creature* creature, const Position& pos, uint32_t stackpos)
 {
+	if (stackpos < 10) {
+		if (!canSee(pos)) {
+			return;
+		}
+
+		NetworkMessage msg;
+		RemoveTileThing(msg, pos, stackpos);
+		writeToOutputBuffer(msg);
+		return;
+	}
+
 	NetworkMessage msg;
-	RemoveTileCreature(msg, creature, pos, stackpos);
+	msg.addByte(0x6C);
+	msg.add<uint16_t>(0xFFFF);
+	msg.add<uint32_t>(creature->getID());
 	writeToOutputBuffer(msg);
 }
 


EPuncker (approved):

MillhioreBT:
Why not simply add:
```cpp
if (!canSee(pos)) {
    return;
}
```

It has nothing wrong and has the same behavior as the code that you have changed, in fact it would be 3 extra lines instead of 14

nekiro:
> Why not simply add:
> 
> ```c++
> if (!canSee(pos)) {
>     return;
> }
> ```
> 
> It has nothing wrong and has the same behavior as the code that you have changed, in fact it would be 3 extra lines instead of 14
> 
> in fact the same guy who published the problem mentioned it:
> PR: #3381
> image

https://github.com/otland/forgottenserver/pull/3383/files#diff-eb01d79b279ebf659ed73216c0b3c490685221c9358e058ce55a1c8ab4e99c35R2456

MillhioreBT:
Yes, is exactly the same
image
image

If what I say is wrong, I will have to buy glasses and not only glasses for the eyes but also for my logic

forgee:
I have been testing both methods and I can't tell a difference either way.

What nekiro said makes sense though. If you just add canSee(pos), it works the same, but always checking the position might be unnecessary.

MillhioreBT:
I'm just giving an opinion, Really i don't get crash when i try to reproduce them with the steps that i mention in PR #3381 so please ignore my suggestions, i can't reproduce the problem with the original source code, i use Tibia Client 10 and not Client 10.98

nekiro:
> Yes, is exactly the same
> image
> image
> 
> If what I say is wrong, I will have to buy glasses and not only glasses for the eyes but also for my logic

No, it's not. I'm not executing "canSee", when stackpos is higher or equal to 10, cause we are removing by creature id and not by position then,

MillhioreBT:
@Nekiro you are right it is not the same.

I remember that this code never caused problems in the past, there is some confirmation that made this happen, it would be good to find out what it was the root problem instead of cutting off the legs of this leak

forgee:
> @nekiro you are right it is not the same.
> 
> I remember that this code never caused problems in the past, there is some confirmation that made this happen, it would be good to find out what it was the root problem instead of cutting off the legs of this leak

You are right, it is a recent issue. I think it was this commit.
https://github.com/otland/forgottenserver/commit/3c6f936eff92041b6dda1df0fd82e3abfc0d08ea#diff-eb01d79b279ebf659ed73216c0b3c490685221c9358e058ce55a1c8ab4e99c35

EPuncker:
confirmed to be working:
https://otland.net/threads/client-crashes-10-98-latest-tfs.275157/page-2#post-2652565

PR #3351, opened 2021-02-26
Use fmt to format messages

ranisalt:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

[fmt](https://fmt.dev) is a library that provides fast and safe string formatting, with a syntax similar to `printf` (or `str.format` for my fellow Pythonists).

This pull request adds it as a dependency, and also changes some usages of stringstreams for formatting where `fmt::format` would suit objectively better. There are many more places there `fmt::format` would fit (just grep for `std::ostringstream` to find), but I can't do that much alone 😛 

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #3385, opened 2021-03-25
Feature/skull system

rookgaard:
### Pull Request Prelude

### Changes Proposed

Introduced skull system described in https://tibia.fandom.com/wiki/Skull_System


rookgaard:
@EPuncker Can you add reviewers?

src/player.cpp
```diff
@@ -3865,14 +3865,49 @@ void Player::addUnjustifiedDead(const Player* attacked)
 
 	sendTextMessage(MESSAGE_EVENT_ADVANCE, "Warning! The murder of " + attacked->getName() + " was not justified.");
 
-	skullTicks += g_config.getNumber(ConfigManager::FRAG_TIME);
+	time_t now = time(NULL), today = (now - 84600), week = (now - (7 * 84600));
```
nekiro, line 3868: NULL > nullptr

src/iologindata.cpp
```diff
@@ -1073,3 +1073,17 @@ void IOLoginData::updatePremiumTime(uint32_t accountId, time_t endTime)
...
+void IOLoginData::getUnjustifiedDates(std::string name, std::vector<time_t>& dateList, time_t _time)
+{
+	Database& db = Database::getInstance();
+	std::ostringstream query;
+	query << "SELECT `time` FROM `player_deaths` WHERE `killed_by` = " << db.escapeString(name) << " AND `unjustified` = 1 AND `time` >= " << (_time - (30 * 86400));
+	DBResult_ptr result = db.storeQuery(query.str());
+
+	if (result) {
+		do {
+			dateList.push_back((time_t)result->getNumber<uint32_t>("time"));
```
nekiro, line 1086: static_cast<time_t>

src/iologindata.cpp
```diff
@@ -1073,3 +1073,17 @@ void IOLoginData::updatePremiumTime(uint32_t accountId, time_t endTime)
 	query << "UPDATE `accounts` SET `premium_ends_at` = " << endTime << " WHERE `id` = " << accountId;
 	Database::getInstance().executeQuery(query.str());
 }
+
+void IOLoginData::getUnjustifiedDates(std::string name, std::vector<time_t>& dateList, time_t _time)
```
nekiro, line 1077: dont prefix variables or arguments with underscore

src/player.cpp
```diff
@@ -3865,14 +3865,49 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+	std::vector<time_t> killsList;
+	killsList.push_back(now); // add current kill to list
+	IOLoginData::getUnjustifiedDates(name, killsList, now); // add kills from last month to list
+	uint32_t
```
nekiro, line 3872: I don't think that's tfs code style like

src/player.cpp
```diff
@@ -3865,14 +3865,49 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+		todayKills = 0,
+		weekKills = 0,
+		monthKills = killsList.size();
 
-	if (getSkull() != SKULL_BLACK) {
-		if (g_config.getNumber(ConfigManager::KILLS_TO_BLACK) != 0 && skullTicks > (g_config.getNumber(ConfigManager::KILLS_TO_BLACK) - 1) * static_cast<int64_t>(g_config.getNumber(ConfigManager::FRAG_TIME))) {
-			setSkull(SKULL_BLACK);
-		} else if (getSkull() != SKULL_RED && g_config.getNumber(ConfigManager::KILLS_TO_RED) != 0 && skullTicks > (g_config.getNumber(ConfigManager::KILLS_TO_RED) - 1) * static_cast<int64_t>(g_config.getNumber(ConfigManager::FRAG_TIME))) {
+	for (std::vector<time_t>::iterator it = killsList.begin(); it != killsList.end(); ++it) {
```
nekiro, line 3877: for (const auto& it : killsList) ?

src/player.cpp
```diff
@@ -3865,14 +3865,49 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+			weekKills++;
+		}
+
+		if ((*it) > today) {
+			todayKills++;
+		}
+	}
+
+	if (getSkull() < SKULL_RED) {
+		uint32_t
```
nekiro, line 3888: same as before

src/player.cpp
```diff
@@ -3865,14 +3865,49 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+			redDailyLimit = g_config.getNumber(ConfigManager::RED_DAILY_LIMIT),
+			redWeekLimit = g_config.getNumber(ConfigManager::RED_WEEKLY_LIMIT),
+			redMonthLimit = g_config.getNumber(ConfigManager::RED_MONTHLY_LIMIT);
+
+		if (todayKills >= redDailyLimit || weekKills >= redWeekLimit || monthKills >= redMonthLimit) {
 			setSkull(SKULL_RED);
+			skullTicks = g_config.getNumber(ConfigManager::RED_SKULL_LENGTH);
+		}
+	} else if (getSkull() == SKULL_RED) {
+		uint32_t
```
nekiro, line 3898: same as before

nekiro (changes requested):

MillhioreBT:
Please use the new string formatting style `fmt`

rookgaard:
@MillhioreBT Where do you want me to use that formatting?

MillhioreBT:
> @MillhioreBT Where do you want me to use that formatting?

Sorry for not specifically suggesting where
In everything that makes use of: `std::ostringstream`

This is suggested in the PR #3351, apparently the idea is to get rid of the use of `std::ostringstream` and using `fmt` instead

rookgaard:
@MillhioreBT In my opinion, as long as there are usages of `ostringstream` inside that file, we shouldn't mix up different approaches. There should be a separate PR to replace all uses of `fmt` in `iologindata.cpp`.

src/player.cpp
```diff
@@ -3865,14 +3865,44 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+	IOLoginData::getUnjustifiedDates(name, killsList, now); // add kills from last month to list
+	int32_t todayKills = 0, weekKills = 0, monthKills = killsList.size();
 
-	if (getSkull() != SKULL_BLACK) {
-		if (g_config.getNumber(ConfigManager::KILLS_TO_BLACK) != 0 && skullTicks > (g_config.getNumber(ConfigManager::KILLS_TO_BLACK) - 1) * static_cast<int64_t>(g_config.getNumber(ConfigManager::FRAG_TIME))) {
-			setSkull(SKULL_BLACK);
-		} else if (getSkull() != SKULL_RED && g_config.getNumber(ConfigManager::KILLS_TO_RED) != 0 && skullTicks > (g_config.getNumber(ConfigManager::KILLS_TO_RED) - 1) * static_cast<int64_t>(g_config.getNumber(ConfigManager::FRAG_TIME))) {
+	for (const auto& it : killsList) {
+		if (it > week) {
+			weekKills++;
```
ranisalt, line 3876:
```suggestion
			++weekKills;
```

src/player.cpp
```diff
@@ -3865,14 +3865,44 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+		}
+
+		if (it > today) {
+			todayKills++;
```
ranisalt, line 3880:
```suggestion
			++todayKills;
```

src/iologindata.cpp
```diff
@@ -1073,3 +1073,17 @@ void IOLoginData::updatePremiumTime(uint32_t accountId, time_t endTime)
 	query << "UPDATE `accounts` SET `premium_ends_at` = " << endTime << " WHERE `id` = " << accountId;
 	Database::getInstance().executeQuery(query.str());
 }
+
+void IOLoginData::getUnjustifiedDates(std::string name, std::vector<time_t>& dateList, time_t offsetTime)
+{
+	Database& db = Database::getInstance();
+	std::ostringstream query;
+	query << "SELECT `time` FROM `player_deaths` WHERE `killed_by` = " << db.escapeString(name) << " AND `unjustified` = 1 AND `time` >= " << (offsetTime - (30 * 86400));
```
ranisalt, line 1081:
```suggestion
	query << "SELECT `time` FROM `player_deaths` WHERE `killed_by` = " << db.escapeString(name) << " AND `unjustified` = 1 AND `time` >= " << offsetTime - 30 * 86400;
```

src/iologindata.h
```diff
@@ -58,6 +58,7 @@ class IOLoginData
 		static void removeVIPEntry(uint32_t accountId, uint32_t guid);
 
 		static void updatePremiumTime(uint32_t accountId, time_t endTime);
+		static void getUnjustifiedDates(std::string name, std::vector<time_t>& dateList, time_t offsetTime);
```
ranisalt, line 61: This function should return the container instead.

```suggestion
		static std::vector<time_t> getUnjustifiedDates(std::string name, time_t offsetTime);
```

ranisalt (changes requested):

src/iologindata.h
rookgaard: @ranisalt Why it should? I'm not asking to deny your suggestion, but as I'm not that experienced in C, I wanted to know difference between returning vector and modyfing declared somewhere else.

src/player.cpp
rookgaard: May I ask what's the difference, when both variables are used outside of a loop?

ranisalt: Semantics. Only use `x++` if you are capturing the previous value, i.e. `y = x++`, since they don't have the same meaning.

If by not outside a loop you mean the `x++` in `for (...; ...; x++)`, note that even in that case the expressions are similar (you could use `++x` without change in behavior)

src/iologindata.h
ranisalt: > but as I'm not that experienced in C,

Specifically because this is not C. In C, you would pass a pointer to the output variable and return the result (e.g. 0 if success, 1 if error), but in C++ we have exceptions, so a function should return its output and throw when an error occurs.

Besides that, there's a myriad of optimizations that can be performed by the compiler in return values that can't be applied with function parameters, because when you return a value you have several constraints (e.g. you don't know if the parameter vector came in empty or with some values, if it was allocated on the stack or heap...).

Overall, it's a better function signature. The function is called "get me something" but it does not return you anything just feels weird in this context :P

ranisalt: By the way, you probably want to take in `const std::string& name` instead of `std::string name` to avoid a copy.

src/player.cpp
```diff
@@ -3865,14 +3865,43 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+		if (it > week) {
+			++weekKills;
+		}
+
+		if (it > today) {
+			++todayKills;
+		}
+	}
+
+	if (getSkull() < SKULL_RED) {
```
nekiro, line 3883: define getSkull into variable to avoid calling it 3 times in worse case scenario

src/player.cpp
```diff
@@ -3865,14 +3865,43 @@ void Player::addUnjustifiedDead(const Player* attacked)
...
+		if (
```
nekiro, line 3884: this if is super ugly, put it in one line, you could define these config values into variables to make it shorter if needed

nekiro (changes requested):

ranisalt (commented):
I would rather have the control be done in Lua instead, but I consider it an improvement over the current situation.

MillhioreBT:
> I would rather have the control be done in Lua instead, but I consider it an improvement over the current situation.

I also support the idea of ​​handling this in lua, as this seems to be a feature that anyone would want to manipulate for their own purposes and modifications ;), but as rani says, this is fine in any case

nekiro:
It actually shouldn't be hard to move to lua, it's bunch of queries and some logic and we have ondeath/onpreparedeath/onkill events, so you can slap skull system there easily.

EPuncker (approved):
Lets leave the Lua part to another PR shall we? 😛 I'm in favor of this change as it is an improvement of the current one

rookgaard:
@nekiro @DSpeichert

Znote (commented):
I must admit I like the OT skull system more, I have had lots of fun on some custom rpg styled pvp servers that had like a constant 2-4 hour to loose 1 frag, without any daily/weekly/monthly limits. As it opens up for the possibility to kill lots of people daily without getting red/black skull, while still preventing "mass pk incidents" where someone would slay all lowbies in the area. 

So I think you should re-implement the previous behavior as an optional toggle in config that just uses timeToDecreaseFrags. Retain "backwards compatibility" with skull system, perhaps we can call it `oldOTSkullSystem = bool`
Instead of fixed red/black skull duration, make the duration be based on remaining frags (when you have 0 frags, you loose red/black skull). 

If I have 2 hour to decrease frag, how would I do it with this new skull system? Go 24 hour (daily limit) / 2 = 12 frags to red skull, set weekly and monthly to 9000+ to avoid those calculations. The risk is here that someone will spam UE and kill 12 people right off the bat, whereas with the retro OT skull system they had to "space out" their killing spree throughout the day. (Since they would get red skull after 3 kills, but are relatively quickly loosing frags).

ranisalt:
I don't see it being one system vs the other but rather as different parameters to the same system. We could even ship presets, e.g. make `skulls` in `config.lua` be a table and allow to set `skulls = skullPreset.default` or `skulls = skullPreset.legacy`

omarcopires:
If this is the correct behavior there is no reason not to add it to 1.5, and for those who like the old version, they could simply use version 1.4 along with the 10.98 protocol.

omarcopires (dismissed):

DSpeichert (dismissed):

DSpeichert:
We could leave both systems in place simultaneously - have per day/week/month limits + frag decay and red skull threshold.

But this is okay new behavior as-is.

EPuncker (dismissed):

EPuncker:
@DSpeichert I rebased it and solved conflicts.

ramon-bernardo (approved):

src/player.cpp
```diff
@@ -4039,17 +4039,49 @@ void Player::addUnjustifiedDead(const Player* attacked)
 
 	sendTextMessage(MESSAGE_EVENT_ADVANCE, "Warning! The murder of " + attacked->getName() + " was not justified.");
 
-	skullTicks += g_config.getNumber(ConfigManager::FRAG_TIME);
+	Skulls_t playerSkull = getSkull();
+	if (playerSkull == SKULL_ORANGE) {
```
yamaken93, line 4043: As far i understand this orange skull check here does not make sense. The orange skull must be checked on the attacked player (the player which has been killed) and its a "client skull" (a skull that only appear to you at certain conditions like yellow skull, party skull, warmode skull etc) manipulated by the Creature::getSkullClient function.

Felipe21:
when merged is going to be added?

ranisalt:
I believe this system can be implemented in Lua. When a player is killed by another, recalculate the skull; using a storage value to keep track of skull duration and an event to clear it if the player is online when it expires, or check on login. We already have skull functions in Lua.

Issue #3378, opened 2021-03-22
Save a house every X minutes after an item inside it is moved / added / removed.


brunotonin:
### Explanation of what you want to do that is currently impossible
Save a house every X minutes after an item inside it is moved / added / removed.
This would be very good for cases of crash on the server, the players do not lose the items in their homes.

### Desired functionality
I think the ideal would be that every time an item is changed in the house, schedule to save the items in the house in X minutes, to avoid many interactions with the DB. event of saving the previous house is canceled and another one is created, or the time of the previous one is reset.

MillhioreBT:
It sounds interesting, but I don't know how efficient this would be.

However, it is assumed that the server should not suffer crashes if it is in production, it would be a very specific case and very very rare, and usually it is our own fault, you know some script with infinite loops, some incorrect call to a already  deleted object in C or something like that

but it would be nice to add a function that does this work and you would work this case yourself from a script in the onItemMoved event

brunotonin:
> saveHouse(...)

Really the ideal would be that it does not happen, however many new servers have this problem and when it happens with big servers the impact of this is gigantic and gives a huge headache, because normally the server save happens every 24 hours, and in an eventual crash, it could lose a lot in the houses.

About adding a function` saveHouse(...)`, I think that would solve it because its implementation would be simple and it makes sense to have a function like that inside tfs.

brunotonin:
it would be interesting to do the following test to simulate what would happen if you ran the save in a house that lost its owner.

```
addEvent(funtion(houseId)
    local house = House(houseId)
    house:save()
end, 2000, house:getId())

house:setOwner(0)
```

DSpeichert:
This is request for an anti-pattern. There are advantages but it would also make it easier to exploit any crash if only partial state is saved. Think about cloning items: item is moved to a house, house is saved but player is not (upon a crash) and you have item cloning.

I think this feature, which as pointed out should never actually be useful in production, has potential to do even more harm.

Kamenuvol:
With two players you can exploit crashes too. Player A gives Item to Player B and Player B logs off. Server crash and both players now have the Item.
Using @DSpeichert logic, it should save players and houses only on server save.

brunotonin:
> This is request for an anti-pattern. There are advantages but it would also make it easier to exploit any crash if only partial state is saved. Think about cloning items: item is moved to a house, house is saved but player is not (upon a crash) and you have item cloning.
> 
[...]

PR #3386, opened 2021-03-26
New method 'save' for House Class

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
* `house:save()`
This method saves the house in the database, there is more information about it in the Issue where this change is suggested

**Issues addressed:** #3378

### Notes
I already took care of testing it in two situations:
* Add items to the house and caused the server to crash and as expected these items were lost
* Add items in the house but this time I used the new save method for this house and then I made the server crash and the items were not lost they were successfully saved in the house

### DIFF
--- a/src/iomapserialize.cpp
+++ b/src/iomapserialize.cpp
@@ -23,6 +23,8 @@
 #include "game.h"
 #include "bed.h"
 
+#include <fmt/format.h>
+
 extern Game g_game;
 
 void IOMapSerialize::loadHouseItems(Map* map)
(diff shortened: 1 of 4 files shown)


ranisalt (approved):
Good job

MillhioreBT:
There is the possibility of cloning items in very specific circumstances if the scripter does not save the house together with the player, so I do not know what to do about it so that it is understood that the player should also be saved, who seems to move this function to Player.save?

this was a suggestion from @EPuncker is good really

MillhioreBT:
There is the possibility of cloning items in very specific circumstances if the scripter does not save the house together with the player, so I do not know what to do about it so that it is understood that the player should also be saved, who seems to move this function to Player.save?

this was a suggestion from @EPuncker is good really

DSpeichert (approved):

PR #3387, opened 2021-03-26
Fix vcpkg workflow's cmake args

marmichalski:
Vcpkg cmake configuration was wrong, therfore the server was always compiled with LuaJIT as it was available and preferred. This PR fixes the use of `USE_LUAJIT `.

Windows fails with LuaJIT off. Fixing this is not in scope of this PR, IMHO.

https://github.com/otland/forgottenserver/actions/runs/689487844

### DIFF
--- a/.github/workflows/build-vcpkg.yml
+++ b/.github/workflows/build-vcpkg.yml
@@ -109,8 +109,8 @@ jobs:
         with:
           useVcpkgToolchainFile: true
           buildDirectory: ${{ runner.workspace }}/build
-          cmakeBuildType: ${{ matrix.buildtype }}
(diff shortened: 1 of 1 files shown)


DSpeichert (changes requested):
I don't see how this fixed `USE_LUAJIT`. I do see the extra `CMAKE_BUILD_TYPE`, although I can't tell now if we were simply not actually building separate Debug/Release builds properly.

marmichalski:
See: https://github.com/otland/forgottenserver/runs/2370092436?check_suite_focus=true#step:10:50

This is supposed to be built without luajit, but it's not.

DSpeichert (approved):

PR #3391, opened 2021-04-01
Respect FLAG_NOLIMIT when adding items to store inbox

nekiro:
Currently adding items into store inbox does not respect passing FLAG_NOLIMIT and this is obviously wrong, this pr fixes it.

### DIFF
--- a/src/storeinbox.cpp
+++ b/src/storeinbox.cpp
@@ -23,7 +23,7 @@
 
 StoreInbox::StoreInbox(uint16_t type) : Container(type, 20, true, true) {}
 
-ReturnValue StoreInbox::queryAdd(int32_t, const Thing& thing, uint32_t, uint32_t, Creature*) const
+ReturnValue StoreInbox::queryAdd(int32_t, const Thing& thing, uint32_t, uint32_t flags, Creature*) const
 {
 	const Item* item = thing.getItem();
 	if (!item) {
--- a/src/storeinbox.cpp
+++ b/src/storeinbox.cpp
@@ -38,13 +38,15 @@ ReturnValue StoreInbox::queryAdd(int32_t, const Thing& thing, uint32_t, uint32_t
 		return RETURNVALUE_CANNOTPICKUP;
 	}
 
-	if (!item->isStoreItem()) {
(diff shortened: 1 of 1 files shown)


MillhioreBT:
@nekiro where does the idea of ​​restricting moving containers with items inside come from?
I have no idea of ​​the behavior of this tray in Tibia RL, but I would love to create containers with rewards and send them to this tray, since this tray ignores the capacity it is perfect for this purpose

nekiro:
> @nekiro where does the idea of ​​restricting moving containers with items inside come from?
> I have no idea of ​​the behavior of this tray in Tibia RL, but I would love to create containers with rewards and send them to this storeInbox, since this storeInbox ignores the capacity it is perfect for this purpose
> 
> Regarding what I comment on the forum, that I cannot move item from this storeInbox, it still happens to me, I will try to verify my sources rigorously to see if there is something different in my custom sources

This is ciptibia behaviour.

MillhioreBT:
> > @nekiro where does the idea of ​​restricting moving containers with items inside come from?
> > I have no idea of ​​the behavior of this tray in Tibia RL, but I would love to create containers with rewards and send them to this storeInbox, since this storeInbox ignores the capacity it is perfect for this purpose
> > Regarding what I comment on the forum, that I cannot move item from this storeInbox, it still happens to me, I will try to verify my sources rigorously to see if there is something different in my custom sources
> 
> This is ciptibia behaviour.

Any official source?
is that it seems impossible to me to know this if we cannot execute code in tibia RL, we know that we cannot move articles into this container for obvious reasons, but extrapolate that it is impossible to move containers with articles just because Tibia does not sell containers filled with articles It does not seem like a correct way to predict that this behavior is like this

Anyway, it is not something important for everyone, I am already complaining a lot and the truth is that you did a great job

nekiro:
https://tibia.fandom.com/wiki/Your_Store_Inbox

MillhioreBT:
You only need to add the name of the parameter of the function, in this case `flags`
https://github.com/otland/forgottenserver/blob/f4cb840d1107eea53af3cf943e49245e722379c5/src/storeinbox.cpp#L26

MillhioreBT (approved):

EPuncker (approved):

PR #3392, opened 2021-04-04
Change int to uint16_t in house class for consistency

nekiro:
As we use precise data types in all parts of the sources. I think we should replace this int with corresponding u16 or even u8 too for consistency.


ranisalt (approved):
It can be `int8_t` or `uint8_t` since it never grows > 100, indeed.

DSpeichert (approved):

PR #3285, opened 2021-01-24
Use STL algorithms instead of loop in vocations.cpp

ranisalt:
With C++14, we will replace these `decltype(vocationsMap)::value_type` with `auto`

PR #3393, opened 2021-04-04
Vocation.getVocationId Use lowercase names too

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
In previous versions you could write the names of the vocations in lowercase, at this time the name must exactly match the name, which is not very comfortable.

### Example
`spell:vocation("sorcerer")` -- Error
`spell:vocation("Sorcerer)` -- Correct because in the xml file it is written that way

Now with these changes the two options are valid

it is worth mentioning that previously this already worked, but with the changes of PR #3285 it is now no longer possible

### DIFF
--- a/src/vocation.cpp
+++ b/src/vocation.cpp
@@ -161,8 +161,10 @@ Vocation* Vocations::getVocation(uint16_t id)
 
 int32_t Vocations::getVocationId(const std::string& name) const
 {
-	auto it = std::find_if(vocationsMap.begin(), vocationsMap.end(), [&name](decltype(vocationsMap)::value_type it) {
-		return strcasecmp(it.second.name.c_str(), name.c_str()) == 0;
+	auto it = std::find_if(vocationsMap.begin(), vocationsMap.end(), [&name](auto it) {
+		return name.size() == it.second.name.size() && std::equal(name.begin(), name.end(), it.second.name.begin(), [](char a, char b) {
+			return std::tolower(a) == std::tolower(b);
+			});
 	});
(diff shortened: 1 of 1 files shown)


ranisalt (changes requested):
Please avoid using raw C functions, use [this pattern](https://github.com/otland/forgottenserver/blob/f2d74c67959308636a7a725de3bb5af9492b9d61/src/game.cpp#L440-L442) instead

src/vocation.cpp
```diff
@@ -161,8 +161,10 @@ Vocation* Vocations::getVocation(uint16_t id)
 
 int32_t Vocations::getVocationId(const std::string& name) const
 {
-	auto it = std::find_if(vocationsMap.begin(), vocationsMap.end(), [&name](decltype(vocationsMap)::value_type it) {
-		return strcasecmp(it.second.name.c_str(), name.c_str()) == 0;
+	auto it = std::find_if(vocationsMap.begin(), vocationsMap.end(), [&name](auto it) {
+		return name.size() == it.second.name.size() && std::equal(name.begin(), name.end(), it.second.name.begin(), [](char a, char b) {
+			return std::tolower(a) == std::tolower(b);
+			});
```
ranisalt, line 167: I believe the indentation is wrong here, it should be 1 level to the left

ranisalt (changes requested):

DSpeichert (approved):

ranisalt (approved):
Cool

nekiro (approved):

Issue #3397, opened 2021-04-05
"defense name="outfit"" only works after you do "/reload monsters"


andremiles1:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
**Any** monster which use spell outfit change on itself (under the _defenses_ tag) only start doing so after getting reloaded.
An excellent example to observe this bug is with the monster "Dark Apprentice", which transforms itself on a frog defensively, and attacks by transforming others in a broom.

1. Create /m dark apprentice, or go to any map respawn of this monster.
2. Do a utevo res "dark apprentice and start attacking the real one. They never defensively transform themselves on a frog, but transforms others/itself on a broom when attacking. So the issue is only on the _defenses_ tag.
3. After a while, just do a /reload monsters
4. They will start transforming themselves on a frog after the reload.

### Expected behaviour
Monsters should change its outfit passively/defensively when specified to do so.

### Actual behaviour
They do, but only after a forced reload on the monsters files.

EPuncker:
yeah, I've just tested and this is indeed a bug:
```<defenses armor="15" defense="15">
		<defense name="healing" interval="2000" chance="15" min="30" max="40">
			<attribute key="areaEffect" value="greenshimmer" />
		</defense>
		<defense name="outfit" interval="2000" chance="5" target="1" monster="green frog" duration="3000" />
	</defenses>```

the healing works fine while the outfit is only triggered after the /reload monsters

MillhioreBT:
The probability of them transforming is ridiculously low, even this monster may never transform perfectly within 5 minutes or more
If you increase the probability to 100% you will notice that it is not necessary to reload, nor to place the monsters in specific situations

they transform perfectly when they are thinking
I will continue testing and investigating a little more

MillhioreBT:
Update: problem only occurs when `forceMonsterTypesOnLoad` is true
do your tests and confirm this

EPuncker:
> Update: problem only occurs when `forceMonsterTypesOnLoad` is true
> do your tests and confirm this

nice catch, exactly, the problem only occurs when `forceMonsterTypesOnLoad` is true

EPuncker:
[...]

PR #3404, opened 2021-04-07
Correct loadMonsterType from Xml

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
We must save in advance a physical reference of the xml file of the monsters, and then in another loop iterate over all the references to load each monster file, in this way we make sure that if a monsterType has not yet been loaded we can load the monsterType with its reference already saved previously

I'm not 100% sure this is the right way to do it, so if you have any suggestions, just give me your suggestions.

**Issues addressed:** #3397
This case only occurs when we want to know information about a type of monster that has not loaded yet.
In the example of the green frog, it was obvious that the `Dark Apprentice` could not know anything about `Green Frog` because it had not loaded yet, and since we are forcing the direct loading of the monsters then we evaded the saving of the reference of the xml file

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -77,12 +77,7 @@ bool Monsters::loadFromXml(bool reloading /*= false*/)
 	bool forceLoad = g_config.getBoolean(ConfigManager::FORCE_MONSTERTYPE_LOAD);
 
 	for (auto it : unloadedMonsters) {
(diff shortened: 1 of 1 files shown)


EvilHero90 (approved):
Looks good to me

EPuncker (approved):

PR #3412, opened 2021-04-16
The storeitems must have a weight in your inventory

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
I just realized that in Tibia RL, the objects contained in the StoreInbox* also have weight in the player's inventory.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -438,6 +438,11 @@ void Player::updateInventoryWeight()
 			inventoryWeight += item->getWeight();
 		}
 	}
+
+	StoreInbox* storeInbox = getStoreInbox();
+	if (storeInbox) {
+		inventoryWeight += storeInbox->getWeight();
+	}
 }
 
 void Player::addSkillAdvance(skills_t skill, uint64_t count)
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -439,8 +439,7 @@ void Player::updateInventoryWeight()
 		}
 	}
 
-	StoreInbox* storeInbox = getStoreInbox();
-	if (storeInbox) {
+	if (StoreInbox* storeInbox = getStoreInbox()) {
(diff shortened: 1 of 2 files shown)


DSpeichert (dismissed):

EPuncker (dismissed):

src/player.cpp
```diff
@@ -438,6 +438,11 @@ void Player::updateInventoryWeight()
 			inventoryWeight += item->getWeight();
 		}
 	}
+
+	StoreInbox* storeInbox = getStoreInbox();
+	if (storeInbox) {
```
nekiro, line 443: if (StoreInbox* storeInbox = getStoreInbox()) {

src/storeinbox.cpp
```diff
@@ -54,15 +54,31 @@ ReturnValue StoreInbox::queryAdd(int32_t, const Thing& thing, uint32_t, uint32_t
 
 void StoreInbox::postAddNotification(Thing* thing, const Cylinder* oldParent, int32_t index, cylinderlink_t)
 {
-	if (parent != nullptr) {
-		parent->postAddNotification(thing, oldParent, index, LINK_PARENT);
+	Cylinder* topParent = getTopParent();
```
nekiro, line 57: what's the point of these changes?

MillhioreBT: Since we are 100% sure that the parent of this container will always be the player, then we can reduce the code to how it was before and only change the cylinder link, I think this should be enough, I'm going to try it

src/player.cpp
MillhioreBT: of course, how could I forget to compact this into one line ;)

nekiro (approved):
Looks like I missed this too.

LGTM

EPuncker (approved):

DSpeichert (approved):

PR #3417, opened 2021-04-20
Added limit of items per tile

EPuncker:
### Pull Request Prelude

### Changes Proposed
Added configurable limit of items per tile exactly as it was on 0.x series!

If you have any change suggestion, please use the _**inline code suggestion**_ feature for faster changes


nekiro:
This is easily doable, just with onMoveItem event in lua

DSpeichert (commented):
LGTM, although I'm not sure if it should be in Lua instead or not. Looks quite elegant here. Also - do we need it?

MillhioreBT:
This could be in a lua script, for example:

**data/scripts/eventcallbacks/tile_limits.lua**
```lua
local tileLimit = 0
local protectionTileLimit = 0
local houseTileLimit = 0

local ec = EventCallback

function ec.onMoveItem(player, item, count, fromPosition, toPosition, fromCylinder, toCylinder)
	local tile = toPosition:getTile()
	if tile then
		local itemLimit = tile:getHouse() and houseTileLimit or tile:hasFlag(TILESTATE_PROTECTIONZONE) and protectionTileLimit or tileLimit
		if itemLimit > 0 and tile:getThingCount() > itemLimit and item:getType():getType() ~= ITEM_TYPE_MAGICFIELD then
			player:sendCancelMessage("You can not add more items on this tile.")
			return false
		end
	end
	return true
end

ec:register(-1)
```

However your code in the sources is correct, I will check it thoroughly later sure ;)

src/tools.cpp
```diff
@@ -1228,6 +1228,9 @@ const char* getReturnMessage(ReturnValue value)
 		case RETURNVALUE_ITEMCANNOTBEMOVEDTHERE:
 			return "This item cannot be moved there.";
 
+		case RETURNVALUE_YOUCANNOTADDMOREITEMSONTHISTILE:
+			return "You can not add more items on this tile.";
```
idontreallywolf, line 1232:
```suggestion
			return "You cannot add more items on this tile.";
```

nekiro: why?

EPuncker:
@nekiro @DSpeichert we don't need it, but it wouldn't harm as it is only a few lines, and yes there were a person asking for it in the forum, also after taking a look at 0.4 config.lua it was like the only "useful" setting they had...

and yes, I was not sure if C++ or Lua, but after thinking Lua is way simpler and one could even remove it totally, so I'm doing a new commit with Lua change

EPuncker:
okay, better thinking, perhaps we should just release it in the forum instead

PR #3421, opened 2021-04-21
Rename cannotpushallcreatures to canpushallcreatures

nekiro:
Current PlayerFlag_CanPushAllCreatures groups.xml string is misleading, it should be the opposite way.

### DIFF
--- a/data/XML/groups.xml
+++ b/data/XML/groups.xml
@@ -24,7 +24,7 @@
 			<flag cannotbebanned="1" />
 			<flag cannotbepushed="1" />
 			<flag hasinfinitecapacity="1" />
-			<flag cannotpushallcreatures="1" />
+			<flag canpushallcreatures="1" />
 			<flag cantalkredprivate="1" />
 			<flag cantalkredchannel="1" />
 			<flag talkorangehelpchannel="1" />
@@ -65,7 +65,7 @@
 			<flag cannotbebanned="1" />
 			<flag cannotbepushed="1" />
 			<flag hasinfinitecapacity="1" />
-			<flag cannotpushallcreatures="1" />
+			<flag canpushallcreatures="1" />
 			<flag cantalkredprivate="1" />
 			<flag cantalkredchannel="1" />
 			<flag talkorangehelpchannel="1" />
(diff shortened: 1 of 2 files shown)


EPuncker:
missing groups.xml file change

MillhioreBT:
Changes are also required here
https://github.com/otland/forgottenserver/blob/670b0ca111cbc72440b60e43c627716de88aee9f/data/XML/groups.xml#L27
https://github.com/otland/forgottenserver/blob/670b0ca111cbc72440b60e43c627716de88aee9f/data/XML/groups.xml#L68

DSpeichert:
Is this in use? If so, it would be backwards incompatible.

nekiro:
> Is this in use? If so, it would be backwards incompatible.

Just the string is renamed not the actual flag, this bug appeared when I added new groups parsing

EPuncker (approved):

MillhioreBT (approved):

DSpeichert:
I see the flag being renamed in this PR, is it reverting to the name that was used before?

MillhioreBT:
> I see the flag being renamed in this PR, is it reverting to the name that was used before?

What happens is that it is a name that does not mean what the flag actually does:
image
the flag is understood as a flag that allows you to move all creatures, while its name in xml, implies that it is a flag that does not allow you to move all creatures

nekiro:
> I see the flag being renamed in this PR, is it reverting to the name that was used before?

No name was used before as I said before this was added by me and I mispelled this flag name.

DSpeichert:
> No name was used before as I said, this was added by me and I mispelled this flag name.

Okay, so here's what I was asking: since i can see you are changing the line in XML, it existed before. Looking through git blame, I found commit 7f8002e14 (#2546) which is probably where you're saying you added it. However, it was 2 years ago. So for two years, we had this other, confusing name for the flag. Is that right?

If so, that does open up the discussion whether anyone over 2 years got used to using that name you're now changing.

nekiro:
> > No name was used before as I said, this was added by me and I mispelled this flag name.
> 
> Okay, so here's what I was asking: since i can see you are changing the line in XML, it existed before. Looking through git blame, I found commit [7f8002e](https://github.com/otland/forgottenserver/commit/7f8002e1479b75e6be1349cec80089e192118072) (#2546) which is probably where you're saying you added it. However, it was 2 years ago. So for two years, we had this other, confusing name for the flag. Is that right?
> 
> If so, that does open up the discussion whether anyone over 2 years got used to using that name you're now changing.

Wow, it was 2 years ago. I didn't realize that. How weird I thought it was recent change wtf...

EPuncker:
I see no reason for all this hassle, I'm in favor of this PR, its not like people wouldn't know how to fix this on their end

DSpeichert (approved):

Issue #3346, opened 2021-02-24
Possible memory leak - depots not getting deleted from memory at player deletion


djseban:
Hey guys,

When working on one feature, I was poking around the code of TFS and in Player class destructor I spotted one thing:
```
Player::~Player()
{
	for (Item* item : inventory) {
		if (item) {
			item->setParent(nullptr);
			item->decrementReferenceCounter();
		}
	}

	for (const auto& it : depotLockerMap) {
		it.second->removeInbox(inbox);
		it.second->decrementReferenceCounter();
	}

	inbox->decrementReferenceCounter();

	storeInbox->setParent(nullptr);
	storeInbox->decrementReferenceCounter();

	setWriteItem(nullptr);
	setEditHouse(nullptr);
}
```

While in depotLockerMap the inbox is removed, the depot chests of player (depotChests std::map) remain untouched. Theoretically depot chests should get deleted due to 
`it.second->decrementReferenceCounter();`
being called, but quick modification of decrementReferenceCounter in item.h
```
void decrementReferenceCounter() {
   if (getID() == ITEM_DEPOT) {
      std::cout << "decrement depot " << referenceCounter <<std::endl;
   }
   if (--referenceCounter == 0) {
      delete this;
   }
}
```
reveals that in fact no depot chest is getting deleted (we don't get any cout in terminal). Moreover, by accessing previously saved pointer to some Item that was in the depot of deleted player, we are able to retrieve it. The easiest way I found to reproduce this is following:
1. Place one item of choice in depot chest. The item has to remain on first slot of depot chest.
2. Execute following lua code:
```
local dpId = 3
local item = player:getDepotChest(dpId):getItem(0) 
local pos = player:getPosition() 
player:remove() 
addEvent(function() Tile(pos):addItemEx(item:clone()) end, 5000)
```
3. Lua code will throw you out of a game. Log in into the same or the other character to witness item being created on the tile you logged out with first character.
Expected behavior of the TFS is to crash - due to accessing the item that doesn't really exist in memory, which it does, if we try the same thing, but with the inbox of a player:
```
local item = player:getInbox():getItem(0) 
local pos = player:getPosition() 
player:remove() 
addEvent(function() Tile(pos):addItemEx(item:clone()) end, 5000)
```

[...]

PR #3423, opened 2021-04-22
Use shared_ptr for depot locker and depot chests

nekiro:
This should fix memory leak pointed out in #3346 
I didn't test this code, so testing is welcome!

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -67,7 +67,6 @@ Player::~Player()
 
 	for (const auto& it : depotLockerMap) {
 		it.second->removeInbox(inbox);
-		it.second->decrementReferenceCounter();
 	}
 
 	inbox->decrementReferenceCounter();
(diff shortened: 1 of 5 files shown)


MillhioreBT:
Simply perfect and a way forward to change all pointers to smart

src/player.cpp
```diff
@@ -782,35 +781,32 @@ DepotChest* Player::getDepotChest(uint32_t depotId, bool autoCreate)
 {
 	auto it = depotChests.find(depotId);
 	if (it != depotChests.end()) {
-		return it->second;
+		return it->second.get();
```
ranisalt, line 784: What about returning a smart pointer from this function? Would it be too much to change?

MillhioreBT: Would this make the path in the stack cleaner?

nekiro: No idea. I didn't check.

MillhioreBT: I'll do the test and let you know the result

MillhioreBT: I have done several tests and it seems that as it is now it is very good, since in case of directly handling `DepotChest_ptr` as a return value, it seems to require adding many `.get()` in many parts of the code

The only way would be to create some kind of operator function that does the `.get` transformation directly with the `= ` operator.

EvilHero90: Should be adressed in another PR I'd say, this is a bug fix so we shouldn't start fiddling around in to much corners

EvilHero90 (approved):
Code wise this looks fine to me

EPuncker (approved):

ranisalt (approved):

ranisalt:
Why is the pipeline failing though?

nekiro:
> Why is the pipeline failing though?

It's some vcpkg thing that fails, code compiled fine cause I tested the compilation at least xD

PR #3427, opened 2021-04-24
some minor fixes

emil92b:


DSpeichert:
You should split those changes into separate PRs, explaining exactly why you're proposing each change.

nekiro:
Looks like a dead PR to me, if you think otherwise, comment or reopen.

PR #3430, opened 2021-04-24
Change ostringstream to fmt format

ramon-bernardo:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
- Change ostringstream to fmt format.

#### Use fmt format:
- [x] for [house door description.](https://github.com/otland/forgottenserver/pull/3430/commits/73d26c9712cf8d78c67c43e5c27b735562537fb3)
- [x] for [messages](https://github.com/otland/forgottenserver/pull/3430/commits/6078559dafb58048ecdd038ec88cc687f9207600).
- [x] for [declare file directories](https://github.com/otland/forgottenserver/pull/3430/commits/cbaa5425896bc314f9b046906140e98c2b09ac0b).
- [x] for [network messages](https://github.com/otland/forgottenserver/pull/3430/commits/af257667b2587cc1339c447b81b1f4e152a1c7e9).
- [x] for [querys](https://github.com/otland/forgottenserver/pull/3430/commits/4f3a3e03663d5338593f5749c78dc2372d39b25c).   (savePlayer() [query](https://github.com/otland/forgottenserver/blob/08218895164f3142d135774a3787ddc1b7ef0c78/src/iologindata.cpp#L697-L787)  kept ostringstream.)

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #3432, opened 2021-04-25
Fix use-after-move and improve code style on actions

ranisalt:
### Pull Request Prelude

### Changes Proposed

- Fix inverted logical conditions
- Avoid usage of private static members
- Remove redundant else statement
- Use fmt for formatting hotkey messages

**Issues addressed:** 
None


src/actions.cpp
```diff
@@ -218,17 +219,17 @@ bool Actions::registerLuaEvent(Action* event)
...
-				auto result = useItemMap.emplace(*i, std::move(*action));
-				if (!result.second) {
-					std::cout << "[Warning - Actions::registerLuaEvent] Duplicate registered item with id: " << *i << " in range from id: " << v.front() << ", to id: " << v.at(v.size() - 1) << std::endl;
-					continue;
-				}
+		}
+
+		auto v = action->getItemIdRange();
+		for (auto i = v.begin(); i != v.end(); i++) {
+			auto result = useItemMap.emplace(*i, std::move(*action));
```
ranisalt, line 226: By the way, isn't this using `*action` after moving? I just noticed it lol

ranisalt:
OK I noticed this code had a lot of reinvention of wheel such as `v.at(vector.size() - 1)` instead of `v.back()` or iterating over indexes when a range was available and I decided to slip that in. Also fixed several use-after-move instances

src/actions.cpp
```diff
@@ -26,6 +26,7 @@
 #include "game.h"
 #include "pugicast.h"
 #include "spells.h"
+#include <fmt/format.h>
```
ramon-bernardo, line 29: Why use this in each file instead [this](https://github.com/otland/forgottenserver/pull/3430/commits/5a15e2021e4b932144acf848e283ae7e963231f1#diff-2aece0739543e6f81501e211e382a9e264966cdddecea19ed6182e61c2c0e4c5R46)?

ranisalt: More includes = more compile time

nekiro (dismissed):

src/actions.cpp
```diff
@@ -210,67 +211,37 @@ bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node)
 bool Actions::registerLuaEvent(Action* event)
 {
 	Action_ptr action{ event };
-	if (action->getItemIdRange().size() > 0) {
-		if (action->getItemIdRange().size() == 1) {
-			auto id = action->getItemIdRange().front();
-			auto result = useItemMap.emplace(id, std::move(*action));
+	if (!action->getItemIdRange().empty()) {
+		const auto& v = action->getItemIdRange();
+		for (auto id : v) {
```
nekiro, line 216: what if we used better names instead of "v"?

ranisalt: Absolutely. Renamed to `range`

nekiro (dismissed):
Perfection

nekiro (approved):

PR #3441, opened 2021-05-04
Add premiumToBuyHouse boolean on buyhouse command

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Add premiumToBuyHouse boolean on buyhouse command. (buyhouse.lua)
- Just change the variable to 'only premium' buy house. 

### DIFF
--- a/data/talkactions/scripts/buyhouse.lua
+++ b/data/talkactions/scripts/buyhouse.lua
@@ -1,10 +1,12 @@
+local premiumToBuyHouse = true
+
 function onSay(player, words, param)
 	local housePrice = configManager.getNumber(configKeys.HOUSE_PRICE)
 	if housePrice == -1 then
 		return true
 	end
 
-	if not player:isPremium() then
+	if not player:isPremium() and premiumToBuyHouse then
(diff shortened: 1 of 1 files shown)


EPuncker:
can you also add **levelToBuyHouse** and perhaps a setting to limit the amount of houses owned by player/account? (this last one is optional)

EPuncker (dismissed):

data/talkactions/scripts/buyhouse.lua
```diff
@@ -1,10 +1,20 @@
...
 		return true
 	end
 
-	if not player:isPremium() then
+	if player:getLevel() < config.level then
+		player:sendCancelMessage("You need level " .. config.level .. " or higher to buy a house.")
+		return false
+	end
+
+	if not player:isPremium() and config.onlyPremium then
```
MillhioreBT, line 17: This is fine, but if you first check `config.onlyPremium` it would be the correct way to write it, since once it is false, `isPremium` will never be called because it will never be necessary

EPuncker (approved):

PR #3442, opened 2021-05-04
Add spell max range

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Add max range when throw spell

#### TibiaRL
- Player cannot throw spell "Exura sio", if target is not on screen

#3427 @emil92b

### DIFF
--- a/src/spells.cpp
+++ b/src/spells.cpp
@@ -1037,7 +1037,7 @@ bool InstantSpell::canThrowSpell(const Creature* creature, const Creature* targe
 	const Position& fromPos = creature->getPosition();
 	const Position& toPos = target->getPosition();
 	if (fromPos.z != toPos.z ||
-			(range == -1 && !g_game.canThrowObjectTo(fromPos, toPos, checkLineOfSight)) ||
+			(range == -1 && !g_game.canThrowObjectTo(fromPos, toPos, checkLineOfSight, 7, 5)) ||
 			(range != -1 && !g_game.canThrowObjectTo(fromPos, toPos, checkLineOfSight, range, range))) {
 		return false;
 	}


EPuncker (approved):

MillhioreBT:
The default range is 8, 6 why change to 7, 5? if 8, 6 are considered to be in the field of vision

ramon-bernardo:
> The default range is 8, 6 why change to 7, 5? if 8, 6 are considered to be in the field of vision

I tested it TibiaRl, can only use spell with player on screen.
  ```![image](https://user-images.githubusercontent.com/53269692/117215547-bb2ea380-add4-11eb-9aff-2caca0972c81.png)
![image](https://user-images.githubusercontent.com/53269692/117215570-c4b80b80-add4-11eb-8c16-16d684425b5e.png)
  ```

MillhioreBT:
In that case, then okay Although I do not know how this applies for users who use OTC with the modified viewport

Znote:
We should make our hardcoded viewport values configurable, and this should be calculated on the viewport values instead of being hardcoded.

ramon-bernardo:
> We should make our hardcoded viewport values configurable, and this should be calculated on the viewport values instead of being 7, 5.

I'll try, good idea!

nekiro:
> > We should make our hardcoded viewport values configurable, and this should be calculated on the viewport values instead of being 7, 5.
> 
> I'll try, good idea!

I would say it's a thing for another pull request, let it be for this one.

ramon-bernardo:
Links with viewport; [canUseFar()](https://github.com/otland/forgottenserver/blob/aea764cb8d8ea5da41e050d302b99033be1008cf/src/actions.cpp#L311), [playerUseWithCreature()](https://github.com/otland/forgottenserver/blob/aea764cb8d8ea5da41e050d302b99033be1008cf/src/game.cpp#L2201)

Znote (approved):

ranisalt:
I believe when #3456 is merged, you can use the constants here too instead of 7 and 5

DSpeichert (approved):

PR #3448, opened 2021-05-07
Fixed overwriting weapons with Revscript

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
With these changes you will be able to overwrite weapons with Revscripts, since currently when doing so the server suffers from a crash for referencing an already released weapon

I have tried these changes and they work great, but if you have any other way to approach the problem feel free to say something about it or even propose other changes.

Credits: @SaiyansKing

### Steps to follow to check if this solves the problem
Once the changes are applied you can try creating a weapon like in Revscript like this:
`data/scripts/snowball.lua`
```lua
local weapon = Weapon(WEAPON_DISTANCE)
weapon:id(2111)
weapon:action("removecount")
weapon:register()
```

Remember not to remove this weapon from `weapons.xml`
```xml
<distance id="2111" action="removecount" /> <!-- snowball -->
```

Once the server is starting it should not cause a crash!
Finally with this we show that the changes allow us to even create the same weapon in xml and revscript without any problem

**Issues addressed:** Fixes possible Weapon null dereference
**Backward compatibility:** Yes
**Compatibility with XML:** Yes

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -16041,8 +16041,13 @@ int LuaScriptInterface::luaWeaponAction(lua_State* L)
 int LuaScriptInterface::luaWeaponRegister(lua_State* L)
 {
 	// weapon:register()
-	Weapon* weapon = getUserdata<Weapon>(L, 1);
-	if (weapon) {
+	Weapon** weaponPtr = getRawUserdata<Weapon>(L, 1);
+	if (!weaponPtr) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	if (auto* weapon = *weaponPtr) {
 		if (weapon->weaponType == WEAPON_DISTANCE || weapon->weaponType == WEAPON_AMMO) {
 			weapon = getUserdata<WeaponDistance>(L, 1);
 		} else if (weapon->weaponType == WEAPON_WAND) {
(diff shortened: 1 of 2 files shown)


EPuncker (approved):

src/luascript.cpp
```diff
@@ -16064,6 +16069,7 @@ int LuaScriptInterface::luaWeaponRegister(lua_State* L)
 
 		weapon->configureWeapon(it);
 		pushBoolean(L, g_weapons->registerLuaEvent(weapon));
+		*weaponPtr = nullptr; // Remove luascript reference
```
ranisalt, line 16072: Doesn't this leak the valued pointed to?

Znote (approved):
Seems to work great, I'm able to edit weapons on the fly and reload between attacks and it just works fine without any crashing. 

image

src/luascript.cpp
MillhioreBT: Yeah, I wonder if it is also possible to do the same in this way:
```cpp
weapon = nullptr;
// instead
*weaponPtr = nullptr;
```
because we have already pointed to this pointer in the variable `auto weapon`

EvilHero90 (approved):
I'll approve it for now because it's a solution which atleast doesn't crash the server, but this should be further investigated and optimized at some point.

PR #3457, opened 2021-05-13
Add config to show/hidden player log in console

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Add bool `showPlayerLogInConsole` to show or hidden player log in console.

Log: `Player has logged in/out`.

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -93,6 +93,7 @@ yellAlwaysAllowPremium = false
 forceMonsterTypesOnLoad = true
 cleanProtectionZones = false
 luaItemDesc = false
+showPlayerLogInConsole = true
 
 -- VIP
 vipFreeLimit = 20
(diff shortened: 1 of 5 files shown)


MillhioreBT:
I prefer this in LUA ;)

DSpeichert:
Is there actual reason why not seeing those actions in the log is so badly desired?

Znote:
> Is there actual reason why not seeing those actions in the log is so badly desired?

This behavior can easily be done in Lua I think using onLogin and onLogout, which allows for easier modification (if you etc also want to display a players session duration, level gain etc). So you could disable this in config.lua and write your own login/logout print behavior. Currently this would end up with double lines? (print from source, and print from Lua). 

Since this can be done in Lua, I'm actually for removing it entirely from source. But being able to just disable it from source works as well. 

Not seeing the logs at all, might help reveal error messages on high density servers with lots of players. Especially if a server is not broadcasting the console to a file.

DSpeichert (approved):

PR #3461, opened 2021-05-15
Protocol lua API

EvilHero90:
This implements a new Protocol service, which gives us a bridge to interact with the server directly with an outside API of your preference with lua code.

**_What does that mean do you think?_**
Basicly we can now send lua code directly from an outside API into the server, a good example would be that we have an administration page on our AAC's where we insert lua code and then send it through the API to the server and it immediatly gets executed, this would make shops etc. a lot more efficient and easier to handle, or you can even administrate your server from your website without the need to touch any kind of files on your server.

**_Why did I chose protocol over http server?_**
Well... I thought why should we implement an entire http server when our protocol can easily handle all this already?

_**What are your final plans with this?**_
This actually opens us doors to quite a few new options, if we ever fully convert everything to revscriptsys then we are able to deploy servers easily without even the need to have any lua file on your hdd, we can easily fetch them directly from any website we want such as github (github raw's to be precise) or otland (if we add support for it)
I know this kinda sounds complex but it would also make it a lot easier for people sharing their scripts to keep them updated for all people who use them as you'll always have the newest version then.

**_Do you provide a sample API?_**
@Znote wrote a basic php API for testing purposes, which we will extend on
https://github.com/EvilHero90/TFS-lua-API/tree/main

This is how the current structure looks like: (might still change while WIP)
```c++
/*
* List of Packets used for communication
* --------------------------------------------------------------------
* 
* Server ----> API (structure of packet)
* --------------------------------------------------------------------
* 100 => ping
* 101 => sending a callback message (string)
* 102 => sending lua error back to API (string)
* 103 => request to exchange lua code, if the API has any
* --------------------------------------------------------------------
* 
* API ----> Server (structure of packet)
* --------------------------------------------------------------------
* 100 => pong
* 101 => sending raw string lua code with immediate execution (string[name], string[data])
* 102 => save server
* 103 => clean server
* 104 => close server
* 105 => start raid (string)
* 200 => reload scripts
* --------------------------------------------------------------------
*/
```
Here is a screenshot from a testing API @Znote made for me (thanks btw)
luaapi1

This also supports sending error messages back to the API if there are any
errorresponse

I'm completly open for suggestions and such, for now this is WIP and probably a few things will change here and there but it's ready to be tested.


data/globalevents/scripts/serversave.lua
```diff
@@ -1,4 +1,4 @@
-local function ServerSave()
+function ServerSave()
```
nekiro, line 1: I doubt this change is related?

src/events.cpp
```diff
@@ -133,6 +133,12 @@ bool Events::load()
 			} else {
 				std::cout << "[Warning - Events::load] Unknown monster method: " << methodName << std::endl;
 			}
+		} else if (className == "Game") {
+			if (methodName == "onLuaApiResponse") {
+				info.MiscOnLuaApiResponse = scriptInterface.getEvent(methodName);
```
nekiro, line 138: Misc -> misc

src/events.h
```diff
@@ -65,6 +65,9 @@ class Events
 		// Monster
 		int32_t monsterOnDropLoot = -1;
 		int32_t monsterOnSpawn = -1;
+
+		// Misc
+		int32_t MiscOnLuaApiResponse = -1;
```
nekiro, line 70: Same as above, lowercase first letter

src/luascript.cpp
```diff
@@ -340,6 +340,42 @@ int32_t LuaScriptInterface::loadFile(const std::string& file, Npc* npc /* = null
 	return 0;
 }
 
+std::string LuaScriptInterface::loadString(const std::string& string, const std::string& fileName)
+{
+	//loads string as a chunk at stack top
+	int ret = luaL_loadstring(luaState, string.c_str());
+	if (ret != 0) {
+		lastLuaError = popString(luaState);
+		return lastLuaError;
```
nekiro, line 349: Shouldnt this use the reportError method and return boolean?

src/protocolluaapi.cpp
```diff
@@ -0,0 +1,131 @@
...
+		return;
+	}
+
+	auto recvbyte = msg.get<uint16_t>();
+	auto name = msg.getString();
+	auto data = msg.getString();
+	g_events->eventMiscOnLuaApiResponse(recvbyte, name, data);
+
+	switch (recvbyte) {
+		case 100: {
```
nekiro, line 71: Hex values?

src/protocolluaapi.cpp
```diff
@@ -0,0 +1,131 @@
...
+			g_dispatcher.addTask(createTask(std::bind(&ProtocolLuaApi::sendCallbackMessage, std::static_pointer_cast<ProtocolLuaApi>(shared_from_this()),
+				"transmission disconnected")));
+	}
+}
+
+// packet 100
+void ProtocolLuaApi::sendPing()
+{
+	setResponse(false);
+	auto output = OutputMessagePool::getOutputMessage();
```
nekiro, line 98: Isnt output pointer?

nekiro (changes requested):

data/globalevents/scripts/serversave.lua
EvilHero90: It is actually, I re use this function in `data/events/scripts/game.lua` so I didn't have to write it twice

src/events.cpp
EvilHero90: I completly forgot to convert this, thanks for pointing it out

src/events.h
EvilHero90: same as above

src/luascript.cpp
EvilHero90: I could use reportError here yea, I basicly tried to mimic the loadFile version where it was basicly made like this.
For the returning as bool, no that wont work because I'm pushing the value as a string on that we can send the error message to the API

src/protocolluaapi.cpp
EvilHero90: We can make it hex once the testing stage is through, not sure if we change the values at some point, or what do you think?

src/protocolluaapi.cpp
EvilHero90: nope ^^

EvilHero90:
Here is the basic php page @Znote made, if anyone wants to give it a shot.
https://github.com/EvilHero90/TFS-lua-API/tree/main

MillhioreBT:
And what about security regarding this API?
I say it has always been known that web pages that are not well protected can be easily hacked and therefore in that case someone could run Lua code in the engine and do what they want

It is just something that crossed my mind, but surely we will not need a sandbox to execute the code sent from this API

EvilHero90:
This is as secure as it can get, we already limit the connection to localhost or bind it to a specific ip, if someone by accident opens the port, it would still refuse the connection due to ip restriction.
Security wise it shouldn't be our problem to begin with, if someone gets hold of your website, then you are completly doomed anyway.

src/protocolluaapi.cpp
Erza: Client packets are parsed using hex values too, so for consistency's sake it should stay that way.

EvilHero90: As I mentioned, I'll change them to hex once they're 100% set in stone and wont get changed

config.lua.dist
```diff
@@ -20,10 +20,12 @@ expFromPlayersLevelRange = 75
 -- NOTE: maxPlayers set to 0 means no limit
 -- NOTE: allowWalkthrough is only applicable to players
 ip = "127.0.0.1"
+luaApiIp = "127.0.0.1"
 bindOnlyGlobalAddress = false
 loginProtocolPort = 7171
 gameProtocolPort = 7172
 statusProtocolPort = 7171
+luaApiProtocolPort = 7179
```
Erza, line 28: Why not continue the port range normally, so 7173? Kinda random to just jump from 7172 to 7179 as default value

DSpeichert:
Considering how even cipbia moved to standardized protocols like ProtoBuffers, I am very unconvinced it's a good idea to design a new protocol for this, even if it's so simple and pure as this one.

Znote:
> Considering how even cipbia moved to standardized protocols like ProtoBuffers, I am very unconvinced it's a good idea to design a new protocol for this, even if it's so simple and pure as this one.

Can you refer to a PR that uses protobuffers?
I find this API to be small, fresh, flexible and consistent with our current protocols, so what is not to love?

MillhioreBT:
I have been investigating how to work with ProtoBuffers and it seems to be a pretty good technology, although it is robust, it seems very redundant, I think I should investigate more and do real tests with TFS, to be able to comment on the potential that it can bring to TFS, but for now this new protocol looks very promising and simple

DSpeichert:
> Can you refer to a PR that uses protobuffers?
> I find this API to be small, fresh, flexible and consistent with our current protocols, so what is not to love?

What do you mean? There's been talks about that in other PRs and an attempt to use protobuffers for other serialization. Technically, for APIs there's a standard based around protobuffers called gRPC.

What's not to love? I personally don't like any part of this. While it's lighter than using a full-fledged existing solution, designing a custom raw/binary protocol requires much more than this to make it well defined. It's simply not necessary. I'm already confused by the documentation: what is "API", what is "server"?

How would you handle authentication? Encryption, when desired? HTTP would allow wrapping in a reverse proxy with mutual TLS auth or other known solutions, a binary TCP protocol isn't easy to wrap to add auth/encryption, which eventually someone is going to want (there's already tons of threads on the forums where the website and server run on different hosts in different places).

Why only consider "website"? There used to be a TFS Admin protocol and we can do better this time. Same with the status protocol, which originally had a few binary "commands" (e.g. list players) but then switched to XML. gRPC/ProtoBuffers didn't exist back then.

I'm not saying ProtoBuffers/gRPC is the only way, it's just an example that I think would work if there is a good way to integrate it with Lua in TFS. It is known to work and has support in many languages. HTTP wouldn't be bad either. A custom protocol these days is simply rarely a good choice because designing, documenting and implementing one from scratch is no small feat.

EvilHero90:
> 
> 
> > Can you refer to a PR that uses protobuffers?
> > I find this API to be small, fresh, flexible and consistent with our current protocols, so what is not to love?
> 
> What do you mean? There's been talks about that in other PRs and an attempt to use protobuffers for other serialization. Technically, for APIs there's a standard based around protobuffers called gRPC.
> 
> What's not to love? I personally don't like any part of this. While it's lighter than using a full-fledged existing solution, designing a custom raw/binary protocol requires much more than this to make it well defined. It's simply not necessary. I'm already confused by the documentation: what is "API", what is "server"?
>
Ok so you rather want to add an entire http server or gRPC handling, which we both don't use anyway and bloat tfs more up than it already is? we rely on protocol anyway so we should make use of it, instead of waiting another 10 years for someone to re write everything to gRPC.
> How would you handle authentication? Encryption, when desired? HTTP would allow wrapping in a reverse proxy with mutual TLS auth or other known solutions, a binary TCP protocol isn't easy to wrap to add auth/encryption, which eventually someone is going to want (there's already tons of threads on the forums where the website and server run on different hosts in different places).
>
This is already limited to either localhost or to bind a specific IP to be able to communicate, so there is no need for authentication.
Everything you mentioned here is part of the API anyway, if the API is poorly protected that's none of our concern.
> Why only consider "website"? There used to be a TFS Admin protocol and we can do better this time. Same with the status protocol, which originally had a few binary "commands" (e.g. list players) but then switched to XML. gRPC/ProtoBuffers didn't exist back then.
>
I've never stated anywhere that I only consider "websites" we just decided to quickly show of a simple API which can be easily included into any AAC, if you rather want to write an API in c# or any other language then everyone can feel free to do that.
> I'm not saying ProtoBuffers/gRPC is the only way, it's just an example that I think would work if there is a good way to integrate it with Lua in TFS. It is known to work and has support in many languages. HTTP wouldn't be bad either. A custom protocol these days is simply rarely a good choice because designing, documenting and implementing one from scratch is no small feat.
>
"designing, documenting and implementing" all of what you mentioned there is even more work for what you suggest, we seem to be forgetting that "tfs" is a game server and not an allround package which should include several things in one program.

config.lua.dist
EvilHero90: there was no specific reason for 7179 but there was a reason to not chose 7173 which was that quite some servers use cast systems and some of them rely on port 7173

UReddington:
I'm using this with a discord bot and it works like a charm, just an observation, it seems like the ip verification is broken, i've changed the ip on the config.lua to an external host and it wasn't working at all (i had to just remove the localhost check, with this it works as intended)

src/protocolluaapi.cpp
```diff
@@ -0,0 +1,131 @@
...
+* 104 => close server
+* 105 => start raid (string)
+* 200 => reload scripts
+* --------------------------------------------------------------------
+*/
+
+void ProtocolLuaApi::onRecvFirstMessage(NetworkMessage& msg)
+{
+	// we only allow connection from the ip set in config.lua "luaApiIp = "xx.xxx.x.xx"" or from localhost.
+	if (convertIPToString(getIP()) != g_config.getString(ConfigManager::LUA_API_IP) || convertIPToString(getIP()) != "127.0.0.1") {
```
DSpeichert, line 59:
```suggestion
	if (convertIPToString(getIP()) != g_config.getString(ConfigManager::LUA_API_IP) && convertIPToString(getIP()) != "127.0.0.1") {
```

DSpeichert:
Quick and simple is exactly what it is. It works, I'm sure (though I haven't tried myself).

Personally, I would prefer to add entire HTTP stack and it's been proposed in #2010. While the amount of initial "bloat" is larger indeed, I believe the flexibility and power of a standard HTTP interface will be worth it.

What I see here is a potential for blocking limitations (like the aforementioned authentication/encryption issues) that aren't easy to change later. Adding byte-encoded API functions (?) is very crude. This protocol that we keep calling "protocol" here is literally pushing bytes back and forth, leaving the rest to a developer. The developer is usually more of an end-user who would prefer to use something more common that is well known or well documented (such as HTTP) instead of figuring out a raw byte-based custom "protocol". Eventually, people will want authentication, or JSON encoding, or multiplexing, compression, streaming, etc. I think it's easier to adapt based on a standard.

Long story short, all that bloat that you don't have here might eventually be necessary down there later.
We may be waiting for 10 years for something good and that's exactly why we should carefully think before adding anything. There is no time to go back and redo.

MillhioreBT:
> I'm using this with a discord bot and it works like a charm, just an observation, it seems like the ip verification is broken, i've changed the ip on the config.lua to an external host and it wasn't working at all (i had to just remove the localhost check, with this it works as intended)

Very Nice! working perfectly
<details>
<summary> GIF test discord BOT </summary>

GIF 03-06-2021 04-58-07 p  m

</details>

Znote:
> Personally, I would prefer to add entire HTTP stack and it's been proposed in #2010. While the amount of initial "bloat" is larger indeed, I believe the flexibility and power of a standard HTTP interface will be worth it.

I have to agree, I am hyped for an HTTP stack API like this one. Perhaps @EvilHero90 could attempt to continue @djarek's work? 

Quote @djarek 
> There are significant differences between the old Beast API and the new one, so this needs a rewrite.

This along with radio silence for 2 years made the HTTP API a pipe-dream. I would prefer this one, but I'll take anything over nothing.

MillhioreBT:
> > Personally, I would prefer to add entire HTTP stack and it's been proposed in #2010. While the amount of initial "bloat" is larger indeed, I believe the flexibility and power of a standard HTTP interface will be worth it.
> 
> I have to agree, I am hyped for an HTTP stack API like this one. Perhaps @EvilHero90 could attempt to continue @djarek's work?
> 
> Quote @djarek
> 
> > There are significant differences between the old Beast API and the new one, so this needs a rewrite.
> 
> This along with radio silence for 2 years made the HTTP API a pipe-dream. I would prefer http api over this one, but I'll take anything over nothing.

If @djarek changes are more secure, I will support the idea, I would like to know more about it to help with this!

Codinablack:
> Following some recent feedback, i would suggest adding a config variable to prevent the direct code execution (config variable, just a thing that would kill the connection if they doesn't want that), as we have already a trust problem of what he owners can execute if they use third party tools

I agree with this, the server owner should have the right to choose if they want this enabled or not, if it gets added.

Zbizu:
> This is as secure as it can get, we already limit the connection to localhost or bind it to a specific ip, if someone by accident opens the port, it would still refuse the connection due to ip restriction.
> Security wise it shouldn't be our problem to begin with, if someone gets hold of your website, then you are completly doomed anyway.

What if the server gets attacked by a spoofed signal?

Zbizu:
imo this would be great if the packet was treated as text so the end user would be able to choose whether he wants to loadstring or just use the api a different way

EvilHero90:
> 
> 
> imo this would be great if the packet was treated as text (like opcodes right now) so the end user would be able to choose whether he wants to loadstring or just use the api in a different way
> 
> it would also be far safer than immediate execution and would also open some new possibilities like live information about ongoing events or new ways to interact with otclient/ot lists

this is already implemented now, we can basicly parse http requests, this way you can even add 3rd party software which connects through your api and give it a certain access to do stuff

src/protocolluaapi.cpp
```diff
@@ -0,0 +1,232 @@
...
+		// now we get the data as string
+		auto dataString = msg.getString();
+		// add everything to map
+		dataMap.emplace(headerString, dataString);
+	}
+
+	/* We are searching if we can even find a registered GlobalEvent with the name
+	*  This way we could push app names onward from the API to only execute the GlobalEvent which belongs to this app
+	*/
+	auto http = g_globalEvents->getEventMap(GLOBALEVENT_HTTPREQUEST);
```
nekiro, line 185: isnt this an unnecessary copy?

src/protocolluaapi.cpp
```diff
@@ -0,0 +1,232 @@
...
+	if (http.find(name) != http.end()) {
+		g_globalEvents->getEventMap(GLOBALEVENT_HTTPREQUEST).at(name).executeHttpRequest(name, dataMap);
```
nekiro, line 187: using at but not capturing the exception anyway, so why using at?

src/protocolluaapi.cpp
```diff
@@ -0,0 +1,232 @@
...
+			g_dispatcher.addTask(createTask(std::bind(&ProtocolLuaApi::sendCallbackMessage, std::static_pointer_cast<ProtocolLuaApi>(shared_from_this()),
+				appName, "pong was successful")));
+			return;
+		}
+		// This will execute whatever data it gets through the string directly as lua code via "dostring" function
+		case 101: {
+			// We use fileName, to let the lua interface know how this file is named, incase it throws an error so we get a name to it.
+			auto fileName = msg.getString();
+			auto data = msg.getString();
+			auto returnvalue = g_scripts->executeString(data, fileName);
```
nekiro, line 143: returnValue

nekiro (commented):
Not a fan of all the "auto" there in the code. I prefer to visible types when I look at variables, but that's my opinion.

src/protocolluaapi.cpp
EvilHero90: yea thanks, fixed in latest commit

src/protocolluaapi.cpp
EvilHero90: actually I just noticed that find holds the pointer already, that way I don't even need to use `at` in this case

src/protocolluaapi.cpp
EvilHero90: thanks

src/protocolluaapi.cpp
EvilHero90: ok nvm, I just noticed I've already tried that and build failed (windows doesn't throw any error tho)

EvilHero90:
We wanted to modernize code and `auto` was one of the things which we talked about that we should use as much as we can, only if it makes the code completly unreadable but I agree in lots of cases I would prefer a strict type aswell

src/protocolluaapi.cpp
```diff
@@ -182,9 +182,9 @@ void ProtocolLuaApi::parseHttpRequest(std::string& appName, std::string& name, N
 	/* We are searching if we can even find a registered GlobalEvent with the name
 	*  This way we could push app names onward from the API to only execute the GlobalEvent which belongs to this app
 	*/
-	auto http = g_globalEvents->getEventMap(GLOBALEVENT_HTTPREQUEST);
+	auto& http = g_globalEvents->getEventMap(GLOBALEVENT_HTTPREQUEST);
 	if (http.find(name) != http.end()) {
-		g_globalEvents->getEventMap(GLOBALEVENT_HTTPREQUEST).at(name).executeHttpRequest(name, dataMap);
```
ranisalt: Don't call `find` twice, reuse the iterator instead.

src/protocolluaapi.cpp
nekiro: const auto& ?

src/protocolluaapi.cpp
EvilHero90: I guess you probably commented the moment I've added the commit, as this was already resolved ^^

src/protocolluaapi.cpp
EvilHero90: nope throws an error aswell, I'll take a closer look soon

PR #3465, opened 2021-05-18
configurable data directory

rookgaard:
### Pull Request Prelude

### Changes Proposed

Have you ever had thought you wanted to name each datapack for each server you have, while having the same engine? Now, using idea from 0.X, you can!

Changes I proposed replaces all occurences of hardcoded "data/" to string loaded from config file with default value as "data/", so from perspective of regular TFS user nothing will change, but give opportunity for those who (like me), wants to rename "data" directory.

### DIFF
--- a/src/baseevents.cpp
+++ b/src/baseevents.cpp
@@ -20,10 +20,12 @@
 #include "otpch.h"
 
 #include "baseevents.h"
+#include "configmanager.h"
 
 #include "pugicast.h"
 #include "tools.h"
 
+extern ConfigManager g_config;
 extern LuaEnvironment g_luaEnvironment;
 
 bool BaseEvents::loadFromXml()
@@ -34,7 +36,7 @@ bool BaseEvents::loadFromXml()
 	}
 
 	std::string scriptsName = getScriptBaseName();
-	std::string basePath = "data/" + scriptsName + "/";
+	std::string basePath = g_config.getString(ConfigManager::DATA_DIRECTORY) + scriptsName + "/";
 	if (getScriptInterface().loadFile(basePath + "lib/" + scriptsName + ".lua") == -1) {
 		std::cout << "[Warning - BaseEvents::loadFromXml] Can not load " << scriptsName << " lib/" << scriptsName << ".lua" << std::endl;
 	}
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -20,11 +20,13 @@
 #include "otpch.h"
 
 #include "chat.h"
+#include "configmanager.h"
 #include "game.h"
 #include "pugicast.h"
 #include "scheduler.h"
 
 extern Chat* g_chat;
+extern ConfigManager g_config;
 extern Game g_game;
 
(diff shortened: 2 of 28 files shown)


rookgaard:
@EPuncker Can you add reviewers please?

MillhioreBT:
I think that with Git we can already handle different branches of the same project and just changing branches is enough to test different data packages, at least that's how I do it

rookgaard:
> I think that with Git we can already handle different branches of the same project and just changing branches is enough to test different data packages, at least that's how I do it

That means, changes I proposed are exactly for you! :D

EPuncker:
I like this idea because right now for example I have two different servers that use the same source but different datapacks and it will fit just nicely

MillhioreBT (approved):
Just because it is a more comfortable way to do it for anyone, I accept it!

src/baseevents.cpp
```diff
@@ -34,7 +36,7 @@ bool BaseEvents::loadFromXml()
 	}
 
 	std::string scriptsName = getScriptBaseName();
-	std::string basePath = "data/" + scriptsName + "/";
+	std::string basePath = g_config.getString(ConfigManager::DATA_DIRECTORY) + scriptsName + "/";
```
ranisalt, line 39: Please use a [`std::filesystem::path`](https://en.cppreference.com/w/cpp/filesystem/path) instead

src/chat.cpp
```diff
@@ -287,10 +289,12 @@ Chat::Chat():
 
 bool Chat::load()
 {
+	std::string dataDirectory = g_config.getString(ConfigManager::DATA_DIRECTORY);
+	std::string chatchannelsFile = dataDirectory + "chatchannels/chatchannels.xml";
```
ranisalt, line 293: Same as above

src/script.cpp
```diff
@@ -41,7 +41,7 @@ bool Scripts::loadScripts(std::string folderName, bool isLib, bool reload)
 {
 	namespace fs = boost::filesystem;
```
ranisalt, line 42: You can now change this to be `std::filesystem`, we used `boost` because C++11 didn't have `filesystem` but C++17 has.

ranisalt (changes requested):
The suggestion applies to all changes you made.

However, I suggest to pass the data folder in the command line (e.g. `tfs --data path/to/data` defaulting to `$CWD/data`)

kornholi:
What do you think about a `ConfigManager::getDataPath` that returns `std::filesystem::path`? Then you could use `getDataPath() / "foo.xml"` etc.

DSpeichert:
We have `data/`, `config.lua` and `key.pem`.

You'd still need to change the working directory to be able to choose different config.lua and (rarely) different key.pem.
So having "same engine", how is this better than specifying e.g. `WorkingDirectory=/path/to...` in a systemd unit?

Zbizu:
Why not go further and create a login server?

DSpeichert:
This seems abandoned after a number of thoughtful reviews, so I'm closing this. Feel free to respond or reopen.

PR #3466, opened 2021-05-21
remove unused methods

rookgaard:
### Pull Request Prelude

### Changes Proposed

Removes old, unused functions.

**Issues addressed:** 

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -44,13 +44,6 @@ Account IOLoginData::loadAccount(uint32_t accno)
 	return account;
 }
 
-bool IOLoginData::saveAccount(const Account& acc)
-{
-	std::ostringstream query;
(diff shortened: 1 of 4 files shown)


rookgaard:
@EPuncker Can you add reviewers please?

nekiro:
Why? Is there any advantage to remove two methods? Which later might be used or by someone that use this engine?

rookgaard:
> Why? Is there any advantage to remove two methods? Which later might be used or by someone that use this engine?

@nekiro That's the point. Anyone who uses and will use this engine won't have any chance to make use of:
- `saveAccount` method, because after premium days behaviour were changed https://github.com/otland/forgottenserver/blob/master/src/iologindata.cpp#L1072-L1077 were added (which do the same)
- `getInstantSpellById` as that method is not used in that engine

ranisalt (approved):
If someone wants to use dead code like this they can keep their fork or copy from the history, it's not obliterated from existence

EPuncker (approved):

Issue #3234, opened 2020-12-02
Add hardness parameter to CONDITION_DRUNK


EPuncker:
### Explanation of what you want to do that is currently impossible
I want us to be able to control the chance that the character will random-step, right now its hard-coded to 20% (?) chance

### Desired functionality
- be able to control the chance to any desirable number between 1 to 100 by using the CONDITION_DRUNK
- be able to control it through monster file as well

### Our current code
```
void Creature::onWalk(Direction& dir)
{
	if (hasCondition(CONDITION_DRUNK)) {
		uint32_t r = uniform_random(0, 20);
		if (r <= DIRECTION_DIAGONAL_MASK) {
			if (r < DIRECTION_DIAGONAL_MASK) {
				dir = static_cast<Direction>(r);
			}
			g_game.internalCreatureSay(this, TALKTYPE_MONSTER_SAY, "Hicks!", false);
		}
	}
}
```
```
		} else if (tmpName == "drunk") {
			int32_t duration = 10000;

			if (spell->duration != 0) {
				duration = spell->duration;
			}

			Condition* condition = Condition::createCondition(CONDITIONID_COMBAT, CONDITION_DRUNK, duration, 0);
			combat->addCondition(condition);
```

EPuncker:
added on https://github.com/otland/forgottenserver/commit/488e37ab860db11baa8d2b3f5d6bbc4b6cd1a4a1

PR #3467, opened 2021-05-21
feat: add optional drunkenness parameter

amatria:
### Pull Request Prelude

### Changes Proposed
Add optional `drunkenness` parameter to `CONDITION_DRUNK`.

```xml
<attack name="drunk" interval="2000" chance="10" range="7" duration="5000" drunkenness="80">
	<attribute key="shootEffect" value="energy" />
</attack>
<!-- or -->
<attack name="drunk" interval="2000" chance="10" range="7" duration="5000">
	<attribute key="shootEffect" value="energy" />
</attack>
```

```lua
local drunk = Condition(CONDITION_DRUNK)
drunk:setParameter(CONDITION_PARAM_TICKS, 60000)
drunk:setParameter(CONDITION_PARAM_DRUNKENNESS, 80)
-- or
local drunk = Condition(CONDITION_DRUNK)
drunk:setParameter(CONDITION_PARAM_TICKS, 60000)
```

```lua
monster.attacks = {
	{name = "drunk", interval = 2000, chance = 10, range = 7, duration = 5000, drunkenness = 80},
	-- or
	{name = "drunk", interval = 2000, chance = 10, range = 7, duration = 5000}
}
```

**Issues addressed:** #3234 


src/condition.cpp
```diff
@@ -202,8 +202,11 @@ Condition* Condition::createCondition(ConditionId_t id, ConditionType_t type, in
 		case CONDITION_SPELLGROUPCOOLDOWN:
 			return new ConditionSpellGroupCooldown(id, type, ticks, buff, subId, aggressive);
 
-		case CONDITION_INFIGHT:
 		case CONDITION_DRUNK:
+			param = (param <= 0 ? 25 : param);
```
MillhioreBT, line 206: Why this?

MillhioreBT: This is really interesting and I'll have to try it sometime ;)

src/condition.cpp
```diff
@@ -1734,3 +1733,54 @@ bool ConditionSpellGroupCooldown::startCondition(Creature* creature)
...
+
+bool ConditionDrunk::updateCondition(const Condition* addCondition)
+{
+	const ConditionDrunk& conditionDrunk = static_cast<const ConditionDrunk&>(*addCondition);
+	return conditionDrunk.drunkness > drunkness;
+}
+
+void ConditionDrunk::addCondition(Creature* creature, const Condition* condition)
+{
+	const ConditionDrunk& conditionDrunk = static_cast<const ConditionDrunk&>(*condition);
```
nekiro, line 1755: same, just use pointer?

src/condition.cpp
```diff
@@ -1734,3 +1733,54 @@ bool ConditionSpellGroupCooldown::startCondition(Creature* creature)
...
+		return false;
+	}
+
+	creature->setDrunkness(drunkness);
+	return true;
+}
+
+bool ConditionDrunk::updateCondition(const Condition* addCondition)
+{
+	const ConditionDrunk& conditionDrunk = static_cast<const ConditionDrunk&>(*addCondition);
```
nekiro, line 1749: Use pointer

src/condition.cpp
```diff
@@ -202,8 +202,11 @@ Condition* Condition::createCondition(ConditionId_t id, ConditionType_t type, in
 		case CONDITION_SPELLGROUPCOOLDOWN:
 			return new ConditionSpellGroupCooldown(id, type, ticks, buff, subId, aggressive);
 
-		case CONDITION_INFIGHT:
 		case CONDITION_DRUNK:
```
nekiro, line 205: use {}

src/condition.cpp
```diff
@@ -1734,3 +1733,54 @@ bool ConditionSpellGroupCooldown::startCondition(Creature* creature)
...
 	return true;
 }
+
+bool ConditionDrunk::startCondition(Creature* creature)
+{
+	if (!Condition::startCondition(creature)) {
+		return false;
+	}
+
+	creature->setDrunkness(drunkness);
```
nekiro, line 1743: every single occurence of drunkness rename to drunkenes, drunkness is incorrect

src/creature.cpp
```diff
@@ -229,15 +229,19 @@ void Creature::onWalk()
...
 {
-	if (hasCondition(CONDITION_DRUNK)) {
-		uint32_t r = uniform_random(0, 20);
-		if (r <= DIRECTION_DIAGONAL_MASK) {
-			if (r < DIRECTION_DIAGONAL_MASK) {
-				dir = static_cast<Direction>(r);
-			}
-			g_game.internalCreatureSay(this, TALKTYPE_MONSTER_SAY, "Hicks!", false);
-		}
+	uint32_t r;
```
nekiro, line 232: when you are actually editing this, rename this to random or rand

nekiro (changes requested):

src/condition.cpp
amatria: > Why this?

To ensure backwards compatibility with the LUA interface. For instance, think of the `data/actions/scripts/other/fluids.lua` file. It looks something like this:

```lua
local drunk = Condition(CONDITION_DRUNK)
drunk:setParameter(CONDITION_PARAM_TICKS, 60000)

...
```

The parameter `CONDITION_PARAM_DRUNKENNESS` is never set. So, when this call to the LUA interface reaches that point in the code, the `param` value is 0, which would not cause a `drunkenness` effect on the creature. Therefore, I added this sanity check in order to fall back to the default parameter before the changes, 25, in case this condition is met.

(I hope it makes sense :p)

src/condition.cpp
amatria: Actually, I double checked before renaming and the correct spelling is: drunkenness. :p.

src/condition.cpp
nekiro: You can just set the default value in header file to 25 instead of checking it

src/condition.cpp
nekiro: Question still present

amatria: See bd2bb53. I squashed the commits and had to force an update. This snippet, indeed as GitHub points out, is outdated.

src/condition.cpp
amatria: Are you sure?

I could set it to 25 in the header file, but when the `createCondition` method calls the `ConditionDrunk` constructor it will overwrite it with a 0 because the default `param` value is, indeed, 0.

```c++
ConditionDrunk(ConditionId_t id, ConditionType_t type, int32_t ticks, bool buff, uint32_t subId, uint32_t drunkenness, bool aggressive = false) :
	Condition(id, type, ticks, buff, subId, aggressive), drunkenness(drunkenness) {}
```

nekiro (dismissed):

src/condition.cpp
nekiro: Well, you can just move the assignment to contructor block and check if drunkenness is 0

amatria: True. Let me change it real quick and the PR should be good to merge.

Thank you very much :p.

src/creature.h
```diff
@@ -512,6 +519,7 @@ class Creature : virtual public Thing
 		int32_t varSpeed = 0;
 		int32_t health = 1000;
 		int32_t healthMax = 1000;
+		uint32_t drunkenness = 0;
```
nekiro, line 522: Move it to the rest of uint32_t values

amatria: Done 👍 .

src/creature.cpp
```diff
@@ -229,15 +229,19 @@ void Creature::onWalk()
...
 {
-	if (hasCondition(CONDITION_DRUNK)) {
-		uint32_t r = uniform_random(0, 20);
-		if (r <= DIRECTION_DIAGONAL_MASK) {
-			if (r < DIRECTION_DIAGONAL_MASK) {
-				dir = static_cast<Direction>(r);
-			}
-			g_game.internalCreatureSay(this, TALKTYPE_MONSTER_SAY, "Hicks!", false);
-		}
+	uint32_t rand;
```
MillhioreBT, line 232: why create this variable, if there is a case where it is not even used?

amatria: At the end of the day, once it goes through the compiler, it will be translated into the same machine instructions.

For instance, the following two snippets:
```c
void foo() {
    int bar;

    if (rand() % 100 > 49) {
        return;
    }

    bar = 10;
    printf("%d", bar);

    return;
}
```
```c
void foo() {
    if (rand() % 100 > 49) {
        return;
    }

    int bar = 10;
    printf("%d", bar);

    return;
}
```
Are translated into the very same machine instructions
```asm
.LC0:
        .string "%d"
foo:
        sub     rsp, 8 ; allocate space in the stack for bar
        call    rand
        ; handle if
        cmp     eax, 49
        jle     .L5
        add     rsp, 8 ; deallocate b
        ret
.L5:
        ; handle b = 10 and printf
        add     rsp, 8 ; deallocate b
        jmp     printf
```

It is just a matter of style and giving "context" to a variable.

I have double checked the coding style guidelines and I have seen no entry for "variable locality" or so, but if you believe it will make the method more readable I can push the change to the branch.

ranisalt: > I have double checked the coding style guidelines and I have seen no entry for "variable locality" or so

https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide#local-variables

amatria: You just tilted me, not gonna lie. I appreciate the "effort" but your ---zero value--- comment clearly demonstrates you haven't thoroughly read the coding style guidelines. First of all, local variable does not equal variable locality. Secondly, because that very section lacks wording, the following snippet is completely correct under the TFS coding style guidelines:

```c
void foo() {
    int bar = 10;

    if (rand() % 100 > 49) {
        return;
    }

    printf("%d", bar);
    return;
}
```

Again, it does not solve the issue with the variable locality.

src/creature.cpp
```diff
@@ -229,15 +229,19 @@ void Creature::onWalk()
...
+
+	if (!hasCondition(CONDITION_DRUNK)) {
+		return;
 	}
+
+	rand = uniform_random(0, 99);
```
nekiro, line 238: uint32_t rand = 
and delete the variable above. Honestly I don't care if its in our coding styl guideline or not, but it just seems wrong, there is no point to explicitly define it like that, if we know the value at the time of declaration

nekiro (changes requested):

src/creature.cpp
amatria: Sure. I just pushed the changes. Let me know if there is anything else that doesn't meet the coding style criteria.

src/condition.cpp
```diff
@@ -1734,3 +1732,56 @@ bool ConditionSpellGroupCooldown::startCondition(Creature* creature)
...
+
+bool ConditionDrunk::updateCondition(const Condition* addCondition)
+{
+	const ConditionDrunk* conditionDrunk = static_cast<const ConditionDrunk*>(addCondition);
+	return conditionDrunk->drunkenness > drunkenness;
+}
+
+void ConditionDrunk::addCondition(Creature* creature, const Condition* condition)
+{
+	const ConditionDrunk* conditionDrunk = static_cast<const ConditionDrunk*>(condition);
```
nekiro, line 1754: move it above `setTicks(conditionDrunk->getTicks());`

src/condition.cpp
```diff
@@ -1734,3 +1732,56 @@ bool ConditionSpellGroupCooldown::startCondition(Creature* creature)
...
+{
+	creature->setDrunkenness(0);
+}
+
+uint32_t ConditionDrunk::getIcons() const
+{
+	return ICON_DRUNK;
+}
+
+bool ConditionDrunk::setParam(ConditionParam_t param, int32_t value) {
```
nekiro, line 1774: new line on { like the other methods

src/condition.h
```diff
@@ -429,4 +429,27 @@ class ConditionSpellGroupCooldown final : public ConditionGeneric
 		}
 };
 
+class ConditionDrunk final : public Condition
+{
+	public:
+		ConditionDrunk(ConditionId_t id, ConditionType_t type, int32_t ticks, bool buff, uint32_t subId, uint32_t drunkenness, bool aggressive = false) :
+			Condition(id, type, ticks, buff, subId, aggressive),
+			drunkenness(drunkenness == 0 ? 25 : drunkenness) {}
```
nekiro, line 437: I still think it would be better to move the assignment to scope and use something like
```
if (drunkenness != 0) {
     this->drunkenness = drunkeness;
}
```

nekiro (changes requested):

src/condition.h
amatria: Hard agree. Its more clear that way.

I just updated the branch with the changes.

nekiro (approved):
LGTM

DSpeichert (dismissed):

EPuncker (commented):
just a few random thoughts:

- wouldn't uint8_t suits better than uint32_t for drunknness?
- did you changed the default value from 20 to 25? why?

missing code in [register_monster_type](https://github.com/otland/forgottenserver/blob/master/data/scripts/lib/register_monster_type.lua#L272) :

```
						if attack.name == "drunk" and attack.condition.drunkenness then
spell:setConditionDrunkenness(attack.condition.drunkenness)
						end
```

amatria:
@EPuncker 

> wouldn't uint8_t suits better than uint32_t for drunknness?

You are right. In fact the `drunkenness` parameter takes values between 0 and 100. Let me change it real quick.

> did you changed the default value from 20 to 25? why?

The default value was 25. `r` took values between 0 and 20 and the creature would move randomly if `r <= DIRECTION_DIAGONAL_MASK`. Provided that `DIRECTION_DIAGONAL_MASK` is 4, then the probability of moving randomly is `5 / 21 = 0.238 ~= 0.25`.

> ```
> if attack.name == "drunk" and attack.condition.drunkenness then
> 	spell:setConditionDrunkenness(attack.condition.drunkenness)
> end
> ```

I completely missed this one. It has something to do with the `revscripts` right?

I'll update the branch in a few minutes. I'm very much sorry if the reviewers have to re-approve the PR again. My apologies :p.

(Also, for some reason I can't respond to your comment above)

Znote:
Your messing up the review process with all those force-pushes, its better to retain all the commits so we can see how each review feedback is addressed. When we merge the PR, we will squash everything into 1 commit, so don't worry about keeping a nice commit log here.

marmichalski:
> Your messing up the review process with all those force-pushes, its better to retain all the commits so we can see how each review feedback is addressed. When we merge the PR, we will squash everything into 1 commit, so don't worry about keeping a nice commit log here.
> 
> If we cant see each commit separately, its harder for us to follow along with the changes.

https://github.com/otland/forgottenserver/compare/373d8d82d89efcf84c4d40c4aac82448ea6adb8b..397939dc756001a50b8946c9a8ebd37eceb00cba

nekiro (dismissed):

EPuncker (dismissed):

src/creature.cpp
```diff
@@ -229,15 +229,17 @@ void Creature::onWalk()
...
+	if (!hasCondition(CONDITION_DRUNK)) {
+		return;
 	}
+
+	uint8_t rand = uniform_random(0, 99);
+	if (rand > getDrunkenness()) {
+		return;
+	}
+
+	dir = static_cast<Direction>(rand % 4);
```
ranisalt, line 241: If you put a chance of 1%, the drunk player will always "random" step north, because the value of `rand` here will always be 0. You need to re-roll the die to get the direction. Preferably, use the enum values instead of naked integers.

ranisalt (changes requested):

src/creature.cpp
amatria: I'd like to hear a second opinion on this one (@nekiro @EPuncker).

First of all, why would you set a 1% chance of drunkenness in the first place? Let alone, any value <5-10%? Moreover,
assuming you have a created a drunk condition with a drunkenness' value of 1% and, assuming again, it lasts for at least 200 SQMs, would you be able to tell wether your 2~4 drunk steps moved you to the north? Is it probable that you move 200 SQMs in a single direction?

Also, "rolling the die" again consumes CPU. Is it worth delaying the response time of the server in order to provide a, so called, "more realistic drunk effect" in such an edge case?

As for using enum values, do you think `rand % DIRECTION_DIAGONAL_MASK` would make the method more readable? In my opinion it adds a, rather useless, level of indirection. Are those values going to change in the foreseable future?

ranisalt: > First of all, why would you set a 1% chance of drunkenness in the first place?

Well, because it is allowed. I don't know what are reasonable values for this, to be fair.

> Also, "rolling the die" again consumes CPU

I wouldn't even worry about it, as the [Mersenne Twister, available as `std::mt19937`, can generate a random number in 6 cycles](https://github.com/s9w/articles/blob/master/perf%20cpp%20random.md). You could generate a million random numbers and the response time will barely be affected, even using the slower `uniform_int_distribution`.

You can, however, generate a number between 0 and 400, use the last 2 bits for the direction and the high bits for chance, i.e.

```cpp
	uint16_t rand = uniform_random(0, 399);
	if (rand / 4 > getDrunkenness()) {
		return;
	}

	dir = static_cast<Direction>(rand % 4);
```

This is still guaranteed to be uniformly distributed.

> As for using enum values, do you think `rand % DIRECTION_DIAGONAL_MASK` would make the method more readable?

Not more readable, but more semantic. I meant `uniform_random(DIRECTION_NORTH, DIRECTION_WEST)`, but thinking again it doesn't help that much indeed, so dismiss it.

data/scripts/lib/register_monster_type.lua
```diff
@@ -335,6 +335,9 @@ registerMonsterType.attacks = function(mtype, mask)
 					if attack.shootEffect then
 						spell:setCombatShootEffect(attack.shootEffect)
 					end
+					if attack.name == "drunk" and attack.drunkenness then
+						spell:setConditionDrunkenness(attack.drunkenness)
+					end
```
soul4soul, line 340: Now that drunk is a condition the implementation of attacks for it in both xml and RevScriptSys feels inconsistent.

Should a "drunkcondition" attack be accepted for XML monsters to be consistent with other conditions?

On the RevScriptSys side, this PR requires `attack type` to now be set to condition_drunk for drunk attacks which is breaking change for RevScriptSys, not that many people use it. I just as much rather see the check be changed to `if attack.drunkenness then` and require attack type be set correctly but I know other devs here are pretty picky about that stuff. To be backwards compatible can we do something like this? 

```suggestion
					if attack.name == "drunk" or (attack.drunkenness and attack.type == CONDITION_DRUNK) then
     					spell:setConditionType(CONDITION_DRUNK)
						spell:setConditionDrunkenness(attack.drunkenness)
					end
```

The current way to cause drunk would continue working, and to cause drunk using `condition` attacks would work.

```
monster.attacks = {
	{name = "drunk", interval = 2000, chance = 10, range = 7, duration = 5000}
}
```
and
```
monster.attacks = {
	{name = "condition", type = CONDITION_DRUNK, interval = 2000, chance = 10, range = 7, duration = 5000, drunkenness = 80}
}
```

src/creature.cpp
MillhioreBT: Any reason not to trust `uniform_random`?

amatria: > You can, however, generate a number between 0 and 400, use the last 2 bits for the direction and the high bits for chance, i.e.

I like this solution very much indeed. Let me update the branch.

data/scripts/lib/register_monster_type.lua
amatria: I doubled checked this comment but failed to understand it at first.

I have now checked again the file `data/monster/lua/#example.lua` and you are right:
```lua
monster.attacks = {
	{name = "drunk", interval = 2000, chance = 10, range = 7, duration = 5000}
}
```
This should be the right way to declare a `drunk` type of attack. I thought it should follow the same syntax as in the XML files. I'll update the branch right away.

However, in my opinion the LUA logic must be:
```lua
if attack.name == "drunk" then
	spell:setConditionType(CONDITION_DRUNK)
	-- it is not mandatory to set a drunkenness value, by default it falls back to a 25% value
	if attack.drunkenness then
		spell:setConditionDrunkenness(attack.drunkenness)
	end
end
```

ranisalt: @amatria if you wish. You can update it at a later date, that would be very helpful.

MillhioreBT:
Tested and working great, I love this change
If you have something else to improve, do it, and have my approval, this change is very nice

nekiro (approved):

ranisalt (approved):
LGTM, my only concern is the magic number 25

EPuncker (approved):

PR #3468, opened 2021-05-22
Missing `self` in function Player.getWeaponType

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
This function is not used anywhere and that is why we had not realized that the function lacks the parameter `self` :D


nekiro:
It's not used, so it does not have to be there.

EPuncker:
> It's not used, so it does not have to be there.

it is used [here](https://github.com/otland/forgottenserver/blob/master/data/lib/compat/compat.lua#L518)

nekiro:
> > It's not used, so it does not have to be there.
> 
> it is used [here](https://github.com/otland/forgottenserver/blob/master/data/lib/compat/compat.lua#L518)

In method, self is not used

nekiro (approved):

EPuncker (approved):

PR #3473, opened 2021-06-03
Updated shovel function so the player goes down automatically when opening a hole

EPuncker:
### Pull Request Prelude

### Changes Proposed
When opening a hole now, the player goes down automatically
yes, that is how it works now. at least from what I could test

### DIFF
--- a/data/actions/lib/actions.lua
+++ b/data/actions/lib/actions.lua
@@ -199,6 +199,7 @@ function onUseShovel(player, item, fromPosition, target, toPosition, isHotkey)
 
 		toPosition.z = toPosition.z + 1
 		tile:relocateTo(toPosition)
+		player:teleportTo(toPosition, false)
 		player:addAchievementProgress("The Undertaker", 500)
 	elseif target.itemid == 20230 then -- swamp digging
 		if (player:getStorageValue(PlayerStorageKeys.swampDigging)) <= os.time() then


soul4soul:
I can confirm this changed in cipbia, I tested a shovel spot east of thais.

DSpeichert:
Is it all the holes working like that now or just the new ones?

EPuncker:
> Is it all the holes working like that now or just the new ones?

actually I tested these {468, 481, 483}, I was not able to test 7932, 23712 but I doubt their behavior will differ

Znote:
I don't remember this behavior, at which protocol was this introduced?

MillhioreBT:
What if I just want to open the hole? It is quite strange how Cip forces us to go down through the hole just by opening it, I don't like how it works, but if we must respect the mechanics of Tibia, then there is no other option but to follow them to the letter, probably I will not use this in my custom server :v

DSpeichert (approved):

soul4soul:
@EPuncker Do you have premium? Did they make the same change to hidden tombs in ank?

nekiro:
I would say this is useless feature and changes nothing relevant and literally forces people to go down if they open a hole, so they have to rope up. Imagine you have cave with monsters and you shovel it down and you die, because it walks you in automatically XD

EPuncker:
> @EPuncker Do you have premium? Did they make the same change to hidden tombs in ank?

no I don't have premium :( but you have a point now

> I would say this is useless feature and changes nothing relevant and literally forces people to go down if they open a hole, so they have to rope up. Imagine you have cave with monsters and you shovel it down and you die, because it walks you in automatically XD

well, just like the change they did so we can't rope up monsters anymore, but indeed I'll add a small config in the file for it

data/actions/lib/actions.lua
```diff
@@ -199,6 +199,7 @@ function onUseShovel(player, item, fromPosition, target, toPosition, isHotkey)
 
 		toPosition.z = toPosition.z + 1
 		tile:relocateTo(toPosition)
```
marmichalski, line 201: Not related to this PR, but I've just checked how it behaves in Tibia and...

Screenshot 2021-06-21 at 14 45 49

Item(s) blocking loose stone pile is not relocated 🤔.
I'd see what happens with players, but I have no friends 😢

marmichalski: Screenshot 2021-06-21 at 14 54 41

Fields are also not removed. They are removed `onStepIn`. Relocation happens `onStepIn` too. 🤔

EPuncker: 🤔

EPuncker: also desert quest entrance stone pile doesn't make player go down on opening and I have no idea why

marmichalski: The sand stone pile also relocates the items when using shovel. #cip quality 😎 
Anyway, the auto-enter when using shovel was introduced around 2016, knowing cip they just forgot to update every item to act the same 🤷

DSpeichert (approved):

Zbizu:
Was this tested in situation when there's teleport or stairs instead of rope spot?

marmichalski:
> Was this tested in situation when there's teleport or stairs instead of rope spot?

Teleport or stairs? Where one would find such a place? Desert dungeon hole has ladder, one of the holes I've tested has rope spot (rotworm cave near dwarven bridge), the other has ladder with energy field (hidden hole north of "tiny" amazon camp near venore, tunnel to witches place).

Zbizu:
@marmichalski ghostland demon skeleton cave

marmichalski:
> @marmichalski ghostland demon skeleton cave if Im not misremembering
> 
> plus I meant testing on tfs if it doesn't bug

There is regular stairs "hole", if that's the place: https://youtu.be/rGSULvV4L18?t=84

EPuncker:
I don't think this is necessary after all

Issue #3411, opened 2021-04-15
Walking should NOT be canceled when doing any actions


EPuncker:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce
1. Walk into a straight line
2. Do any action (use item, use spell) while walking

### Expected behaviour
The character should perform the action and continue moving

### Actual behaviour
The character stops moving

### Reproduction environment
You can easily see the difference by testing cipbia and tfs

EPuncker:
it is client sided 😢

marmichalski:
You sure?
https://github.com/otland/forgottenserver/blob/48167ef5fee895f8195eebd82d15e6d01cece159/src/actions.cpp#L432-L435

EPuncker:
> You sure?
> https://github.com/otland/forgottenserver/blob/48167ef5fee895f8195eebd82d15e6d01cece159/src/actions.cpp#L432-L435

yes, it still stops the walking, you can notice its client sided because on OTC it doesn't stop, while on cip client it does

Codinablack:
This occurs in both useItem and useItemEx, to me, it makes no sense why stopWalk would be used in either case.

Is it truly preferred that one must stop for use of any and all items, I haven't played the original in a long time, but I always remembered being able to run while using items. 

The fact that in otlclient the player isn't stopped blows my mind, especially when I commented out both stopWalk() calls in useItem, and useItemEx, and it no longer interfered with auto walk, my character walked to where I clicked %100 of the time with 10.98 cip client, and without the comments, player stopped short everytime. 

Seems like an easy fix for the actions and to be able to close #3464 

I didn't see the stopWalk() called at all in spells, nor did I see any obvious or apparent reasons as to why it would stop you while casting except that perhaps most the spells check for position, maybe that has something to do with it.

Is there any reason we shouldn't implement the fix for #3464 by removing those lines? I assure you, it most definitely allows cip client to keep moving.

Znote:
Submit pull request and let reviewers test suggested changes. If its a stable improvement it will get merged.

EPuncker:
last time I tested this after @marmichalski pointed that out, the character still stopped walking (note: using arrows to walk, not mapclick)

Codinablack:
[...]

PR #3474, opened 2021-06-08
fixed stopped walking onuseitem/itemex

Codinablack:
### Pull Request Prelude

- [ x] I have read and understood the [contribution guidelines][cont] before making this PR.

### Changes Proposed

**Issues addressed:** 
fixes issue #3411


Codinablack:
> Does this work with the Cip client? I can't test it now, but with OTC it works

Yes this definitely works with cip client from my own testing. This PR does however require a review, so if you feel up to it, test for yourself so we can get the bug finally closed and behind us.

nekiro:
If you use hotkeys, client will stop you anyway (cip client < 11 version)

MillhioreBT:
> If you use hotkeys, client will stop you anyway (cip client < 11 version), but for other things like manually using, it should work.

confirmed!

Codinablack:
I will look into how to fix the issue with hotkeys then , and rebase, and post a new pr when I have some time

nekiro:
> I will look into how to fix the issue with hotkeys then , and rebase, and post a new pr when I have some time

but the thing is you cant. It's client sided.

Codinablack:
If that is the case, and both ways are bugged, server and client, should we not at least fix the server side?
So if it is an item not being used with a hotkey, as we do allow checking "hotkey" use in onUseItem() anyways, the user doesn't have the problem? 

Maybe a user wants to completely disallow all hotkey usage on their server... then the bug would only exist on our end right?
I personally prefer hotkey usage disabled, but surely I'm not the only one. 

I have seen, without the changes, the problem exists even outside of hotkey use, and since it exists outside of the client, is it not our responsibility to fix the problem?

If the dev's disagree, then I will close, but if not, then I believe this PR is good for merger, causes no bugs that I have found, and fixes our end of issue #3411

DSpeichert (approved):

DSpeichert:
I agree that map-click walk does not need to be cancelled when using items (manually, not via hotkey if that's what client controls anyway).

Issue #3476, opened 2021-06-08
liquid containers ignoring subtype parameter in monster loot


Xikini:
### Before creating an issue, please ensure:
- [yes ] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [yes ] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. put subtype="0" for a liquid container item, such as golden mug in a creature loot table
2. kill monster
3. find water in container, regardless of subtype

### Expected behaviour

find water in container, regardless of subtype

### Actual behaviour

subtype should change the liquid in the container, and lack of a subtype should default to an empty container

EPuncker:
fixed on https://github.com/otland/forgottenserver/commit/9d01be7061c96e7bd46e8c68529f412888cfb2d7

PR #3477, opened 2021-06-08
Fix subType in loot

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
**Issues addressed:** #3476

### BTW
We will probably have to add `subType="0"` to many liquid containers in all monsters

### DIFF
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -18,7 +18,7 @@ function Container.createLootItem(self, item)
 	end
 
 	if itemCount > 0 then
-		local tmpItem = Game.createItem(item.itemId, math.min(itemCount, 100))
+		local tmpItem = Game.createItem(item.itemId, item.subType ~= -1 and item.subType or math.min(itemCount, 100))
 		if not tmpItem then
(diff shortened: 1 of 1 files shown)


nekiro:
This is very problematic, whole the subtype thing is bad implementation. I would say just create new variables for xml files like fluidType, charges etc and leave subType like it's used now and then edit monsters xml files to drop items correctly.

MillhioreBT:
> This is very problematic, whole the subtype thing is bad implementation. I would say just create new variables for xml files like fluidType, charges etc and leave subType like it's used now and then edit monsters xml files to drop items correctly.

I think it is a good option

MillhioreBT:
Anyone who wants to have an opinion about what @nekiro is proposing can do so

EPuncker:
I agree with nekiro, and everything will be okay as long as we keep backward compatibility

EPuncker:
fixed on https://github.com/otland/forgottenserver/commit/9d01be7061c96e7bd46e8c68529f412888cfb2d7

PR #3481, opened 2021-06-11
Add possibility to return outfit in onChangeOutfit event

### LINKED ISSUE #3290: EventCallback onChangeOutfit causes Call Stack Overflow

1. edit data/events/events.xml enabling onChangeOutfit (change to 1)
2. in /data/scripts create a new file (any name with lua extension)
3. in this script, add this code:

```
function dump(o)
    if type(o) == 'table' then
       local s = '{ '
       for k,v in pairs(o) do
          if type(k) ~= 'number' then k = '"'..k..'"' end
          s = s .. '['..k..'] = ' .. dump(v) .. ','
       end
       return s .. '} '
    else
       return tostring(o)
    end
 end


local blockChangeColor = EventCallback
    function blockChangeColor.onChangeOutfit(cid, outfit)
        local player = Player(cid)
        print(dump(outfit))
        print("changing outfit")
        outfit.lookHead = 1
        outfit.lookBody = 1
        outfit.lookLegs = 1
        outfit.lookFeet = 1
        player:setOutfit(outfit)
        return true
    end
```
in this script, I tried to block player from changing colors, but keeping the possibility to change addons/looktype.

### Expected behaviour
player:setOutfit(outfit) shouldn't trigger again the onChangeOutfit code, as it was changed by developer and not by client. I believe this function should only be triggered when the client asks to change the outfit, and not the code itself.

### Actual behaviour
player:setOutfit(outfit) is calling onChangeOutfit, which calls player:setOutfit(outfit) and keep this in loop, until stack overflow

nekiro:
Resolve #3290 

This pr adds functionality to set outfit without calling setOutfit inside of the event.\
Simply return changed outfit or false to prevent change like it was before.

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -169,7 +169,7 @@ bool Events::eventMonsterOnSpawn(Monster* monster, const Position& position, boo
 }
 
 // Creature
-bool Events::eventCreatureOnChangeOutfit(Creature* creature, const Outfit_t& outfit)
+bool Events::eventCreatureOnChangeOutfit(Creature* creature, Outfit_t& outfit)
 {
 	// Creature:onChangeOutfit(outfit) or Creature.onChangeOutfit(self, outfit)
 	if (info.creatureOnChangeOutfit == -1) {
@@ -192,7 +192,22 @@ bool Events::eventCreatureOnChangeOutfit(Creature* creature, const Outfit_t& out
 
 	LuaScriptInterface::pushOutfit(L, outfit);
 
-	return scriptInterface.callFunction(2);
+	bool success = true;
+
+	if (scriptInterface.protectedCall(L, 2, 1) != 0) {
+		success = false;
+		LuaScriptInterface::reportError(nullptr, LuaScriptInterface::popString(L));
+	} else {
+		if (LuaScriptInterface::isTable(L, -1)) {
+			outfit = LuaScriptInterface::getOutfit(L, -1);
+		} else {
+			success = LuaScriptInterface::getBoolean(L, -1);
+			lua_pop(L, 1);
+		}
+	}
+
+	scriptInterface.resetScriptEnv();
+	return success;
 }
 
 ReturnValue Events::eventCreatureOnAreaCombat(Creature* creature, Tile* tile, bool aggressive)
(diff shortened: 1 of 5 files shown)


ranisalt:
I think a more elegant solution would be for the event to return the outfit, so it modifies when needed but not call `setOutfit` again. Then, apply the modified (or not) outfit.

MillhioreBT:
> I think a more elegant solution would be for the event to return the outfit, so it modifies when needed but not call `setOutfit` again. Then, apply the modified (or not) outfit.
> 
> This can easily be made backwards compatible if desird.

I love this solution, very elegant and easy

nekiro:
> I think a more elegant solution would be for the event to return the outfit, so it modifies when needed but not call `setOutfit` again. Then, apply the modified (or not) outfit.
> 
> This can easily be made backwards compatible if desird.

This is interesting

Zbizu:
> This is interesting

Health change creature event is made that way and it works really well so outfits can work that way too.

src/events.cpp
```diff
@@ -192,7 +192,22 @@ bool Events::eventCreatureOnChangeOutfit(Creature* creature, const Outfit_t& out
...
 
-	return scriptInterface.callFunction(2);
+	bool success = true;
+
+	if (scriptInterface.protectedCall(L, 2, 1) != 0) {
+		success = false;
+		LuaScriptInterface::reportError(nullptr, LuaScriptInterface::popString(L));
+	} else {
+		if (LuaScriptInterface::isTable(L, -1)) {
+			outfit = LuaScriptInterface::getOutfit(L, -1);
```
MillhioreBT, line 202: getOutfit does not work with relative indices, I suppose that 1 is fine, since the return value will always be in this position, unless there is something else that I do not know?

MillhioreBT (changes requested):
It is the only bad thing that I noticed about this PR, once these two problems are solved and everything will work correctly.

DSpeichert (commented):
@nekiro do you plan to tweak it like @ranisalt mentioned?

nekiro:
If anyone interested in picking this up, feel free

PR #3482, opened 2021-06-12
Revscriptsys monster fixes

soul4soul:
### Pull Request Prelude

### Changes Proposed

Fix a couple issues with registering monsters via revscriptsys system

**Issues addressed:** 

- Couldn't set canWalkOnFIELD flags
- Couldn't set summonable flag
- Couldn't use min/max damage attribute for melee attacks, it was only possible to use attack/skill
- Couldn't set direction of attacks or defenses declared using scripts, this is possible in the XML interface
- Couldn't override the ingame creature name, the value/name used to register the monster is the monster name by default. The ability to override the name is necessary for creatures like "red butterfly" and "demon goblin"

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -123,13 +123,13 @@ registerMonsterType.flags = function(mtype, mask)
 		if mask.flags.staticAttackChance then
 			mtype:staticAttackChance(mask.flags.staticAttackChance)
 		end
-		if mask.flags.canWalkOnEnergy then
+		if mask.flags.canWalkOnEnergy ~= nil then
 			mtype:canWalkOnEnergy(mask.flags.canWalkOnEnergy)
 		end
-		if mask.flags.canWalkOnFire then
+		if mask.flags.canWalkOnFire ~= nil then
 			mtype:canWalkOnFire(mask.flags.canWalkOnFire)
 		end
-		if mask.flags.canWalkOnPoison then
+		if mask.flags.canWalkOnPoison ~= nil then
 			mtype:canWalkOnPoison(mask.flags.canWalkOnPoison)
 		end
 	end
@@ -400,6 +400,9 @@ registerMonsterType.defenses = function(mtype, mask)
 						spell:setType("melee")
 						if defense.attack and defense.skill then
 							spell:setAttackValue(defense.attack, defense.skill)
+						end
+						if attack.minDamage and attack.maxDamage then
+							spell:setCombatValue(attack.minDamage, attack.maxDamage)
 						end
 						if defense.interval then
 							spell:setInterval(defense.interval)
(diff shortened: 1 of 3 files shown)


EPuncker (dismissed):

EvilHero90 (dismissed):
Looks good to me, I'm still kinda curious with the double attack bug as we had a fix for attacks/defenses/loot a while ago when reloading, if you could provide a bit more information I'll gladly take a look at it

data/scripts/lib/register_monster_type.lua
```diff
@@ -378,6 +401,9 @@ registerMonsterType.defenses = function(mtype, mask)
 						if defense.attack and defense.skill then
 							spell:setAttackValue(defense.attack, defense.skill)
 						end
+						if attack.minDamage and attack.maxDamage then
```
MillhioreBT, line 404: XD!

data/scripts/lib/register_monster_type.lua
```diff
@@ -476,6 +502,9 @@ registerMonsterType.defenses = function(mtype, mask)
 					if defense.target then
 						spell:setNeedTarget(defense.target)
 					end
+					if attack.direction then
```
MillhioreBT, line 505: x2

MillhioreBT (changes requested):
put the correct name of the table `attack` chanto to `defense`

data/scripts/lib/register_monster_type.lua
MillhioreBT, line 505: x2

data/scripts/lib/register_monster_type.lua
MillhioreBT: https://github.com/otland/forgottenserver/blob/05f6d203b7956dee0c37091e4b0d4429845d413e/data/scripts/lib/register_monster_type.lua#L505

soul4soul: Ah yes! fixed.

data/scripts/lib/register_monster_type.lua
```diff
@@ -502,7 +502,7 @@ registerMonsterType.defenses = function(mtype, mask)
 					if defense.target then
 						spell:setNeedTarget(defense.target)
 					end
-					if attack.direction then
+					if defense.direction then
 						spell:setNeedDirection(attack.direction)
```
kornholi, line 506: should this be `defense` as well?

kornholi: What about extracting the `MonsterSpell` configuration code into its own function so we can share it between `attacks` and `defenses`?

kornholi: nevermind--I thought these functions were part of your PR for a second.

EPuncker (approved):

data/scripts/lib/register_monster_type.lua
soul4soul: I'm glad you guys are checking!

soul4soul:
> 
> 
> Looks good to me, I'm still kinda curious with the double attack bug as we had a fix for attacks/defenses/loot a while ago when reloading, if you could provide a bit more information I'll gladly take a look at it

I did figure it out after all and it was self caused.

PR #3483, opened 2021-06-13
Fix cotire builds

kornholi:
Cotire must be invoked last on a target or they might get out of sync.

In my case it was trying to build with two different versions of a dependency.


ranisalt:
Since CMake now supports it by default, why not just drop cotire and use the builtin feature?

DSpeichert:
cmake 3.16 is available since Ubuntu 20.04 LTS.

DSpeichert (approved):

PR #3390, opened 2021-03-29
👁 Fix sight line algorithm 👁

MillhioreBT:
### Pull Request Prelude
- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Emulate the behavior with respect to the line of sight of Tibia RL, this topic was already in progress in the following PR #3265, but for some reason the guy who proposed this closed his PR and deleted his repository, I still want to give him credit for inspiring these changes and giving the idea

### Need Testing
I have test this many times, however I do not have a Tibia RL account good enough to run all kinds of tests, it would be good if someone else will test these changes to check that everything is correct

### Any help
Any suggestion or help regarding the current code is welcome, it is possible that much more can be improved

**Issues addressed:** #2891

### GIF's
**Ray Test**
Here I show an example of many rays that verify the accessible places with this algorithm, make your comparisons with Tibia RL
<details>
<summary>Ray animation</summary>

GIF 29-03-2021 07-41-32 p  m
GIF 29-03-2021 07-43-19 p  m

</details>

PR #3487, opened 2021-06-26
full update on isSightClear

### LINKED ISSUE #2727: Wrong behavior of monsters with distance attacks.

### LINKED ISSUE #2891: Directional spells vs. walls behaviour

Zbizu:
### Pull Request Prelude

### Changes Proposed
use Xiaolin Wu algorithm instead of current one to emulate the behaviour of sight line checks more accurately

**Issues addressed:** 
Fixes #2727
Fixes #2891 
- inaccuracy in throwing items, spells and ranged attacks

### Description
credits:
Closes #3390 for steep/slight line functions

long awaited rework of distance attacks with plenty of tests performed
note: the diff file is quite unreadable so I recommend looking into the code itself

changes since #3390:
- changed line functions return value back to booleans
- moved Z checks from Map::canThrowObjectTo to Map::isSightClear for consistency with lua scripts.
- made floorCheck parameter optional for Map::isSightClear
- rewrote Map::isSightClear function to reproduce the logic of throwing between floors more accurately

### Tests performed
(the videos are quite heavy so I uploaded them to imgur)

magic wall test:
https://i.imgur.com/e3YtyOT.mp4

line shape test:
https://i.imgur.com/9Wpv0kJ.mp4

wall interaction test:
https://i.imgur.com/9aOFYVk.mp4

throwing between floors test (had to cut this one because it was too long):
https://i.imgur.com/GY91pf1.mp4

zero crashes during local tests


joseluis2g:
Monster are the same? have you checked it? May you try to do some classic hunt and post the video?
Great PR indeed.

Zbizu:
lines mirror test (note: the asymmetry is intended in some cases):
https://i.imgur.com/h70KBzz.mp4

orc spearman test (note: magic effects come from testing build, they aren't included in this PR)
https://i.imgur.com/Rb72cgo.mp4

in case you want to test the generated path yourself, you have to add this at the beginning of `bool Map::isTileClear(...)`:
```
	g_game.addMagicEffect(Position(x, y, z), CONST_ME_POFF);
```

src/map.cpp
```diff
@@ -529,62 +522,164 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		if (z0 == 0) {
+			return true;
+		}
+
+		//attempt to find a path above the obstacle
+		--z0;
+		sightClear = checkSightLine(x0, y0, x1, y1, z0);
+
+		//check if there are no ground tiles directly above us and the target
+		bool aboveFromPos = isTileClear(x0, y0, z0, true);
```
nekiro, line 635: why not move this into the return statement, same with another bool under this one

src/map.cpp
```diff
@@ -529,62 +522,164 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		bool aboveToPos = isTileClear(x1, y1, z0, true);
+
+		return sightClear && aboveFromPos && aboveToPos;
+	} else {
+		//target is on a different floor
+
```
nekiro, line 641: remove new line

src/map.cpp
```diff
@@ -529,62 +522,164 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		}
+
+		//skip checks for sight line in case fromPos and toPos cross the ground floor
+		if (z0 < 8 && z1 > 7 || z0 > 7 && z1 < 8) {
+			return false;
+		}
+
+		if (z0 > z1) {
+			//target is above us
+
```
nekiro, line 653: remove new line

src/map.cpp
```diff
@@ -529,62 +522,164 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+			int32_t deltaz = Position::getDistanceZ(fromPos, toPos);
+
+			if (deltaz > 1) {
+				return false;
+			}
+
+			--z0;
+			bool aboveFromPos = isTileClear(x0, y0, z0, true);
+			bool sightClear = checkSightLine(x0, y0, x1, y1, z0);
+
```
nekiro, line 663: same thing can be moved to return directly?

src/map.cpp
```diff
@@ -529,62 +522,164 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+			return sightClear && aboveFromPos;
+		} else {
+			//target is below us
+
+			bool sightClear = checkSightLine(x0, y0, x1, y1, z0);
+
```
nekiro, line 669: remove new line also can be moved to if check right?

src/map.cpp
```diff
@@ -529,62 +522,164 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+			if (!sightClear){
```
nekiro, line 670: missing space after )

nekiro (changes requested):

Zbizu:
context for the revert on floor check: spells that use blockWalls attribute use this function with `checkLineOfSight = false`

ranisalt:
I don't get why Wu's algorithm is used here, it's meant for anti-aliasing. We can use [Bresenham's algorithm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm) that draws the exact same line without the extra steps and magic numbers.

Zbizu:
@ranisalt If I remember correctly, it was giving different results in lines mirror test. If you code it more optimal way, I can help you with testing.

ranisalt:
Hmm I think it is related to rounding, as it is considering that the position of a tile is the top left corner of it and not the middle (therefore it is not mirrored accross the X and Y axis).

I have this implementation of  Bresenham's that, as far as I'm concerned, fixes the axis mirroring issue:

```lua
local function getSlightLine(x0, y0, x1, y1)
    if x0 > x1 then
        x0, y0, x1, y1 = x1, y1, x0, y0
    end

    local dx, dy, yi = x1 - x0, y1 - y0, 1
    if dy < 0 then
        yi, dy = -1, -dy
    end

    local d, y = 2 * dy - dx, y0
    for x = x0, x1 do
        if d > 0 then
            y, d = y + yi, d + 2 * (dy - dx)
        else
            d = d + 2 * dy
        end
    end
end

local function getSteepLine(x0, y0, x1, y1)
    if y0 > y1 then
        x0, y0, x1, y1 = x1, y1, x0, y0
    end

    local dx, dy, xi = x1 - x0, y1 - y0, 1
    if dx < 0 then
        xi, dx = -1, -dx
    end

    local d, x = 2 * dx - dy, x0
    for y = y0, y1 do
        if d > 0 then
            x, d = x + xi, d + 2 * (dx - dy)
        else
            d = d + 2 * dx
        end
    end
end

local abs = math.abs
local function getLine(fromPosition, toPosition)
    local z = fromPosition.z

    local x0, y0, x1, y1 = fromPosition.x, fromPosition.y, toPosition.x, toPosition.y
    local dx, dy = x1 - x0, y1 - y0
    if abs(dy) < abs(dx) then
        getSlightLine(x0, y0, x1, y1)
    else
        getSteepLine(x0, y0, x1, y1)
    end
end
```

Though I don't really remember what I changed to fix the line AND it is not throughly tested. @MillhioreBT might have some light on this

itutorial007:
About the "magic numbers"

show

You can see the line when drawn is passing through the tile with the bush.

It isn't actually adding the tile to the line though because its not under half of the tile which is required for the algorithms to add them.

The way tibia is set up - they made it so instances like this would block the sightline to follow their "3d" style.

The "magic number" is required to offset how the line is drawn so it will pick up that tile. Without the offset we would have to tell the code to pick it up in a different way.

1) We pick up all tiles around the line and determine if it should cause a block
2) We add that center tile that isn't being picked up as the code runs and create a hardcode just for that specific tile.

Any other options are going to require more code than just setting an offset. I say this changes it from being a "magic number" to being "the correct number for the offset" so the line will be drawn correctly.

These algorithms are designed to create a straight line, they do not take into account any changes that might be created based on some 3d styling or what ever. So, it needs to be decided if we are going to create all the code needed to pick up the correct tiles (which is more than the algorithms themselves) or if an offset will do.

Zbizu:
@ranisalt your implementation fixes the upper line, but breaks the lower
obraz

your code wrapped to action in case you want to test it yourself:
```
function getSlightLine_r(x0, y0, x1, y1, z)
    if x0 > x1 then
        x0, y0, x1, y1 = x1, y1, x0, y0
    end

    local dx, dy, yi = x1 - x0, y1 - y0, 1
    if dy < 0 then
        yi, dy = -1, -dy
    end

    local d, y = 2 * dy - dx, y0
    for x = x0, x1 do
        if d > 0 then
            y, d = y + yi, d + 2 * (dy - dx)
        else
            d = d + 2 * dy
        end
		Position(x, y, z):sendMagicEffect(CONST_ME_POFF)
    end
end

function getSteepLine_r(x0, y0, x1, y1, z)
    if y0 > y1 then
        x0, y0, x1, y1 = x1, y1, x0, y0
    end

    local dx, dy, xi = x1 - x0, y1 - y0, 1
    if dx < 0 then
        xi, dx = -1, -dx
    end

    local d, x = 2 * dx - dy, x0
    for y = y0, y1 do
        if d > 0 then
            x, d = x + xi, d + 2 * (dx - dy)
        else
            d = d + 2 * dx
        end
		Position(x, y, z):sendMagicEffect(CONST_ME_POFF)
    end
end

function getLine_r(fromPosition, toPosition)
    local z = fromPosition.z

    local x0, y0, x1, y1 = fromPosition.x, fromPosition.y, toPosition.x, toPosition.y
    local dx, dy = x1 - x0, y1 - y0
    if math.abs(dy) < math.abs(dx) then
        return getSlightLine_r(x0, y0, x1, y1, fromPosition.z)
    else
        return getSteepLine_r(x0, y0, x1, y1, fromPosition.z)
    end
end

local b_ranisalt = Action()

function b_ranisalt.onUse(player, item, fromPosition, target, toPosition, isHotkey)
	if not target then
		return false
	end
	
	local pos = target:getPosition()
	if not pos then
		return false
	end
	
	getLine_r(player:getPosition(), pos)
	return true
end

b_ranisalt:id(24174)
b_ranisalt:allowFarUse(true)
b_ranisalt:blockWalls(false)
b_ranisalt:checkFloor(false)

b_ranisalt:register()
```

src/map.cpp
```diff
@@ -529,62 +529,146 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
-	Position start(fromPos.z > toPos.z ? toPos : fromPos);
-	Position destination(fromPos.z > toPos.z ? fromPos : toPos);
-
-	const int8_t mx = start.x < destination.x ? 1 : start.x == destination.x ? 0 : -1;
-	const int8_t my = start.y < destination.y ? 1 : start.y == destination.y ? 0 : -1;
-
-	int32_t A = Position::getOffsetY(destination, start);
-	int32_t B = Position::getOffsetX(start, destination);
-	int32_t C = -(A * destination.x + B * destination.y);
+	if (blockFloor && tile->getGround()){
```
nekiro, line 539: space before {

src/map.cpp
```diff
@@ -529,62 +529,146 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		return false;
+	}
 
-	while (start.x != destination.x || start.y != destination.y) {
-		int32_t move_hor = std::abs(A * (start.x + mx) + B * (start.y) + C);
-		int32_t move_ver = std::abs(A * (start.x) + B * (start.y + my) + C);
-		int32_t move_cross = std::abs(A * (start.x + mx) + B * (start.y + my) + C);
+	return !(tile->hasProperty(CONST_PROP_BLOCKPROJECTILE));
```
nekiro, line 543: parentheses are redundant

src/map.cpp
```diff
@@ -529,62 +529,146 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+{
+	float dx = x1 - x0;
+	float dy = y1 - y0;
+	float grad = (dx == 0) ? 1 : dy / dx;
+	float xy = y0 + grad;
+
+	for (uint16_t y = x0 + 1; y < x1; ++y) {
+		uint16_t newX = std::floor(xy);
+		bool tileClear = isTileClear(newX, y, z);
+		if (!tileClear) {
```
nekiro, line 556: `!isTileClear(std::floor(xy), y, z);` ?

src/map.cpp
```diff
@@ -529,62 +529,146 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+{
+	float dx = x1 - x0;
+	float dy = y1 - y0;
+	float grad = (dx == 0) ? 1 : dy / dx;
+	float xy = y0 + grad;
+
+	for (uint16_t x = x0 + 1; x < x1; ++x) {
+		uint16_t newY = std::floor(xy + 0.1);
+		bool tileClear = isTileClear(x, newY, z);
+		if (!tileClear) {
```
nekiro, line 575: same as above

src/map.cpp
```diff
@@ -529,62 +529,146 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+
+		//no obstacles above floor 0 so we can throw above the obstacle
+		if (z0 == 0) {
+			return true;
+		}
+
+		//check if tiles above us and target are clear and check for clear sight between them
+		--z0;
+		return checkSightLine(x0, y0, x1, y1, z0) && isTileClear(x0, y0, z0, true) && isTileClear(x1, y1, z0, true);
+	} else {
```
nekiro, line 636: this else is pretty much redundant if we gonna return in first if anyway, so I would remove it and just continue the code

src/map.cpp
```diff
@@ -529,62 +529,146 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+
+		if (z0 > z1) {
+			//target is above us
+			if (Position::getDistanceZ(fromPos, toPos) > 1) {
+				return false;
+			}
+
+			//check a tile above us and the path to target
+			--z0;
+			return isTileClear(x0, y0, z0, true) && checkSightLine(x0, y0, x1, y1, z0);
```
nekiro, line 655: same thing, why else if you return

nekiro (changes requested):

ranisalt:
> @ranisalt your implementation fixes the upper line, but breaks the lower

:angry: :angry: :angry: :angry:

src/map.cpp
```diff
@@ -630,45 +625,42 @@ bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floo
...
 
-		if (z0 > z1) {
-			//target is above us
-			if (Position::getDistanceZ(fromPos, toPos) > 1) {
-				return false;
-			}
+	//skip checks for sight line in case fromPos and toPos cross the ground floor
+	if (z0 < 8 && z1 > 7 || z0 > 7 && z1 < 8) {
+		return false;
+	}
```
MillhioreBT: This mechanic should not be here, since this check is only done to throw items
What this function of verifying the line of sight gives me to understand is knowing exactly that, if the line of sight is clean

MillhioreBT:
btw, does this fix the behavior of monsters attacking from a distance, or that's a point to be covered separately in another PR?
What difference does this PR have with the previous one? I am seeing that for the tests you are using a different lua function than the one you use in the sources

src/map.cpp
Zbizu: If you call isSightClear with sameFloor = false from lua script, it doesn't go through Map::canThrowObjectTo. This prevents iterating over path when there is no reason to do it.

Zbizu:
The tests were made by adding magic effect to isTileClear in cpp file. The difference is different order of search in isSightClear and some fixes to it (your code searched same floor twice when z0 = z1).

nekiro (dismissed):
LGTM, from looking at the code. I did NOT test it.

src/map.cpp
```diff
@@ -529,62 +529,138 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		xy += grad;
+	}
 
-		if (start.x != destination.x && (start.y == destination.y || move_ver > move_hor || move_ver > move_cross)) {
-			start.x += mx;
-		}
+	return true;
+}
+
+bool Map::getSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const
```
ranisalt, line 563: Just for clarity I would name these `checkSlightLine` and `checkSteepLine` as you don't *get* them.

src/map.h
```diff
@@ -247,6 +247,16 @@ class Map
...
+		/**
+		  * Checks if there are no obstacles on that position
+		  *	\param blockFloor counts the ground tile as an obstacle
+		  *	\returns The result if there is an obstacle or not
+		  */
+		bool isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor = false) const;
+
+		bool getSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const;
+		bool getSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const;
+
```
ranisalt, line 259: I don't think these belong here. They can be in an anonymous namespace only in the .cpp file for Very Aggressive Optimization :tm:

src/map.cpp
```diff
@@ -529,62 +529,138 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+{
+	float dx = x1 - x0;
+	float dy = y1 - y0;
+	float grad = (dx == 0) ? 1 : dy / dx;
```
ranisalt, line 567:
```suggestion
	float slope = (dx == 0) ? 1 : dy / dx;
```

src/map.cpp
```diff
@@ -529,62 +529,138 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+	float xy = y0 + grad;
```
ranisalt, line 568:
```suggestion
	float yi = y0 + slope;
```

src/map.cpp
```diff
@@ -529,62 +529,138 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
 
-		const Tile* tile = getTile(start.x, start.y, start.z);
-		if (tile && tile->hasProperty(CONST_PROP_BLOCKPROJECTILE)) {
+	for (uint16_t x = x0 + 1; x < x1; ++x) {
+		if (!isTileClear(x, std::floor(xy + 0.1), z)) {
```
ranisalt, line 571:
```suggestion
		if (!isTileClear(x, std::floor(yi + 0.1), z)) {
```

Again the magic number surfaces. Did you test with 0.5 instead of 0.1? Or using `std::round` instead of `std::floor`?

src/map.cpp
```diff
@@ -529,62 +529,138 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+bool Map::getSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const
+{
+	float dx = x1 - x0;
+	float dy = y1 - y0;
+	float grad = (dx == 0) ? 1 : dy / dx;
+	float xy = y0 + grad;
 
-		if (start.y != destination.y && (start.x == destination.x || move_hor > move_ver || move_hor > move_cross)) {
-			start.y += my;
+	for (uint16_t y = x0 + 1; y < x1; ++y) {
```
ranisalt, line 553: I think there is a mistake here. This y = x does not smell good, am I missing something? Shouldn't it be `y = y0 + 1` instead?

src/map.cpp
```diff
@@ -529,62 +529,138 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool floorCheck /*= false*/) const
 {
-	if (floorCheck && fromPos.z != toPos.z) {
+	uint16_t x0 = fromPos.x;
+	uint16_t y0 = fromPos.y;
+	uint8_t  z0 = fromPos.z;
+
+	uint16_t x1 = toPos.x;
+	uint16_t y1 = toPos.y;
+	uint8_t  z1 = toPos.z;
```
ranisalt, line 608: Just use `fromPos.x`, `toPos.y` etc., the only aux values you need is `z0` as you modify it

src/map.cpp
```diff
@@ -529,62 +529,138 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool
...
+		}
+
+		//no obstacles above floor 0 so we can throw above the obstacle
+		if (z0 == 0) {
+			return true;
+		}
+
+		//check if tiles above us and the target are clear and check for a clear sight between them
+		--z0;
+		return checkSightLine(x0, y0, x1, y1, z0) && isTileClear(x0, y0, z0, true) && isTileClear(x1, y1, z0, true);
```
ranisalt, line 630:
```suggestion
		return isTileClear(x0, y0, z0, true) && isTileClear(x1, y1, z0, true) && checkSightLine(x0, y0, x1, y1, z0);
```

This way you immediately short circuit if the origin or destination are not clear, without checking the entire path.

You can also move the checking of the origin floor to the `checkSteep/SlightLine` functions if you start from `x0` and `y0` instead of `x0 + 1` and `y0 + 1`

ranisalt (changes requested):
I have some terminology suggestions and I think one of the functions is not working properly

src/map.h
Zbizu: I can't access `isTileClear`. It asks me for `Map` object.

src/map.cpp
Zbizu: coordinate swap goes there meaning that `x` will be used as `y`

after fixing, it's same for both:
`for (uint16_t x = x0 + 1; x < x1; ++x) {`

but the checks are:
`!isTileClear(x, std::floor(yi + 0.1), z)`
`!isTileClear(std::floor(yi), x, z)`

investigating the "magic numbers" now

src/map.cpp
Zbizu: with `std::round` (no "magic numbers") it behaves same as Bresenham's algo - the lines are completely different, vastly limiting our ability to throw diagonally near obstacles

with `std::floor` (no "magic numbers"), I don't experience the ["bug"](https://github.com/otland/forgottenserver/pull/3223#issuecomment-739238206) at all, but it might be compiler dependent

src/map.cpp
Zbizu: checking starting and ending spots will bug in the situation when both target and caster are on the same floor so I'll skip this one

src/map.h
nekiro: There should be no problem, it's direct call to another class

src/map.cpp
```diff
@@ -543,38 +543,38 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
...
-	float dx = x1 - x0;
-	float dy = y1 - y0;
-	float grad = (dx == 0) ? 1 : dy / dx;
-	float xy = y0 + grad;
-
-	for (uint16_t y = x0 + 1; y < x1; ++y) {
-		if (!isTileClear(std::floor(xy), y, z)) {
-			return false;
+namespace {
+	bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
```
ranisalt, line 547: No indentation and a line break here pls

src/map.cpp
```diff
@@ -543,38 +543,38 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
...
+			}
+			yi += slope;
 		}
-		xy += grad;
-	}
 
-	return true;
+		return true;
+	}
 }
```
ranisalt, line 578: Also a line break here

src/map.cpp
```diff
@@ -544,37 +544,39 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
...
 
-		return true;
+bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
+{
+	float dx = x1 - x0;
+	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
+	float yi = y0 + slope;
+
+	for (uint16_t x = x0 + 1; x < x1; ++x) {
+		if (!g_game.map.isTileClear(std::floor(yi), x, z)) {
```
ranisalt, line 555: This is still probably incorrect. You are using y in place of x. Is this the correct code?

Zbizu: Judging by amount of tests I performed throwing in every possible direction without any bugs, it is.

src/map.cpp
Zbizu: I found the cause of the rounding error. It's the loss of precition at large positions (32k~). This is why I wasn't able to reproduce it earlier.

obraz

Zbizu:
After some testing, new algorithm produces #3497 problem.

Here are all places at which the problem may occur:
```actions.cpp(315) :    ranged use/test with checkFloor = false also (not tested)
game.cpp  (1048) :    place hangable object (note: some walls don't support hangable objects, make sure you're using the right one) (not tested)
game.cpp(2600)   :    request trade (not tested)
game.cpp(2745)   :    accept trade (not tested)
monster.cpp(1035):pushing nearby objects (canPushItems is needed) (not tested)
spells.cpp(749)  :    throwing a rune spell (confirmed to occur)
spells.cpp(1030) :    throwing an instant spell (confirmed to occur)
spells.cpp(1031) :    throwing an instant spell (confirmed to occur)```

occurences I omitted:
```game.cpp(3571)   :    definition
game.cpp(3574)   :    definition
game.h(419)      :    declaration
map.cpp(502)     :    definition
map.h(247)       :    declaration```

Most of these problems can be resolved by `isSightClear` instead of `canThrowObjectTo` as the latter was meant to be used for item throwing and we need same floor sight line only.

**Comment by Zbizu** [comment]
- fixed previously mentioned problem
- adjusted failed trade messages
- reduced amount of checks in Map::canThrowObjectTo
- fixed item throwing for large Z differences

tests performed:
- requesting trade (between floors, between obstacles, clear path, obstacle present and clear +1 floor, walking away)
- orc spearmans attacking my summon from various angles
- spamming ue
- throwing a fire wall spell in all angles
- attempts to reproduce #3497 problem
- throwing gold coins on same floor, between floors (upper floor, lower floor)

Haven't got a single crash/bug during testing. Waiting for a review.

File: src/game.cpp [hunk 25]
```diff
@@ -2742,8 +2747,8 @@ void Game::playerAcceptTrade(uint32_t playerId)
 		return;
 	}
 
-	if (!canThrowObjectTo(tradePartner->getPosition(), player->getPosition())) {
-		player->sendCancelMessage(RETURNVALUE_CREATUREISNOTREACHABLE);
+	if (!canThrowObjectTo(tradePartner->getPosition(), player->getPosition()), true, true) {
```
nekiro, line 2750: you broke this. forgot to remove ")"?

Zbizu: autocompletion broke when I was putting the second boolean and both me and the compiler overlooked that 😅

nekiro (dismissed):
LGTM
I didn't test it in-game.

Zbizu:
made sure I'm on the right branch
compiled (commit name: correcting misplaced parenthesis)
made sure I'm launching the right executable
retook the tests (everything ok except trade - discovered a bug unrelated to my changes)

videos of tests:

ranged monster pathfinding:
https://user-images.githubusercontent.com/5205079/126899681-a45f62df-c894-49d2-8e19-9dd5e235f3d6.mp4
https://user-images.githubusercontent.com/5205079/126899682-085129ab-2f92-4cd8-a652-a07e109cf1a2.mp4

trade:
https://user-images.githubusercontent.com/5205079/126899705-7bb4a440-71a6-4146-82e6-715ac159b4d8.mp4
the bug:
- start a trade with clear sight line
- move to a place with obstacle between you
- you get cancel message, but the trade window doesn't disappear

throwing a firewall:
https://user-images.githubusercontent.com/5205079/126899824-8053d6bb-9807-40ee-8c70-f503e06e30d4.mp4

throwing exura sio:
https://user-images.githubusercontent.com/5205079/126899873-22d2f035-eca2-4c6d-bfe6-b53f00c5a8d9.mp4

throwing exori vis/sd/firewall:
https://user-images.githubusercontent.com/5205079/126899896-b3937c39-a24c-4cff-bcb0-6446056659ba.mp4

throwing gfb:
https://user-images.githubusercontent.com/5205079/126899952-407ef064-9ba5-46dc-afc8-93c9506e7531.mp4

throwing from large heights:
https://user-images.githubusercontent.com/5205079/126899981-fa284b6b-57f0-4896-bc61-f70849591c07.mp4

throwing to large heights:
https://user-images.githubusercontent.com/5205079/126899994-8eb02f79-6e17-470e-807b-0258da2f4716.mp4

throwing to same floor (with empty tiles on the path):
https://user-images.githubusercontent.com/5205079/126900008-cdd86548-95e2-4a49-854e-ec3d7e3c024f.mp4

throwing to lower floor (unreachable):
https://user-images.githubusercontent.com/5205079/126900027-e0fe15fc-ec0a-4167-a38e-22deccf5cb8b.mp4

throwing spells/items at Z = 0
https://user-images.githubusercontent.com/5205079/126900058-41ce5003-0034-4622-8085-846ec66ed56b.mp4

UE spam stress test:
https://user-images.githubusercontent.com/5205079/126900091-5247e397-361f-47a9-89de-ec00557c3af8.mp4

interaction with obstacles(imgur because long video):
https://i.imgur.com/esFtDdJ.mp4

rounding the values at various coordinates(imgur because long video):
https://i.imgur.com/I1X9xPG.mp4

nekiro:
Does the change have any impact on sight line code performance?

Zbizu:
Changes in this commit:
- fixed trade bug
- tweaked message types
- updates trade messages to most recent ones

ranisalt:
You should have done that in a separate pull request. Now it contains several unrelated changes which is not good for git history.

Zbizu:
Should I revert or you `git cherry-pick` that into two commits?

Znote:
> Should I revert or you `git cherry-pick` that into two commits?

You can do both, something like this:
```sh
# add otland as remote if you havent already
git remote add otland https://github.com/otland/forgottenserver.git

# grab fresh version of master branch, create a new PR branch from it
git fetch otland master
git checkout otland/master
git switch -c trade-bug-tweaks

# Find hash for trade bug fix and cherry pick it
git log -3 newSightClear
git cherry-pick 598f69584586fe9a4cce6418a8c9ee048193bae8
git push -u origin trade-bug-tweaks

# <enter github and submit PR>

# revert unrelated commit from sightclear and update PR
git checkout newSightClear
git revert 598f69584586fe9a4cce6418a8c9ee048193bae8
git push origin newSightClear
```

Zbizu:
trade tweaks moved to another PR

anything else before it's ready to merge?

Znote:
Ohh well, done is done. But from now on, please don't use force-push. (ever). 
Your breaking the dev environment for PR reviewers. You should have done a normal revert commit instead.

Zbizu:
Couldn't get the revert done otherwise. It just didn't show up on files changed.

ranisalt (approved):
Code is spotless, but I can't test the effect thoroughly. @MillhioreBT has done a lot of work on this, can you check if the line of sight is correct?

Zbizu:
resolved the conflict

DSpeichert (approved):

nekiro (approved):

Issue #3486, opened 2021-06-26
storage keys for accounts


Zbizu:
### Explanation of what you want to do that is currently impossible
player:getAccount():setStorageValue(key, value)
or in case creating account mt wasn't worth it:
player:setAccountStorage(k, v)

### Desired functionality
- store data tied to account rather than characters
- save when a character is logging out

### Available workarounds
sql queries

### Prior art
Imagine an event you can do with multiple chars because the progress is shared.

EPuncker:
done at https://github.com/otland/forgottenserver/commit/49976cf151fb07e5c89282941c34da85055c54e7

PR #3490, opened 2021-06-27
feat: add account-level storage values

amatria:
### Pull Request Prelude

### Changes Proposed
Add account-level storage values:
```lua
Game.setAccountStorageValue(accountId, key, value)
Game.getAccountStorageValue(accountId, key)
-- force a save
Game.saveAccountStorageValues()
```

**Issues addressed:** 
 #3486

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -96,6 +96,14 @@ CREATE TABLE IF NOT EXISTS `account_ban_history` (
   FOREIGN KEY (`banned_by`) REFERENCES `players` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
+CREATE TABLE IF NOT EXISTS `account_storage` (
+	`account_id` int NOT NULL,
+	`key` int unsigned NOT NULL,
+	`value` int NOT NULL,
+	PRIMARY KEY (`account_id`, `key`),
+	FOREIGN KEY (`account_id`) REFERENCES `accounts`(`id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
+
 CREATE TABLE IF NOT EXISTS `ip_bans` (
   `ip` int unsigned NOT NULL,
   `reason` varchar(255) NOT NULL,
@@ -349,7 +357,7 @@ CREATE TABLE IF NOT EXISTS `towns` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
-INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '28'), ('motd_hash', ''), ('motd_num', '0'), ('players_record', '0');
+INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '29'), ('motd_hash', ''), ('motd_num', '0'), ('players_record', '0');
 
 DROP TRIGGER IF EXISTS `ondelete_players`;
 DROP TRIGGER IF EXISTS `oncreate_guilds`;
--- a/schema.sql
+++ b/schema.sql
@@ -97,11 +97,11 @@ CREATE TABLE IF NOT EXISTS `account_ban_history` (
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
 CREATE TABLE IF NOT EXISTS `account_storage` (
-	`account_id` int NOT NULL,
-	`key` int unsigned NOT NULL,
-	`value` int NOT NULL,
-	PRIMARY KEY (`account_id`, `key`),
-	FOREIGN KEY (`account_id`) REFERENCES `accounts`(`id`) ON DELETE CASCADE
+  `account_id` int NOT NULL,
+  `key` int unsigned NOT NULL,
+  `value` int NOT NULL,
+  PRIMARY KEY (`account_id`, `key`),
+  FOREIGN KEY (`account_id`) REFERENCES `accounts`(`id`) ON DELETE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
 CREATE TABLE IF NOT EXISTS `ip_bans` (
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4417,31 +4417,21 @@ void Game::setAccountStorageValue(const uint32_t accountId, const uint32_t key, 
 	accountStorageMap[accountId][key] = value;
 }
 
-bool Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key, int32_t& value) const
+uint32_t Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key) const
 {
 	const auto& accountMapIt = accountStorageMap.find(accountId);
 	if (accountMapIt != accountStorageMap.end()) {
 		const auto& storageMapIt = accountMapIt->second.find(key);
 		if (storageMapIt != accountMapIt->second.end()) {
-			value = storageMapIt->second;
-			return true;
+			return storageMapIt->second;
 		}
 	}
-
-	value = -1;
-	return false;
+	return -1;
 }
 
 size_t Game::getNumberOfPlayersByAccount(const uint32_t accountId) const
 {
-	size_t ret = 0;
-	for (const auto& it : players) {
-		if (it.second->getAccount() == accountId) {
-			ret++;
-		}
-	}
-
-	return ret;
+	return std::count_if(players.begin(), players.end(), [=](const auto& it) { return it.second->getAccount() == accountId; });
 }
 
 void Game::startDecay(Item* item)
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4407,6 +4407,45 @@ void Game::addDistanceEffect(const SpectatorVec& spectators, const Position& fro
 	}
 }
 
+void Game::setAccountStorageValue(const uint32_t accountId, const uint32_t key, const int32_t value)
+{
+	if (value == -1) {
+		accountStorageMap[accountId].erase(key);
+		return;
+	}
+
+	accountStorageMap[accountId][key] = value;
+}
+
+bool Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key, int32_t& value) const
+{
+	const auto& accountMapIt = accountStorageMap.find(accountId);
+
+	if (accountMapIt != accountStorageMap.end()) {
+		const auto& storageMapIt = accountMapIt->second.find(key);
+		if (storageMapIt != accountMapIt->second.end()) {
+			value = storageMapIt->second;
+			return true;
+		}
+	}
+
+	value = -1;
+	return false;
+}
+
+size_t Game::getNumberOfPlayersByAccount(const uint32_t accountId) const
+{
+	size_t ret = 0;
+
+	for (const auto& it : players) {
+		if (it.second->getAccount() == accountId) {
+			ret++;
+		}
+	}
+
+	return ret;
+}
+
 void Game::startDecay(Item* item)
 {
 	if (!item || !item->canDecay()) {
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4420,7 +4420,6 @@ void Game::setAccountStorageValue(const uint32_t accountId, const uint32_t key, 
 bool Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key, int32_t& value) const
 {
 	const auto& accountMapIt = accountStorageMap.find(accountId);
-
 	if (accountMapIt != accountStorageMap.end()) {
 		const auto& storageMapIt = accountMapIt->second.find(key);
 		if (storageMapIt != accountMapIt->second.end()) {
@@ -4436,7 +4435,6 @@ bool Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key, 
 size_t Game::getNumberOfPlayersByAccount(const uint32_t accountId) const
 {
 	size_t ret = 0;
-
 	for (const auto& it : players) {
 		if (it.second->getAccount() == accountId) {
 			ret++;
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -127,6 +127,7 @@ void Game::setGameState(GameState_t newState)
 
 			loadMotdNum();
 			loadPlayersRecord();
+			loadAccountStorageValues();
 
 			g_globalEvents->startup();
 			break;
@@ -182,6 +183,10 @@ void Game::saveGameState()
 	}
 
 	std::cout << "Saving server..." << std::endl;
+
+	if (!saveAccountStorageValues()) {
+		std::cout << "[Error - Game::saveGameState] Failed to save account-level storage values." << std::endl;
+	}
 
 	for (const auto& it : players) {
 		it.second->loginPosition = it.second->getPosition();
@@ -4417,7 +4422,7 @@ void Game::setAccountStorageValue(const uint32_t accountId, const uint32_t key, 
 	accountStorageMap[accountId][key] = value;
 }
 
-uint32_t Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key) const
+int32_t Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key) const
 {
 	const auto& accountMapIt = accountStorageMap.find(accountId);
 	if (accountMapIt != accountStorageMap.end()) {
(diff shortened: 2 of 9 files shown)


MillhioreBT:
It doesn't sound like a bad idea, I really like it a lot, I'm going to think about it a bit, and check your code ;)

omarcopires:
What happens if I'm logged in with two users on the same account?

MillhioreBT:
You'd better load the storages into an AccountStorageMap cache.

* We load the storages only if the player who enters is the first character of the account who enters.
* We save the storages only if the player who disconnects is the only character on the online account.

schema.sql
```diff
@@ -96,6 +96,14 @@ CREATE TABLE IF NOT EXISTS `account_ban_history` (
   FOREIGN KEY (`banned_by`) REFERENCES `players` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
+CREATE TABLE IF NOT EXISTS `account_storages` (
```
marmichalski, line 99: Please update db_version on line 360.

src/iologindata.cpp
```diff
@@ -44,6 +44,32 @@ Account IOLoginData::loadAccount(uint32_t accno)
...
+	if (value != -1) {
+		query = fmt::format("INSERT INTO `account_storages` (`account_id`, `key`, `value`) VALUES ({:d}, {:d}, {:d}) ON DUPLICATE KEY UPDATE `value` = {:d}", accno, key, value, value);
+	} else {
+		query = fmt::format("DELETE FROM `account_storages` WHERE `account_id` = {:d} AND `key` = {:d}", accno, key);
+	}
+
+	return Database::getInstance().executeQuery(query);
+}
+
+bool IOLoginData::getAccountStorageValue(const uint32_t accno, const uint32_t key, int32_t& value)
```
marmichalski, line 60: I don't think that returning the value by referenced parameter is good. Return value does not have to be a boolean at all, you can distinguish missing storage (-1) from set storage (>= 0) easily.

Anyway, this is probably true for players storage too, I don't see how mixing int32 with uint32 for value is acceptable.

Zbizu:
Can't review from a phone, but basically:

account_storage (not storages) as database table for consistency with player_storage

src/iologindata.cpp
amatria: I agree with you indeed. Do you think it is a good idea I change the `player:setStorageValue` implementation too within this PR?

marmichalski: I'd consider changing return type a breaking change and I have no idea what the policy here is (if any). This should probably be done in [2.0](https://github.com/otland/forgottenserver/projects/3).

amatria:
So, as @MillhioreBT said. The proposed solution would be as follows:
1. Create a `AccountStorageRepository` class. This class will be in charge of storing a `storageMap` for each account logged in into the game. Also, it will store information about how many players from the same account are actually logged in into the game.
2. Once a player logs in into the game, it will push into the repository its account storage values.
3. If another player from the very same account logs in too, the repository will give him a reference to the already existing `storageMap`.
4. Players will now be able to consistently modify the account storages values.
5. Once all players from one account have logged out of the game, it will remove the shared reference from the repository and proceed to store it into the database.
Thoughts? Does it even make any sense to you? :p

Zbizu:
it could be like this:

> map: accountId to account struct/object
> 
> account struct/object:
> - id
> - playersActive
> - storages(map: key to value)
> 
> login: map[accountId] exists and playersActive +1 or create map[accountId]
> logout: playersActive -1, save to database, if playersActive 0 then destroy

though I'm not sure how stable will that be when someone decides to spam relogs
on the other hand keeping the table in memory for entire server uptime might be ram consuming so I suggest adding lastUpdated field that would be helpful to determine whether the object can still be used soon or could be destroyed safely (and not destroying on logout).

MillhioreBT:
I think it is better to have a map in the *Game class* and use this map to store the storages

<details>
<summary>iologindata.cpp</summary>

* load
```c++
//load account storage map
if (g_game.getPlayersByAccount(accountId).empty()) {
	if (result = db.storeQuery(fmt::format("SELECT `key`, `value` FROM `account_storage` WHERE `account_id` = {:d}", accountId))) {
		do {
			g_game.setAccountStorageValue(accountId, result->getNumber("key"), result->getNumber("value"));
		} while (result->next());
	}
}
```

* save
```c++
//save account storage map
if (g_game.getPlayersByAccount(accountId).size() == 1) {
	if (!db.executeQuery(fmt::format("DELETE FROM `account_storage` WHERE `account_id` = {:d}", accountId))) {
		return false;
	}

	const auto& accountStorageMap = g_game.accountStorageMap.find(accountId);
	if (accountStorageMap == g_game.accountStorageMap.end()) {
		return false;
	}

	query.str(std::string());
	DBInsert accountStorageQuery("INSERT INTO `account_storage` (`account_id`, `key`, `value`) VALUES ");
	for (const auto& it : accountStorageMap->second) {
		query << accountId << ',' << it.first << ',' << it.second;
		if (!accountStorageQuery.addRow(query)) {
			return false;
		}
	}

	if (!accountStorageQuery.execute()) {
		return false;
	}

	g_game.accountStorageMap.erase(accountId);
}
```
</details>

<details>
<summary>game.cpp</summary>

```c++
void Game::setAccountStorageValue(uint32_t accountId, const std::string key, const std::string& value)
{
	accountStorageMap[accountId][key] = value;
}

bool Game::getAccountStorageValue(uint32_t accountId, const std::string key, std::string& value) const
{
	auto accountMapIt = accountStorageMap.find(accountId);
	if (accountMapIt != accountStorageMap.end()) {
		auto storageMapIt = accountMapIt->second.find(key);
		if (storageMapIt == accountMapIt->second.end()) {
			value = "-1";
			return false;
		}

		value = storageMapIt->second;
		return true;
	}

	return false;
}
```
</details>

<details>
<summary>game.h</summary>

```h
// public:
std::unordered_map<uint32_t, StorageMap> accountStorageMap;
```

</details>

amatria:
@MillhioreBT
And therefore change the Lua script interface methods to
```lua
Game.setAccountStorageValue(accId, key, value)
Game.getAccountStorageValue(accId, key)
```
right?

amatria:
I have just squashed all the WIP commits and the PR should now be ready for a comprehensive review.

src/game.cpp
```diff
@@ -4407,6 +4407,45 @@ void Game::addDistanceEffect(const SpectatorVec& spectators, const Position& fro
...
+		return;
+	}
+
+	accountStorageMap[accountId][key] = value;
+}
+
+bool Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key, int32_t& value) const
+{
+	const auto& accountMapIt = accountStorageMap.find(accountId);
+
```
nekiro, line 4423: remove new line

src/game.cpp
```diff
@@ -4407,6 +4407,45 @@ void Game::addDistanceEffect(const SpectatorVec& spectators, const Position& fro
...
+	}
+
+	value = -1;
+	return false;
+}
+
+size_t Game::getNumberOfPlayersByAccount(const uint32_t accountId) const
+{
+	size_t ret = 0;
+
```
nekiro, line 4439: remove new line

src/iologindata.cpp
```diff
@@ -536,6 +536,15 @@ bool IOLoginData::loadPlayer(Player* player, DBResult_ptr result)
 		} while (result->next());
 	}
 
+	// load account storage map
+	if (g_game.getNumberOfPlayersByAccount(player->getAccount()) == 0) {
```
nekiro, line 540: wouldn't find by account id be enough here? Why checking the players count at all?

src/iologindata.cpp
```diff
@@ -832,6 +841,31 @@ bool IOLoginData::savePlayer(Player* player)
 		return false;
 	}
 
+	// save account storage values
+	if (g_game.getNumberOfPlayersByAccount(player->getAccount()) == 1) {
```
nekiro, line 845: same here, just check if account id key exist and save it

nekiro (changes requested):

src/iologindata.cpp
amatria: Indeed, it works when you load the account storage map for the first time. You call the `g_game.getPlayerByAccount(player->getAccount())` and you get a `nullptr` value. However, to be able to save the account storages only once, you need to know how many players are online from the very same account. In this case, the `g_game.getPlayerByAccount(player->getAccount())` call won't help you at all. You will just receive a reference to a player (which, in fact, can be the already logged out player). But, are there more online?

amatria: Also, `g_game.getPlayerByAccount(player->getAccount())` is an O(n) function too [1]. So, I don't see any big improvement over using the new function either.

===
[1] Worst case scenario.

src/iologindata.cpp
MillhioreBT: What we want is to avoid saving the storage map several times, since we can save it only once, when in fact all the players of the same account have disconnected, otherwise every time a player disconnects we will be saving the storages, I suppose this is implemented like this in case several characters of the same account can connect at the same time

MillhioreBT:
The idea is very good, and thanks for following my suggestions <3 It only remains to wait for the suggestions of the other developers, to know if they have better ideas, or perhaps a way to relocate the functions or their names, but in my opinion this is functional, it only remains to keep testing and wait!

schema.sql
```diff
@@ -96,6 +96,14 @@ CREATE TABLE IF NOT EXISTS `account_ban_history` (
   FOREIGN KEY (`banned_by`) REFERENCES `players` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
+CREATE TABLE IF NOT EXISTS `account_storage` (
+	`account_id` int NOT NULL,
```
EPuncker, line 100: indentation in sql file is made using spaces

marmichalski:
I do not believe that those "micro-optimizations" are of any good. There are a lot of other places that executes bigger sql queries instantly. MySQL is not a bottleneck here, those are simple insert and select statements and current solution introduces synchronization problems when you'd like to use account_storage in, for example, your web app. Saving account_storage when there is only one player online per-account is IMHO also wrong, consider the web app sync + server allowing MC on the same account scenario.

tl;dr; no-cache should be the way to go IMHO, account storage was implemented in many servers as instant queries and generated no problems, one of those was Classicus with hundreds of players.

amatria:
>I do not believe that those "micro-optimizations" are of any good

I would not consider them "micro-optimizations". Indeed, I would say that is a rather bold claim. I know MySQL implements caching mechanisms to optimize fetching and pulling data in and out of the relations. However, is it even as fast as accessing a hash table in user space? Also, what if the game server was sitting in a DMZ and the database was running behind a firewall to prevent attacks? Would you still consider this solution a "micro-optimization". It may sound bizarre, but it is a completely feasible scenario.

Nevertheless, I get your point. I liked the queries approach too. It was dead simple, clean and effective. So... I guess we have to see what the other developers think about the matter.

>current solution introduces synchronization problems when you'd like to use account_storage in, for example, your web app

And so it does for player storages. Doesn't it?

marmichalski:
Sure it's not that fast, but does it have to be? It's fast enough. 

Check iomarket. Yes, I'd still consider this micro optimizations compared to other "instant" queries. I'm ignoring the DMZ part, as probably many people will have to google what that even means 😉

Player storages are at least saved with each player save.

amatria:
>Sure it's not that fast, but does it have to be? It's fast enough.

Gotchu :p. To be honest, I'm with you with the queries approach (and there's no sync problem), even though I like the other solution a lot too. So, well, let's see what the others have to say. 

>Player storages are at least saved with each player save.

Yeah, they are. But they also follow the same DELETE and INSERT pattern as in this solution. Therefore, if a value is stored from outside of the C++ code, it will overwrite/get rid of the result stored directly into the database.

Zbizu:
A possible sync system could be an agreement between aac developers and us.

The aac can have setAccountStorage method in PHP, a method that would queue the storages for updating and the server could pick, let's say, 20 results every 15 seconds. This solution is already used for website shop systems.

marmichalski:
HI @amatria, I'm back after a bad sleep 😎
I did not mean to sound hostile, so if I did, I apologise.

> 
> > Player storages are at least saved with each player save.
> 
> Yeah, they are. But they also follow the same DELETE and INSERT pattern as in this solution. Therefore, if a value is stored from outside of the C++ code, it will overwrite/get rid of the result stored directly into the database.

That's a side-effect I did not even think about. Imagine web-shop account-wide offer, think outfits. Account storage could be leveraged for that.  
But that's now what my original thought of "desync" was about. There are servers that never restart (Cyntara?), allowing multiple characters to be online on the same account (Cyntara?) and AFK-botting/training/whatever (Cyntara?). Now, let's have consider a situation, where somebody is "abusing" these things and is online for hundreds of hours - his account storage never persists into db.

I know, this might be super-rare situation, so let me introduce another one: saving server in gdb after crash.
Sometimes (always?) you want to call g_game.saveGameState after a crash and (most of the times?) it works. Now, there is more than 1 character online per many accounts, none of those accounts storage will persist.

So, consider the following changes:
* Loading storage:
  * Always try to load account storage, if the entry is already set, no-op
* Saving storage:
  * Perhaps introduce "dirty" state for AccountStorage and make every change to that (new?) object marking it as dirty
  * Save when dirty, reset dirty flag, therefore it will be saved only once, even if there are 100 characters online from that account
* [Optionally] If memory is so precious, cleanup the AccountStorage object on-logout (or player destructor?)

What does it solve? Iterating all online players when loading/saving one. Preventing storage being not loaded/saved in certain scenarios.

- Also, you are erasing the storages on save - but save does not mean that player logs out 🤔 
- "Virtual" players are not supported
- dunno, did not review whole code 🍨 

Have a nice day.

src/game.cpp
```diff
@@ -4407,6 +4407,43 @@ void Game::addDistanceEffect(const SpectatorVec& spectators, const Position& fro
...
+size_t Game::getNumberOfPlayersByAccount(const uint32_t accountId) const
+{
+	size_t ret = 0;
+	for (const auto& it : players) {
+		if (it.second->getAccount() == accountId) {
+			ret++;
+		}
+	}
+
+	return ret;
```
ranisalt, line 4444: Idea:

```
	return std::count_if(players.begin(), players.end(), [=](auto& player) { return player->getAccount() == accountId });
```

src/game.cpp
```diff
@@ -4407,6 +4407,43 @@ void Game::addDistanceEffect(const SpectatorVec& spectators, const Position& fro
...
+{
+	if (value == -1) {
+		accountStorageMap[accountId].erase(key);
+		return;
+	}
+
+	accountStorageMap[accountId][key] = value;
+}
+
+bool Game::getAccountStorageValue(const uint32_t accountId, const uint32_t key, int32_t& value) const
```
ranisalt, line 4420: Return a `pair<bool, int32_t>` or just return `-1` as it was defined to be the invalid value

ranisalt (changes requested):

src/iologindata.cpp
ranisalt: > otherwise every time a player disconnects we will be saving the storages

Is this a problem? If the server crashes, at least *some* of the values will be updated from the session. It's a write-through cache of sorts...

src/iologindata.cpp
ranisalt: >  [1] Worst case scenario.

It is actually always linear, but the problem persists. Is there nothing on a player object that tags it is online or not?

src/game.cpp
MillhioreBT: yeah, better code ^ thanks Rani you always beautiful :p

src/iologindata.cpp
amatria: Okay. I have to agree with you. We can get rid of this function and just go like:
1. Load only once using `Game::getPlayerByAccount(uint32_t accountId)` [1].
2. Always store the values.

===
[1] Checking if a player is online is of no use.

amatria: Okay. I like the write-through idea. However, we can't get rid of this function just yet. We need it in order to delete the account storage map object, once all players have logged out of the account.

To sum up, I believe the final solution would be as follows (I'll update the branch right away):
1. Load the values either using the new function or `Game::getPlayerByAccount(uint32_t acc) == nullptr`. And actually, I'll go with the second option. It is not exactly an O(n) operation when there are one or more players from the same account logged in into the game.
2. Always save the storages.
3. Delete the account storage map reference using the `Game::getNumPlayersByAccount(uint32_t acc)` method.

amatria:
@marmichalski 
>I did not mean to sound hostile, so if I did, I apologise.

No, you didn't. No worries :p.

As for the topic. I believe that implementing a dirty state is far too complex for this use case. Also, with the latest commit this problem is long gone. We simply save the values every time a player logs out. And we needn't to implement a dirty cache. I appreciate your extensive response though. Time is valuable. Not that many people would write such a comprehensive and educated comment.

Take very good care of yourself ❤️.

src/iologindata.cpp
```diff
@@ -832,6 +841,31 @@ bool IOLoginData::savePlayer(Player* player)
...
+			return false;
+		}
+	}
+
+	if (!accountStorageQuery.execute()) {
+		return false;
+	}
+
+	if (g_game.getNumberOfPlayersByAccount(player->getAccount()) == 1) {
+		g_game.accountStorageMap.erase(player->getAccount());
```
marmichalski, line 866: This is not a place to do it, as I have mentioned in my other comment. Calling this method does not mean that player is logging out. You are erasing the storage with every save of that player (player:save(), saveServer()).

amatria: True. I missed that one (twice, lmao).

nekiro:
Let's stop microoptimizing and just make storages save in game class, load it all once on startup and then just save it all once on shutdown. No problems with keepin track of anything and it works just like that.

MillhioreBT:
I was thinking that this should not be implemented, as it can be done perfectly through pure Lua!
However, @nekiro idea is good, although I would like them to be global storages and there we would store the account storages
The current global storage methods do not make sense XD, nobody uses them because they are useless `Game.setStorageValue / Game.getStorageValue`, these methods should be in the sources and have a real storage in the database as before, in this way you can do something like what I did on my server:

```lua
function Player:setAccountStorageValue(key, value)
	return Game.setStorageValue(("AccountStorage%s:%s"):format(self:getAccountId(), key), value)
end

function Player:getAccountStorageValue(key)
	return Game.getStorageValue(("AccountStorage%s:%s"):format(self:getAccountId(), key))
end
```

Zbizu:
@MillhioreBT tbh the longer this discussion goes on, the more I agree with you regarding writing that purely in lua

amatria:
Well, I just pushed the changes nekiro proposed. However, if you want to change the whole implementation to pure Lua, I'd rather close this PR and open a new one. So... It's up to you. I like how this one turned out eventually.

MillhioreBT:
> Well, I just pushed the changes nekiro proposed. However, if you want to change the whole implementation to pure Lua, I'd rather close this PR and open a new one. So... It's up to you. I like how this one turned out eventually.

Let someone else review this, and it depends on what the PRO's say, you make the decision to close the PR and do something simpler in lua, or just do something better, and I apologize for writing so much and extending this discussion so much

amatria:
What's good people 🌚! It's been 10 days since I last updated this PR. I understand it's summer and we all are busy doing RL stuff etc. However, I would appreciate if further action was taken. Are we closing this PR and implementing the changes in Lua or approving it instead?

Have a lovely weekend :p.

DSpeichert (approved):

nekiro (approved):
LGTM

src/game.cpp
```diff
@@ -4407,6 +4412,73 @@ void Game::addDistanceEffect(const SpectatorVec& spectators, const Position& fro
...
+		return false;
+	}
+
+	if (!db.executeQuery("DELETE FROM `account_storage`")) {
+		return false;
+	}
+
+	for (const auto& accountIt : g_game.accountStorageMap) {
+		if (accountIt.second.empty()) {
+			break;
```
MillhioreBT, line 4464: If you break the loop here, then the other accounts will not be saved, you should use continue

PR #3493, opened 2021-07-01
fix spells interaction with walls

Zbizu:
### Pull Request Prelude

### Changes Proposed

- remove the check for toPos for instant spells
- move the centerPos for isSightClear checks to avoid casting waves through walls

**Issues addressed:** #2891

~~Note: Not tested yet, side effects unknown.~~


Zbizu:
tested, doesn't fix the issue

the problem is that check for instant spells calls canDoCombat
https://github.com/otland/forgottenserver/blob/aa6d3584bf6a6327e65d2e3519c4fdd26672ef53/src/spells.cpp#L699 

which checks for blockprojectile
https://github.com/otland/forgottenserver/blob/aa6d3584bf6a6327e65d2e3519c4fdd26672ef53/src/combat.cpp#L280

but in order to remove it, I need to make sure that onAreaCombat gets executed somehow

Zbizu:
removing the check for possibility to cast the spell made it also possible to cast the spell through wall because isSightClear was checking the tile in front of the player instead of the player himself so I fixed that as well

MillhioreBT:
I'm a little concerned about `Combat :: canDoCombat` why should it be removed from here? Doesn't this prevent the spell from doing unnecessary checks?

At first glance it seems a simple solution, but not very pretty

Zbizu:
@MillhioreBT canDoCombat is checking the first tile in front of the caster. Removing that check from canDoCombat can bug some other function that uses it. Removing canDoCombat from this place doesn't do anything as other checks were perfomed earlier and pz check will be performed when checking isAggressive parameter.

Zbizu:
Tested my changes on a build from November 2020 now, works fine with diagonal fire/poison/energy wall runes.

nekiro:
What if you throw field rune at stairs or so, wont it go down?

Zbizu:
> What if you throw field rune at stairs (floorchange item) or so, wont it go down/up?

not enough room error
rune doesn't disappear
rl: fires shooteffect, tfs: doesn't
both send poff

Zbizu:
testing update:
spells can be fired to any place that doesn't block projectile, this includes fences and windows

field spells fire shoot effect, but poff, return "there is not enough room" and don't consume the rune when they can't be placed (eg. when you click on opened window)

area spells can be thrown to any tile that doesn't block projectile, the charge is consumed regardless of spell area

going to look into the code later

Zbizu:
Looks like spells and isSightClear are directly tied. This limits my ability to work on them separately. I'll give that a closer look when working with #3487.

Zbizu:
Delayed until new isSightClear gets merged. Files conflict very likely so I'll rebase later.

src/combat.cpp
```diff
@@ -66,8 +66,11 @@ MatrixArea createArea(const std::vector<uint32_t>& vec, uint32_t rows)
 	return area;
 }
 
-std::vector<Tile*> getList(const MatrixArea& area, const Position& targetPos)
+std::vector<Tile*> getList(const MatrixArea& area, const Position& targetPos, const Direction dir)
 {
+	Position casterPos(targetPos.x, targetPos.y, targetPos.z);
+	casterPos = getNextPosition(dir, casterPos);
```
ranisalt, line 72:
```suggestion
	auto casterPos = getNextPosition(dir, targetPos);
```

ranisalt (changes requested):

src/combat.cpp
Zbizu: why `auto` and not `const Position`?

ranisalt: It's personal, but it's shorter and just as clear. `getNextPosition` should return a `Position`, so why write it twice?

EPuncker (approved):

Zbizu:
Something is stuck in github mergability verification. I addressed the requested changes already.

ranisalt (approved):

ranisalt:
> Something is stuck in github mergability verification

My blessing was required, Github doesn't automatically dismiss change requests

PR #3497, opened 2021-07-03
fix targeted spells bug

Zbizu:
### Pull Request Prelude

### Steps to reproduce
1. go to some place on the surface
2. spawn some wall with /i
3. put some creature on the other side
4. cast a missile rune or a targeted strike spell

### Expected Result
spell cancelled

### Actual Result
spell goes through wall when there are no obstacles above because it's treated like item throwing

### Changes Proposed
stop treating throwing spells like throwing items

**Issues addressed:** see **steps to reproduce**
### Reproduction environment
This bug was discovered while working with isSightClear, but is not related to map.cpp so no conflict will arise between these pulls.
This bug can be reproduced with any build. It's also present in my build from November 2020 and was there for unknown amount of time so you can reproduce it with your own tfs builds as well.

### DIFF
--- a/src/spells.cpp
+++ b/src/spells.cpp
@@ -746,7 +746,13 @@ bool Spell::playerRuneSpellCheck(Player* player, const Position& toPos)
 		return false;
 	}
 
-	if (range != -1 && !g_game.canThrowObjectTo(playerPos, toPos, true, range, range)) {
+	if (range != -1 && (Position::getDistanceX(playerPos, toPos) > range || Position::getDistanceY(playerPos, toPos) > range)) {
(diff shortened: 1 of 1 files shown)


MillhioreBT:
Is this really happening? I was so curious that I wanted to try it, and it doesn't seem to work the way you describe it here, at least for me. I have always seen that the projectiles respect the walls and the line of sight

Zbizu:
I'll record a video when I'm home.

Zbizu:
Weird thing, I'm not able to reproduce that bug anymore. I'll continue testing my other PRs to see what might be the cause.

Zbizu:
My bad, had executable from another branch running.

PR #3499, opened 2021-07-10
RevScriptSys Monster enhanced attack conditions

### LINKED ISSUE #3485: Revscriptsys Monster
Well, I don't know if this is supposed to happen, but, when I try to register a monster using revscript and register a scripted spell such as sudden death rune, rage of the skies, energy strike, it doesn't work: it causes no damage.

### Steps to reproduce
1. Register a monster using revscriptsys
2. Set up the monster to use a scripted spell (sudden death rune or rage of the skies, for example)
3. Wait until the monster cast it, you'll notice that it'll cause 0 damage

### Expected behaviour
Cause the damage registered as minDamage and maxDamage

### Actual behaviour
The monster is causing 0 damage

### Environment
Windows 10

soul4soul:
### Pull Request Prelude

### Changes Proposed
**Issues addressed:** 

- Fix #3485 
- Refactor parsing revscriptsys monster attack and defense to use a common helper function
- Allow conditions to be used with all predefine creature abilities. Conditions can be used with "melee", "combat", "speed", "condition", "outfit", "invisible", "*field", etc.. Conditions won't work with named spells such as "energy strike".

I verified conditions worked using these attack types. 
```
monster.attacks = {
	{name = "melee", attack = 10, skill = 100, condition = { type = CONDITION_FIRE, minDamage = 50, maxDamage = 200, duration = 5000, interval = 1000}},
        {name = "condition", chance = 1000, range = 1, condition = { type = CONDITION_ENERGY, minDamage = 10, maxDamage = 20, duration = 5000, interval = 1000}},
        {name = "combat", type = COMBAT_LIFEDRAIN, chance = 100, interval = 2*1000, minDamage = 0, maxDamage = -10, target = true, range = 7, condition = { type = CONDITION_CURSED, minDamage = 10, maxDamage = 20, duration = 5000, interval = 1000}},
        {name = "speed", chance = 100, interval = 2000, speed = -500, target = true, range = 7, condition = { type = CONDITION_DROWN, minDamage = 20, maxDamage = 20, duration = 5000, interval = 1000}},
        {name = "firefield", chance = 100, interval = 2000, target = true, range = 7, condition = { type = CONDITION_FREEZING, minDamage = 20, maxDamage = 50, duration = 5000, interval = 1000}}
}
```

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -277,32 +277,10 @@ function AbilityTableToSpell(ability)
 			if ability.effect then
 				spell:setCombatEffect(ability.effect)
 			end
-			if ability.condition then
-				if ability.condition.type then
-					spell:setConditionType(ability.condition.type)
-				end
-				local startDamage = 0
-				if ability.condition.startDamage then
-					startDamage = ability.condition.startDamage
-				end
-				if ability.condition.minDamage and ability.condition.maxDamage then
-					spell:setConditionDamage(ability.condition.minDamage, ability.condition.maxDamage, startDamage)
-				end
-				if ability.condition.duration then
-					spell:setConditionDuration(ability.condition.duration)
-				end
-				if ability.condition.interval then
-					spell:setConditionTickInterval(ability.condition.interval)
-				end
-			end
 		else
 			spell:setType(ability.name)
 			if ability.type then
-				if ability.name == "condition" then
-					spell:setConditionType(ability.type)
-				else
-					spell:setCombatType(ability.type)
-				end
+				spell:setCombatType(ability.type)
 			end
 			if ability.interval then
 				spell:setInterval(ability.interval)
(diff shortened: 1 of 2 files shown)


data/scripts/lib/register_monster_type.lua
```diff
@@ -394,120 +277,119 @@ registerMonsterType.defenses = function(mtype, mask)
...
-					end
-					if defense.direction then
-						spell:setNeedDirection(defense.direction)
+				local spell = AbilityTableToSpell(defense)
+				mtype:addDefense(spell)
+			end
+		end
+	end
+end
+function AbilityTableToSpell(ability)
```
EPuncker, line 286: shouldn't this be added a few lines above before it is used?

MillhioreBT: The libraries are loaded first, so there is no problem with the order of declaration, however it does not seem intuitive and it makes more sense to have it declared before it is used.

soul4soul: Didn't even think about it I'm so use to languages which don't care. Moved the function above the usage.

data/scripts/lib/register_monster_type.lua
```diff
@@ -259,127 +259,120 @@ registerMonsterType.immunities = function(mtype, mask)
...
+			if ability.name == "drunk" then
+				spell:setConditionType(CONDITION_DRUNK)
+				if ability.drunkenness then
+					spell:setConditionDrunkenness(ability.drunkenness)
 				end
 			end
+		end
+		if ability.condition then
+			if ability.condition.type then
+				print(ability.condition.type)
```
MillhioreBT, line 335: and this?

MillhioreBT (changes requested):

data/scripts/lib/register_monster_type.lua
MillhioreBT: and why not make the `AbilityTableToSpell` function local? from what I see, it will only be used in this file

EPuncker (approved):

EPuncker (approved):

PR #3513, opened 2021-07-28
Make removeCondition support userdata

nekiro:
This pr makes the creature:removeCondition being able to take condition userdata as the first parameter.
Additionaly it hides condition:delete from lua, because calling this on added condition to creature produces a crash.
condition:delete isn't really needed as __gc will collect our conditions anyway.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2725,7 +2725,6 @@ void LuaScriptInterface::registerFunctions()
 	registerClass("Condition", "", LuaScriptInterface::luaConditionCreate);
 	registerMetaMethod("Condition", "__eq", LuaScriptInterface::luaUserdataCompare);
 	registerMetaMethod("Condition", "__gc", LuaScriptInterface::luaConditionDelete);
-	registerMethod("Condition", "delete", LuaScriptInterface::luaConditionDelete);
 
 	registerMethod("Condition", "getId", LuaScriptInterface::luaConditionGetId);
 	registerMethod("Condition", "getSubId", LuaScriptInterface::luaConditionGetSubId);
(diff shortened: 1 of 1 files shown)


ranisalt (approved):
Interesting

MillhioreBT:
It does not seem intuitive to me to obtain the user data to eliminate a type of condition
But I guess that doesn't concern me.

Regarding the elimination of the method `delete`, it seems very successful, in fact I love it, and all `userdatas` should be deleted by itself, there is no need to have methods like this

nekiro:
> It does not seem intuitive to me to obtain the `userdata` to eliminate a type of condition
> But I guess that doesn't concern me.
> 
> Regarding the elimination of the method `delete`, it seems very successful, in fact I love it, and all `userdatas` should be deleted by itself, there is no need to have methods like this

So what you do if you use getCondition to obtain condition and then want to delete it?

MillhioreBT (approved):

src/luascript.cpp
```diff
@@ -7714,17 +7713,23 @@ int LuaScriptInterface::luaCreatureAddCondition(lua_State* L)
...
+		uint32_t subId = getNumber<uint32_t>(L, 4, 0);
+		condition = creature->getCondition(conditionType, conditionId, subId);
+	}
 
-	ConditionType_t conditionType = getNumber<ConditionType_t>(L, 2);
-	ConditionId_t conditionId = getNumber<ConditionId_t>(L, 3, CONDITIONID_COMBAT);
-	uint32_t subId = getNumber<uint32_t>(L, 4, 0);
-	Condition* condition = creature->getCondition(conditionType, conditionId, subId);
 	if (condition) {
 		bool force = getBoolean(L, 5, false);
```
kornholi, line 7734: I don't think this will work with a `condition`?

MillhioreBT: Why do you think that? At first I saw it wrong too, but then I realized that if it is possible to pass the condition and eliminate it correctly, since it is verified in the list of conditions of the creature and if it matches in the search then it will eliminate it. 

However if you pass a condition other than the owner then nothing will happen, but who would do that?

MillhioreBT: The force parameter is the only detail, thanks @kornholi for letting us see this detail

PR #3525, opened 2021-08-01
Optimize memory access on XTEA

ranisalt:
### Pull Request Prelude

### Changes Proposed

Apparently, the way the code is currently written prevents some vectorization. When using `memcpy`, the compiler is able to figure it out that this is a vector operation and just does what it should. [With this change, there is potential for 1.6x speedup, that is 40% faster (click for benchmark)](https://quick-bench.com/q/eeJ_NmwJKOgxe9Hp-IqayVygiyQ)

This is also more succint.

### DIFF
--- a/src/xtea.cpp
+++ b/src/xtea.cpp
@@ -26,34 +26,9 @@
 
 namespace xtea {
 
-namespace {
-
-constexpr uint32_t delta = 0x9E3779B9;
-
-template<typename Round>
-void apply_rounds(uint8_t* data, size_t length, Round round)
-{
-	for (auto j = 0u; j < length; j += 8) {
-		uint32_t left = data[j+0] | data[j+1] << 8u | data[j+2] << 16u | data[j+3] << 24u,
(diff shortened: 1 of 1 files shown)


ranisalt:
This is amazing, when we first started optimizing this code in ~2018 the original/naive version was just not vectorizable and was 5x slower. Now, this is basically the same code as the original, just with interleaved loops, and it is now faster than the version that was 5x faster. Compilers have evolved so much...

kornholi (approved):
If we're going to assume endianness, why not change `data` to `uint32_t*` and skip the `memcpy`s? It compiles down to the same code.

DSpeichert (approved):

MillhioreBT (approved):

ranisalt:
> If we're going to assume endianness, why not change `data` to `uint32_t*` and skip the `memcpy`s

It may trigger undefined behavior, although a bit unclear if it would in this case. I had a chat with @djarek about it, and went with the safe route.

DSpeichert (approved):

PR #3529, opened 2021-08-01
Fix saving account storages

nekiro:
Break breaks the loop, so it cancels the save for all accounts, it should be continue for empty storages.


EPuncker (approved):

amatria:
Lmao. I was about to open a PR with this very same fix. Thank you very much @nekiro and @MillhioreBT. I was out in a little road trip and couldn't take care of it until today. Again, thank you :p.

PR #3533, opened 2021-08-02
Drop loot when creature kills other creature

### LINKED ISSUE #1918: Monster walk through bug
1. Log into a player, or summon a monster
2. Create (with /m) 10 rats. 2 Will start random step.
3. Create a monster that can push (Cyclops/Dragon)
### Expected behaviour

I'm not sure as I don't play official Tibia. Something along the line of pushing the weaker monster out of the way or side stepping until there's room is what I think is acceptable in this case.
### Actual behaviour

Stronger monster will try to walk through weaker monster, deleting him (weaker monster) in the process

nekiro:
Probably fix #1918
Right now creature will kill if there is no space, if there is it will get pushed.
If you know correct behaviour let me know.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1107,7 +1107,6 @@ void Monster::pushCreatures(Tile* tile)
 				}
 
 				monster->changeHealth(-monster->getHealth());
-				monster->setDropLoot(false);
 				removeCount++;
 			}
 


EPuncker (approved):

EPuncker:
I confirm this is working, now the creature is not deleted, but killed instead.

andersonfaaria:
Hi @nekiro thanks for bringing this to tfs but I have a quick question if you don't mind.
Any particular reason why didn't brought the part that prevents the corpse to be deleted?
During your tests there was any wrong behavior situation happening?

https://github.com/opentibiabr/otservbr-global/pull/1200/files

nekiro:
Why would the corpse get deleted though? Monsters are only pushing elevated items/non walkable like chests, crates, pots and so. Corpse is regular item, which is not even getting pushed by monsters at least it shouldn't.

andersonfaaria:
hmm, I get what you're saying. To be honest I've made this code in a partnership with @kotzlety about 4 or 5 years ago and I don't remember exactly why we put this check. Perhaps it was a redundance to avoid thinking of all situations where a corpse could be CONST_PROP_BLOCKPATH/CONST_PROP_BLOCKSOLID or back then we did had a situation like this in mind to account for.

Looking at it right now, perhaps back then I had misunderstood the code thinking of a corpse that could be deleted if it was on top or below a chest.

In either case, I can't think of any need for this part right now. Thanks for clarifying!

Issue #3535, opened 2021-08-02
New death system goals


Zbizu:
Since new death system is being worked on, I propose using this issue as both the checklist and a discussion thread.

Here are the things that I consider should be accomplished with the new system:

- [ ] do not send 0 hp packet (it disconnects the client?)
- [ ] do not kick the player on death
- [ ] trigger onCreatureDisappear
- [ ] add more post death messages regarding lost stats
- [ ] make legacy death system possible to use (with config value)
- [ ] allow rooking
- [ ] make rooking configurable
- [ ] trigger onLogin
- [ ] move to lua everything that doesn't need to be hardcoded

Zbizu:
I'm thinking of mixing that with #3251 to get a proper (and intended) client freeze on legacy death screen, but I'm not yet sure how to implement it.

PR #3537, opened 2021-08-03
make death screen optional (WIP)

Zbizu:
### Pull Request Prelude

### Changes Proposed
- remove death screen (to do: make it toggleable)
- rewire everything so it works without destroying the references
- TO DO: move to lua everything that doesn't need to be hardcoded
- TO DO: add enableDeathScreen (bool) to config.lua and restore the legacy death screen when enabled

**Issues addressed:**
#3535

Current status of the PR:
- you don't get disconnected on death anymore
- the death window doesn't show up
- everything else should be working as usual. If it doesn't, report it here

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -722,7 +722,7 @@ void Creature::onDeath()
 		setMaster(nullptr);
 	}
 
-	if (droppedCorpse) {
+	if (droppedCorpse && this->getPlayer() == nullptr) {
 		g_game.removeCreature(this, false);
 	}
 }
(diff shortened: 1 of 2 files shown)


Erza:
The death window should still show up, but the client should not be disconnected upon clicking "Ok" in the death window. That is the correct behaviour, see https://youtu.be/j1TYV8ZMU5M?t=21

Ezipsoft:
In Real Tibia you don't leave the party anymore when you die.

Ezipsoft:
Another problem would be the conditions, when the player dies you must remove the conditions from him.

Issue #3528, opened 2021-08-01
Halberds aren't unique right now.


Alkenix:
Fix in items.xml (line: 20730 & 20733):
`<item id="12046" article="a" name="halberd" />
	<item fromid="12047" toid="12048" name="halberds" />
	<item fromid="12049" toid="12050" article="a" name="heavy weapon rack" />
<item id="12051" article="a" name="halberd" />`

Zbizu:
use the template next time

see "files changed" section of this pr to see how to fix that: https://github.com/otland/forgottenserver/pull/3508
(the solution is using itemid instead of string)

Alkenix:
w8 w8 w8.. All i did was placing an zombie on my map on an uptodate tfs 1.3 server (incl. EPuncker's latest fix).

EPuncker:
lets blame forceMonsterTypesOnLoad to only work with monsters that are on map, and since I have no monster on my map I forgot this :D will fix it soon

Alkenix:
Same with grave flowers (vamps). ^^
Line: 35897

Zbizu:
I'd blame using strings in loot honestly.

Identifier's (id's) purpose is to make things possible to (as the name suggests) identify. Using strings is counterintuitive.

Considering that most of the duplicated loot can be resolved by parsing the first occurrence, I suggest stopping the search on first pickupable item. Alternatively the display of the warning can be a config value.

nekiro:
As I said some time ago, we should drop names support for items everywhere, loot is one of the things. Name is not unique. Id is.

Zbizu:
What about allowing some setting to disable warnings? eg about empty spawns or ambiguous item names with a simple true/false?

Alkenix:
Gonna need to correct this issue..
There are MANY as HELL items doubled in items.xml.
I have no clue why for example an arrow or power bolt needs 2 IDs, but i corrected () all of them for me right now.^^

If someone knows how to change monster loot to ID instead of the name, I'm gonna start that for mine soon.

Alkenix:
[#3540](https://github.com/otland/forgottenserver/pull/3540)

PR #3538, opened 2021-08-03
Use item id instead of name in monster loot

marmichalski:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** #3528 


marmichalski:
If you prefer, I can keep (& add missing) comments with item name 🤔

Zbizu:
Why did you remove the comments on item ids? It tells exactly what the user is editing and makes the whole thing readable.

If you're using a script to convert that, call ItemType(itemId):getName() on each item id.

marmichalski:
> Why did you remove the comments on item ids? It tells exactly what the user is editing and makes the whole thing readable.
> 

https://github.com/otland/forgottenserver/pull/3538#issuecomment-891816241

> If you're using a script to convert that, call ItemType(itemId):getName() on each item id to generate comments again.

Do you just type things for the heck of typing?

soul4soul:
I'd rather have the item names and run the risk of duplicate item name. At a minimum the item names have to be kept as comments to make editing the files manageable.

marmichalski:
Will this be stalled until merge conflicts arise?

Erza:
To help with your decision: id's should be used because of name duplication - you can't reliably use item names as loot because multiple items can have the same name and the engine will pick the first id that it finds.

Now when you look at https://github.com/otland/forgottenserver/pull/3540, non-unique item names use id's instead, but for one, that's inconsistent, and for two, nobody wants to maintain maintain that. It's not future-proof, if item names ever get corrected and one ends up with a duplicate name

Just use id's across the board and merge this PR.

marmichalski:
@EPuncker beep boop

Erza:
smh

EPuncker:
@DSpeichert this can be merged, I can't approve it because my browser lag with so many files

DSpeichert:
I think someone suggested keeping both `id` and `name` parameter, instead of moving the name out to a comment. It's easier to write and it's also more portable. IIRC, the engine will prefer `id`, if exists, before reverting to name.

marmichalski:
kthxbye

PR #3540, opened 2021-08-04
Fixed all unique item names on monster loot

### LINKED ISSUE #3528: Halberds aren't unique right now.

EPuncker:
### Pull Request Prelude

### Changes Proposed
Fixed all non unique item names, also prepared some for future updates as well

#Disclaimer
I've not added item names in comment, will do in a later monsters update or you can help me by pushing a commit to this PR

**Issues addressed:** 
fixes #3528 

### DIFF
--- a/data/monster/monsters/annihilon.xml
+++ b/data/monster/monsters/annihilon.xml
@@ -63,17 +63,17 @@
 		<item name="violet gem" chance="16666" />
 		<item name="yellow gem" chance="20000" />
 		<item name="green gem" chance="12500" />
-		<item id="2156" chance="20000" />
+		<item id="2156" chance="20000" /><!-- red gem -->
 		<item name="blue gem" chance="20000" />
-		<item id="2381" chance="20000" />
+		<item id="2381" chance="20000" /><!-- halberd -->
 		<item name="guardian halberd" chance="20000" />
 		<item name="heavy mace" chance="25000" />
-		<item id="2514" chance="4166" />
+		<item id="2514" chance="4166" /><!-- mastermind shield -->
 		<item name="guardian shield" chance="7692" />
-		<item id="2519" chance="11111" />
+		<item id="2519" chance="11111" /><!-- crown shield -->
 		<item name="demon shield" chance="4166" />
-		<item id="2528" chance="9090" />
-		<item id="2547" countmax="94" chance="16666" />
+		<item id="2528" chance="9090" /><!-- tower shield -->
+		<item id="2547" countmax="94" chance="16666" /><!-- power bolt -->
 		<item name="soul orb" countmax="5" chance="20000" />
 		<item name="demon horn" countmax="2" chance="12500" />
 		<item name="infernal bolt" countmax="46" chance="20000" />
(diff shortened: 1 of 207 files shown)


ranisalt:
This is a subset of #3538, though looking at it I like the idea of deprecating `name` and using only `id`. Makes it future-proof in case new items are added.

DSpeichert:
Without adding names in comments, this makes the files un-editable by humans. A fancy "monster editor" could automatically look up IDs from items.xml, but so far the use of tools wasn't necessary.

Alkenix:
So no one wants to allow the merge, right now? This could help alot.

Zbizu:
@DSpeichert what about #3538? haven't personally looked at every single file, but it seems to be what you want to accomplish (if I understood your intentions correctly here)

DSpeichert:
Yes, #3538 does seem to get it "right".
Every PR that rewrites so many files usually takes time to decide, since it dilutes history - we want to be sure it is definitely better than current version.

jo3bingham:
While I understand the argument for human-readability, I have to agree with `id`s over `name`s. But why not both? If we're going to have an `id` attribute with the name of the item in a comment, why not just have both an `id` and `name` attribute? File size will be relatively the same and the code already handles both cases (and gives `id` precedence). If TFS ever moves passed the 10.98 protocol, any added items that a name aren't known can just omit the `name` attribute and it will still work fine.

The only other suggestion I have is to make the warning message less vague. `Non-unique loot item "name"` means nothing to me without looking at the code. It should tell the user the item name, monster file, and how to fix it: `Non-unique loot item "name" in "/full_path/monster.xml". Multiple items were found with this name so the loot item was skipped. Fix: add the item 'id' to the loot item.`

soul4soul:
> 
> 
> While I understand the argument for human-readability, I have to agree with `id`s over `name`s. But why not both? If we're going to have an `id` attribute with the name of the item in a comment, why not just have both an `id` and `name` attribute? File size will be relatively the same and the code already handles both cases (and gives `id` precedence). If TFS ever moves passed the 10.98 protocol, any added items that a name aren't known can just omit the `name` attribute and it will still work fine.
> 
> The only other suggestion I have is to make the warning message less vague. `Non-unique loot item "name"` means nothing to me without looking at the code. It should tell the user the item name, monster file, and how to fix it: `Non-unique loot item "name" in "/full_path/monster.xml". Multiple items were found with this name so the loot item was skipped. Fix: add the item 'id' to the loot item.`

I do like that idea. I never looked at the code carefully enough to realize both work with id having precedence. In the monster converter I recently shared I convert the comment to a attribute I called "description". The reason being that is was more controlled to have the valuable information in an easily accessibly format supported manner instead of a free form comment. Knowing this I will change to using the name attribute and write both for my tool.

marmichalski:
I do not see any reason for neither name in comment nor as a redundant attribute. It's not like you're using item's names in other XML files... 🤷

EPuncker:
lets use the other one

DSpeichert (changes requested):
Since #3538 is closed, next step here is to include name as attribute instead of comment (have both name and id). I'd approve that.

DSpeichert (approved):

PR #3542, opened 2021-08-04
fix client crashes on "set outfit" window requests

### LINKED ISSUE #3541: "set outfit" crashes the client when currently worn outfit is lookTypeEx

### Steps to reproduce
1. Player("PlayerName"):setOutfit({lookTypeEx = 1284})
2. right click -> setOutfit

### Expected behaviour
set outfit window pops up

### Actual behaviour
(ctrl+C didn't work on that window for some reason)
obraz

### Environment
rl client / windows / 10.98

It works normally when my outfit is something from the list. I didn't modify outfits.xml either.

Zbizu:
### Pull Request Prelude

### Changes Proposed
- do not send outfit window when there are no outfits available
- send first outfit available when player's outfit is item id (lookTypeEx)

**Issues addressed:** client crash in both situations, closes #3541

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2778,10 +2778,21 @@ void ProtocolGame::sendHouseWindow(uint32_t windowTextId, const std::string& tex
 
 void ProtocolGame::sendOutfitWindow()
 {
+	const auto& outfits = Outfits::getInstance().getOutfits(player->getSex());
+	if (outfits.size() == 0) {
+		return;
+	}
+
 	NetworkMessage msg;
 	msg.addByte(0xC8);
 
 	Outfit_t currentOutfit = player->getDefaultOutfit();
+	if (currentOutfit.lookType == 0) {
+		Outfit_t newOutfit;
+		newOutfit.lookType = outfits.front().lookType;
+		currentOutfit = newOutfit;
+	}
+
 	Mount* currentMount = g_game.mounts.getMountByID(player->getCurrentMount());
 	if (currentMount) {
 		currentOutfit.lookMount = currentMount->clientId;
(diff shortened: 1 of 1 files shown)


src/protocolgame.cpp
```diff
@@ -2778,10 +2778,21 @@ void ProtocolGame::sendHouseWindow(uint32_t windowTextId, const std::string& tex
...
+	const auto& outfits = Outfits::getInstance().getOutfits(player->getSex());
+	if (outfits.size() == 0) {
+		return;
+	}
+
 	NetworkMessage msg;
 	msg.addByte(0xC8);
 
 	Outfit_t currentOutfit = player->getDefaultOutfit();
+	if (currentOutfit.lookType == 0) {
```
nekiro, line 2790: why this? If you have looktype of item then there should be no problem with sending current looktype even if you look as item, cause item looktype is also stored in Outfit_t

Zbizu: 1. tfs doesn't support adding item outfits, but you can set one through lua
2. I'm not even sure if item clientId can be sent in outfit window packet without crashing the client
3. very likely. At this moment I don't have enough time to reproduce the full outfit window packet.

EPuncker (approved):

src/protocolgame.cpp
marmichalski: OTClient supports displaying item id in outfit window, might be good to add a client check before forcing random outfit.

Anyway, I believe the `AddOutfit` should not be called within this method (unless it's an OTClient), just write the bytes like the client expects them.

DSpeichert: I'm going to guess that the Outfit window doesn't support "item" look, but it would be nice to know for sure.

DSpeichert (approved):

src/protocolgame.cpp
marmichalski: It does not: https://github.com/Cavitt/TibiaFlashClient/blob/master/scripts/tibia/network/Communication.as#L5950

PR #3543, opened 2021-08-06
Fix interaction of ghost mode & invisibility conditions

Erza:
### Pull Request Prelude

### Changes Proposed

Previously, the creature outfit would get messed up when equipping/dequipping stealth rings while in ghost mode, or switching ghost mode while having a stealth ring equipped

**Issues addressed:** none

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -9988,6 +9988,7 @@ int LuaScriptInterface::luaPlayerSetGhostMode(lua_State* L)
 
 	Tile* tile = player->getTile();
 	const Position& position = player->getPosition();
+	const bool isInvisible = player->isInvisible();
 
 	SpectatorVec spectators;
 	g_game.map.getSpectators(spectators, position, true, true);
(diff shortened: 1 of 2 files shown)


DSpeichert (approved):

EPuncker (approved):

src/luascript.cpp
```diff
@@ -9988,6 +9988,7 @@ int LuaScriptInterface::luaPlayerSetGhostMode(lua_State* L)
 
 	Tile* tile = player->getTile();
 	const Position& position = player->getPosition();
+	const bool isInvisible = player->isInvisible();
```
nekiro, line 9991: whats the point of declaring this if its gonna be used only once anyway?

nekiro (changes requested):

src/luascript.cpp
Erza: So isInvisible() doesn't get called multiple times while iterating through the spectators, that would be wasteful

nekiro: true, didn't see the loop.

nekiro (approved):

ranisalt:
Thanks!

PR #3545, opened 2021-08-07
Remove condition by Condition parameters (fixes #3544)

### LINKED ISSUE #3544: creature:removeCondition doesn't work properly with userdata

1. Add the following test code at `data/scripts/talkactions/condition.lua`
```lua
local condition = Condition(CONDITION_HASTE)
condition:setParameter(CONDITION_PARAM_TICKS, 10000)
condition:setFormula(0.3, -12, 0.3, -12)

local talk = TalkAction("/test")
	if player:hasCondition(CONDITION_HASTE) then
		print("remove")
		player:removeCondition(condition)
	else
		print("add")
		player:addCondition(condition)
	end
	return false
end

talk:register()
```

2. Run the `/test` command twice in game within 10 seconds, the condition should be added on the first execution and then removed on the second

### Expected behaviour
The haste condition should be added and removed respectively

### Actual behaviour
The haste condition is only added, but removing fails, even though `remove` is being printed in the console

### Related commits
https://github.com/otland/forgottenserver/commit/923f448f402392ee0c27632326da901947a724a8
https://github.com/otland/forgottenserver/commit/5d946f178e3f9e1ba48ba96e47ca7291bc4b6a4c

### Problematic code
It fails to find the condition with `std::find` in `Creature::removeCondition`
```cpp
void Creature::removeCondition(Condition* condition, bool force/* = false*/)
{
	auto it = std::find(conditions.begin(), conditions.end(), condition);
	if (it == conditions.end()) {
		return;
	}
```

Erza:
### Pull Request Prelude

### Changes Proposed

A copy of the Condition object was added to the creature in `LuaScriptInterface::luaCreatureAddCondition`, therefore `Creature::removeCondition` could not remove the condition properly, as the Condition object that was constructed from Userdata did not match the one stored in the creature's condition list.

Also applied const correctness.

**Issues addressed:** closes #3544 

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -7764,29 +7764,32 @@ int LuaScriptInterface::luaCreatureRemoveCondition(lua_State* L)
 {
 	// creature:removeCondition(conditionType[, conditionId = CONDITIONID_COMBAT[, subId = 0[, force = false]]])
 	// creature:removeCondition(condition[, force = false])
-	Creature* creature = getUserdata<Creature>(L, 1);
+	Creature* const creature = getUserdata<Creature>(L, 1);
 	if (!creature) {
 		lua_pushnil(L);
 		return 1;
 	}
 
-	Condition* condition = nullptr;
-
+	Condition* creatureCondition = nullptr;
 	bool force = false;
 
 	if (isUserdata(L, 2)) {
-		condition = getUserdata<Condition>(L, 2);
+		const Condition* const condition = getUserdata<Condition>(L, 2);
+		const ConditionType_t conditionType = condition->getType();
+		const ConditionId_t conditionId = condition->getId();
+		const uint32_t subId = condition->getSubId();
+		creatureCondition = creature->getCondition(conditionType, conditionId, subId);
 		force = getBoolean(L, 3, false);
 	} else {
-		ConditionType_t conditionType = getNumber<ConditionType_t>(L, 2);
-		ConditionId_t conditionId = getNumber<ConditionId_t>(L, 3, CONDITIONID_COMBAT);
-		uint32_t subId = getNumber<uint32_t>(L, 4, 0);
-		condition = creature->getCondition(conditionType, conditionId, subId);
+		const ConditionType_t conditionType = getNumber<ConditionType_t>(L, 2);
+		const ConditionId_t conditionId = getNumber<ConditionId_t>(L, 3, CONDITIONID_COMBAT);
+		const uint32_t subId = getNumber<uint32_t>(L, 4, 0);
+		creatureCondition = creature->getCondition(conditionType, conditionId, subId);
 		force = getBoolean(L, 5, false);
 	}
 
-	if (condition) {
-		creature->removeCondition(condition, force);
+	if (creatureCondition) {
+		creature->removeCondition(creatureCondition, force);
 		pushBoolean(L, true);
 	} else {
 		lua_pushnil(L);


MillhioreBT:
If you use logic, you will realize that this PR does not make any sense

This function has two behaviors:
* Remove by userdata: this gives us to understand that we need to pass the correct user data stored in the player conditions, if you pass a random or self-generated condition then it is very obvious that it will not remove anything since you random condition is not a player condition

* Remove by type: this is the behavior of a lifetime, you specify the search parameters, and if these parameters match any condition stored in the player then we obtain this condition and eliminate it

**¿How to use the function correctly with userdata?**
```lua
creature:removeCondition(creature:getCondition(...), ...)
```

*If you still insist, you can add an overload in lua and it looks much better, although counterintuitive and can lead people to think incorrectly, it is already out of basic logic.*

Erza:
> If you use logic, you will realize that this PR does not make any sense
> 
> This function has two behaviors:
> 
> * Remove by userdata: this gives us to understand that we need to pass the correct user data stored in the player conditions, if you pass a random or self-generated condition then it is very obvious that it will not remove anything since you random condition is not a player condition
> * Remove by type: this is the behavior of a lifetime, you specify the search parameters, and if these parameters match any condition stored in the player then we obtain this condition and eliminate it
> 
> **¿How to use the function correctly with userdata?**
> 
> ```lua
> creature:removeCondition(creature:getCondition(...), ...)
> ```
> 
> _If you still insist, you can add an overload in lua and it looks much better, although counterintuitive and can lead people to think incorrectly, it is already out of basic logic._

Read https://github.com/otland/forgottenserver/issues/3544#issuecomment-894699556

MillhioreBT:
> > If you use logic, you will realize that this PR does not make any sense
> > This function has two behaviors:
> > 
> > * Remove by userdata: this gives us to understand that we need to pass the correct user data stored in the player conditions, if you pass a random or self-generated condition then it is very obvious that it will not remove anything since you random condition is not a player condition
> > * Remove by type: this is the behavior of a lifetime, you specify the search parameters, and if these parameters match any condition stored in the player then we obtain this condition and eliminate it
> > 
> > **¿How to use the function correctly with userdata?**
> > ```lua
> > creature:removeCondition(creature:getCondition(...), ...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > _If you still insist, you can add an overload in lua and it looks much better, although counterintuitive and can lead people to think incorrectly, it is already out of basic logic._
> 
> Read [#3544 (comment)](https://github.com/otland/forgottenserver/issues/3544#issuecomment-894699556)

And I understand what you mean by using a condition and then extracting its properties and then finding the player's condition and so bla bla bla
I thought about this first than you, before this feature was added, but then I realized that it is just an additional complication
For example, what happens if we pass the correct user data? in this case we have to look for it again and it is like a pretty silly code

In case you are too lazy to write 20 more letters you can do a help function:
`data/lib/core/condition.lua`
```lua
function Condition:unpack()
	return self:getType(), self:getId(), self:getSubId()
end
```

`script.lua`
```lua
creature:removeCondition(condition:unpack())
```

DSpeichert (approved):
I'm not entirely against it, although we may be requiring in other places the Lua developer to also pass actual condition, not a replica. This behavior does not hurt though.

src/luascript.cpp
```diff
@@ -7764,29 +7764,32 @@ int LuaScriptInterface::luaCreatureRemoveCondition(lua_State* L)
 {
 	// creature:removeCondition(conditionType[, conditionId = CONDITIONID_COMBAT[, subId = 0[, force = false]]])
 	// creature:removeCondition(condition[, force = false])
-	Creature* creature = getUserdata<Creature>(L, 1);
+	Creature* const creature = getUserdata<Creature>(L, 1);
```
ranisalt, line 7767: This is not necessary

ranisalt, line 7767: This is not necessary

Codinablack:
All checks passing, all reviews passing. LGTM

PR is from 2021, its amazing its still able to be merged... Can we do so now?

MillhioreBT (approved):

Issue #3552, opened 2021-08-11
NPC's have a (Lua) memory leak & crash potential


Erza:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Stick this code in a talkaction
```lua
local tests = {
    npc = 1,
    monster = 2,
    item = 3,
}

local activeTest = tests.npc

local monsterName = "Enslaved Dwarf"
local npcName = "Banker"
local itemId = 2148

local function runTest(playerPos)
    for i = 1, 5000 do
        if activeTest == tests.npc then
            local npc = Game.createNpc(npcName, playerPos)
            npc:remove()
        elseif activeTest == tests.monster then
            local monster = Game.createMonster(monsterName, playerPos)
            monster:remove()
        elseif activeTest == tests.item then
            local item = Game.createItem(itemId, 100, playerPos)
            item:remove()
        end
    end
end

function onSay(player, words, param)
    local playerPos = player:getPosition()
    
    local start = os.clock()
    runTest(playerPos)
    local reloadStart = os.clock()
    Game.reload(RELOAD_TYPE_TALKACTIONS)
    local testEnd = os.clock()
    print("Test #1: Test: "..reloadStart - start.." - Reload: "..testEnd - reloadStart.." - Total: "..testEnd - start)

    start = os.clock()
    runTest(playerPos)
    reloadStart = os.clock()
    Game.reload(RELOAD_TYPE_TALKACTIONS)
    testEnd = os.clock()
    print("Test #2: Test: "..reloadStart - start.." - Reload: "..testEnd - reloadStart.." - Total: "..testEnd - start)

    start = os.clock()
    runTest(playerPos)
    reloadStart = os.clock()
    Game.reload(RELOAD_TYPE_TALKACTIONS)
    testEnd = os.clock()
    print("Test #3: Test: "..reloadStart - start.." - Reload: "..testEnd - reloadStart.." - Total: "..testEnd - start)

    playerPos:sendMagicEffect(CONST_ME_BLOCKHIT)
    print(("-"):rep(50))
    return false
end
```
2. Run the talkaction a few times and keep an eye on the memory usage of the process

### Expected behaviour
NPC's cleanly deallocate their memory upon being removed

### Actual behaviour
[...]

PR #3553, opened 2021-08-11
first attempt at clearing npc events(callbacks) from scriptinterface …

yamaken93:
…lua events table

### Pull Request Prelude

### Changes Proposed

Remove/clear the events in the events table so resources can be freed on the lua side.

**Issues addressed:** 
#3552

### DIFF
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -1110,6 +1110,14 @@ NpcEventsHandler::NpcEventsHandler(const std::string& file, Npc* npc) :
 	}
 }
 
+NpcEventsHandler::~NpcEventsHandler()
+{
+	std::array<int32_t, 7> eventArray = {creatureSayEvent, creatureDisappearEvent, creatureAppearEvent, creatureMoveEvent, playerCloseChannelEvent, playerEndTradeEvent, thinkEvent};
+	for (int32_t eventId : eventArray) {
+		scriptInterface->deleteEvent(eventId);
+	}
+}
+
 bool NpcEventsHandler::isLoaded() const
 {
 	return loaded;
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -423,6 +423,34 @@ int32_t LuaScriptInterface::getMetaEvent(const std::string& globalName, const st
 	return runningEventId++;
 }
 
+bool LuaScriptInterface::deleteEvent(int32_t eventId)
+{
+	if (eventId == -1) {
+		return false;
+	}
+
+	//get our events table
+	lua_rawgeti(luaState, LUA_REGISTRYINDEX, eventTableRef);
+	if (!isTable(luaState, -1)) {
+		lua_pop(luaState, 1);
+		return false;
+	}
+
+	lua_rawgeti(luaState, -1, eventId);
+	if (!isFunction(luaState, -1)) {
+		lua_pop(luaState, 2);
+		return false;
+	}
+	lua_pop(luaState, 1);
+
+	lua_pushnil(luaState);
+	lua_rawseti(luaState, -2, eventId);
+
+	lua_pop(luaState, 1);
+
+	return true;
+}
+
 const std::string& LuaScriptInterface::getFileById(int32_t scriptId)
 {
 	if (scriptId == EVENT_ID_LOADING) {
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -218,6 +218,7 @@ class LuaScriptInterface
 		int32_t getEvent(const std::string& eventName);
 		int32_t getEvent();
 		int32_t getMetaEvent(const std::string& globalName, const std::string& eventName);
+		bool deleteEvent(int32_t eventId);
 
 		static ScriptEnvironment* getScriptEnv() {
 			assert(scriptEnvIndex >= 0 && scriptEnvIndex < 16);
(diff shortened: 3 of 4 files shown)


yamaken93:
```
~MonsterInfo() {
			if (!scriptInterface) {
				return;
			}
			std::array<int32_t, 5> eventArray = {creatureSayEvent, creatureDisappearEvent, creatureAppearEvent, creatureMoveEvent, thinkEvent};
			for (int32_t eventId : eventArray) {
				scriptInterface->deleteEvent(eventId);
			}
		};
```
I still need to find a way to add this kinda of code to MonsterInfo so we can prevent the leak for monsters events too.

src/npc.cpp
```diff
@@ -1110,6 +1112,14 @@ NpcEventsHandler::NpcEventsHandler(const std::string& file, Npc* npc) :
 	}
 }
 
+NpcEventsHandler::~NpcEventsHandler()
+{
+	std::array<int32_t, 7> eventArray = {creatureSayEvent, creatureDisappearEvent, creatureAppearEvent, creatureMoveEvent, playerCloseChannelEvent, playerEndTradeEvent, thinkEvent};
```
nekiro, line 1117: static?

Erza:
I used an empty NPC script file with these contents
```lua
function onCreatureAppear(cid)
end
function onCreatureDisappear(cid)
end

function onCreatureSay(cid, type, msg)
end

function onThink()
end
```

And `data/npc/lib/npc.lua` only had the following content, in order to allocate some memory and then see if it will get free'd
```lua
local bigTable = {}
local index = 1
for i = 1, 10000 do
    local tbl = {}
    local tmpIndex = 1
    
    for i = 1, 100 do
        tbl[tmpIndex] = ("%s%u%s%u"):format(("a"):rep(100), math.random(1000000000000), ("b"):rep(100), math.random(1000000000000))
        tmpIndex = tmpIndex + 1
    end

    bigTable[index] = tbl
    index = index + 1
end
```

I then ran the test from #3552 again, with 5000 iterations each, totaling 15k NPC's. This is the memory graph


The second tiny dip in the graph is due to me running `/reload npc` after the test, so `lua_gc` will get called. As we can see, the `bigTable` object's don't get cleaned up. There must be something more to this issue.

Erza:
New findings: it seems like Lua is reserving a certain amount of memory that it then re-uses, only allocating more when there is no more free room.

## With this PR
### Callbacks are unreferenced, memory usage stops increasing, GC times stay consistent - 690MB~ of memory used by the process after running the test 6x with 15k NPC's each


## Without this PR
### Callbacks stay referenced forever, memory usage keeps increasing, GC times increase as well 1.5GB~ of memory used by the process after running the test 4x with 15k NPC's each


Now we just need to unreference all the other places where `LuaScriptInterface::getEvent` or `LuaScriptInterface::getMetaEvent` are called

src/npc.cpp
ranisalt: @nekiro a static in the *destructor* means it will always try to delete the events of the first NpcEventsHandler to ever be deleted.

No need for the temporary though
```cpp
for (int32_t eventId : {creatureSayEvent, creatureDisappearEvent, creatureAppearEvent, creatureMoveEvent, playerCloseChannelEvent, playerEndTradeEvent, thinkEvent}) {
```

src/npc.cpp
```diff
@@ -36,12 +36,14 @@ void Npcs::reload()
...
 
-	delete Npc::scriptInterface;
+	NpcScriptInterface* oldScriptInterface = Npc::scriptInterface;
 	Npc::scriptInterface = nullptr;
 
 	for (const auto& it : npcs) {
 		it.second->reload();
 	}
+
+	delete oldScriptInterface;
```
ranisalt, line 46: Unique ptr?

nekiro: +1

MillhioreBT: +2

ranisalt: image

Erza (changes requested):

src/baseevents.cpp
```diff
@@ -167,13 +168,28 @@ bool Event::loadCallback()
...
 }
 
+CallBack::~CallBack()
+{
+	try {
+		if (scriptInterface && scriptId) {
+			scriptInterface->deleteEvent(scriptId);
+		}
+	} catch (const std::exception& e) {
+		fmt::printf("CallBack::~CallBack(): its look like we have a dangling scriptInterface in a combat callback.");
```
EPuncker, line 178:
```suggestion
		fmt::printf("CallBack::~CallBack(): its look like we have a dangling scriptInterface in a combat callback.");
		std::cout << "CallBack::~CallBack(): It looks like we have a dangling scriptInterface in a combat callback." << std::endl;
```

src/baseevents.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include "pugicast.h"
 #include "tools.h"
+#include "fmt/printf.h"
```
EPuncker, line 26:
```suggestion
```

src/baseevents.cpp
```diff
@@ -167,13 +168,28 @@ bool Event::loadCallback()
 	return true;
 }
 
+CallBack::~CallBack()
+{
+	try {
+		if (scriptInterface && scriptId) {
+			scriptInterface->deleteEvent(scriptId);
+		}
+	} catch (const std::exception& e) {
```
EPuncker, line 177: > 'e': unreferenced local variable

PR #3559, opened 2021-08-13
Npc improvements

Erza:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:**
1. NPC's would take 1 unnecessary step upon spawning, even if there was nobody to see it
2. NPC's did not behave correctly when being spawned like `/s Banker`, you would watch the NPC take 1 step and be in idle state, never moving again until a player relogged or moved within it's visible range.
3. Compare creature pointers instead of ID's

### DIFF
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -119,13 +119,21 @@ void Npc::reload()
 	reset();
 	load();
 
+	SpectatorVec players;
+	g_game.map.getSpectators(players, getPosition(), true, true);
+	for (const auto& player : players) {
+		spectators.insert(player->getPlayer());
+	}
+
+	updateIdleStatus();
+
+	if (walkTicks > 0 && !spectators.empty()) {
+		addEventWalk();
+	}
+
 	// Simulate that the creature is placed on the map again.
 	if (npcEventHandler) {
 		npcEventHandler->onCreatureAppear(this);
-	}
-
-	if (walkTicks > 0) {
-		addEventWalk();
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/npc.cpp
```diff
@@ -255,7 +264,16 @@ void Npc::onCreatureAppear(Creature* creature, bool isLogin)
 	Creature::onCreatureAppear(creature, isLogin);
 
 	if (creature == this) {
-		if (walkTicks > 0) {
+		SpectatorVec players;
```
MillhioreBT, line 267: the redundancy of the word `player` `players` `player`

src/npc.cpp
```diff
@@ -413,8 +431,12 @@ bool Npc::getNextStep(Direction& dir, uint32_t& flags)
 	return getRandomStep(dir);
 }
 
-void Npc::setIdle(bool idle)
+void Npc::setIdle(const bool idle)
```
MillhioreBT, line 434: I don't know if the const is necessary, but I like it anyway ;)

src/npc.cpp
Erza: What?

src/npc.cpp
Erza: It's not necessary, just good practice to use const correctness

https://lmgtfy.app/?q=const+correctness+c%2B%2B

src/npc.cpp
MillhioreBT: Don't worry, I was thinking of changing the name of the vector, but I already checked it well and there is nothing wrong, excuse me

DSpeichert (approved):

EPuncker (approved):

PR #3561, opened 2021-08-14
Prevent possible crash in the event onSpawn

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
If the monster stops being inactive in the onSpawn event, and it cannot be placed in the world, then we will have a null pointer in the verification vector of creatures, with this small change we will avoid a possible crash

### Ways to make the monster stop being inactive
```lua
function Monster:onSpawn(position, startup, artificial)
   self:addHealth(1)
   return true
end
```
```lua
function Monster:onSpawn(position, startup, artificial)
   self:setIdle(false)
   return true
end
```
and any function that makes idle false

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -655,7 +655,9 @@ void Monster::setIdle(bool idle)
 	isIdle = idle;
 
 	if (!isIdle) {
-		g_game.addCreatureCheck(this);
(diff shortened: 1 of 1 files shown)


MillhioreBT:
I'll keep it closed, because PR #3563 handles this case perfectly.

PR #3563, opened 2021-08-14
Monster_ptr and a better way and complete to handle the onSpawn event

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
* `createMonster` now returns a `unique_ptr`
* `onSpawn` is now a metatable event `Creature`
* `onSpawn` it only works for Npc's / Monsters, as `onLogin` does its job well and we don't want anything redundant
* the `onSpawn` event now fires with calls to `Map::placeCreature`
* the `onSpawn` event only fires if the creature can be placed
* add the new `Creature.getMonster` and `Creature.getNpc` methods to be able to work comfortably with the new `onSpawn` method
* The `Monster` Destructor now makes the monster inactive to avoid any possible null pointers in the creatureCheckList


EPuncker (approved):

MillhioreBT:
Closed because it is not what I really wanted to propose, I was wrong!

PR #3576, opened 2021-08-17
Update prevent infinite loop teleports

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Currently you can still create infinite teleports loops, it is a matter of generating the loop in a given position and making two positions not related to the initial one enter an infinite loop, With the small changes that I made, now no type of immense or infinite loops will be a problem, also now everything is verified in a loop, eliminating the recursion of the code.

### DIFF
--- a/src/teleport.cpp
+++ b/src/teleport.cpp
@@ -70,27 +70,6 @@ void Teleport::addThing(Thing* thing)
 	return addThing(0, thing);
 }
 
-bool Teleport::isHuge(const Tile* destTile)
-{
-	Teleport* teleport = destTile->getTeleportItem();
-	uint8_t tries = 0;
-	do {
-		const Position& nextDestPos = teleport->getDestPos();
-		if (getPosition() == nextDestPos || ++tries >= 32) {
-			return true;
-		}
-
-		if (const Tile* tempTile = g_game.map.getTile(nextDestPos)) {
-			teleport = tempTile->getTeleportItem();
-		} else {
-			break;
-		}
-
-	} while (teleport);
-
-	return false;
-}
-
 void Teleport::addThing(int32_t, Thing* thing)
 {
 	Tile* destTile = g_game.map.getTile(destPos);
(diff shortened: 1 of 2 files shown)


src/teleport.cpp
```diff
@@ -70,19 +70,24 @@ void Teleport::addThing(Thing* thing)
 	return addThing(0, thing);
 }
 
-bool Teleport::checkInfinityLoop(Tile* destTile)
+bool Teleport::isHuge(const Tile* destTile)
```
Erza, line 73: There is no reason to change the name of the method, especially if it's not a very descriptive name

src/teleport.cpp
```diff
@@ -70,19 +70,24 @@ void Teleport::addThing(Thing* thing)
...
 {
-	if (!destTile) {
-		return false;
-	}
```
Erza, line 77: Leave the pointer check in, since `g_game.map.getTile` can return a nullptr and will crash the server

src/teleport.cpp
```diff
@@ -70,19 +70,24 @@ void Teleport::addThing(Thing* thing)
...
-
-	if (Teleport* teleport = destTile->getTeleportItem()) {
+	Teleport* teleport = destTile->getTeleportItem();
```
Erza, line 75: Check if `teleport` is a nullptr, since we're not guaranteed to have a teleport item on `destTile`

Erza (changes requested):

ranisalt:
To be honest, is this a real issue? The only way to do make an infinite teleport is by changing the map to do so, or making it through a script, therefore it must be done by the server owner. I would only add guards for things that may be triggered by the players, but the owners should be able to do whatever they want - even if that means crashing their own server.

src/teleport.cpp
MillhioreBT: The function was not designed to be passed a nullptr, since this verification is done before in line 92

Erza: True, I looked at the diff in split view and looked at the code on the wrong side. The old code needed it because it was recursing with `return checkInfinityLoop(g_game.map.getTile(nextDestPos));`

DSpeichert (approved):

EPuncker (approved):

EPuncker (approved):

PR #3580, opened 2021-08-18
Don't show scripts loading by default

EPuncker:
### Pull Request Prelude

### Changes Proposed
In the first place I don't know why is this even an option, it might have made sense in the beginning but now that we have only a few scripts, the console is already spamming lots of loaded files (that we don't need to know, since if they are put inside **scripts** folder they are supposed to be loaded, and if there is any error the console will report with despite showScriptsLogInConsole configuration). We don't have a list of actions.xml entries loaded, etc, and while I'm not requesting the removal of this setting, I'm just wanting to set it off by default. People can still use print() or whatsoever if they want to display their custom mod/system on the console.


Zbizu (approved):
LGTM

MillhioreBT (approved):

Erza (approved):

DSpeichert (approved):

PR #3583, opened 2021-08-20
Clear Npc spectators on reset

Erza:
Followup to https://github.com/otland/forgottenserver/pull/3559

### Pull Request Prelude

### Changes Proposed

Clear spectator set on Npc resets, e.g. on reloading, before we re-populate the spectator set with surrounding players
In practice it's not an issue, since we can never double-populate, since std::set entries are unique

**Issues addressed:**

### DIFF
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -112,6 +112,7 @@ void Npc::reset()
 
 	parameters.clear();
 	shopPlayerSet.clear();
+	spectators.clear();
 }
 
 void Npc::reload()


MillhioreBT (approved):

MillhioreBT:
As you say it is not necessary, but is it better to convert this to a vector? And then `clear` would make more sense?

ranisalt:
> but is it better to convert this to a vector

It needs to be unique. Maybe unordered_set, but at this level there isn't really a big difference, as Spectator sets never grow large enough to be noticeably slow.

(this comment is bound to age badly)

marmichalski (approved):

EPuncker (approved):

nekiro (approved):

PR #3584, opened 2021-08-21
Add attack speed attribute to items.xml

### LINKED ISSUE #3395: attackSpeed via items.lua feature.
https://otland.net/threads/c-item-attack-speed-attr-key-not-recognized.262272/

Would be nice to have this on TFS.

Code is old and need a edit.

marmichalski:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** closes #3395.

🤷


Dockerfile
```diff
@@ -19,7 +19,7 @@ COPY cmake /usr/src/forgottenserver/cmake/
 COPY src /usr/src/forgottenserver/src/
 COPY CMakeLists.txt /usr/src/forgottenserver/
 WORKDIR /usr/src/forgottenserver/build
-RUN cmake .. && make
+RUN cmake .. && make -j $(($(nproc) / 2 + 1))
```
Zbizu, line 22: is that intended or pushed by accident?

marmichalski: Intended, pushed as separate commit: https://github.com/otland/forgottenserver/pull/3584/commits

marmichalski:
I've included basic attributes (`/attr <name>,<value>`) talkaction with this PR for easier testing. I can move it to it's own PR, if you prefer, as it's not connected in any way with this PR, it's just here for convenience.

src/luascript.cpp
```diff
@@ -2033,6 +2037,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("Game", "getWorldType", LuaScriptInterface::luaGameGetWorldType);
 	registerMethod("Game", "setWorldType", LuaScriptInterface::luaGameSetWorldType);
 
+	registerMethod("Game", "getItemAttributeByName", LuaScriptInterface::luaGameGetItemAttributeByName);
```
marmichalski, line 2040: no idea where functions from `tools.cpp` should go. `getReturnMessage` is in `Game`, so I've gone with it for `stringToItemAttribute` too.

nekiro:
One pull request one change, so if you add attack speed remove all other code you pushed.
It looks okay, but the lowest you can go is creature think interval, which is 1000 ms.

marmichalski:
> One pull request one change, so if you add attack speed remove all other code you pushed.
> It looks okay, but the lowest you can go is creature think interval, which is 1000 ms.

Not really true (about the interval)

nekiro:
> > One pull request one change, so if you add attack speed remove all other code you pushed.
> > It looks okay, but the lowest you can go is creature think interval, which is 1000 ms.
> 
> Not really true (about the interval). I'll split the changes when I'm home, no problem.

Not true why? doAttacking method executes every 1000 ms, cause its called from checkCreatures loop.

Dockerfile
ranisalt: I'd rather pass `MAKEFLAGS` as an env/arg and not try to anything smart _inside_ the container.

marmichalski:
> > > One pull request one change, so if you add attack speed remove all other code you pushed.
> > > It looks okay, but the lowest you can go is creature think interval, which is 1000 ms.
> > 
> > 
> > Not really true (about the interval). I'll split the changes when I'm home, no problem.
> 
> Not true why? doAttacking method executes every 1000 ms, cause its called from checkCreatures loop, which loops every 1000 ms.


https://user-images.githubusercontent.com/57528542/130335292-ef454fa8-56dd-41d3-99c9-917be33d043f.mov

Anyway, I do agree that a sane minimum could be set - but not sure what should it be: 250? 100?

nekiro:
> > > > One pull request one change, so if you add attack speed remove all other code you pushed.
> > > > It looks okay, but the lowest you can go is creature think interval, which is 1000 ms.
> > > 
> > > 
> > > Not really true (about the interval). I'll split the changes when I'm home, no problem.
> > 
> > 
> > Not true why? doAttacking method executes every 1000 ms, cause its called from checkCreatures loop, which loops every 1000 ms.
> 
>  Screen.Recording.2021-08-21.at.23.31.38.mov 
> Anyway, I do agree that a sane minimum could be set - but not sure what should it be: 250? 100?

Wait what, you want to lower onthink interval? That's a big nono.

marmichalski:
> > > > > One pull request one change, so if you add attack speed remove all other code you pushed.
> > > > > It looks okay, but the lowest you can go is creature think interval, which is 1000 ms.
> > > > 
> > > > 
> > > > Not really true (about the interval). I'll split the changes when I'm home, no problem.
> > > 
> > > 
> > > Not true why? doAttacking method executes every 1000 ms, cause its called from checkCreatures loop, which loops every 1000 ms.
> > 
> > 
> > Screen.Recording.2021-08-21.at.23.31.38.mov
> > Anyway, I do agree that a sane minimum could be set - but not sure what should it be: 250? 100?
> 
> Wait what, you want to lower onthink interval? That's a big nono.

No. The attack speed that can be set.

EPuncker:
> but not sure what should it be: 250? 100?

200 should be enough

EPuncker (dismissed):

src/luascript.cpp
```diff
@@ -1510,6 +1510,10 @@ void LuaScriptInterface::registerFunctions()
 	registerEnum(ITEM_ATTRIBUTE_CHARGES)
 	registerEnum(ITEM_ATTRIBUTE_FLUIDTYPE)
 	registerEnum(ITEM_ATTRIBUTE_DOORID)
+	registerEnum(ITEM_ATTRIBUTE_DECAYTO)
```
EPuncker, line 1513: doesn't these belong to the attr PR? 😄

marmichalski: Nope, not really. Those are just missing enums, they are not used in `/attr`.

EPuncker:
Okay, I've tested this today and everything is working just fine, I just have two things to point out:

1. cap the minimum value to 100 (silently?)
2. display the attackspeed attribute at item description? (ASPD:? AS:? Atk Speed:? idk)

marmichalski:
> Okay, I've tested this today and everything is working just fine, I just have two things to point out:
> 
> 1. cap the minimum value to 100 (silently?)
> 2. display the attackspeed attribute at item description? (ASPD:? AS:? Atk Speed:? idk)

1.
```
[Warning - Items::parseItemNode] AttackSpeed lower than 100 for item: 2400
```

2.
```
You see a crossbow (Range:5, Atk Spd:0.1s). //100
You see a club (Atk:7, Def:7, Atk Spd:1s). //1000
You see a club (Atk:7, Def:7, Atk Spd:5.5s).//5500
You see a club (Atk:7, Def:7, Atk Spd:101.521s).// 101521
```

Could `::setprecision` before outputting it, but I'd say... it doesn't matter, whose gonna set attack speed to some random number 🤷

DSpeichert (approved):

PR #3585, opened 2021-08-21
Allow to disable pvp per vocation

marmichalski:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 

Closes (supersedes) PR #3143.

### DIFF
--- a/data/XML/vocations.xml
+++ b/data/XML/vocations.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <vocations>
-	<vocation id="0" clientid="0" name="None" description="none" gaincap="10" gainhp="5" gainmana="5" gainhpticks="6" gainhpamount="1" gainmanaticks="6" gainmanaamount="1" manamultiplier="4.0" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" allowpvp="0" fromvoc="0">
+	<vocation id="0" clientid="0" name="None" description="none" gaincap="10" gainhp="5" gainmana="5" gainhpticks="6" gainhpamount="1" gainmanaticks="6" gainmanaamount="1" manamultiplier="4.0" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" allowPvp="0" fromvoc="0">
 		<formula meleeDamage="1.0" distDamage="1.0" defense="1.0" armor="1.0" />
 		<skill id="0" multiplier="1.5" />
 		<skill id="1" multiplier="2.0" />
--- a/data/XML/vocations.xml
+++ b/data/XML/vocations.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <vocations>
-	<vocation id="0" clientid="0" name="None" description="none" gaincap="10" gainhp="5" gainmana="5" gainhpticks="6" gainhpamount="1" gainmanaticks="6" gainmanaamount="1" manamultiplier="4.0" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" fromvoc="0">
+	<vocation id="0" clientid="0" name="None" description="none" gaincap="10" gainhp="5" gainmana="5" gainhpticks="6" gainhpamount="1" gainmanaticks="6" gainmanaamount="1" manamultiplier="4.0" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" allowpvp="0" fromvoc="0">
 		<formula meleeDamage="1.0" distDamage="1.0" defense="1.0" armor="1.0" />
 		<skill id="0" multiplier="1.5" />
 		<skill id="1" multiplier="2.0" />
--- a/src/vocation.h
+++ b/src/vocation.h
@@ -86,6 +86,10 @@ class Vocation
 			return fromVocation;
 		}
 
+		bool allowsPvp() const {
+			return allowPvp;
+		}
+
 		float meleeDamageMultiplier = 1.0f;
 		float distDamageMultiplier = 1.0f;
 		float defenseMultiplier = 1.0f;
@@ -116,6 +120,8 @@ class Vocation
 
 		uint8_t soulMax = 100;
 		uint8_t clientId = 0;
+
+		bool allowPvp = true;
 };
 
 class Vocations
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -327,7 +327,7 @@ bool Combat::isProtected(const Player* attacker, const Player* target)
 		return true;
 	}
 
-	if (attacker->getVocationId() == VOCATION_NONE || target->getVocationId() == VOCATION_NONE) {
+	if (!attacker->getVocation()->allowsPvp() || !target->getVocation()->allowsPvp()) {
 		return true;
 	}
 
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2625,6 +2625,8 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("Vocation", "getDemotion", LuaScriptInterface::luaVocationGetDemotion);
 	registerMethod("Vocation", "getPromotion", LuaScriptInterface::luaVocationGetPromotion);
 
+	registerMethod("Vocation", "allowsPvp", LuaScriptInterface::luaVocationAllowsPvp);
+
 	// Town
 	registerClass("Town", "", LuaScriptInterface::luaTownCreate);
 	registerMetaMethod("Town", "__eq", LuaScriptInterface::luaUserdataCompare);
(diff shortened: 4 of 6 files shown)


src/vocation.h
```diff
@@ -116,6 +120,8 @@ class Vocation
 
 		uint8_t soulMax = 100;
 		uint8_t clientId = 0;
+
+		bool _protected = false;
```
marmichalski, line 124: 💩... but `protected` is reserved within that scope 🤷

EPuncker (approved):

src/vocation.h
Erza: `isProtected`?

src/vocation.h
```diff
@@ -86,6 +86,10 @@ class Vocation
 			return fromVocation;
 		}
 
+		bool isProtected() const {
+			return _protected;
```
nekiro, line 90: please don't use variables starting with underscore

nekiro (changes requested):

src/vocation.h
ranisalt: Not possible in this scenario

ranisalt:
Protected from _what_?

src/vocation.h
nekiro: why?

Erza: It's an access modifier keyword like `public` & `private`

EPuncker (dismissed):

ranisalt (commented):
I find negated booleans to be confusing. Personally, I would pick `enablePvP` or `enabledPvP` instead, defaulting to true. Nonetheless it is a cool feature, more meaningful than checking for `VOCATION_NONE`

Erza (dismissed):

nekiro (dismissed):

marmichalski:
@ranisalt I'm thinking about moving it to something similar to monster `flags`. So you could define them like, for example:
```xml
<flags>
  <flag droploot="0" />
  <flag somethingpvp="0" />
  <flag juniorbank="0" /> <!-- will need to think about a better name 😎 -->
</flags>
```

This will allow to replace all `VOCATION_NONE` checks with proper flag checks.

WDYT?

ranisalt:
> I'm thinking about moving it to something similar to monster `flags`

This is a very good idea, indeed!

src/vocation.h
```diff
@@ -116,6 +120,8 @@ class Vocation
 
 		uint8_t soulMax = 100;
 		uint8_t clientId = 0;
+
+		bool pvp = true;
```
marmichalski, line 124: dunno, `canPvp`, `hasPvp`, `enabledPvp`, `allowPvp`, `somethingsomething`, `canEngagePvp`...

marmichalski: same goes for flag name, im keeping this as draft and awaiting ideas for a nice property, flag and method names 😎

soul4soul: From most to least I like `canEngagePvp`, `allowPvp`, and `enabledPvp`.

marmichalski: And how do you like the one I went with?

EPuncker (dismissed):

nekiro (approved):

Erza (approved):

data/XML/vocations.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <vocations>
 	<vocation id="0" clientid="0" name="None" description="none" gaincap="10" gainhp="5" gainmana="5" gainhpticks="6" gainhpamount="1" gainmanaticks="6" gainmanaamount="1" manamultiplier="4.0" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" fromvoc="0">
+		<flags>
+			<flag allowPvp="0" />
+		</flags>
```
DSpeichert, line 6: I don't think it makes sense to have this parent-child structure of "list of flags", where each flag contains attributes... and it can actually contain many such attributes, because why not?

Specifically, the design here does not prevent this:
```xml
<flags>
	<flag allowPvp="0" allowSomething="1" />
	<flag allowSomethingElse="0" />
</flags>
```

There's really not good way to avoid that, because this (below) is less flexible, although does solve this problem somewhat:
```xml
<flags>
	<flag>allowPvp</flag>
	<flag>allowSomethingElse</flag>
</flags>
```

The problem is that everything would be a boolean now, and can't be defaulted to true if not specified.

So I think the simplest way to do that is less nesting:
```xml
<flags allowPvp="0" something="1" somethingElse="1" />
```

DSpeichert (changes requested):
XML design is ambiguous

data/XML/vocations.xml
Erza: Well, you approved it for groups too, which uses the same design https://github.com/otland/forgottenserver/pull/2546

marmichalski: IMHO it's better to be consistent with what's there instead of introducing new design for every new feature.

With that said, I could make it consistent with what's already in vocations, skills, so it would be repeated `flag` nodes, without nesting:
```xml
<flag name/id="allowPvp" value="1" />
```

Or keep it consistent with other flag declarations in monsters and groups, and leave it as is.

marmichalski: Anyway, I don't see whats ambiguous. Your example:
```xml
<flags>
	<flag allowPvp="0" allowSomething="1" />
	<flag allowSomethingElse="0" />
</flags>
```

Is not a valid use case, `allowSomething` will be ignored as the flag takes _only_ the first attribute.

If you really wanna do something new here, then maybe:
```xml
<flags>
  <allowPvp />
  <!-- which is equivalent to -->
  <allowPvp>1</allowPvp>
</flags>
```

Defining a flag without value makes it "set" (true) by default.

Anyway, I'm not even sure about the `flags` name, since I might want to add explicit `protectionLevel` per vocation, which would not really fit into `flags`. Maybe `attributes`? Or `features`?
```xml
<features>
  <pvp>0</pvp>
  <lootDrop>1</lootDrop>
  <protectionLevel>40</protectionLevel>
  <skillLossReduction>10</skillLossReduction>
</features>
```

🤷

ranisalt: Oh my god XML is so damn ugly

DSpeichert: > Is not a valid use case, allowSomething will be ignored as the flag takes only the first attribute.

That's exactly the ambiguity problem. It's not obvious to the user. What you're describing is an implementation limitation.

We all know XML design is a problem and #2546 showed a similar problem. If you want to follow that other style for consistency, that is fine as well. I think what made the difference there is the vast number of flags, where here we only have one for now?

On the other hand, if you call it features, which really turn into parameters... then why call it anything at all? Everything inside vocation definition is a parameter/property/feature/flag/option... whatever. Maybe just keep adding extra attributes to the top level `<vocation/>`? Ugly, but it already is anyway, and simple. You're right that these flags aren't really distinctive "flags".

marmichalski: > That's exactly the ambiguity problem. It's not obvious to the user. What you're describing is an implementation limitation.

It's not really ambiguous with current implementation... You have a nested single flag under flags, so why would you think you can assign more than one flag attribute to flag node? 🤔 
It's kinda really explicit, flags collection containing entries describing single flag.

It seems like you'd be most comfortable with how skills are defined, full example here:
```xml
<vocation id="1" ...>
	<formula meleeDamage="1.0" distDamage="1.0" defense="1.0" armor="1.0" />
	<flag id="allowPvp" value="0" />
	<skill id="0" multiplier="1.5" />
	<skill id="1" multiplier="2.0" />
	<skill id="2" multiplier="2.0" />
</vocation>
```

Flattening flags might be actually a good idea, as in your own example:
```xml
<flags allowPvp="0" something="1" somethingElse="1" />
```

But we might want to resolve first how do we want to call it, if these are not gonna be only simple boolean flags. Personally I would rather not add more attributes to the root node (vocation).

soul4soul: We already have flags that work exactly like this for monsters too. I don't see a problem being consistent with what we already have implemented. Not all of monster flags are boolean. Some are numeric like targetdistance and staticattack. Seems better off to add it as is and look at it again in 2.0 if that is the timeframe for really switching away from XML.

MillhioreBT: I support the idea of ​​squashing it into one line:
`<flags allowPvp="0" something="1" somethingElse="1" />`
is consistent with what we already have in this xml file

marmichalski:
@DSpeichert any decision? This is blocking new pull requests for me 🙄 .

data/XML/vocations.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <vocations>
 	<vocation id="0" clientid="0" name="None" description="none" gaincap="10" gainhp="5" gainmana="5" gainhpticks="6" gainhpamount="1" gainmanaticks="6" gainmanaamount="1" manamultiplier="4.0" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" fromvoc="0">
```
DSpeichert, line 3: Since we're considering just one new attribute in this PR, I think it's best to add it to this ugly-looking giant `vocation` element as `allowPvp="0/1"`.

I don't think we'll find a pattern that is consistent for different files because there's too much history. Inventing a new "flags" thing at this point seems unnecessary to me.

EPuncker: > Since we're considering just one new attribute in this PR, I think it's best to add it to this ugly-looking giant `vocation` element as `allowPvp="0/1"`.
> 
> I don't think we'll find a pattern that is consistent for different files because there's too much history. Inventing a new "flags" thing at this point seems unnecessary to me.

but if I'm not mistaken he intends to add new flags in new PR, related to bank restriction (and drop loot?)

DSpeichert: Maybe. Still, I see no difference between them and the existing attributes. Should we move all existing ones into "flags"? Of course not. So what's the point of adding "flags". Also, I think flags are mostly boolean because it's a reference to... flags, that are usually raised (or not). The only real flags in TFS were [the ones stored as bitmask](https://my-aac.org/flags/). We're not doing that here and hopefully we never do that.

DSpeichert (approved):

PR #3589, opened 2021-08-21
Add monster challengeable flag

marmichalski:
### Pull Request Prelude

Requested some time ago in https://otland.net/threads/monster-exeta-res.277481/

### Changes Proposed

Ability to make monsters immune to challenge (for example the spell).

**Issues addressed:** 

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1041,7 +1041,7 @@ void LuaScriptInterface::registerFunctions()
 	//doTargetCombat(cid, target, type, min, max, effect[, origin = ORIGIN_SPELL[, blockArmor = false[, blockShield = false[, ignoreResistances = false]]]])
 	lua_register(luaState, "doTargetCombat", LuaScriptInterface::luaDoTargetCombat);
 
-	//doChallengeCreature(cid, target)
+	//doChallengeCreature(cid, target[, force = false])
 	lua_register(luaState, "doChallengeCreature", LuaScriptInterface::luaDoChallengeCreature);
 
 	//addEvent(callback, delay, ...)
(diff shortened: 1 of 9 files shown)


EPuncker (approved):

src/monster.cpp
```diff
@@ -1957,12 +1957,16 @@ void Monster::changeHealth(int32_t healthChange, bool sendHealthChange/* = true*
...
 }
 
-bool Monster::challengeCreature(Creature* creature)
+bool Monster::challengeCreature(Creature* creature, bool force/* = false*/)
 {
 	if (isSummon()) {
 		return false;
 	}
 
+	if (!mType->info.isChallengeable && !force) {
```
MillhioreBT, line 1966: Here `force` doesn't really work as its name implies, regardless of whether `force` is `true` or `false`, it will not change any behavior

`force` is supposed to be a parameter that forces the creature regardless of whether it is challengeable or no
image

marmichalski: True, thanks for spotting 🙉

marmichalski: Actually, it works as expected 🤔 

if monster **is not** challengeable **AND** call **is not** forced, then and only then **return false**.

So, examples:

| Challengeable | Force | Will exit early? |
|---------------|-------|----------------|
| Y                      | N        | N                       |
| Y                      | Y        | N                       |
| N                      | N        | Y                       |
| N                      | Y        | N                       |

src/luascript.cpp
```diff
@@ -12887,6 +12888,23 @@ int LuaScriptInterface::luaMonsterTypeIsAttackable(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaMonsterTypeIsChallengeable(lua_State* L)
+{
+	// get: monsterType:isChallengeable() set: monsterType:isChallengeable(bool)
```
DSpeichert, line 12893: Do we use this pattern of getter/setter being the same function name anywhere else?
It's simple and nice, but could be confusing to read sometimes.

DSpeichert (commented):
LGTM, one comment

src/luascript.cpp
marmichalski: Yes, I think it was introduced with the revscriptsys.

PR #3592, opened 2021-08-22
Add monster rename functionality

### LINKED ISSUE #3564: Add monster:setName(name, description)
Rename an ingame creature(monster) name by using a script command

### Available workarounds
none

### Prior art
this has been done before, but need to fix so you don't need to relog to see the new name
https://otland.net/threads/setcreaturename-monster-setname-for-tfs-1-2.245108/

marmichalski:
### Pull Request Prelude

### Changes Proposed

It would be great if someone could test how it behaves on cip client, as the `changeOnMap` packet behaves differently than in OTClient. OTClients downside is that the creature is removed from the battle list, this would probably have to be fixed there. If the monster has been selected as a target, it stays selected. Unfortunately, I can't test with the official client, so... @EPuncker, maybe you could? 😎 

**Issues addressed:** Closes #3564.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -46,7 +46,7 @@ Monster* Monster::createMonster(const std::string& name)
 
 Monster::Monster(MonsterType* mType) :
 	Creature(),
-	strDescription(mType->nameDescription),
+	nameDescription(mType->nameDescription),
 	mType(mType)
 {
 	defaultOutfit = mType->info.outfit;
@@ -80,6 +80,41 @@ void Monster::addList()
 void Monster::removeList()
 {
 	g_game.removeMonster(this);
+}
+
+const std::string& Monster::getName() const
+{
+	if (name.empty()) {
+		return mType->name;
+	}
+	return name;
+}
+
+void Monster::setName(const std::string& name)
+{
+	if (getName() == name) {
+		return;
+	}
+
+	this->name = name;
+
+	// NOTE: Due to how client caches known creatures,
+	// it is not feasible to send creature update to everyone that has ever met it
+	SpectatorVec spectators;
+	g_game.map.getSpectators(spectators, position, true, true);
+	for (Creature* spectator : spectators) {
+		if (Player* tmpPlayer = spectator->getPlayer()) {
+			tmpPlayer->sendUpdateTileCreature(this);
+		}
+	}
+}
+
+const std::string& Monster::getNameDescription() const
+{
+	if (nameDescription.empty()) {
+		return mType->nameDescription;
+	}
+	return nameDescription;
 }
 
 bool Monster::canSee(const Position& pos) const
--- a/src/monster.h
+++ b/src/monster.h
@@ -64,17 +64,19 @@ class Monster final : public Creature
 			}
 		}
 
-		void removeList() override;
 		void addList() override;
+		void removeList() override;
 
-		const std::string& getName() const override {
-			return mType->name;
-		}
-		const std::string& getNameDescription() const override {
-			return mType->nameDescription;
-		}
+		const std::string& getName() const override;
+		void setName(const std::string& name);
+
+		const std::string& getNameDescription() const override;
+		void setNameDescription(const std::string& nameDescription) {
+			this->nameDescription = nameDescription;
+		};
+
 		std::string getDescription(int32_t) const override {
-			return strDescription + '.';
+			return nameDescription + '.';
 		}
 
 		CreatureType_t getType() const override {
@@ -180,7 +182,8 @@ class Monster final : public Creature
 		CreatureHashSet friendList;
 		CreatureList targetList;
 
-		std::string strDescription;
+		std::string name;
+		std::string nameDescription;
 
 		MonsterType* mType;
 		Spawn* spawn = nullptr;
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1115,7 +1115,7 @@ void LuaScriptInterface::registerFunctions()
 	registerEnum(AMMO_THROWINGKNIFE)
 	registerEnum(AMMO_STONE)
 	registerEnum(AMMO_SNOWBALL)
-	
+
 	registerEnum(BUG_CATEGORY_MAP)
 	registerEnum(BUG_CATEGORY_TYPO)
 	registerEnum(BUG_CATEGORY_TECHNICAL)
@@ -2514,6 +2514,8 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("Monster", "isMonster", LuaScriptInterface::luaMonsterIsMonster);
 
 	registerMethod("Monster", "getType", LuaScriptInterface::luaMonsterGetType);
+
+	registerMethod("Monster", "rename", LuaScriptInterface::luaMonsterRename);
 
 	registerMethod("Monster", "getSpawnPosition", LuaScriptInterface::luaMonsterGetSpawnPosition);
 	registerMethod("Monster", "isInSpawnRange", LuaScriptInterface::luaMonsterIsInSpawnRange);
@@ -10235,6 +10237,24 @@ int LuaScriptInterface::luaMonsterGetType(lua_State* L)
 	} else {
 		lua_pushnil(L);
 	}
+	return 1;
+}
+
+int LuaScriptInterface::luaMonsterRename(lua_State* L)
+{
+	// monster:rename(name[, nameDescription])
+	Monster* monster = getUserdata<Monster>(L, 1);
+	if (!monster) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	monster->setName(getString(L, 2));
+	if (lua_gettop(L) >= 3) {
+		monster->setNameDescription(getString(L, 3));
+	}
+
+	pushBoolean(L, true);
 	return 1;
 }
 
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -1034,6 +1034,8 @@ class LuaScriptInterface
 
 		static int luaMonsterGetType(lua_State* L);
 
+		static int luaMonsterRename(lua_State* L);
+
 		static int luaMonsterGetSpawnPosition(lua_State* L);
 		static int luaMonsterIsInSpawnRange(lua_State* L);
 
--- a/src/player.h
+++ b/src/player.h
@@ -139,8 +139,8 @@ class Player final : public Creature, public Cylinder
 		const std::string& getName() const override {
 			return name;
 		}
-		void setName(std::string name) {
-			this->name = std::move(name);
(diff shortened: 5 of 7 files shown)


EPuncker:
image
working, the only "issue" is:

> if you summon a monster, walk offscreen, rename monster, walk back into view, the monster name will not change
> because you aren't classified as a spectator when the renaming happened

image

marmichalski:
I don't think there will be an easy way to solve it, but I'll give it a look.

Zbizu:
I thought about it earlier. My approach would be adding field "displayName" to monster class (or even creature). It would function like item attribute name - if set, would be sent instead of default monster name everywhere. If not - it would display base name. A method `creature:getBaseName()` could be introduced to get monster base name.

It could even work with player tags/titles because packets work with creature ids.

marmichalski:
> I thought about it earlier. My approach would be adding field "displayName" to monster class (or even creature). It would function like item attribute name - if set, would be sent instead of default monster name everywhere. If not - it would display base name. A method `creature:getBaseName()` could be introduced to get monster base name.
> 
> It could even work with player tags/titles put in front of their names because packets work with creature ids.

And how that differs from current implementation? You don't send a name for a known creature, that's the issue here. You have to trick the client to replace it without making it unpleasant for players.

Xikini:
Just wanted to note that the first picture is with cip client, second picture is with OTC.

Wasn't sure how to mc the cip client, so wasn't able to test spectators.

marmichalski:
> Just wanted to note that the first picture is with cip client, second picture is with OTC.
> 
> Wasn't sure how to mc the cip client, so wasn't able to test spectators.

The "caster" is also considered spectator.

Zbizu:
> And how that differs from current implementation? You don't send a name for a known creature, that's the issue here. You have to trick the client to replace it without making it unpleasant for players.
it would allow renaming players and npcs as well

btw, the problem here is "known" flag I believe (see https://github.com/otland/forgottenserver/blob/e1b5a991d03ed0be4ec46e15f7336d0546eb6917/src/protocolgame.cpp#L2946)

If you clear "known" list when renaming the creature, it should update for everyone.

marmichalski:
> > And how that differs from current implementation? You don't send a name for a known creature, that's the issue here. You have to trick the client to replace it without making it unpleasant for players.
> > it would allow renaming players and npcs as well
> 
> btw, the problem here is "known" flag I believe (see
> 
> https://github.com/otland/forgottenserver/blob/e1b5a991d03ed0be4ec46e15f7336d0546eb6917/src/protocolgame.cpp#L2946
> 
> )
> If you clear "known" list when renaming the creature, it should update for everyone.

Yes, and you'll duplicate the creature in the client, since it also has it's known map 🙄

Zbizu:
so there's no way for the client to "unlearn" the creature?

Xikini:
> > Just wanted to note that the first picture is with cip client, second picture is with OTC.
> > Wasn't sure how to mc the cip client, so wasn't able to test spectators.
> 
> The "caster" is also considered spectator.

Ah, sorry for not being more specific.
Meant I was unable to login with a second character to re-create the same situation as in the OTC client.
So I'm unsure if the demon would get rename on the cip client with that specific scenario.

nekiro:
> so there's no way for the client to "unlearn" the creature?

Yes, there is a way. Simply send remove creature and then sendAddCreature

marmichalski:
> > > Just wanted to note that the first picture is with cip client, second picture is with OTC.
> > > Wasn't sure how to mc the cip client, so wasn't able to test spectators.
> > 
> > 
> > The "caster" is also considered spectator.
> 
> Ah, sorry for not being more specific.
> Meant I was unable to login with a second character to re-create the same situation as in the OTC client.
> So I'm unsure if the demon would get rename on the cip client with that specific scenario.

It would not get renamed, since the packet to replace creature with itself is sent only to spectators. AFAIK client won't accept/will ignore "changeOnMap" packet for a position it does not see, that's the reason I've said it won't be easy to solve, if even possible without adding some 💩 logic that will be of purpose only for that scenario (which might not even be a real problem).

ranisalt:
> I don't think there will be an easy way to solve it, but I'll give it a look.

Either keeping a separate list of every spectator the monster ever had, or sending the updated name to every new spectator when they enter the screen

EPuncker (approved):

MillhioreBT:
> > I don't think there will be an easy way to solve it, but I'll give it a look.
> 
> Either keeping a separate list of every spectator the monster ever had, or sending the updated name to every new spectator when they enter the screen

Maintain a list in the monster of those viewers who already know this monster and then by updating the name we remove this creature from all viewers who ever knew it

marmichalski:
> > > I don't think there will be an easy way to solve it, but I'll give it a look.
> > 
> > 
> > Either keeping a separate list of every spectator the monster ever had, or sending the updated name to every new spectator when they enter the screen
> 
> Maintain a list in the monster of those spectators who already know this monster and then by updating the name we remove this creature from all spectators who ever knew it

That's not an easy way. That's a shitty way and it's not possible, anyway. You can't remove a creature that's not visible, client will crash. At least not without hacks.

IMHO this is going to turn into a big pile of smelly poopoo just to solve a problem that might never actually happen 🙄

I'll browse flash client sources again this weekend. If there is no decent solution, I'll just leave it as-is.

Zbizu:
technically you can send a packet that the creature is on your face and another one to make it disappear, but the tricky part is that the data about who knows the creature is scattered around protocol objects.

MillhioreBT:
> > > > I don't think there will be an easy way to solve it, but I'll give it a look.
> > > 
> > > 
> > > Either keeping a separate list of every spectator the monster ever had, or sending the updated name to every new spectator when they enter the screen
> > 
> > 
> > Maintain a list in the monster of those spectators who already know this monster and then by updating the name we remove this creature from all spectators who ever knew it
> 
> That's not an easy way. That's a shitty way and it's not possible, anyway. You can't remove a creature that's not visible, client will crash. At least not without hacks.
> 
> IMHO this is going to turn into a big pile of smelly poopoo just to solve a problem that might never actually happen 🙄
> 
> I'll browse flash client sources again this weekend. If there is no decent solution, I'll just leave it as-is.

If it cannot be solved for all viewers then this functionality is useless, it would be very crappy to add a function that does not do what it is supposed to do And the client does not crash when we remove a known creature, regardless of whether or not it is in the field of vision.

marmichalski:
> > > > > I don't think there will be an easy way to solve it, but I'll give it a look.
> > > > 
> > > > 
> > > > Either keeping a separate list of every spectator the monster ever had, or sending the updated name to every new spectator when they enter the screen
> > > 
> > > 
> > > Maintain a list in the monster of those spectators who already know this monster and then by updating the name we remove this creature from all spectators who ever knew it
> > 
> > 
> > That's not an easy way. That's a shitty way and it's not possible, anyway. You can't remove a creature that's not visible, client will crash. At least not without hacks.
> > IMHO this is going to turn into a big pile of smelly poopoo just to solve a problem that might never actually happen 🙄
> > I'll browse flash client sources again this weekend. If there is no decent solution, I'll just leave it as-is.
> 
> If it cannot be solved for all viewers then this functionality is useless, it would be very crappy to add a function that does not do what it is supposed to do And the client does not crash when we remove a known creature, regardless of whether or not it is in the field of vision.

with stackpos: https://github.com/Cavitt/TibiaFlashClient/blob/master/scripts/tibia/network/Communication.as#L2161
without stackpos: https://github.com/Cavitt/TibiaFlashClient/blob/master/scripts/tibia/network/Communication.as#L2191

Take it or close it, just don't keep it open for years.

MillhioreBT:
I do not have the power to make that decision, but for me it seems bad that we cannot change the name of monster for all the spectators.

and please do not bother, that's why we are here to discuss a bit and think of some way to solve the problems, I have noticed that CipClient has many limitations, if we used OTC it would be a different story

marmichalski:
You can change for **spectators**. It does not cover the, IMHO ultra-rare, situation where you go and see the monster, then leave and come back after it was renamed. If it was renamed, then it most probably got killed already when you're back. 🙉

src/monster.cpp
```diff
@@ -82,6 +82,40 @@ void Monster::removeList()
...
 
+const std::string& Monster::getName() const
+{
+	if (name.empty()) {
+		return mType->name;
+	}
+	return name;
+}
+
+void Monster::setName(std::string name)
```
Erza, line 93: Const reference for name argument?
`void Monster::setName(const std::string& name)`

src/monster.h
```diff
@@ -64,17 +64,19 @@ class Monster final : public Creature
...
 
-		void removeList() override;
 		void addList() override;
+		void removeList() override;
+
+		const std::string& getName() const override;
+		void setName(std::string name);
+
+		const std::string& getNameDescription() const override;
+		void setNameDescription(std::string nameDescription) {
```
Erza, line 74: Const reference for nameDescription argument?
`void setNameDescription(const std::string& nameDescription)`

src/monster.cpp
marmichalski: I've inlined it, for whatever reason, with how it is in player class:

https://github.com/otland/forgottenserver/blob/802ea5e92a30f23bf09175319b3d86c00bbe3160/src/player.h#L142-L147

🤷 

Can of course change that to const ref, it that's what wanted.

nekiro:
Please don't force push the branch. It destroys git log and it's harder to review what changed

marmichalski:
> Please don't force push the branch. It destroys git log and it's harder to review what changed

https://github.com/otland/forgottenserver/compare/e1b5a991d03ed0be4ec46e15f7336d0546eb6917..c8a2beb829463aba4a9a1ec1208e45bba716d087

src/monster.cpp
nekiro: I think const ref is better than copying to move it

Erza: > I think const ref is better than copying to move it

Technically we can do both

DSpeichert:
I'm okay with merging this but it needs a comment in the code describing the limitation of this feature. We don't want someone to start considering this as a bug tomorrow, rather a well-discussed limitation that probably isn't worth chasing, but someone might be surprised.

Do we still need it to be tested?

EPuncker:
> I'm okay with merging this but it needs a comment in the code describing the limitation of this feature. We don't want someone to start considering this as a bug tomorrow, rather a well-discussed limitation that probably isn't worth chasing, but someone might be surprised.
> 
> Do we still need it to be tested?

Me and xikini have tested this and its fine, other than what they are calling limitation, which will only occur in very specific scenarios

EPuncker (approved):

EPuncker (approved):

DSpeichert (approved):

PR #3596, opened 2021-08-24
Implement loaded mounts limit

### LINKED ISSUE #3578: Client crash if you have more than 255 mounts

1. Add a few custom mounts to your mounts.xml file until you have 254 or more
2. Open your server
3. Login your (GM) character and open outfit window

### Expected behaviour
no client crash

### Actual behaviour
client crash

I think we should have a limit check as we have for outfits [here](https://github.com/otland/forgottenserver/blob/master/src/protocolgame.cpp#L2807)

marmichalski:
### Pull Request Prelude

### Changes Proposed
Since mount id is declared as `uint8_t`, protocol check is not needed, since we can report invalid configuration during loading.

**Issues addressed:** Closes #3578.

### DIFF
--- a/src/mounts.cpp
+++ b/src/mounts.cpp
@@ -40,8 +40,19 @@ bool Mounts::loadFromXml()
 	}
 
 	for (auto mountNode : doc.child("mounts").children()) {
+		uint16_t nodeId = pugi::cast<uint16_t>(mountNode.attribute("id").value());
+		if (nodeId == 0 || nodeId > std::numeric_limits<uint8_t>::max()) {
+			std::cout << "[Notice - Mounts::loadFromXml] Mount id \"" << nodeId << "\" is not within 1 and 255 range" <<  std::endl;
+			continue;
+		}
+
+		if (getMountByID(nodeId)) {
+			std::cout << "[Notice - Mounts::loadFromXml] Duplicate mount with id: " << nodeId << std::endl;
+			continue;
+		}
+
 		mounts.emplace_back(
-			static_cast<uint8_t>(pugi::cast<uint16_t>(mountNode.attribute("id").value())),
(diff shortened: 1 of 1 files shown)


MillhioreBT:
The limit check should be done in the construction of the packet just as it is done with the outfits.

There is not much science with that, in any case if you think it is better to do it directly in the load, then do it for the outfits too.
Anyway, I would like to see someone else's opinion

marmichalski:
> The limit check should be done in the construction of the packet just as it is done with the outfits.
> 
> There is not much science with that, in any case if you think it is better to do it directly in the load, then do it for the outfits too.
> Anyway, I would like to see someone else's opinion

Why would it? Mounts, for some reason, have `id` that's an uint8_t. If you have more than 255, it will just overflow and you will have duplicated mounts with the same id. It's limited """by design""" 😎

ranisalt (dismissed):

Erza (approved):

EPuncker (approved):
Shouldn't we warn the user that the max id is 255 too, to prevent future confusion? (if and when we get to that point) 😄

marmichalski:
> Shouldn't we warn the user that the max id is 255 too, to prevent future confusion? (if and when we get to that point) 😄

id 255 is allowed as it fits in u8. 0 also fits, but to keep mounts size limited to 254, it's treated as invalid

EPuncker:
> Oh, I think I've understood now what you want - to display the allowed range in notice?

exactly

DSpeichert (approved):

marmichalski:
```
[Notice - Mounts::loadFromXml] Mount id "0" is not within 1 and 255 range
[Notice - Mounts::loadFromXml] Mount id "256" is not within 1 and 255 range
```

EPuncker (approved):

PR #3597, opened 2021-08-24
Add forced argument to on logout creature event

### LINKED ISSUE #2147: onLogout should have second parameter isForced!

marmichalski:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** Closes #2147

### DIFF
--- a/data/creaturescripts/scripts/logout.lua
+++ b/data/creaturescripts/scripts/logout.lua
@@ -1,4 +1,4 @@
-function onLogout(player)
+function onLogout(player, forced)
 	local playerId = player:getId()
 	if nextUseStaminaTime[playerId] then
 		nextUseStaminaTime[playerId] = nil
--- a/src/creatureevent.cpp
+++ b/src/creatureevent.cpp
@@ -139,12 +139,12 @@ bool CreatureEvents::playerLogin(Player* player) const
 	return true;
 }
 
-bool CreatureEvents::playerLogout(Player* player) const
+bool CreatureEvents::playerLogout(Player* player, bool forced/* = false*/) const
 {
 	//fire global event if is registered
 	for (const auto& it : creatureEvents) {
 		if (it.second.getEventType() == CREATURE_EVENT_LOGOUT) {
-			if (!it.second.executeOnLogout(player)) {
+			if (!it.second.executeOnLogout(player, forced)) {
 				return false;
 			}
 		}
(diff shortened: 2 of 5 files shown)


MillhioreBT:
Any real life examples? I do not see any use for this parameter, if it is forced or not forced, what happens?

marmichalski:
> Any real life examples? I do not see any use for this parameter, if it is forced or not forced, what happens?

Can't you imagine anything? 😛 

First thing that came to mind is do not let GM to kick a player that has engaged pvp.

marmichalski:
> > Any real life examples? I do not see any use for this parameter, if it is forced or not forced, what happens?
> 
> Can't you imagine anything? 😛
> 
> First thing that came to mind is do not let GM to kick a player that has engaged pvp (which could also be done in kick script, but whatever).

That's a bad example actually, because Player::kickPlayer ignores the event return value. 🤷

marmichalski:
@DSpeichert / @EPuncker if this is not wanted, could the issue be closed along with this PR?

EPuncker (approved):
well, I see no issue in having it, but lets see what @DSpeichert thinks

DSpeichert:
Hey, I haven't had the chance to think about it earlier.

It looks good but I think the primary purpose is for it to be used in scripts. Therefore, it would be good to extensively document when `forced` is `true` or `false`. Naturally, client-initiated logout is "not forced", and server initiated logout is "forced" (but is it always?). Now, is noPongKick a client or server initiated? Someone might be x-logging on purpose.

Therefore, maybe, and I know it doesn't necessarily make it better, we should have a "Logout cause", maybe an ENUM of "client" / "server" / "timeout" (noPongKick) ?

If it's supposed to be useful and make a stable scripting interface, we should think about it at the design stage - which is now.

Zbizu:
imo `creature:remove(forced?)` and closing the server should be only cases when forced is true

MillhioreBT:
> Hey, I haven't had the chance to think about it earlier.
> 
> It looks good but I think the primary purpose is for it to be used in scripts. Therefore, it would be good to extensively document when `forced` is `true` or `false`. Naturally, client-initiated logout is "not forced", and server initiated logout is "forced" (but is it always?). Now, is noPongKick a client or server initiated? Someone might be x-logging on purpose.
> 
> Therefore, maybe, and I know it doesn't necessarily make it better, we should have a "Logout cause", maybe an ENUM of "client" / "server" / "timeout" (noPongKick) ?
> 
> If it's supposed to be useful and make a stable scripting interface, we should think about it at the design stage - which is now.

It is a perfect idea.

PR #3604, opened 2021-08-27
Add monsters walking back to spawn

marmichalski:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Monsters will now walk back to their spawn position when lured away and no targets are around. See the attached videos for more details. When the path gets blocked, monster will just stop.

`spawnwalk.mp4` - monster does not walk back when in allowed range, and walk backs when goes outside of it.
`spawnwalk_block.mp4` - monster cancels walk to spawn when target appears | stop when walking back and path gets blocked.

https://user-images.githubusercontent.com/57528542/131175910-c9ec25c6-89cc-4dda-bee6-df0428018947.mp4

https://user-images.githubusercontent.com/57528542/131175972-af2d6fb3-69c3-4b32-91b1-d8c96ee8cdec.mp4

**Issues addressed:** Part of #922, #3600

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #3610, opened 2021-08-28
Revert "Add monsters walking back to spawn"

nekiro:
Reverts otland/forgottenserver#3604

It doesn't work properly.


MillhioreBT (approved):
These changes do not fully satisfy what we would expect from this mechanism.

marmichalski:
so why not just fix it? 🙄

MillhioreBT:
> so why not just fix it? 🙄

If you still keep the branch, do it, fix it and then push the changes, i suppose this revert is in case you don't decide to fix it.

marmichalski:
This happens because of how poor A* is in TFS. I'll just chunk the path, so it does not have to check high count of nodes at once 🙄.

lgrossi:
Would be nice to know the reason of the revert. That way people of other repositories can see it and maybe help to improve/fix. It's also an amazing way of keeping log for the future. 😁

marmichalski:
> Would be nice to know the reason of the revert. That way people of other repositories can see it and maybe help to improve/fix. It's also an amazing way of keeping log for the future. 😁

https://github.com/otland/forgottenserver/pull/3604#issuecomment-907679749
https://github.com/otland/forgottenserver/pull/3610#issuecomment-907689912

PR with solution has been opened.

lgrossi:
> > Would be nice to know the reason of the revert. That way people of other repositories can see it and maybe help to improve/fix. It's also an amazing way of keeping log for the future. 😁
> 
> [#3604 (comment)](https://github.com/otland/forgottenserver/pull/3604#issuecomment-907679749)
> [#3610 (comment)](https://github.com/otland/forgottenserver/pull/3610#issuecomment-907689912)
> 
> PR with solution has been opened.

Thanks for linking it, I think it's important to have some visibility in the revert PR.

DSpeichert:
Reverting isn't necessary if the author is providing a fix quickly.

PR #3611, opened 2021-08-28
 Make monster loot count accept count higher than 100

### LINKED ISSUE #2859: Loot countmax "issue" when using values bigger than 100

soul4soul:
### Pull Request Prelude

### Changes Proposed

Make monster loot count accept count higher than 100
Credits Epuncker and MillhioreBT. 

Reopening this PR (original was #2875), it was discussed in #2906 to use a solution where the count is rolled once. If you want to roll multiple times just add multiple item entries.

**Issues addressed:** 
fixes #2859

### DIFF
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -18,40 +18,47 @@ function Container.createLootItem(self, item)
 	end
 
 	if itemCount > 0 then
-		local tmpItem = Game.createItem(item.itemId, math.min(itemCount, 100))
-		if not tmpItem then
+		local tmpItem = {}
+		while itemCount > 100 do
+			local itm = self:addItem(item.itemId, 100)
+			if itm then
+				tmpItem[#tmpItem + 1] = itm
+			end
+			itemCount = itemCount - 100
+		end
+
+		if itemCount > 0 then
+			local itm = self:addItem(item.itemId, itemCount)
+			if itm then
+				tmpItem[#tmpItem + 1] = itm
+			end
+		end
+
+		if #tmpItem == 0 then
 			return false
 		end
 
-		if tmpItem:isContainer() then
-			for i = 1, #item.childLoot do
-				if not tmpItem:createLootItem(item.childLoot[i]) then
-					tmpItem:remove()
-					return false
+		for index, tmp in pairs(tmpItem) do
+			if tmp:isContainer() then
+				for i = 1, #item.childLoot do
+					if not tmp:createLootItem(item.childLoot[i]) then
+						tmp:remove()
+						return false
+					end
 				end
 			end
 
-			if #item.childLoot > 0 and tmpItem:getSize() == 0 then
-				tmpItem:remove()
-				return true
+			if item.subType ~= -1 then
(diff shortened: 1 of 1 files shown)


MillhioreBT:
I think in the discussion it is mentioned that several roll should be made for every 100 stackable items, so a formula is required for this, not simply increasing the total number, and increasing the roll does not seem intuitive, people expect everything to be automatic and it should work as usual.

Anyway I am not very clear about this

soul4soul:
> 
> 
> I think in the discussion it is mentioned that several roll should be made for every 100 stackable items, so a formula is required for this, not simply increasing the total number, and increasing the roll does not seem intuitive, people expect everything to be automatic and it should work as usual.
> 
> Anyway I am not very clear about this

Read https://github.com/otland/forgottenserver/pull/2906#issuecomment-617757708
And even more so read https://github.com/otland/forgottenserver/pull/2906#issuecomment-621500554

MillhioreBT:
In any case I think it is easier to leave the code like this and just add the loop ;)
only 3 lines more
<details>
<summary>container.lua</summary>

```lua
function Container.createLootItem(self, item)
		return true
	end

	local itemCount = 0
		if ItemType(item.itemId):isStackable() then
			itemCount = randvalue % item.maxCount + 1
		else
			itemCount = 1
		end
	end

	while itemCount > 0 do
		local count = math.min(100, itemCount)
		local tmpItem = Game.createItem(item.itemId, count)
		if not tmpItem then
			return false
		end

		if tmpItem:isContainer() then
			for i = 1, #item.childLoot do
				if not tmpItem:createLootItem(item.childLoot[i]) then
					tmpItem:remove()
					return false
				end
			end

			if #item.childLoot > 0 and tmpItem:getSize() == 0 then
				tmpItem:remove()
				return true
			end
		end

		if item.subType ~= -1 then
			tmpItem:setAttribute(ITEM_ATTRIBUTE_CHARGES, item.subType)
		end

		if item.actionId ~= -1 then
			tmpItem:setActionId(item.actionId)
		end

		if item.text and item.text ~= "" then
			tmpItem:setText(item.text)
		end

		local ret = self:addItemEx(tmpItem)
		if ret ~= RETURNVALUE_NOERROR then
			tmpItem:remove()
		end

		itemCount = itemCount - count
	end
	return true
end
```

</details>

MillhioreBT (approved):
LGTM

This would be a strange but very nice little feature for our custom monsters.

EPuncker (approved):

Issue #3600, opened 2021-08-25
List of missing TFS features from Cipbia up to version 10.98


EPuncker:
Just for reference, someone asked me to keep it here too. (for a more in-depth information, refer to [this thread](https://otland.net/threads/tfs-1-x-list-of-missing-tfs-features-from-cipbia.266762/))

## Disclaimer
As I said, **it is just for reference**, it doesn't mean we will be implementing any of it at all, but feel free to make a PR if you want to try.
## Features

- [ ] Implemented the **You cannot use objects that fast.** message if the player use an object more than 2 times per second. [8.72]
- [ ] All players now gain 10 second invisibility to creatures upon login, but as soon as the player has moved or attacked, creatures will again see the character. [9.5]
- [ ] In the case of disconnection from server, players now have a greater chance of survival. [9.5]
- Characters are now ignored by creatures for 30 seconds upon the time the server recognizes the client has lost connection. This leaves less time a character is vulnerable: 30 seconds (knight), 20 seconds (paladin) or 10 seconds (druid/sorcerer). Note that this is invisibility, not invincibility - area attacks will affect the player still.
- [ ] After dying and re-logging your server log is kept. [9.8] #3535 
- [ ] Reward system for bosses. [10.2]
- [ ] The Adventurer's Blessing was introduced. ([more info](https://tibia.fandom.com/wiki/Adventurer%27s_Blessing)) [10.55]
- [ ] Guild Bank. [10.7] #2214 
- [ ] Delayed spawn feature. [10.7]
- [ ] The Loyalty System was introduced. ([more info](https://tibia.fandom.com/wiki/Loyalty_System)) [10.74]
- [ ] The Store is introduced along with its currency Tibia Coins. [10.8]
- [ ] [Gold Pouch](https://tibia.fandom.com/wiki/Gold_Pouch) added [10.97]

completed:
- [x] Lured creatures walk back to their spawn places as soon as they lose their target. [8.7] done at #3604
- [x] The Depot locker was remade ([pic here](https://i.imgur.com/QlxLvzB.png)). It now contains a Depot Chest(vol: 18) (with 18 numbered depot boxes inside, each one hold infinite amount of items, paged each 32 items, and limited only by the depot limit being free or premy), Inbox (where parcels, letters, Market purchases/expired offers, house items go, and you can't move any item inside it/to it). [9.4]
[...]

PR #3612, opened 2021-08-28
Make monsters walk to spawn by chunked path

marmichalski:
### Pull Request Prelude

### Changes Proposed

Monster will now think a little more about where he wants to go. Also added lua function that will force the monster to walk back, even when config is set to 0. Config value is used for lured monsters, also added a bit more checks about summons/not-placed-by-map-monsters.

**Issues addressed:** #3600.

Supersedes PR: #3610.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -211,11 +211,11 @@ void Creature::onWalk()
 				forceUpdateFollowPath = true;
 			}
 		} else {
+			stopEventWalk();
+
 			if (listWalkDir.empty()) {
 				onWalkComplete();
 			}
-
-			stopEventWalk();
 		}
 	}
 
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -482,17 +482,13 @@ void Monster::onCreatureLeave(Creature* creature)
 	//update targetList
 	if (isOpponent(creature)) {
 		removeTarget(creature);
-		if (targetList.empty()) {
+		updateIdleStatus();
+
+		if (!isSummon() && targetList.empty()) {
 			int32_t walkToSpawnRadius = g_config.getNumber(ConfigManager::DEFAULT_WALKTOSPAWNRADIUS);
 			if (walkToSpawnRadius > 0 && !Position::areInRange(position, masterPos, walkToSpawnRadius, walkToSpawnRadius)) {
-				std::vector<Direction> dirList;
-				if (getPathTo(masterPos, dirList, 0, 0, true, true)) {
-					startAutoWalk(dirList);
-					return;
-				}
+				walkToSpawn();
 			}
-
-			updateIdleStatus();
 		}
 	}
 }
@@ -681,10 +677,6 @@ void Monster::updateIdleStatus()
 		idle = std::find_if(conditions.begin(), conditions.end(), [](Condition* condition) {
 			return condition->isAggressive();
 		}) == conditions.end();
-	}
-
-	if (idle) {
-		idle = listWalkDir.empty();
 	}
 
 	setIdle(idle);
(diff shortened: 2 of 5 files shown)


src/creature.cpp
```diff
@@ -211,11 +211,11 @@ void Creature::onWalk()
 				forceUpdateFollowPath = true;
 			}
 		} else {
+			stopEventWalk();
+
 			if (listWalkDir.empty()) {
 				onWalkComplete();
 			}
-
-			stopEventWalk();
```
marmichalski, line 218: Walk should actually be completed (`onWalkCompleted`) after the event is stopped. 🙄
This also makes it possible for monster to walk by smaller paths.

src/monster.cpp
```diff
@@ -482,17 +482,13 @@ void Monster::onCreatureLeave(Creature* creature)
 	//update targetList
 	if (isOpponent(creature)) {
 		removeTarget(creature);
-		if (targetList.empty()) {
+		updateIdleStatus();
+
+		if (!isSummon() && targetList.empty()) {
 			int32_t walkToSpawnRadius = g_config.getNumber(ConfigManager::DEFAULT_WALKTOSPAWNRADIUS);
 			if (walkToSpawnRadius > 0 && !Position::areInRange(position, masterPos, walkToSpawnRadius, walkToSpawnRadius)) {
```
MillhioreBT, line 489: the Position::areInRange overload is unnecessary, we already have: `Position::areInRange<x, y, z>(pos1, pos2)` you should delete it and use the one that already exists.

src/monster.cpp
```diff
@@ -1024,11 +1016,41 @@ void Monster::onThinkYell(uint32_t interval)
...
+		return false;
+	}
+
+	int32_t distance = std::max(Position::getDistanceX(position, masterPos), Position::getDistanceY(position, masterPos));
+	if (distance == 0) {
+		return false;
+	}
+
+	std::vector<Direction> dirList;
+	if (!getPathTo(masterPos, dirList, 0, std::max<int32_t>(0, distance - 5), true, true, distance)) {
```
marmichalski, line 1031: Last argument, `distance`, should probably be removed, forgot to cleanup. Will do tomorrow. 💩

src/monster.cpp
```diff
@@ -1024,11 +1016,41 @@ void Monster::onThinkYell(uint32_t interval)
 	}
 }
 
+bool Monster::walkToSpawn()
+{
+	if (walkingToSpawn || !spawn || !targetList.empty()) {
+		return false;
+	}
+
+	int32_t distance = std::max(Position::getDistanceX(position, masterPos), Position::getDistanceY(position, masterPos));
```
MillhioreBT, line 1025: use templates `std::max<int32_t>`

src/monster.cpp
```diff
@@ -1024,11 +1016,41 @@ void Monster::onThinkYell(uint32_t interval)
...
+		return false;
+	}
+
+	std::vector<Direction> dirList;
+	if (!getPathTo(masterPos, dirList, 0, std::max<int32_t>(0, distance - 5), true, true, distance)) {
+		return false;
+	}
+
+	walkingToSpawn = true;
+	startAutoWalk(dirList);
```
MillhioreBT, line 1036: don't use a temp vector if you have access to the object's private data
example:
```c++
listWalkDir.clear();
if (getPathTo(masterPos, listWalkDir, 0, 0, true, true)) {
	startAutoWalk();
}
```

if in the future you want to use getPathTo from a place where you don't have access to a creature's private data, then use the temporary vector, but don't forget to move it correctly with `std::move()`

This is an optimization that was done a few months ago, please use it correctly.

MillhioreBT (changes requested):

src/monster.cpp
marmichalski: You can't use it for non const value 🙄

MillhioreBT: are you changing the value of `DEFAULT_WALKTOSPAWNRADIUS`?
```c++
const int32_t walkToSpawnRadius = g_config.getNumber(ConfigManager::DEFAULT_WALKTOSPAWNRADIUS);
```

marmichalski: Nice. Now try to compile it.

src/monster.cpp
marmichalski: It actually has to stay.

EPuncker (approved):

PR #3601, opened 2021-08-25
Add ignore spawn block monster flag

marmichalski:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

**Issues addressed:** #3600 (spawn block, black knight)
**Supersedes:** #2964

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #3617, opened 2021-08-30
Add base no pong kick time to vocations

marmichalski:
### Pull Request Prelude

### Changes Proposed

Allows to set base no pong kick time per vocation, default is 60 seconds (60000).

**Issues addressed:** #3600

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -881,13 +881,24 @@ void Player::sendPing()
 		setAttackedCreature(nullptr);
 	}
 
-	if (noPongTime >= 60000 && canLogout()) {
-		if (g_creatureEvents->playerLogout(this)) {
-			if (client) {
-				client->logout(true, true);
-			} else {
-				g_game.removeCreature(this, true);
-			}
+	int32_t noPongKickTime = vocation->getNoPongKickTime();
+	if (pzLocked && noPongKickTime < 60000) {
+		noPongKickTime = 60000;
+	}
+
+	if (noPongTime >= noPongKickTime) {
+		if (isConnecting || getTile()->hasFlag(TILESTATE_NOLOGOUT)) {
+			return;
+		}
+
+		if (!g_creatureEvents->playerLogout(this)) {
+			return;
+		}
+
+		if (client) {
+			client->logout(true, true);
+		} else {
+			g_game.removeCreature(this, true);
 		}
 	}
 }
@@ -3745,23 +3756,6 @@ bool Player::hasOutfit(uint32_t lookType, uint8_t addons)
 		}
 	}
 	return false;
-}
-
-bool Player::canLogout()
-{
-	if (isConnecting) {
-		return false;
-	}
-
-	if (getTile()->hasFlag(TILESTATE_NOLOGOUT)) {
-		return false;
-	}
-
-	if (getTile()->hasFlag(TILESTATE_PROTECTIONZONE)) {
-		return true;
-	}
-
-	return !isPzLocked() && !hasCondition(CONDITION_INFIGHT);
 }
 
 void Player::genReservedStorageRange()
--- a/data/XML/vocations.xml
+++ b/data/XML/vocations.xml
@@ -10,7 +10,7 @@
 		<skill id="5" multiplier="1.5" />
 		<skill id="6" multiplier="1.1" />
 	</vocation>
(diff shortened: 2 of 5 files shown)


src/player.cpp
```diff
@@ -3747,23 +3753,6 @@ bool Player::hasOutfit(uint32_t lookType, uint8_t addons)
 	return false;
 }
 
-bool Player::canLogout()
```
marmichalski, line 3750: This method has only been used for no-pong kick.

src/player.cpp
```diff
@@ -3747,23 +3753,6 @@ bool Player::hasOutfit(uint32_t lookType, uint8_t addons)
...
-
-	if (getTile()->hasFlag(TILESTATE_NOLOGOUT)) {
-		return false;
-	}
-
-	if (getTile()->hasFlag(TILESTATE_PROTECTIONZONE)) {
-		return true;
-	}
-
-	return !isPzLocked() && !hasCondition(CONDITION_INFIGHT);
```
marmichalski, line 3764: I have not transfered the in-fight check, as this was keeping player online forever 🤷

I do not know if being pz-locked disabled the kick, @EPuncker?

EPuncker: in pz locked situation the kick time is always 60 seconds

marmichalski: alright, i'll update it to work like that later 💩

src/player.cpp
```diff
@@ -881,13 +881,19 @@ void Player::sendPing()
...
 	}
 
-	if (noPongTime >= 60000 && canLogout()) {
-		if (g_creatureEvents->playerLogout(this)) {
-			if (client) {
-				client->logout(true, true);
-			} else {
-				g_game.removeCreature(this, true);
-			}
+	if (noPongTime >= (pzLocked ? 60000 : 60000 + vocation->getNoPongKickTimeChange()))) {
```
marmichalski, line 884: Well, I guess it should be std::max here, since someone can increase the kick time for given vocation 🤔

And perhaps no pong kick with pz lock should be configurable.

Zbizu:
`noPongKickTime` in xml instead of difference would be more intuitive to work with

marmichalski:
> `noPongKickTime` in xml instead of difference would be more intuitive to work with

Dunno, one day you might want to implement mount reducing kick time, an item, outfit, player attribute, whatever. Unless you want to treat vocation as a base for kick time?

mattyx14 (approved):

Zbizu:
> > `noPongKickTime` in xml instead of difference would be more intuitive to work with
> 
> Dunno, one day you might want to implement mount reducing kick time, an item, outfit, player attribute, whatever. Unless you want to treat vocation as a base for kick time?

I meant vocation as a base.

marmichalski:
I personally do not agree with vocation being the base. Nevertheless, I've updated it 🙄

DSpeichert (commented):
I'm unsure about this. All looks good but how do we decide if something should be per-vocation-configurable instead of globally configurable?
We could have experience/loot/skill rate per-vocation, or pvp-mode per vocation. Is there a particular use case this makes sense per-vocation?

soul4soul:
> 
> 
> I'm unsure about this. All looks good but how do we decide if something should be per-vocation-configurable instead of globally configurable?
> We could have experience/loot/skill rate per-vocation, or pvp-mode per vocation. Is there a particular use case this makes sense per-vocation?

This feature is to mirror cip behavior at the vocation level. From the mentioned issue


> In the case of disconnection from server, players now have a greater chance of survival. [9.5]
> Characters log off after a shorter period of time depending on vocation: a knight (or characters of no vocation) will log off after 60 seconds, a paladin 50 and a druid/sorcerer 40. Before, it was 60 seconds for all vocations.

DSpeichert:
> This feature is to mirror cip behavior at the vocation level. From the mentioned issue

I wasn't aware of this. Makes perfect sense now.

DSpeichert (approved):

marmichalski:
> We could have experience/loot/skill rate per-vocation, [...]

No, but we could have the ability to modify these rates by vocation.

EPuncker (approved):

soul4soul:
It would be nice to have some documentation defining the difference between breaking changes and change in behavior. It should be more clear to contributors what the difference between the two are and what is accepted under each scenario. Bonus points to include the differences between TFS 1.X series vs future TFS 2.0. 

Today It seems random and is loosely gauged by perceived impact. For example in diff #3601, the justification to change BK flags had none and the justification not to add the flag to other monsters was because of "breaking change". In this diff we are changing the kick times per vocation with no justification.

In my own mind, I have only been thinking about breaking change as breaking the scripting API which should be avoided for TFS 1.x series. Nealy everything else I was considering a change in behavior meaning the script interface is the same or expanded and some part of the system now works differently.

PR #3635, opened 2021-09-08
Removed unused 'rateSpawn' setting

EPuncker:
### Pull Request Prelude

### Changes Proposed
This setting name is misleading and it actually **breaks** a behavior instead of making anything else...

for example: if you make a spawn with 5 rats with spawn time of 60 seconds and kill all of them at same time:

actual behavior: rats will respawn one by one every 60 seconds

behavior after this commit: rats will respawn at same time after 60 seconds (like rl tibia AND as anyone would expect)


gerotib:
I think that its wrong if you dont kill all the monsters at the same time.

Example:
- two monsters in single spawn with respawn time 60 seconds
- kill one of them and then second one after 50 seconds
- change floor so they can respawn
- both monsters will respawn after 10 seconds (counting from first kill)

Second monster will be spawned only after 10 seconds (or less) after being killed, which is not correct I think.

EPuncker:
> I think that its wrong if you dont kill all the monsters at the same time.
> 
> Example:
> 
> * two monsters in single spawn with respawn time 60 seconds
> * kill one of them and then second one after 50 seconds
> * change floor so they can respawn
> * both monsters will respawn after 10 seconds (counting from first kill)
> 
> Second monster will be spawned only after 10 seconds (or less) after being killed, which is not correct I think.

that is actually true, thanks for the detailed information. this will be put on hold until further review

whatthefs:
everytime i visit this repository i am fucking baffled. what the fuck is this change? jesus fucking christ what has happened to the quality of this project

had to create a github account just to comment on this abomination, and there's so many ridicilous changes like this recently

PR #3645, opened 2021-09-14
Fix getDirectionTo method

### LINKED ISSUE #3643: ultimate spells area bug

1. walk around and cast area spell (mas flam/mas tera)

Note: strike, wave and beam spells work fine for some reason

### Expected behaviour
spells get cut properly when walls block them

### Actual behaviour
the spell thinks, the caster is at x-1, y-1 from real caster position
obraz
obraz
obraz
obraz

### Reproduction environment
master branch, recent builds

I'm aware that I might be the person who broke it (in #3493), but I'm not really sure what is the cause / how to fix it.
edit: posted a PR with a fix. I'm no longer getting this bug with these changes.

nekiro:
This pull request fixes getDirectionTo helper method, which returns some weird direction instead of none if from == to.

By the way it also fixes #3643

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -461,6 +461,10 @@ Position getNextPosition(Direction direction, Position pos)
 
 Direction getDirectionTo(const Position& from, const Position& to)
 {
+	if (from == to) {
+		return DIRECTION_NONE;
+	}
+	
 	Direction dir;
 
 	int32_t x_offset = Position::getOffsetX(from, to);


EPuncker (approved):

Zbizu (approved):
Solves the issue while also fixes getDirectionTo.

Has potential to break something that was relying on frompos=topos -> nw in the future, but we can change things to mind the correct implementation later.

PR #3648, opened 2021-09-14
Add missing 'params' field to pushInstantSpell table

emil92b:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 
params field was missing

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -925,7 +925,7 @@ void LuaScriptInterface::pushCombatDamage(lua_State* L, const CombatDamage& dama
 
 void LuaScriptInterface::pushInstantSpell(lua_State* L, const InstantSpell& spell)
 {
-	lua_createtable(L, 0, 6);
+	lua_createtable(L, 0, 7);
 
 	setField(L, "name", spell.getName());
 	setField(L, "words", spell.getWords());
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -933,6 +933,7 @@ void LuaScriptInterface::pushInstantSpell(lua_State* L, const InstantSpell& spel
 	setField(L, "mlevel", spell.getMagicLevel());
 	setField(L, "mana", spell.getMana());
 	setField(L, "manapercent", spell.getManaPercent());
+	setField(L, "params", spell.getHasParam());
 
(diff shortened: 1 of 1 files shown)


src/luascript.cpp
```diff
@@ -933,6 +933,7 @@ void LuaScriptInterface::pushInstantSpell(lua_State* L, const InstantSpell& spel
 	setField(L, "mlevel", spell.getMagicLevel());
 	setField(L, "mana", spell.getMana());
 	setField(L, "manapercent", spell.getManaPercent());
+	setField(L, "params", spell.getHasParam());
```
EPuncker, line 936: I think you need to change line 928 value from 6 to 7 too, since now table will have 7 values

MillhioreBT (commented):
I don't see much use for this, I think none, but if you have an example please I want to see
I like to learn from other people who know more ;)

EPuncker (approved):

src/luascript.cpp
```diff
@@ -925,14 +925,15 @@ void LuaScriptInterface::pushCombatDamage(lua_State* L, const CombatDamage& dama
...
-	lua_createtable(L, 0, 6);
+	lua_createtable(L, 0, 7);
 
 	setField(L, "name", spell.getName());
 	setField(L, "words", spell.getWords());
 	setField(L, "level", spell.getLevel());
 	setField(L, "mlevel", spell.getMagicLevel());
 	setField(L, "mana", spell.getMana());
 	setField(L, "manapercent", spell.getManaPercent());
+	setField(L, "params", spell.getHasParam());
```
Zbizu, line 936: `hasParam` is a boolean type, not number. I suggest renaming the field to `hasparam` and removing `> 0` from the other commit.

Zbizu (changes requested):

src/luascript.cpp
emil92b: i don't agree with the rename, unless you do it in spells.xml aswell
but il fix the other commit sure

Zbizu: understandable
looks good to me now

Zbizu (approved):

EPuncker (approved):

PR #3662, opened 2021-09-17
Fix potential crash in Player::getClientIcons

dbjorkholm:
Fixes a potential crash by checking if **tile** is not a null pointer in `Player::getClientIcons`


Zbizu (approved):
I can't imagine a scenario in which tile would be a nullptr, but it's good to have nonetheless

EPuncker (approved):

DSpeichert (approved):

PR #3664, opened 2021-09-17
Reduce broadcast range for container updates

Erza:
### Pull Request Prelude

### Changes Proposed

Containers can never be opened if a player is more than 1 SQM away from it, hence we are doing unnecessary work here by sending updates to clients that are too far away from the container

**Issues addressed:** 

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -221,7 +221,7 @@ bool Container::isHoldingItem(const Item* item) const
 void Container::onAddContainerItem(Item* item)
 {
 	SpectatorVec spectators;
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

DSpeichert (approved):

nekiro:
why are we looping twice though in these methods? Instead of once and execute both methods (sendUpdate, onUpdate)

Erza:
> why are we looping twice though in these methods? Instead of once and execute both methods (sendUpdate, onUpdate)

I don't know either, but it is like that all over the codebase. Once I read a comment somewhere on this repo about there being a possible race condition if it's not done this way? Might be wrong tho

PR #3669, opened 2021-09-18
Remove useless item transform constructors

Kamenuvol:
There is no need at all to have this. You cannot get those items normally, and there are other items with ```transformEquipTo```
that are not considered problems.


EPuncker:
will test it soon

DSpeichert:
It might be there to ensure a "valid" item is created when creating it by name instead of ID (where order of them could otherwise yield the "wrong" one). Probably doesn't hurt to leave it in there, even if it does not cover everything.

EPuncker (approved):
ok tested and works just fine, maybe it was needed years ago, but not anymore

ranisalt (approved):

EPuncker (approved):

Issue #3492, opened 2021-06-28
Update to version 12


Zbizu:
### Explanation of what you want to do that is currently impossible
I've heard that one of the biggest things that is preventing us from moving to version 12 was the lack of open sourced client.

What about this one?
https://github.com/opentibiabr/otclient/ (related commit: https://github.com/opentibiabr/otclient/commit/69a98992e36622eceb99a90c4f4d28ad59579ad6)

### Desired functionality
Getting out of year 2016.

### Available workarounds
using br/spanish server engines

### Prior art
not sure what to put here, is their protocolgame file ok?
https://github.com/opentibiabr/otservbr-global/blob/c6a1b67335cae2a4e2d389c436c71d3f3cde3ed7/src/server/network/protocol/protocolgame.cpp#L634

gesior:
> ### Prior art
> not sure what to put here, is their protocolgame file ok?
> https://github.com/opentibiabr/otservbr-global/blob/c6a1b67335cae2a4e2d389c436c71d3f3cde3ed7/src/server/network/protocol/protocolgame.cpp#L634

It's not. Most of that code gives random crashes as soon as you get 100+ online. More info: https://github.com/opentibiabr/otservbr-global/issues/2670
EVERY time I open otservbr code, I find new critical bug in 5 minutes! Every new system [bestiary/store/market/prey] they coded is sooo easy to crash/abuse/DoS.

That's a difference. TFS is stable. There are no new features, no rl map, but all other servers are based on it.

Will there be any 12+ engine with code quality like TFS? I don't think so. To create all features 10.99 -> 12.x, it will take ~2000 hours. Who will spend 2000 hours on "open source" project that is mostly used to generate money?
I get orders from few popular 12+ servers: "fix me X crash bug in otservbr, but DONT SHARE IT with others". Around 20% of OTS owners are interested in sharing fixes, rest is ready to pay more, to make bug fixes only for them. They use open source software, but they are against open source community.

lgrossi:
I love how you sustain that a vodebase with 0% test coverage and 153 open issues without any real active development work going on is stable.
Without statistics and real data, and also an constructive attitude your comment is just toxic, trying to discredit people that try to help. This is as bad to open source community as leeshers and otadmins that pay to not share. 
That's one of the main reasons why open tibia is dead, however it's to late to you guys to wake up and see it. 

R.I.P. ❤

Zbizu:
[...]

PR #3671, opened 2021-09-19
early support for web login

Zbizu:
### Pull Request Prelude

### Changes Proposed
- web login support (standalone login with 10.98 client still possible with account name)
- search both `name` and `email` columns when logging in (login with email is webservice only)
- changed the way of handling checksums

**Issues addressed:** one step closer to #3492

### Notes
I know these changes are far from perfect, therefore I'm waiting for reviews and opinions
Changes based on [opentibiabr/canary](https://github.com/opentibiabr/canary/) code


Fukotland:
zbizu na prezydenta

Zbizu:
Fixed the bugs and walked a bit around the temple.

7.6, 8.6, 9.7, 10.99, 12.71: only client with protocol 10.98 allowed
10.98 vanilla: OK (walked a bit as a player, killed 2 rats, eaten a cheese, cast light healing, life is good)
10.98 otc: OK (performed same test as above)

Waiting for a review.

src/connection.h
```diff
@@ -68,13 +82,20 @@ class Connection : public std::enable_shared_from_this<Connection>
...
+			writeTimer(init_io_service),
+			service_port(std::move(init_service_port)),
+			socket(init_io_service) {
+			connectionState = CONNECTION_STATE_PENDING;
+			packetsSent = 0;
+			timeConnected = time(nullptr);
+			receivedFirst = false;
+			serverNameTime = 0;
+			receivedName = false;
+			receivedLastChar = false;
```
ranisalt, line 97: What's the point of using the body of the constructor instead of using initialization? Also just use default values for those trivial values (false and 0)

ranisalt (changes requested):

src/connection.h
nekiro: I would say there is no point and I hate it

Zbizu: I pushed a commit and performed same tests as earlier with same results. Is it ok now?

src/connection.cpp
```diff
@@ -148,6 +157,35 @@ void Connection::parseHeader(const boost::system::error_code& error)
 		return;
 	}
 
+	if (!receivedLastChar && connectionState == CONNECTION_STATE_CONNECTING_STAGE2) {
+		uint8_t* msgBuffer = msg.getBuffer();
+
+		if (!receivedName && msgBuffer[1] == 0x00) {
+			receivedLastChar = true;
+		} else {
+			std::string serverName = g_config.getString(ConfigManager::SERVER_NAME) + "\n";
```
nekiro, line 166: this is not used?

src/connection.cpp
```diff
@@ -182,32 +220,25 @@ void Connection::parsePacket(const boost::system::error_code& error)
...
-	}
+	// Read potential checksum bytes
+	uint32_t recvPacket = msg.get<uint32_t>();
 
 	if (!receivedFirst) {
-		// First message received
 		receivedFirst = true;
 
 		if (!protocol) {
+			// Skip deprecarted checksum bytes (with clients that aren't using it in mind)
```
nekiro, line 234: deprecated*

src/connection.cpp
```diff
@@ -182,32 +220,25 @@ void Connection::parsePacket(const boost::system::error_code& error)
...
+			uint16_t len = msg.getLength();
+			if (len < 280 && len != 151) {
```
nekiro, line 236: I would like to know what are these numbers xD

Zbizu: 151 is packet length for 8.6 and 9.7 (and probably all the clients in the middle)
clients that use checksums have packet lengths larger than 280 although I'm not sure how was that figured out

src/connection.cpp
Zbizu: Leftover from removing a check for matching server name.

src/connection.cpp
```diff
@@ -108,21 +105,33 @@ void Connection::accept(Protocol_ptr protocol)
 {
 	this->protocol = protocol;
 	g_dispatcher.addTask(createTask(std::bind(&Protocol::onConnect, protocol)));
+	connectionState = CONNECTION_STATE_CONNECTING_STAGE2;
```
ranisalt, line 108: Can we get more descriptive names for these stages?

src/connection.cpp
```diff
@@ -108,21 +105,33 @@ void Connection::accept(Protocol_ptr protocol)
...
+		if (!receivedLastChar && receivedName && connectionState == CONNECTION_STATE_CONNECTING_STAGE2) {
+			// Read size of the first packet
+			boost::asio::async_read(socket,
+				boost::asio::buffer(msg.getBuffer(), 1),
+				std::bind(&Connection::parseHeader, shared_from_this(), std::placeholders::_1));
+		} else {
+			// Read size of the first packet
+			boost::asio::async_read(socket,
+				boost::asio::buffer(msg.getBuffer(), NetworkMessage::HEADER_LENGTH),
+				std::bind(&Connection::parseHeader, shared_from_this(), std::placeholders::_1));
```
ranisalt, line 133: You can use a lambda here, I think it will be way cleaner, e.g.

```cpp
		auto bufferLength = !receivedLastChar && receivedName && connectionState == CONNECTION_STATE_CONNECTING_STAGE2 ?  1 : NetworkMessage::HEADER_LENGTH
			// Read size of the first packet
		boost::asio::async_read(socket,
			boost::asio::buffer(msg.getBuffer(), bufferLength),
			[=, connection = shared_from_this()](auto buf) { connection->parseHeader(buf); });
```

Not 100% sure about the syntax but you'll get the idea

src/connection.h
```diff
@@ -121,8 +135,12 @@ class Connection : public std::enable_shared_from_this<Connection>
 		time_t timeConnected;
 		uint32_t packetsSent = 0;
 
-		bool closed = false;
+		int8_t connectionState = CONNECTION_STATE_PENDING;
```
ranisalt, line 138:
```suggestion
		ConnectionState_t connectionState = CONNECTION_STATE_PENDING;
```

src/connection.cpp
Zbizu: I don't understand this enough to make that change. I'm not that good at c++

ranisalt: I believe it is right now, just copy it over the whole if/else block

src/connection.cpp
EPuncker:
```
			CONNECTION_STATE_OPEN,
			CONNECTION_STATE_IDENTIFYING,
			CONNECTION_STATE_READINGS,
			CONNECTION_STATE_CLOSED
```

PR #3673, opened 2021-09-20
unnecessary loop

Diego-OT:
### Pull Request Prelude

### Changes Proposed

remove same loops for same situations

**Issues addressed:** 

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -568,9 +568,6 @@ bool Game::placeCreature(Creature* creature, const Position& pos, bool extendedP
 		if (Player* tmpPlayer = spectator->getPlayer()) {
 			tmpPlayer->sendCreatureAppear(creature, creature->getPosition(), true);
 		}
-	}
-
-	for (Creature* spectator : spectators) {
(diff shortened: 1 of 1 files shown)


MillhioreBT:
If I'm not mistaken these loops are separated because the players need to update certain things of the creature that appears before another event called from onAppear modifies it.

At least that's the only reason I see two loops there, otherwise multiple players could get different properties of the creature that appears if it is modified by some creature that captures the event before another player next in the vector can update it

I was wondering the same thing myself when I saw two loops there, but then I remembered that the guys at TFS are not dumb.

Diego-OT:
Hm, you may be right but I couldn't see the reason of these duplicated loops (Im still learning). I assume that there are places that need two loops (in case of Game::removeCreature but I didn't touch), but looking into that case below, we can see that vector spectators only sending msgs (nothing related in update something):

https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L3514

DSpeichert:
Sounds like this isn't correct then?

PR #3677, opened 2021-09-22
Add missing brackets

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed

- Add missing brackets.

### DIFF
--- a/src/bed.cpp
+++ b/src/bed.cpp
@@ -61,8 +61,9 @@ Attr_ReadValue BedItem::readAttr(AttrTypes_t attr, PropStream& propStream)
 			return ATTR_READ_CONTINUE;
 		}
 
-		default:
+		default: {
 			break;
+		}
 	}
 	return Item::readAttr(attr, propStream);
 }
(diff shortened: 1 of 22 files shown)


Zbizu:
While this for sure DO follow tfs code convention, I personally find brackets ugly on small switch statements. Perhaps an exception in the convention should be added?

MillhioreBT:
I love the brackets but I cannot deny that in some places it is not even necessary, as it only increases the illegibility of the code.

ramon-bernardo:
> While this for sure DO follow tfs code convention, I personally find brackets ugly on small switch statements. Perhaps an exception in the convention should be added?



> I love the brackets but I cannot deny that in some places it is not even necessary, as it only increases the illegibility of the code.

I prefer same-line return, but this manually = bad, i think.
Why dont we have something like clang-format here?

soul4soul:
> 
> 
> > While this for sure DO follow tfs code convention, I personally find brackets ugly on small switch statements. Perhaps an exception in the convention should be added?
> 
> > I love the brackets but I cannot deny that in some places it is not even necessary, as it only increases the illegibility of the code.
> 
> I prefer same-line return, but this manually = bad, i think. clang-format? hm

Check the clang-format branch, it's been abandoned and waiting for someone to bring it back to life. https://github.com/otland/forgottenserver/tree/clang_format

PR #3679, opened 2021-09-22
added isKey itemtype

emil92b:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 
now you can add numbers to a key using /i silver key, 1234 for example.

### DIFF
--- a/data/talkactions/scripts/create_item.lua
+++ b/data/talkactions/scripts/create_item.lua
@@ -26,10 +26,14 @@ function onSay(player, words, param)
 		return false
 	end
 
+	local keyNumber = 0
 	local count = tonumber(split[2])
 	if count then
 		if itemType:isStackable() then
 			count = math.min(10000, math.max(1, count))
+		elseif itemType:isKey() then
+			keyNumber = count
+			count = 1
 		elseif not itemType:isFluidContainer() then
 			count = math.min(100, math.max(1, count))
 		else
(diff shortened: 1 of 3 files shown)


EPuncker (dismissed):

MillhioreBT (approved):

Zbizu (approved):
tbh I expected this game to have more keys, but either way it delivered
obraz

new `/i` works fine (tested with stackable items, unstackable items, fluid containers and keys)

by the way, it's redundant as you can simply `itemType:getType() == ITEM_TYPE_KEY`

Zbizu (changes requested):
source edit is redundant as you can simply define `function ItemType:isKey() return self:getType() == ITEM_TYPE_KEY end`

DSpeichert (changes requested):
drop the C++ change in favor of Lua as zbizu pointed out

data/lib/core/itemtype.lua
```diff
@@ -14,3 +14,7 @@ local slotBits = {
 function ItemType.usesSlot(self, slot)
 	return bit.band(self:getSlotPosition(), slotBits[slot] or 0) ~= 0
 end
+
+function ItemType:isKey()
+	return self:getType() == ITEM_TYPE_KEY 
+end
```
DSpeichert, line 20: missing EOL

EPuncker, line 20:
```suggestion
end

```

PR #3691, opened 2021-09-25
Create ring attribute for monster spells

soul4soul:
### Pull Request Prelude

### Changes Proposed

While I really don't like the creature spell system and think it should be deprecated in favor of revscriptsys spells + revscriptsys monsters it was easy and convenient to make a ring attribute to support spells for the tibia monsters which have a ring shaped attack.

```
XML
<attack name="fire" interval="1000" chance="100" ring="2" target="0" min="-1" max="-2">
			<attribute key="shootEffect" value="fire" />
			<attribute key="areaEffect" value="firearea" />
		</attack>

LUA
monster.attacks = {
	{name = "combat", type = COMBAT_LIFEDRAIN, chance = 100, interval = 1000, ring = 4, target = true, minDamage = -1, maxDamage = -1, effect = CONST_ME_PURPLEENERGY}
}
```

I made ring=2 a cross because that is another common shape and radius=2 is already a plus shape and radius=3 is already an exori shape. With this addition we cover all the spell shapes used on tibiawiki.

Test with both XML and revscriptsys
riing2
ring3
ring4

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -339,6 +339,9 @@ registerMonsterType.attacks = function(mtype, mask)
 					if attack.radius then
 						spell:setCombatRadius(attack.radius)
 					end
+					if attack.ring then
+						spell:setCombatRing(attack.ring)
+					end
 					if attack.minDamage and attack.maxDamage then
 						if attack.name == "combat" then
 							spell:setCombatValue(attack.minDamage, attack.maxDamage)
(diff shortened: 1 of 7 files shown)


omarcopires:
Very good! Could you also add the spread?

EPuncker (approved):

soul4soul:
> 
> 
> Very good! Could you also add the spread?

What spread?

The ring attribute works just like radius so it can be used on self or on target with an optional range.

EPuncker:
> > Very good! Could you also add the spread?
> 
> What spread?
> 
> The ring attribute works just like radius so it can be used on self or on target with an optional range.

I think he means that effect that most of new monsters are using, their draw a line of effect to the player and anything between it takes damage

EPuncker:
also seems like it can target multiple players:
2021-09-25 19_01_26-(3) Tibia #375- HUNT Aparições Dark Thais- Mirrored Nightmare - (STREAMER Nawtsc
https://youtu.be/36w6sr24Z_M?t=2032

PR #3693, opened 2021-09-26
Fix nullptr dereference in action:register()

### LINKED ISSUE #3692: Can't run as Debug (Windows)

1. Open project in Visual Studio (2019)
2. Select Debug x64 configuration
3. Set working dir to ".."

### Expected behaviour
Should run just fine

### Actual behaviour
Acess memory violation o.O
image

### Environment
Windows 10 x64

nekiro:
g_actions->registerLuaEvent(action) frees action pointer, but then we proceed to dereference it, which is in first place redundant.

Probably fix #3692

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -15726,9 +15726,6 @@ int LuaScriptInterface::luaActionRegister(lua_State* L)
 			return 1;
 		}
 		pushBoolean(L, g_actions->registerLuaEvent(action));
-		action->clearActionIdRange();
-		action->clearItemIdRange();
-		action->clearUniqueIdRange();
 	} else {
 		lua_pushnil(L);
 	}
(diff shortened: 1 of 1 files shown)


ranisalt:
Same pattern exists when registering movements. Does that trigger the null pointer dereference?

In that case, those methods (`clearXxxIdRange`) are unused elsewhere and could be removed. However, I think that implies a leak.

nekiro:
> Same pattern exists when registering movements. Does that trigger the null pointer dereference?
> 
> In that case, those methods (`clearXxxIdRange`) are unused elsewhere and could be removed. However, I think that implies a leak.

There is also weird check there "isScripted" which does exactly the same thing as non scripted 🤔

hugoj-goncalves:
I know nothing about c++, the change allowed me to debug just fine

if it helps, the way it is in the @nekiro 's downport works without complains also (actions part)
image
https://github.com/nekiro/TFS-1.4-Downgrades/blob/58517dd44f21af683c1a60cfabfbb408ae640aee/src/luascript.cpp#L15092

EPuncker (approved):

ranisalt:
> In that case, those methods (`clearXxxIdRange`) are unused elsewhere and could be removed

Go ahead and clean them up :)

PR #3694, opened 2021-09-27
fix crash and leak in action:register and moveevent:register

yamaken93:
### Pull Request Prelude

### Changes Proposed
Better registerLuaEvent code, allow to register the same action/moveevent to aid, uid, item id and position.
No use after delete anymore in action:register and moveevent:register.
No memory leak/waste of memory since we borrow/steal the vector from the original event making the future copies copy empty vector.

**Issues addressed:** 
#3692

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -211,37 +211,45 @@ bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node)
 bool Actions::registerLuaEvent(Action* event)
 {
 	Action_ptr action{ event };
-	if (!action->getItemIdRange().empty()) {
-		const auto& range = action->getItemIdRange();
-		for (auto id : range) {
-			auto result = useItemMap.emplace(id, *action);
-			if (!result.second) {
-				std::cout << "[Warning - Actions::registerLuaEvent] Duplicate registered item with id: " << id << " in range from id: " << range.front() << ", to id: " << range.back() << std::endl;
-			}
+
+	auto itemIdRange = action->borrowItemIdRange();
+	auto uniqueIdRange = action->borrowUniqueIdRange();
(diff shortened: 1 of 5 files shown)


EPuncker (approved):
this fixed the crash for me

DSpeichert (approved):
I think "steal" fits better than "borrow" here. You have no intention of giving them back ;)

ranisalt (commented):
I'm unsure about this. If I recall correctly, using a variable after moving it is undefined behavior.

It would probably be better to create a new one on the stack, swap with the one on the current object, and return the one on the stack (it will elide the copy/move):

```cpp
std::vector newVector{};
std::swap([this->]oldVector, newVector);
return newVector;
```

MillhioreBT:
> 

I think the compiler knows if this should be returned with or without movement samantic, but since I can't confirm it, I think it would be best to do that @ranisalt suggested.

Codinablack:
> I'm unsure about this. If I recall correctly, using a variable after moving it is undefined behavior.
> 
> It would probably be better to create a new one on the stack, swap with the one on the current object, and return the one on the stack (it will elide the copy/move):
> 
> ```c++
> std::vector newVector{};
> std::swap([this->]oldVector, newVector);
> return newVector;
> ```

So basically something like this is needed to fix this PR

```c++
std::vector<uint32_t> borrowItemIdRange() {
    std::vector<uint32_t> temp;
    temp.swap(itemIdRange);
    return temp;
}
```

Right? If all the new "borrowed" methods looked like this, or along these lines (obviously variable names would change), then the PR would be solid?

Codinablack:
[4742](https://github.com/otland/forgottenserver/pull/4742) Solves this one, we can close.

PR #3699, opened 2021-09-28
Possibility of defining currencies

MillhioreBT:

### Changes Proposed
* Define new currencies or configure existing ones
* The npc bank has been ignored because it treats currencies very specifically and we still do not have a convention of what currencies should be called
* The method `Game.getCurrenyItems` returns a table with the `ItemType` of each currency, this table is ordered from lowest to highest , so the first currency will always be the most cheapest and the expensive will be the last
* Added a methods for `ItemType` and `Item`, this method is `getWorth`


data/scripts/actions/tools/gold_converter.lua
```diff
@@ -32,5 +28,12 @@ function goldConverter.onUse(player, item, fromPosition, target, toPosition, isH
 	return true
 end
 
-goldConverter:id(26378)
+local currencyItems = Game.getCurrenyItems()
+for index, currency in pairs(currencyItems) do
+	local back, to = currencyItems[index-1], currencyItems[index+1]
+	local currencyId = currency:getId()
+	config[currencyId] = { changeBack = back and back:getId(), changeTo = to and to:getId() }
+	goldConverter:id(currencyId)
```
EPuncker, line 36: this is the tool id, not currency ID 😛

MillhioreBT: ajajajjajaja yes, sorry ;)

data/lib/core/player.lua
```diff
@@ -171,40 +171,20 @@ function Player.canCarryMoney(self, amount)
...
-			totalWeight = totalWeight + type_gold:getWeight(count)
-			amount = amount - count
-			inventorySlots = inventorySlots + 1
+	local currencyItems = Game.getCurrencyItems()
+	for index = #currencyItems, 1, -1 do
+		local currency = currencyItems[index]
+		-- Add currency coins to totalWeight and inventorySlots
+		local worth = currency:getWorth()
+		local currencyCoins = math.floor(amount / worth)
+		if crystalCoins > 0 then
```
soul4soul, line 180:
```suggestion
		if currencyCoins > 0 then
```

src/items.cpp
```diff
@@ -1368,6 +1370,17 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 					break;
 				}
 
+				case ITEM_PARSE_WORTH: {
+					uint64_t worth = pugi::cast<uint64_t>(valueAttribute.value());
+					if (currencyItems.find(worth) != currencyItems.end()) {
+						std::cout << "[Warning - Items::parseItemNode] Duplicated currency item " << id << " with worth: " << worth << std::endl;
```
soul4soul, line 1376:
```suggestion
						std::cout << "[Warning - Items::parseItemNode] Duplicated currency worth. Item " << id << " redefines worth " << worth << std::endl;
```
Trying to make it more clear that the thing which is duplicate is the worth and not the item id.

src/items.cpp
```diff
@@ -1368,6 +1370,17 @@ void Items::parseItemNode(const pugi::xml_node& itemNode, uint16_t id)
 					break;
 				}
 
+				case ITEM_PARSE_WORTH: {
```
soul4soul, line 1373: Should a check be added to ensure worth is a power of 10? The change gold script won't work with any other values.

MillhioreBT: That does not concern us, whoever wants to implement an strange currency should know that he must create his own script to handle said currencies

soul4soul: It should concern us... Over the last few months many diffs were accepted to help those who want to run a server from making mistakes. This is another area we should protect server admins from making a mistake and dealing with the headache of figuring out what is wrong since the system only works with power's of 10. At a minimum that information could be added to the PR description. I myself thought I could use any value until reading the code.

PR #3700, opened 2021-09-28
addTask(callback, ...) lua function

yamaken93:
### Pull Request Prelude

### Changes Proposed

implements addTask lua c function so we can access the engine g_dispatcher.addTask function like addEvent and g_scheduler.addEvent.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1053,6 +1053,9 @@ void LuaScriptInterface::registerFunctions()
 	//stopEvent(eventid)
 	lua_register(luaState, "stopEvent", LuaScriptInterface::luaStopEvent);
 
+	// addTask(callback, ...)
+	lua_register(luaState, "addTask", LuaScriptInterface::luaAddTask);
+
 	//saveServer()
 	lua_register(luaState, "saveServer", LuaScriptInterface::luaSaveServer);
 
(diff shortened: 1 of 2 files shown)


Erza:
Can you describe an example scenario where having this functionality would be useful?

yamaken93:
> Can you describe an example scenario where having this functionality would be useful?
If  you want the function(task) to be executed in the next frame(dispatcher cycle) asap without any delay.

Erza:
> > Can you describe an example scenario where having this functionality would be useful?
> 
> If you want the function(task) to be executed in the next frame(dispatcher cycle) asap without any delay.

Sure, but in which example scenario would you want to do that?

Kamenuvol:
addEvent with 1ms is not the same?

MillhioreBT:
the dispatcher executes tasks in another thread is it correct?

nekiro:
> the dispatcher executes tasks in another thread is it correct?

Yes, dispatcher and scheduler are using their own thread.

yamaken93:
I think i may drop this pr and then change addEvent to use g_dispatcher.addTask when the delay is 0.

yamaken93:
So we should add addTask or internally when someone call addEvent with 0 interval parameter it should add the task to dispatcher?

DSpeichert:
I'd say the latter.

yamaken93:
I still think we should add this addTask function. Why? Its problematic to change this ancient addEvent function parameters, since its has been a given for ages the minimum delay is 100 ms and its also problematic to under the hood make addEvent to add the task to dispatcher instead of scheduler because for the API sake the user script should be able to call addEvent and in the next line  call stopEvent and its not possible or easier to implement that dispatcher has cancellable tasks.
So the only two options for me is, add this addTask function or remove the 100 ms minimum delay from addEvent which will work basically as adding the task to the dispatcher when delay is 0 but will keep the stopEvent function working.

DSpeichert:
Remove the minimum delay?

Wasn't there a reason this was implemented though?

yamaken93:
> Remove the minimum delay?
> 
> Wasn't there a reason this was implemented though?

Probably a protection for performance reasons.

yamaken93:
@nekiro @DanielChabrowski @MillhioreBT @ranisalt I'm willing to finally finish this. We should decide, add this api or remove the minimum delay from addEvent or add something like addEventEx whithout the minimum delay.

MillhioreBT:
> @nekiro @DanielChabrowski @MillhioreBT @ranisalt I'm willing to finally finish this. We should decide, add this api or remove the minimum delay from addEvent or add something like addEventEx whithout the minimum delay.

I like `addTask` and am using it in some of my scripts. If `addEventEx` works all the same I have no complaints about it, I'll take it.

yamaken93:
> > @nekiro @DanielChabrowski @MillhioreBT @ranisalt I'm willing to finally finish this. We should decide, add this api or remove the minimum delay from addEvent or add something like addEventEx whithout the minimum delay.
> 
> I like `addTask` and am using it in some of my scripts. If `addEventEx` works all the same, I have no complaints about it, I'll take it.

addEventEx would accept below 100 ms delay as parameter so it can work like addTask (with more overhead thanks to the timer system) if the delay is 0 and you will be able to cancel it if you call stopEvent before the task is added to the dispatcher.

PR #3705, opened 2021-09-29
Another type of search for Game.getSpectators

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
New type of search for the method `Game.getSpectators`
Types:
* `GROUND_FLOOR`: This represents what was previously false, we search all viewers on the same level Z.
* `UNDERGROUND_FLOOR`: This represents what was true before, we look for all the spectators in all the Z levels.
* `SURFACE_GROUND`: This is a new type of search, we search all spectators from level 0 to 7 only.

**Issues addressed:** #3322

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -712,7 +712,7 @@ void Combat::doCombat(Creature* caster, Creature* target) const
 	} else {
 		if (!params.aggressive || (caster != target && Combat::canDoCombat(caster, target) == RETURNVALUE_NOERROR)) {
 			SpectatorVec spectators;
-			g_game.map.getSpectators(spectators, target->getPosition(), true, true);
+			g_game.map.getSpectators(spectators, target->getPosition(), UNDERGROUND_FLOOR, true);
 
 			if (params.origin != ORIGIN_MELEE) {
 				for (const auto& condition : params.conditionList) {
@@ -779,7 +779,7 @@ void Combat::doCombat(Creature* caster, const Position& position) const
 
 		const int32_t rangeX = maxX + Map::maxViewportX;
 		const int32_t rangeY = maxY + Map::maxViewportY;
-		g_game.map.getSpectators(spectators, position, true, true, rangeX, rangeX, rangeY, rangeY);
+		g_game.map.getSpectators(spectators, position, UNDERGROUND_FLOOR, true, rangeX, rangeX, rangeY, rangeY);
 
 		postCombatEffects(caster, position, params);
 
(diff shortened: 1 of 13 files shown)


Diego-OT:
Suggestions: 

UNDERGROUND_FLOOR -> ALL_FLOORS
SURFACE_GROUND -> SURFACE_FLOORS
GROUND_FLOOR -> SAME_FLOOR

kornholi:
`UNDERGROUND_FLOOR` is misleading as it's only looking at two floors above and below while underground, not all of them. I don't have a better name, though.

Xikini:
Some alternatives for UNDERGROUND_FLOOR

EXTENDED_FLOORS
QUINTUPLE_FLOORS
MULTI_FLOOR
5_FLOOR_SEARCH
NEARBY_FLOORS

All of the above in arbitrary anyway.
Why can't we change it to be user defined instead of hardcoded?
range_z,

MillhioreBT:
> Some alternatives for UNDERGROUND_FLOOR
> 
> EXTENDED_FLOORS QUINTUPLE_FLOORS MULTI_FLOOR 5_FLOOR_SEARCH NEARBY_FLOORS
> 
> All of the above in arbitrary anyway. Why can't we change it to be user defined instead of hardcoded? range_z,

I've been thinking about it, already @yamaken93 suggested it too, it's not a bad idea so I'll think about it, I would also like to be able to define the Z range manually, I was just trying to keep the cache in the viewers, but I think because of how they go things, that cache will probably be removed later as apparently it is useless

ramon-bernardo:
Dont forget `getSpectators` in
```
data/lib/compat/compat.lua
data/actions/scripts/other/die.lua
```

DSpeichert (changes requested):
Here are my name picks:

* `SAME_FLOOR` - that one is easy
* `ABOVE_GROUND` - because it's not just ground floor, it's all above it (floors 0-7)
* `UNDER_GROUND` - the special case that is weird and can't be summarized concisely.

Specifically, I think `GROUND_FLOOR` is misleading.

Codinablack:
SUB_SURFACE would work for the third param and I think its pretty clear and concise in what it does, especially with the other options being like Dspeichert said:

SAME_FLOOR - that one is easy
ABOVE_GROUND - because it's not just ground floor, it's all above it (floors 0-7)

Anyway it goes, if it gets changed with new params, We can just update the wiki to make sure its clear for what is each params intended use 

**Alternatively**, we could just include a min and max for the z,  like the x and y already has, with the minimum being the lowest floor to search, and max being the top floor to search, minimum defaulting to the current floor, and maximum defaulting to highest floor in map or same floor as well.

ranisalt:
> I think a raw api which allows you set from position and to position is good and versatile.

I second this. Trying to imagine every scenario will just lead to duplication and confusion.

PR #3707, opened 2021-09-30
Removing automatic shutdown from launcher

omarcopires:
### Pull Request Prelude

### Changes Proposed

When presenting any problems at startup, the executable closed automatically and this made it difficult to read the console, having to generate a log in another way, with this change this no longer happens.


MillhioreBT:
Currently we can see the registry if we run the server from a terminal and not just executing the file directly with a double click, I don't know if it's a good idea or the right thing to do in TFS not to turn off the scheduler when the server die

I just want to believe that those `shutdown` are there for some reason in the first place, otherwise then what are they for?

Zbizu:
It's easier for restarter on Linux I believe

PR #3737, opened 2021-10-15
Replace with iterator on updateItemWeight function

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Use iterator instead of recursive function on updateItemWeight.

I know it depends on countless variables to fail, but it's just suggestion/questioning.

@SaiyansKing 

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -149,8 +149,9 @@ bool Container::unserializeItemNode(OTB::Loader& loader, const OTB::Node& node, 
 void Container::updateItemWeight(int32_t diff)
 {
 	totalWeight += diff;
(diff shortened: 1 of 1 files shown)


ramon-bernardo:
> When exactly this could happen?

What I can say is: it depends. It may happen? [maybe](https://stackoverflow.com/a/30209404).

DSpeichert:
I think you should start with a bug report that explains why this change would make sense before jumping to a PR.

ramon-bernardo:
Testing with the compilation flags -O3 -std=c++20 -stdlib=libstdc++, i obtained the same result using iteration or recursion, the compiled code is the same. :anguished

PR #3740, opened 2021-10-16
Fix int_fast16_t overflow

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Fix int_fast16_t overflow


soul4soul (dismissed):

DSpeichert:
From the commit message, PR message and the code, I still have no idea what the problem is and what the fix fixes.

ranisalt:
> From the commit message, PR message and the code, I still have no idea what the problem is and what the fix fixes.

The range arguments are int32, but the result variable is int16, therefore overflow may occur if the range does not fit 16 bits. It doesn't usually happen, because the range is usually small (this function is heavy af, it usually runs with the viewport dimensions).

DSpeichert:
@ranisalt `int` or `int32_t` to keep it the same?

ranisalt:
`auto` to avoid conversions, but `int32_t` between those

DSpeichert (approved):

EPuncker (approved):

PR #3749, opened 2021-10-18
Bumped engine version to 1.5

EPuncker:
1.5 will be the development stage, and the next release will be 1.6

### Pull Request Prelude

### DIFF
--- a/src/definitions.h
+++ b/src/definitions.h
@@ -21,7 +21,7 @@
 #define FS_DEFINITIONS_H_877452FEC245450C9F96B8FD268D8963
 
 static constexpr auto STATUS_SERVER_NAME = "The Forgotten Server";
-static constexpr auto STATUS_SERVER_VERSION = "1.4";
+static constexpr auto STATUS_SERVER_VERSION = "1.5";
 static constexpr auto STATUS_SERVER_DEVELOPERS = "Mark Samman";
 
 static constexpr auto CLIENT_VERSION_MIN = 1097;


Zbizu:
imo it's worth to apply some bugfixes before closing that chapter

EPuncker:
> imo it's worth to apply some bugfixes before closing that chapter

it wont be closed, it will still receive bug fixes, we are changing it to 1.5 just to differentiate release from development

Znote:
Hmm, is it possible to make it 1.5_YearMonth? (Where YearMonth represents the time it was compiled, or last commit?)

> Maybe we should just put `master branch` as a value there and not skip versions?
> 
> I know we discussed that before and that was agreed upon, however using 1.5 as undetermined/unreleased indicator is still not helpful. It needs to be replaced at release time regardless.

Technically, this would make it more difficult, because we wouldn't be able to easily distinguish etc version 1.5 and 1.7 if they are both development versions just known as master. Then I would flag it at least as 1.5_dev or 1.5_master or 1.5_git to emphasis that its in the middle of a dev process.

Zbizu:
You're overcomplicating this.

DSpeichert:
> Technically, this would make it more difficult, because we wouldn't be able to easily distinguish etc version 1.5 and 1.7 if they are both development versions just known as master. Then I would flag it at least as 1.5_dev or 1.5_master or 1.5_git to emphasis that its in the middle of a dev process.

The version isn't there to help distinguish anything. 1.5_dev is just as vague as anything else, and it's wrong.
My suggestion to basically remove it (or say "master", which says nothing) is dictated by the fact that cmake has a git hook to stamp builds with git hash/version, which is visible in any cmake-built artifacts (such as from CI).

We do not need to duplicate that effort manually.

Znote:
Hmm ok, but we should definitely merge a change to `STATUS_SERVER_VERSION` before merging in the protocol 12 PR. Whatever it is. :P

Zbizu:
@Znote I can't find it, could you tag it?

Zbizu (approved):

Znote (approved):

Znote:
> @Znote I can't find it, could you tag it?

What I meant is that I don't think we should merge in protocol 12 into what is currently considered version 1.4. So we should merge this PR first, or change it to master and merge it before we introduce protocol 12. 

@DSpeichert 
This version number appears in the console when you start the server, and in statusprotocol. (for etc OT server lists)
When people run into problems and create support threads, they might grab this version number to tell what they are running. I find it to be useful to determine compatibility. "1.5" is vague, but better than nothing, and much better than 1.4 as that would be misleading at this point. Especially if protocol 12 is just around the corner.

DSpeichert (approved):

Issue #3742, opened 2021-10-17
yet another trade bug


Zbizu:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce
1. create bp (2365), add 19x norseman doll (8982) - this has to be only item on your character equipped
2. create bp (1988) and equip it with another character (this has to be only item in your equipment)
3. start a trade using bp of holding
4. trade
5. accept trade

### Expected behaviour
either cancelling the trade or swapping the backpacks

### Actual behaviour
one of backpacks is transfered,
one of backpacks gets deleted (server side)

### Environment
repository:
19 commits behind master (forked at #3176, probably unrelated to that bug), very likely to occur in most recent commit as well considering that the function wasn't touched since that commit.

platform:
Windows 7, visual studio 2019

### Notes
suggestion: make the "query" not searching the currently traded containers

PR #3750, opened 2021-10-19
fix disappearing backpack during trade

Zbizu:

### Changes Proposed
prevent a trade that would result in one of trade containers disappearing in situation when both players try to trade equipped backpack

expected result: the player who accepts the trade should see "there is not enough room" error

**Issues addressed:** #3742

### Notes
needs testing


Zbizu:
> That solves it, but shouldn't we find bug in internalMoveItem?

Adding item to container normally terminates the trade (or at least it should). In this situation trade was already completing so it couldn't terminate. The item moves by cloning and deleting itself, but the other item was added before the container deleted itself so the item was lost. I believe there won't be any more problems after very specific situation which is trade got resolved.

PR #3752, opened 2021-10-20
Fix memory leak on luaPlayerShowTextDialog - double ref

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Fix memory leak on luaPlayerShowTextDialog - double ref

- `Player::setWriteItem` and `Item::CreateItem` will add reference so we'll end up with 2 references and since we'll have 2 references the memory allocated will never be destroyed.

@SaiyansKing 


Zbizu:
wasn't there some way to check if item is virtual? Could save effort setting all those variables.

DSpeichert (approved):

PR #3781, opened 2021-11-06
Move onShareExperience callback from party event

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Move onShareExperience callback from party event.
- Add party folder with default file for onShareExperience event callback.

### DIFF
--- a/data/scripts/eventcallbacks/party/default_onShareExperience.lua
+++ b/data/scripts/eventcallbacks/party/default_onShareExperience.lua
@@ -1,0 +1,29 @@
+local ec = EventCallback
+
+ec.onShareExperience = function(self, exp)
+	local sharedExperienceMultiplier = 1.20 --20%
+	local vocationsIds = {}
+	local rawExp = exp
+
+	local vocationId = self:getLeader():getVocation():getBase():getId()
+	if vocationId ~= VOCATION_NONE then
+	table.insert(vocationsIds, vocationId)
+	end
+
+	for _, member in ipairs(self:getMembers()) do
+	vocationId = member:getVocation():getBase():getId()
+	if not table.contains(vocationsIds, vocationId) and vocationId ~= VOCATION_NONE then
+	table.insert(vocationsIds, vocationId)
+	end
+	end
+
+	local size = #vocationsIds
+	if size > 1 then
+	sharedExperienceMultiplier = 1.0 + ((size * (5 * (size - 1) + 10)) / 100)
+	end
+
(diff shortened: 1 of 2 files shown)


data/scripts/eventcallbacks/party/default_onShareExperience.lua
```diff
@@ -0,0 +1,29 @@
+local ec = EventCallback
+
+ec.onShareExperience = function(self, exp)
+	local sharedExperienceMultiplier = 1.20 --20%
+	local vocationsIds = {}
+	local rawExp = exp
+
+	local vocationId = self:getLeader():getVocation():getBase():getId()
+	if vocationId ~= VOCATION_NONE then
+	table.insert(vocationsIds, vocationId)
```
EPuncker, line 10:
```suggestion
		table.insert(vocationsIds, vocationId)
```

EPuncker: same with others

data/scripts/eventcallbacks/party/default_onShareExperience.lua
```diff
@@ -0,0 +1,29 @@
+local ec = EventCallback
+
+ec.onShareExperience = function(self, exp)
```
ramon-bernardo, line 3: or `function(self, exp, rawExp)`?

MillhioreBT: rawExp is a control variable that is used to know the real experience before any call in the eventcallback system, can be ignored if you don't need it ;)

MillhioreBT (approved):
LGTM, including all event events in eventcallback is essential to be able to have full control of these events from revscript.

DSpeichert:
Again - what's the point of reorganizing the datapack to move existing metatable scripts to "callbacks"?

If we had a plan to move everything and deprecate old system or standardize on one then it would make sense but it feels like those changes are suggested randomly and are unnecessary. This datapack (default one) showcases all features, not just the latest shiny callback.

Unless there's some change in behavior that I'm not seeing here, then please tell me.

MillhioreBT:
Don, You are absolutely right, there is no urgent need to move all systems to eventcallbacks, the only real function that can be taken advantage of by moving certain things to eventcallbacks is that when we create a new eventcallbacks script of the same type we can register the event before the default and control what will happen much earlier

However this is something the user can do personally on their own custom server and the tfs master doesn't need it at all

closing PR is not bad, merge it either

ramon-bernardo:
Is it time to merge it now? I can open another PR in a few days with this.

:triangular_flag_on_post: I'll map it for the future.

PR #3799, opened 2021-11-23
Change default value to NULL for conditions

nekiro:
Honestly, idk why it's NOT NULL as this is binary, but it only triggers me when I manually add entries thru phpmyadmin and it spits out errors, cause conditions cannot be null.

Default value for this should be null, this should not affect anything in loading and will fix that annoyance.

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -37,7 +37,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `posx` int NOT NULL DEFAULT '0',
   `posy` int NOT NULL DEFAULT '0',
   `posz` int NOT NULL DEFAULT '0',
-  `conditions` blob NOT NULL,
+  `conditions` blob DEFAULT NULL,
   `cap` int NOT NULL DEFAULT '400',
   `sex` int NOT NULL DEFAULT '0',
   `lastlogin` bigint unsigned NOT NULL DEFAULT '0',


schema.sql
```diff
@@ -37,7 +37,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `posx` int NOT NULL DEFAULT '0',
   `posy` int NOT NULL DEFAULT '0',
   `posz` int NOT NULL DEFAULT '0',
-  `conditions` blob NOT NULL,
+  `conditions` blob DEFAULT NULL,
```
marmichalski, line 40: This seems like a sane change, but considering there are already projects using that field, for example web AACs as string, it might be reasonable to do `DEFAULT ''`?

EPuncker (approved):

DSpeichert:
Hold up! Where the migration is?

nekiro:
> Hold up! Where the migration is?

It does not break anything, so there is none.

Zbizu:
would conflict with upcoming version 12 migrations

DSpeichert:
> > Hold up! Where the migration is?
> 
> It does not break/change anything, so there is none. I don't think we need migration for that xD

It does change schema, so now you introduce divergence... wouldn't hurt to have a migration for that.

PR #3801, opened 2021-11-24
Magicwall shouldn't disappear on forced step-in

omarcopires:
### Pull Request Prelude

### Changes Proposed 
This is a pull request that corrects the operation according to the revision made in #3039


Zbizu:
There's nothing wrong with your changes, but I think that a larger configuration should be made for pvp system so server owners can choose build their own pvp type easily just by changing values in config (the same way skull system is configurable now).

omarcopires:
> There's nothing wrong with your changes, but I think that a larger configuration should be made for pvp system so server owners can choose build their own pvp type easily just by changing values in config (the same way skull system is configurable now).

Yes, I agree with you, but it doesn't fit this change as it's a bug.

PR #3826, opened 2021-12-04
Fix condition update stats

ArturKnopik:
Condition doesn't update all stats, for example magic level

try: 
local condition = createConditionObject(CONDITION_ATTRIBUTES)
condition:setParameter(CONDITION_PARAM_STAT_MAGICPOINTSPERCENT, 200)
condition:setParameter(CONDITION_PARAM_TICKS, conditionTime)

### DIFF
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -500,6 +500,7 @@ void ConditionAttributes::updateStats(Player* player)
 
 	if (needUpdateStats) {
 		player->sendStats();
+		player->sendSkills();
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Zbizu (approved):

EPuncker (approved):

src/condition.cpp
```diff
@@ -578,6 +579,7 @@ void ConditionAttributes::endCondition(Creature* creature)
 
 		if (needUpdateStats) {
 			player->sendStats();
+			player->sendSkills();
```
marmichalski, line 582: You are sending skills twice in certain scenarios here. 🙄

ArturKnopik:
i will look on this and fix if need in free time

PR #3833, opened 2021-12-05
CreatureEvent onParsePacket

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
This is basically the same as Nekiro proposed in his PR, However, I made some small modifications that make the onParsePacket event have the same possibilities as the other events.

previously we could not create two events with the same recvbyte, now we can.
* This is because now the message is handled by a shared smart pointer, this guarantees that the object will not be deleted manually from lua causing (problems or crashing) in the following events of the same recvbyte.

The way we create network messages has been changed, now they will always be created wrapped in a shared pointer, nothing changes and everything remains the same, this was changed only to have compatibility with this new onParsePacket event

The method `creature:getEvents(type)` now has a new parameter `uint8_t` `creature:getEvents(type[, recvbyte = 0])`
* This new parameter will be mandatory if the event type is `CREATURE_EVENT_PARSE_PACKET` this will return the list of parse_packet events with the specified recvbyte
* why not return the whole list of parse_packets? In my opinion, it is only necessary to return events that are actually running, no one should register a parse_packet event with recvbyte 0

### Notes
Maybe this does not seem necessary today, since the handling of bytes is always faster and cleaner from the sources, however, this opens the possibility of experimentation, this means that anyone can investigate and test the message structures, and maybe create something beautiful, who knows? XD

**PR Base addressed:** #2978

**Credits:** @nekiro

# Examples

<details>
<summary>classic interface: parsepacket.lua</summary>

```lua
function onParsePacket(player, recvbyte, msg)
	-- structure for parseUseItem packet, recbyte: 130 (0x82)
	local pos = msg:getPosition()
	local clientId = msg:getU16()
	local stackpos = msg:getByte()
	local index = msg:getByte()
	print(("Player: %s wants to use item at position (%d, %d, %d) with sprite id %d"):format(player:getName(), pos.x, pos.y, pos.z, clientId))
	msg:delete()
end

```

</details>

<details>
<summary>revscript interface: parsepacket.lua</summary>

```lua
local creatureEvent = CreatureEvent("ParsePacket")
function creatureEvent.onParsePacket(player, recvbyte, msg)
	-- structure for parseUseItem packet, recbyte: 130 (0x82)
	local pos = msg:getPosition()
	local clientId = msg:getU16()
	local stackpos = msg:getByte()
	local index = msg:getByte()
	print(("Player: %s wants to use item at position (%d, %d, %d) with sprite id %d"):format(player:getName(), pos.x, pos.y, pos.z, clientId))
	msg:delete()
end

creatureEvent:recvbyte(130)
creatureEvent:resgister()
```

</details>

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -5492,6 +5492,25 @@ void Game::playerAnswerModalWindow(uint32_t playerId, uint32_t modalWindowId, ui
 	}
 }
 
+void Game::playerExecuteParsePacketEvent(uint32_t playerId, uint8_t recvbyte, const NetworkMessage& msg)
+{
+	Player* player = getPlayerByID(playerId);
+	if (!player) {
+		return;
+	}
+
+	CreatureEventList creatureEventList = player->getCreatureEvents(CREATURE_EVENT_PARSE_PACKET, recvbyte);
+	if (creatureEventList.empty()) {
+		return;
+	}
+
+	NetworkMessage_ptr message = std::make_shared<NetworkMessage>(msg);
+	for (CreatureEvent* creatureEvent : creatureEventList) {
+		creatureEvent->executeParsePacket(player, message);
+		message->setBufferPosition(1);
+	}
+}
+
 void Game::addPlayer(Player* player)
 {
 	const std::string& lowercase_name = asLowerCaseString(player->getName());
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -5506,7 +5506,7 @@ void Game::playerExecuteParsePacketEvent(uint32_t playerId, uint8_t recvbyte, co
 
 	NetworkMessage_ptr message = std::make_shared<NetworkMessage>(msg);
 	for (CreatureEvent* creatureEvent : creatureEventList) {
-		creatureEvent->executeParsePacket(player, message);
+		creatureEvent->executeParsePacket(player, recvbyte, message);
 		message->setBufferPosition(1);
 	}
 }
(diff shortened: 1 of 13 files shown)


MillhioreBT:
I will make the interface in Events, since it seems more correct that the bytes are received by players...
I will do it in another PR, then I will see your opinions

omarcopires:
This is something of the utmost importance for the upcoming 12.x updates, thank you so much for bringing it to us.
Just a question, could it be done in revscript since it's something new?

MillhioreBT:
> This is something of the utmost importance for the upcoming 12.x updates, thank you so much for bringing it to us.

Thanks, and if I agree with you, it will be something necessary at the apogeeof this version.

I will open another PR with onParsePacket in Events, to see which one is the most voted

nekiro:
No idea what is the purpose of this PR, this is copied #2978 with changes in luascript to support shared_ptrs on NetworkMessages.
I updated my PR from while ago and it should be ready to merge.

src/game.cpp
```diff
@@ -5492,6 +5492,25 @@ void Game::playerAnswerModalWindow(uint32_t playerId, uint32_t modalWindowId, ui
...
+
+	CreatureEventList creatureEventList = player->getCreatureEvents(CREATURE_EVENT_PARSE_PACKET, recvbyte);
+	if (creatureEventList.empty()) {
+		return;
+	}
+
+	NetworkMessage_ptr message = std::make_shared<NetworkMessage>(msg);
+	for (CreatureEvent* creatureEvent : creatureEventList) {
+		creatureEvent->executeParsePacket(player, recvbyte, message);
+		message->setBufferPosition(1);
```
nekiro, line 5510: this doesn't look good, when lua calls message:delete()

PR #3837, opened 2021-12-07
libs: warn user about duplicate storage value

omarcopires:
### Pull Request Prelude

### Changes Proposed 
The script checks for duplicate storages when starting the server and also added the AccountStorageKeys


lgrossi:
You have 16 storages, why do you need that at all?
The place you copied it has 10k+ storages, so it's impossible to do a visual check, but here is just random code without a goal.

It's a nice approach anyway, just make sure you put the credits.
https://github.com/opentibiabr/otservbr-global/commit/fc987f273d3bad2b09e94fabe058bebade449678

Also the complexity analisys and why is it ok to have it, even if it's not lightweight, is a good think to keep imho.

Also Wolf and I found a faster way of doing that, after we implemented this one. 
If you have any doubts about it, let me know.

dudantas:
> You have 16 storages, why do you need that at all? The place you copied it has 10k+ storages, so it's impossible to do a visual check, but here is just random code without a goal.
> 
> It's a nice approach anyway, just make sure you put the credits. [opentibiabr/otservbr-global@fc987f2](https://github.com/opentibiabr/otservbr-global/commit/fc987f273d3bad2b09e94fabe058bebade449678)
> 
> Also the complexity analisys and why is it ok to have it, even if it's not lightweight, is a good think to keep imho.
> 
> Also Wolf and I found a faster way of doing that, after we implemented this one. If you wanna chat about it, to gather some ideas, let me know.

I particularly like the idea of contributions between repositories a lot, after all, everything is used by the community in the end. I just think it's funny that the owner of the pull request only lives on the otland forum saying that "everything on otbr is buggy", but in all his pull requests that I saw here, they were otbr things. Note that all pull requests from otbr are "cherry pick" from tfs, we require credit, link to original pr and author. I think this is an important step towards a more united community!

Very unethical, in my opinion.

amatria:
You can do it in `O(n)` time using a hash table. Sorting may feel "cheap", but it is never as "cheap" as implementing the right algorithm.

Zbizu:
I was thinking about that yesterday, what a coincidence!

Personally I'd do it this way:
`PlayerStorageKeys.register(key, alias)`
It would search the table with keys first, then:

- if available: add to it, then set is as `PlayerStorageKeys.alias = key`
- if already taken: send error

amatria:
The straightforward solution:
```lua
local storage_keys = {
  foobar = 1,
  foo = 2,
  bar = 3, 
  baz = 4, 
  qux = 1
}
local storage_map = {}

for _, key in pairs(storage_keys) do
  if storage_map[key] == nil then
      storage_map[key] = 1
  else
      print(key .. " is duplicated")
  end
end
```

Also, why is `extractValues` recursive? If you know in advance your input is a table of tables, process it as it is. Recursion implies pushing and poping data to and from the stack. It hurts performance a lot. If you are maximizing throughput, avoid unnecessary complexity at all times.

omarcopires:
> The straightforward `O(n)` solution:
> 
> ```lua
> local storage_keys = {
>   foobar = 1,
>   foo = 2,
>   bar = 3, 
>   baz = 4, 
>   qux = 1
> }
> local storage_map = {}
> 
> for _, key in pairs(storage_keys) do
>   if storage_map[key] == nil then
>     storage_map[key] = 1
>   else
>     print(key .. " is duplicated")
>   end
> end
> ```
> 
> Also, why is `extractValues` recursive? If you know in advance your input is a table of tables, process it as it is. Recursion implies pushing and poping data to and from the stack. It hurts performance a lot. If you are maximizing throughput, avoid unnecessary complexity at all times.

I applied the change, but I'm not sure how it works.

amatria:
You have not tested the changes, have you? You are iterating over a table of tables:
```lua
local keys = {AccountStorageKeys, GlobalStorageKeys, PlayerStorageKeys}
```
Your code does not work, nor will it ever flag a storage key as duplicated.

What I would do:
1. Understand the code I posted above. Read as much information as you need, but try and understand it.
2. Play with a Lua REPL to check you have fully understood how it functions.
3. Now that you understand the code, encapsulate the algorithm in a new auxiliary function.
4. And now, call the function several times to check for duplicates in the different tables. First, find the duplicates in the `AccountStorageKeys` table. Next, find the duplicates in the `GlobalStorageKeys` table. And finally, find the duplicates in the `PlayerStorageKeys` table.

Do not commit changes if you do not fully understand your code.

PR #3834, opened 2021-12-05
EventCallback Lib a more modern and easy way to configure

MillhioreBT:
### Pull Request Prelude
- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
I changed the metatable a bit, now we can access it in a more beautiful way.
These changes allow a special configuration for network events, more information about this special in PR #3838

**Issues addressed:** Nothing!

PR #3838, opened 2021-12-07
onParsePacket in Events/Callback

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
This event allows filtering all packets captured by a player.

If #3834 is merged then we can use EventCallback to have concrete callbacks for each type of recvbyte, here is an example:
```lua
local ec = EventCallback

function ec.onParsePacket(player, recvbyte, msg)
    print("TEST 1", player, recvbyte, msg:getByte())
end

ec:recvbyte(0x90)
ec:register(1)

function ec.onParsePacket(player, recvbyte, msg)
    print("TEST 2", player, recvbyte, msg:getByte())
end

ec:recvbyte(0x90)
ec:register(2)
```
If no recvbyte is declared, then the event will be called regardless of the recvbyte, example:
```lua
local ec = EventCallback

function ec.onParsePacket(player, recvbyte, msg)
    print("TEST 1", player, recvbyte, msg:getByte())
end

ec:register(1)

function ec.onParsePacket(player, recvbyte, msg)
    print("TEST 2", player, recvbyte, msg:getByte())
end

ec:register(2)
```

If someone wants to intentionally delete the network message, he can do it, however the interpreter will not collapse, since the object will have been correctly deleted and no attempt will be made to access a dead object, this has always been the case.

The method `delete` exists only for network messages that we create ourselves and we want to delete at a certain point to save the garbage collector work, because it is not really mandatory to have to clean the object manually.

Regardless of the PR preferred by the majority and whatever the one that merges, I want to give all the credits to @nekiro 

### DIFF
--- a/data/events/events.xml
+++ b/data/events/events.xml
@@ -31,6 +31,7 @@
 	<event class="Player" method="onLoseExperience" enabled="0" />
 	<event class="Player" method="onGainSkillTries" enabled="1" />
 	<event class="Player" method="onWrapItem" enabled="1" />
+	<event class="Player" method="onParsePacket" enabled="0" />
 
 	<!-- Monster methods -->
 	<event class="Monster" method="onDropLoot" enabled="1" />
--- a/data/events/scripts/player.lua
+++ b/data/events/scripts/player.lua
@@ -211,3 +211,6 @@ function Player:onWrapItem(item)
 		end
 	end
 end
+
+function Player:onParsePacket(recvbyte, msg)
+end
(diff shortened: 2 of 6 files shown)


nekiro:
So every single message is pushed to lua regardless of it being used or not?
You also read the same message from two sources if you decide to read something in lua and c++

Zbizu:
Why not pushing it at the end of switch statement (in default option), for unhandled headers only?

We don't need to parse something that is already coded or even redirected to other event (like player use item or player move)

PR #3844, opened 2021-12-13
Fix the client crash when you open the missions window

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
The client 12.81 now needs the index of the mission that is sent to him, I do not know if it is the best way or if we need to identify the missions in another way that is not the index of the list

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -2366,8 +2366,10 @@ void ProtocolGame::sendQuestLine(const Quest* quest)
 	msg.add<uint16_t>(quest->getID());
 	msg.addByte(quest->getMissionsCount(player));
 
+	uint16_t missionId = 0;
 	for (const Mission& mission : quest->getMissions()) {
 		if (mission.isStarted(player)) {
+			msg.add<uint16_t>(++missionId);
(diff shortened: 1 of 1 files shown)


Zbizu:
You may track quests progress now. This is why it got an extra value. Overall quest log should have been rewritten, but I remember that the maintainers desire to keep compatibility blocked the pr in the past.

omarcopires:
Maybe it's better to look this way rather than a partial solution, of course it's just a suggestion, but it would be interesting if you bring us your version of the code.

If possible take a look at this:
https://github.com/igorzerafusion/optimized_forgottenserver/commit/7b23b3503b123dc41ca86b3d48ae6557851ba988
https://github.com/igorzerafusion/optimized_forgottenserver/commit/63503e15a5513e9f83adf9477a343e86649b396f
https://github.com/igorzerafusion/optimized_forgottenserver/commit/c8f4d429abec31a23c45f3a3c31b49379ac6c900

EPuncker (approved):
Indeed there is a client crash (you can test it by having a non empty quest log), and this code does fix it.

ranisalt:
And this has broken the other way around to OTClient :laughing:

PR #3851, opened 2021-12-15
Fix skill status when item decays

Diego-OT:
Steps to reproduce the bug:

Equip an item that give you some bonus skill and decay to another item or 0.
When it decay, you will continue with that bonus.

part of fix:
https://github.com/opentibiabr/canary/blob/77e037dd72063279754e6efa8ae97a5843981111/src/game/game.cpp#L6474

full fix:
checks if the item is being carried by the player and also check if the item is equipped to avoid skill changes when the item decay inside backpack (not equipped).

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4718,11 +4718,57 @@ void Game::startDecay(Item* item)
 	}
 }
 
+static bool isEquipped(Player* player, Item* item)
+{
+	for (int slot = CONST_SLOT_FIRST; slot <= CONST_SLOT_LAST; ++slot) {
+		if (player->getInventoryItem(static_cast<slots_t>(slot)) == item) {
+			return true;
+		}
+	}
+	return false;
+}
+
 void Game::internalDecayItem(Item* item)
 {
 	const int32_t decayTo = item->getDecayTo();
 	if (decayTo > 0) {
-		startDecay(transformItem(item, decayTo));
+		if (Player* player = item->getHoldingPlayer(); player && isEquipped(player, item)) {
+			const ItemType& it = Item::items[item->getID()];
+
+			bool needUpdateSkills = false;
+			for (int i = SKILL_FIRST; i <= SKILL_LAST; ++i) {
+				if (it.abilities && it.abilities->skills[i] != 0) {
+					needUpdateSkills = true;
+					player->setVarSkill(static_cast<skills_t>(i), -it.abilities->skills[i]);
+				}
+			}
+
+			bool needUpdateStats = false;
+			for (int s = STAT_FIRST; s <= STAT_LAST; ++s) {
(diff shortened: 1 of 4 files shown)


src/player.cpp
```diff
@@ -4705,3 +4705,21 @@ void Player::updateRegeneration()
...
+{
+	bool isEquiped = false;
+	if (!item) {
+		return isEquiped;
+	}
+
+	static const slots_t allSlots[] = {CONST_SLOT_RIGHT, CONST_SLOT_LEFT, CONST_SLOT_HEAD, CONST_SLOT_NECKLACE, CONST_SLOT_ARMOR, CONST_SLOT_LEGS, CONST_SLOT_FEET, CONST_SLOT_RING};
+	for (slots_t slot : allSlots) {
+		Item* checkitem = inventory[slot];
+		if (checkitem and (checkitem == item)) {
```
Znote, line 4719:
```suggestion
		if (checkitem && (checkitem == item)) {
```

src/game.cpp
```diff
@@ -4534,6 +4534,42 @@ void Game::internalDecayItem(Item* item)
...
+		if (player && player->isEquippedItem(item)) {
+			bool needUpdateSkills = false;
+			for (int32_t i = SKILL_FIRST; i <= SKILL_LAST; ++i) {
+				if (it.abilities && it.abilities->skills[i] != 0) {
+					needUpdateSkills = true;
+					player->setVarSkill(static_cast<skills_t>(i), -it.abilities->skills[i]);
+				}
+			}
+
+			if (needUpdateSkills) {
```
Codinablack, line 4547: This update here is an un-needed duplicate. You already update at the end where it should be.

src/player.cpp
```diff
@@ -4705,3 +4705,21 @@ void Player::updateRegeneration()
 		condition->setParam(CONDITION_PARAM_MANATICKS, vocation->getManaGainTicks() * 1000);
 	}
 }
+
+bool Player::isEquippedItem(const Item* item) const
```
Codinablack, line 4709: For this function, TFS already has a way of checking if an item is one of those slots, Player::getInventoryItem. Utilize that method for this one.

Codinablack (changes requested):
Looks good overall. The changes I requested are nothing serious, just reducing code duplication. Fix those and I will approve.

ranisalt:
@Codinablack fixed your suggestions and rebased to fix conflicts

Codinablack (approved):
Looks good to me!

Codinablack:
> Looks good to me!

I stand corrected. This fix looks like it would serve as a good fix, if skill status didn't already change during decay by calling postRemoveNotification, because of that, this PR isn't necessary.

nekiro:
Stats/skills are properly updated for decaying items (they get called from postRemoveNotification)

PR #3857, opened 2021-12-17
Feat Boosted creature

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Added the functionality to boosted creature, the code works correctly, but it should be improved.
Currently the system changes the monster at every server startup, but this must be changed to every 24 hours afterwards, that way it's easier to do tests.

Reference material https://tibia.fandom.com/wiki/Boosted_Creature

### DIFF
--- a/data/creaturescripts/scripts/login.lua
+++ b/data/creaturescripts/scripts/login.lua
@@ -14,8 +14,8 @@ function onLogin(player)
 	player:sendTextMessage(MESSAGE_STATUS_DEFAULT, loginStr)
 
 	-- Bosted Creature
-	if Game.getBoostMonster():lower() ~= 'none' and MonsterType(Game.getBoostMonster()) then
-		player:sendTextMessage(MESSAGE_BOOSTED_CREATURE, string.format("Today's boosted creature: %s\nBoosted creatures yield more experience points, carry more loot than usual and respawn at a faster rate.", MonsterType(Game.getBoostMonster()):getName()))
+	if Game.getBoostedMonster():lower() ~= 'none' and MonsterType(Game.getBoostedMonster()) then
+		player:sendTextMessage(MESSAGE_BOOSTED_CREATURE, string.format("Today's boosted creature: %s\nBoosted creatures yield more experience points, carry more loot than usual and respawn at a faster rate.", MonsterType(Game.getBoostedMonster()):getName()))
 	end
 
 	-- Promotion
--- a/data/creaturescripts/scripts/login.lua
+++ b/data/creaturescripts/scripts/login.lua
@@ -13,9 +13,9 @@ function onLogin(player)
 	end
 	player:sendTextMessage(MESSAGE_STATUS_DEFAULT, loginStr)
 
-	-- Bosted Creatures
+	-- Bosted Creature
 	if Game.getBoostMonster():lower() ~= 'none' and MonsterType(Game.getBoostMonster()) then
-		player:sendTextMessage(MESSAGE_EVENT_ADVANCE, string.format("Today's boosted creature: %s\nBoosted creatures yield more experience points, carry more loot than usual and respawn at a faster rate.", MonsterType(Game.getBoostMonster()):getName()))
+		player:sendTextMessage(MESSAGE_BOOSTED_CREATURE, string.format("Today's boosted creature: %s\nBoosted creatures yield more experience points, carry more loot than usual and respawn at a faster rate.", MonsterType(Game.getBoostMonster()):getName()))
 	end
 
 	-- Promotion
--- a/data/creaturescripts/scripts/login.lua
+++ b/data/creaturescripts/scripts/login.lua
@@ -13,6 +13,11 @@ function onLogin(player)
 	end
 	player:sendTextMessage(MESSAGE_STATUS_DEFAULT, loginStr)
 
+	-- Bosted Creatures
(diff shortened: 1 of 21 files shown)


data/creaturescripts/scripts/login.lua
```diff
@@ -13,6 +13,11 @@ function onLogin(player)
 	end
 	player:sendTextMessage(MESSAGE_STATUS_DEFAULT, loginStr)
 
+	-- Bosted Creatures
+	if Game.getBoostMonster():lower() ~= 'none' and MonsterType(Game.getBoostMonster()) then
+		player:sendTextMessage(MESSAGE_EVENT_ADVANCE, string.format("Today's boosted creature: %s\nBoosted creatures yield more experience points, carry more loot than usual and respawn at a faster rate.", MonsterType(Game.getBoostMonster()):getName()))
```
EPuncker, line 18:
```suggestion
		player:sendTextMessage(MESSAGE_BOOSTED_CREATURE, string.format("Today's boosted creature: %s\nBoosted creatures yield more experience points, carry more loot than usual and respawn at a faster rate.", MonsterType(Game.getBoostMonster()):getName()))
```

data/creaturescripts/scripts/login.lua
```diff
@@ -13,6 +13,11 @@ function onLogin(player)
 	end
 	player:sendTextMessage(MESSAGE_STATUS_DEFAULT, loginStr)
 
+	-- Bosted Creatures
```
EPuncker, line 16:
```suggestion
	-- Boosted Creature
```

src/game.cpp
```diff
@@ -5751,3 +5752,78 @@ bool Game::reload(ReloadTypes_t reloadType)
...
+
+	return true;
+}
+
+void Game::setBoostMonster(std::string monstername, uint16_t monsterid)
+{
+	Database& db = Database::getInstance();
+	std::ostringstream query;
+
+	query << "UPDATE `server_config` SET `value` = '" << monsterid << "' WHERE `config` = 'boost_monster'";
```
EPuncker, line 5788: use fmt?

src/luascript.cpp
```diff
@@ -2135,6 +2135,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("Game", "getPlayerCount", LuaScriptInterface::luaGameGetPlayerCount);
 	registerMethod("Game", "getNpcCount", LuaScriptInterface::luaGameGetNpcCount);
 	registerMethod("Game", "getMonsterTypes", LuaScriptInterface::luaGameGetMonsterTypes);
+	registerMethod("Game", "getMonstersRace", LuaScriptInterface::luaGameGetMonstersRace);
```
EPuncker, line 2138: getMonstersByRace maybe?

data/scripts/eventcallbacks/monster/default_onDropLoot.lua
```diff
@@ -5,6 +5,11 @@ ec.onDropLoot = function(self, corpse)
 		return
 	end
 
+	local percent = 1.0
```
EPuncker, line 8: I didn't see where this is being used yet

data/scripts/eventcallbacks/player/default_onLookInBattleList.lua
```diff
@@ -3,10 +3,15 @@ local ec = EventCallback
 ec.onLookInBattleList = function(self, creature, distance)
 	local description = "You see " .. creature:getDescription(distance)
 	if self:getGroup():getAccess() then
+		if thing:isMonster() then
```
EPuncker, line 6:
```suggestion
		if creature:isMonster() then
```

EPuncker (changes requested):
A few things:

- what is this BOOSTED_MONSTERS_IMAGES_URL?
- change all **boost** to **boosted** where applicable
- where is the faster respawn rate?
- is the system that doesn't allow same creature to be selected more than once in a month added?
- loot bonus seems to not be added
- config.lua entries where not added

now a few optional suggestions:
- boosted monster black list
- talkaction to set specific monster by name or random if not provided name

data/scripts/eventcallbacks/monster/default_onDropLoot.lua
omarcopires: I ended up skipping this step, but I believe everything is fine now.

src/game.cpp
omarcopires: Sorry but I don't know how to do this!

Zbizu:
What is the url here for? I don't see any protocol part that uses it.

If it uses external url only to get a picture, can't we just use outfit system from server aac? I'm pretty sure we can cooperate with aac devs in that matter.

omarcopires:
- what is this BOOSTED_MONSTERS_IMAGES_URL?
They are used to send the lookType and lookTypeEx to the database, so the website can read the monster and the corresponding outfit.

- is the system that doesn't allow same creature to be selected more than once in a month added?
Yes, monsters are not repeated, a new one is always set every day, as long as there are more than two monsters registered with raceId. However there is a problem, if only one monster has raceId the server freezes on startup.

- loot bonus seems to not be added
it was applied

- config.lua entries where not added
it was applied

- change all boost to boosted where applicable
it was applied

- where is the faster respawn rate
not added yet

Zbizu:
personally, I'd just push boosted creature name + outfit to the database and let a system like gesiorAAC handle making the picture on PHP level

PR #3860, opened 2021-12-17
feature: kill tracking

ArturKnopik:
image

based on https://github.com/otland/forgottenserver/pull/3757
Co-Authored-By: @SaiyansKing and @omarcopires 
- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

PR #3861, opened 2021-12-18
bugfix and optimizations: kill tracking

Znote:
Followup on #3860

Bugfix:
* returns `false` if it fails to set `monsterType` (instead of returning nil when it finds it)
* `Player.updateKillTracker` now returns status of operation as a bool, instead of void
* `monster:getType()` instead of `monster.getType()`

Optimizations:
* 1 function instead of 2
* Invokes `getType()` 1 time instead of 2 times. 
* Invokes `getOutfit()` 1 time instead of 6 times. 
* instantiate monsterType as a local
* no need to confirm that networkMessage is set in conditional statements, since it is created locally
* no need for `elseif`, or a conditional statement to determine need to send message to self, since it has already returned something otherwise.

### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 

### DIFF
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -313,43 +313,44 @@ function Player.getWeaponType(self)
 	return WEAPON_NONE
 end
 
-local function prepareUpdateKillTrackerMsg(monster, corpse)
-	monsterType = monster:getType()
-	if monsterType then
-		return nil
-	end
-	local networkMessage = NetworkMessage()
-	networkMessage:addByte(0xD1)
-	networkMessage:addString(monster:getName())
-	networkMessage:addU16(monsterType:getOutfit().lookType or 19)
-	networkMessage:addByte(monsterType:getOutfit().lookHead)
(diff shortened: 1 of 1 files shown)


data/lib/core/player.lua
```diff
@@ -313,43 +313,44 @@ function Player.getWeaponType(self)
...
-		msg:delete()
-	end
+    local monsterType = monster:getType()
+    if not monsterType then
+        return false
+    end
+    
+    local monsterOutfit = monsterType:getOutfit()
+
+    local networkMessage = NetworkMessage()
```
omarcopires, line 324: we could just use `msg`, as is done in sources

EPuncker (approved):

DSpeichert (approved):

data/lib/core/player.lua
Znote: Yeah, but msg is ambiguous and could just mean a regular string/text. 
Since this is a network message, why not call it so?

PR #3863, opened 2021-12-19
Moved market description to Lua

EPuncker:
### Pull Request Prelude

### Changes Proposed
Moved market item description to Lua.
- added "chance to hit" and "range" that have been missing for years
- added ItemType:getWeaponString()
- added itemType:getClassification()


EPuncker:
yeah it is actually tested and working:
2021-12-18 23_59_59-Tibia - Perflex

Zbizu (approved):
Let's go. Tier system is waiting.

data/scripts/eventcallbacks/player/default_onLookInMarket.lua
```diff
@@ -0,0 +1,319 @@
...
+			response:addU64(stats.lowestPrice)
+		else
+			response:addByte(0x00)
+		end
+	end
+
+	response:sendToPlayer(self)
+end
+
+ec:register()
```
DSpeichert, line 319: This file looks double-indented compared to other Lua files, no?

Zbizu: `do...end` clauses were used to limit the scope of local variables. This has been done in libs already way before this pull request.

nekiro: why though?

MillhioreBT: In this file it is not necessary to use `do ... end` clauses
There is not a single reason for it

MillhioreBT (changes requested):

DSpeichert:
@EPuncker needs rebase

EPuncker:
> @EPuncker needs rebase

fixed.

DSpeichert (dismissed):
2x as much code in Lua as C++, huh?

data/scripts/eventcallbacks/player/default_onLookInMarket.lua
Znote: I agree the `do .. end` local scope limitation seems pointless. Is the overhead so big that we shouldn't deconstruct them naturally at the end of function, and force it lots of times throughout the execution? Is this some shenanigans to keep memory overhead as low as possible? I'm not to sure how Lua works, but to me it sounds like your making the function slower by calling the garbage collector way too often. 
And indentation size seems to be set to 8, it should be 4.

Zbizu:
> 2x as much code in Lua as C++, huh?

A few features were missing or had inaccurate descriptions in the C++ version. For example distance weapons show their modifiers now.

src/luascript.cpp
```diff
@@ -42,6 +42,7 @@
 #include "globalevent.h"
 #include "script.h"
 #include "weapons.h"
+#include "iomarket.h"
```
DSpeichert, line 45: Aren't these alphabetically ordered?

DSpeichert (changes requested):

DSpeichert (approved):

PR #3865, opened 2021-12-19
fix #3842: item attributes + trade bug

### LINKED ISSUE #3842: item attributes + trade bug

### Summary
Changing item attribute mid-trade doesn't cancel the trade. This can result in such situation:
- assume the server has some kind of upgrade system
- players trade the upgraded item
- player1 (the buyer) looks at the item and sees stats
- player2 (the seller) uses some kind of gem to remove the attributes from the traded item
- trade doesn't get cancelled
- both parties accept the trade
- player1 buys item with no attributes thinking that he bought upgraded one

1. start trade as player 1
2. join trade as player 2
3. change traded item attribute
4. look at the item in trade window

### Expected behaviour
trade stops before step 4 happens

### Actual behaviour
trade continues

ArturKnopik:
fix #3842

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -478,10 +478,6 @@ bool Actions::useItemEx(Player* player, const Position& fromPos, const Position&
 	}
 
 	if (action->executeUse(player, item, fromPos, action->getTarget(player, creature, toPos, toStackPos), toPos, isHotkey)) {
-		Thing* tradeItem = action->getTarget(player, creature, toPos, toStackPos);
-		if (tradeItem  && tradeItem->getItem() && tradeItem->getItem() == player->getTradeItem()) {
-			g_game.playerCloseTrade(player->getID());
-		}
 		return true;
 	}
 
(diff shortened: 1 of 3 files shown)


Zbizu:
personally I'd make `setAttribute` cancel the trade through `item->getHoldingPlayer` and (if player and trade) cancelling the trade

ArturKnopik:
nice tip, i will try fix this that way, should we keep this fix to?

Zbizu:
> nice tip, i will try fix this that way, should we keep this fix to?

I can't think of a scenario in which this would be necessary.
- Moving the item cancels the trade already.
- Removing the item cancels it too.
- Changing the item count/subType cancels the trade as well.

With setAttribute cancelling the trade, I can't think of any situation in which your fix would be triggered.

ArturKnopik:
make trade using item A, use item B(example: upgrade stone) on item A, trade will be canceled and item A upgraded

MillhioreBT:
I think it is not necessary to add this, in fact adding an attribute in the middle of a transaction is very normal.

Zbizu:
That's what onTradeCompleted is for

PR #3874, opened 2021-12-26
FIX: Container will not decay in house after server load

hansjny:
### Pull Request Prelude

- [ x] I have followed [proper The Forgotten Server code styling][code].
- [ x] I have read and understood the [contribution guidelines][cont] before making this PR.

### Changes Proposed

This change makes sure containers decay (usually corpses). Previously **_only_** the contents of a container would be marked to decay by the `container::startDecaying()` function call. This issue is obvious when decyable containers are loaded from the database or other places. 

The issue of unrotting corpses is reproducable when placing a corpse in a house before killing the server. When the server goes back online the corpse will never decay.  This is because the house-loading code in  `IOMapSerialize::loadItem` calls `item->startDecaying();` on all items, and currently this fails to add the corpse itself to the state of decaying items.  This is probably an oversight, as `item::startDecaying();` already calls `g_game.startDecay(this);`, but `container::startDecaying()` does not. 

**Issues addressed:** 

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -724,6 +724,7 @@ void Container::internalAddThing(uint32_t, Thing* thing)
 
 void Container::startDecaying()
 {
+	g_game.startDecay(this);
 	for (Item* item : itemlist) {
 		item->startDecaying();
 	}
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -725,6 +725,7 @@ void Container::internalAddThing(uint32_t, Thing* thing)
 void Container::startDecaying()
 {
 	Item::startDecaying();
+	
 	for (Item* item : itemlist) {
 		item->startDecaying();
 	}
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -724,7 +724,7 @@ void Container::internalAddThing(uint32_t, Thing* thing)
 
 void Container::startDecaying()
 {
-	g_game.startDecay(this);
+	Item::startDecaying();
 	for (Item* item : itemlist) {
 		item->startDecaying();
 	}


nekiro:
Isn't this correct behaviour though?
I remember seeing people having corpses inside houses and they never rotted.

hansjny:
> Isn't this correct behaviour though? I remember seeing people having corpses inside houses and they never rotted.

Corpses do rot in houses, because corpses are added to the decay list when a creature is killed. The problem is when the server loads the items back into the game, then all items are added to the decay list except corpses. Remember this only happens if an unrotten corpse is inside a house and you kill the server before it has a chance to rot. 

I also remember seeing corpses inside houses, but I believe those are the ones placed by the map editor and then a player has moved them into a house. These corpses have no decay attribute

Zbizu:
> Isn't this correct behaviour though? I remember seeing people having corpses inside houses and they never rotted.

Those are naturally spawning corpses that do not decay (eg dead humans named "skeleton" from some dragon caves)

MillhioreBT:
This mechanic of the corpses in the houses I like uwu, I guess it's not correct and I'm going to miss it XD

Erza:
> This mechanic of the corpses in the houses I like uwu, I guess it's not correct and I'm going to miss it XD

Why do you always have to add your useless comments? Would be great if you could shut up if you have nothing of value to add, instead of giving people notifications.

On topic: The corpses that don't decay in houses are specific item id's that never decay. The idea of the PR is correct. You can still have corpses in houses, they just need to be the proper id and have correct decay attributes.

For example in the 4 corners here https://tibiamaps.io/map#32598,31890,12:3 there are 4 dead dragon corpses that never decay. The client id's are 4286 and 4288.

EPuncker (approved):

omarcopires:
In rl tibia it is possible to keep monster corpses at home, without them decomposing.

hansjny:
> In rl tibia it is possible to keep monster corpses at home, without them decomposing. Exemple: https://www.youtube.com/watch?v=gLIRSLNf88E

Please read the above conversations. **_This is not about being able to keep corpses in houses._** The creatures in the house there is not killed by players, but placed by the map editor. These creatures do not have a decay attribute. 

The title might be misleading. **The problem is not about corpses **.  No container, whatever the type, will decay inside a house.

src/container.cpp
```diff
@@ -724,6 +724,7 @@ void Container::internalAddThing(uint32_t, Thing* thing)
 
 void Container::startDecaying()
 {
+	g_game.startDecay(this);
```
nekiro, line 727: Item::startDecaying()?
This is override method, so making use of the base class method would make more sense imo.

hansjny: Yes. Much cleaner. Fixed.

hansjny:
Here is a more visual example of the problem:
https://streamable.com/27edkz

Zbizu (dismissed):

src/container.cpp
```diff
@@ -724,6 +724,7 @@ void Container::internalAddThing(uint32_t, Thing* thing)
 
 void Container::startDecaying()
 {
+	Item::startDecaying();
```
nekiro, line 727: sorry for nitpicking but I would love a new line under this line

nekiro (approved):

MillhioreBT (approved):
It works fine, although it still does not work for Depots.

EPuncker (approved):

DSpeichert (approved):

EPuncker:
@DSpeichert shouldn't this PR be aimed to tfs master instead of 1.4 branch? I noticed it just now

DSpeichert (changes requested):
It should be targeting master.

hansjny:
> It should be targeting master.

ok, wont fix, i have patched this on my local version so ill close the pr.

Erza:
> > It should be targeting master.
> 
> ok, wont fix, i have patched this on my local version so ill close the pr.

So? Just had to fix the target branch and it could have been merged, smh

hansjny:
> 

Then tell me that. I retargeted and got loads of other commits, probably because the branches are not in sync.

Issue #3884, opened 2022-01-05
revscript outfit spell


emil92b:
### Explanation of what you want to do that is currently impossible

be able to read "monster" and "item", currently target is set to "invisible" instead

https://github.com/otland/forgottenserver/blob/master/data/scripts/lib/register_monster_type.lua#L280-L335
https://github.com/otland/forgottenserver/blob/master/src/monsters.cpp#L707-L717
https://github.com/otland/forgottenserver/blob/master/src/monsters.cpp#L390-L413 - xml reference

{name = "outfit", interval = 2000, chance = 17, **monster = "rabbit"**, duration = 20000, range = 7, target = true, effect = CONST_ME_MAGIC_BLUE}

{name = "outfit", interval = 2000, chance = 1, **item = 3976**, duration = 4000, range = 7, target = true, effect = CONST_ME_MAGIC_BLUE},

theshibbies:
> ### Explanation of what you want to do that is currently impossible
> be able to read "monster" and "item", currently target is set to "invisible" instead
> 
> https://github.com/otland/forgottenserver/blob/master/data/scripts/lib/register_monster_type.lua#L280-L335 https://github.com/otland/forgottenserver/blob/master/src/monsters.cpp#L707-L717 https://github.com/otland/forgottenserver/blob/master/src/monsters.cpp#L390-L413 - xml reference
> 
> {name = "outfit", interval = 2000, chance = 17, **monster = "rabbit"**, duration = 20000, range = 7, target = true, effect = CONST_ME_MAGIC_BLUE}
> 
> {name = "outfit", interval = 2000, chance = 1, **item = 3976**, duration = 4000, range = 7, target = true, effect = CONST_ME_MAGIC_BLUE},

We can add a luaMonsterSpellSetConditionOutfit function and add it into register_monster_type.lua.

However, an issue lies with trying to get a monster types outfit when the file hasn't been loaded yet. For instance, blue djinn has trouble loading rabbit looktypes because the blue djinn file is loaded far before the rabbit. 

Reading the lua files lines via io.open is possible but seems a bit hacky to me, interested to hear others solutions.

emil92b:
that didn't fix it for me, they are still invisible

emil92b:
@theshibbies thanks, that worked for me! you should do a PR mby?

Zbizu:
why don't you just use looktypes?

PR #3903, opened 2022-01-20
Revscript MonsterType Spell Fixes

theshibbies:
### Pull Request Prelude

**Protocol version**
Any

### Changes Proposed
Spells aren't handled correctly in register_monster_type.lua, Not exactly sure if this is the fully correct way to handle these, but at least I am bringing the issue to everyones attention.

**Issues addressed:** 
Within revscript monstertype lua files, spells named "condition" wouldn't be registered and would cause errors. Spells named "drunk" would create condition twice, once at line 281 and again at line 331, which would cause all sorts of crashes. You can reproduce these errors by adding a monster lua file and adding these types of spells below.

`{name ="condition", interval = 2000, chance = 30, type = CONDITION_POISON, startDamage = 0, tick = 4000, minDamage = -440, maxDamage = -520, radius = 5, target = false, effect = CONST_ME_POISONAREA}`

`{name ="drunk", interval = 2000, chance = 10, drunkenness = 25, duration = 5000, range = 7, shootEffect = CONST_ANI_ENERGY}`

The easiest way is to convert the current monster xml files into revscript using this tool:
https://otland.net/threads/ot-monster-converter.277546/

**Note:** You may need to create multiple files containing drunk spells in order to make the crash happen, not sure why though. Sometimes it crashes, sometimes it doesn't, the more files the easier it is to replicate.


EPuncker:
great! I would like to hear @soul4soul input on this as well

emil92b:
https://github.com/otland/forgottenserver/blob/master/data/scripts/lib/register_monster_type.lua#L337-L354
`{name ="condition", interval = x, chance = x, condition = { type = x , startDamage = x, minDamage = x, maxDamage = x, duration = x, interval = x}`

theshibbies:
Pretty sure you'll still get an error from here.
https://github.com/otland/forgottenserver/blob/b73ac8c77bd81471eef587f9d0e7c3b83256895c/data/scripts/lib/register_monster_type.lua#L281

emil92b:
https://github.com/otland/forgottenserver/pull/3499
Allow conditions to be used with all predefine creature abilities. Conditions can be used with "melee", "combat", "speed", "condition", "outfit", "invisible", "*field", etc.. Conditions won't work with named spells such as "energy strike".

theshibbies:
I see. Well either way it still causes an error.
image

theshibbies:
> #3499 Allow conditions to be used with all predefine creature abilities. Conditions can be used with "melee", "combat", "speed", "condition", "outfit", "invisible", "*field", etc.. Conditions won't work with named spells such as "energy strike".

Alright, great. I guess @soul4soul will have to check into adapting his system for this format. I reverted the changes for that part. However, the drunk spells are still an issue.

data/monster/lua/#example.lua
```diff
@@ -56,18 +56,22 @@ monster.loot = {
...
+	{name = "combat", type = COMBAT_LIFEDRAIN, chance = 10, interval = 2*1000, length = 8, spread = 0, minDamage = -300, maxDamage = -490, effect = CONST_ME_PURPLEENERGY},
 	{name = "speed", chance = 15, interval = 2*1000, speed = -700, radius = 1, target = true, duration = 30*1000, effect = CONST_ME_MAGIC_RED},
 	{name = "firefield", chance = 10, interval = 2*1000, range = 7, radius = 1, target = true, shootEffect = CONST_ANI_FIRE},
-	{name = "combat", type = COMBAT_LIFEDRAIN, chance = 10, interval = 2*1000, length = 8, spread = 0, minDamage = -300, maxDamage = -490, effect = CONST_ME_PURPLEENERGY}
+	{name = "condition", chance = 1000, range = 1, condition = { type = CONDITION_ENERGY, minDamage = 10, maxDamage = 20, duration = 5000, interval = 1000}},
+	{name = "drunk", chance = 10, interval = 2000, drunkenness = 25, duration = 5000, range = 7, shootEffect = CONST_ANI_ENERGY},
+	{name =	"outfit", chance = 10, interval = 2000, item = 7172, duration = 8000, range = 7, effect = CONST_ME_MAGIC_BLUE}
```
soul4soul, line 66: Maybe leave this one off until the outfit attack is fixed?  #3884

soul4soul (commented):
Yeah #3499 has the best examples, for real tibia spells and conditions only these two formats would be used

This gives melee with a condition
`{name = "melee", attack = 10, skill = 100, condition = { type = CONDITION_FIRE, minDamage = 50, maxDamage = 200, duration = 5000, interval = 1000}},`

This is an attack that causes a condition only.
`{name = "condition", chance = 1000, range = 1, condition = { type = CONDITION_ENERGY, minDamage = 10, maxDamage = 20, duration = 5000, interval = 1000}},`

Looking at the source code of the converter it looks correct but I'll actually test it when I get a chance.

soul4soul (approved):

soul4soul:
This could use a rebase and add back the outfits spells and it would be good to go.

ranisalt (dismissed):
Rebased against master

data/scripts/lib/register_monster_type.lua
```diff
@@ -331,12 +331,9 @@ local function AbilityTableToSpell(ability)
 			if outfit then
 				spell:setOutfit(outfit)
 			end
-			if ability.name == "drunk" then
-				spell:setConditionType(CONDITION_DRUNK)
-				if ability.drunkenness then
-					spell:setConditionDrunkenness(ability.drunkenness)
-				end
-			end
+      if ability.drunkenness then
```
EPuncker, line 334: indentation

EPuncker (approved):

PR #3915, opened 2022-01-29
Actions cleanup

DSpeichert:
### Pull Request Prelude

### Changes Proposed

* Don't run non-release actions on tags
* Run specific release-building actions on tags
* Updated `lukka/run-vcpkg` and `lukka/run-cmake` to latest version compatible (there is newer v10 that would require changes to cmake)
* Removed Travis CI and AppVeyor configs as we no longer use these services
* Updated README with up-to-date badges

The packaged MacOS archive seems odd, or specifically, the build artifacts for MacOS coming from cmake are "odd", but I'm unable to verify that. You can check it out at https://github.com/DSpeichert/forgottenserver/releases/tag/v0.1.1


Znote (approved):
image

[Works](https://github.com/Znote/forgottenserver/actions/runs/1783438479), but got a warning we might want to consider:

Build with vcpkg : .github#L1
windows-latest workflows will use windows-2022 soon. For more details, see https://github.com/actions/virtual-environments/issues/4856

Also, vcpkg takes forever to compile on every action (12-25 minutes), we might want to upgrade to 10.1 and utilize the artifact cache, if I understand correctly it will prevent us rebuilding the same version of vcpkg on every action?

And since Macos isnt really that used, perhaps we can skip the luajit off checks and just keep luajit on for macos, saving us 2 steps?

DSpeichert:
@Znote I have a follow-up where I will investigate what to do with MacOS further.

PR #3922, opened 2022-02-04
Upgrade raids to lua under globalevents

andremiles1:
### Pull Request Prelude

### Changes Proposed

Hello guys.
I was thinking... The raid system being separated from globalevents is kind of redundant. Globalevents does the same, and it's not xml so you can do better, more detailed scripts for certain raids. I've included this script I made, but this is just an example, since it lacks delayed raids or multiple stages raids. So the script needs to be upgraded. Keep in mind that my script is designed run with a chance of no raids occurring, so an interval of 30 minutes is advised.
It's also a decision on if we should remove the raid system we currently use, so let's discuss this, please.

**Issues addressed:** 


EPuncker:
#1813

DSpeichert (changes requested):
I think your indentation is off.

ranisalt:
I have updated the changes to mirror what would change in #1813 and added a message to raids.xml

ranisalt (approved):

EPuncker:
closed as #1813 were implemented

PR #3964, opened 2022-02-21
according to the name getNextPosition should return something

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
The name of this function is somewhat misleading, since in Lua we are used to functions starting with `get` returning something, so I've added the `return self`

before:
```lua
position:getNextPosition(player:getDirection())
local tile = Tile(position)
```

after:
```lua
local tile = Tile(player:getPosition():getNextPosition(player:getDirection()))
```

This does not break compatibility with previous versions and gives us a new possibility, which should always have been the case.

### DIFF
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -16,6 +16,7 @@ function Position:getNextPosition(direction, steps)
 		self.x = self.x + offset.x * steps
 		self.y = self.y + offset.y * steps
 	end
+	return self
 end
 
 function Position:moveUpstairs()


EPuncker (approved):

ranisalt:
Agreed, and I think it should not modify the value of `self`.

nekiro:
> Agreed, and I think it should not modify the value of `self`.

this was always weird to me too, though, if people use that and you change that it will break a lot of stuff xD

MillhioreBT:
> Agreed, and I think it should not modify the value of `self`.

I agree with you, however as nekiro says this would break a lot of scripts that users use on their servers and a lot of otland content, we have to live with this detail

DSpeichert (commented):
Add a comment that this function behaves oddly but the behavior is now documented and preserved.

ranisalt (approved):

PR #3966, opened 2022-02-22
Removed draw well from teleport action

EPuncker:
### Pull Request Prelude

### Changes Proposed
No draw well at all should be teleporting players a floor down, that is not default action and would lead to scenarios like this:
2022-02-22 03_40_09-baiakNFT 8 6

### DIFF
--- a/data/actions/actions.xml
+++ b/data/actions/actions.xml
@@ -152,6 +152,7 @@
 	<!-- Other -->
 	<action itemid="430" script="other/teleport.lua" />
 	<action itemid="486" script="other/snow_heap.lua" />
+	<action itemid="1369" script="other/draw_well.lua" />
 	<action itemid="1386" script="other/teleport.lua" />
 	<action fromid="1728" toid="1731" script="other/watch.lua" />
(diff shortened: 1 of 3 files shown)


MillhioreBT (dismissed):

omarcopires:
Isn't there a way to check if there's a floor available instead of removing the functionality?

ranisalt:
> Isn't there a way to check if there's a floor available instead of removing the functionality?

This should be added specifically to each well that needs it, by action instead of itemid

EPuncker:
@ranisalt ok, ready to go and tested, and the map part was already merged

omarcopires:
The value is already used in data/lib/core/actionIds
`sandHole = 100, -- hidden sand hole`

ranisalt:
I thought it would use actionid instead of itemid. In this case I don't see any difference from having it inside `teleport.lua` or not.

ranisalt:
It will conflict with sandhole as @omarcopires said, if you use a shovel on the draw well and the ground is sand.

EPuncker:
ok fixed

ranisalt (approved):

PR #3977, opened 2022-02-24
Removing authors file

omarcopires:
There is no reason to keep this file, especially as it is a somewhat impartial contributors file. Several other people contribute to the project directly or indirectly and this file does not list all of them!
The fairest way to credit contributors is present on github itself and is evident on the first page of the repository, you can access it through this link: https://github.com/otland/forgottenserver/graphs/contributors


EPuncker:
authors file will stay because that is what most of projects use it nowadays and it was requested/suggested by project admins.
- yes, I agree the file have lots of missing names, it was generated in a hurry by request
- it will be reworked in the future
- its not the final/definitive version as it is intended to go with "releases" (master branch is a Work In Progress)
- someone complained on other commit that we added name of people that only contributed once or something and they should be removed... no, AUTHORS file is intended to have the names of everyone that ever contributed to the project code, unfortunately we don't have SVN names saved...

https://opensource.google/documentation/reference/releasing/authors

soul4soul:
Revision history already contains the information so I don't see the point of AUTHORS file.

ranisalt:
> Revision history already contains the information so I don't see the point of AUTHORS file.

It does not, as @EPuncker noted the past contributors from before the repository was moved to git are not included. Also, if someone downloads the zipped code from github, it will not include the list of commits/authors

MillhioreBT:
I don't see what's wrong with this file, no matter if it exists or not, we still have the list of github contributors, I'm not offended by not being listed, Anyway, I'm not a TFS developer, I only help with basic things and I'm very happy to help as much as possible.
close pr? XD

marmichalski:
> > Revision history already contains the information so I don't see the point of AUTHORS file.
> 
> It does not, as @EPuncker noted the past contributors from before the repository was moved to git are not included. Also, if someone downloads the zipped code from github, it will not include the list of commits/authors

That's the thing I always do when downloading something. Snoop for the authors file.
What if I wanted my name removed/pseudonymised/anonymised/whatever from this file (repository)? Now you'll have to rewrite the whole repository history anyway.

DSpeichert:
You're proposing a change that is exactly opposite to what was just introduced. If you think the file has wrong contents, you could offer improvements. Opening a PR to delete the whole thing without prior discussion is not helpful.

PR #3983, opened 2022-03-03
Monster onSpawned equivalent to onLogin in monsters

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes propose a new event for monsters: `onSpawned`.
many people still do not fully understand the event `onSpawn`, This event is used to verify certain things of a monster that does not yet exist, this monster, since it does not exist, has certain restrictions, such as not having a defined ID and other things.

With this new event we will make sure to act on a monster that already exists and was successfully placed on the map.

We currently can't do this with auto-spawned monsters with the spawn system:
```lua
local m = Game.createMonster(...)
if m then
-- m:somethingMethod(...)
end
```
with this event it will be possible to handle the monster properly without having to opt for a custom generation system from lua.

### Notes
it is possible that there is a different way to deal with this small inconvenience when using `onSpawn`, I even have some other ideas like returning a function in `onSpawn` to act on the correctly spawned monster, something like this:
```lua
local ec = EventCallback

function ec.onSpawn(monster, position, startup, artificial)
	local mType = monster:getType()
	if mType:getName() == "Demon" then
		return false
	end

	return function (monster, position, startup, artificial)
		-- Yeah, I'm not a demon and I'm already alive on the map :D
	end
end

ec:register(--[[0]])
```
however I think it's too much trouble to create the interface to return a function, save the reference on the side of C++, and then call her back.., but we can try it if this PR is not very satisfactory, maybe other ideas?

I personally find the idea of ​​returning a function sexy, this way we would avoid having another event.

### DIFF
--- a/data/events/events.xml
+++ b/data/events/events.xml
@@ -39,4 +39,5 @@
 	<!-- Monster methods -->
 	<event class="Monster" method="onDropLoot" enabled="1" />
 	<event class="Monster" method="onSpawn" enabled="0" />
+	<event class="Monster" method="onSpawned" enabled="0" />
 </events>
--- a/data/events/scripts/monster.lua
+++ b/data/events/scripts/monster.lua
@@ -15,3 +15,9 @@ function Monster:onSpawn(position, startup, artificial)
 		return true
 	end
 end
+
+function Monster:onSpawned(position, startup, artificial, forced)
+	if EventCallback.onSpawned then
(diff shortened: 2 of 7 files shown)


MillhioreBT:
I'll close it, because there seems to be a simple and easy solution for this with the same onSpawn event, we'll see...

PR #3987, opened 2022-03-06
Replace CMake with Meson

ranisalt:
Will write proposal shortly


ranisalt:
For fucks sake github controls are so terrible. Accidentally opened this way before it's ready.

PR #3993, opened 2022-03-07
Normalize include guards to FS_<FILENAME>_H

ranisalt:
### Pull Request Prelude

### Changes Proposed

Let's have some standardization here. Prepended `FS_` to every guard to avoid name clashes on common names.

We could alternatively use `#pragma once` instead, the code style suggests it but it was never really used.

### DIFF
--- a/src/account.h
+++ b/src/account.h
@@ -1,8 +1,8 @@
 // Copyright 2022 The Forgotten Server Authors. All rights reserved.
 // Use of this source code is governed by the GPL-2.0 License that can be found in the LICENSE file.
 
-#ifndef FS_ACCOUNT_H_34817537BA2B4CB7B71AA562AFBB118F
-#define FS_ACCOUNT_H_34817537BA2B4CB7B71AA562AFBB118F
+#ifndef FS_ACCOUNT_H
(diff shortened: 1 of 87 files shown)


EPuncker (approved):

Erza:
I like the pragma once idea, it's a compiler extension, but basically every single compiler supports it, and it's cleaner than these custom include guards.

https://en.wikipedia.org/wiki/Pragma_once#Portability

ranisalt:
Just to summarize:

Include guards
- Defined in the standard :+1: 
- Optimized by compilers :+1: 
- Works with symlinks (that we should not use anyway) :shrug: 
- Allows importing the same file from multiple places (useful for libraries, not for us) :shrug: 
- Error prone (mismatched `#ifndef`/`#define` tokens, missing `#endif`) :-1: 
- Prone to collision (i.e. two files using the same token) :-1: 

`#pragma once`
- Optimized by definition :+1: 
- Two lines less :shrug: 
- Does not work with symlinks :shrug: 
- Never collides, but allows importing twice from different places :shrug: 
- Not standard (but available in every compiler that makes sense) :-1:

PR #3996, opened 2022-03-09
Fixed variable type in npc.cpp

### LINKED ISSUE #3980: OnCreatureSay out of range value error

1. Go to any shop NPC and say trade

### Expected behaviour
no error

### Actual behaviour
download

EPuncker:
### Pull Request Prelude

### Changes Proposed
Changed from uint to int because we pass -1 at buy/sell

**Issues addressed:** 
Closes #3980

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -3570,8 +3570,8 @@ void ProtocolGame::AddShopItem(NetworkMessage& msg, const ShopInfo& item)
 
 	msg.addString(item.realName);
 	msg.add<uint32_t>(it.weight);
-	msg.add<uint32_t>(item.buyPrice == std::numeric_limits<int64_t>::max() ? 0 : item.buyPrice);
-	msg.add<uint32_t>(item.sellPrice == std::numeric_limits<int64_t>::max() ? 0 : item.sellPrice);
+	msg.add<uint32_t>(item.buyPrice == -1 ? 0 : item.buyPrice);
+	msg.add<uint32_t>(item.sellPrice == -1 ? 0 : item.sellPrice);
 }
 
 void ProtocolGame::parseExtendedOpcode(NetworkMessage& msg)
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -3570,8 +3570,8 @@ void ProtocolGame::AddShopItem(NetworkMessage& msg, const ShopInfo& item)
 
 	msg.addString(item.realName);
 	msg.add<uint32_t>(it.weight);
-	msg.add<uint32_t>(item.buyPrice == -1 ? 0 : item.buyPrice);
-	msg.add<uint32_t>(item.sellPrice == -1 ? 0 : item.sellPrice);
+	msg.add<uint32_t>(std::max<uint32_t>(item.buyPrice, 0));
+	msg.add<uint32_t>(std::max<uint32_t>(item.sellPrice, 0));
 }
 
 void ProtocolGame::parseExtendedOpcode(NetworkMessage& msg)
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -3570,8 +3570,8 @@ void ProtocolGame::AddShopItem(NetworkMessage& msg, const ShopInfo& item)
 
 	msg.addString(item.realName);
 	msg.add<uint32_t>(it.weight);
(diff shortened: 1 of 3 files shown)


ranisalt (dismissed):

nekiro:
> For consistency with the expected value `uint32_t` we can't use `int32_t` since we would lose half of the available numbers, so we should use `int64_t` in this case

Or stop passing -1

src/protocolgame.cpp
```diff
@@ -3570,8 +3570,8 @@ void ProtocolGame::AddShopItem(NetworkMessage& msg, const ShopInfo& item)
 
 	msg.addString(item.realName);
 	msg.add<uint32_t>(it.weight);
-	msg.add<uint32_t>(item.buyPrice == std::numeric_limits<uint32_t>::max() ? 0 : item.buyPrice);
-	msg.add<uint32_t>(item.sellPrice == std::numeric_limits<uint32_t>::max() ? 0 : item.sellPrice);
+	msg.add<uint32_t>(item.buyPrice == std::numeric_limits<int64_t>::max() ? 0 : item.buyPrice);
+	msg.add<uint32_t>(item.sellPrice == std::numeric_limits<int64_t>::max() ? 0 : item.sellPrice);
```
ranisalt, line 3574: `-1` in both cases

ranisalt (changes requested):

src/protocolgame.cpp
EPuncker: isn't that going to affect the check here?
https://github.com/otland/forgottenserver/blob/33bca9feb5d3e0bf03f450f035cbcc7590b054ae/src/protocolgame.cpp#L1810

ranisalt: I meant, use -1 instead of `std::numeric_limits<int64_t>::max()`. It would work previously because `static_cast<uint32_t>(-1) == std::numeric_limits<uint32_t>::max()` but it is not true with `int64_t`

nekiro: honestly I would drop these checks entirely, why are we checking that?
use `std::max<uint32_t>(item.sellPrice, 0);`?

ranisalt: Even better

ranisalt (approved):
LGTM

ramon-bernardo:
Shouldn't we use `price > 0`?
```
auto it = std::find_if(shopItemList.begin(), shopItemList.end(), [itemId](const ShopInfo& shopInfo) { return shopInfo.itemId == itemId && shopInfo.sellPrice != 0; });
```
```

return shopInfo.itemId == itemId && shopInfo.buyPrice != 0 && (!itemType.isFluidContainer() || shopInfo.subType == subType);
```
and `price < 1` or `price <= 0`
```
bool Player::hasShopItemForSale() {
    ...
    if (shopInfo.sellPrice == 0) { ... }
    ...
}
```

src/protocolgame.cpp
```diff
@@ -3570,8 +3570,8 @@ void ProtocolGame::AddShopItem(NetworkMessage& msg, const ShopInfo& item)
 
 	msg.addString(item.realName);
 	msg.add<uint32_t>(it.weight);
-	msg.add<uint32_t>(item.buyPrice == std::numeric_limits<uint32_t>::max() ? 0 : item.buyPrice);
-	msg.add<uint32_t>(item.sellPrice == std::numeric_limits<uint32_t>::max() ? 0 : item.sellPrice);
+	msg.add<uint32_t>(std::max<uint32_t>(item.buyPrice, 0));
+	msg.add<uint32_t>(std::max<uint32_t>(item.sellPrice, 0));
```
marmichalski, line 3574: This does not work as you think it does. This first converts both values to the template type (uint32_t). Since the buyPrice/sellPrice can be -1, this will just overflow and return that value instead of 0.

nekiro: Then just use int32 instead of u32

MillhioreBT: > This does not work as you think it does. This first converts both values to the template type (uint32_t). Since the buyPrice/sellPrice can be -1, this will just overflow and return that value instead of 0.
GIF 19-07-2022 02-47-55 p  m

Confirmed!
I guess a more explicit check should be done since buyPrice and sellPrice are of type `int64_t`

A possible solution:
```cpp
std::max<int64_t>(std::min<int64_t>(number, std::numeric_limits<uint32_t>::max()), 0);
```

PR #4002, opened 2022-03-10
[WIP] Update blessing

ArturKnopik:
Update blessing to 12+ version

new config:
enableUnknownPackage - enable printing unknow package sended to server, default disabled

DONE: 
image

TODO:
Send blessing window
image

possible problems:
slightly different behavior from Global

Need help with query cleaning old blessing value (all players)


MillhioreBT:
You should add 1 feature per PR to make it more readable
here you are adding an update to the blessings system and on top of that an extra with a setting to enable printing of unknown packets

in TFS we have a lua style for lua codes
you are explicitly comparing `if value == true then` when you can simply do `if value then`
and for loops you should omit the increment value if it is equal to the default
`for i = 1, max_n, 1 do` when you can just do `for i = 1, max_n do`

btw, Why add the setting to enable printing of unknown packages? this can be done manually in protocolgame.cpp file and compiled in 5 seconds, Adding this check to parsePacket doesn't look good and we don't even have an onParsePacket event to have a real reason to use it, and in case of implementing this event, it could have a configuration to monitor all packets.

just my opinion <3, please don't take this the wrong way, I'm glad you're adding 12+ stuff to TFS
and sorry for my english

joseluis2g:
this blessings1 to n naming is kinda ugly imo, what about using blobs for this data, its not like we read this anywhere else.

ArturKnopik:
I don't want to compile engine every time to turn on unknown package, for me it can help develop protocol staff faster, if needed it can be removed

ArturKnopik:
> this blessings1 to n naming is kinda ugly imo, what about using blobs for this data, its not like we read this anywhere else.

i have lack of knowledge in mysql

PR #4005, opened 2022-03-11
Allow to get monster storages in the onDeath event

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Important: This change is only mergeable if #3700 is merged.
This allows us to use monster storages in the onDeath event.

### What happens?
what happens is that the event [`onDeath` to remove monster storages](https://github.com/otland/forgottenserver/blob/cf28d1cfa1f56813ab35f7e6b7f11dc040fa7214/data/scripts/monsters/monster_storages.lua#L20) it is activated when the monster dies but not always this event is activated at the end, so if we try to get the storage in another onDeath event there is a possibility of getting an error because the storage was already removed earlier.

By using the `addTask` function we make sure that deleting the monster's storage happens after the interpreter finishes executing all onDeath events.

### DIFF
--- a/data/scripts/monsters/monster_storages.lua
+++ b/data/scripts/monsters/monster_storages.lua
@@ -17,7 +17,9 @@ end
 local creatureEvent = CreatureEvent("MonsterStorages")
 
 function creatureEvent.onDeath(self)
-    MonsterStorages[self:getId()] = nil
+    addTask(function (id)
+        MonsterStorages[id] = nil
+    end, self:getId())
     return true
 end
 


nekiro:
No, thats a big no, I dont know how addTask work exactly, cause I didnt check, but if you would put addEvent like you did with addTask that would be a huge performance issue, with many players killing creatures frequently. I guess it isnt that bad if you register storage value for only monsters using them, but with more creatures marked with monster storages this will still be an issue

MillhioreBT:
> No, thats a big no, I dont know how addTask work exactly, cause I didnt check, but if you would put addEvent like you did with addTask that would be a huge performance issue, with many players killing creatures frequently. I guess it isnt that bad if you register storage value for only monsters using them, but with more creatures marked with monster storages this will still be an issue

It's not like 10000 tasks are going to be added to the queue XD, literally once the task enters the queue and the monster dies once its death is finished the task is executed and not even 2 will accumulate XD

PR #4010, opened 2022-03-13
Use a variable as a reference of database instance

ramon-bernardo:
### Pull Request Prelude

### (Reverted - Changes Proposed)
- Use ref variable with database instance.
- Code more clean.

### Changes Proposed
- Use a variable as a reference of db instance instead of 2?

### DIFF
--- a/src/ban.cpp
+++ b/src/ban.cpp
@@ -94,5 +94,6 @@ bool IOBan::isIpBanned(uint32_t clientIP, BanInfo& banInfo)
 
 bool IOBan::isPlayerNamelocked(uint32_t playerId)
 {
-	return Database::getInstance().storeQuery(fmt::format("SELECT 1 FROM `player_namelocks` WHERE `player_id` = {:d}", playerId)).get();
+	Database& db = Database::getInstance();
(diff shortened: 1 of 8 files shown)


nekiro:
Why? If that variable isn't reused then what's the point of defining it?

kygov:
It could be "more clean" with changes that Saiyans did, instead calling Database::getInstance just g_database.X.

https://github.com/andersonfaaria/optimized_forgottenserver/blob/master/src/ban.cpp#L100

ranisalt:
It would be more clean without the globals. Just pass the reference around

ramon-bernardo:
`g_database`, `Database::getInstance()` or `Database& db = Database::getInstance()`?

MillhioreBT:
it's not necessary at all, and it doesn't look cleaner either, I just see more unnecessary instructions
`Create a reference variable`, `Destroy the reference when the function returns` XD

DSpeichert (commented):
I don't think it's worth making a change.

ramon-bernardo:
Alright, Thanks!

PR #4015, opened 2022-03-15
Refactor getHoldingPlayer method

nekiro:
If player is holding an item then top parent of that item is always player, which means, we don't need to check every single parent going up the hierarchy.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -336,17 +336,15 @@ uint16_t Item::getSubType() const
 	return count;
 }
 
-Player* Item::getHoldingPlayer() const
+const Player* Item::getHoldingPlayer() const
 {
-	Cylinder* p = getParent();
-	while (p) {
-		if (p->getCreature()) {
-			return p->getCreature()->getPlayer();
-		}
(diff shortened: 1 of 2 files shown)


EPuncker (dismissed):

src/item.cpp
```diff
@@ -336,17 +336,15 @@ uint16_t Item::getSubType() const
...
-		p = p->getParent();
+	const Cylinder* topParent = getTopParent();
+	if (!topParent) {
+		return nullptr;
 	}
-	return nullptr;
+
+	const Creature* topParentCreature = topParent->getCreature();
+	return topParentCreature ? topParentCreature->getPlayer() : nullptr;
 }
```
ranisalt, line 348: This `get<Type>()` pattern aged like milk. You should be able to just return `return dynamic_cast<Player*>(getTopParent())`;

MillhioreBT (approved):

EPuncker (approved):

DSpeichert (approved):

PR #4022, opened 2022-03-18
Move all system headers to PCH

ranisalt:
### Pull Request Prelude

### Changes Proposed

Further improves #4019 by adding all system headers to the precompiled header. According to [this and several other sources](https://stackoverflow.com/questions/23904353/what-to-add-to-precompiled-headers), every included header that is not expected to change often - that is, usually all system headers - are good candidates.

There is a notable improvement from ~2:40 to less than 2 minutes build in my machine.

### DIFF
--- a/src/otpch.h
+++ b/src/otpch.h
@@ -7,24 +7,49 @@
 // Definitions should be global.
 #include "definitions.h"
 
+// System headers required in headers should be included here.
 #include <algorithm>
+#include <array>
+#include <atomic>
+#include <bitset>
 #include <boost/asio.hpp>
+#include <boost/iostreams/device/mapped_file.hpp>
+#include <boost/lockfree/stack.hpp>
+#include <boost/variant.hpp>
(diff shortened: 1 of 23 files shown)


ranisalt:
@DSpeichert was faster than me, sorry for the second PR :P

DSpeichert (approved):

src/otpch.h
```diff
@@ -7,24 +7,49 @@
...
 #include <string>
 #include <thread>
 #include <unordered_map>
 #include <unordered_set>
+#include <utility>
+#include <valarray>
+#include <variant>
 #include <vector>
 
+#if __has_include("luajit/lua.hpp")
```
nekiro, line 49: didn't we drop this __has_include thing for some reason?

DSpeichert: I don't remember exactly, but it could have something to do with the CMAKE's `-DUSE_LUAJIT` not working right.

PR #4036, opened 2022-03-22
Added Player:getAttackFactor()

Codinablack:
### Pull Request Prelude

- [x ] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Adds lua method for Player:getAttackFactor()

**Issues addressed:** None

### DIFF
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -372,3 +372,12 @@ function Player.addAllMounts(self)
 		self:addMount(mounts[mount].id)
 	end
 end
+
+function Player.getAttackFactor(self)
+	local factors = {
+		FIGHTMODE_ATTACK = 1.0,
+		FIGHTMODE_BALANCED = 1.2,
+		FIGHTMODE_DEFENSE = 2.0
+	}
+	return factors[self:getFightMode()]
+end


EPuncker (approved):

kygov:
I would use this function from source https://github.com/otland/forgottenserver/pull/4036, doesnt make sense making it twice.

Codinablack:
> I would use this function from source #4036, doesnt make sense making it twice.

What do you mean? Use it from source, yes it exists in the source, but there is no api for calling the method in lua, you are saying I should push the source method to lua? That makes no sense to me, as there is no need at all to alter source just to open this function up to the lua api when it is easily replicated in lua. I guess its preference.

gg-martins091:
> 

You need to think about the maintainability of the codebase long term. If we need to change the behaviour of this feature in the future, we would need to *remember* to do so in both places, thats simply a bad practice, especially when talking about such a small feature.

Codinablack:
> > That makes no sense to me, as there is no need at all to alter source just to open this function up to the lua api when it is easily replicated in lua. I guess its preference.
> 
> You need to think about the maintainability of the codebase long term. If we need to change the behaviour of this feature in the future, we would need to _remember_ to do so in both places, thats simply a bad practice, especially when talking about such a small feature that is easily reusable.

Ok so because something exists in the source already, we shouldn't have it in lua as well? If TFS followed that logic we wouldn't even have a lua API, furthermore, we most definitely wouldn't have a hundred functions written and rewritten three or for different ways just so everyone's scripts can work (compat.lua). W/e though, I only even wrote the function because someone else suggested it and convinced me they could help get it approved, which didn't happen so I'm closing it.

PR #4042, opened 2022-03-24
Add missing breaks in readAttr switch

nekiro:
Like the title says, I missed them when added reflect and boost


nekiro:
Duplicate of #4041

PR #4049, opened 2022-03-29
Ignore potions checks if group has access

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Now, players with access on your group ignore potions check (level and vocation).

### DIFF
--- a/data/actions/scripts/other/potions.lua
+++ b/data/actions/scripts/other/potions.lua
@@ -129,7 +129,7 @@ function onUse(player, item, fromPosition, target, toPosition, isHotkey)
 	end
 
 	local potion = potions[item:getId()]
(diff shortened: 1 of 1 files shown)


omarcopires:
The same should work for spells, ammo, and gear. Not sure of current behavior!

EPuncker (approved):

EPuncker:
> The same should work for spells, ammo, and equipments. Not sure of current behavior!

ignorespellcheck
ignoreweaponcheck

😃

ramon-bernardo:
That was quick, we can use flag here, but would add c++ code.

EPuncker:
> That was quick, we can use flag here, but would add c++ code.

there is no need 😄 same thing with level/access doors

PR #4051, opened 2022-03-30
Use built-in traceback lua method

nekiro:
Use built-in traceback from lua api


ranisalt (approved):
Great!

viperthepala:
Change breaks build process on linux (Debian 10) 
luascript.cpp:442:2: error: 'luaL_traceback' was not declared in this scope.

PR #4053, opened 2022-03-31
Fix addEvent scheduler values being passed by reference not value

### LINKED ISSUE #4031: after recent updates, npc does not respond to hi
after recent updates, npc does not respond to hi

tfs 1.5
otclient 12.81

no errors in the tfs console or otclient

nekiro:
Fixes #4031

Apparently (from what I read on the internet) std::bind copies values, #4017 introduces usage of lambdas and this was changed to ref [&] and broke npcs and probably a lot more, so the solution is to bring back passing by value.


EPuncker (approved):

ranisalt:
😡

PR #4057, opened 2022-04-01
Removed MOTD

### LINKED ISSUE #4056: [discussion] MOTD is gone from client
Remember that message of the day that used to be displayed before getting to the char select window?

It's gone on QT client, and replaced with this:

And it doesn't receive any info from server side (unless we add HTTP API).

What should we do with the [leftover code TFS-side](https://github.com/otland/forgottenserver/search?q=motd)?

EPuncker:
### Pull Request Prelude

### Changes Proposed
**GUILD MOTD HASN'T BEEN TOUTCHED!**

For those who don't know what MOTD is, it was the small "Message Of The Day" window that appeared once a day before your character select window...

It was replaced in QT client by a new widget on client bottom, and it doesn't need any server-sided info:
image

I removed leftover code as it is not present in QT client anymore (since version 11?)

OTC can still be edited easily to have it client-sided only. https://github.com/edubart/otclient/blob/9c9b85ac5e9ae3cbacfc2d4993517c886d8881cd/modules/gamelib/protocollogin.lua#L186-L189

**Issues addressed:** 
closes #4056 

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -291,12 +291,6 @@ bool ConfigManager::load()
 	return true;
 }
 
-bool ConfigManager::reload()
-{
-	bool result = load();
-	return result;
-}
-
 static std::string dummyStr;
 
 const std::string& ConfigManager::getString(string_config_t what) const
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -238,7 +238,6 @@ bool ConfigManager::load()
 	string[OWNER_EMAIL] = getGlobalString(L, "ownerEmail", "");
 	string[URL] = getGlobalString(L, "url", "");
 	string[LOCATION] = getGlobalString(L, "location", "");
-	string[MOTD] = getGlobalString(L, "motd", "");
 	string[WORLD_TYPE] = getGlobalString(L, "worldType", "pvp");
 
 	integer[MAX_PLAYERS] = getGlobalNumber(L, "maxPlayers");
@@ -295,9 +294,6 @@ bool ConfigManager::load()
 bool ConfigManager::reload()
 {
 	bool result = load();
-	if (transformToSHA1(getString(ConfigManager::MOTD)) != g_game.getMotdHash()) {
-		g_game.incrementMotdNum();
-	}
 	return result;
 }
 
--- a/src/protocolstatus.cpp
+++ b/src/protocolstatus.cpp
@@ -129,6 +129,9 @@ void ProtocolStatus::sendStatusString()
 	map.append_attribute("width") = std::to_string(mapWidth).c_str();
 	map.append_attribute("height") = std::to_string(mapHeight).c_str();
 
+	pugi::xml_node motd = tsqp.append_child("motd");
+	motd.text() = "N/A";
+
 	std::ostringstream ss;
 	doc.save(ss, "", pugi::format_raw);
 
@@ -157,6 +160,7 @@ void ProtocolStatus::sendInfo(uint16_t requestedInfo, const std::string& charact
 
 	if (requestedInfo & REQUEST_MISC_SERVER_INFO) {
 		output->addByte(0x12);
+		output->addString("N/A"); //MOTD
 		output->addString(g_config.getString(ConfigManager::LOCATION));
 		output->addString(g_config.getString(ConfigManager::URL));
 		output->add<uint64_t>((OTSYS_TIME() - ProtocolStatus::start) / 1000);
--- a/src/protocolstatus.cpp
+++ b/src/protocolstatus.cpp
@@ -129,9 +129,6 @@ void ProtocolStatus::sendStatusString()
 	map.append_attribute("width") = std::to_string(mapWidth).c_str();
 	map.append_attribute("height") = std::to_string(mapHeight).c_str();
 
-	pugi::xml_node motd = tsqp.append_child("motd");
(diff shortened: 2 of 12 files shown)


src/configmanager.cpp
```diff
@@ -295,9 +294,6 @@ bool ConfigManager::load()
 bool ConfigManager::reload()
 {
 	bool result = load();
-	if (transformToSHA1(getString(ConfigManager::MOTD)) != g_game.getMotdHash()) {
-		g_game.incrementMotdNum();
-	}
 	return result;
```
ramon-bernardo, line 297: 😄 
```suggestion
	return load();

File: src/protocolstatus.cpp [hunk 2]
```diff
@@ -129,9 +129,6 @@ void ProtocolStatus::sendStatusString()
 	map.append_attribute("width") = std::to_string(mapWidth).c_str();
 	map.append_attribute("height") = std::to_string(mapHeight).c_str();
 
-	pugi::xml_node motd = tsqp.append_child("motd");
-	motd.text() = g_config.getString(ConfigManager::MOTD).c_str();
-
```
Erza, line 134: I'd recommend at least returning an empty string here, othewise you might break server lists like otservlist or such.

src/protocolstatus.cpp
```diff
@@ -160,7 +157,6 @@ void ProtocolStatus::sendInfo(uint16_t requestedInfo, const std::string& charact
 
 	if (requestedInfo & REQUEST_MISC_SERVER_INFO) {
 		output->addByte(0x12);
-		output->addString(g_config.getString(ConfigManager::MOTD));
```
Erza, line 163: This will break server lists for sure, if any use the binary protocol.

src/configmanager.cpp
ranisalt: This function can be removed altogether. Just `load()` again.

Zbizu:
protocolstatus can return "N/A" string

DSpeichert (changes requested):
I'd leave the config option just to send it in XML and binary status protocol, at least while server lists use it.

Zbizu:
> I'd leave the config option just to send it in XML and binary status protocol, at least while server lists use it.

server lists are unlikely to back off considering that many servers are still running 8.6 and 10.98

EPuncker:
re-review please 😄

ranisalt (approved):
LGTM

src/protocolstatus.cpp
```diff
@@ -160,7 +160,7 @@ void ProtocolStatus::sendInfo(uint16_t requestedInfo, const std::string& charact
...
+		output->addString("N/A"); //MOTD
```
Zbizu, line 163: `// MOTD`

Zbizu (changes requested):

src/protocolstatus.cpp
EPuncker: don't worry, @ranisalt clang will fix that in the future I hope

DSpeichert:
> > I'd leave the config option just to send it in XML and binary status protocol, at least while server lists use it.
> 
> server lists are unlikely to back off considering that many servers are still running 8.6 and 10.98

That's why I think it's a feature to still be able to configure MOTD - as long as serverlists continue to show it, it's better to put something in there than just "N/A".

EPuncker:
> > > I'd leave the config option just to send it in XML and binary status protocol, at least while server lists use it.
> > 
> > 
> > server lists are unlikely to back off considering that many servers are still running 8.6 and 10.98
> 
> That's why I think it's a feature to still be able to configure MOTD - as long as serverlists continue to show it, it's better to put something in there than just "N/A".

nooo, it was a client feature, not otservlist feature, they can start to parse the new "MOTD" from server login.php now, it have lots of useful info: boosted creature id, players online, twitch viewers, event schedule etc

DSpeichert (approved):

LOLlolgrb:
It's still used by every single otlist, otclient is still using it. So why would it be removed?!? Jesus christ this repository has really fallen off the last couple of years...

Can we also update the readme to say that we're aiming to be a copy of tibia - not an open source mmorpg emulator and that you need to use the real tibia client to connect these days? That seems more accurate these days.

EPuncker:
> It's still used by every single otlist, otclient is still using it. So why would it be removed?!? Jesus christ this repository has really fallen off the last couple of years...
> 
> Can we also update the readme to say that we're aiming to be a copy of tibia - not an open source mmorpg emulator and that you need to use the real tibia client to connect these days? That seems more accurate these days.

if you are not happy, create your own fork and make your own tfs, enough with the drama... seems like some people just like following tfs/otland to show their disliking/hating/disaproval while contributing with nothing...

LOLlolgrb:
> > It's still used by every single otlist, otclient is still using it. So why would it be removed?!? Jesus christ this repository has really fallen off the last couple of years...
> > Can we also update the readme to say that we're aiming to be a copy of tibia - not an open source mmorpg emulator and that you need to use the real tibia client to connect these days? That seems more accurate these days.
> 
> if you are not happy, you can fork the repository and revert the patch, or not apply it to your current branch.

If you don't want any users, keep your great work up.

The beauty with this PR is that we can finally see what servers are using TFS 1.5. Let me create a server list in a year that filters with this new dumb MOTD message, and let's see how many servers are actively using this distro, it won't be many.

ranisalt:
@LOLlolgrb I am really eager to see your results, please ping me in the future.

PR #4067, opened 2022-04-04
monster loot charges

### LINKED ISSUE #3881: monster loot charges

currently if you dont set any charge it will be 1x by default so we should use `<attribute key="charges" value="200" />` in items.xml instead to make the charges max on drop

`monster.loot = {
{id = 2170, chance = 99999}
}`

emil92b:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 
loot with charges didn't set the charges to max

closes #3881

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -202,13 +202,11 @@ registerMonsterType.loot = function(mtype, mask)
 			if loot.aid or loot.actionId then
 				parent:setActionId(loot.aid or loot.actionId)
 			end
+			local charges = loot.charges or ItemType(loot.id):getCharges()
 			if loot.subType or loot.charges then
-				parent:setSubType(loot.subType or loot.charges)
+				parent:setSubType(loot.subType or charges)
 			else
-				local charges = ItemType(loot.id):getCharges()
-				if charges > 0 then
-					parent:setSubType(charges)
-				end
+				parent:setSubType(charges)
 			end
 			if loot.text or loot.description then
 				parent:setDescription(loot.text or loot.description)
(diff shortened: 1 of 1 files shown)


data/scripts/lib/register_monster_type.lua
```diff
@@ -225,6 +232,10 @@ registerMonsterType.loot = function(mtype, mask)
 					end
 					if children.subType or children.charges then
 						child:setSubType(children.subType or children.charges)
+					else
+						if charges > 0 then
+							child:setSubType(charges)
```
MillhioreBT, line 237: is it normal to place the default charges from parent to child? shouldn't I get the charges of the itemtype of the child id in question?

Zbizu: nah, should make a new variable from `ItemType(child.id):getCharges()`

emil92b: correct now?

Zbizu: haven't tested, but looks ok

emil92b: Untitled

Zbizu (dismissed):

data/scripts/lib/register_monster_type.lua
```diff
@@ -204,6 +204,11 @@ registerMonsterType.loot = function(mtype, mask)
 			end
 			if loot.subType or loot.charges then
```
nekiro, line 205: define charges as
`local charges = loot.charges or ItemType(loot.id):getCharges()`
then just use charges everywhere where loot.charges is used
charges > 0 check is probably redundant as it wont change anything, unless you want to save that single call

PR #4078, opened 2022-04-11
fix #4077 (monster id 0 in onSpawn event)

### LINKED ISSUE #4077: monster:onSpawn creature id is 0

### Steps to reproduce
1. create a script:
```
	local ec = EventCallback
	ec.onSpawn = function(self, position, startup, artificial)
		print(self:getId())
		return true
	end
	ec:register()
```
2, start the server

### Expected behaviour
output: monster id

### Actual behaviour
output: 0
note: monster methods work normally, but this could break a script that would rely on monster id

Zbizu:

### Changes Proposed
set monster ID before calling the script
**Issues addressed:** #4077

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -6,6 +6,7 @@
 #include "events.h"
 
 #include "item.h"
+#include "monster.h"
 #include "player.h"
 
 Events::Events() :
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -149,6 +149,9 @@ bool Events::eventMonsterOnSpawn(Monster* monster, const Position& position, boo
 	lua_State* L = scriptInterface.getLuaState();
 	scriptInterface.pushFunction(info.monsterOnSpawn);
 
+	// set monster ID earlier than usual so we can reference it in the script
+	monster->setID();
+	
 	LuaScriptInterface::pushUserdata<Monster>(L, monster);
 	LuaScriptInterface::setMetatable(L, -1, "Monster");
 	LuaScriptInterface::pushPosition(L, position);


nekiro:
This event is called too early, with that fix you are setting id twice, It will work cause setId has id == 0 check, but anyway that's workaround and is probably problematic

conde2:
In my opinion we should only call onSpawn after it successfully spawn.

conde2:
Can't we just call setId on each creature class consctructor?

Zbizu:
> I believe it sets the id after creature is placed to avoid waste of ids for unplaced creatures, this is though very rare, when creature is created but not placed, so your suggestion might be actually quite good, unless I missed something

the id pool is big enough to run the server for several months without restarting, even high exp ue spam
will never get exhausted if you restart server daily

Codinablack:
Would it not be best, for the end user, if "onSpawn" was a void, with the whole created monster passed after creation, and then there be another event "onPrepareSpawn" which could have more information such as,
 onPrepareSpawn(monsterType, spawnId, spawnBlock, centerPos, radius, #aliveInSpawn, startup)

ramon-bernardo (approved):

ramon-bernardo:
Simplest solution. Good!

> @Zbizu 
> the id pool is big enough to run the server for several months without restarting, even on high exp ue spam server the pool will never get exhausted if you restart server daily

I agree.

Any reason not to merge?

Zbizu:
@ramon-bernardo 
> Any reason not to merge?

Divided opinions on the solution.

I won't be working on this pr anymore so if you guys are not interested in merging it in its current state, you can close it.

ranisalt:
If no one is willing to dedicate time, I'd rather have the bug fixed.

PR #4079, opened 2022-04-17
Use string views to avoid copying strings

### LINKED ISSUE #4064: Change return type of getScriptBaseName to string_view
`getScriptBaseName` functions exist in event types (`Actions`, `Weapons`, etc) and return a constant string referring to the name of the class. It is possible to, instead of returning `std::string`, return `std::string_view` which is more lightweight and avoids copying data on every call (a `string` has to be created, where a `string_view` will just be a pointer to static data).

Tip: just search for `getScriptBaseName`, it's very likely that a search and replace is enough. Optionally, move all definitions to cpp files.

ranisalt:
### Pull Request Prelude

### Changes Proposed

Avoids some string copies by using `string_view` to wrap existing strings instead of copying to a new string instance. Most notably, avoids copying strings from database column names and query results, and avoids copying strings when reading prop streams.

Additionally, throws in a bit of code modernization by replacing output parameters with returning pairs and destructuring assignments.

**Issues addressed:**
Closes #4064 

### DIFF
--- a/src/baseevents.cpp
+++ b/src/baseevents.cpp
@@ -17,7 +17,7 @@ bool BaseEvents::loadFromXml()
 		return false;
 	}
 
-	std::string scriptsName = getScriptBaseName();
+	std::string scriptsName{getScriptBaseName()};
 	std::string basePath = "data/" + scriptsName + "/";
 	if (getScriptInterface().loadFile(basePath + "lib/" + scriptsName + ".lua") == -1) {
 		std::cout << "[Warning - BaseEvents::loadFromXml] Can not load " << scriptsName << " lib/" << scriptsName
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -44,8 +44,6 @@ void Actions::clear(bool fromLua)
 
 LuaScriptInterface& Actions::getScriptInterface() { return scriptInterface; }
 
-std::string Actions::getScriptBaseName() const { return "actions"; }
-
 Event_ptr Actions::getEvent(const std::string& nodeName)
 {
 	if (!caseInsensitiveEqual(nodeName, "action")) {
@@ -539,8 +537,6 @@ bool Action::loadFunction(const pugi::xml_attribute& attr, bool isScripted)
 	}
 	return true;
 }
-
-std::string Action::getScriptEventName() const { return "onUse"; }
 
 ReturnValue Action::canExecuteAction(const Player* player, const Position& toPos)
 {
--- a/src/actions.h
+++ b/src/actions.h
@@ -54,7 +54,7 @@ public:
 	ActionFunction function;
 
 private:
-	std::string getScriptEventName() const override;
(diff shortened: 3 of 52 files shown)


src/baseevents.cpp
```diff
@@ -17,7 +17,7 @@ bool BaseEvents::loadFromXml()
 		return false;
 	}
 
-	std::string scriptsName = getScriptBaseName();
+	std::string scriptsName{getScriptBaseName()};
 	std::string basePath = "data/" + scriptsName + "/";
```
ranisalt, line 21: Thinking about using `std::filesystem::path` here

DSpeichert (dismissed):

EPuncker (dismissed):

ranisalt:
Any server owner with a relatively large server willing to try and share the results? Can't properly measure with local benchmarks :sweat_smile:

EPuncker (dismissed):

yamaken93:
If we rely on fmt instead of c++ 20 formatting api, we can use fmt::memory_buffer buffer + fmt::format_to + string_view to prevent copies in queries and other stuff. Here a really small optimization based on this concept:

```
	fmt::memory_buffer buffer;
	const ItemType& it = Item::items[item->getID()];
	if (!it.showCount) {
		fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("Using one of {:s}..."), item->getName());
	} else if (count == 1) {
		fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("Using the last {:s}..."), item->getName());
	} else {
		fmt::format_to(std::back_inserter(buffer), FMT_COMPILE("Using one of {:d} {:s}..."), count, item->getPluralName());
	}
	player->sendTextMessage(MESSAGE_HOTKEY_PRESSED, std::string_view(buffer.data(), buffer.size()));
```

This is a "avoid" copying the final string at all costs approach. The fmt::memory_buffer is a 500 stack allocated buffer that is used internally by fmt::format function and here we are using it directly to avoid the copy (and extra allocations if the final string is bigger than 500 bytes) from the stack to a new allocated in the heap string which would happen if we used fmt::format function.

ranisalt:
@yamaken93 this is a good idea in general, but I would be careful not to overuse it

yamaken93:
> @yamaken93 this is a good idea in general, but I would be careful not to overuse it

I agree, its a "war" on allocation + copy idea. Can be used for strings retrieved from the lua API(when lua calls TFS C functions), strings from the protocol, etc.

ramon-bernardo (dismissed):

EPuncker (approved):

ranisalt:
@EPuncker @ramon-bernardo may I get an updated review? Rebased and fixed a conflict

ranisalt:
Rebased against master

EPuncker (approved):

MillhioreBT (approved):

PR #4083, opened 2022-04-17
Maintenance Mode

omarcopires:
### Pull Request Prelude

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -80,7 +80,7 @@ mysqlSock = ""
 -- may cause high CPU usage with many players and potentially affect performance!
 -- NOTE: forceMonsterTypesOnLoad loads all monster types on startup to validate them.
 -- You can disable it to save some memory if you don't see any errors at startup.
--- When activating maintenance mode, only players with a position above tutors will be able to login to the server
+-- When activating maintenance mode, only players with a position above tutors will be able to login to the server.
 allowChangeOutfit = true
 freePremium = false
 kickIdlePlayerAfterMinutes = 15
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -80,6 +80,7 @@ mysqlSock = ""
 -- may cause high CPU usage with many players and potentially affect performance!
 -- NOTE: forceMonsterTypesOnLoad loads all monster types on startup to validate them.
 -- You can disable it to save some memory if you don't see any errors at startup.
+-- When activating maintenance mode, only players with a position above tutors will be able to login to the server
 allowChangeOutfit = true
 freePremium = false
 kickIdlePlayerAfterMinutes = 15
@@ -96,6 +97,7 @@ premiumToSendPrivate = false
 forceMonsterTypesOnLoad = true
 cleanProtectionZones = false
 showPlayerLogInConsole = true
+maintenanceMode = false
 
 -- VIP and Depot limits
 -- NOTE: you can set custom limits per group in data/XML/groups.xml
--- a/data/creaturescripts/scripts/login.lua
+++ b/data/creaturescripts/scripts/login.lua
@@ -1,5 +1,5 @@
 function onLogin(player)
-	if configManager.getBoolean(configKeys.MAINTENCE_MODE) and player:getGroup():getId() < ACCOUNT_TYPE_TUTOR then
+	if configManager.getBoolean(configKeys.MAINTENANCE_MODE) and player:getGroup():getId() < ACCOUNT_TYPE_TUTOR then
 		return false
 	end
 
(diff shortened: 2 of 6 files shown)


gpedro:
and about /closeserver with `PlayerFlag_CanAlwaysLogin` flag?

data/creaturescripts/scripts/login.lua
```diff
@@ -1,4 +1,8 @@
 function onLogin(player)
+	if configManager.getBoolean(configKeys.MAINTENANCE_MODE) and player:getGroup():getId() < ACCOUNT_TYPE_TUTOR then
```
Erza, line 2: Comparing group ID with account type won't work

src/protocollogin.cpp
```diff
@@ -183,6 +184,12 @@ void ProtocolLogin::onRecvFirstMessage(NetworkMessage& msg)
 		return;
 	}
 
+	if (g_config.getBoolean(ConfigManager::MAINTENANCE_MODE)) {
+		disconnectClient("The Server is in maintenance mode and only players with positions above tutors can login.",
+		                 version);
+		return;
+	}
```
Erza, line 191: This will block anyone from logging in, regardless of account type or CanAlwaysLogin flag

data/creaturescripts/scripts/login.lua
Erza: Doing this in an onLogin script will also trigger other onLogin scripts, and you won't know which one ran first (and potentially already executed code by the time it hits this one)

config.lua.dist
```diff
@@ -80,6 +80,7 @@ mysqlSock = ""
 -- may cause high CPU usage with many players and potentially affect performance!
 -- NOTE: forceMonsterTypesOnLoad loads all monster types on startup to validate them.
 -- You can disable it to save some memory if you don't see any errors at startup.
+-- When activating maintenance mode, only players with a position above tutors will be able to login to the server.
```
Erza, line 83: This is actually already handled here https://github.com/otland/forgottenserver/blob/412950218c5e01062a5e2458d363802a5cea3115/src/protocolgame.cpp#L152-L154

You'd just have to give that flag to groups that you want to be able to log in.

src/protocollogin.cpp
Erza: Also, if anything, this shouldn't be done in protocollogin at all, it should be in protocolgame

config.lua.dist
```diff
@@ -96,6 +97,7 @@ premiumToSendPrivate = false
 forceMonsterTypesOnLoad = true
 cleanProtectionZones = false
 showPlayerLogInConsole = true
+maintenanceMode = false
```
Zbizu, line 100: this one is ok

src/configmanager.cpp
```diff
@@ -230,6 +230,7 @@ bool ConfigManager::load()
 	boolean[REMOVE_ON_DESPAWN] = getGlobalBoolean(L, "removeOnDespawn", true);
 	boolean[PLAYER_CONSOLE_LOGS] = getGlobalBoolean(L, "showPlayerLogInConsole", true);
 	boolean[TWO_FACTOR_AUTH] = getGlobalBoolean(L, "enableTwoFactorAuth", true);
+	boolean[MAINTENANCE_MODE] = getGlobalBoolean(L, "maintenanceMode", false);
```
Zbizu, line 233: this one can stay too

src/configmanager.h
```diff
@@ -51,6 +51,7 @@ class ConfigManager
 		REMOVE_ON_DESPAWN,
 		PLAYER_CONSOLE_LOGS,
 		TWO_FACTOR_AUTH,
+		MAINTENANCE_MODE,
```
Zbizu, line 54: this one is fine

src/luascript.cpp
```diff
@@ -2107,6 +2107,7 @@ void LuaScriptInterface::registerFunctions()
 	registerEnumIn("configKeys", ConfigManager::SERVER_SAVE_CLOSE);
 	registerEnumIn("configKeys", ConfigManager::SERVER_SAVE_SHUTDOWN);
 	registerEnumIn("configKeys", ConfigManager::ONLINE_OFFLINE_CHARLIST);
+	registerEnumIn("configKeys", ConfigManager::MAINTENANCE_MODE)
```
Zbizu, line 2110: this one is fine too

src/protocollogin.cpp
```diff
@@ -183,6 +184,12 @@ void ProtocolLogin::onRecvFirstMessage(NetworkMessage& msg)
 		return;
 	}
 
+	if (g_config.getBoolean(ConfigManager::MAINTENANCE_MODE)) {
```
Zbizu, line 187: this one is not needed if you make this line
https://github.com/otland/forgottenserver/blob/master/src/otserv.cpp#L374

depend on config value + give tutors flag `canAlwaysLogin`

the server could start in `GAME_STATE_CLOSED` state, this way it would be identical to `/closeserver` command

data/creaturescripts/scripts/login.lua
Zbizu, line 2: this one won't be needed if you apply other changes from my review

Zbizu (commented):
Overall I like the idea, just make some changes and resubmit

Issue #4086, opened 2022-04-20
onSpawn return false breaks spawnCheck


Codinablack:
### Before creating an issue, please ensure:
- [x ] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [ x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Add "if startup then return false end" to any onSpawn Script
2. Start server
3. Observe

### Expected behaviour
Monsters spawning during startup should not spawn, this is correct. However after x amount of time determined by spawnblock's interval, monster should try to spawn again.

### Actual behaviour
Monsters won't try to respawn after x amount of time. Meaning if you return false once in onSpawn at all, even if it's just during startup, then monsters will stop calling for a spawn check for that location in spawns.xml. 

### Environment
Windows 10 VS 2022 Pro compiler in 64x release mode.

### Reproduction environment
No specific scripts or anything fancy required, just return false for any scenario in onSpawn, and if that monster was spawning naturally from spawns.xml it will no longer even try to spawn again.

PR #4087, opened 2022-04-20
Fix onSpawn return false and missing creature id errors

Codinablack:
Co-Authored-By: Forgee <3932953+forgee@users.noreply.github.com>
 Special thanks to @MillhioreBT for her advice on where to put monster->decrementReferenceCounter();

### Pull Request Prelude

### Changes Proposed

[#4086]
[#4077]

This fixes the monster not being assigned a creature id during "onSpawn" event, it safely removes the monster if returned false, and fixes the problem of returning false with onSpawn breaking the internal spawn check system. 

Spent much time finding #4086 when trying to solve #4077, but it was a problem unknown and this pr has been tested pretty well, no errors found. 

### DIFF
--- a/src/spawn.cpp
+++ b/src/spawn.cpp
@@ -314,9 +314,6 @@ bool Spawn::spawnMonster(uint32_t spawnId, MonsterType* mType, const Position& p
                          bool startup /*= false*/)
 {
 	std::unique_ptr<Monster> monster_ptr(new Monster(mType));
-	if (!g_events->eventMonsterOnSpawn(monster_ptr.get(), pos, startup, false)) {
-		return false;
-	}
 
 	if (startup) {
 		// No need to send out events to the surrounding since there is no one out there to listen!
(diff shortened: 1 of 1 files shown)


ramon-bernardo (approved):
LGTM

amatria:
Just out of curiosity, why is the `spawnMonster` method creating a `std::unique_pointer` to immediately release the ownership of the stored object? Why bother creating an auto managed pointer if it is not going to be used at all? Am I missing something?

nekiro:
> Just out of curiosity, why is the `spawnMonster` method creating a `std::unique_pointer` to immediately release the ownership of the stored object? Why bother creating an auto managed pointer if it is not going to be used at all? Am I missing something?

To avoid memory leaks, when placeCreature fails

src/spawn.cpp
```diff
@@ -336,6 +333,12 @@ bool Spawn::spawnMonster(uint32_t spawnId, MonsterType* mType, const Position& p
 	monster->setSpawn(this);
 	monster->setMasterPos(pos);
 	monster->incrementReferenceCounter();
+	if (!g_events->eventMonsterOnSpawn(monster, pos, startup, false)) {
```
nekiro, line 336: not a fan of this, placing a monster to delete it, sounds like a waste

Codinablack: Then return false won't be fixed. 🤷‍♂️

PR #4089, opened 2022-04-23
Adding exceptions for odd door ids

torresmon:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Adding exceptions for odd door ids: 12692, 12695, 12701, 12703.

**Issues addressed:** 

Some door ids don't follow the general rule

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4090, opened 2022-04-24
Odd door ids potential solution

torresmon:
### Pull Request Prelude

### Changes Proposed
Adding a function that deal with any odd door id.

**Issues addressed:** 

Odd door ids making doors behave abnormally.
https://github.com/otland/forgottenserver/pull/4089#issuecomment-1107533786
Prevents the need to add exceptions for every door that don't follow the general id rule.
As a feature it could be used to make other items behave like normal doors I guess.

### DIFF
--- a/data/scripts/actions/others/doors.lua
+++ b/data/scripts/actions/others/doors.lua
@@ -53,6 +53,33 @@ local function findPushPosition(creature, round)
 	end
 end
 
+--[[
+First value should be lockedDoor or closedDoor(if no lock), second closedDoor, third openDoor.
+Odd ID doors must be added to this table and their respective ones. IDs don't matter, only their order.
+]]
+local oddDoors = {
+	[1] = {12692, 12692, 12695}, [2] = {12701, 12701, 12703}, [3] = {33431, 33430, 33433}, [4] = {33429, 33428, 33432}
+}
+
+local function findOddDoor(itemId, vi)
+	local k = 1
+	local v = 1
+	local newId = 0
+	for k, _ in pairs(oddDoors) do
+		for v, _ in pairs(oddDoors[k]) do
+			if itemId == oddDoors[k][v] then
+				if vi == nil then
+					return true
+				else
+					newId = oddDoors[k][vi]
+					return newId
+				end
+			end
+		end
+	end
+	return false
+end
+
 local door = Action()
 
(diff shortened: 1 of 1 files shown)


ranisalt:
What does this fix?

torresmon:
some new doors don't follow the common pattern id+1, id-1 and stuff, they end up transforming in stairs, walls and other stuff that has the id of that transformation

ranisalt:
In this case, instead of this "odd door" thing, all doors should be a pair of (open, closed) ids, even the ones that are just +1. Or, as I see, there is a third door for when it's locked, then that should too be added to all doors

torresmon:
In the past days I rewrote the main code many times searching for a solution, some of the attempts were huge list of all door ids kinda like this one, it proved to be a pain to manipulate(with my limited knowledge at least) and the function would need to go through like 250+ door ids to find the last one. I think the main function works very well for 99% of cases and function to find a exception is all it needs

Zbizu:
@ranisalt no need to map everything, mapping exceptions is enough
something like:
`if someTable[id] then item:transform(someTable[id]) else item:transform(itemId+1) end`

torresmon:
`item:transform(itemId+1)`  this is as smart as it's simple. I think it's the work of a talented doorsmith and it should be used unless the exception becomes the rule

torresmon:
> @ranisalt no need to map everything, mapping exceptions is enough something like:
> 
> ```lua
> if someTable[itemId] then
>     item:transform(someTable[itemId])
> else
>     item:transform(itemId+1)
> end
> ```

Windows also have some random ids and they are manipulated like you showed in the example, The windows list have one entry for each state: `[openWindow] = {closedWindw}, [closedWindow] = {openWindow}` and the "if someTable[itemId]". This is simple and works well for normalDoors too. I just don't know how i'd go about making the same work when the door have three possible states. The indexing function was the best I could think of

Zbizu:
@torresmon 
> I just don't know how i'd go about making the same work when the door have three possible states. 

four tables:
- doors open (transform to closed)
- doors closed (transform to open)
- doors locked ("it is locked" + key script)
- irregular transformations - as in example above - regular transformation if someTable[itemId] not present

torresmon:
yeah... i was trying to avoid doing that. i'll think about it some more.
thanks!

PR #4098, opened 2022-04-27
Registered missing weapons to destroy action

EPuncker:
### Pull Request Prelude

### Changes Proposed
TFS only had weapons up to **id 2453**, registered to the destroy action script (responsible for allowing players to destroy furniture using weapons), with this update, ALL missing weapons were added, and the script was moved to revscriptsys as well.

list with id/name for reference: https://pastebin.com/GQKeNGAD


EPuncker:
need testing, but should work

torresmon:
tested with random items with Ids lower and higher than 2453. tried using weapons from hand, backpack, hotkey, ground, browse field and inside corpses. No errors found, everything seems to be working as intended.

Codinablack:
I'm pretty sure weapon.onUse() is called when the weapon gets used as a weapon, in other words, when attacking something, since you can't attack items, I don't see this as being possible with this event, instead, should still be an **Action**, with **item.onUse()**, that would then have the desired outcome for this PR.

DSpeichert:
> I'm pretty sure weapon.onUse() is called when the weapon gets used as a weapon, in other words, when attacking something, since you can't attack items, I don't see this as being possible with this event, instead, should still be an **Action**, with **item.onUse()**, that would then have the desired outcome for this PR.

So you're saying this should not work. But it does? I'm confused but what you're saying is an important behavior detail to understand.

EPuncker:
> > I'm pretty sure weapon.onUse() is called when the weapon gets used as a weapon, in other words, when attacking something, since you can't attack items, I don't see this as being possible with this event, instead, should still be an **Action**, with **item.onUse()**, that would then have the desired outcome for this PR.
> 
> So you're saying this should not work. But it does? I'm confused but what you're saying is an important behavior detail to understand.

it was just a misunderstanding, it can be merged

DSpeichert (approved):

PR #4101, opened 2022-04-28
Merge loops over spectators in Game::placeCreature

vccolombo:
### Pull Request Prelude

### Changes Proposed

Game::placeCreature method has two different loops in sequence over the
same collection.

Merge them into a single one.

I was reading part of the code for the last 2 or 3 days and saw this code chunk, so I though about opening this PR. 
This code is kinda ancient, and I'm thinking if there is a reason for these two loops to be separate? The two reasons I could think about are 1. sendCreatureAppear has a side effect that we need to call it for everyone in this case before using onCreatureAppear; or 2. It's better to sendCreatureAppear to everyone as soon as possible for some network performance reason.
Is my change valid or there is a reason to be kept like it is today?

Thanks for your time

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -537,9 +537,6 @@ bool Game::placeCreature(Creature* creature, const Position& pos, bool extendedP
 		if (Player* tmpPlayer = spectator->getPlayer()) {
 			tmpPlayer->sendCreatureAppear(creature, creature->getPosition(), magicEffect);
 		}
-	}
-
-	for (Creature* spectator : spectators) {
 		spectator->onCreatureAppear(creature, true);
 	}
 


nekiro:
There was a discussion about this before and no one know why it's like that, but there was certainly a reason to write two separate loops instead of one, right now I didn't really check if this changes anything, but can potentially break things

ramon-bernardo (approved):
LGTM

'It seems to be okay to remove the loop.

MillhioreBT:
It looks good, but if this is changed, it will change the way players are notified of a appear creature.
especially when dealing with monsters, since they call the onAppear method, within this event you can do certain things with the monster that other players won't be able to see, because they haven't been notified in the loop yet.

And it's possible that some things will break too as nekiro said, but I can't confirm that right per now.

It might look silly to have two of the same loops, but this ensures that all players know about the creature, before the creature triggers the possible onAppear callback.

I think I've already mentioned this in another PR#, I don't remember what it was... however if I'm wrong or if the problem I mention is not relevant let me know.

vccolombo:
I see. I was not able to find this previous discussion on the topic, and also couldn't grasp if there was something that I wasn't understanding that caused this necessity to separate the loops
I think in this case, it's better to not apply the changes. It's just a very minor code cleaning, and is not worth any risk of causing new issues

Thanks everyone! :)

PR #4103, opened 2022-04-28
fix revscriptsys monster condition crashes

### LINKED ISSUE #4102: CONDITION_DRUNK crash in registerMonsterType

### Steps to reproduce
1. create a lua monster with attack involving condition drunk, for example
```{name ="drunk", interval = 2000, chance = 100, effect = CONST_ME_STUN, shootEffect = CONST_ANI_BURSTARROW, radius = 3, range = 4, target = true},```
2. start a server
3. summon that monster
4. do `/reload scripts` multiple times

### Expected behaviour
lua error?

### Actual behaviour
server crash
reproduction rate is roughly 50% for reload and roughly 15-20% for startup
exact line that leads to crashing:
https://github.com/otland/forgottenserver/blob/a8e5007a4c577a1c5f7e9e47f04e5b3a2e6e5a06/src/actions.cpp#L35

### Environment
master branch at c506a2f (most recent commit)
no idea how long this was bugged, might be related to https://github.com/otland/forgottenserver/issues/3692 due to the fact that both are crashing in actions interface

### PR DESCRIPTION
**Pull request by Zbizu**
closes #4102

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -620,25 +620,9 @@ bool Monsters::deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std
 			combat->setArea(area);
 		}
 
-		if (spell->conditionType != CONDITION_NONE) {
-			ConditionType_t conditionType = spell->conditionType;
-
-			uint32_t tickInterval = 2000;
-			if (spell->tickInterval != 0) {
-				tickInterval = spell->tickInterval;
-			}
-
-			int32_t conMinDamage = std::abs(spell->conditionMinDamage);
-			int32_t conMaxDamage = std::abs(spell->conditionMaxDamage);
-			int32_t startDamage = std::abs(spell->conditionStartDamage);
-
-			Condition* condition =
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by MillhioreBT** [comment]
Some conditions already named like `drunk` are already treated in a particular way, but if you also add the type of condition it will try to create another condition that will be added to the list, when in fact we only want one...

image

so the problem can be easily solved from lua, since apparently things can be done that should not be done

**Review by DSpeichert — APPROVED** [review]

**Comment by DSpeichert** [comment]
Formatting does need fixing though.

File: src/monsters.cpp [hunk 1]
```diff
@@ -765,6 +753,23 @@ bool Monsters::deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std
 			          << " - Unknown spell name: " << spell->name << std::endl;
 		}
 
+		if (!conditionDeclared && spell->conditionType != CONDITION_NONE) {
```
nekiro, line 756: Isnt it enough to move this as last else if?
Then just remove that condition declared

PR #4108, opened 2022-04-30
Add XML parsing bindings for Lua

ranisalt:
### Pull Request Prelude

### Changes Proposed

Adds XML parsing functions to Lua script interface, allowing us to navigate XML files without having to have that functionality inside the engine. I have implemented the smallest subset of features that was enough to parse any of the XML files we currently use for configuration. The intent is to have a migration path from XML without dropping any functionality immediately, and still benefitting from the cleaner engine.

### DIFF
--- a/data/scripts/lib/xml.lua
+++ b/data/scripts/lib/xml.lua
@@ -1,0 +1,18 @@
+local boolTrue = {"1", "y", "Y", "t", "T"}
+
+function tobool(str)
+    return str and table.contains(boolTrue, str[0])
+end
+
+function XMLNode.children(self)
+    local node = self:firstChild()
+    return function()
+        if not node then
+            return nil
+        end
+
+        local prevNode = node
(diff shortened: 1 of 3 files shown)


EPuncker (approved):
I've tested it, and its awesome!

data/scripts/lib/xml.lua
```diff
@@ -0,0 +1,18 @@
...
+    return function()
+        if not node then
+            return nil
+        end
+
+        local prevNode = node
+        node = node:nextSibling()
+        return prevNode
+    end
+end
```
ranisalt, line 18: In case anyone wonders, this implements the iterator protocol (which is super weird in Lua, but w/e) so one can use:
```lua
for child in node:children() do
  ...
end
```

LukSrT (approved):

DSpeichert (approved):

PR #4115, opened 2022-05-07
cipbia conditions by default

soul4soul:
### Pull Request Prelude

### Issues Addressed

I was playing with conditions in TFS to learn more about how they work in order to improve my monster converter. Overall the built in monster condition systems seem to behave in unexpected ways and isn't very useful in the current behavior.

 - Monster spells condition system in xml/revscriptsys is lackluster and likely misleading to what users would expect. All conditions defined with the monster system make use of the default Log formula. 
   - There are 89 uses of condition spells of which 66 are position condition (NOTE: I'm not counting conditions added on melee damage)
   - The other 23 uses of condition are using the wrong DOT formula and based on the mistakes in the monster attacks for these conditions its clear the authors of the monster files were confused. ie;
     - koshei the deathless, cursecondition has damage set to -54. This is about the max damage per turn that the condition deals on cipbia. It's wrong because min/max damage should be set to the total damage dealt by the attack and use expo damage instead of log
     -  Inky, energycondition has damage set to 75. In cipbia Inky causes energy condition with 3 turns at 25damage per turn which is expected for energy condition but not what the TFS engine does. Clearly user expectations don't match.
     - None of the 3 uses of drowncondition have a damage specified
   - conditions added on melee default the tick time to 2s for all condition types unless overridden but conditions caused by spells used the tick time as defined in cipbia.
- The lua `Creature:addDamageCondition` isn't any better
  - 30 conditions are defined using LUA scripts
  - 8 curse conditions that are defined using LUA all which over damage.
  - In the lua system the DAMAGELIST_EXPONENTIAL_DAMAGE way over damages the target. This effects all curse spell scripts. For example setting the damage range to 1000,1000 causes over 4k damage to the target. The formula in cip is to increase damage by 20% per turn and a random 1-120% increase for the last turn.
  - In the lua system the DAMAGELIST_LOGARITHMIC_DAMAGE way over damages the target. Hardcoding a damage of 400 causes over 7k damage. I'm not sure why this even exists when there is a working log formula built into the engine.	

### Changes made in this PR to address the issues
- Updated logic for generating conditions in monsters.cpp to be exactly the same for all ways conditions could be added to attacks (xml monster melee and spells, lua revscriptsys monster melee and spells, and lua spells). Note its still possible for dev's to write their own custom conditions and those will remain unaffected by this PR.
- updated condition attacks in monster spells to work the same as cipbia by default
- Added a condition damage formula for a constant DOT
- Added a condition damage formula for a exponential DOT at rate of 20% (curse condition)
- Moved all 30~ condition attacks from lua scripts to the xml monster definition file
- Improved Charged Energy Elemental, Dracola, Dwarf Henchmen, Tremorak, Pythius the rotten, and melee poison damage of many other creatures
- Update formulas for player DOT spells so they calculate the total damage instead of the "max hit" to be consistent with the way spells are defined.

### Issues not addressed
 - Conditions aren't ticking on monsters when players are off screen. When I'm on my god character and attack a monster with a condition spell (`###1`), the condition doesn't tick until I summon a monster with utevo res. Removing the monster stop the condition tick, summoning the monster back will continue the tick.
 - The logic for replacing conditions doesn't match cipbia. For example, In cipbia curse damage doesn't restart at 1 when you get cursed again. Maybe some changes can be made in `ConditionDamage::updateCondition` to improve the behavior. (#2651 is related in nature but not exactly the same)


src/monsters.cpp
```diff
@@ -82,11 +82,103 @@ ConditionDamage* Monsters::getDamageCondition(ConditionType_t conditionType, int
...
-	condition->setParam(CONDITION_PARAM_STARTVALUE, startDamage);
 	condition->setParam(CONDITION_PARAM_DELAYED, 1);
+
+	uint32_t defaultTickInterval = 2000;
+	uint32_t defaultTickInterval2 = 0;
+	// -1 = skip invalid condition
+	// 0 = constant
+	// 1 = log
+	// 2 = 20% growth
+	int32_t damageType = -1;
```
ramon-bernardo, line 93: enum instead of var?

src/monsters.cpp
```diff
@@ -82,11 +82,72 @@ ConditionDamage* Monsters::getDamageCondition(ConditionType_t conditionType, int
...
+		damagePerRound = startDamage;
+	}
+
+	// Maybe better to just check if condition is dazzled? It would make more sense if we need special block for bleeding too
+	if ((damageType == CONDITION_DAMAGE_FORMULA_CONSTANT) && (defaultTickInterval2 != 0)) {
+		rounds = normal_random(minDamage, maxDamage) / damagePerRound;
+		for (int32_t i = 0; i <= rounds; i++) {
+			int32_t randTickInterval = uniform_random(tickInterval, defaultTickInterval2);
+			condition->addDamage(1, randTickInterval, damagePerRound * -1);
+		}
```
ramon-bernardo, line 141: Execute addDamage with rounds=0? HM

```suggestion
		for (uint32_t i = 0; i <= rounds; i++) {
			int32_t randTickInterval = uniform_random(tickInterval, defaultTickInterval2);
			condition->addDamage(1, randTickInterval, damagePerRound * -1);
		}
```

or
```suggestion
		do {
			int32_t randTickInterval = uniform_random(tickInterval, defaultTickInterval2);
			condition->addDamage(1, randTickInterval, damagePerRound * -1);
		} while (rounds-- != 0);
```

src/monsters.cpp
```diff
@@ -82,11 +82,72 @@ ConditionDamage* Monsters::getDamageCondition(ConditionType_t conditionType, int
...
+
+	uint32_t defaultTickInterval = 2000;
+	uint32_t defaultTickInterval2 = 0;
+	ConditionDamageFormula_t damageType = CONDITION_DAMAGE_FORMULA_CONSTANT;
+	int32_t rounds = 0;
+	int32_t damagePerRound = 0;
+
+	if (conditionType == CONDITION_FIRE) {
+		defaultTickInterval = 9000;
+		damageType = CONDITION_DAMAGE_FORMULA_CONSTANT;
```
ramon-bernardo, line 95: always CONDITION_DAMAGE_FORMULA_CONSTANT

ramon-bernardo (commented):
Some suggestions 😃

src/monsters.cpp
soul4soul: yeah, I wanted to be explicit for each condition to make it easier to read but I can remove them if its not prefer.

src/monsters.cpp
```diff
@@ -82,11 +82,72 @@ ConditionDamage* Monsters::getDamageCondition(ConditionType_t conditionType, int
...
+		defaultTickInterval = 10000;
+		defaultTickInterval2 = 12000;
+		damageType = CONDITION_DAMAGE_FORMULA_CONSTANT;
+		damagePerRound = 20;
+	} else if (conditionType == CONDITION_BLEEDING) {
+		defaultTickInterval = 4000;
+		damageType = CONDITION_DAMAGE_FORMULA_CONSTANT;
+		// Not clear on how this works, it might be random damage per turn between 1 and 19?
+		damagePerRound = 15;
+	}
```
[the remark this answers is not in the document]
ramon-bernardo: Sorry, the space comes after the key

data/monster/monsters/dracola.xml
```diff
@@ -43,9 +43,6 @@
 		<attack name="lifedrain" interval="3000" chance="20" length="8" spread="3" min="-300" max="-600">
 			<attribute key="areaEffect" value="greenshimmer" />
 		</attack>
-		<attack name="drowncondition" interval="1000" chance="20" length="8" spread="3">
-			<attribute key="areaEffect" value="poff" />
-		</attack>
```
ramon-bernardo, line 48: Why?

src/monsters.cpp
```diff
@@ -82,11 +82,71 @@ ConditionDamage* Monsters::getDamageCondition(ConditionType_t conditionType, int
...
+		damagePerRound = startDamage;
+	}
+
+	// Maybe better to just check if condition is dazzled? It would make more sense if we need special block for bleeding too
+	if (defaultTickInterval2 > 0 && damageType == CONDITION_DAMAGE_FORMULA_CONSTANT) {
+		rounds = normal_random(minDamage, maxDamage) / damagePerRound;
+		do {
+			int32_t randTickInterval = uniform_random(tickInterval, defaultTickInterval2);
+			condition->addDamage(1, randTickInterval, damagePerRound * -1);
+		} while (rounds-- != 0);
```
ramon-bernardo, line 141: * suggestion
```
while (rounds-- != 0)
```
it's confusing, you can opt out, anyway

data/monster/monsters/dracola.xml
ramon-bernardo, line 48: cip removed this?

ramon-bernardo (commented):
Some suggestions

data/spells/spells.xml
```diff
@@ -768,33 +768,6 @@
...
-	<instant name="war golem electrify" words="###22" aggressive="1" blockwalls="1" needtarget="1" needlearn="1" script="monster/war_golem_electrify.lua" />
-	<instant name="souleater drown" words="###23" aggressive="1" blockwalls="1" needtarget="1" needlearn="1" script="monster/souleater_drown.lua" />
-	<instant name="mutated bat curse" words="###24" direction="1" aggressive="1" blockwalls="1" needtarget="0" needlearn="1" script="monster/mutated_bat_curse.lua" />
-	<instant name="vulcongra soulfire" words="###25" aggressive="1" blockwalls="1" needtarget="0" needlearn="1" script="monster/vulcongra_soulfire.lua" />
-	<instant name="lava golem soulfire" words="###26" aggressive="1" blockwalls="1" needtarget="0" needlearn="1" script="monster/lava_golem_soulfire.lua" />
-	<instant name="magma crawler soulfire" words="###27" aggressive="1" blockwalls="1" needtarget="0" needlearn="1" script="monster/magma_crawler_soulfire.lua" />
 	<instant name="war golem skill reducer" words="###28" direction="1" aggressive="1" blockwalls="1" needtarget="0" needlearn="1" script="monster/war_golem_skill_reducer.lua" />
```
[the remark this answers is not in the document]
soul4soul: No problem with renumbering, I wasn't sure how people would feel about it.

data/monster/monsters/dracola.xml
soul4soul: I couldn't find any information about this attack for this monster and it wasn't defined with any damage so the spell had no effect. If you know this monster causes drown condition and the amount of total damage I can add it back.

src/condition.cpp
```diff
@@ -1494,23 +1500,56 @@ uint32_t ConditionDamage::getIcons() const
 	return icons;
 }
 
-void ConditionDamage::generateDamageList(int32_t amount, int32_t start, std::list<int32_t>& list)
+void ConditionDamage::generateDamageList(int32_t amount, int32_t start, std::list<int32_t>& list, ConditionDamageFormula_t damageFormula)
 {
-	amount = std::abs(amount);
-	int32_t sum = 0;
-	double x1, x2;
+	if (damageFormula == CONDITION_DAMAGE_FORMULA_CONSTANT) {
```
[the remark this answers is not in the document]
soul4soul: I can't tell what you are recommending here but I moved `amount = std::abs(amount);` to the first line of the function.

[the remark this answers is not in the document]
ramon-bernardo:
```suggestion
	switch (damageFormula) {
	    case CONDITION_DAMAGE_FORMULA_CONSTANT: break;
	    case CONDITION_DAMAGE_FORMULA_LOG: break;
	    case CONDITION_DAMAGE_FORMULA_GROWTH: break;
	    default: break;
	}
```

src/monsters.cpp
```diff
@@ -82,11 +82,74 @@ ConditionDamage* Monsters::getDamageCondition(ConditionType_t conditionType, int
...
-	condition->setParam(CONDITION_PARAM_MAXVALUE, maxDamage);
-	condition->setParam(CONDITION_PARAM_STARTVALUE, startDamage);
 	condition->setParam(CONDITION_PARAM_DELAYED, 1);
+
+	uint32_t defaultTickInterval = 2000;
+	uint32_t defaultTickInterval2 = 0;
+	ConditionDamageFormula_t damageType = CONDITION_DAMAGE_FORMULA_CONSTANT;
+	int32_t rounds = 0, damagePerRound = 0;
+
+	if (conditionType == CONDITION_FIRE) {
```
ramon-bernardo, line 92: same here
```suggestion
	switch (conditionType) {
	    case CONDITION_FIRE: break;
	    case CONDITION_POISON: break;
	    case CONDITION_ENERGY: break;
	    case CONDITION_DROWN: break;
	    case CONDITION_FREEZING: break;
	    case CONDITION_DAZZLED: break;
	    case CONDITION_BLEEDING: break;
	    default: break;
	}
```

raymondtfr:
I added the "creature:addDamageCondition" code in Lua because the idea was to move more abd more things to Lua, right? Has it changed? (Ive been away for a while now, so no idea)

I researched hiw these conditions worked in the Testserver when these new conditions were (spells) were added and I added a formula for it (exponential damage for curse conditions). I am pretty sure they work as intended.

The firmula for log damage (for poison and bleeding conditions) was published by a user of tibiawikia in a discussion we were having regarding formulas for spells, armor values research, etc.

The reason for adding the log formula was because of keeping more stuff in Lua and this one is easier to read and more accurate.

The info you can read in tibiawikia about utori spells were researched by me at a time when one of tibiawikia's admins would ask Cip to grant me access to the Testservers.

Regarding the "Envenom (utori pox)" spell you mentioned it "does not take player level into account". I confirmed it does take player level into account (as all other utori spells do) but very minor. Whichever formula Cip developers use for this part only does a very small diference. The most approx. case I could find was "(level / 80)", after numerous researches, data gathering.

PR #4118, opened 2022-05-08
Update compat.lua

### LINKED ISSUE #4109: Add table.maxn to compat.lua
Just a reminder to add it to compat.lua as it was deprecated in lua 5.2

something like `table.maxn = table.maxn or function(t) return #t end` should do the trick

and yes, it is used in several datapacks npcs

EPuncker:
### Pull Request Prelude

### Changes Proposed
Added two functions to compat.lua that were deprecated or had its name changed in Lua 5.2

**Issues addressed:** 
closes #4109 

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -1546,3 +1546,15 @@ end
 
 -- this is a fix for lua52 or higher which has the function renamed to table.unpack, while luajit still uses unpack
 if not unpack then unpack = table.unpack end
+
+if not loadstring then loadstring = load end
+
+function table.maxn(t)
+	local max = 0
+	for k in pairs(t) do
+		if type(k) == "number" and k > max then
+			max = k
+		end
+	end
+	return max
+end


Zbizu (approved):
works, demo to test:

https://www.lua.org/cgi-bin/demo

```lua
t = {
[1] = 30,
["abc"] = 70,
[14] = 100,
["def"] = 10000
}

function table.maxn(t)
	local max = 0
	for k in pairs(t) do
		if type(k) == "number" and k > max then
			max = k
		end
	end
	return max
end

print(table.maxn(t))
```

MillhioreBT (approved):

ranisalt:
Doesn't Lua publish some global compatibility code? Since they make so many breaking changes from one version to the other, I feel like the language developers should provide that path

EPuncker:
> Doesn't Lua publish some global compatibility code? Since they make so many breaking changes from one version to the other, I feel like the language developers should provide that path

there is this: http://lua-users.org/wiki/LuaVersionCompatibility but I don't know

DSpeichert (approved):

omarcopires:
Works fine for me, could someone merge it?

PR #4123, opened 2022-05-10
Fix: Prey window

omarcopires:
### Pull Request Prelude

### Changes Proposed 
The prey window should not be sent to players with no vocation in their actual behavior. With this change, the system correctly sends the system, in cases of customized servers it is at the discretion of the administrator.


Zbizu:
doesn't matter, prey system is not even implemented yet

Zbizu:
> There is also no reason to send the wrong window

it's not even window, it's a byte in context menu
0x00 will give you "prey system disabled" kind of response, 0x01 will give you error because prey metadata is missing

nekiro:
I'm against it, redundant change and somewhat indicates that we have something with prey going on, while we dont

Issue #3871, opened 2021-12-21
List of missing TFS features from all versions


omarcopires:
**Version 8.72**
- [ ] Implemented the You cannot use objects that fast. message if the player use an object more than 2 times per second.

**Version 9.05**
- [ ] All players now gain 10 second invisibility to creatures upon login, but as soon as the player has moved or attacked, creatures will again see the character.
- [ ] In the case of disconnection from server, players now have a greater chance of survival:
- Characters are now ignored by creatures for 30 seconds upon the time the server recognizes the client has lost connection. This leaves less time a character is vulnerable: 30 seconds (knight), 20 seconds (paladin) or 10 seconds (druid/sorcerer). Note that this is invisibility, not invincibility - area attacks will affect the player still.

**Version 9.08**
- [ ] After dying and re-logging your server log is kept. #3535

**Version 10.20**
- [ ] Reward system for bosses (https://otland.net/threads/reward-chest-boss-reward-tfs-1-2.233397/).

**Version 10.55**
- [ ] The Adventurer's Blessing was introduced (https://tibia.fandom.com/wiki/Adventurer%27s_Blessing).

**Version 10.70**
- [ ] Guild Bank. #2214
- [ ] Delayed spawn feature (partially added, still missing the magic effect display)

**Version 10.74**
- [ ] The Loyalty System was introduced.

**Version 10.80**
- [ ] The Store is introduced along with its currency Tibia Coins.

**Version 10.97**
- [ ] Gold Pouch (https://tibia.fandom.com/wiki/Gold_Pouch).

**Version 11.02**
- [ ] Introduction of the Prey System, which will allow players to obtain a random hunting bonus that only applies while hunting a particular monster. This bonus can be a boost of the damage dealt, a reduction of the damage taken, a boost of experience or a boost in loot.
- [ ] Introduction of a new enchanting system called Imbuing, i.e the ability to augment certain equipment with elemental damage, elemental protection, skill boosts (including speed), critical hit, hit points leech and mana leech. There'll be 63 different imbuements.

**Version 11.10**
- [ ] VIP groups: VIP List has now the following 3 default groups: Friends, Enemies and Trading Partners. Additionally, Premium Account players can create up to 5 other groups.

**Version 11.30**
[...]

PR #4148, opened 2022-05-29
Added Loyalty System (10.74)

reyaleman:
### Pull Request Prelude

### Changes Proposed
* Implemented [LOYALTY SYSTEM](https://tibia.fandom.com/wiki/Loyalty_System)
* System is optional, can be disabled in config.lua (loyaltyEnabled = false)
* Loyalty bonuses are customizables in config.lua
* Modified Player::GetPercentLevel to return floats (client support decimal numbers on skill bars)
* Loyalty Titles onLook
* Loyalty message bonus on serverlog (Player:onLogin script)
* Kept skill & tries untouched, loyalty is calculated when the player is loaded.
* Loyalty points are updated on startup but only if it's been more than 1 day since last update (server config)
* Added 3 lua functions (Player:getLoyaltyStatus, Player:addLoyalty & Player:removeLoyalty)
* Added migration file to update database (column "loyalty_points" added to accounts table and inserted server config "loyalty_update" value to update loyalty on a daily basis)

#### Missing:
* Social loyalty badges (social not implemented yet)
* Loyalty highscores (highscores not implemented yet)

loyalty

**Issues addressed:** 
#3871 Version 10.74 (The Loyalty System was introduced.)

Any comments would be highly appreciated. 😃 

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -135,6 +135,22 @@ experienceStages = {
 	{ minlevel = 101, multiplier = 3 }
 }
 
+-- Loyalty system
+loyaltyEnabled = true
+loyaltyBonuses = {
+	{ min = 0, max = 360, bonus = 0 },
+	{ min = 360, max = 720, bonus = 0.05 },
+	{ min = 720, max = 1080, bonus = 0.1 },
+	{ min = 1080, max = 1440, bonus = 0.15 },
+	{ min = 1440, max = 1800, bonus = 0.2 },
+	{ min = 1800, max = 2160, bonus = 0.25 },
+	{ min = 2160, max = 2520, bonus = 0.3 },
+	{ min = 2520, max = 2880, bonus = 0.35 },
+	{ min = 2880, max = 3240, bonus = 0.4 },
+	{ min = 3240, max = 3600, bonus = 0.45 },
+	{ min = 3600, bonus = 0.5 }
+}
+
 -- Rates
 -- NOTE: rateExp is not used if you have enabled stages above
 rateExp = 5
--- a/data/globalevents/scripts/startup.lua
+++ b/data/globalevents/scripts/startup.lua
@@ -42,4 +42,18 @@ function onStartup()
 		local position = town:getTemplePosition()
 		db.query("INSERT INTO `towns` (`id`, `name`, `posx`, `posy`, `posz`) VALUES (" .. town:getId() .. ", " .. db.escapeString(town:getName()) .. ", " .. position.x .. ", " .. position.y .. ", " .. position.z .. ")")
 	end
+	
+	--  loyalty system (update points)
+	if configManager.getBoolean(configKeys.LOYALTY_SYSTEM) then
+		local updatedAt = 0
+		local resultId = db.storeQuery("SELECT `value` FROM `server_config` WHERE `config` = 'loyalty_updated'")
+		if resultId then
+			updatedAt = result.getNumber(resultId, "value")
+			result.free(resultId)
+		end	
+		if (os.time() - updatedAt) >= 86400 then
+			db.asyncQuery("UPDATE `accounts` SET `loyalty_points` = `loyalty_points` + 1 WHERE `premium_ends_at` > " .. os.time())
+			db.query("UPDATE `server_config` SET `value` = " .. os.time() .. " WHERE `config` = 'loyalty_updated'")
+		end
+	end
 end
--- a/schema.sql
+++ b/schema.sql
@@ -5,6 +5,7 @@ CREATE TABLE IF NOT EXISTS `accounts` (
   `secret` char(16) DEFAULT NULL,
   `type` int NOT NULL DEFAULT '1',
   `premium_ends_at` int unsigned NOT NULL DEFAULT '0',
+  `loyalty_points` smallint unsigned NOT NULL DEFAULT '0',
   `email` varchar(255) NOT NULL DEFAULT '',
   `creation` int NOT NULL DEFAULT '0',
   PRIMARY KEY (`id`),
@@ -362,7 +363,7 @@ CREATE TABLE IF NOT EXISTS `towns` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
-INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '32'), ('players_record', '0');
+INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '32'), ('players_record', '0'), ('loyalty_updated', '0');
 
 DROP TRIGGER IF EXISTS `ondelete_players`;
 DROP TRIGGER IF EXISTS `oncreate_guilds`;
--- a/schema.sql
+++ b/schema.sql
@@ -363,7 +363,7 @@ CREATE TABLE IF NOT EXISTS `towns` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
-INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '32'), ('players_record', '0'), ('loyalty_updated', '0');
+INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '33'), ('players_record', '0'), ('loyalty_updated', '0');
 
 DROP TRIGGER IF EXISTS `ondelete_players`;
 DROP TRIGGER IF EXISTS `oncreate_guilds`;
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -78,6 +78,32 @@ ConfigManager::ConfigManager() { string[CONFIG_FILE] = "config.lua"; }
 
 namespace {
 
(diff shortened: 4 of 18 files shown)


schema.sql
```diff
@@ -362,7 +363,7 @@ CREATE TABLE IF NOT EXISTS `towns` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
-INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '32'), ('players_record', '0');
+INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '32'), ('players_record', '0'), ('loyalty_updated', '0');
```
DSpeichert, line 366:
```suggestion
INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '33'), ('players_record', '0'), ('loyalty_updated', '0');
```

I think you forgot to bump the schema version here.

DSpeichert (changes requested):

schema.sql
reyaleman: thanks! I'll fix it in the next commit, I'm still waiting more reviews

src/player.cpp
```diff
@@ -4835,3 +4960,39 @@ void Player::updateRegeneration()
...
+	}
+
+	loyaltyPoints += points;
+	return true;
+}
+
+std::string Player::getLoyaltyTitleDescription() const
+{
+	if (loyaltyPoints <= 50) {
+		return "a Scout of Tibia";
```
omarcopires, line 4977: maybe we should use world name or citizen instead of tibia

reyaleman: okay, I'll use ConfigManager::SERVER_NAME

src/player.cpp
```diff
@@ -564,6 +579,65 @@ void Player::removeSkillTries(skills_t skill, uint64_t count, bool notify /* = f
...
+}
+
+void Player::updateLoyaltySkill(uint8_t skill)
+{
+	if (skill > SKILL_LAST) {
+		return;
+	}
+
+	skills[skill].bLevel = skills[skill].level;
+	skills[skill].bTries = skills[skill].tries;
```
nekiro, line 632: is bLevel supposed to be baseLevel?
If so what do you think about just naming it baseLevel, baseTries and so on

src/player.cpp
```diff
@@ -1778,6 +1853,57 @@ void Player::removeManaSpent(uint64_t amount, bool notify /* = false*/)
...
+		currReqManaSpent = nextReqManaSpent;
+		nextReqManaSpent = vocation->getReqMana(bMagLevel + 1);
+		if (currReqManaSpent >= nextReqManaSpent) {
+			count = 0;
+			break;
+		}
+	}
+
+	magLevelPercent = Player::getPercentLevel(count, nextReqManaSpent);
+	bManaSpent += count;
```
nekiro, line 1880: same here, baseManaSpent?

src/player.cpp
```diff
@@ -4835,3 +4961,39 @@ void Player::updateRegeneration()
...
+{
+	if (loyaltyPoints == 0) {
+		return false;
+	}
+
+	loyaltyPoints += points;
+	return true;
+}
+
+std::string Player::getLoyaltyTitleDescription() const
```
nekiro, line 4975: this should be a "tool" method, so place this in tools.cpp and pass loyaltyPoints, you can also put everything into a map and just iterate it to find proper title (code will be shorter and easier to edit)

src/player.cpp
```diff
@@ -4835,3 +4961,39 @@ void Player::updateRegeneration()
 		condition->setParam(CONDITION_PARAM_MANATICKS, vocation->getManaGainTicks() * 1000);
 	}
 }
+
+bool Player::addLoyalty(int32_t points)
+{
+	if (loyaltyPoints == 0) {
```
nekiro, line 4967: I might not understand how this work, but why would you not allow adding loyalty points if loyaltyPoints is 0?

nekiro (changes requested):
Great stuff

src/player.cpp
nekiro: Not sure, but I think it doesnt matter, so just let it work even with loyalty system disabled

src/configmanager.cpp
```diff
@@ -334,6 +364,19 @@ float ConfigManager::getExperienceStage(uint32_t level) const
 	return std::get<2>(*it);
 }
 
+float ConfigManager::getLoyaltyBonus(uint16_t points) const
+{
+	auto it = std::find_if(loyaltyBonuses.begin(), loyaltyBonuses.end(), [points](LoyaltyBonuses::value_type bonus) {
```
nekiro, line 369: `auto it = std::find_if(loyaltyBonuses.begin(), loyaltyBonuses.end(), [points](const auto& bonus) {`?

src/luascript.cpp
```diff
@@ -10800,6 +10805,49 @@ int LuaScriptInterface::luaPlayerGetIdleTime(lua_State* L)
...
+{
+	// player:removeLoyalty([points = 1])
+	Player* player = getUserdata<Player>(L, 1);
+	if (!player) {
+		lua_pushnil(L);
+	}
+
+	uint16_t points = getNumber<uint16_t>(L, 2, 1);
+	pushBoolean(L, player->addLoyalty(-points));
+
```
nekiro, line 10847: remove newline

src/luascript.cpp
```diff
@@ -10800,6 +10805,49 @@ int LuaScriptInterface::luaPlayerGetIdleTime(lua_State* L)
...
+{
+	// player:addLoyalty([points = 1])
+	Player* player = getUserdata<Player>(L, 1);
+	if (!player) {
+		lua_pushnil(L);
+	}
+
+	uint16_t points = getNumber<uint16_t>(L, 2, 1);
+	pushBoolean(L, player->addLoyalty(points));
+
```
nekiro, line 10833: remove newline

src/player.cpp
```diff
@@ -146,6 +151,16 @@ std::string Player::getDescription(int32_t lookDistance) const
 		} else {
 			s << " has no vocation.";
 		}
+
+		if (g_config.getBoolean(ConfigManager::LOYALTY_SYSTEM) && loyaltyPoints >= 50) {
+			if (sex == PLAYERSEX_FEMALE) {
```
nekiro, line 156: use fmt::format?

src/player.cpp
```diff
@@ -126,6 +126,11 @@ std::string Player::getDescription(int32_t lookDistance) const
 		} else {
 			s << " You have no vocation.";
 		}
+
+		if (g_config.getBoolean(ConfigManager::LOYALTY_SYSTEM) && loyaltyPoints >= 50) {
+			s << " You are " << getLoyaltyTitleDescription(loyaltyPoints) << " "
```
nekiro, line 131: use fmt::format?

src/player.cpp
```diff
@@ -1778,6 +1870,72 @@ void Player::removeManaSpent(uint64_t amount, bool notify /* = false*/)
...
+	if (!isLogin) {
+		float bonusCount = count * (1 + loyaltyBonus);
+
+		float amount, fractional;
+		fractional = std::modf(bonusCount, &amount);
+		manaSpentFractional += std::round(fractional * 100) / 100;
+
+		count = static_cast<uint64_t>(amount);
+
+		if (manaSpentFractional == (int)manaSpentFractional) {
```
nekiro, line 1929: use static_cast and int32_t

omarcopires:
Please don't leave this!!

config.lua.dist
```diff
@@ -135,6 +135,22 @@ experienceStages = {
 	{ minlevel = 101, multiplier = 3 }
 }
 
+-- Loyalty system
+loyaltyEnabled = true
```
ranisalt, line 139: You can drop this flag and let users effectively disable the loyalty system by removing all tiers.

omarcopires:
could you leave some extra functions in LUA to be used in cyclopedia?
`getLoyaltyMagicLevel`, `getLoyaltySkill` and `getLoyaltyTitleDescription`

I also think it would be better to move the descriptions to onLook instead of keeping them directly in the engine

data/globalevents/scripts/startup.lua
```diff
@@ -42,4 +42,18 @@ function onStartup()
 		local position = town:getTemplePosition()
 		db.query("INSERT INTO `towns` (`id`, `name`, `posx`, `posy`, `posz`) VALUES (" .. town:getId() .. ", " .. db.escapeString(town:getName()) .. ", " .. position.x .. ", " .. position.y .. ", " .. position.z .. ")")
 	end
+	
+	--  loyalty system (update points)
+	if configManager.getBoolean(configKeys.LOYALTY_SYSTEM) then
+		local updatedAt = 0
+		local resultId = db.storeQuery("SELECT `value` FROM `server_config` WHERE `config` = 'loyalty_updated'")
```
omarcopires, line 49:
```lua
local resultId = ...
if not resultId then
  return {}
end
```

reyaleman:
I'd have to rewrite this and make a new pull request (taking your comments in consideration), IIRC I stopped this because I was having issues with float values being inexact 👀

PR #4150, opened 2022-06-02
Compute combatTypeToIndex mathematically

nekiro:
Instead of hardcoding conditionals, we can just use logarithm to reverse bit shiffting (multiplying) effectively making this method work properly for new custom types.

### DIFF
--- a/src/tools.cpp
+++ b/src/tools.cpp
@@ -926,34 +926,7 @@ bool booleanString(const std::string& str)
 
 size_t combatTypeToIndex(CombatType_t combatType)
 {
-	switch (combatType) {
-		case COMBAT_PHYSICALDAMAGE:
-			return 0;
-		case COMBAT_ENERGYDAMAGE:
-			return 1;
-		case COMBAT_EARTHDAMAGE:
-			return 2;
-		case COMBAT_FIREDAMAGE:
-			return 3;
-		case COMBAT_UNDEFINEDDAMAGE:
-			return 4;
-		case COMBAT_LIFEDRAIN:
-			return 5;
-		case COMBAT_MANADRAIN:
-			return 6;
-		case COMBAT_HEALING:
-			return 7;
-		case COMBAT_DROWNDAMAGE:
-			return 8;
-		case COMBAT_ICEDAMAGE:
-			return 9;
-		case COMBAT_HOLYDAMAGE:
-			return 10;
-		case COMBAT_DEATHDAMAGE:
-			return 11;
-		default:
-			return 0;
-	}
+	return static_cast<size_t>(std::log2(static_cast<uint64_t>(combatType)));
 }
 
 CombatType_t indexToCombatType(size_t v) { return static_cast<CombatType_t>(1 << v); }


EPuncker (approved):

kygov:
I don't know about this case, but isn't std:log2 very slow?

nekiro:
> I don't know about this case, but isn't std:log2 very slow? And then it's used in TFS in a lot of places, every monster/player attack etc. Probably that's why someone left simple switch instead.

Can't find any relevant information about it being slow nor any performance comparisons, so I guess someone gotta benchmark this first.

ranisalt:
> Did a benchmark (hopefully the right way)

You forgot to deoptimize the return value

https://quick-bench.com/q/saYIiPMo23q71PzltXyhCBd_pok

If you see the disassemble, both the switch and the log2 implementation generate the very same value, because the compiler is able to infer a jump table from the limited input values.

ranisalt (approved):

amatria:
Isn't the benchmark biased? Shouldn't you test the three functions with random input values? If the compiler is generating a lookup table for the switch statement, you are benchmarking the speed of the log2 function against a 100% cache hit and 0% branch misprediction rate lookup table, essentially.

Also, if the compiler is generating a lookup table for the math-only implementation too, I would still go for the switch-case function. I see no benefit in reducing readability if the two versions are going to produce the same assembly.

Anastaciaa:
> Isn't the benchmark biased? Shouldn't you test the three functions with random input values? If the compiler is generating a lookup table for the switch statement, you are benchmarking the speed of the log2 function against a 100% cache hit rate and 0% branch misprediction rate lookup table, essentially.
> 
> Also, if the compiler is generating a lookup table for the math-only implementation too, I would still go for the switch-case function. I see no benefit in reducing readability if the two versions are going to produce the same assembly.

You may be right. https://quick-bench.com/q/8vEZOgx0v6RQ1hT_DzBVZ5Aa-Z8

kygov:
> Isn't the benchmark biased? Shouldn't you test the three functions with random input values? If the compiler is generating a lookup table for the switch statement, you are benchmarking the speed of the log2 function against a 100% cache hit rate and 0% branch misprediction rate lookup table, essentially.
> 
> Also, if the compiler is generating a lookup table for the math-only implementation too, I would still go for the switch-case function. I see no benefit in reducing readability if the two versions are going to produce the same assembly.

You may be right. https://quick-bench.com/q/8vEZOgx0v6RQ1hT_DzBVZ5Aa-Z8

ranisalt:
> Isn't the benchmark biased? Shouldn't you test the three functions with random input values?

That's correct, good catch. It was inlining the log2, and since it is constexpr and always using the same input, it was replaced by the result.

> You may be right - https://quick-bench.com/q/8vEZOgx0v6RQ1hT_DzBVZ5Aa-Z8

Getting a random number is dominating the time - it is super slow compared to the meaningful code itself. If you move `CombatType_t type = static_cast<CombatType_t>(rand() % COMBAT_COUNT);` outside the loop - it will still be a random value, and avoid the aforementioned inlining "bug" - the switch is still an order of magnitude faster.

nekiro:
Alright, closing as slow then, very sad, because this would make one place less to edit to add new combat types.

Issue #4161, opened 2022-06-12
Questions about monster spell range


soul4soul:
If I understand correctly a monster can only see as far as the view port
https://github.com/otland/forgottenserver/blob/master/src/monster.cpp#L98-L101
(There appears to be a typo here too, I believe the last param should be `maxClientViewportY + 1`)

But monster spell range is double view port, should this be changed to `max(maxClientViewportX, maxClientViewportY ) + 1`?
https://github.com/otland/forgottenserver/blame/master/src/monsters.cpp#L127-L133
https://github.com/otland/forgottenserver/blob/master/src/monsters.cpp#L563-L566
It would be nice to add a warning to this condition block, to help maintain clean monster packs, `ghazbaran` has an attack with range 14, which is the only attack I see exceeding the range. Tangent information, In the current monster pack only 5 *attacks* have a range >=8 , while over 800 attacks have a range of exactly 7.

Range seems to work a bit differently for canUseSpell and canUseAttack. These code locations
https://github.com/otland/forgottenserver/blob/master/src/monster.cpp#L846-L849
https://github.com/otland/forgottenserver/blob/master/src/monster.cpp#L879-L880
For canUseSpell 0 means don't check range
For canUseAttack 0 skips the spell?

I see a few options which could be mixed and matched together.
- option A) Add a warning for attack spells with a range of 0
- option B) Don't make range of 0 invalid, although these spells would only be cast when the player is on the creature
- option C) Default the range to some valid value (maybe 1 or viewport)

My preference would be to default the attack range to viewport (it feels more flexible and would be less work to change if someone wanted a different viewport size), make 0 a valid range for canUseSpell and canUseAttack, and show a warning if range > viewport. Given those changes, I would have to do some testing in cipbia but conceivably many attacks in the data pack with range 7 and above could have the range removed and default to the viewport.

EPuncker:
I vote for default to range 1 plus a warning

ramon-bernardo:
> https://github.com/otland/forgottenserver/blob/master/src/monster.cpp#L98-L101
(There appears to be a typo here too, I believe the last param should be maxClientViewportY + 1)

If changed to Y, it will be easier to run away from monsters, vertically

PR #4162, opened 2022-06-12
Limits to spell range to prevent invalid configurations

soul4soul:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** #4161

- Fix monster canSee Range
- Limit spell range to ViewPort
- Add warning to spells which exceed range
- Allow spells to execute with range of 0
- Fixes to attacks with spell range >= 8

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -97,7 +97,7 @@ const std::string& Monster::getNameDescription() const
 
 bool Monster::canSee(const Position& pos) const
 {
-	return Creature::canSee(getPosition(), pos, Map::maxClientViewportX + 1, Map::maxClientViewportX + 1);
+	return Creature::canSee(getPosition(), pos, Map::maxClientViewportX + 1, Map::maxClientViewportY + 1);
 }
 
 bool Monster::canWalkOnFieldType(CombatType_t combatType) const
@@ -846,7 +846,7 @@ bool Monster::canUseAttack(const Position& pos, const Creature* target) const
 		uint32_t distance =
 		    std::max<uint32_t>(Position::getDistanceX(pos, targetPos), Position::getDistanceY(pos, targetPos));
 		for (const spellBlock_t& spellBlock : mType->info.attackSpells) {
-			if (spellBlock.range != 0 && distance <= spellBlock.range) {
+			if (distance <= spellBlock.range) {
 				return g_game.isSightClear(pos, targetPos, true);
 			}
 		}
@@ -876,8 +876,7 @@ bool Monster::canUseSpell(const Position& pos, const Position& targetPos, const 
 		}
 	}
 
-	if (sb.range != 0 &&
-	    std::max<uint32_t>(Position::getDistanceX(pos, targetPos), Position::getDistanceY(pos, targetPos)) > sb.range) {
+	if (std::max<uint32_t>(Position::getDistanceX(pos, targetPos), Position::getDistanceY(pos, targetPos)) > sb.range) {
 		inRange = false;
 		return false;
 	}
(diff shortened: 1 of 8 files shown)


ramon-bernardo (dismissed):

src/monsters.cpp
```diff
@@ -124,10 +124,13 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
...
 	}
 
+	sb.range = std::max<uint32_t>(Map::maxClientViewportX + 1, Map::maxClientViewportY + 1);
 	if ((attr = node.attribute("range"))) {
 		uint32_t range = pugi::cast<uint32_t>(attr.value());
-		if (range > (Map::maxViewportX * 2)) {
-			range = Map::maxViewportX * 2;
+		if (range > std::max<uint32_t>(Map::maxClientViewportX + 1, Map::maxClientViewportY + 1)) {
+			range = std::max<uint32_t>(Map::maxClientViewportX + 1, Map::maxClientViewportY + 1);
+			std::cout << "[Warning - Monsters::deserializeSpell] - " << description
```
ranisalt, line 132: IMO it should either:
1. Warn the user of the mistake, and allow it to proceed - if it does not crash the server
2. Forbid using invalid range values, and refuse to proceed (return false)
But not change the value to something the user explicitly set otherwise.

src/monster.cpp
```diff
@@ -97,7 +97,7 @@ const std::string& Monster::getNameDescription() const
 
 bool Monster::canSee(const Position& pos) const
 {
-	return Creature::canSee(getPosition(), pos, Map::maxClientViewportX + 1, Map::maxClientViewportX + 1);
+	return Creature::canSee(getPosition(), pos, Map::maxClientViewportX + 1, Map::maxClientViewportY + 1);
```
ramon-bernardo, line 100: For monsters, using `maxClientViewportX` seems like correct behavior. Unless Cip says otherwise.

src/monsters.cpp
soul4soul: Changing the value has no impact because its constrained to the creatures max view. I do it to keep the number clean and sanitized for wherever its used in the future. Also this is done in other parts of the code both warning + setting the value (IE when the chance is parsed). If we know how to fix the value on the users behalf we should do that and log a warning.

src/monster.cpp
soul4soul: Good point, I will revert this line and leave it as it alone to mirror cip behavior 🤔 we could use better variable names.

src/monsters.cpp
```diff
@@ -124,10 +124,13 @@ bool Monsters::deserializeSpell(const pugi::xml_node& node, spellBlock_t& sb, co
 		          << " - Missing chance value on non-melee spell: " << name << std::endl;
 	}
 
+	sb.range = std::max<uint32_t>(Map::maxClientViewportX + 1, Map::maxClientViewportY + 1);
 	if ((attr = node.attribute("range"))) {
 		uint32_t range = pugi::cast<uint32_t>(attr.value());
-		if (range > (Map::maxViewportX * 2)) {
-			range = Map::maxViewportX * 2;
+		if (range > std::max<uint32_t>(Map::maxClientViewportX + 1, Map::maxClientViewportY + 1)) {
+			range = std::max<uint32_t>(Map::maxClientViewportX + 1, Map::maxClientViewportY + 1);
```
ramon-bernardo, line 131: range = x + x

src/monster.cpp
ramon-bernardo: > For monsters, using `maxClientViewportX` seems like correct behavior. Unless Cip says otherwise.

Why? If Y axis is calculated, players running vertically can escape easily than horizontally.
Monster ex: can see X: 10, see Y: 8.

ramon-bernardo (approved):
LGTM! Great impr, no more spells with invalid range ;)

PR #4152, opened 2022-06-04
WIP: Highscores

reyaleman:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
* Added in-game highscores feature...
* Easy to add custom highscores like "Most deaths, online times, storage checks, etc..."
* configurable (highscores.lua)

Needs testing, suggestions & code review

**Issues addressed:** 
Highscores functionality

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4164, opened 2022-06-14
Client vocation id with const

omarcopires:
### Pull Request Prelude

### Changes Proposed 
The client uses ids of different vocations than what we have reserved, an example of this can be seen at:
https://github.com/otland/forgottenserver/pull/4152/files#diff-d2c7616fe001d599d15716741b30fb0a3dd3a251a45485ed7896eccd0268fd8dR8

#4152 can be updated to use consts instead of numbers.


soul4soul:
Unless you plan to add more here you aren't replacing anything. I think you are better off leaving a comment/code review to add constants to the high scores PR. I don't remember how work vocations in TFS and I know the default pack doesn't have a 3rd promotion but it could be useful to have a helpful function that could handle the 3rd+ promotion level and be smart enough to return the highest available client vocation promotion.

omarcopires:
> Unless you plan to add more here you aren't replacing anything. I think you are better off leaving a comment/code review to add constants to the high scores PR. I don't remember how work vocations in TFS and I know the default pack doesn't have a 3rd promotion but it could be useful to have a helpful function that could handle the 3rd+ promotion level and be smart enough to return the highest available client vocation promotion.

I believe the name is right now, however this should be added to the pull request I mentioned, replacing the vocation numbers with const.

Issue #4076, opened 2022-04-11
OnUse event


EPuncker:
### Explanation of what you want to do that is currently impossible
We have pretty much an event for almost everything, but not one for OnUse.

### Desired functionality
With that event we could script several useful stuff and block/unblock item usage, for example:
- block item from being used at certain conditions
- log item usage
- etc, sorry I'm out of ideas, but I'm pretty sure there are many

where it could be used:
https://otland.net/threads/log-onuse.280529/
#4000 

### Available workarounds
scripting EACH item manually.

PR #4167, opened 2022-06-18
add onUseItem event

amatria:
### Pull Request Prelude

### Changes Proposed

Add `onUseItem` event functionality. It allows to obtain item usage statistics or to block an item from being used under certain conditions.

```lua
function Player:onUseItem(item)
    if player:hasFlag(FLAG_CANT_USE_ITEMS) then
        return false
    end
    print("Player " .. self:getName() .. " used item with ID " .. item:getId())
    return true
end
```

**Issues addressed:**  #4076 

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -2164,6 +2164,10 @@ void Game::playerUseItemEx(uint32_t playerId, const Position& fromPos, uint8_t f
 	player->resetIdleTime();
 	player->setNextActionTask(nullptr);
 
+	if (!g_events->eventPlayerOnUseItem(player, item)) {
+		return;
+	}
+
 	g_actions->useItemEx(player, fromPos, toPos, toStackPos, item, isHotkey);
 }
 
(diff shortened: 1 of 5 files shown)


src/game.cpp
```diff
@@ -2164,6 +2164,10 @@ void Game::playerUseItemEx(uint32_t playerId, const Position& fromPos, uint8_t f
 	player->resetIdleTime();
 	player->setNextActionTask(nullptr);
 
+	if (!g_events->eventPlayerOnUseItem(player, item)) {
```
nekiro, line 2167: move this above
```cpp
player->resetIdleTime();
player->setNextActionTask(nullptr);
```

Erza:
I don't see the point to this. Use an action event if you want to handle item interaction in a specific way. It just seems like an unnecessary duplicate of functionality that is already there.

Zbizu:
Agreed. It's pretty much extra script execution every time you drink a potion or use a rune.

If you want every weapon to have use action by default, move registering weapons to lua and register onUse there.

amatria:
I get your points indeed, @Zbizu @Erza. So... If the general opinion is that we can simply do the same with action scripts, we can close this PR. But let's see what other `devs` think about it.

Thank you for your time 🙂

omarcopires:
Any progress with this? We already have all the events practically, it would be interesting to have the action too.

PR #4173, opened 2022-06-26
minor fixes to outfits.xml

BahamutxD:
Removed duplicate entry for 1415 "Forest Warden"
Set premium yes/no to all outfits following tibiawiki


EPuncker (changes requested):
re-add Nordic Chieftain and we good 😃

EPuncker (approved):

PR #4176, opened 2022-06-29
Possibility to cancel LookDescription by returning an empty string

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Allow cancellation of the description message by returning an empty string.

### DIFF
--- a/data/events/scripts/player.lua
+++ b/data/events/scripts/player.lua
@@ -10,7 +10,10 @@ function Player:onLook(thing, position, distance)
 	if EventCallback.onLook then
 		description = EventCallback.onLook(self, thing, position, distance, description)
 	end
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

MillhioreBT:
> What happens if you actually send an empty string, is that possible? I'm thinking maybe `nil` would be a better value for cancelling.

I also thought so although if we return `nil` we are not sure that a next event will not treat this parameter as a string and it would break backwards compatibility, certain scripts might throw errors that you are treating `nil` as a `string`
This is because the parameters are auto-updated on each iteration of each registered eventcallback.

ramon-bernardo (approved):

PR #4181, opened 2022-07-07
Avoid user timing attack

ramon-bernardo:
### Pull Request Prelude

### Optional - Changes Proposes
- SHA1 it's as fast as rocket, which makes this PR optional.

### For curious ?
- Auth response time varies depending on query and hash-gen.
- If you consult the bank, returning without not gen the password-hash, the time is different if the user exists.
Ex: if user not exists: return select time, if exists: return select time + password hash gen. With this, you can see if user exists in database.

### Future is Argon2 ?
- If the password hash is changed, test verification response time.

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -97,6 +97,8 @@ bool IOLoginData::loginserverAuthentication(const std::string& name, const std::
 uint32_t IOLoginData::gameworldAuthentication(const std::string& accountName, const std::string& password,
                                               std::string& characterName, std::string& token, uint32_t tokenTime)
 {
+	std::string passwordHash = transformToSHA1(password);
+
 	Database& db = Database::getInstance();
 	DBResult_ptr result = db.storeQuery(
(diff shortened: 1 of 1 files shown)


ranisalt:
That code path is not used anymore, since the login server does not support the client protocol.

Zbizu:
Mehah OTC allows legacy login protocol, useful when you don't want to bother with setting up the webservice just to test a few things

https://github.com/mehah/otclient/commit/d9c9a9d913b7bceaf3cee2de7a9c351d4ecd339c#diff-9069838e67c3d247571f4c3e3d140550eb63b59bdb21eb26222ba79ba312b9f7R439

ramon-bernardo:
@ranisalt now that we have login via HTTP, can we consider implementing it?

ranisalt:
> @ranisalt now that we have login via HTTP, can we consider implementing it?

Sure, but I'm skeptical this improves anything. SHA1 is orders of magnitude faster than the network request so if anything the difference will be within statistical margin of error.

PR #4182, opened 2022-07-13
Quest via Lua and dynamic descriptions

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes add the ability to create Quest from Revscripts.
Support for dynamic descriptions via a callback has also been added.
Example of `data/scripts/quests/#dynamic.lua`
```lua
local questII = Game.createQuest("Example Quest II", 1002, 1)

questII:createMission("Example Mission 1", 1002, 1, 10, false, function (player)
    return string.format("Example description %d by %s.", player.storage[1002], player:getName())
end)

questII:register()
```
GIF 13-07-2022 06-19-33 a  m

**Notes:**
Suggestions to improve the code are welcome if possible <3

### DIFF
--- a/src/quests.cpp
+++ b/src/quests.cpp
@@ -8,8 +8,15 @@
 #include "player.h"
 #include "pugicast.h"
 
+uint32_t Quests::questAutoID = 0;
+uint32_t Quests::missionAutoID = 0;
+
 std::string Mission::getDescription(Player* player) const
 {
+	if (callback) {
+		return callback.get()->getDescription(player);
+	}
+
 	int32_t value;
 	player->getStorageValue(storageID, value);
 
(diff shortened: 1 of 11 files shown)


src/quests.cpp
```diff
@@ -232,3 +269,45 @@ bool Quests::isQuestStorage(const uint32_t key, const int32_t value, const int32
 	}
 	return false;
 }
+
+void Quests::createQuest(Quest* quest)
```
ranisalt, line 273: Make it an `unique_ptr`, so it symbolizes that a pointer passed to this function must not be used by the caller (it changes ownership).

MillhioreBT: I've been trying to do what you suggested, but to no avail.
Any attempt to wrap `Quest*` in a unique_ptr resulted in a `C2440` error.

I made some changes, though I don't know if it's still really necessary to wrap Quest* in unique_ptr?

PR #4191, opened 2022-08-11
Remove house from database if not exist on the map

rookgaard:
### Pull Request Prelude

When house ID will change in XML (or house will be deleted at all during map edition) it is ignored during map loads, but remain in database. I don't see a reason why, so I'm proposing following change. Only one `DELETE` query is required, as `tile_store` and `house_lists` tables are connected if parent would be deleted.

### DIFF
--- a/src/iomapserialize.cpp
+++ b/src/iomapserialize.cpp
@@ -265,11 +265,14 @@ bool IOMapSerialize::loadHouseInfo()
 	}
 
 	do {
-		House* house = g_game.map.houses.getHouse(result->getNumber<uint32_t>("id"));
+		uint32_t houseId = result->getNumber<uint32_t>("id");
+		House* house = g_game.map.houses.getHouse(houseId);
 		if (house) {
 			house->setOwner(result->getNumber<uint32_t>("owner"), false);
 			house->setPaidUntil(result->getNumber<time_t>("paid"));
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

ramon-bernardo (approved):

Zbizu (commented):
what if map fails to load? Won't all houses just get wiped?

ranisalt:
> what if map fails to load? Won't all houses just get wiped?

I believe in that case the server is aborted before that function is called, but good point.

I'm wondering if at that point isn't it easier to just assign unique IDs to houses instead, and forget about this synchronization

Zbizu:
> I believe in that case the server is aborted before that function is called, but good point.

What if `forgotten-house.xml` fails to load due to some mistake on the user's side? Are houses safe?

danyelvarejao (approved):

ranisalt:
> What if `forgotten-house.xml` fails to load due to some mistake on the user's side or some weird rme situation? Are houses safe?

In that case, maybe we should validate everything before loading? Definitely, the server should refuse to start - and preferably refuse to touch the database - if anything is out of order.

rookgaard:
> what if map fails to load? Won't all houses just get wiped?

Good point, I will check what happen and will stop making DB changes if something go wrong.

PR #4197, opened 2022-08-18
Outfit(addons) & Mounts database changes

EvilHero90:
### Pull Request Prelude

### Changes Proposed

We are moving Outfits(addons) & Mounts from storagevalues into their own respective database tables.
Increased the mount id size from `uint8_t` to `uint16_t` as we already reached 200 mounts and we are soon hitting a wall.


Zbizu:
imo not worth fixing something that isn't broken

current system has advantages of:
- occupying less space in the database
- scalability - more reserved ranges can be added for familiars, auras, wings, etc and when you have one function in aac, it's super easy to add another without changing the database structure or adding any extra queries
- being tied to quest systems - storage values is a great api and all potential problems you're going to have were already resolved by it, no need to reinvent the wheel

EvilHero90:
> imo not worth fixing something that isn't broken
>
Mounts will soon break, because we exceed the limit we are reserving for it.
> 
> current system has advantages of:
> 
>     * occupying less space in the database
> 
It might save space in database but hey... here we are in 2022 we should start making data easily accessible and not just because we can save a bit of space to make it hard for people to parse data correctly.
>     * scalability - more reserved ranges can be added for familiars, auras, wings, etc and when you have one function in aac, it's super easy to add another without changing the database structure or adding any extra queries
Actually it isn't as scalable as you think, we are reserving only 1000 storagevalues for outfits and only 10 storagevalues for mounts, the moment we hit around 260 mounts we will hit a wall anyway and it wont work correct anymore.
The problem is that you can't say for example "Hey we now have more outfits then we have reserved storagevalues, let's just increase it" because then we would collide with mounts and moving that to a higher number would result in data loss or we would have to migrate aswell.
> 
>     * being tied to quest systems - storage values is a great api and all potential problems you're going to have were already resolved by it, no need to reinvent the wheel
>
I can't really get what you are trying to tell me here, as far as things go no functions have been changed, just the way we store the outfits(addons) / mounts in the database, that's it.
> 
>     * being already used in some servers (someone will have to change his aac code to adapt if this gets implemented)
>
If someone managed to get the parsing with all this weird bitshifting done, then it probably wont take him 2mins to get it changed to this.
There is no need to pull the "backward compatibility" card here

data/migrations/34.lua
```diff
@@ -38,30 +21,27 @@ function onUpdateDatabase()
...
-					table.insert(mounts, i-1)
-				end
-			end
-
-			for k, v in pairs(mounts) do
-				local lshift = bit.lshift(1, (v % 31))
+	-- this equals to our current id range from mounts
+	for i = 1, 200 do
+		local key = mountRange + ((i-1) / 31)
+		local resultId = db.storeQuery(string.format("SELECT `player_id`, `key`, `value` FROM `player_storage` WHERE `key` = %d", key))
```
ranisalt, line 27: Maybe use `key >= 1 AND key <= 200` to avoid running 200 queries?

data/migrations/34.lua
```diff
@@ -1,25 +1,8 @@
...
-	CREATE TABLE IF NOT EXISTS `player_mounts` (
-	`player_id` int NOT NULL DEFAULT '0',
-	`mount_id` smallint unsigned NOT NULL DEFAULT '0',
-	PRIMARY KEY (`player_id`,`key`),
-	FOREIGN KEY (`player_id`) REFERENCES `players`(`id`) ON DELETE CASCADE
-	) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
-
-	ALTER TABLE players
-	ADD currentmount smallint unsigned NOT NULL DEFAULT '0';
-	]])
```
ranisalt: It was better to read with the line breaks :|

data/migrations/34.lua
```diff
@@ -38,30 +21,27 @@ function onUpdateDatabase()
...
+			until not result.next(resultId)
+			result.free(resultId)
+		end
+	end
 
-			db.asyncQuery(string.format("DELETE FROM `player_storage` WHERE `player_id` = %d AND `key` = %d", playerId, key))			
-		until not result.next(resultId)
-		result.free(resultId)
+	for i = 0, 9 do
+		db.asyncQuery(string.format("DELETE FROM `player_storage` WHERE `key` = %d", mountRange + i))
```
ranisalt, line 43: Again, just use >= and <=

ranisalt (changes requested):

data/migrations/34.lua
EvilHero90: problem is I'm getting errors when I had them in brackets but I'll try to make a better formating

data/migrations/34.lua
EvilHero90: nice, should have thought of that idea, thanks!

ramon-bernardo:
Looks good!

schema.sql
```diff
@@ -363,7 +379,7 @@ CREATE TABLE IF NOT EXISTS `towns` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET=utf8;
 
-INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '33'), ('players_record', '0');
+INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '35'), ('players_record', '0');
```
ramon-bernardo, line 382: 34?

EvilHero90: It seems like on last migration someone forgot to update the `db_version` of the `schema.sql` file. 
We are at `34.lua` which results in setting `db_version` to `35`

ramon-bernardo: db_version setting in other migrations is always the penultimate

EvilHero90: after running the `34.lua` the database updates to `35`
db_number

data/migrations/34.lua
```diff
@@ -1,3 +1,44 @@
...
+				if mount ~= 0 then
+					db.query(string.format("INSERT INTO `player_mounts` (`player_id`, `mount_id`) VALUES (%d, %d)", playerId, i))
+				end
+			until not result.next(resultId)
+			result.free(resultId)
+		end
+	end
+
+	-- deleting all outfit & mount storages at once
+	db.asyncQuery(string.format("DELETE FROM `player_storage` WHERE `key` >= %d AND `key` <= %d AND `key` >= %d AND `key` <= %d", outfitRange, outfitRange + 500, mountRange, mountRange + 10))
```
ramon-bernardo, line 41: I think OR is correct instead of AND.
```
DELETE FROM `player_storage`
WHERE (`key` >= 10001000 AND `key` <= 10001000 + 500)
  AND (`key` >= 10002001 AND `key` <= 10002001 + 10)
  ^^^
  ---> OR
```

data/migrations/34.lua
ranisalt: Since you also fetch the `key`, you can use the value returned instead of `i` for the `INSERT`  below

data/migrations/34.lua
EvilHero90: saw it a bit to late, resolved in https://github.com/otland/forgottenserver/pull/4197/commits/a4fde5c6b2c7f95237bba469d3d473e1e6e4d50e

data/migrations/34.lua
EvilHero90: does this look better to you? https://github.com/otland/forgottenserver/pull/4197/commits/218cc896a9082d8ced0f1ce1724a8d03bb7b2644

Erza (dismissed):
Looks good to me

kygov:
Saving/loading players will take even more time. Maybe it's better to solve first https://github.com/otland/forgottenserver/issues/1150 and then add more queries to the database

src/player.cpp
```diff
@@ -3980,9 +3963,9 @@ bool Player::removeOutfit(uint16_t lookType)
 
 bool Player::removeOutfitAddon(uint16_t lookType, uint8_t addons)
 {
-	for (OutfitEntry& outfitEntry : outfits) {
-		if (outfitEntry.lookType == lookType) {
-			outfitEntry.addons &= ~addons;
+	for (auto& outfit : outfits) {
+		if (outfit.first == lookType) {
+			outfit.second &= ~addons;
```
ranisalt, line 3968: Tip for this recurring pattern:

```suggestion
	for (auto& [first, second] : outfits) {
		if (first == lookType) {
			second &= ~addons;
```

(and pick better names :sweat_smile: )

src/player.h
```diff
@@ -1191,7 +1182,8 @@ class Player final : public Creature, public Cylinder
 	std::map<uint32_t, DepotChest*> depotChests;
 	std::map<uint32_t, int32_t> storageMap;
 
-	std::vector<OutfitEntry> outfits;
+	std::map<uint16_t, uint8_t> outfits;
```
ranisalt, line 1185: Any specific reason to move from a vector? It could be a vector of `pair<u16, u8>` if you wanted to drop the struct

src/player.cpp
EvilHero90: wasn't aware that you could do it like this, will change that

src/player.h
EvilHero90: lookup and erase is easier in this case but I'm fine with either way

Erza:
> Saving/loading players will take even more time. Maybe solve first #1150 (like Gesior tried) and then add more queries to the database

No, it would take about the same time, or even be minimally faster. In the storage table it would need to delete & insert just as much as it would have to do that in addons & mounts tables. But the difference with seperate tables is that it would have to mess around with the table indices much less.

src/player.cpp
Erza: > wasn't aware that you could do it like this, will change that

You can also name them like normal variables, so for example something like `for (auto& [playerLookType, playerAddons] : outfits) {` could be even clearer

I think that method was introduced in C++14 or C++17, but I could be wrong

EvilHero90: thanks, it's solved in https://github.com/otland/forgottenserver/pull/4197/commits/83573ead045b1fecca79a482cc4c87197d3f9b6d

EPuncker (approved):

MillhioreBT (dismissed):

src/player.h
MillhioreBT: I agree with @EvilHero90, now with c++20 we can use the `contains` method, and everything is more fancy

ranisalt (approved):

EPuncker (approved):

luanluciano93:
@EvilHero90 can just answer a layman's question? Why in the player's home SAVE are all the records deleted and added to the database? Since changes to these records are just adding and removing outfits and addons, could you explain it to me just for the case study?

PR #4201, opened 2022-08-26
[Feature] Login Message

moviebr:
### Changes Proposed

- Added the possibility to "close" the server with a custom warning.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -239,7 +239,7 @@ bool ConfigManager::load()
 	string[URL] = getGlobalString(L, "url", "");
 	string[LOCATION] = getGlobalString(L, "location", "");
 	string[WORLD_TYPE] = getGlobalString(L, "worldType", "pvp");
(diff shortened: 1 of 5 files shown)


src/configmanager.cpp
```diff
@@ -238,6 +239,7 @@ bool ConfigManager::load()
 	string[URL] = getGlobalString(L, "url", "");
 	string[LOCATION] = getGlobalString(L, "location", "");
 	string[WORLD_TYPE] = getGlobalString(L, "worldType", "pvp");
+	string[BLOCK_LOGIN_TEXT] = getGlobalString(L, "blockLoginText", "pvp");
```
omarcopires, line 242: I think the default message shouldn't just be "pvp" xD

omarcopires:
It's a great feature, but it would be more interesting if it allowed players with roles to login.

src/configmanager.cpp
moviebr: haha I forgot to change it, soon I'll make a commit to change it

marciNekel:
Wouldn't it be better just to change GAME_STATE_CLOSED server message to customizable one?

Issue #3569, opened 2021-08-16
summon - follow/push behaviour


emil92b:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. /summon bug

### Expected behaviour
S = This is where the summon stands related to player
X = Summon cannot **STEP** and player cannot **PUSH** the summon here
P = Player position

![192915484-a7ce2339-7e6c-4071-8a5f-7513b748678d](https://user-images.githubusercontent.com/43324808/213399448-dc166c20-856f-478d-ad6a-dbdc3df6500a.png)

if you try step on **X** position the summon should move "**randomly**" to any of the available tiles, like in this image (yellow squares)
![1](https://user-images.githubusercontent.com/43324808/129499327-465ce205-fc9f-4d5e-88d5-51602da7b470.png)

### Actual behaviour
it's set to 2 sqm regardless where it stands, also seems like it moves to the same position instead of checking what tile is currently available

EPuncker:
most likely to be related to pathfinding #2464

Zbizu:
idk what changes were made, but now they stand too far

expected:
![obraz](https://user-images.githubusercontent.com/5205079/137650631-a65f08c2-a951-4931-9429-123cf2b535fd.png)

actual:
![obraz](https://user-images.githubusercontent.com/5205079/137650636-0368636e-931d-4528-b087-79179c32b2f2.png)

emil92b:
![192915484-a7ce2339-7e6c-4071-8a5f-7513b748678d](https://user-images.githubusercontent.com/43324808/213396365-715e5b85-82c0-4174-aef3-e3a348595337.png)

emil92b:
S = This is where the summon stands related to player
X = Summon cannot **STEP** and player cannot **PUSH** the summon here
P = Player position

![192915484-a7ce2339-7e6c-4071-8a5f-7513b748678d](https://user-images.githubusercontent.com/43324808/213399448-dc166c20-856f-478d-ad6a-dbdc3df6500a.png)

emil92b:
updated first post

floki21uu:
Any possible wayt to solve that?

PR #4202, opened 2022-08-28
A* Pathfinding fixes

NRH-AA:
### Pull Request Prelude
The current A* pathfinding system does not work correctly. This pull request is to fix what is missing and even includes some additional speed to the default A* algorithm.

### Changes Proposed
1) Add targetPos to Creature::getPathTo
2) Fix the f algorithm in Map::getPathMatching
3) Add a sightline check to see if there are obstacles
4) Improve default A* Algorithm based on sightline check

**Issues addressed:** 
#3569 -- Possibly
#2464

Video of Pathfinding
https://youtu.be/TYV4qFLHr-Q

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -505,16 +505,16 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
 		return false;
 	}
 
-	if (isPathFinding) {
+	if (isPathfinding) {
 		const Creature* creature = tile->getTopCreature();
 		if (creature && tile->getTopVisibleCreature(creature) || tile->hasProperty(CONST_PROP_BLOCKPATH) ||
-                    tile->hasProperty(CONST_PROP_BLOCKSOLID)) {
+		    tile->hasProperty(CONST_PROP_BLOCKSOLID)) {
 			return false;
 		}
 	}
 
 	return !tile->hasProperty(CONST_PROP_BLOCKPROJECTILE);
-}
+};
 
 namespace {
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -550,7 +550,7 @@ bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
 
 } // namespace
 
-bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z,
+bool Map::checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z,
 					 bool isPathfinding /*= false*/)
 {
 	if (x0 == x1 && y0 == y1) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -493,7 +493,8 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool 
 	return !checkLineOfSight || isSightClear(fromPos, toPos, sameFloor);
 }
 
-bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= false*/, bool isPathfinding /*= false*/) const
+bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= false*/,
+					  bool isPathfinding /*= false*/) const
 {
 	const Tile* tile = getTile(x, y, z);
 	if (!tile) {
@@ -549,7 +550,8 @@ bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
 
 } // namespace
 
-bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool isPathfinding /*= false*/) const
+bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z,
+					 bool isPathfinding /*= false*/)
 {
 	if (x0 == x1 && y0 == y1) {
 		return true;
@@ -782,7 +784,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 			// The cost to walk to this neighbor
 			const int_fast32_t walkCost = AStarNodes::getMapWalkCost(n, pos);
 			const int_fast32_t speedCost = AStarNodes::getTileWalkCost(creature, tile);
-			const int_fast32_t distEnd = (Position::getDistanceX(pos, targetPos) + Position::getDistanceY(pos, targetPos));
+			const int_fast32_t distEnd = Position::getDistanceX(pos, targetPos) + Position::getDistanceY(pos, targetPos);
 			const int_fast32_t f = distEnd + (walkCost + speedCost);
 
 			if (neighborNode) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -750,7 +750,9 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 			}
 
 			if (fpp.keepDistance && !pathCondition.isInRange(startPos, pos, fpp)) {
-				continue;
+				if (distX < fpp.maxTargetDist && distY < fpp.maxTargetDist) {
+					continue;
+				}
 			}
 
 			// Sight is clear. We shouldn't have to move backwards.
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -806,14 +806,12 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 					// The node on the closed/open list is cheaper than this one
 					continue;
 				}
-
-				// g_game.addMagicEffect(pos, CONST_ME_TELEPORT);
+				
 				neighborNode->f = newf;
 				neighborNode->parent = n;
 				nodes.openNode(neighborNode);
 			} else {
 				// Does not exist in the open/closed list, create a new node
-				g_game.addMagicEffect(pos, CONST_ME_POFF);
 				neighborNode = nodes.createOpenNode(n, pos.x, pos.y, newf);
 				if (!neighborNode) {
 					if (found) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -667,8 +667,14 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	const Position startPos = pos;
 
 	// Don't update path if the target is too far away
-	if (Position::getDistanceX(startPos, targetPos) > fpp.maxSearchDist ||
-	    Position::getDistanceY(startPos, targetPos) > fpp.maxSearchDist) {
+	const int_fast32_t distX = Position::getDistanceX(startPos, targetPos);
+	const int_fast32_t distY = Position::getDistanceY(startPos, targetPos);
+	if (distX > fpp.maxSearchDist || distY > fpp.maxSearchDist) {
+		return false;
+	}
+
+	// We are at the target and don't need to walk away. No need to update path.
+	if (!fpp.keepDistance && fpp.minTargetDist <= 1 && distX < 2 && distY < 2) {
 		return false;
 	}
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -754,7 +754,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 			}
 
 			// Sight is clear. We shouldn't have to move backwards.
-			if (sightClear) {
+			if (sightClear && !fpp.keepDistance) {
 				if (startPos.x == targetPos.x) {
 					// Don't check nodes if start and end pos X are the same and node X is different.
 					if (pos.x != targetPos.x) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -514,7 +514,7 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
 
 namespace {
 
-bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool isPathfinding /*= false*/)
+bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool isPathfinding)
 {
 	float dx = x1 - x0;
 	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
@@ -531,7 +531,7 @@ bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t 
 	return true;
 }
 
-bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool isPathfinding /*= false*/)
+bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool isPathfinding)
 {
 	float dx = x1 - x0;
 	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -807,7 +807,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 					// The node on the closed/open list is cheaper than this one
 					continue;
 				}
-				
+
 				neighborNode->f = newf;
 				neighborNode->parent = n;
 				nodes.openNode(neighborNode);
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -550,8 +550,8 @@ bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
 
 } // namespace
 
-bool Map::checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z,
-					 bool isPathfinding /*= false*/)
+bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z,
+					 bool isPathfinding /*= false*/) const
 {
 	if (x0 == x1 && y0 == y1) {
 		return true;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -586,7 +586,8 @@ bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uin
 	return checkSlightLine(x0, y0, x1, y1, z, isPathFinding);
 }
 
-bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/, bool pathFinding /*= false*/) const
+bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/,
+                       bool pathFinding = false) const
 {
 	// target is on the same floor
 	if (fromPos.z == toPos.z) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -662,6 +662,11 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	Position endPos;
 	const Position startPos = pos;
 
+	// Don't update path if the target is too far away
+	if (Position::getDistanceX(startPos, targetPos) > fpp.maxSearchDist || Position::getDistanceY(startPos, targetPos) > fpp.maxSearchDist) {
+		return false;
+	}
+
 	int32_t bestMatch = 0;
 
 	AStarNodes nodes(pos.x, pos.y);
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -493,7 +493,7 @@ bool Map::canThrowObjectTo(const Position& fromPos, const Position& toPos, bool 
 	return !checkLineOfSight || isSightClear(fromPos, toPos, sameFloor);
 }
 
-bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= false*/) const
+bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= false*/, bool pathFinding /*= false*/) const
 {
 	const Tile* tile = getTile(x, y, z);
 	if (!tile) {
@@ -504,20 +504,31 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
 		return false;
 	}
 
+	if (pathFinding) {
+		if (tile->hasProperty(CONST_PROP_BLOCKPATH) || tile->hasProperty(CONST_PROP_BLOCKSOLID)) {
+			return false;
+		}
+	}
+
 	return !tile->hasProperty(CONST_PROP_BLOCKPROJECTILE);
 }
 
 namespace {
 
-bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
+bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool pathFinding /*= false*/)
 {
 	float dx = x1 - x0;
 	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
 	float yi = y0 + slope;
 
+	bool isPathFinding = false;
+	if (pathFinding) {
+		isPathFinding = true;
+	}
+
 	for (uint16_t x = x0 + 1; x < x1; ++x) {
 		// 0.1 is necessary to avoid loss of precision during calculation
-		if (!g_game.map.isTileClear(std::floor(yi + 0.1), x, z)) {
+		if (!g_game.map.isTileClear(std::floor(yi + 0.1), x, z, false, isPathFinding)) {
 			return false;
 		}
 		yi += slope;
@@ -526,15 +537,20 @@ bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t 
 	return true;
 }
 
-bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
+bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool pathFinding /*= false*/)
 {
 	float dx = x1 - x0;
 	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
 	float yi = y0 + slope;
 
+	bool isPathFinding = false;
+	if (pathFinding) {
+		isPathFinding = true;
+	}
+
 	for (uint16_t x = x0 + 1; x < x1; ++x) {
 		// 0.1 is necessary to avoid loss of precision during calculation
-		if (!g_game.map.isTileClear(x, std::floor(yi + 0.1), z)) {
+		if (!g_game.map.isTileClear(x, std::floor(yi + 0.1), z, false, isPathFinding)) {
 			return false;
 		}
 		yi += slope;
@@ -545,33 +561,43 @@ bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
 
 } // namespace
 
-bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const
+bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool pathFinding /*= false*/) const
 {
 	if (x0 == x1 && y0 == y1) {
 		return true;
 	}
 
+	bool isPathFinding = false;
+	if (pathFinding) {
+		isPathFinding = true;
+	}
+
 	if (std::abs(y1 - y0) > std::abs(x1 - x0)) {
 		if (y1 > y0) {
-			return checkSteepLine(y0, x0, y1, x1, z);
+			return checkSteepLine(y0, x0, y1, x1, z, isPathFinding);
 		}
-		return checkSteepLine(y1, x1, y0, x0, z);
+		return checkSteepLine(y1, x1, y0, x0, z, isPathFinding);
 	}
 
 	if (x0 > x1) {
-		return checkSlightLine(x1, y1, x0, y0, z);
+		return checkSlightLine(x1, y1, x0, y0, z, isPathFinding);
 	}
 
-	return checkSlightLine(x0, y0, x1, y1, z);
+	return checkSlightLine(x0, y0, x1, y1, z, isPathFinding);
 }
 
-bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/) const
+bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/, bool pathFinding /*= false*/) const
 {
 	// target is on the same floor
 	if (fromPos.z == toPos.z) {
 		// skip checks if toPos is next to us
 		if (Position::getDistanceX(fromPos, toPos) < 2 && Position::getDistanceY(fromPos, toPos) < 2) {
 			return true;
+		}
+
+		// Path finding line is clear
+		if (pathFinding) {
+			return checkSightLine(fromPos.x, fromPos.y, toPos.x, toPos.y, fromPos.z, true);
 		}
 
 		// sight is clear or sameFloor is enabled
@@ -643,24 +669,18 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
 	return tile;
 }
 
-bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirList,
+bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::vector<Direction>& dirList,
                           const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const
 {
 	Position pos = creature.getPosition();
 	Position endPos;
-
-	AStarNodes nodes(pos.x, pos.y);
+	const Position startPos = pos;
 
 	int32_t bestMatch = 0;
 
-	static int_fast32_t dirNeighbors[8][5][2] = {
-	    {{-1, 0}, {0, 1}, {1, 0}, {1, 1}, {-1, 1}},    {{-1, 0}, {0, 1}, {0, -1}, {-1, -1}, {-1, 1}},
-	    {{-1, 0}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}}, {{0, 1}, {1, 0}, {0, -1}, {1, -1}, {1, 1}},
-	    {{1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}},  {{-1, 0}, {0, -1}, {-1, -1}, {1, -1}, {-1, 1}},
-	    {{0, 1}, {1, 0}, {1, -1}, {1, 1}, {-1, 1}},    {{-1, 0}, {0, 1}, {-1, -1}, {1, 1}, {-1, 1}}};
-	static int_fast32_t allNeighbors[8][2] = {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}};
+	AStarNodes nodes(pos.x, pos.y);
 
-	const Position startPos = pos;
+	bool sightClear = isSightClear(startPos, targetPos, true, true);
 
 	AStarNode* found = nullptr;
 	while (fpp.maxSearchDist != 0 || nodes.getClosedNodes() < 100) {
@@ -718,7 +738,6 @@ bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirL
 			neighbors = *allNeighbors;
 		}
 
-		const int_fast32_t f = n->f;
 		for (uint_fast32_t i = 0; i < dirCount; ++i) {
 			pos.x = x + *neighbors++;
 			pos.y = y + *neighbors++;
@@ -732,6 +751,39 @@ bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirL
 				continue;
 			}
 
+			// Sight is clear. We shouldn't have to move backwards.
+			if (sightClear) {
+				if (startPos.x == targetPos.x) {
+					// Don't check nodes if start and end pos X are the same and node X is different.
+					if (pos.x != targetPos.x) {
+						continue;
+					}
+				} else if (startPos.y == targetPos.y) {
+					// Don't check nodes if start and end pos Y are the same and node Y is different.
+					if (pos.y != targetPos.y) {
+						continue;
+					}
+				}
+
+				const int_fast32_t startDistXtarget = std::abs(startPos.x - targetPos.x);
+				const int_fast32_t startDistYtarget = std::abs(startPos.y - targetPos.y);
+				const int_fast32_t startDistXNode = std::abs(startPos.x - pos.x);
+				const int_fast32_t startDistYNode = std::abs(startPos.y - pos.y);
+				// We shouldn't check past the targets X or Y
+				if (startDistXtarget < startDistXNode) {
+					continue;
+				} else if (startDistYtarget < startDistYNode) {
+					continue;
+				}
+
+				// We don't need to check behind us. The sight is clear to go forward.
+				const int_fast32_t nodeDistXtarget = std::abs(pos.x - targetPos.x);
+				const int_fast32_t nodeDistYtarget = std::abs(pos.y - targetPos.y);
+				if (startDistXtarget < nodeDistXtarget || startDistYtarget < nodeDistYtarget) {
+					continue;
+				}
+			}
+
 			const Tile* tile;
 			AStarNode* neighborNode = nodes.getNodeByPosition(pos.x, pos.y);
 			if (neighborNode) {
@@ -743,10 +795,11 @@ bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirL
 				}
 			}
 
-			// The cost (g) for this neighbor
-			const int_fast32_t cost = AStarNodes::getMapWalkCost(n, pos);
-			const int_fast32_t extraCost = AStarNodes::getTileWalkCost(creature, tile);
-			const int_fast32_t newf = f + cost + extraCost;
+			// The cost to walk to this neighbor
+			const int_fast32_t G = AStarNodes::getMapWalkCost(n, pos);
+			const int_fast32_t E = AStarNodes::getTileWalkCost(creature, tile);
+			const int_fast32_t H = (Position::getDistanceX(pos, targetPos) + Position::getDistanceY(pos, targetPos));
+			const int_fast32_t newf = H + (G + E);
 
 			if (neighborNode) {
 				if (neighborNode->f <= newf) {
@@ -754,11 +807,13 @@ bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirL
 					continue;
 				}
 
+				// g_game.addMagicEffect(pos, CONST_ME_TELEPORT);
 				neighborNode->f = newf;
 				neighborNode->parent = n;
 				nodes.openNode(neighborNode);
 			} else {
 				// Does not exist in the open/closed list, create a new node
+				g_game.addMagicEffect(pos, CONST_ME_POFF);
 				neighborNode = nodes.createOpenNode(n, pos.x, pos.y, newf);
 				if (!neighborNode) {
 					if (found) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -514,7 +514,7 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
 	}
 
 	return !tile->hasProperty(CONST_PROP_BLOCKPROJECTILE);
-};
+}
 
 namespace {
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -587,7 +587,7 @@ bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uin
 }
 
 bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/,
-                       bool pathFinding = false) const
+                       bool pathFinding /*= false*/) const
 {
 	// target is on the same floor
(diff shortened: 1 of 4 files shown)


gigastar:
As of now the pathfinding system works like dijkstra's pathfinding or w/e. It will get the actual best path because it checks all directions.

With my changes it will work as normal A* only a bit faster and better because of the sightline check. The only problem is, it will walk closer to a wall than it needs to if the target is on the other side.

These changes will make all pathfinding WAY FASTER. Faster meaning a lot less CPU usage. A lot less tiles being checked.

We could tell it to do the default A* if the path is clear and do dijkstra's when it isn't. Even that would help a lot.

gigastar:
I am almost done with code that should allow A* to work even when obstacles are in the way. It should be able to find the correct path a lot better instead of going into walls. Ill keep this updated. Please review so I can make sure the compilation problems were addressed.

src/map.cpp
```diff
@@ -504,20 +504,31 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
...
+bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool pathFinding /*= false*/)
 {
 	float dx = x1 - x0;
 	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
 	float yi = y0 + slope;
 
+	bool isPathFinding = false;
+	if (pathFinding) {
+		isPathFinding = true;
+	}
```
ranisalt, line 527: Could use `pathFinding` straight away? Maybe even rename it to `isPathFinding`

ranisalt:
Fantastic. This bit of code is overdue for an improvement.

gigastar:
Almost have a perfect A* set up.. Just need to add the ability to store costs already used and change the algorithm based on that. Should be able to make it so it will find the best path while using less nodes and whenever there is no obstacles use barley any at all. I will push it here if this hasn't been approved and merged already.

src/map.cpp
```diff
@@ -545,27 +550,29 @@ bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
 
 } // namespace
 
-bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const
+bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z,
```
ranisalt, line 553: Typo here, you need to revert the function name change

conde2:
```I think this should be pushed then we can work on adding something that if the path line is blocked it checks around for tiles that it isn't. Then changes the creatures endPos to that position.

 Then if the target walks out of screen it will remove the endPos. Otherwise once it gets to that position it will check the path to the target again and repeat. That is the best method IMO. This is the first stop though.```

This seems to be the best of the two worlds. Would be nice if you benchmark the algorithms and make a video on a more realistic scenario with a lot of trees and monsters around.

**Comment by gigastar** [comment]
> This seems to be the best of the two worlds. Would be nice if you benchmark the algorithms and make a video on a more realistic scenario with a lot of trees and creatures moving around.

If I get some extra time ill try to do that. I wasn't able to bench mark but I did add in a check for how many tiles were being checked.

The default pathfinding would check up to 94 tiles (if the target was directly behind a obstacles) and always checked more tiles in general. The low end being around 10 avg. (If the target was only a few squares away with no obstacles)

With these changes the most I was able to get it too with 2+ different obstacles  was about 50. 
The same scenario that the previous pathfinding system did 94 only did 35 with this change.
Lastly, the low end being a fraction of the defaults.

File: src/map.cpp [hunk 3]
```diff
@@ -545,27 +550,29 @@ bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
 
 } // namespace
 
-bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const
+bool Map::checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z,
+					 bool isPathfinding /*= false*/)
```
EvilHero90, line 554: forgot `const` at the end

EvilHero90:
Did you already benchmark? I would really like to see the results, because code wise this looks good to me but seeing some benchmark statistics wouldn't be bad at all.

gigastar:
> Did you already benchmark? I would really like to see the results, because code wise this looks good to me but seeing some benchmark statistics wouldn't be bad at all.

I am working on a couple things really quick then ill get some benchmarks going. There is an issue which is if the target goes out of the search distance the code will run the max 100 tiles which I have fixed.

The other thing is if the target is in range but the paths too him are too long (hit max distance) it will also check all 100 nodes looking in all directions for a path to him. I am working out a way to not do that. That is a big part of the performance.

gigastar:
Alright, so after doing some bench marks the results are:

Both systems had a max time of: 0.00100 sec

Difference:
The new pathfinding NEVER clocked at 0.00100 while the path was clear.
The default clocked at 0.00100 multiple times while the path was clear.

Both are about the same when the path is blocked. It is hard to tell because it only shows
0.00100 or 0.00000.

I don't really know the best way to benchmark times though. I did this using: clock_t

conde2:
> Alright, so after doing some bench marks the results are:
> 
> Both systems had a max time of: 0.00100 sec
> 
> Difference: The new pathfinding NEVER clocked at 0.00100 while the path was clear. The default clocked at 0.00100 multiple times while the path was clear.
> 
> Both are about the same when the path is blocked. It is hard to tell because it only shows 0.00100 or 0.00000.
> 
> I don't really know the best way to benchmark times though. I did this using: clock_t



You can use a single header library that does the work for you.

Here you have an example
https://github.com/martinus/nanobench

just include the header in the project and run the function that does benchmark the code

EvilHero90:
You can also use `std::chrono::high_resolution_clock` that gives good results aswell.

gigastar:
Alright, I will add more as they come. Just wanted to update.

This is benchmarks with clear sight and straight line

Default
[Distance]:2[Time]:17000
[Distance]:2[Time]:18200
[Distance]:2[Time]:25100

[Distance]:3[Time]:30400
[Distance]:3[Time]:31400
[Distance]:3[Time]:32700

[Distance]:4[Time]:50000
[Distance]:4[Time]:44800
[Distance]:4[Time]:61500

[Distance]:5[Time]:77800
[Distance]:5[Time]:80800
[Distance]:5[Time]:63200

[Distance]:10[Time]:249000
[Distance]:10[Time]:284500
[Distance]:10[Time]:297900

NEW ASTAR
[Distance]:2[Time]:7100
[Distance]:2[Time]:8200
[Distance]:2[Time]:11900

[Distance]:3[Time]:8700
[Distance]:3[Time]:16200
[Distance]:3[Time]:33000

[Distance]:4[Time]:15300
[Distance]:4[Time]:16800
[Distance]:4[Time]:17000

[Distance]:5[Time]:25700
[Distance]:5[Time]:18500
[Distance]:5[Time]:17200

[Distance]:10[Time]:30200
[Distance]:10[Time]:39000
[Distance]:10[Time]:40100

[Distance]:13[Time]:35500
[Distance]:13[Time]:43500
[Distance]:13[Time]:43900

conde2:
> Alright, I will add more as they come. Just wanted to update.
> 
> This is benchmarks with clear sight and straight line
> 
> https://pastebin.com/KjJ22K7H

The results sounds amazing, which measurement unit did you use?

gigastar:
I am going to note... This still needs more work. It's a lot more efficient than current BUT it is still set up as: Closest Node First

To get a true A* this is what needs to happen:

1) Create a way to store highest cost so far
2) Check if the next nodes cost is more than the highest
3) Check if any other nodes around have the same or lower cost
4) Create new node only if there isn't any with the same or lower
5) Set the new highest cost so far
6) Base the algorith off of:
   -StartDist to TargetDist
   -NodeDist to StartDist
   -WalkCost
   -TileCost

ranisalt:
Food for thought, don't know if we could ever apply that, but Boost provides an A* algorithm: https://www.boost.org/doc/libs/1_80_0/libs/graph/doc/astar_search.html

EvilHero90:
sounds good to me, once you consider it "stable" and it looks good we should merge it, we can still do more optimizations over the time but this is a good start.

gigastar:
She is stable. Once this gets merged I will do another pull for updates to when pathfinding gets called. That will be the questionable one.

Flatlander57:
This really does help. Especially if you made the screen-size bigger for your client.

I've been using basically the same thing since 2018.
https://github.com/otland/forgottenserver/pull/2464

Here is the testing and conversation about these changes on otland.
https://otland.net/threads/optimizing-tfs-pathfinding.257868/

You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)

gigastar:
> This really does help. Especially if you made the screen-size bigger for your client.
> 
> I've been using basically the same thing since 2018. #2464
> 
> Here is the testing and conversation about these changes on otland. https://otland.net/threads/optimizing-tfs-pathfinding.257868/
> 
> You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)

Does your set up act as best node first or does it actually get the best possible node. To check it see if a monster will run into a wall if it is between the target and start position.

Flatlander57:
> > This really does help. Especially if you made the screen-size bigger for your client.
> > I've been using basically the same thing since 2018. #2464
> > Here is the testing and conversation about these changes on otland. https://otland.net/threads/optimizing-tfs-pathfinding.257868/
> > You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)
> 
> Does your set up act as best node first or does it actually get the best possible node. To check it see if a monster will run into a wall if it is between the target and start position.

Mine will do the same exact thing that the normal pathfinding does. Except it checks the "most likely" to work nodes first.
It should get the exact same path that the normal pathfinding system would take, just more efficiently.

gigastar:
> > > This really does help. Especially if you made the screen-size bigger for your client.
> > > I've been using basically the same thing since 2018. #2464
> > > Here is the testing and conversation about these changes on otland. https://otland.net/threads/optimizing-tfs-pathfinding.257868/
> > > You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)
> > 
> > 
> > Does your set up act as best node first or does it actually get the best possible node. To check it see if a monster will run into a wall if it is between the target and start position.
> 
> Mine will do the same exact thing that the normal pathfinding does. Except it checks the "most likely" to work nodes first. It should get the exact same path that the normal pathfinding system would take, just more efficiently.

So, I am almost done with a perfect A* code set up... I am just having some trouble with one thing. West/South the code works perfectly. North/east it breaks for some reason.

Flatlander57:
I don't really know what additional features you are adding.

I didn't want to change the AI of creatures, (They still should choose the fastest path to target, regardless of line of sight, etc)
If you have ideas on how to speed up path-finding beyond checking the "closest" node to the target position first then please do let me know what your ideas are and I can look into them.

Flatlander57:
> > > > This really does help. Especially if you made the screen-size bigger for your client.
> > > > I've been using basically the same thing since 2018. #2464
> > > > Here is the testing and conversation about these changes on otland. https://otland.net/threads/optimizing-tfs-pathfinding.257868/
> > > > You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)
> > > 
> > > 
> > > Does your set up act as best node first or does it actually get the best possible node. To check it see if a monster will run into a wall if it is between the target and start position.
> > 
> > 
> > Mine will do the same exact thing that the normal pathfinding does. Except it checks the "most likely" to work nodes first. It should get the exact same path that the normal pathfinding system would take, just more efficiently.
> 
> So, I am almost done with a perfect A* code set up... I am just having some trouble with one thing. West/South the code works perfectly. North/east it breaks for some reason.

I don't really know what additional features you are adding.

I didn't want to change the AI of creatures, (They still should choose the fastest path to target, regardless of line of sight, etc)
If you have ideas on how to speed up path-finding beyond checking the "closest" node to the target position first then please do let me know what your ideas are and I can look into them.

gigastar:
> > > > > This really does help. Especially if you made the screen-size bigger for your client.
> > > > > I've been using basically the same thing since 2018. #2464
> > > > > Here is the testing and conversation about these changes on otland. https://otland.net/threads/optimizing-tfs-pathfinding.257868/
> > > > > You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)
> > > > 
> > > > 
> > > > Does your set up act as best node first or does it actually get the best possible node. To check it see if a monster will run into a wall if it is between the target and start position.
> > > 
> > > 
> > > Mine will do the same exact thing that the normal pathfinding does. Except it checks the "most likely" to work nodes first. It should get the exact same path that the normal pathfinding system would take, just more efficiently.
> > 
> > 
> > So, I am almost done with a perfect A* code set up... I am just having some trouble with one thing. West/South the code works perfectly. North/east it breaks for some reason.
> 
> I don't really know what additional features you are adding.
> 
> I didn't want to change the AI of creatures, (They still should choose the fastest path to target, regardless of line of sight, etc) If you have ideas on how to speed up path-finding beyond checking the "closest" node to the target position first then please do let me know what your ideas are and I can look into them.

Its mainly how the nodes are created, stored, and looped through. A* when done correctly will only open nodes it needs too. Its possible your set up is correct as far as that but it is still missing efficiency in how it is storing nodes, accessing them, w.e.

Could you add some code and check how many nodes are being looped through? Not just opened but checked in general. Pretty much just add a counter to the start of each neighbor node check.

The amount of nodes checked should be something like this:

DistEnd + distObstacle + distAroundObs(n) * n(o) + (possibleNeighbors - (possibleNeighbors - 1sqm of neighbors)

Distance to end +
distance to obstacles +
distance to get around the obstacle *
the number of those obstacles

So, if the path is 2sqm with 1 wall that needs 4 sqm to pass you should check

2 + (4)(1) * (possibleNeighbors - (possibleNeighbors - 1sqm of neighbors)

So roughly only 18 or so nodes should ever be looked at in that instance. It is hard to explain but that should be all the nodes to the wall, then check both sides, one of the sides would have 1sqm not checked compared to the other because it found the path on the other side first. Then all the nodes to get to the target.

I have a feeling the code you are using is including additional nodes just because it is using the same loop/node access/node storing methods as the default code that was there.

Your code:
1) Parent's neighbors are checked
2) Values are stored for how much the F value is
3) It picks the best node based on F and distance to target

It is faster in some cases because the lowest f and closest node is usually the best to pick. When it isn't though it has to go back through and keep trying until it actually gets the best path. Essentially, it is adding nodes it doesn't need and you ignore them by using the getBestNode algorithm that you are using. So there are a few instances that way more nodes/paths will need to be created and checked.

1) Each parent should have all its neighbors checked
2) Any neighbors that can't be created are ignored
3) It should assign values on the cost of using that parent to get there and how long until it hits the end
4) The best neighbor is added to the list that should be checked.
5) If it finds a open node that has a lower cost based on the parents cost already and the old parents cost
6) Change the parent of the node and change its cost.

Now the difference is: only the nodes we actually know are the best nodes will be checked for its neighbors. The list we loop through will only have to loop through the nodes it HAS to to find the path. It's pretty hard to explain in just text.

The best way to sum it up is: A* never goes on the wrong path finding the end. The codes we have now will go through wrong paths because they are lower cost. Once its path has a higher cost it will start looping through the next path. On the surface its undetectable but once you have 10k algorithms running it is very obvious when you have it working correctly or not.

Another thing to know is A* will not work at all until it is done correctly. 
The code you have is a mix of searching all paths and prioritizing the best path. 
A* is finding the best path based on a best node heuristic.  

It sounds the same but in practice is a whole different story.

I can show you an example of what im doing if you like.

Flatlander57:
> > > > > > This really does help. Especially if you made the screen-size bigger for your client.
> > > > > > I've been using basically the same thing since 2018. #2464
> > > > > > Here is the testing and conversation about these changes on otland. https://otland.net/threads/optimizing-tfs-pathfinding.257868/
> > > > > > You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)
> > > > > 
> > > > > 
> > > > > Does your set up act as best node first or does it actually get the best possible node. To check it see if a monster will run into a wall if it is between the target and start position.
> > > > 
> > > > 
> > > > Mine will do the same exact thing that the normal pathfinding does. Except it checks the "most likely" to work nodes first. It should get the exact same path that the normal pathfinding system would take, just more efficiently.
> > > 
> > > 
> > > So, I am almost done with a perfect A* code set up... I am just having some trouble with one thing. West/South the code works perfectly. North/east it breaks for some reason.
> > 
> > 
> > I don't really know what additional features you are adding.
> > I didn't want to change the AI of creatures, (They still should choose the fastest path to target, regardless of line of sight, etc) If you have ideas on how to speed up path-finding beyond checking the "closest" node to the target position first then please do let me know what your ideas are and I can look into them.
> 
> Its mainly how the nodes are created, stored, and looped through. A* when done correctly will only open nodes it needs too. Its possible your set up is correct as far as that but it is still missing efficiency in how it is storing nodes, accessing them, w.e.
> 
> Could you add some code and check how many nodes are being looped through? Not just opened but checked in general. Pretty much just add a counter to the start of each neighbor node check.
> 
> The amount of nodes checked should be something like this:
> 
> DistEnd + distObstacle + distAroundObs(n) * n(o) + (possibleNeighbors - (possibleNeighbors - 1sqm of neighbors)
> 
> Distance to end + distance to obstacles + distance to get around the obstacle * the number of those obstacles
> 
> So, if the path is 2sqm with 1 wall that needs 4 sqm to pass you should check
> 
> 2 + (4)(1) * (possibleNeighbors - (possibleNeighbors - 1sqm of neighbors)
> 
> So roughly only 18 or so nodes should ever be looked at in that instance. It is hard to explain but that should be all the nodes to the wall, then check both sides, one of the sides would have 1sqm not checked compared to the other because it found the path on the other side first. Then all the nodes to get to the target.
> 
> I have a feeling the code you are using is including additional nodes just because it is using the same loop/node access/node storing methods as the default code that was there.
> 
> Your code:
> 
>     1. Parent's neighbors are checked
> 
>     2. Values are stored for how much the F value is
> 
>     3. It picks the best node based on F and distance to target
> 
> 
> It is faster in some cases because the lowest f and closest node is usually the best to pick. When it isn't though it has to go back through and keep trying until it actually gets the best path. Essentially, it is adding nodes it doesn't need and you ignore them by using the getBestNode algorithm that you are using. So there are a few instances that way more nodes/paths will need to be created and checked.
> 
>     1. Each parent should have all its neighbors checked
> 
>     2. Any neighbors that can't be created are ignored
> 
>     3. It should assign values on the cost of using that parent to get there and how long until it hits the end
> 
>     4. The best neighbor is added to the list that should be checked.
> 
>     5. If it finds a open node that has a lower cost based on the parents cost already and the old parents cost
> 
>     6. Change the parent of the node and change its cost.
> 
> 
> Now the difference is: only the nodes we actually know are the best nodes will be checked for its neighbors. The list we loop through will only have to loop through the nodes it HAS to to find the path. It's pretty hard to explain in just text.
> 
> The best way to sum it up is: A* never goes on the wrong path finding the end. The codes we have now will go through wrong paths because they are lower cost. Once its path has a higher cost it will start looping through the next path. On the surface its undetectable but once you have 10k algorithms running it is very obvious when you have it working correctly or not.
> 
> Another thing to know is A* will not work at all until it is done correctly. The code you have is a mix of searching all paths and prioritizing the best path. A* is finding the best path based on a best node heuristic.
> 
> It sounds the same but in practice is a whole different story.
> 
> I can show you an example of what im doing if you like.

I ran the same test you have (with the rat following you in the video) on mine, with the same walls and it checks the same nodes. (Same animation plays as your video)

My pathfinding system is a little more complex because monsters in my game are able to path up/down stairs and track players from around 50sqm away. Which is why I had to upgrade the pathfinding system to be super-efficient.

It is very simple. There is no way for the pathfinding system to know the correct path, but it is more likely that the correct path will be in the direction of the targetPosition. So simply make it more costly each sqm you go away from the targetPos so when it is selecting a "getBestNode" it always picks the node that is more likely to be the correct path (is closest to the targetPosition)

gigastar:
> > > > > > > This really does help. Especially if you made the screen-size bigger for your client.
> > > > > > > I've been using basically the same thing since 2018. #2464
> > > > > > > Here is the testing and conversation about these changes on otland. https://otland.net/threads/optimizing-tfs-pathfinding.257868/
> > > > > > > You might want to test both versions and see which is better. (I didn't look too far into it to see if there were any meaningful differences)
> > > > > > 
> > > > > > 
> > > > > > Does your set up act as best node first or does it actually get the best possible node. To check it see if a monster will run into a wall if it is between the target and start position.
> > > > > 
> > > > > 
> > > > > Mine will do the same exact thing that the normal pathfinding does. Except it checks the "most likely" to work nodes first. It should get the exact same path that the normal pathfinding system would take, just more efficiently.
> > > > 
> > > > 
> > > > So, I am almost done with a perfect A* code set up... I am just having some trouble with one thing. West/South the code works perfectly. North/east it breaks for some reason.
> > > 
> > > 
> > > I don't really know what additional features you are adding.
> > > I didn't want to change the AI of creatures, (They still should choose the fastest path to target, regardless of line of sight, etc) If you have ideas on how to speed up path-finding beyond checking the "closest" node to the target position first then please do let me know what your ideas are and I can look into them.
> > 
> > 
> > Its mainly how the nodes are created, stored, and looped through. A* when done correctly will only open nodes it needs too. Its possible your set up is correct as far as that but it is still missing efficiency in how it is storing nodes, accessing them, w.e.
> > Could you add some code and check how many nodes are being looped through? Not just opened but checked in general. Pretty much just add a counter to the start of each neighbor node check.
> > The amount of nodes checked should be something like this:
> > DistEnd + distObstacle + distAroundObs(n) * n(o) + (possibleNeighbors - (possibleNeighbors - 1sqm of neighbors)
> > Distance to end + distance to obstacles + distance to get around the obstacle * the number of those obstacles
> > So, if the path is 2sqm with 1 wall that needs 4 sqm to pass you should check
> > 2 + (4)(1) * (possibleNeighbors - (possibleNeighbors - 1sqm of neighbors)
> > So roughly only 18 or so nodes should ever be looked at in that instance. It is hard to explain but that should be all the nodes to the wall, then check both sides, one of the sides would have 1sqm not checked compared to the other because it found the path on the other side first. Then all the nodes to get to the target.
> > I have a feeling the code you are using is including additional nodes just because it is using the same loop/node access/node storing methods as the default code that was there.
> > Your code:
> > ```
> > 1. Parent's neighbors are checked
> > 
> > 2. Values are stored for how much the F value is
> > 
> > 3. It picks the best node based on F and distance to target
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is faster in some cases because the lowest f and closest node is usually the best to pick. When it isn't though it has to go back through and keep trying until it actually gets the best path. Essentially, it is adding nodes it doesn't need and you ignore them by using the getBestNode algorithm that you are using. So there are a few instances that way more nodes/paths will need to be created and checked.
> > ```
> > 1. Each parent should have all its neighbors checked
> > 
> > 2. Any neighbors that can't be created are ignored
> > 
> > 3. It should assign values on the cost of using that parent to get there and how long until it hits the end
> > 
> > 4. The best neighbor is added to the list that should be checked.
> > 
> > 5. If it finds a open node that has a lower cost based on the parents cost already and the old parents cost
> > 
> > 6. Change the parent of the node and change its cost.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now the difference is: only the nodes we actually know are the best nodes will be checked for its neighbors. The list we loop through will only have to loop through the nodes it HAS to to find the path. It's pretty hard to explain in just text.
> > The best way to sum it up is: A* never goes on the wrong path finding the end. The codes we have now will go through wrong paths because they are lower cost. Once its path has a higher cost it will start looping through the next path. On the surface its undetectable but once you have 10k algorithms running it is very obvious when you have it working correctly or not.
> > Another thing to know is A* will not work at all until it is done correctly. The code you have is a mix of searching all paths and prioritizing the best path. A* is finding the best path based on a best node heuristic.
> > It sounds the same but in practice is a whole different story.
> > I can show you an example of what im doing if you like.
> 
> I ran the same test you have (with the rat following you in the video) on mine, with the same walls and it checks the same nodes. (Same animation plays as your video)
> 
> My pathfinding system is a little more complex because monsters in my game are able to path up/down stairs and track players from around 50sqm away. Which is why I had to upgrade the pathfinding system to be super-efficient.
> 
> It is very simple. There is no way for the pathfinding system to know the correct path, but it is more likely that the correct path will be in the direction of the targetPosition. So simply make it more costly each sqm you go away from the targetPos so when it is selecting a "getBestNode" it always picks the node that is more likely to be the correct path (is closest to the targetPosition)

Alright, I tested your code and did some tests. As suspected it is best node first. This is why A* is such a pain. It is very specific. Your code is a lot faster than the default but falls behind this code. It is because of the list of reasons I gave above. Just because you are changing the getBestNode doesn't mean it isn't adding just as many nodes. You are just ignoring them using getBestNode. Which does make it faster but still not as fast as it should be. Also its not getting the actual best path.

This is the results of your pathfinding using the same time measurements.
https://pastebin.com/HGNTVQDf

These are the results of default and mine. I already showed it before. Note that I didn't organize or differentiate what lines were happening in your code. Just know the time that are a lot bigger are ones that the sightline was blocked.

Straight Line + Clear Sight
https://pastebin.com/KjJ22K7H

Diag + Clear Sight
https://pastebin.com/EUPZMQux

Blocked Sight
https://pastebin.com/7fjvtm2H

Doing a straight path your code took: [Distance]:1[Time]:20700

With mine we could see as low as: [Distance]:2[Time]:7100

Defaults we see: [Distance]:2[Time]:17000

Flatlander57:
> > I ran the same test you have (with the rat following you in the video) on mine, with the same walls and it checks the same nodes. (Same animation plays as your video)
> > My pathfinding system is a little more complex because monsters in my game are able to path up/down stairs and track players from around 50sqm away. Which is why I had to upgrade the pathfinding system to be super-efficient.
> > It is very simple. There is no way for the pathfinding system to know the correct path, but it is more likely that the correct path will be in the direction of the targetPosition. So simply make it more costly each sqm you go away from the targetPos so when it is selecting a "getBestNode" it always picks the node that is more likely to be the correct path (is closest to the targetPosition)
> 
> Alright, I tested your code and did some tests. As suspected it is best node first. This is why A* is such a pain. It is very specific. Your code is a lot faster than the default but falls behind this code. It is because of the list of reasons I gave above. Just because you are changing the getBestNode doesn't mean it isn't adding just as many nodes. You are just ignoring them using getBestNode. Which does make it faster but still not as fast as it should be. Also its not getting the actual best path.
> 
> This is the results of your pathfinding using the same time measurements. https://pastebin.com/HGNTVQDf
> 
> These are the results of default and mine. I already showed it before. Note that I didn't organize or differentiate what lines were happening in your code. Just know the time that are a lot bigger are ones that the sightline was blocked.
> 
> Straight Line + Clear Sight https://pastebin.com/KjJ22K7H
> 
> Diag + Clear Sight https://pastebin.com/EUPZMQux
> 
> Blocked Sight https://pastebin.com/7fjvtm2H
> 
> Doing a straight path your code took: [Distance]:2[Time]:20700
> 
> With mine we could see as low as: [Distance]:2[Time]:7100
> 
> Defaults we see: [Distance]:2[Time]:17000
> 
> NOTE** inside the first pastebin with your algorithms results. Add +1 to the distance. I forgot to do that.

On my thread I linked, if you scroll down Mark made changes to make it even faster. That is the code I am using. 

If you use Mark's code and test it and it is slower than yours, then I would be very interested in what you are doing different.
TFS's Original code already does everything you listed above.

1. Each parent should have all its neighbors checked 
-- TFS does this by default

2. Any neighbors that can't be created are ignored 
-- TFS does this by default

3. It should assign values on the cost of using that parent to get there and how long until it hits the end 
-- TFS does this by default

4. The best neighbor is added to the list that should be checked. 
-- all open nodes are looked through to find the "best possible match". I simply changed it to also use distance to targetPos as an additional metric to choose the node most likely to be the correct path.

5. If it finds a open node that has a lower cost based on the parents cost already and the old parents cost 
-- TFS does this by default

6. Change the parent of the node and change its cost. 
-- TFS does this by default

I think all the things are you trying to add is already in TFS by default.
If I am wrong, then let me know.

gigastar:
> I think all the things are you trying to add is already in TFS by default. If I am wrong, then let me know.

You are partially correct. The default code does do all of this. Where the problem comes in is:

1) How many nodes are added that aren't needed
  - When going in a straight line there should only be a line of nodes checked
  - I kind of half assed creating that in my code but it should be done by the algorithm itself
2) How the nodes are created and stored
  - There are different ways of doing the same thing some are a lot faster than others

One of the big problems is A* generally uses something like this to find the distance from the target:
>float getDistance(uint16_t x1, uint16_t y1, uint16_t x2, uint16_t y2) {
>			return sqrtf((x1 - x2)*(x1-x2)+(y1-y2)*(y1-y2));
>}

Also, instead of looping through the nodes like this
AStarNode* AStarNodes::getBestNode(const Position& targetPos)
{
	if (curNode == 0) {
		return nullptr;
	}

	int32_t diffNode = 0;
	int32_t best_node_f = std::numeric_limits<int32_t>::max();
	int32_t best_node = -1;
	for (size_t i = 0; i < curNode; i++) {
		const int_fast32_t Sx = nodes[0].x;
		const int_fast32_t Sy = nodes[0].y;
		const int_fast32_t Cx = nodes[i].x;
		const int_fast32_t Cy = nodes[i].y;
		int32_t SdiffX = std::abs(targetPos.x - Sx);
		int32_t SdiffY = std::abs(targetPos.y - Sy);
		int32_t NdiffX = std::abs(targetPos.x - Cx);
		int32_t NdiffY = std::abs(targetPos.y - Cy);
		diffNode = ((nodes[i].f + (((NdiffX - SdiffX) * 5) + ((NdiffY - SdiffY) * 5))) + (std::max(NdiffX, NdiffY) * 5));
		if (openNodes[i] && diffNode < best_node_f) {
			best_node_f = diffNode;
			best_node = i;
		}
	}

	if (best_node >= 0) {
		return nodes + best_node;
	}
	return nullptr;
}

You can easily use something like this:
openNodes.sort([](const AStarNode* lhs, const AStarNode* rhs) { return lhs->f < rhs->f; });

There is also the issue with TFS code that allows it to run up to 100 nodes. That should never happen. Which is why in the new code im working on this is the while loop everything is nested in

while (!openNodes.empty() || !found) {}

There is a lot of different things to consider and the main one is understanding how many nodes should be getting added, how many of those get checked, and the difference from our code to that.

Flatlander57:
> > I think all the things are you trying to add is already in TFS by default. If I am wrong, then let me know.
> 
> You are partially correct. The default code does do all of this. Where the problem comes in is:
> 
>     1. How many nodes are added that aren't needed
> 
> 
>     * When going in a straight line there should only be a line of nodes checked
> 
>     * I kind of half assed creating that in my code but it should be done by the algorithm itself
> 
> 
>     2. How the nodes are created and stored
> 
> 
>     * There are different ways of doing the same thing some are a lot faster than others
> 
> 
> One of the big problems is A* generally uses something like this to find the distance from the target:
> 
> > float getDistance(uint16_t x1, uint16_t y1, uint16_t x2, uint16_t y2) {
> > return sqrtf((x1 - x2)_(x1-x2)+(y1-y2)_(y1-y2));
> > }
> 
> Also, instead of looping through the nodes like this
> 
> > AStarNode* AStarNodes::getBestNode(const Position& targetPos)
> > {
> > if (curNode == 0) {
> > return nullptr;
> > }
> > int32_t diffNode = 0;
> > int32_t best_node_f = std::numeric_limits<int32_t>::max();
> > int32_t best_node = -1;
> > for (size_t i = 0; i < curNode; i++) {
> > const int_fast32_t Sx = nodes[0].x;
> > const int_fast32_t Sy = nodes[0].y;
> > const int_fast32_t Cx = nodes[i].x;
> > const int_fast32_t Cy = nodes[i].y;
> > int32_t SdiffX = std::abs(targetPos.x - Sx);
> > int32_t SdiffY = std::abs(targetPos.y - Sy);
> > int32_t NdiffX = std::abs(targetPos.x - Cx);
> > int32_t NdiffY = std::abs(targetPos.y - Cy);
> > diffNode = ((nodes[i].f + (((NdiffX - SdiffX) * 5) + ((NdiffY - SdiffY) * 5))) + (std::max(NdiffX, NdiffY) * 5));
> > if (openNodes[i] && diffNode < best_node_f) {
> > best_node_f = diffNode;
> > best_node = i;
> > }
> > }
> > if (best_node >= 0) {
> > return nodes + best_node;
> > }
> > return nullptr;
> > }
> 
> You can easily use something like this:
> 
> > openNodes.sort([](const AStarNode* lhs, const AStarNode* rhs) { return lhs->f < rhs->f; });
> 
> There is also the issue with TFS code that allows it to run up to 100 closed nodes. That should never happen. Which is why in the new code im working on this is the while loop everything is nested in
> 
> > while (!openNodes.empty() || !found) {}
> 
> There is a lot of different things to consider and the main one is understanding how many nodes should be getting added, how many of those get checked, and the difference from our code to that.

You are "kind of" correct that it should open nodes in a strait line towards the target.
But there are different costs to tiles (ground speed, field cost, push creature cost, diagonal move cost, etc), so you still need to check nearby nodes after finding a correct path just to make sure there isn't a cheaper path. You wouldn't want your pathing to choose a slow walking tile when there is a fast tile right next to it, or choose to walk over fire fields when it could easily avoid them, so even after finding a direct path you still need to check at least the adjacent tiles to see if there is a possible faster/cheaper path.

The limit on closed nodes is useful for very specific use cases. For example: The default TFS client (if you are on an open screen with no walls/doodads) will fail to reach the top corner of the screen when "using" the tile due to running out of nodes.

Speaking further on closed nodes... on my server, some creatures can path up to 50 sqm. Meaning a pack of wolves can "smell" you from 50 sqm away, and follow you offscreen waiting for you to get low health so it can attack you while you are weak. When pathing 50 sqm, you can easily close thousands of nodes trying to reach a target which takes way too much time, so you would want a limit where pathing could fail and the wolves would give up on following their target.

The sorting of the openNodes might be a good idea. I would be interested in if that was faster, because you would have to sort the openNodes every time before grabbing the fastest nodes. -- Is it faster to sort, or to loop, I haven't tested so I do not know.

gigastar:
> > > I think all the things are you trying to add is already in TFS by default. If I am wrong, then let me know.
> > 
> > 
> > You are partially correct. The default code does do all of this. Where the problem comes in is:
> > ```
> > 1. How many nodes are added that aren't needed
> > 
> > 
> > * When going in a straight line there should only be a line of nodes checked
> > 
> > * I kind of half assed creating that in my code but it should be done by the algorithm itself
> > 
> > 
> > 2. How the nodes are created and stored
> > 
> > 
> > * There are different ways of doing the same thing some are a lot faster than others
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > One of the big problems is A* generally uses something like this to find the distance from the target:
> > > float getDistance(uint16_t x1, uint16_t y1, uint16_t x2, uint16_t y2) {
> > > return sqrtf((x1 - x2)_(x1-x2)+(y1-y2)_(y1-y2));
> > > }
> > 
> > 
> > Also, instead of looping through the nodes like this
> > > AStarNode* AStarNodes::getBestNode(const Position& targetPos)
> > > {
> > > if (curNode == 0) {
> > > return nullptr;
> > > }
> > > int32_t diffNode = 0;
> > > int32_t best_node_f = std::numeric_limits<int32_t>::max();
> > > int32_t best_node = -1;
> > > for (size_t i = 0; i < curNode; i++) {
> > > const int_fast32_t Sx = nodes[0].x;
> > > const int_fast32_t Sy = nodes[0].y;
> > > const int_fast32_t Cx = nodes[i].x;
> > > const int_fast32_t Cy = nodes[i].y;
> > > int32_t SdiffX = std::abs(targetPos.x - Sx);
> > > int32_t SdiffY = std::abs(targetPos.y - Sy);
> > > int32_t NdiffX = std::abs(targetPos.x - Cx);
> > > int32_t NdiffY = std::abs(targetPos.y - Cy);
> > > diffNode = ((nodes[i].f + (((NdiffX - SdiffX) * 5) + ((NdiffY - SdiffY) * 5))) + (std::max(NdiffX, NdiffY) * 5));
> > > if (openNodes[i] && diffNode < best_node_f) {
> > > best_node_f = diffNode;
> > > best_node = i;
> > > }
> > > }
> > > if (best_node >= 0) {
> > > return nodes + best_node;
> > > }
> > > return nullptr;
> > > }
> > 
> > 
> > You can easily use something like this:
> > > openNodes.sort([](const AStarNode* lhs, const AStarNode* rhs) { return lhs->f < rhs->f; });
> > 
> > 
> > There is also the issue with TFS code that allows it to run up to 100 closed nodes. That should never happen. Which is why in the new code im working on this is the while loop everything is nested in
> > > while (!openNodes.empty() || !found) {}
> > 
> > 
> > There is a lot of different things to consider and the main one is understanding how many nodes should be getting added, how many of those get checked, and the difference from our code to that.
> 
> You are "kind of" correct that it should open nodes in a strait line towards the target. But there are different costs to tiles (ground speed, field cost, push creature cost, diagonal move cost, etc), so you still need to check nearby nodes after finding a correct path just to make sure there isn't a cheaper path. You wouldn't want your pathing to choose a slow walking tile when there is a fast tile right next to it, or choose to walk over fire fields when it could easily avoid them, so even after finding a direct path you still need to check at least the adjacent tiles to see if there is a possible faster/cheaper path.
> 
> The limit on closed nodes is useful for very specific use cases. For example: The default TFS client (if you are on an open screen with no walls/doodads) will fail to reach the top corner of the screen when "using" the tile due to running out of nodes.
> 
> Speaking further on closed nodes... on my server, some creatures can path up to 50 sqm. Meaning a pack of wolves can "smell" you from 50 sqm away, and follow you offscreen waiting for you to get low health so it can attack you while you are weak. When pathing 50 sqm, you can easily close thousands of nodes trying to reach a target which takes way too much time, so you would want a limit where pathing could fail and the wolves would give up on following their target.
> 
> The sorting of the openNodes might be a good idea. I would be interested in if that was faster, because you would have to sort the openNodes every time before grabbing the fastest nodes. -- Is it faster to sort, or to loop, I haven't tested so I do not know.

Sorting is a lot faster than doing a bunch of memory creations and comparisons.

this is how I stop paths from going to far
>// Don't update path if the target is too far away
>	if (Position::getDistanceX(startPos, endPos) > fpp.maxSearchDist || Position::getDistanceY(startPos, endPos) > >fpp.maxSearchDist) {
>		return false;
>}

This is how I have it set up right now... I am missing something but I can't figure it out for the life of me. West and South operations work perfectly. North and east break.

https://pastebin.com/ErkZwuL0

Flatlander57:
> > You are "kind of" correct that it should open nodes in a strait line towards the target. But there are different costs to tiles (ground speed, field cost, push creature cost, diagonal move cost, etc), so you still need to check nearby nodes after finding a correct path just to make sure there isn't a cheaper path. You wouldn't want your pathing to choose a slow walking tile when there is a fast tile right next to it, or choose to walk over fire fields when it could easily avoid them, so even after finding a direct path you still need to check at least the adjacent tiles to see if there is a possible faster/cheaper path.
> > The limit on closed nodes is useful for very specific use cases. For example: The default TFS client (if you are on an open screen with no walls/doodads) will fail to reach the top corner of the screen when "using" the tile due to running out of nodes.
> > Speaking further on closed nodes... on my server, some creatures can path up to 50 sqm. Meaning a pack of wolves can "smell" you from 50 sqm away, and follow you offscreen waiting for you to get low health so it can attack you while you are weak. When pathing 50 sqm, you can easily close thousands of nodes trying to reach a target which takes way too much time, so you would want a limit where pathing could fail and the wolves would give up on following their target.
> > The sorting of the openNodes might be a good idea. I would be interested in if that was faster, because you would have to sort the openNodes every time before grabbing the fastest nodes. -- Is it faster to sort, or to loop, I haven't tested so I do not know.
> 
> Sorting is a lot faster than doing a bunch of memory creations and comparisons.
> 
> this is how I stop paths from going to far
> 
> > // Don't update path if the target is too far away
> > if (Position::getDistanceX(startPos, endPos) > fpp.maxSearchDist || Position::getDistanceY(startPos, endPos) > >fpp.maxSearchDist) {
> > return false;
> > }
> 
> This is how I have it set up right now... I am missing something but I can't figure it out for the life of me. West and South operations work perfectly. North and east break.
> 
> https://pastebin.com/ErkZwuL0

Distance and closed nodes are 2 different metrics.
For example:
image

If you were trying to get from the Red Gem Tile to the Blue Gem Tile in this image, it would take long enough (even with your faster pathfinding) that you would probably want to make the pathfinding give up before it reached the destination.

gigastar:
True, in that case I would say if the nodes being checked are too far then stop the code. Then you can atleast stop it before it checks 100 nodes....

So after doing some more research I found that this code is the most efficient and results in the same thing as the default code for TFS. I thought the path was incorrect but it is actually the best path. Maybe someone can show me an example where dijkstras code grabs a better path than A*.

dijkstra
AStar

gigastar:
Final benchmarks: https://pastebin.com/3wGUdWzm

src/map.cpp
```diff
@@ -504,20 +505,24 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
...
+		return false;
+	}
+
 	return !tile->hasProperty(CONST_PROP_BLOCKPROJECTILE);
 }
 
 namespace {
 
-bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
+bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool isPathfinding /*= false*/)
```
Zbizu, line 517: we add ` /*= false*/` when default value is defined as false in header file (.h)

this function has no declaration in .h file, therefore no default value, therefore ` /*= false*/` should be removed

src/map.cpp
```diff
@@ -526,15 +531,15 @@ bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
 	return true;
 }
 
-bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
+bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool isPathfinding /*= false*/)
```
Zbizu, line 534: same here

src/map.cpp
```diff
@@ -545,27 +550,29 @@ bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
...
-		return checkSlightLine(x1, y1, x0, y0, z);
+		return checkSlightLine(x1, y1, x0, y0, z, isPathfinding);
 	}
 
-	return checkSlightLine(x0, y0, x1, y1, z);
+	return checkSlightLine(x0, y0, x1, y1, z, isPathfinding);
 }
 
-bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/) const
+bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/,
```
Zbizu, line 574: this one is correct because...

src/map.h
```diff
@@ -236,12 +243,14 @@ class Map
 	 *Destination point \param sameFloor checks if the destination is on same
 	 *floor \returns The result if there is no obstacles
 	 */
-	bool isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor = false) const;
-	bool checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const;
+	bool isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor = false,
+	                  bool isPathfinding = false) const;
```
Zbizu, line 247: ... false is declared here as default value

Zbizu (changes requested):

ArturKnopik:
I applied this fix and this one TFS https://github.com/otland/forgottenserver/pull/4205 to my server based TFS(no changes in A* before this fix)
The result: https://www.youtube.com/watch?v=bevXfAApvzE&ab_channel=Krecikondexin

gigastar:
> I applied this fix and this one #4205 to my server based TFS(no changes in A* before this fix) The result: https://www.youtube.com/watch?v=bevXfAApvzE&ab_channel=Krecikondexin

I am on it. Nothing will stop me from getting this system in TFS xD

Thank you zbizu

gigastar:
> I applied this fix and this one #4205 to my server based TFS(no changes in A* before this fix) The result: https://www.youtube.com/watch?v=bevXfAApvzE&ab_channel=Krecikondexin

So I did notice that this same thing happens on real tibia. I haven't played in a couple months so it may have changed. Can you make sure this is actually a problem?

gigastar:
So, one thing we can add to make monsters act a little more normal would be to add a modifier to the algorithm based on which distance is further of X and Y. It will cause the monster to walk a path that is less straight. It does cost some performance but not very much. Even with the modifier it is almost just as fast and still a lot faster than default.

This is what the modifier would look like:
The 10 is the walkCost of walking to a straight tile.

```
if (Position::getDistanceX(startPos, targetPos) > Position::getDistanceY(startPos, targetPos)) {
				if (Position::getDistanceX(pos, startPos) < Position::getDistanceX(startPos, targetPos)) {
					f = f - 10;
				}
			}
			if (Position::getDistanceY(startPos, targetPos) > Position::getDistanceY(startPos, targetPos)) {
				if (Position::getDistanceY(pos, startPos) < Position::getDistanceY(startPos, targetPos)) {
					f = f - 10;
				}
			}
```

Tofame:
https://user-images.githubusercontent.com/80583148/189544376-59a30005-7dc8-40fe-bc77-877fa40bf5e7.mp4

Wolf prefers going around instead of diagonally?

gigastar:
> AAAFilmikOBS_2022-09-11.20-52-45.mp4 
> Wolf prefers going around instead of diagonally?

Yes, I have been working on something for this. You can fix it by changing
>const int_fast32_t f = n->f + distEnd + distStart + (walkCost + speedCost);

but it is a huge cost to performance as it changes it to be more like Dijkstra's again.
I will figure something out if no one else does soon enough.

gigastar:
So there seems to be a problem with the summons. For some reason they stop following after about 5sqm between the master and summon. I am working on a solution.

gigastar:
Wanted to do a benchmark with the recent changes. Here is the results.
https://pastebin.com/jEMxGa9r

Tofame:
broken on tfs 1.4.2, monster is all the time trying to "keep distance" even though he shouldn't so in result he approaches player then thinks - "I should keep distance" goes away and then comes back.... And this is looped behaviour.

floki21uu:
> broken on tfs 1.4.2, monster is all the time trying to "keep distance" even though he shouldn't so in result he approaches player then thinks - "I should keep distance" goes away and then comes back.... And this is looped behaviour.

I do agree.
Here GIF:
brok

PR #4203, opened 2022-08-29
use "set outfit" window for /looktype

Zbizu:
### Pull Request Prelude

### Changes Proposed
/looktype now sends outfit window
/looktype param now sends outfit window with a single outfit
obraz
obraz

### DIFF
--- a/data/talkactions/scripts/looktype.lua
+++ b/data/talkactions/scripts/looktype.lua
@@ -1,4 +1,3 @@
--- keep it ordered
 local invalidTypes = {
 	1, 135, 161, 162, 163, 164, 165, 166, 167, 168, 169, 170, 171, 172, 173,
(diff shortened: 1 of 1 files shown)


Zbizu:
small note: this change targets 12.9, will crash earlier clients because it sends outfits up to gnomprona

soul4soul:
Cool change but it should only send the outfit window when an outfit id isn't passed. I have used this for quick testing of outfit ids and it would be annoying to get that dialog all the time.

data/talkactions/scripts/looktype.lua
```diff
@@ -1,4 +1,3 @@
--- keep it ordered
```
ranisalt, line 1: Doesn't need to keep _sorted_ anymore?

omarcopires:
This doesn't work, it just crashes the client.

PR #4205, opened 2022-08-31
Update pathfinding call

NRH-AA:
### Pull Request Prelude
Monsters update follow path only when onThink() occurs. This causes a lag that makes monsters easy to get away from.
This pull will make them update their path whenever the path they need to get to changes.

### Changes Proposed
Remove path update onThink()
Add a check to when the creature walks to see if it needs to update its path.
If the position it needs to get to hasn't changed it will not update its path.

Watch path update behavior here:
https://youtu.be/TYV4qFLHr-Q

Creature Keep Distance:
https://youtu.be/2FkE23HkhS0

**Issues addressed:** 

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -218,11 +218,6 @@ void Creature::onWalk()
 		eventWalk = 0;
 		addEventWalk();
 	}
-
-	if (!hasFollowPath && !followCreature && !attackedCreature || attackedCreature != followCreature) {
-		goToFollowCreature();
-		return;
-	}
 }
 
 void Creature::onWalk(Direction& dir)
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -137,10 +137,7 @@ void Creature::onThink(uint32_t interval)
 		blockTicks = 0;
 	}
 
-	if (forceUpdatePath) {
-		g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
-		forceUpdatePath = false;
-	} else {
+	if (!forceUpdatePath) {
 		forceUpdatePath = true;
 	}
 
@@ -213,7 +210,7 @@ void Creature::onWalk()
 
 	if (attackedCreature) {
 		const Position& targetPos = attackedCreature->getPosition();
-		if (followPosition != targetPos) {
+		if (followPosition != targetPos || forceUpdatePath) {
 
 			FindPathParams fpp;
 			getPathSearchParams(attackedCreature, fpp);
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -137,6 +137,13 @@ void Creature::onThink(uint32_t interval)
 		blockTicks = 0;
 	}
 
+	if (forceUpdatePath) {
+		g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
+		forceUpdatePath = false;
+	} else {
+		forceUpdatePath = true;
+	}
+
 	// scripting event - onThink
 	const CreatureEventList& thinkEvents = getCreatureEvents(CREATURE_EVENT_THINK);
 	for (CreatureEvent* thinkEvent : thinkEvents) {
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -199,7 +199,7 @@ void Creature::onWalk()
 		addEventWalk();
 	}
 
-	if (followCreature) {
+	if (followCreature && !attackedCreature || attackedCreature != followCreature) {
 		listWalkDir.clear();
 		goToFollowCreature();
 		return;
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -153,6 +153,7 @@ void Creature::onThink(uint32_t interval)
 void Creature::checkPath()
 {
 	if (followCreature && !attackedCreature && followPosition != followCreature->getPosition()) {
+		followPosition = followCreature->getPosition();
 		goToFollowCreature();
 		return;
 	}
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -216,7 +216,8 @@ void Creature::onWalk()
 			FindPathParams fpp;
 			getPathSearchParams(attackedCreature, fpp);
 			const Position& pos = getPosition();
-			if (Position::getDistanceX(pos, targetPos) > fpp.maxSearchDist || Position::getDistanceY(pos, targetPos) > fpp.maxSearchDist) {
+			if (Position::getDistanceX(pos, targetPos) > fpp.maxSearchDist ||
+	                    Position::getDistanceY(pos, targetPos) > fpp.maxSearchDist) {
 				// Attacked Creature has gone too far. Stop trying to get a path.
 				listWalkDir.clear();
 				return;
@@ -926,12 +927,13 @@ bool Creature::setAttackedCreature(Creature* creature)
 			attackedCreature = nullptr;
 			return false;
 		}
-		
+
 		// Target is too far to get a path too. We shouldn't add him as a target.
 		FindPathParams fpp;
 		getPathSearchParams(creature, fpp);
 		const Position& pos = getPosition();
-		if (Position::getDistanceX(pos, creaturePos) > fpp.maxSearchDist || Position::getDistanceY(pos, creaturePos) > fpp.maxSearchDist) {
+		if (Position::getDistanceX(pos, creaturePos) > fpp.maxSearchDist ||
+                    Position::getDistanceY(pos, creaturePos) > fpp.maxSearchDist) {
 			attackedCreature = nullptr;
 			return false;
 		}
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -924,16 +924,6 @@ bool Creature::setAttackedCreature(Creature* creature)
 			return false;
 		}
 
-		// Target is too far to get a path too. We shouldn't add him as a target.
-		FindPathParams fpp;
-		getPathSearchParams(creature, fpp);
-		const Position& pos = getPosition();
-		if (Position::getDistanceX(pos, creaturePos) > fpp.maxSearchDist ||
-		    Position::getDistanceY(pos, creaturePos) > fpp.maxSearchDist) {
-			attackedCreature = nullptr;
-			return false;
-		}
-
 		attackedCreature = creature;
 		followPosition = creaturePos;
 		onAttackedCreature(attackedCreature);
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -217,7 +217,7 @@ void Creature::onWalk()
 			getPathSearchParams(attackedCreature, fpp);
 			const Position& pos = getPosition();
 			if (Position::getDistanceX(pos, targetPos) > fpp.maxSearchDist ||
-	                    Position::getDistanceY(pos, targetPos) > fpp.maxSearchDist) {
+			    Position::getDistanceY(pos, targetPos) > fpp.maxSearchDist) {
 				// Attacked Creature has gone too far. Stop trying to get a path.
 				listWalkDir.clear();
 				return;
@@ -933,7 +933,7 @@ bool Creature::setAttackedCreature(Creature* creature)
 		getPathSearchParams(creature, fpp);
 		const Position& pos = getPosition();
 		if (Position::getDistanceX(pos, creaturePos) > fpp.maxSearchDist ||
-                    Position::getDistanceY(pos, creaturePos) > fpp.maxSearchDist) {
+		    Position::getDistanceY(pos, creaturePos) > fpp.maxSearchDist) {
 			attackedCreature = nullptr;
 			return false;
 		}
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -136,6 +136,10 @@ void Creature::onThink(uint32_t interval)
 		blockCount = std::min<uint32_t>(blockCount + 1, 2);
 		blockTicks = 0;
 	}
+	
+	if (!forceUpdateFollowPath) {
+		forceUpdateFollowPath = true;
+	}
 
 	// scripting event - onThink
 	const CreatureEventList& thinkEvents = getCreatureEvents(CREATURE_EVENT_THINK);
@@ -148,7 +152,9 @@ void Creature::checkPath()
 {
 	if (attackedCreature) {
 		const Position& targetPos = attackedCreature->getPosition();
-		if (followPosition != targetPos) {
+		if (followPosition != targetPos || forceUpdateFollowPath) {
+			forceUpdateFollowPath = false;
+
 			FindPathParams fpp;
 			getPathSearchParams(attackedCreature, fpp);
 			const Position& pos = getPosition();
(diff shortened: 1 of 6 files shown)


src/creature.cpp
```diff
@@ -214,6 +198,31 @@ void Creature::onWalk()
...
+			const Position& pos = getPosition();
+			if (Position::getDistanceX(pos, targetPos) > fpp.maxSearchDist || Position::getDistanceY(pos, targetPos) > fpp.maxSearchDist) {
+				// Attacked Creature has gone too far. Stop trying to get a path.
+				listWalkDir.clear();
+				return;
+			}
+
+			followPosition = attackedCreature->getPosition();
+			listWalkDir.clear();
+			g_dispatcher.addTask(createTask(std::bind(&Game::updateCreatureWalk, &g_game, getID())));
```
ranisalt, line 223: Don't use bind, use a lambda instead.

ranisalt (changes requested):

src/creature.cpp
Zbizu: `g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));`

Zbizu:
there is one more problem: monsters stop when you walk to them, they hit you and you run away
expected behaviour: black square will lit up and monster will start walking instantly

I tried to add `updateCreatureWalk` at the end of doAttacking, but it didn't change anything

gigastar:
> there is one more problem: monsters stop when you walk to them, they hit you and you run away expected behaviour: black square will lit up and monster will start chasing instantly observed: monster stands still until he received new instruction from onThink/checkCreatures
> 
> I tried to add `updateCreatureWalk` at the end of `doAttacking`, but it didn't change anything

What monster are you trying this on?
I don't completely understand the problem.

Should the black square stay on?
Is the monster not following fast enough? when you run away?

Testing some things I noticed when you first walk into the range of the monster
If you run into him he has a delay before he starts walking to you.

It also doesn't show the black square until he walks up and  hits you (that is with non-ranged monsters)
Ranged monsters will show the black square whenever they shoot at you.

Zbizu:
@gigastar I mean melee monsters.

expected:
https://user-images.githubusercontent.com/5205079/188333241-abb9d4d8-e9b9-40ed-8432-3ef8efd7302b.mp4

observed:
https://user-images.githubusercontent.com/5205079/188333264-421899ae-eaab-4962-9dd9-ecf6733790d1.mp4

gigastar:
> @gigastar I mean melee monsters.
> 
> expected:
> 
>  2022-09-04.22-50-06.mp4 
> observed:
> 
>  2022-09-04.22-57-56.mp4

So it looks like they are missing swings maybe? Its still hard to see xD. Can you slow down your GM character so its closer to the same thing

gigastar:
@Zbizu I am pretty sure with these last changes it was fixed. I can't see the issue on my side.

Zbizu:
@gigastar the issue is that monsters stand still for a moment (they shouldn't)

updating monster direction should make them start walking towards target immediately

gigastar:
> ```c++
> g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
> ```

Is this something that didn't happen before? That will help me find a solution. If it was a bug already I think it is linked to something else.

Zbizu:
> I found a solution. The only thing im wondering is... If the monsters turn a direction, but need to turn again before facing the correct way. On real tibia do they walk anyway or do they take an extra second to turn the correct way, then start walking?

it is very hard to make a monster stand still in rl. I only managed to do it once (beginning of first recording, wolf north of my char)

gigastar:
> > I found a solution. The only thing im wondering is... If the monsters turn a direction, but need to turn again before facing the correct way. On real tibia do they walk anyway or do they take an extra second to turn the correct way, then start walking?
> 
> it is very hard to make a monster stand still in rl by walking to him and running away. I only managed to do it once (beginning of first recording, wolf north of my char), as you can see the wolf takes 0.5-1 sec pause at best. Tfs monsters take 2 sec and they do it far more often.
> 
> sry for no review and not responding to whole comment, it's late and I'm going to sleep

With the latest changes I think it is fixed. The only problem right now is:
1) You are to the right of monster
2) You run to the left of it
3) The monster turns north because it got a turn signal right at that time
4) He takes an extra second to turn left and start walking

The only way to fix that is to modify how fast the monster turns, or modify the actual turning. It could be off from what RL tibia is.

gigastar:
I can't find what is causing it...

gigastar:
After some testing it is actually better updating the paths this way. It also fixes the problem zbizu showed.

It was less cpu expensive because instead of updating onWalk where you can have 1000 monsters walking at the same time pushing all those path updates over and over it is a set time. I tested this with at least 300 monsters chasing me and did see a better result.

src/creature.cpp
```diff
@@ -136,6 +136,10 @@ void Creature::onThink(uint32_t interval)
 		blockCount = std::min<uint32_t>(blockCount + 1, 2);
 		blockTicks = 0;
 	}
+	
+	if (!forceUpdateFollowPath) {
+		forceUpdateFollowPath = true;
+	}
```
ranisalt, line 142: Just set it to true directly

src/creature.cpp
```diff
@@ -148,7 +152,9 @@ void Creature::checkPath()
 {
 	if (attackedCreature) {
 		const Position& targetPos = attackedCreature->getPosition();
-		if (followPosition != targetPos) {
+		if (followPosition != targetPos || forceUpdateFollowPath) {
+			forceUpdateFollowPath = false;
+
```
ranisalt, line 157: This will always set the value to `false`:

| followPosition != targetPos | forceUpdateFollowPath | output |
|-|-|-|
| true | true | false |
| true | false | false |
| false | true | false |
| false | false | false |

ranisalt (changes requested):

src/creature.cpp
ranisalt: The comment ended up on the wrong side of the diff

src/creature.cpp
gigastar: ? it should do:

if (followPosition != targetPos) : execute
if (forceUpdateFollowPath) : execute

We always set forceUpdateFollowPath = false because we shouldn't need to update again no matter what.
Other areas in the sources can set forceUpdateFollowPath to true, that's the only reason its there.

Zbizu:
I forgot to answer this one earlier:
> Is this something that didn't happen before? That will help me find a solution. If it was a bug already I think it is linked to something else.

It was bugged in tfs for years. I just thought it to be worth mentioning here because it is very closely related to your changes.
Seems to be working better now. Monsters still get stuck for a while occasionally, but for shorter durations now.

Zbizu (approved):

src/creature.cpp
ranisalt: @gigastar but the current code is setting to false in every possible case (check the truth table). Might as well remove the if.

src/game.cpp
```diff
@@ -3913,6 +3914,23 @@ void Game::checkCreatures(size_t index)
 	cleanup();
 }
 
+void Game::checkCreaturesPath(size_t index)
+{
+	g_scheduler.addEvent(createSchedulerTask(EVENT_CREATURE_PATH_INTERVAL, std::bind(&Game::checkCreaturesPath, this, (index + 1) % EVENT_CREATURECOUNT)));
```
ranisalt, line 3919: Lambda

src/game.cpp
```diff
@@ -78,6 +78,7 @@ void Game::start(ServiceManager* manager)
 		g_scheduler.addEvent(createSchedulerTask(EVENT_LIGHTINTERVAL, [this]() { checkLight(); }));
 	}
 	g_scheduler.addEvent(createSchedulerTask(EVENT_CREATURE_THINK_INTERVAL, [this]() { checkCreatures(0); }));
+	g_scheduler.addEvent(createSchedulerTask(EVENT_CREATURE_PATH_INTERVAL, std::bind(&Game::checkCreaturesPath, this, 0)));
```
ranisalt, line 81: Lambda

ranisalt (changes requested):

src/creature.cpp
gigastar: @ranisalt  I am not understanding you man.

The code works as intended. It executes if the player moves or after the onThink event sets forceUpdatePath = true.

The idea is to make sure its not constantly updating the path. Trying to save as much CPU, memory, ect. as possible.

As long as the player doesnt move, the path should only update after the onThink events sets forceUpdatePath.
If the player moves it will read the position is different and execute.

If targetPos changed OR forceUpdateFollowPath is true then
     updateThePath

if targetPos is the same > check if forceupdateFollowPath is true
     -forceUpdateFollowPath is true : execute
     -forceUpdateFollowPath is false : dont execute

We don't want it to update constantly. If we remove the forceUpdateFollowPath check then when players don't move and a monster already has a path, but that path gets blocked. They won't be able to find their way.

There is a misunderstanding here.

gigastar:
@Zbizu
If you change
> static constexpr int32_t EVENT_CREATURE_PATH_INTERVAL = 20;
to 10 the monsters will be even more responsive when turning. You can play with that depending on how powerful your server is to modify how fast the pathing happens. I think that's good yeah?

gigastar:
This was an issue brought by ralke:

>Hi there! I've just updated everything, saw this post and started the testings. I have found a little issue already, when I use >utevo res "any creature, the creature doesn't follow the player. This started to happen, after applied the commits, is it related >to the latest changes? Regards!

>
>Edit:
>Found another one, when I click on a stair from far distance (with right click, to use it), it says "There is no way". The same >thing happens when I try to use any object from far distance... Here you can see how I applied the whole code.

Both issues were addressed in the last few commits.

src/creature.cpp
ranisalt: My bad. I opened this page several times and it was displaying this:

```
		if (followPosition != targetPos || forceUpdateFollowPath) {
			forceUpdateFollowPath = false;
		}
```

Now I see that it has more within the block, idk what was happening.

gigastar: It's probably mad at me for all the commits at this point.

ranisalt:
Just want to add something super weird that Cipsoft Tibia has (I think it is a regression and should not be considered).

A monster will not walk to the player if the distance to cover is larger than half the viewport size (i.e. distance between the center and the edge). In this video, the distance the Orc Warlord would span over 7 sqm, so it simply did not attack me until it was close enough.

[Screencast from 09-08-2022 11:03:52 AM.webm](https://user-images.githubusercontent.com/1993083/189325155-1ea8f8a0-feb0-452a-8a22-28ec7c6221a2.webm)

gigastar:
> Just want to add something super weird that Cipsoft Tibia has (I think it is a regression and should not be considered).
> 
> A monster will not walk to the player if the distance to cover is larger than half the viewport size (i.e. distance between the center and the edge). In this video, the distance the Orc Warlord would span over 7 sqm, so it simply did not attack me until it was close enough.
> 
>  Screencast.from.09-08-2022.11.03.52.AM.webm

That probably is part of their system to suppress pathfinding calls mixed with making monsters/situations a little easier for the kids.

src/creature.cpp
```diff
@@ -138,26 +138,40 @@ void Creature::onThink(uint32_t interval)
...
 	// scripting event - onThink
 	const CreatureEventList& thinkEvents = getCreatureEvents(CREATURE_EVENT_THINK);
 	for (CreatureEvent* thinkEvent : thinkEvents) {
 		thinkEvent->executeOnThink(this, interval);
 	}
 }
 
+void Creature::checkPath()
+{
+	if (attackedCreature) {
```
4mrcn4, line 155: why nest if you can return when `!attackedCreature`

src/creature.cpp
```diff
@@ -138,26 +138,40 @@ void Creature::onThink(uint32_t interval)
...
+		const Position& targetPos = attackedCreature->getPosition();
+		if (followPosition != targetPos || forceUpdateFollowPath) {
```
4mrcn4, line 157: same here (nesting)

4mrcn4:
Cool effort, keep it up!
Can we see some benchmarks that will highlight the improvements?

src/game.cpp
```diff
@@ -3913,6 +3914,24 @@ void Game::checkCreatures(size_t index)
...
+
+	auto& checkCreatureList = checkCreatureLists[index];
+	auto it = checkCreatureList.begin(), end = checkCreatureList.end();
+	while (it != end) {
+		Creature* creature = *it;
+		if (creature->getHealth() > 0) {
+			creature->checkPath();
+		}
+		++it;
+	}
```
ranisalt, line 3930: Why not using a range-for here?

```
	for (Creature* creature : checkCreatureLists[index]) {
		if (creature->getHealth() > 0) {
			creature->checkPath();
		}
	}
```

gigastar: Oversight

gigastar: Would we still need the cleanup?

Tofame:
Im not sure where should I leave this comment here, or on the other thread of the pathfinding. And also I'm not sure if this was intended or not, but thinking that maybe it wasn't I will post. 
Summon follows his master while always keeping 1 distance. However, if the master approaches him he stays still. Shouldn't he back off 1 cell?
[https://streamable.com/c0ycwk](url)

Zbizu:
> Im not sure where should I leave this comment here, or on the other thread of the pathfinding. And also I'm not sure if this was intended or not, but thinking that maybe it wasn't I will post. Summon follows his master while always keeping 1 distance. However, if the master approaches him he stays still. Shouldn't he back off 1 cell?

You're correct. I have a code working like this on a public repo, but it's not compatible with the one in pr and I don't remember the commits I made anymore.

gigastar:
> Im not sure where should I leave this comment here, or on the other thread of the pathfinding. And also I'm not sure if this was intended or not, but thinking that maybe it wasn't I will post. Summon follows his master while always keeping 1 distance. However, if the master approaches him he stays still. Shouldn't he back off 1 cell?



> > Im not sure where should I leave this comment here, or on the other thread of the pathfinding. And also I'm not sure if this was intended or not, but thinking that maybe it wasn't I will post. Summon follows his master while always keeping 1 distance. However, if the master approaches him he stays still. Shouldn't he back off 1 cell?
> 
> You're correct. I have a code working like this on a public repo, but it's not compatible with the one in this pr and I don't remember the commits I made anymore.
> 
> @ranisalt my guess is they're using dijkstra algo with range limits also either the monster path has reset when you moved or changing walk direction has extra cost

Fixed

gigastar:
https://youtu.be/YuKYgnYujAo

gigastar:
I have been working on updating all the code for this. A lot of the code and be changed and removed. Paths only update if the last path is no longer possible so we don't need any extra code for getting paths that don't use the A* path finding. With the changes to A* it essentially costs nothing 90% of the time and costs a fraction at the more complex path ends.

We can use A* as a bool to tell us if we should be doing a dance step or not as well as a random step if we need it.

soul4soul:
> https://youtu.be/YuKYgnYujAo @Zbizu

When summons stand diagonal they are closer as  just as Zbizu explained.

diag

emil92b:
summon follow pattern:

PR #4207, opened 2022-09-08
Split Lua script interface into modules

### LINKED ISSUE #2169: Split up luascript.cpp
This file is waaaaaay to large and actually has started causing travis builds to fail in my other PRs due to OOM. We need to come up with a convention of placing lua interface functions because the current one is leading to absurd build requirements.

### LINKED ISSUE #3504: A potential solution to luascript.cpp length
This doesn't quite fit to any of the issue templates so I'll try to improvise.

## The problem
luascript.cpp is too long and is causing some (CI?) tools to fail

## Suggested solution
keeping lines such as this one: `registerClass("Creature", "", LuaScriptInterface::luaCreatureCreate);` to maintain the self-documenting style of code

move functions such as this one: `static int luaCreatureCreate(lua_State* L);` to a new file in a separate directory to reduce `luascript.cpp` size

## Example for Creature class
in luascript.cpp:
`#include luascript/creature.h`
and the creature cpp/h would be located in:
- src/luascript/creature.cpp
- src/luascript/creature.h

ranisalt:
### Pull Request Prelude

### Changes Proposed

Splits Lua script interface and its notoriously humongous header file into several modules that implement the Lua interfaces for a single header.

This reduces the amount of code that needs to be recompiled when said header changes, and splits compilation of Lua interfaces into several compilation units, which may be compiled in parallel.

Additionally, since these callbacks are not exported via a header file, the compiler can apply more aggressive optimizations given that the functions are strictly local to the compilation units.

As a nice-to-have, I refactored the interface to allow self-registering of modules, such that adding a new module does not require recompiling everything, just compiling the module and linking it against the server.

**Issues addressed:**
Fixes #2169
Closes #3504 

**Pending:**
There are still some functions being registered in `luascript.cpp`. While it's not so many, and they are quite short, I still want to move them to another file to keep it clean.

Co-Authored-By: Mario <Shadow280690@hotmail.de>


Zbizu:
It's about time. Adding more code to luascript.cpp begun to hang visual studio for several seconds. Also looks cleaner than mine (pic below) looking forward to it.
obraz

On a side note: please avoid creating too many directories. Canary did it and it's painful to navigate through their files now.

Erza:
>tfs::lua::reserveScriptEnv

Needs more namespaces, how about something like `tfs::lua::scriptEnv::reserve`?
Jokes aside, that `tfs::lua` namespace is probably one of the worst named things I've seen in a while, sucks to look at and type out. Why not just name it `Lua::reserveScriptEnv`?

>On a side note: please avoid creating too many directories. Canary did it and it's painful to navigate through their files now.

That's true

ranisalt:
> Needs more namespaces, how about something like `tfs::lua::scriptEnv::reserve`?
> Jokes aside, that `tfs::lua` namespace is probably one of the worst named things I've seen in a while, sucks to look at and type out. Why not just name it `Lua::reserveScriptEnv`?

Ideally, everything would live under the `tfs` namespace, so it would be used as `lua::` in my perfect world :sweat_smile: that's one of the things I wanted feedback on, thanks!

Erza:
> > Needs more namespaces, how about something like `tfs::lua::scriptEnv::reserve`?
> > Jokes aside, that `tfs::lua` namespace is probably one of the worst named things I've seen in a while, sucks to look at and type out. Why not just name it `Lua::reserveScriptEnv`?
> 
> Ideally, everything would live under the `tfs` namespace, so it would be used as `lua::` in my perfect world 😅 that's one of the things I wanted feedback on, thanks!

I don't see the point in even having the `tfs` namespace, it's so redundant, why would we need it?

ranisalt:
> it's so redundant, why would we need it?

To separate things if TFS is used as a library in another project, i.e. to allow using map loading, database handling in another project.

More of a nice-to-have than anything else

conde2:
This is just to reduce the compiling time or is there any more benefits to it?
Because I really like having all the lua API in a single file, whenever I need to look for a function I just need to open a single file and search in it. 

For example tile and positions functions are very often used togheter and I can easily visualize the options I have to solve whatever I'm coding. If this only brings some extra few seconds to compiling time, I'm not a huge fan of this change.

src/spells.cpp
```diff
@@ -1024,22 +1026,24 @@ bool InstantSpell::internalCastSpell(Creature* creature, const LuaVariant& var)
 bool InstantSpell::executeCastSpell(Creature* creature, const LuaVariant& var)
 {
 	// onCastSpell(creature, var)
-	if (!tfs::lua::reserveScriptEnv()) {
+	using namespace tfs;
```
Zbizu, line 1029: wouldn't it be better to declare the namespace on top of the file just once?

ranisalt:
Rebased against master and fixed @Zbizu comment

> This is just to reduce the compiling time or is there any more benefits to it?

It is mostly that, it helps by a huge margin being able to compile these files in parallel instead of one massive 30k+ lines chunk at once.

> Because I really like having all the lua API in a single file, whenever I need to look for a function I just need to open a single file and search in it.

This really should be fixed by having proper documentation, not putting everything in a single file. I agree that documentation is lacking currently.

Kamenuvol:
I'd vote to use otclient LuaEngine and declare the functions in its own class (most are just calling them) and only register in luascript.lua

ranisalt:
> I'd vote to use otclient LuaEngine and declare the functions in its own class (most are just calling them) and only register in luascript.lua

That would be super cool, if it was extracted into its own library.

conde2:
There is already a exported binder that @edubart create based on OTClient take a look: 

https://github.com/edubart/euluna-binder

gigastar:
@ranisalt I downloaded this branch and tried to compile it but got errors. Is there anything that needs to be downloaded through vcpkg or anything?

ranisalt:
@gigastar no, what error?

gigastar:
> @gigastar no, what error?

show

MillhioreBT:
I have the same error with the visual studio 2019 compiler.

Btw, I also like what Erza said, it would be better to call a single namespace, in this case Lua or lua

ranisalt:
Feel free to add VS support, I can not do that as previously said.

ranisalt:
Re-synced to master

EPuncker (dismissed):

ranisalt:
Added files to vc project, but can't test if it works for obvious reasons.

omarcopires:
the log is in Portuguese, but I believe it is not a problem for you for obvious reasons xD

https://pastebin.com/mbXemwbW

ramon-bernardo:
Rebase?

ranisalt:
@ramon-bernardo ain't no way I'm rebasing it :laughing: I prefer if we move as much as possible to Lua, starting with #4764 and then the other classes that don't belong to the core.

ramon-bernardo:
> @ramon-bernardo ain't no way I'm rebasing it 😆 I prefer if we move as much as possible to Lua, starting with #4764 and then the other classes that don't belong to the core.

I'm fine with that. Could we split it into parts? I really want to test this with Visual Studio.

PR #4212, opened 2022-09-21
Fixed argument order on luaDoAreaCombat

EPuncker:
Co-Authored-By: Zbizu <Zbizu@users.noreply.github.com>

### Pull Request Prelude

### Changes Proposed
Fixed argument order from doAreaCombat script command:

`doAreaCombat(cid = 1, type = 2, pos = 3, area = 4, min = 5, max = 6, effect = 7, origin = 8, blockArmor = 9, blockShield = 10, ignoreResistances = 11)`

as you can check above, `blockedByArmor`, `min` and `max` were wrong

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -3666,14 +3666,15 @@ int LuaScriptInterface::luaDoAreaCombat(lua_State* L)
 		CombatParams params;
 		params.combatType = combatType;
 		params.impactEffect = getNumber<uint8_t>(L, 7);
-		params.blockedByArmor = getBoolean(L, 8, false);
-		params.blockedByShield = getBoolean(L, 9, false);
-		params.ignoreResistances = getBoolean(L, 10, false);
+
+		params.blockedByArmor = getBoolean(L, 9, false);
+		params.blockedByShield = getBoolean(L, 10, false);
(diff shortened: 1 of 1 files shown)


ranisalt:
If nobody noticed that before, is anyone even using this function? Can we move it to compat and use a `Combat` instance instead?

EPuncker:
> If nobody noticed that before, is anyone even using this function? Can we move it to compat and use a `Combat` instance instead?

`blockedByArmor` was added quite recently by @nekiro, and the order of `min` and `max` doesn't make any difference at all I suppose

ramon-bernardo (approved):

src/luascript.cpp
```diff
@@ -3666,14 +3666,15 @@ int LuaScriptInterface::luaDoAreaCombat(lua_State* L)
 		CombatParams params;
 		params.combatType = combatType;
 		params.impactEffect = getNumber<uint8_t>(L, 7);
-		params.blockedByArmor = getBoolean(L, 8, false);
-		params.blockedByShield = getBoolean(L, 9, false);
-		params.ignoreResistances = getBoolean(L, 10, false);
+
```
ramon-bernardo, line 3669:
```suggestion
		// getNumber<CombatOrigin>(L, 8);
```

PR #4214, opened 2022-09-21
prevent double call to equipFunction inside MoveEvent::fireEquip

yamaken93:
### Pull Request Prelude

### Changes Proposed

prevent double call to the equipFunction when equipFunction does not return RETURNVALUE_NOERROR.

### DIFF
--- a/src/movement.cpp
+++ b/src/movement.cpp
@@ -998,12 +998,8 @@ ReturnValue MoveEvent::fireEquip(Player* player, Item* item, slots_t slot, bool 
 	if (equipFunction) {
 		ret = equipFunction(this, player, item, slot, isCheck);
 	}
-	if (scripted) {
-		if (ret == RETURNVALUE_NOERROR) {
-			if (!executeEquip(player, item, slot, isCheck)) {
-				ret = RETURNVALUE_CANNOTBEDRESSED;
-			}
-		}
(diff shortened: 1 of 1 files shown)


src/movement.cpp
```diff
@@ -994,16 +994,18 @@ bool MoveEvent::executeStep(Creature* creature, Item* item, const Position& pos)
 
 ReturnValue MoveEvent::fireEquip(Player* player, Item* item, slots_t slot, bool isCheck)
 {
+	ReturnValue ret = RETURNVALUE_NOERROR;
+	if (equipFunction) {
```
nekiro, line 998:
```C++
if (equipFunction) {
    ret = equipFunction(this, player, item, slot, isCheck);
}

if (scripted && ret == RETURNVALUE_NOERROR && !executeEquip(player, item, slot, isCheck)) {
    ret = RETURNVALUE_CANNOTBEDRESSED;
}
```

yamaken93: I thought the same (adding a () around ```ret == RETURNVALUE_NOERROR```) but maybe 3 checks in the same is not aesthetics or debuggable enough...

EvilHero90: suggestion looks fine, even tho it's 3 checks it's readable

EPuncker (approved):

PR #4216, opened 2022-09-23
Monster Weapon weakness

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Add ability to make monsters weak to certain weapons. The same as elements.

**Issues addressed:** 
None

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -14418,9 +14418,9 @@ int LuaScriptInterface::luaMonsterTypeAddElement(lua_State* L)
 	return 1;
 }
 
-int LuaScriptInterface::luaMonsterTypeGetWeaponList(lua_State* L)
+int LuaScriptInterface::luaMonsterTypeGetWeaponWeaknessList(lua_State* L)
 {
-	// monsterType:getWeaponList()
+	// monsterType:getWeaponWeaknessList()
 	MonsterType* monsterType = getUserdata<MonsterType>(L, 1);
 	if (!monsterType) {
 		lua_pushnil(L);
@@ -14435,9 +14435,9 @@ int LuaScriptInterface::luaMonsterTypeGetWeaponList(lua_State* L)
 	return 1;
 }
 
-int LuaScriptInterface::luaMonsterTypeAddWeapon(lua_State* L)
+int LuaScriptInterface::luaMonsterTypeAddWeaponWeakness(lua_State* L)
 {
-	// monsterType:addWeapon(type, percent)
+	// monsterType:addWeaponWeakness(type, percent)
 	MonsterType* monsterType = getUserdata<MonsterType>(L, 1);
 	if (monsterType) {
 		WeaponType_t weapon = getNumber<WeaponType_t>(L, 2);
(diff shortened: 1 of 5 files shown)


ranisalt:
Cool, but the name is terrlble. Maybe `weaponWeakness` instead of just `weapon`, because `getWeaponList` definitely does not recall a weapon _weakness_ list

gigastar:
> Cool, but the name is terrlble. Maybe `weaponWeakness` instead of just `weapon`, because `getWeaponList` definitely does not recall a weapon _weakness_ list

I knew it would come up. I wanted to hear what you guys thought for naming.

src/luascript.cpp
```diff
@@ -3027,6 +3027,9 @@ void LuaScriptInterface::registerFunctions()
 
 	registerMethod("MonsterType", "getElementList", LuaScriptInterface::luaMonsterTypeGetElementList);
 	registerMethod("MonsterType", "addElement", LuaScriptInterface::luaMonsterTypeAddElement);
+	
+	registerMethod("MonsterType", "getWeaponWeaknessList", LuaScriptInterface::luaMonsterTypeGetWeaponWeaknessList);
+	registerMethod("MonsterType", "addWeakness", LuaScriptInterface::luaMonsterTypeAddWeakness);
```
MillhioreBT, line 3032: `addWeakness`? no is `addWeaponWeakness`?

MillhioreBT (changes requested):

src/monster.cpp
```diff
@@ -626,6 +626,23 @@ BlockType_t Monster::blockHit(Creature* attacker, CombatType_t combatType, int32
 				blockType = BLOCK_ARMOR;
 			}
 		}
+		
+		if (const Player* player = attacker->getPlayer()) {
```
4mrcn4, line 630: inverse that condition and early return blockType

src/monster.cpp
```diff
@@ -626,6 +626,23 @@ BlockType_t Monster::blockHit(Creature* attacker, CombatType_t combatType, int32
...
+			int32_t weaponMod = 0;
+			WeaponType_t weapon = player->getWeaponType();
+			auto it = mType->info.weaponMap.find(weapon);
+			if (it != mType->info.weaponMap.end()) {
+				weaponMod = it->second;
+			}
+
+			if (weaponMod != 0) {
```
4mrcn4, line 638: same here

src/monster.cpp
```diff
@@ -626,6 +626,23 @@ BlockType_t Monster::blockHit(Creature* attacker, CombatType_t combatType, int32
...
+				damage = static_cast<int32_t>(std::round(damage * ((100 - weaponMod) / 100.)));
+				if (damage <= 0) {
```
4mrcn4, line 640: and here

src/monster.cpp
gigastar: Wut?

4mrcn4: Instead of writing bloated if statement you can check for falseness and `return blocktype` inside

4mrcn4: so it would be something like
```
if (!(const Player* player = attacker->getPlayer()) {
  return blockType;
}
```

PR #4217, opened 2022-09-24
Force change fishing spot

NRH-AA:
Once a fish is caught you shouldn't gain experience or catch fish.

### Pull Request Prelude
Once a fish is caught the player should be required to change the sqm they are fishing on. They do not gain skill or catch fish until they do.

### Changes Proposed

1) Make tiles that have or dont have fish
2) Tiles with no fish will give no skill
3) Tiles with no fish will decay after 2200 seconds (ice fishing is 900 seconds)
4) Update items.xml add decay on no fish tiles.

**Issues addressed:** 
None


joseluis2g:
is adding a new event just to reset the time the best way to solve this? what about just saving the time plus the cooldown when you catch something and then check if the time is lower

gigastar:
> is adding a new event just to reset the time the best way to solve this? what about just saving the time plus the cooldown when you catch something and then check if the time is lower

It would store every position until the server is reset. I think removing it to save memory is better but I will let the maintainers decide.

soul4soul:
I'm pretty sure on RL tibia the tile id changes when its fished and then decays back to the original id. If I'm remember correctly that is how NGbot knew to show lava on non fishable tiles. When fish is caught transform tile, update items.xml with decay time (60mins) back to original tile.

gigastar:
> I'm pretty sure on RL tibia the tile id changes when its fished and then decays back to the original id. If I'm remember correctly that is how NGbot knew to show lava on non fishable tiles. When fish is caught transform tile, update items.xml with decay time (60mins) back to original tile.

If that's the case we need to pick a water tile for it. @ranisalt @EPuncker @ramon-bernardo @MillhioreBT 
What tile would be best?

EPuncker:
> What tile would be best?

I prefer the way you did, it's easier and clean to implement.

I was going to write a big wall of text explaining why, but lets see what others think

conde2:
You are saving the key of the table using a position table. This won't work since tables cahnges address everytime. 

You should store a unique hash for the position or a string that represents this position.

Transforming the item in this case would be better in my opinion. But the idea is the same.

gigastar:
> You are saving the key of the table using a position table. This won't work since tables cahnges address everytime.
> 
> You should store a unique hash for the position or a string that represents this position.
> 
> Transforming the item in this case would be better in my opinion. But the idea is the same.

I will fix it. Going to wait for some more opinions on how it should be implemented first.

ranisalt:
> I'm pretty sure on RL tibia the tile id changes when its fished and then decays back to the original id.

If that's the case, then the spot becomes unfishable for everyone, not only for the fisher that caught a fish there, correct?

soul4soul:
> > I'm pretty sure on RL tibia the tile id changes when its fished and then decays back to the original id.
> 
> If that's the case, then the spot becomes unfishable for everyone, not only for the fisher that caught a fish there, correct?

correct

gigastar:
I wasn't able to see anything about all players not being able to fish in the spot.

im not sure what this shows

><const code="waterWithFishStart" value="0x11F5"/>
><const code="waterWithFishEnd" value="0x11FA"/>

if I need to modify to change the tile I will.

soul4soul:
> im not sure what this shows
That is the range of fishable client ids 4597-4602 but they need to be converted to server ids.

> Keep in mind RME and most peoples maps probably use all of the water tiles. It might be difficult to add it that way without people having to remap a bunch of stuff. If not let me know.
Sure but if the decay time is added to the other nonfishable water ids then it shouldn't be a problem.

I'm not sure what the nonfishable ids are, its a pain to do but this can be verified and determined with @jo3bingham map tracker.

gigastar:
> Sure but if the decay time is added to the other nonfishable water ids then it shouldn't be a problem.

If im not mistaken there are different water tiles based on how the current is moving, ect. Essentially we would have to find 2 tiles with the same current and set one as the non-fishable.

The issue is we have to change the tile which could mess up the flow of the water and it will look bad then we have to know which tile to set it back too which isn't hard to do but...

If there was 2-3 people fishing in a spot they would make all the tiles change to the 1 tile we make non-fishable and it would look bad. Unless we do what I said before and make multiple non-fish tiles that at least have the same current on the water.

Generally speaking it probably wouldn't be that big of a problem but if someone was to use the fishing system like this and the there was a lot of people playing on it, it would be an issue. So, we can make the system like this and it works fine or we try and go through a whole process which may not be very good in the end anyway.

ranisalt:
> If im not mistaken there are different water tiles based on how the current is moving. Essentially, we would have to find 2 tiles with the same current and set one as the non-fishable.

They are probably in the same order, so it's just a matter of shifting up and down the ID

soul4soul:
> > If im not mistaken there are different water tiles based on how the current is moving. Essentially, we would have to find 2 tiles with the same current and set one as the non-fishable.
> 
> They are probably in the same order, so it's just a matter of shifting up and down the ID. Like doors, mostly

Yup that is what I assume too. 

I'm making this up but something like
4610, 4611, 4612, 4613 = fishable
4614, 4615, 4616, 4617 = nonfishable decay to fishable
4618, 4619, 4620, 4621, 4622, 4623, 4624, 4625 = nonfishable nondecay

But ids needs to be verified on cipbia

data/actions/scripts/tools/fishing.lua
```diff
@@ -1,20 +1,93 @@
...
 
 function onUse(player, item, fromPosition, target, toPosition, isHotkey)
-	local targetId = target.itemid
-	if not table.contains(waterIds, targetId) then
-		return false
+	if not target.itemid then return false end
+	
+	-- Normal Fishing --
+	if table.contains(waterWithFish, target.itemid) or table.contains(waterWithoutFish, target.itemid) then
+		if targetId ~= 7236 and targetId ~= 7237 then
```
ranisalt, line 34: What are those?

ranisalt:
What do you think about breaking up the main `onUse` function into smaller ones that get called depending on the target tile? I find that long function with several indentation levels hard to follow

data/actions/scripts/tools/fishing.lua
```diff
@@ -1,20 +1,93 @@
-local waterIds = {493, 4608, 4609, 4610, 4611, 4612, 4613, 4614, 4615, 4616, 4617, 4618, 4619, 4620, 4621, 4622, 4623, 4624, 4625, 7236, 10499, 15401, 15402}
+-- Water Ids + 1 no fish -1 with fish
+local waterWithFish = {4608, 4610, 4612, 4614, 4616, 4618, 4664, 7236, 15401}
+local waterWithoutFish = {4609, 4611, 4613, 4615, 4617, 4619, 4665, 7237, 15402}
+local waterSpiritId = 10449
```
EPuncker, line 4: sprite?

gigastar: Water spirit, there was code already in the fishing script to allow fishing in a water spirit. I guess you can get some type of rewards.

data/actions/scripts/tools/fishing.lua
gigastar: 7236 and 7237 is ice holes. They don't get a magiceffect sent to them when fishing. This code was already in the original.

data/actions/scripts/tools/fishing.lua
gigastar: Water spirit corpse*

gigastar:
Does anyone know if you can use a mechanical fishing rod to catch normal fish or is it only for a quest?

ranisalt:
> Does anyone know if you can use a mechanical fishing rod to catch normal fish or is it only for a quest?

It is only used during the quest. Doesn't need to be included here, as it only makes sense for global map

data/actions/scripts/tools/fishing.lua
EPuncker: its called water elemental, not water spirit, no?

data/actions/scripts/tools/fishing.lua
```diff
@@ -1,82 +1,112 @@
...
-			if math.random(1, 100) >= 97 then
-				player:addItem(15405, 1)
-				player:addAchievement("Desert Fisher")
-				return true
+					
+			-- Add achievement based on which waterId is being fished on --
+			if target.itemid == 7236 then
+				player:addAchievementProgress("Exquisite Taste", 250)
+			elseif target.itemid == 15401 then
+				player:addAchievementProgress("Desert Fisher")
```
EPuncker, line 81:
```suggestion
				player:addAchievement("Desert Fisher")
```

PR #4219, opened 2022-10-01
Add unit testing framework

ranisalt:
### Pull Request Prelude

### Changes Proposed

This patch adds the support for auto-discovered unit tests by creating files named `test_<modulename>.cpp` i.e. `test_configmanager.cpp` inside `src/tests`. It should allow us to implement unit tests to prevent regressions when PRs change functionality deep in the core, and should also help to document the code with practical examples.

To build tests, enable with `-DENABLE_TESTING=ON` when invoking cmake. To run tests, just execute the `test` target, i.e. `make test` or `ninja test`.

It uses [Boost::Test](https://www.boost.org/doc/libs/1_80_0/libs/test/doc/html/index.html) and a basic test file would be:

```cpp
#define BOOST_TEST_MODULE configmanager

#include "../otpch.h"

#include "../configmanager.h"

#include <boost/test/unit_test.hpp>

BOOST_AUTO_TEST_CASE(free_test_function)
{
    BOOST_TEST(true /* test assertion */);
}
```


ArturKnopik:
is it only for cmake, where is support for VS?

ranisalt:
> is it only for cmake, where is support for VS?

:shrug: can't install, can't add support

You can generate VS makefiles with CMake though

ArturKnopik:
This is only my opinion:
we support cmake and VS building, unit tests should be supported also for cmake and VS

ranisalt:
Feel free to add support and let me know what changes are needed

marmichalski:
Would be good to actually include at least one test in the CI pipeline.

Zbizu:
if there is no built-in way to support this in visual studio, it could be added as a separate project under same solution

ranisalt:
> Would be good to actually include at least one test in the CI pipeline.

Great, I'll think of something small that can work as a proof of concept & example



> if there is no built-in way to support this in visual studio

There probably is support, but I don't use Windows

src/CMakeLists.txt
```diff
@@ -168,4 +167,24 @@ set(tfs_HDR
...
+        Boost::system
+        Boost::iostreams
+        fmt::fmt
+        ${CMAKE_THREAD_LIBS_INIT}
+        ${Crypto++_LIBRARIES}
+        ${LUA_LIBRARIES}
+        ${MYSQL_CLIENT_LIBS}
+        ${PUGIXML_LIBRARIES}
+        )
+target_include_directories(tfslib PRIVATE ${Boost_INCLUDE_DIRS} ${Crypto++_INCLUDE_DIR} ${LUA_INCLUDE_DIR} ${MYSQL_INCLUDE_DIR} ${PUGIXML_INCLUDE_DIR})
```
ranisalt, line 184: I had to add the server itself as a library because tests need to be linked against server code. Since I'm using `OBJECT`, it will not generate a library, and the output executable is the same thing as before.

marmichalski (dismissed):

EPuncker (approved):

ramon-bernardo (approved):

EPuncker (approved):

lgrossi:
I just wanted to say that this is actually a pretty nice initiative. I know that TFS/OTBR iteractions are not always the most friendly ones, but I'm honestly quite glad to see this kind of effort moving forward, regardless of where. Keep it up! 😄

Have you guys managed to integrate it with VS solutions? We went with a full CMake approach for now, tbh I care about GHA telling me that the tests are green more than being able to support it for VS. Would be nice to have it, though, to test against different setups. Curious to hear what have you guys decided about it.

ranisalt:
@lgrossi I tried, and with the VS plugin for CMake I'm able to completely ditch the solution file (which is an hot pile of garbage tbh)

I know @EPuncker couldn't do it because of cryptopp and I didn't manage to find the problem. I would really much like to remove the VS solution as it breaks subdirectories

lgrossi:
@ranisalt yeah, TBH I never used solution, I don't like it, but for compatibility we've been keeping it. Buidling with CMake is easier to setup, more modular, way faster and more reliable. Anyway, thanks for the reply and good luck removing solutions.

PR #4222, opened 2022-10-03
item.custom property fancy access

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes add the ability to get or set a custom attribute using the `__index` and `__newindex` metamethods just like with `player.storage` and `player.accountStorage`

`data/scripts/any.lua`
```lua
print(item.custom[1])
item.custom[1] = 100
print(item.custom.useCount)
item.custom.useCount = 2
item.custom["useCount"] = "uwu"
...
```

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -102,6 +102,17 @@ do
 end
 
 do
+	local function customAttributeProxy(item)
+		return setmetatable({}, {
+			__index = function (self, key)
+				return item:getCustomAttribute(key)
+			end,
+			__newindex = function (self, key, value)
+				item:setCustomAttribute(key, value)
+			end
+		})
+	end
+
 	local function ItemIndex(self, key)
 		local methods = getmetatable(self)
 		if key == "itemid" then
@@ -112,6 +123,8 @@ do
 			return methods.getUniqueId(self)
 		elseif key == "type" then
 			return methods.getSubType(self)
+		elseif key == "custom" then
+			return customAttributeProxy(self)
 		end
 		return methods[key]
 	end


yamaken93:
This is fancy indeed, it even allows you to save the custom table in a local variable and then you can edit the item custom attributes using that custom instance that was created once, the problem is the item not existing when that custom table saved in a local variable is called. Performance is horrible tho.

MillhioreBT:
> This is fancy indeed, it even allows you to save the custom table in a local variable and then you can edit the item custom attributes using that custom instance that was created once, the problem is the item not existing when that custom table saved in a local variable is called. Performance is horrible tho.

The idea of ​​this is to reduce typing both `getCustomAttribute and setCustomAttribute`, it's only useful when doing a quick script. using the returned table to save it in a variable and then using it later is a mistake, in fact to this day and probably will be for a while it's not a good idea to save user data in a variable to use later, it's the same problem of `addEvent`.

yamaken93:
Here the "optimized" version of that code:
```
do
	local function createItemIndex(table)
		local methodGetId = table.getId
		local methodGetActionId = table.getActionId
		local methodGetUniqueId = table.getUniqueId
		local methodGetSubType = table.getSubType

		local item = nil

		local index = function(self, key)
			local ret = item:getCustomAttribute(key)
			item = nil
			return ret
		end

		local newIndex = function(self, key, value)
			item:setCustomAttribute(key, value)
			item = nil
		end

		local customAttributeProxyTable = setmetatable({}, {
			__index = index,
			__newindex = newIndex
		})

		return function(self, key)
			if (key == "itemid") then
				return methodGetId(self)
			elseif (key == "actionid") then
				return methodGetActionId(self)
			elseif (key == "uid") then
				return methodGetUniqueId(self)
			elseif (key == "type") then
				return methodGetSubType(self)
			elseif (key == "custom") then
				item = self
				return customAttributeProxyTable
			end
			return table[key]
		end
	end
	rawgetmetatable("Item").__index = createItemIndex(Item)
	rawgetmetatable("Container").__index = createItemIndex(Container)
	rawgetmetatable("Teleport").__index = createItemIndex(Teleport)
end

```
Its using an "item" upvalue that is reset after index or newindex is called, any usage differently from item.custom["customattribute"] will throw an lua error. I'm already using that optimized version without the custom attribute fancy access.

MillhioreBT:
In order not to create confusion with `custom` property I will close the PR, if I have a better idea that meets the expectations I will open a new one, anyway this is not necessary at all.

PR #4236, opened 2022-11-01
RFC: Replace CMake with Meson

ranisalt:
### Pull Request Prelude

### Changes Proposed

CMake works, and is widely adopted. But one must agree that maintaining it is terrible, configuration is more complicated than necessary and the scripting language is cumbersome. 

[Meson is on the rise](https://gms.tf/the-rise-of-meson.html) and several large projects are using it, such as the entire GNOME ecosystem, DXVK, Mesa, QEMU, and soon Xorg and Zstd; the documentation is covers a lot more and is extensive; the scripting language is cleaner, and the defaults are better.

[It is also faster](https://mesonbuild.com/Simple-comparison.html), _according to their own tests_.

By using subprojects, which meson will download and build automatically when not installed in the system, we could now skip installing fmt, Lua and pugixml, instead building it locally. Since a lot of people still use pre-historic Debian or Ubuntu releases, and even the latest releases usually lag far behind, this is a relief when adopting newer versions.

Hopefully, this is a quality of life improvement for everyone developing TFS.

**What to check:**

- [ ] the same configure options are available AND working
- [ ] release builds generate working binaries
- [ ] the same relevant build flags are used by the compiler


marmichalski:
Maintaining cmake here must be really hard. Last change to the config file was done ~8 months ago and modules were last touched few years ago. Personally, I do not any reason for this change. Keep in mind that most of the consumers of this repository have problems with reading comprehension and you want to throw new build tool at them, because """its on the rise""".

ArturKnopik:
+1

ranisalt:
> Keep in mind that most of the consumers of this repository have problems with reading comprehension and you want to throw new build tool at them, because """its on the rise""".

I think this is a very important argument. I'm not expecting to merge this unless it proves to be easier to handle.

If Meson is widely available, it is actually _easier_ to do everything - there are fewer dependencies to install. "On the rise" was a terrible choice of words, since it is already well established, i.e. is as common as CMake.

meson.build
```diff
@@ -0,0 +1,114 @@
+project('tfs', 'cpp', license: ['gpl-2.0'], default_options: ['cpp_std=c++17'])
+
+fmt_dep = dependency('fmt', version: '>=7.1.0', fallback: ['fmt', 'fmt_dep'])
+pugixml_dep = dependency('pugixml', fallback: ['pugixml', 'pugixml_dep'])
```
eli-schwartz, line 4: You can also pass `default_options: ['default_library=static']` anywhere that you are doing subproject fallbacks. That guarantees the subproject is statically linked by default, which is reasonable for something that won't be shared with system software anyway. It also simplifies installation and/or distribution.

eli-schwartz (commented):
> By using subprojects, which meson will download and build automatically when not installed in the system, we could now skip installing fmt, Lua and pugixml, instead building it locally. Since a lot of people still use pre-historic Debian or Ubuntu releases, and even the latest releases usually lag far behind, this is a relief when adopting newer versions.

Just want to emphasize this is probably one of the main advantages that users would see.

meson.build
ranisalt: Done!

omarcopires:
Taking this opportunity to say that it might be interesting to do something like they did in opentibiabr/canary regarding the compilation of windows that it is no longer necessary to install the vcpkg libraries manually, just open the project folder and all dependencies are installed automatically.
It's just a suggestion, I don't know if it would be viable or even if it would be an improvement and not a return.

ranisalt:
Wow, vcpkg is such low quality garbage. This will never work, I give up trying to work around, a sane OS with a real package manager is required.

Issue #4193, opened 2022-08-14
eventPlayerOnItemMoved never called when item equiped/deequiped by hotkey


ArturKnopik:
ProtocolGame::parsePacket -> case 0x77: parseEquipObject(msg); -> 
void ProtocolGame::parseEquipObject(NetworkMessage& msg) ->  g_game.playerEquipItem -> internalMoveItem do not call eventPlayerOnItemMoved when item moved by hotkey

PR #4238, opened 2022-11-03
call onItemMove and Moved from Game:playerEquipItem

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
**Issues addressed:** #4193

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -1851,11 +1851,22 @@ void Game::playerEquipItem(uint32_t playerId, uint16_t spriteId)
 
 	Item* slotItem = player->getInventoryItem(slot);
 	Item* equipItem = searchForItem(backpack, it.id);
+	Position fromPos, toPos;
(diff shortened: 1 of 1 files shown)


ramon-bernardo (approved):
This fix the problem, looks good!

EPuncker:
great, need to run clang for code style only

MillhioreBT:
> great, need to run clang for code style only

It was broken many PRs ago, I don't know when they merged a PR with code without the required format, but it doesn't come from this PR. the downside of merging PR without c-clang is that it propagates through all subsequent PRs.

If you think it's a good idea that I do the c-clang to the whole source, I'll gladly do it right here in this PR, just let me know.

ranisalt:
@EPuncker we need to do it separately. I don't have this issue locally, format changes nothing, seems to be an issue with the specific version shipped by the OS that runs the action.

PR #4242, opened 2022-11-05
Move hangable check to move action

nekiro:
### Pull Request Prelude

### Changes Proposed
Checking hangables should only happen on move action, there is no point in checking use, look or any other action.

**Issues addressed:** 
In my case (but it wasn't reproduced on clean tfs) this prevented me from using windows on south faced walls, while I was being north of the wall.

This scenario, when standing to north of the wall with window, prevented use on it.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -225,6 +225,21 @@ Thing* Game::internalGetThing(Player* player, const Position& pos, int32_t index
 				} else {
 					thing = tile->getTopVisibleCreature(player);
 				}
+				
+				if (player && tile->hasFlag(TILESTATE_SUPPORTS_HANGABLE)) {
+					// do extra checks here if the thing is accessible
+					if (thing && thing->getItem()) {
+						if (tile->hasProperty(CONST_PROP_ISVERTICAL)) {
+							if (player->getPosition().x + 1 == tile->getPosition().x) {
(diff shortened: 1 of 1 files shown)


ramon-bernardo:
The same thing happens when trying to hang something on the wall. It's a RL behavior?
[north -> south]
image
[south-> north]
image

nekiro:
> The same thing happens when trying to hang something on the wall. It's a RL behavior? Error: [north -> south] image [west -> east] image
> 
> Ok: [south -> north] image
> 
> [east -> west] image

Yes, this is correct, Are you able to hang up something on a wall you don't see IRL? Try washing window without opening it, that's how I understand this and why this is correctly working.

ramon-bernardo (approved):

Felipe21:
this is still bugged as you can see in this video
https://files.fm/f/484ra27m3

nekiro:
> 

This issue is totally unrelated to what this pr tries to "fix".

Felipe21:
> > this is still bugged as you can see in this video
> > https://files.fm/f/484ra27m3
> 
> This issue is totally unrelated to what this pr tries to "fix".

do you know where i can find the PR that fix my issue?

EPuncker (approved):

Issue #4274, opened 2022-12-18
look me 


daniofordon:
tfs 12.86 and otclient meh 12.86

look character

https://prnt.sc/IzzCkB6Pcroe

EPuncker:
@ranisalt can you check this please? seems to be related to that ip commit

ranisalt:
I can't open the image, can you post it directly to Github?

EPuncker:
> I can't open the image, can you post it directly to Github?

![image](https://user-images.githubusercontent.com/4684880/208420678-2bbb4a28-3377-4554-a102-0637091c7f26.png)

daniofordon:
https://user-images.githubusercontent.com/4684880/208420678-2bbb4a28-3377-4554-a102-0637091c7f26.png

ranisalt:
Working on a fix

Zbizu:
@daniofordon please use issue templates provided when creating an issue next time

LuckyM91:
When you are trying to look at yourself that error appears

PR #3952, opened 2022-02-16
IPv6 support

ranisalt:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Rebase and some updates on #2178

Hey, it's ~~2017~~ 2022 so it's high time we implemented support for IPv6 connections. The server now listens on an IPv6 socket unless it's configured to only bind to a specific IPv4 address. The address is now stored as a varchar in the database. The internal representation of the address has been changed to the
asio-native one for simplicity.

Closes #2178

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4283, opened 2022-12-27
Remove usages of Game.convertIpToString and mark as deprecated

ranisalt:
### Pull Request Prelude

### Changes Proposed

Fixes a backwards-incompatibility introduced in #3952 where there are still some functions around that expect numeric IPs, which are not available anymore. Since most of the time the IP comes from `player:getIP()`, which returns string values, this function can be skipped, but if they come from somewhere else it will still try to convert the number.

**Issues addressed:** 
#4274 

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -580,12 +580,10 @@ function getPlayerLearnedInstantSpell(cid, name) local p = Player(cid) return p 
 function isPlayerGhost(cid) local p = Player(cid) return p and p:isInGhostMode() or false end
 function isPlayerPzLocked(cid) local p = Player(cid) return p and p:isPzLocked() or false end
 function isPremium(cid) local p = Player(cid) return p and p:isPremium() or false end
-function getPlayersByIPAddress(ip, mask)
-	if not mask then mask = 0xFFFFFFFF end
-	local masked = bit.band(ip, mask)
+function getPlayersByIPAddress(ip)
 	local result = {}
 	for _, player in ipairs(Game.getPlayers()) do
-		if bit.band(player:getIp(), mask) == masked then
+		if player:getIp() == masked then
 			result[#result + 1] = player:getId()
 		end
 	end
@@ -1290,6 +1288,23 @@ end
 
 function doExecuteRaid(raidName)
 	return Game.startRaid(raidName)
+end
+
+function Game.convertIpToString(ip)
+	print("[Warning - " .. debug.getinfo(2).source:match("@?(.*)") .. "] Function Game.convertIpToString is deprecated and will be removed in the future. Use the return value of player:getIp() instead.")
+
+	if type(ip) == "string" then
+		return ip
+	end
+
+	local band = bit.band
+	local rshift = bit.rshift
+	return string.format("%d.%d.%d.%d",
+		band(ip, 0xFF),
+		band(rshift(ip, 8), 0xFF),
+		band(rshift(ip, 16), 0xFF),
+		rshift(ip, 24)
+	)
 end
 
 numberToVariant = Variant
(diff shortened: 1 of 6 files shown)


EPuncker (dismissed):

data/lib/compat/compat.lua
```diff
@@ -580,12 +580,10 @@ function getPlayerLearnedInstantSpell(cid, name) local p = Player(cid) return p
...
 function isPlayerPzLocked(cid) local p = Player(cid) return p and p:isPzLocked() or false end
 function isPremium(cid) local p = Player(cid) return p and p:isPremium() or false end
-function getPlayersByIPAddress(ip, mask)
-	if not mask then mask = 0xFFFFFFFF end
-	local masked = bit.band(ip, mask)
+function getPlayersByIPAddress(ip)
 	local result = {}
 	for _, player in ipairs(Game.getPlayers()) do
-		if bit.band(player:getIp(), mask) == masked then
+		if player:getIp() == masked then
```
Zbizu, line 586: you have a nil value there, variable `masked` no longer exists within this function scope

Zbizu (changes requested):

data/lib/compat/compat.lua
MillhioreBT: should look something like this right?
```lua
function getPlayersByIPAddress(ip, mask)
	local players = {}
	if type(id) == "string" then
		for _, player in ipairs(Game.getPlayers()) do
			if player:getIp() == ip then
				players[#players + 1] = player:getId()
			end
		end
	else
		if not mask then mask = 0xFFFFFFFF end
		local masked = bit.band(ip, mask)
		for _, player in ipairs(Game.getPlayers()) do
			if bit.band(player:getIp(), mask) == masked then
				players[#players + 1] = player:getId()
			end
		end
	end
	return players
end
```

ranisalt: @xmish yes, that's the reasoning. I just kept this compatibility of accepting numeric IPs for the extremely unlikely, but definitely possible scenario where someone manually blocked an IP by hardcoding it.

EPuncker (approved):

data/lib/compat/compat.lua
ranisalt: @Zbizu hope it is good enough now :sweat_smile:

omarcopires:
The advantage is that the error no longer happens, however the ip appears as follows:
image

ranisalt:
@omarcopires does this print just the return value of `player:getIp()`, or are you still calling `Game.convertIpToString()`?

omarcopires:
> @omarcopires does this print just the return value of `player:getIp()`, or are you still calling `Game.convertIpToString()`?

I'm using the code as in your branch, I didn't make any changes.

ranisalt:
> I'm using the code as in your branch, I didn't make any changes.

As @xmish pointed out, that's a valid IP, and your server console is printing that the function has been deprecated, where it has been called (filename and line), and that you should skip calling it instead. What do you think, is this enough information for the server owner?

omarcopires:
> ```c++
> ::ffff:127.0.0.1
> ```

I didn't realize that was the default behavior, sorry.
This is more than enough.

omarcopires:
> > Estou usando o código como em sua ramificação, eu não fiz nenhuma alteração.
> 
> Como @xmish apontado, esse é um IP válido, e o console do servidor está imprimindo que a função foi preterida, onde foi chamada (nome do arquivo e linha), e que você deve ignorar chamá-la em vez disso. O que você acha, essa informação é suficiente para o proprietário do servidor?



> > I'm using the code as in your branch, I didn't make any changes.
> 
> As @xmish pointed out, that's a valid IP, and your server console is printing that the function has been deprecated, where it has been called (filename and line), and that you should skip calling it instead. What do you think, is this enough information for the server owner?

I didn't realize that was the default behavior, sorry.
This is more than enough.

omarcopires:
> > I'm using the code as in your branch, I didn't make any changes.
> 
> As @xmish pointed out, that's a valid IP, and your server console is printing that the function has been deprecated, where it has been called (filename and line), and that you should skip calling it instead. What do you think, is this enough information for the server owner?

I didn't realize that was the default behavior, sorry.
This is more than enough.

omarcopires (approved):

PR #4293, opened 2023-01-16
Fix: Price of NPC's items

### LINKED ISSUE #4292: Price of incorrect items

1. Open any buy and sell NPC and add a list of products if there are not enough some values will be incorrect.

### Expected behaviour 
Show the correct values.

### Actual behaviour 
Sem título

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Fixes the window for buying and selling items on NPC's.

Credits: @Zbizu

**Issues addressed:** 
Closes #4292

### DIFF
--- a/data/npc/lib/npcsystem/modules.lua
+++ b/data/npc/lib/npcsystem/modules.lua
@@ -815,7 +815,7 @@ if not Modules then
 			if it:getId() ~= 0 then
 				local shopItem = self:getShopItem(itemid, itemSubType)
 				if not shopItem then
-					self.npcHandler.shopItems[#self.npcHandler.shopItems + 1] = {id = itemid, buy = cost, sell = -1, subType = itemSubType, name = realName or it:getName()}
+					self.npcHandler.shopItems[#self.npcHandler.shopItems + 1] = {id = itemid, buy = cost, sell = 0, subType = itemSubType, name = realName or it:getName()}
 				else
 					if cost < shopItem.sell then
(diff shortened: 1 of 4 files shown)


src/enums.h
```diff
@@ -560,12 +560,12 @@ struct ShopInfo
 {
 	uint16_t itemId = 0;
 	int32_t subType = 1;
-	int64_t buyPrice = 0;
-	int64_t sellPrice = 0;
+	uint32_t buyPrice = 0;
+	uint32_t sellPrice = 0;
```
DSpeichert, line 564: Why not uint64_t? Didn't the client start accepting larger numbers recently?

omarcopires: Values currently allowed in version 13.x do not exceed the uint32_t limits.

EPuncker (approved):

data/npc/lib/npcsystem/modules.lua
```diff
@@ -960,7 +960,7 @@ if not Modules then
 			return false
 		end
 
-		if shopItem.buy == -1 then
+		if shopItem.buy == 0 then
```
Kamenuvol, line 963: I think it should be a valid value to buy/sell for free. To not me buyable/sellable the value in Lua table should be nil.

EPuncker: I don't think the client can handle that (haven't tested), since cip sell items for 1 gp at testserver assistant and not 0 (for free)

Issue #4255, opened 2022-11-24
monster atk direction update


emil92b:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. make monster change direction by walking from one side to the other
2. watch it switch two times when it should only switch once

### Expected behaviour
https://user-images.githubusercontent.com/43324808/203817025-ba585909-4f81-49bb-8974-2782c04c393c.mp4

### Actual behaviour
https://user-images.githubusercontent.com/43324808/203818611-fb14a552-3aa1-4fb0-b33e-3cee4490824e.mp4

EPuncker:
this is related to that issue tfs have about monsters being "dumb"

Cesar525:
The expected behavior seems to just have a delay in compare to the tfs.

floki21uu:
the fix is easy
`void Monster::updateLookDirection()
{
	Direction newDir = getDirection();
	if (attackedCreature) {
		const Position& pos = getPosition();
		const Position& attackedCreaturePos = attackedCreature->getPosition();
		int_fast32_t minotaur_x = Position::getOffsetX(attackedCreaturePos, pos);
		int_fast32_t minotaur_y = Position::getOffsetY(attackedCreaturePos, pos);

		int32_t dx = std::abs(minotaur_x);
		int32_t dy = std::abs(minotaur_y);

		if (dx > dy) 
		{
			if (minotaur_x < 0) {
				newDir = DIRECTION_WEST;
			} else {
				newDir = DIRECTION_EAST;
			}
		} 
		else if (dx < dy)
		{
			if (minotaur_y < 0) {
				newDir = DIRECTION_NORTH;
			} else {
				newDir = DIRECTION_SOUTH;
			}
		} 
		else 
		{
			Direction dir = getDirection();

			if (minotaur_x < 0 && minotaur_y < 0) 
			{
				if (dir == DIRECTION_SOUTH)
				{
					newDir = DIRECTION_WEST;
				} 
				else if (dir == DIRECTION_NORTH)
				{
					newDir = DIRECTION_WEST;
				} 
				else if (dir == DIRECTION_EAST)
				{
					newDir = DIRECTION_WEST;
				}
			} 
			else if (minotaur_x < 0 && minotaur_y > 0)
			{
[...]

PR #4296, opened 2023-01-21
Improve monsters turning to target behaviour

marmichalski:
### Pull Request Prelude

### Changes Proposed

* Simplify updateLookDirection logic removing 2-step turning in some cases.
* Update look direction on successful attack or when melee was out of reach

There is still some room for improvement when it comes to ranged creatures more than necessarily turning to you while they're running to keep distance. I may or may not fix it in separate PR, as this is another issue.

@emil92b pls test

**Issues addressed:** #4255

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -800,7 +800,7 @@ void Monster::doAttacking(uint32_t interval)
 		return;
 	}
 
-	bool updateLook = true;
+	bool lookUpdated = false;
 	bool resetTicks = interval != 0;
 	attackTicks += interval;
 
@@ -816,9 +816,9 @@ void Monster::doAttacking(uint32_t interval)
 
 		if (canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
 			if (spellBlock.chance >= static_cast<uint32_t>(uniform_random(1, 100))) {
-				if (updateLook) {
+				if (!lookUpdated) {
 					updateLookDirection();
-					updateLook = false;
+					lookUpdated = true;
 				}
 
 				minCombatValue = spellBlock.minCombatValue;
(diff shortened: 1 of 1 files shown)


emil92b:
seem to work for melee, not sure about ranged, also noticed in real tibia they move more frequently, but thats another issue too. btw you don't need to rename and bool flip

marmichalski:
> btw you don't need to rename and bool flip

Without flipping the bool they would still be turning twice most of the times, as this method is fired more often than the actual attack. And it seemed like they turn on attack in vanilla. 🙈

emil92b:
> > btw you don't need to rename and bool flip
> 
> Without flipping the bool they would still be turning twice most of the times, as this method is fired more often than the actual attack. And it seemed like they turn on attack in vanilla. 🙈

isn't it same logic tho, only difference is this `lastMeleeAttack == 0` which i assume takes care of the double turning

marmichalski:
> > > btw you don't need to rename and bool flip
> > 
> > 
> > Without flipping the bool they would still be turning twice most of the times, as this method is fired more often than the actual attack. And it seemed like they turn on attack in vanilla. 🙈
> 
> isn't it same logic tho, only difference is this `lastMeleeAttack == 0` which i assume takes care of the double turning

Nope. With `true` as default the creature would turn anytime this method is called, which IIRC was through game.checkCreatures. `lastMeleeAttack == 0` is for ranged/non-hostile creatures to face you - this means the target was out of range for melee, or the creature has no melee at all (and also that it's the first ever melee attack of this monster, but thats of no importance).

ranisalt:
I prefer the new name nevertheless. `lookUpdated` is clear, it *has been updated*, while `updateLook` to me means that it *should be updated*, which is not the case

EPuncker (approved):

PR #4309, opened 2023-02-01
Libs: warn user about duplicate storage value

omarcopires:
### Pull Request Prelude

### Changes Proposed 
- The script checks for duplicate stores when starting the server.
- Adds the recently introduced AccountStorageKeys variable.

**Note:** It may not be useful for small storages, but for larger servers like realmap it is very useful.

Example:
image


omarcopires:
I'm thinking maybe it would be interesting to check the actionId and uniqueId of your check variables, I'd like to hear from some of you.

src/configmanager.cpp
```diff
@@ -230,6 +230,7 @@ bool ConfigManager::load()
 	boolean[REMOVE_ON_DESPAWN] = getGlobalBoolean(L, "removeOnDespawn", true);
 	boolean[PLAYER_CONSOLE_LOGS] = getGlobalBoolean(L, "showPlayerLogInConsole", true);
 	boolean[TWO_FACTOR_AUTH] = getGlobalBoolean(L, "enableTwoFactorAuth", true);
+	boolean[TWO_FACTOR_AUTH] = getGlobalBoolean(L, "enableTwoFactorAuth", true);
```
EPuncker, line 233: hmm

config.lua.dist
```diff
@@ -96,6 +96,7 @@ premiumToSendPrivate = false
 forceMonsterTypesOnLoad = true
 cleanProtectionZones = false
 showPlayerLogInConsole = true
+checkDuplicatesValues = false
```
ranisalt, line 99: I think this is not clear enough. How about `checkDuplicateStorageKeys`? Also, some description would be cool, but a name change is enough

data/global.lua
```diff
@@ -231,3 +231,24 @@ function getPlayerDatabaseInfo(name_or_guid)
...
+	-- Retrieve the table to be checked
+	local keys = _G[varName]
+	-- Create a table to keep track of the keys already seen
+	local seen = {}
+	-- Iterate over the keys in the table
+	for k, v in pairs(keys) do
+		-- Check if a key has already been seen
+		if seen[v] then
+			-- If it has, return true and the duplicate key
+			return true, "Duplicate key found: " .. v
```
ranisalt, line 247: Don't return it, and let it find all duplicates before returning. That way, if the key is duplicated more than once, it will be caught at the same time instead.

Given that you are printing the message, maybe return a list of duplicate values and build the message at call point.

data/globalevents/scripts/startup.lua
```diff
@@ -42,4 +41,16 @@ function onStartup()
...
+		-- List of table names to be checked for duplicates
+		local variableNames = {"AccountStorageKeys", "PlayerStorageKeys", "GlobalStorageKeys", "actionIds", "uniqueIds"}
+		-- Loop through the list of table names
+		for _, variableName in ipairs(variableNames) do
+			-- Call the checkDuplicatesValues function for each table
+			local hasDuplicates, message = checkDuplicatesValues(variableName)
+			-- Print the result of the check for each table
+			print(">> Checking " .. variableName .. ": " .. message)
+		end
+	end
```
ranisalt, line 55: Hmm I believe duplicate storage values to be a critical error (it will lead to bugs, 100% sure) so it can crash the server and prevent it from starting. If the server owner wants to use duplicate values they can always disable the check, so if enabled, it should prevent invalid server state.

ranisalt (changes requested):
Interesting, and definitely will help prevent nasty bugs.

data/global.lua
omarcopires: done!

config.lua.dist
omarcopires: Done!

data/globalevents/scripts/startup.lua
omarcopires: I don't know how to shut down the server, because `Game.setGameState(GAME_STATE_CLOSED)` didn't work in this case.

ranisalt: Hmm, leave it as is for now, I don't think we have a "shutdown" function in Lua

EPuncker (approved):

ranisalt (approved):

Issue #4310, opened 2023-02-01
Unable to cast runes with no vocation restrictions


xmish:
When player tries to use rune that is for all vocations (empty vocationMap) he gets an error:
`You have the wrong vocation to cast this spell.`

spells.xml:
```XML
<rune group="support" spellid="14" name="Chameleon" id="2291" charges="1" lvl="5" maglv="4" cooldown="1600" groupcooldown="1600" enabled="1" script="runes/support/chameleon.lua" />
```

Most likely this is the cause (#4226):
https://github.com/otland/forgottenserver/blob/07c27637fedc454ad78388beb46db114d8e1b31e/src/spells.cpp#L594

had to change it back to make it work again:
```cpp
	} else if (!vocSpellMap.empty() && vocSpellMap.find(player->getVocationId()) == vocSpellMap.end()) {
```
or to:
```cpp
	} else if (!vocationSpellMap.empty() && !hasVocationSpellMap(player->getVocationId())) {
```

PR #4311, opened 2023-02-01
Fix spell vocation check

xmish:
### Pull Request Prelude

### Changes Proposed
Revert vocationSpellMap check logic to allow use ~~spells~~/runes with no `<vocation>` specified
e.g.
```XML
<rune group="support" spellid="14" name="Chameleon" id="2291" charges="1" lvl="5" maglv="4" cooldown="1600" groupcooldown="1600" enabled="1" script="runes/support/chameleon.lua" />
```

**Issues addressed:** 
#4310

### DIFF
--- a/src/spells.cpp
+++ b/src/spells.cpp
@@ -1052,11 +1052,9 @@ bool InstantSpell::canCast(const Player* player) const
 
 	if (isLearnable()) {
 		return player->hasLearnedInstantSpell(getName());
-	} else {
-		return hasVocationSpellMap(player->getVocationId());
 	}
 
-	return false;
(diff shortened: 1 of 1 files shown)


MillhioreBT:
Just follow this PR, and do the same: #4262

MillhioreBT:
> @MillhioreBT yea first had such idea to fix it like that but then method name would no longer be valid as even if set is empty it gives a “true”

If you don't like that solution, you can use yours, but then you should remove `vocationSpellMap.empty()` inside the function as otherwise you'll be double checking the same thing.

When I have time I will change the other function to be consistent with the name as in this case.
If you want to do it you would be perfect ;)

PR #4314, opened 2023-02-03
Prevent map loading crash

maattch:
### Pull Request Prelude

### Changes Proposed
Prevent crash when loading map and it does not exists, may happen if you misstyped the map name in config.lua

**Issues addressed:**
If the map does not exists, server just crash without telling you why.

### DIFF
--- a/src/fileloader.cpp
+++ b/src/fileloader.cpp
@@ -11,17 +11,23 @@ namespace OTB {
 
 constexpr Identifier wildcard = {{'\0', '\0', '\0', '\0'}};
 
-Loader::Loader(const std::string& fileName, const Identifier& acceptedIdentifier) : fileContents(fileName)
+Loader::Loader(const std::string& fileName, const Identifier& acceptedIdentifier)
 {
+	try {
+		fileContents.open(fileName); // may throw if file does not exists
+	} catch (...) {
+		throw OTBMException(OPEN_FAILED);
+	}
+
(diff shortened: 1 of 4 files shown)


src/fileloader.h
```diff
@@ -27,14 +27,36 @@ struct Node
...
+			case OPEN_FAILED: {
+				w = "Failed to open OTBM file";
+				break;
+			}
+			default: {
+				w = "OTBM Exception";
+				break;
+			}
+		}
+	}
```
ranisalt, line 54: Pass the error message in a constructor instead. Also, `std::string_view` over `const char*` please

src/fileloader.cpp
```diff
@@ -11,25 +11,31 @@ namespace OTB {
 
 constexpr Identifier wildcard = {{'\0', '\0', '\0', '\0'}};
 
-Loader::Loader(const std::string& fileName, const Identifier& acceptedIdentifier) : fileContents(fileName)
+Loader::Loader(const std::string& fileName, const Identifier& acceptedIdentifier)
 {
+	try {
+		fileContents.open(fileName); // may throw if file does not exists
```
ranisalt, line 17: I would rather pass the `std::ifstream` as a parameter so that it throws before the constructor is called, and in here it is guaranteed to be open.

EPuncker (approved):

PR #4323, opened 2023-02-17
Fix issue with position calculation in move function

### LINKED ISSUE #4316: Lua Script Error: moveUpstairs

1. Use a ladder to climb

### Expected behaviour 
Climb the ladder and return no error.

### Actual behaviour 
It climbs the ladder and returns an error to the console
```Lua Script Error: [Action Interface]
data/actions/scripts/others/teleport.lua:onUse
LuaScriptInterface::getNumber(). Argument -1 has out-of-range value for unsigned short: -1
stack traceback:
        data/lib/core/position.lua:38: in function 'moveUpstairs'
        data/actions/scripts/others/teleport.lua:4: in function <data/actions/scripts/others/teleport.lua:2>```

### PR DESCRIPTION
**Pull request by omarcopires**
### Pull Request Prelude

### Changes Proposed 
This pull request fixes an issue with the move function that was causing incorrect position calculation. The issue was due to the use of the + operator to add a Position object to a directionOffset object, which was resulting in unexpected behavior. The correct behavior is to create a new Position object with the correct coordinates.

To fix the issue, the code has been modified to create a new Position object using the x and y values of the current position, and the x and y values of the directionOffset object. The z value is left unchanged since we're assuming the character is moving on the same floor.

This fix has been tested and verified to work correctly in a variety of scenarios.

**Issues addressed:** 
Close #4316

### DIFF
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -27,7 +27,7 @@ function Position:moveUpstairs()
 
 	self.z = self.z - 1
 
-	local defaultPosition = self + Position.directionOffset[DIRECTION_SOUTH]
+	local position = Position(self.x + Position.directionOffset[direction].x, self.y + Position.directionOffset[direction].y, self.z)
 	local toTile = Tile(defaultPosition)
 	if not toTile or not toTile:isWalkable() then
 		for direction = DIRECTION_NORTH, DIRECTION_NORTHEAST do

### CODE REVIEW

**Comment by projectMWX** [comment]
Just to clarify, made this fix from my phone laying in bed. Was going to check to see if there was a magic method set anywhere and if so, why the addition isn't applying properly. But if not, Omarcopires is happy and tested with no problems.

**Comment by omarcopires** [comment]
> Just to clarify, made this fix from my phone laying in bed. Was going to check to see if there was a magic method set anywhere and if so, why the addition isn't applying properly. But if not, Omarcopires is happy and tested with no problems.

I correctly applied the fix, but now it's working as it should.

**Comment by projectMWX** [comment]
> 

True, so just adding z = 0, in the Position.directionOffset table should be sufficient.

**Comment by MillhioreBT** [comment]
This is not correct, the Position are tables, it is enough that the first table has the Position metatable to call the `__add` metamethod.

as Position are tables also tables can be Position.
I don't know what solves this, if there is nothing to solve, in fact the code you propose is wrong.

**Comment by projectMWX** [comment]
Have you even bothered to test the code? I wrote it from my phone laying in bed to help a dude on OTAcademy and to be quite honest, didn't want to be tagged into this PR or anything affiliated with TFS for this exact reason.

As it stands, you can't create a Position with a negative value, it's that simple. The code I wrote for him fixes it. It's up to you to include it or figure out a way to include it in the metamethod.

File: data/lib/core/position.lua [hunk 1]
```diff
@@ -35,7 +35,7 @@ function Position:moveUpstairs()
 				direction = DIRECTION_WEST
 			end
 
-			local position = self + Position.directionOffset[direction]
+			local position = Position(self.x + Position.directionOffset[direction].x, self.y + Position.directionOffset[direction].y, self.z)
```
ranisalt, line 38: `Position:getNextPosition(direction)` already does that, so

```lua
local position = self:getNextPosition(direction)
```

should have the same result, but cleaner

projectMWX: This will not work by the looks of it. It seems that Position:getNextPosition(direction) applies the offsets to itself, so on the next iteration in the loop, it will then be applying the next offsets to that and not the original position.

projectMWX:
[omarcopires](https://github.com/omarcopires) Are you able to test this. Make a ladder, and upstairs block every tile around the hole except for the SW corner.

nekiro:
> > 
> 
> Position.directionOffset is not a valid Position even with the z co-ord. The meta requires both sides of the addition to be Positions. And we can't create Position offsets i.e Position(-1, 0 0) as they require unsigned co-ordinates

you can create them explicitly like 65535, y, z, if z is 0 then __add will add nothing to the position, so its safe to use 0 as z

projectMWX:
> > > 
> > 
> > 
> > Position.directionOffset is not a valid Position even with the z co-ord. The meta requires both sides of the addition to be Positions. And we can't create Position offsets i.e Position(-1, 0 0) as they require unsigned co-ordinates
> 
> You can create them explicitly like 65535, y, z, if z is 0 then `__add` will add nothing to the `z` position, so its safe to use 0 as z -1 probably underflows later in the c++ world and does addition on 65535 anyway, which overflows the number, so it should work, but someone has to check it.

I see the logic, but definitely isn't working based on what I've seen. I don't have a copy of latest to test, but the reason it's gone un-noticed for this long is due to the fact that most stairs wont have a default south blocked tile. I think the lua fails and halts the script before the server can process the underflow/overflow, but it's only a guess.

data/lib/core/position.lua
omarcopires: I don't think it's ideal behavior.
image

projectMWX: > I don't think it's ideal behavior. image

Your :isWalkable() method is clearly busted. Or that table item doesn't have TILESTATE_BLOCKSOLID or TILESTATE_IMMOVABLEBLOCKSOLID.

nekiro:
> > > > 
> > > 
> > > 
> > > Position.directionOffset is not a valid Position even with the z co-ord. The meta requires both sides of the addition to be Positions. And we can't create Position offsets i.e Position(-1, 0 0) as they require unsigned co-ordinates
> > 
> > 
> > You can create them explicitly like 65535, y, z, if z is 0 then `__add` will add nothing to the `z` position, so its safe to use 0 as z -1 probably underflows later in the c++ world and does addition on 65535 anyway, which overflows the number, so it should work, but someone has to check it.
> 
> I see the logic, but definitely isn't working based on what I've seen. I don't have a copy of latest to test, but the reason it's gone un-noticed for this long is due to the fact that most stairs wont have a default south blocked tile. I think the lua fails and halts the script before the server can process the underflow/overflow, but it's only a guess.

There is no checking in lua that would "halt" the execution, calls from lua go directly to C lua api, because `__add` is registered there.
-1 gives a warning, because of the code that checks number validity in `getNumber<T>`, this is "recent" addition.
Even though it gives that warning the code will still go thru and work.
The code above that doesn't work is because of the addition to raw table, which does not trigger `__add` on position metatables, because the second argument is not a metatable.
The fix in this case would be to replace raw tables in `Position.directionOffset` to use Position metatables and no other changes to the code.

```
Position.directionOffset = {
	[DIRECTION_NORTH] = Position(0, 65535, 0),
	[DIRECTION_EAST] = Position(1, 0, 0),
	[DIRECTION_SOUTH] = Position(0, 1, 0),
	[DIRECTION_WEST] = Position(65535, 0, 0),
	[DIRECTION_SOUTHWEST] = Position(65535, 1, 0),
	[DIRECTION_SOUTHEAST] = Position(1, 1, 0),
	[DIRECTION_NORTHWEST] = Position(65535, 65535, 0),
	[DIRECTION_NORTHEAST] = Position(1, 65535, 0),
}
```
By defining it like that you can see, how **ugly** this hack actually is.

To be entirely honest about how this should be handled. TFS should have `Point` (or something with a better name) metatable, which would define `__add` `__sub` and other operators to work directly with positions, in this case we could simply do `Point(0, 2)` or `Point(-1, 0, 0)` and add it to position metatable. It would accept signed integers, so no issues with that. Setting -1 to Position is an ugly and weird hack to make it seem like its an offset.

data/lib/core/position.lua
ranisalt: > It seems that Position:getNextPosition(direction) applies the offsets to itself

Ah this awful behavior. But you can clone the position:
```lua
local position = Position(self)
position:getNextPosition(direction)
```

omarcopires: > > It seems that Position:getNextPosition(direction) applies the offsets to itself
> 
> Ah this awful behavior. But you can clone the position:
> 
> ```lua
> local position = Position(self)
> position:getNextPosition(direction)
> ```

works fine

MillhioreBT (approved):

EPuncker (approved):

PR #4326, opened 2023-02-18
added 3 party events

reyaleman:
### Pull Request Prelude

### Changes Proposed

Added 3 party events

- Party.onInvite(self, player)
- Party.onRevokeInvitation(self, player)
- Party.onPassLeadership(self, player)

**Issues addressed:** 

### DIFF
--- a/data/events/events.xml
+++ b/data/events/events.xml
@@ -11,6 +11,9 @@
 	<event class="Party" method="onLeave" enabled="0" />
 	<event class="Party" method="onDisband" enabled="0" />
 	<event class="Party" method="onShareExperience" enabled="1" />
+	<event class="Party" method="onInvite" enabled="0" />
+	<event class="Party" method="onRevokeInvitation" enabled="0" />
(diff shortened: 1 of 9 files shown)


MillhioreBT:
I like `onInvite`, makes a lot of sense.
However the other two methods, mmmm are not very consistent.
I don't see valid reasons to use them.

Deny a player to revoke an invitation? because?
Deny one player passing the leadership to another? because?

Suppose I return false in the `Party.onPassLeadership` event then I can't make anyone else a leader anymore, but if the leader dies or the player gets kicked off the server what happens? rule is violated, leadership now belongs to the next member...

I think my mind is very closed right now, so I hope you can show some good examples to learn a bit about how to use these two methods.

Tofame:
@MillhioreBT deny...? Are those only for denying, or are those to run a script when something is happening e.g. when inviting.
If the latter then those could be useful for writing some otclient modules.

EPuncker (approved):

PR #4332, opened 2023-02-21
Add option to load offline player in lua

alfuveam:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 

If you need a send parcel or other things in **.lua script** to offline player has a issues.

1. In this exemple a mis position of mailbox cause a item lost.
```lua
   local item_parcel = Game.createItem(ITEM_PARCEL, 1)
   local item_label = Game.createItem(ITEM_LABEL, 1)
   item_label:setAttribute(ITEM_ATTRIBUTE_TEXT, "Alfuveam")
   item_parcel:addItemEx(item_label)
   item_parcel:addItemEx(item)
   item_parcel:moveTo(Position(MAILBOX_POSITION))
```

2. Player need to be online to send a item to mailbox in .lua if don't use a example 1.
```lua
   local player = Player("id or guid or name or userdata")
   local depot = player:getInbox()
   local addItem = depot:addItemEx(item, INDEX_WHEREEVER, FLAG_NOLIMIT)
```
**Look to the past:**

In source files when a online playerOne send a parcel to offline playerTwo the server load player to get a inbox , look this function Mailbox::sendItem.
In market has a same implementation, example in function IOMarket::processExpiredOffers load player to send money and item.

**The implementation**

First check if player is in memory to prevent a no necessary database call.
To save all changes in player object is necessary a call :save() 

1. Example to send things to offline player in .lua.
```lua
   local player = loadPlayer("id, guid or name")
   local depot = player:getInbox()
   local addItem = depot:addItemEx(item, INDEX_WHEREEVER, FLAG_NOLIMIT)
   player:save()
```

**Side effects**

- [ ] Create infinite player objects if found player in database?

**Bug found in tests**
- [ ] If call a method :remove() the server crash, need a set a virtual position.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -1107,6 +1107,9 @@ void LuaScriptInterface::registerFunctions()
 	// sendGuildChannelMessage(guildId, type, message)
 	lua_register(luaState, "sendGuildChannelMessage", LuaScriptInterface::luaSendGuildChannelMessage);
 
+	//loadPlayer(id, guid or name)
+	lua_register(luaState, "loadPlayer", LuaScriptInterface::luaLoadPlayer);
+
 	// isScriptsInterface()
 	lua_register(luaState, "isScriptsInterface", LuaScriptInterface::luaIsScriptsInterface);
 
@@ -4088,6 +4091,44 @@ int LuaScriptInterface::luaSendGuildChannelMessage(lua_State* L)
 	channel->sendToAll(message, type);
 	pushBoolean(L, true);
 	return 1;
+}
+
+int LuaScriptInterface::luaLoadPlayer(lua_State* L)
+{
+	// loadPlayer(id, guid or name)
+	Player* player;
+	bool player_found = true;
+	if (isNumber(L, 1)) {
+		const uint32_t playerId = getNumber<uint32_t>(L, 1);
+		if (playerId >= CREATURE_ID_MIN && playerId <= Player::playerIDLimit) {
+			player = g_game.getPlayerByID(playerId);
+		} else {
+			player = g_game.getPlayerByGUID(playerId);
+		}		
+		if(!player) {
+			player = new Player(nullptr);
+			if(!IOLoginData::loadPlayerById(player, playerId)){
+				player_found = false;
+			}
+		}
+	} else if (isString(L, 1)) {
+		const std::string playerName = getString(L, 1);
+		player = g_game.getPlayerByName(playerName);
+		if (!player) {
+			player = new Player(nullptr);
+			if(!IOLoginData::loadPlayerByName(player, playerName)){
+				player_found = false;
+			}
+		}
+	}
+
+	if (player_found) {
+		pushUserdata<Player>(L, player);
+		setMetatable(L, -1, "Player");
+	} else {
+		lua_pushnil(L);
+	}
+	return 1;	
 }
 
 int LuaScriptInterface::luaIsScriptsInterface(lua_State* L)
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -452,6 +452,7 @@ private:
 
 	static int luaSendChannelMessage(lua_State* L);
 	static int luaSendGuildChannelMessage(lua_State* L);
+	static int luaLoadPlayer(lua_State* L);
 
 	static int luaIsScriptsInterface(lua_State* L);
 


Zbizu:
you need to create `:unload()` method to destroy the object after it's no longer in use

EPuncker:
don't forget to check code style with clang

conde2:
Better naming would be OfflinePlayer instead of TempPlayer

src/luascript.cpp
```diff
@@ -1107,6 +1107,9 @@ void LuaScriptInterface::registerFunctions()
 	// sendGuildChannelMessage(guildId, type, message)
 	lua_register(luaState, "sendGuildChannelMessage", LuaScriptInterface::luaSendGuildChannelMessage);
 
+	//loadPlayer(id, guid or name)
+	lua_register(luaState, "loadPlayer", LuaScriptInterface::luaLoadPlayer);
```
nekiro, line 1111: this is not neeeded, delete

src/luascript.cpp
```diff
@@ -2707,6 +2710,12 @@ void LuaScriptInterface::registerFunctions()
 
 	registerMethod("Player", "getIdleTime", LuaScriptInterface::luaPlayerGetIdleTime);
 
+	// TempPlayer
+	registerClass("TempPlayer", "Player", LuaScriptInterface::luaTempPlayerCreate);
```
nekiro, line 2714: name it OfflinePlayer for clarity

src/luascript.cpp
```diff
@@ -4025,6 +4034,44 @@ int LuaScriptInterface::luaSendGuildChannelMessage(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaLoadPlayer(lua_State* L)
```
nekiro, line 4037: as above not needed

src/luascript.cpp
```diff
@@ -10824,6 +10871,55 @@ int LuaScriptInterface::luaPlayerGetIdleTime(lua_State* L)
 	return 1;
 }
 
+//	TempPlayer
```
nekiro, line 10874: remove tab

src/luascript.cpp
```diff
@@ -10824,6 +10871,55 @@ int LuaScriptInterface::luaPlayerGetIdleTime(lua_State* L)
...
+int LuaScriptInterface::luaTempPlayerCreate(lua_State* L)
+{
+	// TempPlayer(id or name)
+	Player* player = new Player(nullptr);
+	bool player_found = true;
```
nekiro, line 10879: camelCase
also this is leaking memory.

I would suggest code:
```c++
std::unique_ptr<Player> player = std::make_unique<Player>(nullptr);
if (isNumber(L, 2)) {
	const uint32_t playerId = getNumber<uint32_t>(L, 2);
	if (!IOLoginData::loadPlayerById(player, playerId)) {
		lua_pushnil(L);
		return 1;
	}
} else if (isString(L, 2)) {
	const std::string& playerName = getString(L, 2);
	if (!IOLoginData::loadPlayerByName(player, playerName)) {
		lua_pushnil(L);
		return 1;
	}
}

if (player) {
	pushUserdata<Player>(L, player.release());
	setMetatable(L, -1, "OfflinePlayer");
} else {
	lua_pushnil(L);
}
```

src/luascript.cpp
```diff
@@ -10824,6 +10871,55 @@ int LuaScriptInterface::luaPlayerGetIdleTime(lua_State* L)
...
+	} else {
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
+int LuaScriptInterface::luaTempPlayerDelete(lua_State* L)
+{
+	// tempPlayer:free()
+	Creature** creaturePtr = getRawUserdata<Creature>(L, 1);
```
nekiro, line 10904: why not getting player as the userdata?

nekiro (changes requested):

src/luascript.cpp
alfuveam: ok

src/luascript.cpp
```diff
@@ -10824,6 +10830,45 @@ int LuaScriptInterface::luaPlayerGetIdleTime(lua_State* L)
 	return 1;
 }
 
+// OfflinePlayer
+int LuaScriptInterface::luaOfflinePlayerCreate(lua_State* L)
```
MillhioreBT, line 10834: 1) If you pass a value other than a number or string then a null Player is returned to Lua, (I don't know how that behaves, but it shouldn't be like that).
2) You can get players that are online, it doesn't make sense because this object is supposed to be for offline players.
3) Use camelCase format.

Example:
```c++
// OfflinePlayer
int LuaScriptInterface::luaOfflinePlayerCreate(lua_State* L)
{
	// OfflinePlayer(guid or name)
	auto offlinePlayerPtr = std::make_unique<Player>(nullptr);
	if (const uint32_t guid = getNumber<uint32_t>(L, 2)) {
		if (g_game.getPlayerByGUID(guid) || !IOLoginData::loadPlayerById(offlinePlayerPtr.get(), guid)) {
			lua_pushnil(L);
			return 1;
		}
	} else if (const std::string& name = getString(L, 2); !name.empty()) {
		if (g_game.getPlayerByName(name) || !IOLoginData::loadPlayerByName(offlinePlayerPtr.get(), name)) {
			lua_pushnil(L);
			return 1;
		}
	}

	if (offlinePlayerPtr) {
		pushUserdata<Player>(L, offlinePlayerPtr.release());
		setMetatable(L, -1, "OfflinePlayer");
	} else {
		lua_pushnil(L);
	}
	return 1;
}
```

src/luascript.cpp
```diff
@@ -10824,6 +10830,45 @@ int LuaScriptInterface::luaPlayerGetIdleTime(lua_State* L)
...
+			return 1;
+		}
+	}
+
+	pushUserdata<Player>(L, player_ptr.release());
+	setMetatable(L, -1, "OfflinePlayer");
+	return 1;
+}
+
+int LuaScriptInterface::luaOfflinePlayerDelete(lua_State* L)
```
MillhioreBT, line 10857: `delete` instead `free`, just for consistency with TFS nomenclature.

In my personal opinion I would use `remove` this way I override the `remove` method of Creatures, since this method causes interpreter crashes with this object OfflinePlayer, so no one would use it accidentally.

Example:
```c++
registerMetaMethod("OfflinePlayer", "__gc", LuaScriptInterface::luaOfflinePlayerRemove);

registerMethod("OfflinePlayer", "remove", LuaScriptInterface::luaOfflinePlayerRemove);

int LuaScriptInterface::luaOfflinePlayerRemove(lua_State* L)
{
	// offlinePlayer:remove() offlinePlayer:__gc()
	Player** offlinePlayerPtr = getRawUserdata<Player>(L, 1);
	if (auto offlinePlayer = *offlinePlayerPtr) {
		g_game.ReleaseCreature(offlinePlayer);
		*offlinePlayerPtr = nullptr;
		pushBoolean(L, true);
		return 1;
	}

	lua_pushnil(L);
	return 1;
}
```

If this object is based on the Player/Creaturewe should treat the object as such, so use `g_game.ReleaseCreature`

However, I'm not from the development team so don't take my suggestions too seriously and my personal opinions 😊.

nekiro: ReleaseCreature is not used for creatures that are loaded like that, they are not stored anywhere so there is no need to delay the deletion. Delete operator is ok

Issue #4304, opened 2023-01-28
/reload libs breaks 'lookAt' funcionality


xmish:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. type `/reload libs` (`Reloaded libs.` shows)
2. `onLook` event callback no longer works (`function Player:onLook(thing, position, distance)` is called but results with empty description)
4. No text is shown on look at items

### Expected behaviour

Reloads does not break `lookAt` functionality
### Actual behaviour

No lookAt text is shown after `reload libs`.
### Environment

- Windows 10 x64 (10.0.19045.2486)
- x64 Debug build of latest commit from master: #4289 45f59f6a7d131a1c4dcf5ff50a225a6199f17230

PR #4343, opened 2023-03-08
Fix reload global

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Because the global reload in [luascript.cpp](https://github.com/otland/forgottenserver/blob/dccc3bcc4be7e712dbcd6daf12c14553da604589/src/luascript.cpp#L4926) reloads the internal libraries of the revscript environment, [the tables where the eventcallbacks are stored are cleaned up](https://github.com/otland/forgottenserver/blob/master/data/scripts/lib/event_callbacks.lua), so we must reload the scripts after the global reload.

I also modified the code of the `Game.relaod` method a bit so that it would return 1 or 2 values ​​as the case may be, since previously a boolean was being pushed to the stack that was not even returned, [because only 1 value was being returned](https://github.com/otland/forgottenserver/blob/dccc3bcc4be7e712dbcd6daf12c14553da604589/src/luascript.cpp#L4931).

**Issues addressed:** #4304

According to the Issue lookAt fails, and yes, it's true, but so did all the related events, so it's a big problem if someone uses reloaded global 😩

### DIFF
--- a/data/talkactions/scripts/reload.lua
+++ b/data/talkactions/scripts/reload.lua
@@ -79,6 +79,11 @@ function onSay(player, words, param)
 	end
 
 	Game.reload(reloadType)
+	if reloadType == RELOAD_TYPE_GLOBAL then
+		-- we need to reload the scripts as well
+		Game.reload(RELOAD_TYPE_SCRIPTS)
+	end
+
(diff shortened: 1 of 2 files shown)


EPuncker (approved):

EPuncker:
@xmish can you check this please?

Issue #4199, opened 2022-08-22
[Discussion] getNumber and getBoolean defaultRet


conde2:
What I would like to discuss here is if the function getNumber and getBoolean  should return the default value when the argument passed through lua is nil. What is the impact of such change?

foo(param1, [param2, [param3]])
foo(1, nil, 3) -- in this case the default value will not be assigned to param2, instead it will assign 0

```cpp
	template <typename T>
	static T getNumber(lua_State* L, int32_t arg, T defaultValue)
	{
		const auto parameters = lua_gettop(L);
		if (parameters == 0 || arg > parameters) {
			return defaultValue;
		}
		return getNumber<T>(L, arg);
	}
```

Solution
```cpp
	template <typename T>
	static T getNumber(lua_State* L, int32_t arg, T defaultValue)
	{
		const auto parameters = lua_gettop(L);
		if (parameters == 0 || arg > parameters || lua_isnil(L, arg)) {
			return defaultValue;
		}
		return getNumber<T>(L, arg);
	}
```

https://github.com/otland/forgottenserver/blob/master/src/luascript.h#L280

ranisalt:
I would rather implement that behavior in Lua, but since we don't raise exceptions on failure, one can just use `getNumber(x) or defaultValue` for the same behavior.

MillhioreBT:
If we put ourselves in that we would also have to verify if it is a table, function, string or boolean.

If someone used the `string.format` like this: `string.format(nil, "name")` maybe you should learn the basics of Lua, before trying to use the language, the TFS API is not strongly protected against human errors.

I am in favor of protecting ourselves from all kinds of errors, but it is true, adding many checks to getNumber could result in milliseconds of extra delays in most scripts.

You can work the default value from lua:
```lua
local n = tonumber(var) or 0
```

conde2:
> I would rather implement that behavior in Lua.
> 
> What is the use case you have in mind?

I'm implementing a HTTP Client in TFS, and I have a struct with default values for the HTTP Client and requests. My function receive a Lua table with the configs (similar what is done with outfits) and fills up this struct. I'm using the getField and if the field is nil then I use the defaultValue so I don't need to fill my Lua table with all config settings.

I will not bother answering the other guy above.

MillhioreBT:
easy

example:
```lua
foo(param1, [param2, [param3]])
foo(param1, tonumber(param2) or YOUR_DEFAULT_VALUE, param3) -- no sources changes required.
```

[...]

PR #4348, opened 2023-03-14
A better check for getNumber/getBoolean with defaultRet

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Currently the check evaluates if `arg` is greater than the stack size or if the stack is empty, which is wrong in many cases, this is used so rarely that I didn't even realize it, but it makes particularly sense. when dealing with `defaultRet`

**Issues addressed:** #4199

### DIFF
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -276,8 +276,7 @@ public:
 	template <typename T>
 	static T getNumber(lua_State* L, int32_t arg, T defaultValue)
 	{
-		const auto parameters = lua_gettop(L);
-		if (parameters == 0 || arg > parameters) {
+		if (lua_isnumber(L, arg) == 0) {
 			return defaultValue;
 		}
 		return getNumber<T>(L, arg);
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

ranisalt (approved):
Neat, so we don't need to check if `arg` is a valid stack index?

EPuncker:
thoughts @conde2 ?

conde2:
I would check for nil or not present parameter and then return the default value, otherwise if the argument is from a different type than call reportErrorFunc with an error telling the argument type is wrong.

For example:
```C++
		if (lua_isnoneornil(L, arg) == 1) {
			return defaultValue;
		}

		if (lua_isnumber(L, arg) == 0) {
			reportErrorFunc(L,
			                fmt::format("Argument {} has wrong value type {} expected number", arg, lua_typename(L, lua_type(L, arg))));
			return defaultValue;
		}
```

MillhioreBT:
> Neat, so we don't need to check if `arg` is a valid stack index?

No need, lua already does that. if the index is invalid return 0
if the index is valid and there is no numeric data type return 0.
otherwise return 1

PR #4351, opened 2023-03-18
Fixed exp stages ignoring last stage

EPuncker:
### Pull Request Prelude

### Changes Proposed
Fixed exp stages ignoring last stage

### Context

> if you have x5 in config
> (not staged rate)
> last stage will be ignored and use that value instead
> because it checks min < level < max
> and max is 0
> in last stage
> so it never satisfies
> and defaults to RATE_EXP

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -90,9 +90,10 @@ ExperienceStages loadLuaStages(lua_State* L)
 	lua_pushnil(L);
 	while (lua_next(L, -2) != 0) {
 		const auto tableIndex = lua_gettop(L);
-		auto minLevel = LuaScriptInterface::getField<uint32_t>(L, tableIndex, "minlevel");
-		auto maxLevel = LuaScriptInterface::getField<uint32_t>(L, tableIndex, "maxlevel");
-		auto multiplier = LuaScriptInterface::getField<float>(L, tableIndex, "multiplier");
+		auto minLevel = LuaScriptInterface::getField<uint32_t>(L, tableIndex, "minlevel", 1);
+		auto maxLevel =
+		    LuaScriptInterface::getField<uint32_t>(L, tableIndex, "maxlevel", std::numeric_limits<uint32_t>::max());
+		auto multiplier = LuaScriptInterface::getField<float>(L, tableIndex, "multiplier", 1);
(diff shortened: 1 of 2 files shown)


ranisalt:
I would fix it differently, by setting the default to maximum possible value when loading:

```cpp
			if (auto maxLevelAttribute = stageNode.attribute("maxlevel")) {);
				maxLevel = pugi::cast<uint32_t>(maxLevelAttribute.value());
			} else {
				maxLevel = std::numeric_limits<uint32_t>::max();
			}
```

To avoid branching

Zbizu:
speaking of exp stages,
these three variables probably aren't used by anything anymore:
https://github.com/otland/forgottenserver/blob/e288183a5a459f5c5f7a470dcd9001e6cd745de0/src/game.h#L586

ranisalt:
@EPuncker the code goes in loading the stages, not where you put it

src/configmanager.cpp
```diff
@@ -328,7 +328,8 @@ bool ConfigManager::getBoolean(boolean_config_t what) const
 float ConfigManager::getExperienceStage(uint32_t level) const
 {
 	auto it = std::find_if(expStages.begin(), expStages.end(), [level](ExperienceStages::value_type stage) {
```
ranisalt, line 330: While at that you can change it here to use `auto`
```suggestion
	auto it = std::find_if(expStages.begin(), expStages.end(), [level](auto stage) {
```

And I THINK this is possible:
```suggestion
	auto it = std::find_if(expStages.begin(), expStages.end(), [level](auto&& [minLevel, maxLevel]) {
```
or
```suggestion
	auto it = std::find_if(expStages.begin(), expStages.end(), [level](auto stage) {
            auto&& [minLevel, maxLevel] = stage;
```

EPuncker: feel free to commit on this branch (please)

ranisalt (approved):
image

nekiro:
I think it still doesnt work. With last stages without maxLevel it returned 0 for me, which failed to find the stage and defaulted back to flat expRate.
I added check for 0 and it works, but most likely that getField doesnt actually do what it should do with that defaultValue

PR #4355, opened 2023-03-24
[New Methods] get/set exhaustion

omarcopires:
### Pull Request Prelude

### Changes Proposed 
This pull request adds two new functions to the Player class: setExhaustion and getExhaustion. These functions allow for easy management of player exhaustion values, which can be useful for limiting the frequency of certain actions or events in a game. The setExhaustion function takes two arguments: the value to set and the duration of the exhaustion in seconds. The getExhaustion function takes one argument, the value to retrieve, and returns the remaining duration of the exhaustion in seconds.

**Example usage:**
Set a 10 second exhaustion for the player
```lua
player:setExhaustion(1234, 10)
```

Check if the player is still exhausted for value 1234
```lua
if player:getExhaustion(1234) > 0 then
    player:sendTextMessage(MESSAGE_EVENT_ADVANCE, "You are still exhausted.")
end
```

### DIFF
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -393,3 +393,15 @@ function Player.isPromoted(self)
 	local fromVocId = vocation:getDemotion():getId()
 	return vocation:getId() ~= fromVocId
 end
+
+function Player.setExhaustion(self, value, time)
+	return self:setStorageValue(value, time + os.time())
+end
+
+function Player.getExhaustion(self, value)
+	local storage = self:getStorageValue(value)
+	if storage <= 0 then
+		return 0
+	end
+	return storage - os.time()
+end
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -394,14 +394,10 @@ function Player.isPromoted(self)
 	return vocation:getId() ~= fromVocId
 end
 
-function Player.setExhaustion(self, value, time)
-	return self:setStorageValue(value, time + os.time())
+function Player.setExhaustion(self, key, seconds)
+	return self:setStorageValue(key, os.time() + seconds)
 end
 
-function Player.getExhaustion(self, value)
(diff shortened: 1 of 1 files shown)


MillhioreBT:
There's nothing wrong with it, but it would look nicer like this:
```lua
function Player.setExhaustion(self, key, seconds)
	return self:setStorageValue(key, os.time() + seconds)
end

function Player.getExhaustion(self, key)
	return math.max(self:getStorageValue(key) - os.time(), 0)
end
```

omarcopires:
> There's nothing wrong with it, but it would look nicer like this:
> 
> ```lua
> function Player.setExhaustion(self, key, seconds)
> 	return self:setStorageValue(key, os.time() + seconds)
> end
> 
> function Player.getExhaustion(self, key)
> 	return math.max(self:getStorageValue(key) - os.time(), 0)
> end
> ```

I applied the suggestions, but there is no way for me to test its operation at the moment.

DSpeichert:
The name may be confusing - it appears to be related to the game's exhaust mechanism, but it's not.

ranisalt:
What's that for? I didn't get it, if it's just a helper to set a storage to a relative point in time, I would just name it `setStorageTimeInterval` or something more specific, as @DSpeichert pointed out it is confusing.

Also, I would prefer that it used milliseconds instead, it opens up more possibilities

omarcopires:
> What's that for? I didn't get it, if it's just a helper to set a storage to a relative point in time, I would just name it `setStorageTimeInterval` or something more specific, as @DSpeichert pointed out it is confusing.
> 
> Also, I would prefer that it used milliseconds instead, it opens up more possibilities

I just imported the function from another datapack, and I noticed that it is used in some rlmap and custom maps that have leverage systems for bosses, I don't know if it would be ideal to change the name of the functions. perhaps adding comments explaining would be enough.

Issue #4359, opened 2023-03-29
Cant login to game when server is built in debug mode


ArturKnopik:
Build server on debug mode and try login (windows, VS Studio 19)

Character name: God
![image](https://user-images.githubusercontent.com/17506599/228654614-a03efa8b-2e87-499c-b422-d9b95e7b12b3.png)

Note:
![image](https://user-images.githubusercontent.com/17506599/228655502-a06e5095-573e-41f5-bd79-094352a2e182.png)
result->getString("name") returns a garbage value
characterName has valid character name

Note:
Release mode works fine

ArturKnopik:
@ranisalt i have question related to this defect and string_view changes, did you try test string_view on debug mode? 
I assume this only works by compiler optimizations... 
of course, I'm not an expert, but rather a beginner, so I won't argue, after the merge of the last changes, everything broke
Also bestiary after master merge somehow become broken (i cant retrive race name from network msg)

PR #4360, opened 2023-04-01
Fix string view in game world auth

ArturKnopik:
Fix string_view pointing to non-existent string
### Pull Request Prelude

**Issues addressed:** 
https://github.com/otland/forgottenserver/issues/4359

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -92,7 +92,7 @@ bool IOLoginData::loginserverAuthentication(const std::string& name, const std::
 	return true;
 }
 
-std::pair<uint32_t, std::string_view> IOLoginData::gameworldAuthentication(std::string_view accountName,
+GameWorldAuthenticationData_t IOLoginData::gameworldAuthentication(std::string_view accountName,
                                                                            std::string_view password,
                                                                            std::string_view characterName,
                                                                            std::string_view token, uint32_t tokenTime)
@@ -102,7 +102,7 @@ std::pair<uint32_t, std::string_view> IOLoginData::gameworldAuthentication(std::
 	    fmt::format("SELECT `id`, `password`, `secret` FROM `accounts` WHERE `name` = {:s} OR `email` = {:s}",
 	                db.escapeString(accountName), db.escapeString(accountName)));
 	if (!result) {
-		return std::make_pair(0, characterName);
+		return {0, ""};
 	}
 
 	// two-factor auth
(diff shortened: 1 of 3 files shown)


ranisalt:
No need for most of these changes; just changing the return type from `string_view` to `string` in `gameworldAuthentication` and `move`ing it into the lambda is enough.

ArturKnopik:
i tried it your way but i failed, i don't know why but i got an empty string after escaping gameworldAuthentication method(int onRecvFirstMessage method), before adding the string to the pair the string had a valid value

ArturKnopik:
updated, issue with @ranisalt approach was std::tie and characterName variable, characterName was string_view not std::string

reyaleman:
> updated, issue with @ranisalt approach was std::tie and characterName variable, characterName was string_view not std::string

I just changed std::tie to std::pair on protocolgame.cpp (L472) and it worked 😕 


```cpp
std::tie(accountId, characterName) = ...
```

```cpp
std::pair(accountId, characterName) = ...
```

ranisalt:
I would try with

```
auto [accountId, name] = gameworldAuthentication(...);
```

then `move` `name` into the lambda if necessary.

> `std::pair(accountId, characterName) = ...`

Did it _compile_ or did it _work_? I think that's creating a new pair with `accountId` (which is zero) and `characterName` (the unnormalized name) and then assigning it, but this shouldn't override the previous values, and you'll get the `accountId == 0` and the unmodified `characterName` instead.

reyaleman:
> > `std::pair(accountId, characterName) = ...`
> 
> Did it _compile_ or did it _work_? I think that's creating a new pair with `accountId` (which is zero) and `characterName` (the unnormalized name) and then assigning it, but this shouldn't override the previous values, and you'll get the `accountId == 0` and the unmodified `characterName` instead.

you're right... it doesn't override previous values for `accountId` and `characterName`.

Ignore my previous comment 👍

ranisalt:
To be honest, reading the character name would just prevent a hopelessly broken client from not being able to log in, because by default the Tibia client and OTClient send the name exactly as the login server responds.

We can skip that code and return solely the account ID, the code gets much simpler. Along these lines:

```cpp
	uint32_t accountId = IOLoginData::gameworldAuthentication(accountName, password, characterName, token, tokenTime);
	if (accountId == 0) {
		disconnectClient("Account name or password is not correct.");
		return;
	}

	g_dispatcher.addTask([=, thisPtr = getThis(), characterName = std::move(characterName)]() {
		thisPtr->login(characterName, accountId, operatingSystem);
	});
```

Alternatively, we can instead return the character ID and not the name, and then `login` takes it, since the very first thing it does is `getPlayerByName`.

What do you think?

ArturKnopik:
characterName is string_view, string_view is moveable? I don't think so... but i can be wrong

ranisalt:
> characterName is string_view, string_view is moveable? I don't think so... but i can be wrong

Right, it is not, so keep `std::string{characterName}` to copy it

PR #4362, opened 2023-04-02
Hide unknow packet warning

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Disable by default the Unknow Packet warning preventing spam in the console.

image


ranisalt:
You could create no-op packet handlers for these packet types instead:

```lua
local noop = PacketHandler(0x1C)
function noop.onReceive() end
noop.register()
```

omarcopires:
> You could create no-op packet handlers for these packet types instead:

Thank you for suggesting this approach. While creating no-op packet handlers could indeed help to hide missing packets, it would only be a workaround rather than a proper solution, as it wouldn't actually add the missing packets to the codebase. Moreover, it could make the code harder to maintain and understand in the long term, as it would introduce unnecessary complexity and potentially obscure the actual packet flow.

Instead, I believe the best approach would be to hide the warning message only in production, and keep it visible in debugging mode, so that developers can identify missing packets and add them properly, as you mentioned. This would ensure a cleaner and more reliable codebase, while still allowing for the necessary flexibility to add new features.

That being said, I respect your opinion and appreciate your input. However, I won't be implementing this approach in this pull request, as I believe there are better ways to handle missing packets in the codebase.

MillhioreBT:
> > You could create no-op packet handlers for these packet types instead:
> 
> Thank you for suggesting this approach. While creating no-op packet handlers could indeed help to hide missing packets, it would only be a workaround rather than a proper solution, as it wouldn't actually add the missing packets to the codebase. Moreover, it could make the code harder to maintain and understand in the long term, as it would introduce unnecessary complexity and potentially obscure the actual packet flow.
> 
> Instead, I believe the best approach would be to hide the warning message only in production, and keep it visible in debugging mode, so that developers can identify missing packets and add them properly, as you mentioned. This would ensure a cleaner and more reliable codebase, while still allowing for the necessary flexibility to add new features.
> 
> That being said, I respect your opinion and appreciate your input. However, I won't be implementing this approach in this pull request, as I believe there are better ways to handle missing packets in the codebase.

I like your way of thinking, and by the way the suggestion that ranisalt told you, was missing the two points:
```lua
noop:register() -- instead noop.register()
```

ranisalt:
@omarcopires my point is that it does not make sense to move packet handling to Lua, then create a flag in C++ to pass it to Lua to disable logging. Just comment out that line, I would approve it. We can uncomment it when debugging it necessary :)

EPuncker (approved):

ranisalt (approved):

PR #4366, opened 2023-04-02
Perform gameworld authentication using character ID

### LINKED ISSUE #4359: Cant login to game when server is built in debug mode

ranisalt:
### Pull Request Prelude

### Changes Proposed

Instead of performing gameworld authentication and login using the character name, instead read the character ID as early as possible and use that instead. This reduces the need to pass strings around.

**Issues addressed:**
Fixes #4359 
Closes #4360

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -92,17 +92,17 @@ bool IOLoginData::loginserverAuthentication(const std::string& name, const std::
 	return true;
 }
 
-std::pair<uint32_t, std::string_view> IOLoginData::gameworldAuthentication(std::string_view accountName,
-                                                                           std::string_view password,
-                                                                           std::string_view characterName,
-                                                                           std::string_view token, uint32_t tokenTime)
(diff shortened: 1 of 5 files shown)


EPuncker (approved):

ranisalt:
> I think this need to be more benchmarked, because having to perform a JOIN is probably worse then passing some strings around.

This is doing a join on indexed columns. It will be equal or faster than running two separate queries.

I did the test anyway, and on my machine the join takes as much time as each of the separate queries; therefore it halves the running time.

ranisalt:
@ArturKnopik could you check if this also fixes #4359 or if #4360 is still required?

krecikondexin:
@ranisalt i will check it today

ArturKnopik:
Debug build on windows works without issue, i can login to the game server without any problem

ranisalt:
Thanks for the review!

Issue #4288, opened 2023-01-05
Memory leaks


ArturKnopik:
### Before creating an issue, please ensure:
- [X] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [X] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
cmake -DCMAKE_BUILD_TYPE=Debug ..
make
<copy tfs to valid folder>
valgrind --leak-check=full ./tfs
wait for full startup
ctrl+c

### Expected behaviour
no memory leak

### Actual behaviour
memory leaks
[leaks.txt](https://github.com/otland/forgottenserver/files/10356258/leaks.txt)

### Environment
VirtualBox
artur@artur-VirtualBox:/media/sf_forgottenserver_AK$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy

EPuncker:
@ranisalt can you check this please?

ranisalt:
@EPuncker definitely, when I return from vacation by the end of the month :) it's one of my main objectives around here, fixing these leaks

ArturKnopik:
These leaks are at server startup and shutdown, I haven't checked how it looks from the player's perspective (when the player moves, performs actions, etc...), we have a lot of lambdas that have not been checked

Zbizu:
> These leaks are at server startup and shutdown, I haven't checked how it looks from the player's perspective (when the player moves, performs actions, etc...), we have a lot of lambdas that have not been checked

what about reload?

ArturKnopik:
will be nice to have some list of features to test with description how to test it if needed 
i expect really big list

Zbizu:
@ArturKnopik 
from what I see in the lambda commit itself:
1. going to sleep in bed
2. send guild motd to player
3. connection opening and closing
4. parse packet and handling individual packets (sending instructions to game thread)
5. connection::internalSend
6. creature walk
7. "extra swing" (stairhop?)
8. creature death
9. creature::addCondition
10. 3x `g_scheduler.addEvent(createSchedulerTask(walkDelay, [=, id = getID()] () { g_game.forceRemoveCondition(id, type); }));`
11. databasetasks::runtask
12. checkSpawn cycle
13. raid system
14. login/relog
15. game/player cpp files
[...]

PR #4380, opened 2023-04-06
Raise C++ standard to C++20

ranisalt:
### Pull Request Prelude

### Changes Proposed

Raise the required C++ standard from C++17 to C++20. It has the following benefits (not exhaustive):

- Potentially improved compile times using concepts instead of SFINAE for code where we use `std::enable_if` 
- More expressive code through the use of [ranges](https://en.cppreference.com/w/cpp/ranges) (not range-for), views and [spans](https://en.cppreference.com/w/cpp/container/span)
- Native `std::format` that can replace fmt
- `constexpr if` and `consteval` to perform more work during compile time
- Coroutines, which can help writing generator functions where we currently generate a whole container of items before processing it (e.g. recursively navigating a container)

See more [language features](https://en.cppreference.com/w/cpp/20) and [compiler support](https://en.cppreference.com/w/cpp/compiler_support/20).

Minimum compiler version with decent C++20 support is gcc 8 and clang 6, with gcc 9 and clang 8 being mostly feature-complete.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,7 @@ option(ENABLE_TESTING "Build unit tests" OFF)
 
 include_directories(${Boost_INCLUDE_DIRS} ${Crypto++_INCLUDE_DIR} ${LUA_INCLUDE_DIR} ${MYSQL_INCLUDE_DIR} ${PUGIXML_INCLUDE_DIR})
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 add_subdirectory(src)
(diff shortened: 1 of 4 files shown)


Erza:
> Native std::format that can replace fmt
I wish. Last time I checked a few months ago, GCC still didn't support std::format. I switched to C++20 on a project because I wanted to use it, and then found out that GCC *still* did not support the C++20 text formatting at the end of 2022....

GCC 13 finally supports it now, but it's not even released yet.

EPuncker (changes requested):
missing vsproject changes

ranisalt:
> GCC 13 finally supports it now, but it's not even released yet.

It's just wishful thinking. We will still use libfmt, since we use colors, but maybe one day `std::format` matures enough :P hopefully the rest of the reasons are appealing enough

ArturKnopik:
a lot of problems with lambdas, maybe the introduction of the C++20 standard will fix memory leaks by the way (https://github.com/otland/forgottenserver/issues/4288)

EPuncker (approved):

PR #4385, opened 2023-04-07
Refactor variable name for Gamemaster outfit in player protocol

### LINKED ISSUE #4383: Crash when opening outfits window with GM

1. Login with a GM account.
2. Open customization window or ctrl + r.

### Expected behaviour 
The outfits and mounts window should open normally.

### Actual behaviour 
The client crashes when opening.

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Changed Gamemaster outfit creation to use a const static string for outfit name.

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -3234,7 +3234,8 @@ void ProtocolGame::sendOutfitWindow()
 
 	std::vector<ProtocolOutfit> protocolOutfits;
 	if (player->isAccessPlayer()) {
-		protocolOutfits.emplace_back("Gamemaster", 75, 0);
+		static const std::string gamemasterOutfitName = "Gamemaster";
+		protocolOutfits.emplace_back(gamemasterOutfitName, 75, 0);
 	}
 
 	protocolOutfits.reserve(outfits.size());


ranisalt:
Why?

omarcopires:
> Why?

The main reason for changing the code in this way is to improve code efficiency and readability.
By introducing a static constant variable with the gamemaster outfit name, we're reducing the number of times the string "Gamemaster" needs to be created and destroyed during program execution. This can be helpful in situations where creating temporary objects can lead to performance bottlenecks.
Additionally, introducing a variable with a descriptive name makes the code more readable and easier to understand. The use of descriptive variables can make the code easier to maintain, especially when multiple parts of the code need to reference the same string.
In summary, the proposed change is a good programming practice that improves code efficiency and readability.

ranisalt:
The literal string already lives in static storage and does not have to be instantiated. Storing in a `static std::string` moves instantiation to runtime, in fact losing performance and wasting more space.

I can't see how `gamemasterOutfitName` is less descriptive than `"Gamemaster"`, but I don't have a strong opinion to block based on that. In that case, `constexpr auto gamemasterOutfitName = "Gamemaster"` is better than a `static std::string`.

(also ChatGPT is too wordy, it writes too much to say too little)

PR #4261, opened 2022-12-11
Quests moved to Lua

MillhioreBT:
### Pull Request Prelude
- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
* Moved the Quest and Mission system to Lua completely.
* No compatibility with the old quests.xml file (If someone wants to do it in another PR, they can do it with the help of these methods included in this PR #4108)
* The Quest Tracker of each player that is tracking a quest will be reset if you use `/reload scripts or all` (This is because we don't want ghost quests that have been removed, changed, or replaced.)
* To make this possible, 4 new events have been created: (`onStorageUpdate`, `onQuestTracker`, `onQuestLog`, `onQuestLine`

If you have suggestions or find any bugs please let me know, I've tested it, but your revisions are always needed.

**Issues addressed:** Nothing!

PR #4388, opened 2023-04-07
Quest moved to lua (Part 2)

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
This is the new version of PR #4261, everything is neater and more consistent with namespaces as you suggested @ranisalt

The tracker now updates if the value is -1


EPuncker (approved):

omarcopires (approved):

dex-89:
I have only one question, what is the purpose of transferring quests and other elements from the engine to lua?

MillhioreBT:
> I have only one question, what is the purpose of transferring quests and other elements from the engine to lua?

The reason for moving this to lua are several:

* Quests in lua are versatile, you can create your own quests based on your own criteria and not just storages.
* Being in lua we are freeing load on sources, you don't need the wonderful low level power of C++ for this nonsense.
* The packet event manager in lua needs work and here we are giving you one.
* Everything that can be handled in lua should be in lua and not in the sources.
* Etc...

Zbizu (changes requested):

data/scripts/lib/event_callbacks.lua
```diff
@@ -57,6 +57,7 @@ ec.onLoseExperience = {[2] = 1}
 ec.onGainSkillTries = {[3] = 1}
 ec.onWrapItem = {}
 ec.onInventoryUpdate = {}
+ec.onUpdateStorage = {}
```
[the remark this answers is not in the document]
ranisalt: `onInventoryUpdate` is wrong. It's always `onVerbSubject` for all other events.

Zbizu:
setting storage value to `-1` does not update the quest in tracker

data/lib/core/quests.lua
```diff
@@ -0,0 +1,113 @@
...
+                return self.descriptions[current]
+            end
+        end
+    else
+        for current = self.endValue, self.startValue, -1 do
+            if value == current then
+                return self.descriptions[current]
+            end
+        end
+    end
```
Zbizu, line 110:
```suggestion
    if self.ignoreEndValue then
        for current = self.endValue, self.startValue, -1 do
            if value >= current and self.descriptions then
                return self.descriptions[current]
            end
        end
    else
        for current = self.endValue, self.startValue, -1 do
            if value == current and self.descriptions then
                return self.descriptions[current]
            end
        end
    end
```

this should get rid of this error
```
data/lib/core/quests.lua:107: attempt to index field 'descriptions' (a nil value)
stack traceback:
        [C]: in function '__index'
        data/lib/core/quests.lua:107: in function 'getDescription'
        data/lib/core/player.lua:752: in function 'sendQuestLine'
        ...cripts\eventcallbacks\player\default_onStorageUpdate.lua:43: in function 'callback'
        C:\ot\____SERVER\ots\data\scripts/lib\event_callbacks.lua:170: in function 'onQuestRequest'
        data/events/scripts/player.lua:658: in function <data/events/scripts/player.lua:656>
```

I'm not exactly sure how to reproduce it, here are my prompts
```
3:20 [0]: > Player("Player2"):setStorageValue(1003, 4)
03:20 [0]: > Player("Player2"):setStorageValue(1003, 3)
03:20 [0]: > Player("Player2"):setStorageValue(1003, 1)
03:20 [0]: > Player("Player2"):setStorageValue(1003, 2)
03:20 [0]: > Player("Player2"):setStorageValue(1002, 1)
```
and my config
```lua
Game.createQuest("Example Quest II", {
    storageId = 1002,
    storageValue = -1,

    missions = {
        {
            name = "Example Mission 1",
            storageId = 1002,
            startValue = -1,
            endValue = 10,
            ignoreEndValue = false,
            description = function(player)
                return string.format("Example description %d by %s.", player.storage[1003], player:getName())
            end
        },
        {
            name = "Example Mission 2",
            storageId = 1003,
            startValue = 1,
            endValue = 3,
            ignoreEndValue = false,
            description = "Mission 2"
        }
    }
}):register()
```

src/luascript.cpp
```diff
@@ -2112,6 +2112,8 @@ void LuaScriptInterface::registerFunctions()
 	registerEnumIn("configKeys", ConfigManager::SERVER_SAVE_SHUTDOWN);
 	registerEnumIn("configKeys", ConfigManager::ONLINE_OFFLINE_CHARLIST);
 	registerEnumIn("configKeys", ConfigManager::CHECK_DUPLICATE_STORAGE_KEYS);
+	registerEnumIn("configKeys", ConfigManager::QUEST_TRACKER_FREE_LIMIT);
+	registerEnumIn("configKeys", ConfigManager::QUEST_TRACKER_PREMIUM_LIMIT);
```
Zbizu, line 2116: this column is for boolean keys, the proposed config keys should be two blocks below (right above "os")

Zbizu (changes requested):

data/lib/core/quests.lua
MillhioreBT: I think -1 was never an accepted value, but it can be accepted by simply removing the "if value == -1 then" check in the onUpdateStorage event.

I also like the idea of ​​being able to set -1 by default to have quests that are always on and can be tracked from -1

data/scripts/quests/load_questsxml.lua
```diff
@@ -0,0 +1,40 @@
+local questXml = XMLDocument("data/XML/quests.xml")
```
ranisalt, line 1: What if the file does not exist?

data/scripts/eventcallbacks/player/default_onUpdateStorage.lua
```diff
@@ -0,0 +1,29 @@
...
+        return
+    end
+
+    local playerId = player:getId()
+    local now = os.mtime()
+    if not lastQuestUpdate[playerId] then
+        lastQuestUpdate[playerId] = now
+    end
+
+    if now - lastQuestUpdate[playerId] <= 0 and Game.isQuestStorage(key, value, oldValue) then
```
Zbizu, line 16: this never triggers, should be like this:
```suggestion
    if lastQuestUpdate[playerId] - now <= 0 and Game.isQuestStorage(key, value, oldValue) then
```

Zbizu (changes requested):

Zbizu (approved):
looks good so far (local tests)
If you wait one week, I can post results from production also.

EPuncker (approved):

EPuncker (approved):

Zbizu (approved):
I've ran it successfully on production for 2 days and haven't had any error reports or server issues related to it.

PR #4395, opened 2023-04-08
EventCallback Improvements, Oops is now called Event

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Finally a better name for this system, or rather, this should have always been the name, since we're really dealing with `Events`, and `Events` are already callbacks.

Here the optimization of PR #4237 is maintained, which by the way the idea of ​​maintaining a local was horrible hahahaha

A new global table (`hasEvent`) was added that functions as a proxy with a metatable with the metamethod __index to check if the events exist, thus maintaining the logic of not creating global IDs for each event.

The Events are now created from making a call to `Event`
```lua
local event = Event()
```
these objects can still be reused as before with the global `EventCallback` table

There is no compatibility with the older EventCallback, yes or yes you have to rename Event and call it to get the object.
Whoever wants to make the changes here is an example:
```lua
local ec = EventCallback
-- To
local event = Event()
```

**Notes:**
* Overall this PR makes ~EventCallback~ now feel more natural and consistent with Revscript.
* I don't know if others will, but I plan to update all the scripts that I have published in OTLand to be compatible with these changes.


MillhioreBT:
I have added a Global variable with the above name, for backwards compatibility.
now there is no need to change anything, your scripts will continue to work as usual, however it is recommended to create the new events with the new base

EPuncker (approved):

data/scripts/lib/event_callbacks.lua
Shawak, line 1: Shouldn't this file also be called 'event.lua' then?

EPuncker (dismissed):
It's cool for me, I would like to wait for @ranisalt opinion on this as well

EPuncker (approved):

EvilHero90 (approved):

EPuncker (approved):

PR #4356, opened 2023-03-27
[Feature] Bestiary

ArturKnopik:
Bestiary system

bestiary_tfs
image
image

### Pull Request Prelude

- [X] I have followed [proper The Forgotten Server code styling][code].
- [X] I have read and understood the [contribution guidelines][cont] before making this PR.
- [X] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Bestiary basic mechanism

PR #4402, opened 2023-04-12
Bestiary in Lua

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes are inspired by the PR linked as Based.
Basically, this PR adds the possibility of partially enabling the bestiary system, more information needs to be added, but this is a good starting point.

The `MonsterType` function now also accepts a raceId to obtain the MonsterType linked to that raceId.
New function `Game.getBestiary()` that returns a table with other tables containing name and list of MonsterTypes, both compatible with the Lua `#` operator.
Compatible with Revscript.

**Based:** #4356

### DIFF
--- a/data/scripts/creaturescripts/bestiary_kills.lua
+++ b/data/scripts/creaturescripts/bestiary_kills.lua
@@ -1,22 +1,21 @@
 local BESTIARY_PARTY_KILL_SHARING = true
 
 local function getKillersForBestiary(monster)
-	local players = monster:getKillers(true)
 	if not BESTIARY_PARTY_KILL_SHARING then
-		return players[1] and {players[1]} or {}
+		return monster:getKillers(true)
 	end
 
 	local killers = {}
-	for _, killer in pairs(players) do
+	for _, killer in pairs(monster:getKillers(true)) do
 		local party = killer:getParty()
 		if party then
 			for _, member in pairs({party:getLeader(), unpack(party:getMembers())}) do
 				if party:isMemberSharingExp(member) then
-					killers[#killers + 1] = member
+					killers[member:getId()] = member
 				end
 			end
 		else
-			killers[#killers + 1] = killer
+			killers[killer:getId()] = killer
 		end
 	end
 	return killers


data/scripts/creaturescripts/bestiary_kills.lua
```diff
@@ -0,0 +1,47 @@
+local PARTY_MEMBERS_ENABLED = false
+
+local function kill(player, monsterName, raceId, unlockLimits)
+    local oldKills = player:getBestiaryKills(raceId)
```
EPuncker, line 4: use tab instead of spaces for indentation

data/scripts/creaturescripts/bestiary_kills.lua
```diff
@@ -0,0 +1,47 @@
+local PARTY_MEMBERS_ENABLED = false
```
omarcopires, line 1: Any reason not to keep it in config.lua?

data/scripts/creaturescripts/bestiary_kills.lua
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    local info = monster:getType():getBestiaryInfo()
+    if info.raceId == 0 then
+        return true
+    end
+
+    local monsterName = monster:getName()
+    local unlockLimits = {info.prowess, info.expertise, info.mastery}
+    if PARTY_MEMBERS_ENABLED then
```
ranisalt, line 31: Note that the system in Tibia is not like that, only party members that actually damaged the creature will get kills counted. If you just stand there watching, it will not count.

To add further complications to the mix, this is only when shared exp is enabled. If it is not enabled, only the highest damaging player gets the kill, just like it would be outside a party.

There's the scenario where both happen (some members of a party damage the creature, some members out of the party damage it).

MillhioreBT: I think it's perfect, it's just what I wanted to read

data/scripts/creaturescripts/bestiary_kills.lua
MillhioreBT: There is no reason, maybe the idea of ​​having the boolean in config.lua is a good idea to maintain consistency with other systems

config.lua.dist
```diff
@@ -161,6 +161,10 @@ removeOnDespawn = true
 walkToSpawnRadius = 15
 monsterOverspawn = false
 
+-- Bestiary
+bestiaryPartyKillSharing = false
+bestiaryKillsPoints = 1
```
EPuncker, line 166:
```suggestion
bestiaryPointsPerKill = 1
```

EPuncker: the party kill sharing I agree, but multiple points per kill is a very popular event, but I'm fine with either outcome

omarcopires:
After killing some Dragon Lords using `bestiaryPointsPerKill = 100` the client is crashing and returning the following console error:
```
Lua Script Error: [Event Interface]
data/events/scripts/player.lua:Player@onNetworkMessage
LuaScriptInterface::getNumber(). Argument 2 has out-of-range value for unsigned char: 512
stack traceback:
        ...os Git\forgottenserver\data\scripts\network\bestiary.lua:156: in function 'handler'
        data/events/scripts/player.lua:358: in function <data/events/scripts/player.lua:351>

Lua Script Error: [Event Interface]
data/events/scripts/player.lua:Player@onNetworkMessage
LuaScriptInterface::getNumber(). Argument 2 has out-of-range value for unsigned short: -10
stack traceback:
        ...os Git\forgottenserver\data\scripts\network\bestiary.lua:157: in function 'handler'
        data/events/scripts/player.lua:358: in function <data/events/scripts/player.lua:351>
```

MillhioreBT:
> 

Oops sorry, I forgot not all items are counted here, I'll fix this in a bit, thanks for reminding me <3

omarcopires:
Okay, no errors were returned in any of the bestiary steps and no crash occurred to me.
On the client side everything works as it should.

image

omarcopires:
All bestiary steps work as I expect;
The description works as expected;
`bestiaryPointsPerKill` works as expected;
`bestiaryPartyKillSharing` works as expected in both scenarios.

On the client side everything is as it should, I believe it can be merged as it is. If there is no change suggested by any staff member.

**Thanks Arthur and Sarah for the content.**

omarcopires (approved):

EPuncker (dismissed):

config.lua.dist
```diff
@@ -161,6 +161,10 @@ removeOnDespawn = true
 walkToSpawnRadius = 15
 monsterOverspawn = false
 
+-- Bestiary
+bestiaryPartyKillSharing = true
+bestiaryPointsPerKill = 1
```
ranisalt, line 166: I don't think we need these options. The code is easily editable if someone wants special behavior, and it just creates configuration creep - more configuration that almost no one will change, and if they need to change, they have to change some of the code anyway.

data/lib/core/creature.lua
```diff
@@ -178,3 +178,29 @@ end
...
+	local inFightTicks = configManager.getNumber(configKeys.PZ_LOCKED)
+	local timeNow = os.mtime()
+	for cid, cb in pairs(self:getDamageMap()) do
+		local creature = Creature(cid)
+		if creature and creature ~= self and (timeNow - cb.ticks) <= inFightTicks then
+			killers[#killers + 1] = creature
+		end
+	end
+	return killers
+end
```
ranisalt, line 193: What if this function and the one below are merged? Return a table of killers sorted by most damage first. It doesn't add to the code complexity though, so if it the code ends up ugly, just dismiss that 😅

data/lib/core/player.lua
```diff
@@ -393,3 +393,50 @@ function Player.isPromoted(self)
...
+function Player.getUnlockedBestiary(self, monsterTypes)
+	local count = 0
+	for _, monsterType in pairs(monsterTypes) do
+		local info = monsterType:getBestiaryInfo()
+		if info.raceId ~= 0 and self:getStorageValue(PlayerStorageKeys.bestiaryKillsBase + info.raceId) >= info.prowess then
+			count = count + 1
+		end
+	end
+	return count
+end
```
ranisalt, line 406: What is this function used for?

data/scripts/creaturescripts/bestiary_kills.lua
```diff
@@ -0,0 +1,44 @@
...
+		local mostDamagePlayer = monster:getMostDamagePlayer()
+		return mostDamagePlayer and {mostDamagePlayer} or {}
+	end
+
+	local killers = {}
+	for _, killer in pairs(monster:getKillers()) do
+		if killer:isPlayer() then
+			local party = killer:getParty()
+			if party then
+				for _, member in pairs({party:getLeader(), unpack(party:getMembers())}) do
```
ranisalt, line 12: Neat! 👍

data/scripts/network/bestiary.lua
```diff
@@ -0,0 +1,193 @@
...
+	if progress > 3 then
+		response:addByte(0)
+		response:addByte(1)
+	end
+
+	response:sendToPlayer(player)
+	response:delete()
+end
+
+handler:register()
```
ranisalt, line 193: I would rather have 1 file = 1 packet type to avoid long files with too much behavior

src/luascript.cpp
```diff
@@ -14923,6 +14988,32 @@ int LuaScriptInterface::luaMonsterTypeChangeTargetSpeed(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaMonsterTypeBestiaryInfo(lua_State* L)
+{
+	// get: monsterType:bestiaryInfo() set: monsterType:bestiaryInfo(info)
```
ranisalt, line 14993: Why not calling it just `bestiary()`?

data/monster/lua/#example.lua
```diff
@@ -32,6 +32,18 @@ monster.flags = {
...
 }
 
+monster.bestiary = {
+	class = "Dragon",
+	raceId = 39,
+	prowess = 50,
+	expertise = 500,
+	mastery = 1000,
+	charmPoints = 25,
+	stars = 3,
```
ranisalt, line 42: In the client, this is called "difficulty" and has the levels "harmless", "trivial", "easy", "medium", "hard" and "challenging" (0 to 5 stars)

image

EPuncker (dismissed):

MillhioreBT:
Points will no longer be counted twice for those who deal damage to monsters and are in the same party
The behavior without sharing in a party was also fixed, which is basically giving the point to everyone who participated in the death of the monsters.

omarcopires (approved):

omarcopires:
The only issue I had found was about the multiplication in party cases, but it has now been fixed.
Everything is still working as it should after the last fixes.

EPuncker (approved):

PR #4412, opened 2023-04-20
Improvements that the Bestiary needs.

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes improve the readability of the code and raceId has been moved to a better place.
now the occurrence must be set with a name and not with a raw number. (`common`, `uncommon` `ect...`)
Thanks to ranisalt and StreamSide for all the information and suggestions.

### DIFF
--- a/data/lib/core/bestiary.lua
+++ b/data/lib/core/bestiary.lua
@@ -1,0 +1,34 @@
+local bestiaryData = {
+	-- dummy tables for very rare and none
+	[-2] = {kills = {0, 0, 0}},
+	[-1] = {kills = {2, 3, 5}},
+
+	-- number of stars (from harmless to challenging)
+	-- rarity 4 triggers rare values
+	[0] = {kills = {5, 10, 25},  charmPoints = 1},
+	[1] = {kills = {10, 100, 250}, charmPoints = 5},
+	[2] = {kills = {25, 250, 500},  charmPoints = 15},
+	[3] = {kills = {50, 500, 1000}, charmPoints = 25},
+	[4] = {kills = {100, 1000, 2500}, charmPoints = 50},
+	[5] = {kills = {200, 2000, 5000}, charmPoints = 100}
+}
+
+function MonsterType:getBestiaryKills()
+	if self:raceId() == 0 then
+		return bestiaryData[-2].kills
+	end
+
+	local info = self:bestiaryInfo()
+	if info.occurrence == 3 then
+		return bestiaryData[-1].kills
+	end
+	return bestiaryData[info.difficulty].kills
+end
+
+function MonsterType:getBestiaryCharmPoints()
+	if self:raceId() == 0 then
+		return 0
+	end
+	local info = self:bestiaryInfo()
+	return bestiaryData[info.difficulty].charmPoints * (info.occurrence == 3 and 5 or 2)
+end
--- a/src/monsters.h
+++ b/src/monsters.h
@@ -12,9 +12,6 @@ class LuaScriptInterface;
 
 const uint32_t MAX_LOOTCHANCE = 100000;
 
-static constexpr int32_t BESTIARY_MAX_DIFFICULTY = 5;
-static constexpr int32_t BESTIARY_MAX_OCCURRENCE = 4;
-
 struct LootBlock
 {
 	uint16_t id;
@@ -98,13 +95,8 @@ struct voiceBlock_t
 struct BestiaryInfo
 {
 	std::string className = "";
-	uint32_t raceId = 0;
-	uint32_t prowess = 0;
-	uint32_t expertise = 0;
-	uint32_t mastery = 0;
-	uint32_t charmPoints = 0;
-	uint32_t difficulty = 0;
-	uint32_t occurrence = 0;
+	uint8_t difficulty = 0;
+	uint8_t occurrence = 0;
 	std::string locations = "";
 };
 
@@ -186,6 +178,7 @@ public:
 
 	std::string name;
 	std::string nameDescription;
+	uint32_t raceId = 0;
 
 	MonsterInfo info;
 	BestiaryInfo bestiaryInfo;
@@ -256,7 +249,6 @@ public:
 
 	MonsterType* getMonsterType(uint32_t raceId);
 	bool addBestiaryMonsterType(const MonsterType* monsterType);
-	bool isValidBestiaryInfo(const BestiaryInfo& bestiaryInfo) const;
 
 	std::unique_ptr<LuaScriptInterface> scriptInterface;
 	std::map<std::string, MonsterType> monsters;
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -515,9 +515,8 @@ function Player.addBestiaryKills(self, raceId)
 
 	local kills = self:getBestiaryKills(raceId)
 	local newKills = kills + 1
-	local bestiaryInfo = monsterType:getBestiaryInfo()
-	for _, totalKills in pairs({bestiaryInfo.prowess, bestiaryInfo.expertise, bestiaryInfo.mastery}) do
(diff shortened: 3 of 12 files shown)


omarcopires:
Everything works as it should, I did the same tests as in the previous pull.

omarcopires (dismissed):

EPuncker (dismissed):

ArturKnopik:
maybe let's leave the hybrid option, if there are no blocks responsible for setting kills, let's set it by default depending on the number of stars, if this data is present, then instead of the default ones, let's use these
this will allow us to generate custom bestiaries records(example: 1000-5000-15000 kills)

MillhioreBT:
> maybe let's leave the hybrid option, if there are no blocks responsible for setting kills, let's set it by default depending on the number of stars, if this data is present, then instead of the default ones, let's use these this will allow us to generate custom bestiaries records(example: 1000-5000-15000 kills)

It's very easy to create a new table based on names or raceId or whatever you want, and hook it up to the getBestiaryKills method for custom behavior.
image

data/lib/core/bestiary.lua
```diff
@@ -0,0 +1,34 @@
...
+
+	-- number of stars (from harmless to challenging)
+	-- rarity 4 triggers rare values
+	[0] = {kills = {5, 10, 25},  charmPoints = 1},
+	[1] = {kills = {10, 100, 250}, charmPoints = 5},
+	[2] = {kills = {25, 250, 500},  charmPoints = 15},
+	[3] = {kills = {50, 500, 1000}, charmPoints = 25},
+	[4] = {kills = {100, 1000, 2500}, charmPoints = 50},
+	[5] = {kills = {200, 2000, 5000}, charmPoints = 100}
+}
```
ranisalt, line 14: What if I want to create a monster with a custom amount of kills per level?

I'm OK with defaulting to Tibia behavior, but not as much with only being able to have Tibia behavior.

MillhioreBT: I don't know if the one [I posted above](https://github.com/otland/forgottenserver/pull/4412#issuecomment-1517881166) is enough.

soul4soul:
I understand the value in having a table with bestiary data because there are only a few combination of values but personally I'd rather it all bestiary data live directly in the creature files. On top of that I have additional bias to have it in the creature files to make it easier for my monster converter to work.

MillhioreBT:
> I understand the value in having a table with bestiary data because there are only a few combination of values but personally I'd rather all bestiary data live directly in the creature files as it is before the PR. On top of that I have additional bias to have it in the creature files to make it easier for my monster converter to work.

I don't know who in their right mind would customize over 800 monster files XD. That's what the table is for, so that only 4 lines can control those 800 files.

But to be honest, I'm not sure if it's worth having these configurations in each file, or if it's better to use Lua for this. There are many ways to get monsters with customized kills.

I hope other people can suggest or give ideas...

ranisalt:
> On top of that I have additional bias to have it in the creature files to make it easier for my monster converter to work.

How does it work with monsters created in Lua?

soul4soul:
> > On top of that I have additional bias to have it in the creature files to make it easier for my monster converter to work.
> 
> How does it work with monsters created in Lua?

I created lua binding that match TFS monsters revsctiptsys https://github.com/soul4soul/ot-monster-converter/pull/84. For all the formats I support the converter is generally only able to look directly at the monster files not any of the supporting files.

> > I understand the value in having a table with bestiary data because there are only a few combination of values but personally I'd rather all bestiary data live directly in the creature files as it is before the PR. On top of that I have additional bias to have it in the creature files to make it easier for my monster converter to work.
> 
> I don't know who in their right mind would customize over 800 monster files XD. That's what the table is for, so that only 4 lines can control those 800 files.
> 
> But to be honest, I'm not sure if it's worth having these configurations in each file, or if it's better to use Lua for this. There are many ways to get monsters with customized kills.
> 
> I hope other people can suggest or give ideas...

True, it will be rare that an OT owner will customize each monster. I suspect people will be less likely to touch the default monsters and instead have custom values for their custom creatures. Still, as of now all data about a creature is defined directly in the creature file I'd prefer to keep it that way.

omarcopires:
I agree with what @MillhioreBT says, it doesn't make sense to manipulate more than 800 files to configure 4 or 5 different behaviors if we can simply customize some desired monster just by adding a few extra lines to the code. Maybe just leaving an example script will be enough.

As for the changes to this pull request, everything works as it should and unless someone has any suggestions or code hints it's ready to be merged.

EPuncker:
yeah please let push that in, there is no need to delay things more

soul4soul:
> I agree with what @MillhioreBT says, it doesn't make sense to manipulate more than 800 files to configure 4 or 5 different behaviors if we can simply customize some desired monster just by adding a few extra lines to the code. Maybe just leaving an example script will be enough.
> 
> As for the changes to this pull request, everything works as it should and unless someone has any suggestions or code hints it's ready to be merged.

Either way you need to touch 800 files to add the difficulty and occurrence. Another PR that was opened already had to work done, making the changes isn't a problem. I won't reiterate my reasons I already mentioned above.

omarcopires:
> > I agree with what @MillhioreBT says, it doesn't make sense to manipulate more than 800 files to configure 4 or 5 different behaviors if we can simply customize some desired monster just by adding a few extra lines to the code. Maybe just leaving an example script will be enough.
> > As for the changes to this pull request, everything works as it should and unless someone has any suggestions or code hints it's ready to be merged.
> 
> Either way you need to touch 800 files to add the difficulty and occurrence. Another PR that was opened already had to work done, making the changes isn't a problem. I won't reiterate my reasons I already mentioned above.

It is due to this kind of selfish behavior that people distance themselves from TFS. These comments only hinder development, adding nothing except for blocking the entire request and causing the developer to abandon it.

It would be easy to resolve if you simply submitted your version of the code and let the team discuss which version should be merged. Instead, you just block something that would be useful for the community by saying, "I won't reiterate my reasons I already mentioned above."

soul4soul:
> > > I agree with what @MillhioreBT says, it doesn't make sense to manipulate more than 800 files to configure 4 or 5 different behaviors if we can simply customize some desired monster just by adding a few extra lines to the code. Maybe just leaving an example script will be enough.
> > > As for the changes to this pull request, everything works as it should and unless someone has any suggestions or code hints it's ready to be merged.
> > 
> > 
> > Either way you need to touch 800 files to add the difficulty and occurrence. Another PR that was opened already had to work done, making the changes isn't a problem. I won't reiterate my reasons I already mentioned above.
> 
> It is due to this kind of selfish behavior that people distance themselves from TFS. These comments only hinder development, adding nothing except for blocking the entire request and causing the developer to abandon it.
> 
> It would be easy to resolve if you simply submitted your version of the code and let the team discuss which version should be merged. Instead, you just block something that would be useful for the community by saying, "I won't reiterate my reasons I already mentioned above."

Having a difference of opinions doesn't make behavior selfish. I'm not part of TFS development team they are free to merge this PR. I do try and contribute to the community and I'm sharing that by going this route it will detract from the features I will be able to support with my tool. Albeit my tool has few users and most are one times users. 

My previous comment was only to rebut your argument that my suggestions requires 800 files to be edited as not being a good one. Both approaches requires 800 files to be manipulate. As for not relisting my reasons, why would I do that? They are one the same page a few comments up, I assume those who needed to read them have done so already.

src/monsters.h
```diff
@@ -97,7 +97,11 @@ struct BestiaryInfo
 	std::string className = "";
 	uint8_t difficulty = 0;
 	uint8_t occurrence = 0;
+	std::tuple<uint32_t, uint32_t, uint32_t> totalKills;
```
ranisalt, line 100: Do you see any advantages on using a tuple vs three separate `uint32_t`s?

MillhioreBT: I don't think it has any advantage other than looking prettier 😊

PR #4421, opened 2023-04-23
feature: in-game highscores

reyaleman:
### Pull Request Prelude

### Changes Proposed
- Added in-game highscores feature...
- Easy to add custom highscores like "Most deaths, online times, storage checks, etc..."

Needs testing, suggestions & code review

**Issues addressed:** 


ranisalt:
It seems that the list of vocations is defined by the server. In this scenario, we do not have to lock to the regular vocations, since we can just use what vocations are set in vocations.xml

omarcopires:
All right on the client side.

image

EPuncker (approved):

omarcopires (dismissed):

data/lib/core/core.lua
```diff
@@ -19,3 +19,4 @@ dofile('data/lib/core/teleport.lua')
 dofile('data/lib/core/tile.lua')
 dofile('data/lib/core/vocation.lua')
 dofile('data/lib/core/quests.lua')
+dofile('data/lib/core/highscores.lua')
```
ranisalt, line 22: This should be kept in alphabetical order.

`quests.lua` is out of place @MillhioreBT please move it to the alphabetical order

EPuncker (approved):

DSpeichert (dismissed):

data/lib/core/player.lua
```diff
@@ -534,3 +534,63 @@ function Player.sendBestiaryMilestoneReached(self, raceId)
...
+
+	for _, vocation in ipairs(vocations) do
+		msg:addU32(vocation:getId())
+		msg:addString(vocation:getName())
+	end
+
+	msg:addU32(params.vocation)
+
+	msg:addByte(#HIGHSCORES_CATEGORIES)
+	for id, category in ipairs(HIGHSCORES_CATEGORIES) do
```
ranisalt, line 565: Why does this one in particular use `ipairs` instead of `pairs`?

data/lib/core/highscores.lua
```diff
@@ -0,0 +1,242 @@
...
+	[HIGHSCORES_CATEGORY_ACHIEVEMENTS] = [[
+		SELECT `id`, `name`, `vocation`, `level`, `storages`.`value` AS `points`
+		FROM `player_storage` AS `storages`
+		LEFT JOIN `players` ON `players`.`id` = `storages`.`player_id`
+		WHERE `storages`.`key` = ]] .. PlayerStorageKeys.achievementsTotal .. [[
+		AND `deletion` = 0 %s
+		ORDER BY `points` DESC, `name` ASC LIMIT ]] .. highscoresMaxResults
+}
+
+do
```
MillhioreBT, line 120: Why is it a global function?
The do-end block is also unnecessary.

data/lib/core/highscores.lua
```diff
@@ -0,0 +1,242 @@
...
+	local extras = {"AND `group_id` NOT IN (" .. table.concat(highscoresExcludedGroups, ",") .. ")"}
+	if self.params.vocation ~= 0xFFFFFFFF then
+		local vocations = relatedIdsVocation[self.params.vocation] or {VOCATION_NONE}
+		table.insert(extras, "AND `vocation` IN(" .. table.concat(vocations, ",") .. ")")
+	end
+
+	local resultId = db.storeQuery(string.format(query, table.concat(extras, " ")))
+	if resultId then
+		repeat
+			table.insert(entries, {
```
MillhioreBT, line 209: Here, I could perfectly use `t[#t + 1]`. It's faster and recommended by Lua.

data/lib/core/highscores.lua
```diff
@@ -0,0 +1,242 @@
+function Vocation.getRelated(self)
```
omarcopires, line 1: Any reason not to move to vacation.lua?

data/global.lua
```diff
@@ -112,6 +112,10 @@ table.contains = function(array, value)
 	return false
 end
 
+table.slice = function(source, from, to)
```
MillhioreBT, line 115: It's better to call it an `array` since it only works with arrays. `source` sounds like it could work with anything.

src/luascript.cpp
```diff
@@ -11957,6 +11976,18 @@ int LuaScriptInterface::luaVocationGetPromotion(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaVocationGetFromVocation(lua_State* L)
+{
+	// vocation:getFromVocation()
```
MillhioreBT, line 11981: `getDemotion` works in the same way. Why have two methods that do the same thing? Maybe it's better to add this other name to the same method?

`core/vocation.lua`
`Vocation.getFromVocation = Vocation.getDemotion`

data/lib/core/player.lua
MillhioreBT: For this particular case it is fine to use ipairs, the difference is in nanoseconds. My apolgies to ranisalt for making him doubt about this. The only advantage of pairs is that when iterating through tables with nil we can iterate over the `nil` values.

data/lib/core/highscores.lua
ranisalt: It needs to work more before it goes on vacation

ranisalt: (it's because highscores get loaded before vocations. we really need to start using `require`)

reyaleman: What if I make a global function `setUpHighscores` to initialize those arrays (`clientIdsVocation` and `relatedIdsVocation`), call the function on the startup global event and move `Vocation.getRelated` and `Vocation.getFromVocation` to `vocation.lua`? @MillhioreBT

PR #4425, opened 2023-04-24
fix ProtocolOutfit const reference

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
We now construct a string instead of holding a constant reference to a string.

**Issues addressed:** #4383

### DIFF
--- a/src/outfit.h
+++ b/src/outfit.h
@@ -26,11 +26,11 @@ struct Outfit
 
 struct ProtocolOutfit
 {
-	ProtocolOutfit(const std::string& name, uint16_t lookType, uint8_t addons) :
-	    name(name), lookType(lookType), addons(addons)
+	ProtocolOutfit(std::string_view name, uint16_t lookType, uint8_t addons) :
+	    name{name}, lookType{lookType}, addons{addons}
 	{}
 
-	const std::string& name;
+	std::string name;
 	uint16_t lookType;
 	uint8_t addons;
 };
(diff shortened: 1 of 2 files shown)


src/outfit.h
```diff
@@ -26,11 +26,11 @@ struct Outfit
...
 struct ProtocolOutfit
 {
-	ProtocolOutfit(const std::string& name, uint16_t lookType, uint8_t addons) :
-	    name(name), lookType(lookType), addons(addons)
+	ProtocolOutfit(std::string_view name, uint16_t lookType, uint8_t addons) :
+	    name{name}, lookType{lookType}, addons{addons}
 	{}
 
-	const std::string& name;
+	std::string name;
```
ranisalt, line 33: This is the actual fix^

src/protocolgame.cpp
```diff
@@ -3236,12 +3236,12 @@ void ProtocolGame::sendPodiumWindow(const Item* item)
 
 	// add GM outfit for staff members
 	std::vector<ProtocolOutfit> protocolOutfits;
+	protocolOutfits.reserve(outfits.size());
```
ranisalt, line 3239: Reserve can be removed, it's not that many outfits and not too often called to warrant

src/outfit.h
MillhioreBT: Thanks for confirming, I didn't know if this would cause any controversy or problem.

ranisalt (approved):

EPuncker (approved):

PR #4435, opened 2023-04-30
Cleanup GitHub workflows using CMake presets

ranisalt:
### Pull Request Prelude

### Changes Proposed

Adds presets for different build types, and stop requiring ninja by allowing CMake to find a suitable build tool. Updates github workflows to build & test on all presets.

### DIFF
--- a/.github/workflows/build-vcpkg.yml
+++ b/.github/workflows/build-vcpkg.yml
@@ -73,7 +73,7 @@ jobs:
         uses: lukka/run-cmake@v10
         with:
           buildPreset: vcpkg
-          buildPresetAdditionalArgs: "['--config ${{ matrix.buildtype }}']"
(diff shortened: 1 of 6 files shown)


.github/workflows/build-vcpkg.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         name: [ubuntu, ubuntu-clang, macos-clang, windows-msvc]
```
ranisalt, line 30: I think we should get rid of `ubuntu-clang` here. Doesn't cover anything (MacOS already builds with clang) and adds 3 more builds that slow the workflow. cc @DSpeichert

DSpeichert: OK

EPuncker (approved):

DSpeichert (approved):

PR #4437, opened 2023-05-01
Update lua-syntax action to latest version

EPuncker:
### Pull Request Prelude

### Changes Proposed
Update lua-syntax action to latest version, Action now runs on node16 as node 12 actions are marked as deprecated:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

also bumped Lua version from 5.3.5 to 5.4 (it will default to latest 5.4.* version)

**Issues addressed:** 
> [luac](https://github.com/otland/forgottenserver/actions/runs/4848872310/jobs/8640352902)
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: leafo/gh-actions-lua@v8.0.0. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

### DIFF
--- a/.github/workflows/lua-syntax.yml
+++ b/.github/workflows/lua-syntax.yml
@@ -16,9 +16,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: leafo/gh-actions-lua@v8.0.0
+      - uses: leafo/gh-actions-lua@v10.0.0
         with:
-           luaVersion: "5.3.5"
+           luaVersion: "5.4"
 
       - name: Test Lua syntax
         run: find data/ -name '*.lua' -print0 | xargs -0 -n1 luac -p
--- a/.github/workflows/lua-syntax.yml
+++ b/.github/workflows/lua-syntax.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: leafo/gh-actions-lua@v10.0.0
+      - uses: leafo/gh-actions-lua@v10
         with:
            luaVersion: "5.4"
 


.github/workflows/lua-syntax.yml
```diff
@@ -16,9 +16,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: leafo/gh-actions-lua@v8.0.0
+      - uses: leafo/gh-actions-lua@v10.0.0
```
ranisalt, line 19:
```suggestion
      - uses: leafo/gh-actions-lua@v10
```

So it picks up latest v10 automatically (e.g. 10.1, 10.0.1)

MillhioreBT (approved):
LGTM

.github/workflows/lua-syntax.yml
```diff
@@ -16,9 +16,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: leafo/gh-actions-lua@v8.0.0
+      - uses: leafo/gh-actions-lua@v10
         with:
-           luaVersion: "5.3.5"
+           luaVersion: "5.4"
```
kornholi, line 21: LuaJIT barely supports any 5.2 and 5.3 features. I wonder if we should test with 5.1 instead.

ranisalt: Or get rid of LuaJIT and go full Lua 5.4. LuaJIT at this point is a single-maintainer project with no releases for the past 6 years. It's not like we're doing scientific computing or other CPU-heavy workloads that would benefit from it, Lua 5.4 is good enough.

PR #4442, opened 2023-05-02
Feat: client exp display

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Now send rate as expected instead of always sending a fixed value to the player.

image


src/player.h
```diff
@@ -1110,6 +1110,12 @@ class Player final : public Creature, public Cylinder
 
 	const std::map<uint8_t, OpenContainer>& getOpenContainers() const { return openContainers; }
 
+	uint16_t getBaseExpGain() const { return baseExpGain; }
+	void setBaseExpGain(uint16_t value)
+	{
+		baseExpGain = std::min<uint16_t>(std::numeric_limits<uint16_t>::max(), value);
```
jo3bingham, line 1116:
```suggestion
		baseExpGain = value;
```
`std::min` is pointless as `value` can't exceed the max value of `uint16_t` because it is a `uint16_t`.

omarcopires: I have almost no knowledge of C, thanks for the help.

jo3bingham (approved):
This LGTM, but you may need approval from someone on the dev team before it can be merged.

EPuncker (dismissed):
if jo3s agree, I agree as well :)

src/player.h
```diff
@@ -1255,6 +1258,7 @@ class Player final : public Creature, public Cylinder
 	uint16_t lastStatsTrainingTime = 0;
 	uint16_t staminaMinutes = 2520;
 	uint16_t maxWriteLen = 0;
+	uint16_t baseExpGain = 100;
```
marmichalski, line 1261: This is misleading as this does not affect exp gain rate at all and is only used for display purposes

marmichalski: This is in no way related to players baseExpGain, if you set it to 200 player won't start getting more XP. All it does is update value in Client (display).

Calling player:setBaseExpGain does not affect the exp gain.

omarcopires: any name suggestions?

omarcopires:
I can't think of better names for the functions, this is the best I could come up with, hope that's good enough.

EPuncker (dismissed):

EPuncker (approved):

DSpeichert (approved):

PR #4449, opened 2023-05-09
Fix missing Luarocks in actions

ranisalt:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:**

Fixes the master branch issue with missing luarocks when installing luacheck. Additionally, since it reports _a shitload of warnings_, I disabled some false-positives, but there is still a lot of work to do.

### DIFF
--- a/.github/workflows/lua-check.yml
+++ b/.github/workflows/lua-check.yml
@@ -5,13 +5,17 @@ on:
     branches:
       - master
       - v*
+
     paths:
-      - 'data/**.lua'
-      - 'config.lua.dist'
+      - .github/workflows/lua-check.yml
+      - config.lua.dist
+      - "data/**.lua"
+
   pull_request:
     paths:
-      - 'data/**.lua'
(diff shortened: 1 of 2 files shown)


ranisalt:
cc @shabbir23ah

EPuncker (approved):

.github/workflows/lua-check.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         run: luarocks install luacheck
 
       - name: Run Luacheck
-        run: luacheck data config.lua.dist
+        run: luacheck data config.lua.dist || true
```
marmichalski, line 44: ... won't that ignore errors also? You could try having some fun with --only or --ignore, or just add all ignored errors to luacheck config

ranisalt: Yes, but currently the data folder triggers 1800+ errors from several different checks. It's too much to ignore 😆 we should just try to fix those before enforcing

PR #4453, opened 2023-05-14
Bye Warning C4266

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
I'm not sure, but some expert can say if it's correct.
I'm using 'using' to specify the use of base methods, so the compiler doesn't issue warnings anymore.

### Notes
We just need to get rid of the C4244 warning.
It's quite bothersome because not all of them show up, so we have to compile multiple times and discover them one by one.

### DIFF
--- a/src/housetile.h
+++ b/src/housetile.h
@@ -13,6 +13,8 @@ class HouseTile final : public DynamicTile
 public:
 	HouseTile(int32_t x, int32_t y, int32_t z, House* house);
 
+	using DynamicTile::internalAddThing;
+
 	// cylinder implementations
 	ReturnValue queryAdd(int32_t index, const Thing& thing, uint32_t count, uint32_t flags,
 	                     Creature* actor = nullptr) const override;
@@ -23,7 +25,6 @@ public:
 	                        Creature* actor = nullptr) const override;
 
 	void addThing(int32_t index, Thing* thing) override;
-	using DynamicTile::internalAddThing;
 	void internalAddThing(uint32_t index, Thing* thing) override;
 
 	House* getHouse() const { return house; }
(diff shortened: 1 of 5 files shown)


EPuncker (approved):

ranisalt:
Not sure this is the correct solution, but at least these methods should be moved to the top (right after constructors)

MillhioreBT:
> Not sure this is the correct solution, but at least these methods should be moved to the top (right after constructors)

The other option is to create the method and call the parent method explicitly.
```cpp
void onWalk(...) override { Creature::onWalk(...); }
```

I don't know which is better, but `using` seems simple to me.
honestly didn't know that `using` could be used for this.

Erza:
> > Not sure this is the correct solution, but at least these methods should be moved to the top (right after constructors)
> 
> The other option is to create the method and call the parent method explicitly.
> 
> ```c++
> void onWalk(...) override { Creature::onWalk(...); }
> ```
> 
> I don't know which is better, but `using` seems simple to me. honestly didn't know that `using` could be used for this.

I ran into this too, that's the solution I went with, but it's super stupid that we even have to do this.
There is also an insane amount of Warning C4242 in Visual Studio, e.g. `warning C4242: 'argument': conversion from 'uint32_t' to 'uint16_t', possible loss of data`

I've spent hours addressing these, but I feel like whenever I fix like 20 of them, 30 new instances show up in the next build. What happened that these suddenly started showing up? The last time the warning level for Visual Studio was adjusted was by Mark in 2013. https://github.com/otland/forgottenserver/blame/master/vc17/theforgottenserver.vcxproj#L145

EPuncker (approved):

Issue #4467, opened 2023-05-28
Deleting items on Quest Doors


Feeltz:
### Before creating an issue, please ensure:
- [x] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Log into regular character (not working for GM)
2. Use any quest door (with red handle)
3. While standing on the quest door tile put any item on the tile
4. Move in direction of closing door

Video: https://youtu.be/SI1V0Up_qZQ

### Expected behaviour
Items are pushed on the tile with character

### Actual behaviour
Character moves on tile, and items are destroyed

If you move to the opposite direction, then items are relocated correctly. If you move to the position where items are pushed then items dissapper.

EPuncker:
@Feeltz can you please test #4468 and let is know if this fixes the issue?

Feeltz:
Yes, it fixes the problem.

PR #4468, opened 2023-05-28
Fix closing door

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
I added the flag `FLAG_IGNOREBLOCKCREATURE`, This is because we are checking if the items can be moved to the next tile. Otherwise, they won't move because queryAdd returns `RETURNVALUE_NOTENOUGHROOM` since the player is already on the tile.

**Issues addressed:** #4467

### DIFF
--- a/data/scripts/movements/closing_door.lua
+++ b/data/scripts/movements/closing_door.lua
@@ -1,5 +1,4 @@
 local closingDoor = MoveEvent()
-closingDoor:type("stepout")
 
 function closingDoor.onStepOut(creature, item, position, fromPosition)
 	local tile = Tile(position)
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

MillhioreBT:
> We should remove obsolete doRelocate(position, newPosition) Also this loop
> 
> ```lua
> local i, tileItem, tileCount = 1, true, tile:getThingCount()
> while tileItem and i < tileCount do
> 	tileItem = tile:getThing(i)
> 	if tileItem and tileItem:getUniqueId() ~= item.uid and tileItem:getType():isMovable() then
> 		tileItem:remove()
> 	else
> 		i = i + 1
> 	end
> end
> ```
> 
> looks weird. Maybe something like this?
> 
> ```lua
> for _, item in pairs(tile:getItems()) do
> 	if item and item:getType():isMovable() then
> 		item:remove()
> 	end
> end
> ```

So officially, creatures are now considered items 🤣

There's no need to change anything in this PR, we're just fixing the bug, that's the point of this PR.

Feeltz:
So what we are removing if script ends if there is any creature?
```lua
if tile:getCreatureCount() > 0 then
	return true
end
	```

MillhioreBT:
> So what we are removing if script ends if there is any creature?
> 
> ```lua
> if tile:getCreatureCount() > 0 then
> 	return true
> end
> ```

I understand that there may be other ways to do it, but as it is, it works well. I don't want to touch those things in this PR. Feel free to open a PR changing this logic.

Feeltz:
It is called Do Not Toutch It Driven Development. Not good. Don't understand why we should make more PRs instead of doing it right at first.

PR #4475, opened 2023-06-01
feat: client bless status

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Add blessings status to the client.

image

**NOTES:**
- The client identifies as green all blessings currently available in the game, but TFS doesn't have all of them, what I did was decrease the value of `>= 7` to `>=5` in `blessingStatus`.
thanks to @ArturKnopik for writing the code https://github.com/otland/forgottenserver/pull/4002/


EPuncker (dismissed):

data/lib/core/player.lua
```diff
@@ -616,3 +616,27 @@ function Player.sendHighscores(self, entries, params)
...
+	end
+	return blessings
+end
+
+function Player.updateClientBlessStatus(self)
+	local msg = NetworkMessage()
+	msg:addByte(0x9C)
+
+	local blessCount = self:getBlessingCount()
+	local blessingStatus = blessCount >= 5 and 3 or (blessCount > 0 and 2 or 1)
```
ranisalt, line 635: Please move this logic to a local function

EPuncker (approved):

ArturKnopik:
My proposal...
blessStatusBits is for setting up on hover bless description on bless status icon
image

```
local blessStatusBits = {
	-- [-1] = 1 -- adventurer blessing, not supported
	-- [0] = 2 -- twist of fate, not supported
	-- IDK if index/values is valid for TFS
	[1] = 4, -- wisdom of solit
	[2] = 8, -- spark of phoenix
	[3] = 16, -- fire of the suns
	[4] = 32, -- spiritual shield
	[5] = 64, -- the embrace of tibia
	-- [6] = 128, -- heart of mountain, not supported
	-- [7] = 256 -- blood of mountain, not supported
}

function Player.getBlessingCount(self)
	local blessings = 0
	local bits = 0
	for i = 1, 6 do
		if self:hasBlessing(i) then
			blessings = blessings + 1
			bits = bits + blessStatusBits[i]
		end
	end
	return blessings, bits
end

local function blessStatus(blessCount)
	if blessCount >= 5 then
		return 3
	elseif blessCount > 0 then
		return 2
	else
		return 1
	end
end

function Player.updateClientBlessStatus(self)
	local msg = NetworkMessage()
	msg:addByte(0x9C)

	local blessCount, bits = self:getBlessingCount()
	local blessingStatus = blessStatus(blessCount)
	msg:addU16(bits)
	msg:addByte(blessingStatus)

	msg:sendToPlayer(self)
	msg:delete()
	return true
end
```

ArturKnopik:
gold border is enabled only if adventurer blessing bit is present

data/lib/core/player.lua
```diff
@@ -616,3 +616,51 @@ function Player.sendHighscores(self, entries, params)
...
 end
+
+local blessStatusBits = {
+	-- [-1] = 1 -- adventurer blessing, not supported
+	-- [0] = 2 -- twist of fate, not supported
+	[1] = 4, -- wisdom of solit
+	[2] = 8, -- spark of phoenix
+	[3] = 16, -- fire of the suns
+	[4] = 32, -- spiritual shield
+	[5] = 64, -- the embrace of tibia
```
EPuncker, line 627:
```suggestion
	[5] = 64, -- the embrace of the world 
```

data/lib/core/player.lua
```diff
@@ -616,3 +616,51 @@ function Player.sendHighscores(self, entries, params)
 	msg:delete()
 	return true
 end
+
+local blessStatusBits = {
+	-- [-1] = 1 -- adventurer blessing, not supported
+	-- [0] = 2 -- twist of fate, not supported
+	[1] = 4, -- wisdom of solit
```
EPuncker, line 623:
```suggestion
	[1] = 4, -- wisdom of solitude
```

EPuncker (dismissed):

EPuncker (dismissed):

data/lib/core/constants.lua
```diff
@@ -30,3 +32,13 @@ SCREENSHOT_TYPE_SKILLUP = 12
 
 SCREENSHOT_TYPE_FIRST = SCREENSHOT_TYPE_ACHIEVEMENT
 SCREENSHOT_TYPE_LAST = SCREENSHOT_TYPE_SKILLUP
+
+BLESS_TYPE_ADVENTURER = -1 -- not supported
```
Zbizu, line 36: you have a negative value there

overall, I believe that this should be an enum in `const.cpp` pushed to Lua through `registerEnum` in `luascript.cpp`

Zbizu (changes requested):

data/lib/core/constants.lua
omarcopires: The numbering is actually incorrect, it should be 1 and not -1. However, I see no reason to move to the engine as it is not being used anywhere.

data/lib/core/player.lua
```diff
@@ -674,3 +664,47 @@ function Player.getTotalDefense(self)
...
+	[2] = BLESS_TYPE_SPARK_OF_PHOENIX,
+	[3] = BLESS_TYPE_FIRE_OF_THE_SUNS,
+	[4] = BLESS_TYPE_SPIRITUAL_SHIELD,
+	[5] = BLESS_TYPE_THE_EMBRACE_OF_THE_WORLD
+}
+
+function Player.getBlessingCount(self)
+	local blessings = 0
+	local bits = 0
+	for i = 1, SERVER_BLESSINGS_COUNT, 1 do
```
MillhioreBT, line 679: image
this 1 is unnecessary, this third parameter is only used when we want to use a number other than 1, since it is 1 by default

```lua
for i = 1, SERVER_BLESSINGS_COUNT, 1 do
-- The same:
for i = 1, SERVER_BLESSINGS_COUNT do
```

MillhioreBT (changes requested):

ranisalt:
Let's move blessings to storages instead and add support for the two enhanced ones. No need to be an engine feature

omarcopires:
> Let's move blessings to storages instead and add support for the two enhanced ones. No need to be an engine feature

this request is just to send the bless status to the client, we can add the new ones and migrate to storages in a new pull request

ArturKnopik:
> Let's move blessings to storages instead and add support for the two enhanced ones. No need to be an engine feature
it seems to me that 90% of feature can be written on storages >.<

data/lib/core/constants.lua
```diff
@@ -30,3 +32,13 @@ SCREENSHOT_TYPE_SKILLUP = 12
...
+
+BLESS_TYPE_ADVENTURER = 1 -- not supported
+BLESS_TYPE_TWIST_OF_FATE = 2 -- not supported
+BLESS_TYPE_WISDOM_OF_SOLITUDE = 4
+BLESS_TYPE_SPARK_OF_PHOENIX = 8
+BLESS_TYPE_FIRE_OF_THE_SUNS = 16
+BLESS_TYPE_SPIRITUAL_SHIELD = 32
+BLESS_TYPE_THE_EMBRACE_OF_THE_WORLD = 64
+BLESS_TYPE_HEART_OF_MOUNTAIN = 128 -- not supported
+BLESS_TYPE_BLOOD_OF_MOUNTAIN = 256 -- not supported
```
ranisalt, line 44: Why are these not supported?

omarcopires: not implemented in tfs, for they are the new blessings.

PR #4479, opened 2023-06-02
blessings updated

reyaleman:
### Pull Request Prelude

### Changes Proposed
Modified how blessings are handled in the server.

1. moved blessings to storages and removed `blessings` column from `players` table.
2. added missing blessings (adventurer's, twist of fate and enhanced blessings)
4. added blessings status (little green icon next to equipment)
5. added blessings window
6. added blessings history
7. blessings can have "charges" (1 charge is removed when dying)

Ps. I saw another PR was opened about the very same issue, so you can ignore this one if you guys want to.

**Issues addressed:** 


data/migrations/35.lua
```diff
@@ -1,3 +1,42 @@
...
+				end
+			end
+		until not result.next(resultId)
+		result.free(resultId)
+
+		local stmtLen = string.len(stmt)
+		if stmtLen > 66 then
+			stmt = string.sub(stmt, 1, stmtLen - 1)
+			db.query(stmt)
+		end
```
DSpeichert, line 26: Something looks wrong here - you're building one huge insert and then... trimming it to 66 chars?

reyaleman: nope, only removing the last comma so the insert query doesn't fails

data/lib/core/player.lua
```diff
@@ -674,3 +664,72 @@ function Player.getTotalDefense(self)
...
+end
+
+function Player.addBlessingsHistory(self, event, type)
+	return db.query("INSERT INTO `blessings_history` (`player_id`, `type`, `event`, `created_at`) VALUES (" .. self:getGuid() .. ", " .. type .. ", " .. db.escapeString(event) .. ", " .. os.time() .. ")")
+end
+
+function Player.getBlessingsHistory(self)
+	local history = {}
+	local resultId = db.storeQuery("SELECT `type`, `event`, `created_at` FROM `blessings_history` WHERE `player_id` = " .. self:getGuid() .. " ORDER BY `created_at` DESC")
+	if resultId then
```
ranisalt, line 698: It would look cleaner if you invert this condition:

```lua
local resultId = ...
if not resultId then
  return {}
end

local history = {}
repeat
  ...
until ...
return history
```

To avoid an indent level

data/scripts/actions/others/blessing_charms.lua
```diff
@@ -1,9 +1,11 @@
 local items = {
-	[11260] = {text = "The Spiritual Shielding protects you.", id = 1, effect = CONST_ME_LOSEENERGY},
-	[11259] = {text = "The Embrace of the World surrounds you.", id = 2, effect = CONST_ME_MAGIC_BLUE},
-	[11261] = {text = "The Fire of the Suns engulfs you.", id = 3, effect = CONST_ME_MAGIC_RED},
-	[11262] = {text = "The Wisdom of Solitude inspires you.", id = 4, effect = CONST_ME_MAGIC_GREEN},
-	[11258] = {text = "The Spark of the Phoenix emblazes you.", id = 5, effect = CONST_ME_FIREATTACK}
+	[11260] = {name = "Spiritual Shielding", text = "The Spiritual Shielding protects you.", id = BLESSING_SPIRITUAL_SHIELDING, effect = CONST_ME_LOSEENERGY},
+	[11259] = {name = "Embrace of the World", text = "The Embrace of the World surrounds you.", id = BLESSING_EMBRACE_OF_THE_WORLD, effect = CONST_ME_MAGIC_BLUE},
```
ranisalt, line 3: I see what you did there 😂

data/migrations/35.lua
reyaleman: 66 it's the `stmt` length without any **inserted** values so if the `stmt` length is 66 we won't execute the insert query nor trim the string, but if it's greater than 66 it means that we have `rows` being inserted so we trim the `stmt` string and  remove the last comma

```sql
INSERT INTO `player_storage` (`player_id`, `key`, `value`) VALUES -- (66 length)
(1, 10003003, 1),
(1, 10003004, 1),
```
we'd have to remove the last comma so the query can be executed.
```sql
INSERT INTO `player_storage` (`player_id`, `key`, `value`) VALUES -- (66 length)
(1, 10003003, 1),
(1, 10003004, 1)
```

we could make an array for rows being inserted and concatenate them

```lua
local inserts = {}

inserts[#inserts + 1] = "..."

...

local stmt = "INSERT INTO ... VALUES "
if #inserts > 0 then
	stmt = stmt .. table.concat(inserts, ",")
	db.query(stmt)
end
```

data/lib/core/player.lua
reyaleman: yep, looks way cleaner

thanks 😁

data/migrations/35.lua
ranisalt: That would be good. Or even better, read all player ids and blessings first into a table, then you can skip building the query if there are no results i.e.

```lua
		rows = {}
		repeat
			local playerId = result.getNumber(resultId, "id")
			local blessings = result.getNumber(resultId, "blessings")
			for id = 0, 4 do
				if bit.band(blessings, 2 ^ id) ~= 0 then
					rows[#rows + 1] = {playerId, id}
				end
			end
		until not result.next(resultId)
```

or whichever structure you prefer :)

data/migrations/35.lua
```diff
@@ -7,22 +7,21 @@ function onUpdateDatabase()
...
 		result.free(resultId)
-
-		local stmtLen = string.len(stmt)
-		if stmtLen > 66 then
-			stmt = string.sub(stmt, 1, stmtLen - 1)
-			db.query(stmt)
+		
+		local stmt = "INSERT INTO `player_storage` (`player_id`, `key`, `value`) VALUES "
+		if #rows > 0 then
+			db.query(stmt .. table.concat(rows, ","))
```
ranisalt, line 24: Excellent!

data/scripts/network/blessings_window.lua
```diff
@@ -44,15 +44,17 @@ local function sendBlessingsWindow(player)
 	local equipmentLoss = {100, 70, 45, 25, 10, 0, 0, 0}
 	local equipmentPercent = equipmentLoss[#blessings + 1]
 
+	local playerSkull = player:getSkull()
+	playerSkull = (playerSkull == SKULL_RED or playerSkull == SKULL_BLACK)
```
ranisalt, line 48: Please don't reuse variables for different meanings. You can use `if playerSkull >= SKULL_RED` below

src/player.cpp
```diff
@@ -4785,3 +4794,45 @@ void Player::updateRegeneration()
...
+	int32_t oldValue;
+	getStorageValue(key, oldValue);
+	if (oldValue == -1) { // storage doesn't exists
+		oldValue = 0;
+	}
+
+	addStorageValue(key, oldValue + count);
+}
+
+void Player::removeBlessing(uint8_t blessingId, int32_t count /* = 1*/) { addBlessing(blessingId, count * -1); }
```
ranisalt, line 4811:
```suggestion
void Player::removeBlessing(uint8_t blessingId, int32_t count /* = 1*/) { addBlessing(blessingId, -count); }
```

src/player.cpp
```diff
@@ -4785,3 +4794,45 @@ void Player::updateRegeneration()
...
+
+uint8_t Player::getBlessing(uint8_t blessingId) const
+{
+	int32_t value;
+	getStorageValue(PSTRG_BLESSINGS_RANGE_START + blessingId, value);
+	if (value < 0) {
+		return 0;
+	}
+
+	return value;
```
ranisalt, line 4821: `return std::max(value, 0)` and get rid of the branch

src/luascript.cpp
```diff
@@ -10374,56 +10386,56 @@ int LuaScriptInterface::luaPlayerSetPremiumEndsAt(lua_State* L)
 	return 1;
 }
 
-int LuaScriptInterface::luaPlayerHasBlessing(lua_State* L)
+int LuaScriptInterface::luaPlayerAddBlessing(lua_State* L)
```
ranisalt, line 10389: This can be written in pure Lua now, manipulating the storages.

reyaleman: okay... should I update the `int LuaScriptInterface::luaPlayerSetStorageValue(lua_State* L)` method to ignore blessings range?

```cpp
if (IS_IN_KEYRANGE(key, RESERVED_RANGE)) {
	reportErrorFunc(L, fmt::format("Accessing reserved range: {:d}", key));
	pushBoolean(L, false);
	return 1;
}
```

ranisalt: No, if you do that it will be impossible to set them anyway :P

ranisalt: Ah sorry, I thought you were asking if the blessings storage range should be added to the protected list, not the other way around. Yes, it should be removed from the reserved range, since there is a key to access them that is very clear (`PlayerStorageKeys.blessingsRangeStart`).

PR #4481, opened 2023-06-07
getOutfitByLookType optimization

lyuz1n:
### Pull Request Prelude

### Changes Proposed

I noticed this method in perf top of a server with hundreds players online, so these changes should improve performance.

**Issues addressed:** 


danyelvarejao (changes requested):

src/outfit.cpp
```diff
@@ -40,31 +40,28 @@ bool Outfits::loadFromXml()
...
-		    outfitNode.attribute("name").as_string(), pugi::cast<uint16_t>(lookTypeAttribute.value()),
-		    outfitNode.attribute("premium").as_bool(), outfitNode.attribute("unlocked").as_bool(true));
+		std::string name = outfitNode.attribute("name").as_string();
+		uint16_t lookType = pugi::cast<uint16_t>(lookTypeAttribute.value());
+		bool premium = outfitNode.attribute("premium").as_bool();
+		bool unlocked = outfitNode.attribute("unlocked").as_bool(true);
+
+		Outfit outfit = Outfit(name, lookType, premium, unlocked);
+		outfits[type].emplace_back(outfit);
+		mappedOutfits[lookType] = &outfit;
```
ranisalt, line 50: This will store pointers to the stack, which will become invalid as soon as the block ends.

MillhioreBT (changes requested):

src/outfit.h
```diff
@@ -52,6 +52,7 @@ class Outfits
 
 private:
 	std::vector<Outfit> outfits[PLAYERSEX_LAST + 1];
+	std::unordered_map<uint16_t, Outfit*> mappedOutfits;
```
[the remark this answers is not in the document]
lyuz1n: thanks a lot!

MillhioreBT:
I think all that remains is to apply the Cclang format

danyelvarejao (approved):

omarcopires:
Just change the spacing for the code below and it will be enough to clang.
```cpp
const uint16_t lookType = pugi::cast<uint16_t>(lookTypeAttribute.value());
		outfits[type].emplace_back(outfitNode.attribute("name").as_string(), lookType,
		                           outfitNode.attribute("premium").as_bool(),
		                           outfitNode.attribute("unlocked").as_bool(true));
```

 ̶I̶ ̶t̶h̶o̶u̶g̶h̶t̶ ̶t̶h̶a̶t̶ ̶g̶i̶t̶h̶u̶b̶ ̶a̶u̶t̶o̶m̶a̶t̶i̶c̶a̶l̶l̶y̶ ̶a̶p̶p̶l̶i̶e̶s̶ ̶t̶h̶e̶ ̶f̶o̶r̶m̶a̶t̶t̶i̶n̶g̶ ̶t̶o̶ ̶t̶h̶e̶ ̶p̶u̶l̶l̶ ̶r̶e̶q̶u̶e̶s̶t̶.̶

MillhioreBT (dismissed):
LGTM

EPuncker (approved):

ranisalt:
Looking at it deeper, it seems that we do not need the outfits vector, we can just have a map of looktype to outfit and be done. The only functions that need _all outfits by sex_ are related to the outfit window in the client, in this case you can just get all outfits and filter by sex since it's not performance critical, it would be better to avoid the "duplication".

So I'd say get rid of the vector, and move the outfits to a (unordered_)map instead.

lyuz1n:
> Looking at it deeper, it seems that we do not need the outfits vector, we can just have a map of looktype to outfit and be done. The only functions that need _all outfits by sex_ are related to the outfit window in the client, in this case you can just get all outfits and filter by sex since it's not performance critical, it would be better to avoid the "duplication".
> 
> So I'd say get rid of the vector, and move the outfits to a (unordered_)map instead.

good idea! thanks!

src/outfit.h
```diff
@@ -46,12 +46,11 @@ class Outfits
...
 	bool loadFromXml();
 
-	const Outfit* getOutfitByLookType(PlayerSex_t sex, uint16_t lookType) const;
 	const Outfit* getOutfitByLookType(uint16_t lookType) const;
-	const std::vector<Outfit>& getOutfits(PlayerSex_t sex) const { return outfits[sex]; }
+	std::vector<Outfit> getOutfits(PlayerSex_t sex) const;
 
 private:
-	std::vector<Outfit> outfits[PLAYERSEX_LAST + 1];
+	std::map<uint16_t, std::pair<PlayerSex_t, Outfit>> outfits;
```
MillhioreBT, line 53: better to use an unordered map and not use a pair as value
```cpp
std::unordered_map<uint16_t, Outfit> outfits;
```

src/outfit.cpp
```diff
@@ -40,31 +40,34 @@ bool Outfits::loadFromXml()
...
 }
 
-const Outfit* Outfits::getOutfitByLookType(PlayerSex_t sex, uint16_t lookType) const
+const Outfit* Outfits::getOutfitByLookType(uint16_t lookType) const
 {
-	for (const Outfit& outfit : outfits[sex]) {
-		if (outfit.lookType == lookType) {
-			return &outfit;
-		}
+	const auto& it = outfits.find(lookType);
```
MillhioreBT, line 55: It is not necessary to use const& for iterators returned by standard methods.
```cpp
const Outfit* Outfits::getOutfitByLookType(uint16_t lookType) const
{
	auto it = outfits.find(lookType);
	if (it != outfits.end()) {
		return &it->second;
	}
	return nullptr;
}
```

src/outfit.cpp
```diff
@@ -40,31 +40,34 @@ bool Outfits::loadFromXml()
 			continue;
 		}
 
-		outfits[type].emplace_back(
-		    outfitNode.attribute("name").as_string(), pugi::cast<uint16_t>(lookTypeAttribute.value()),
-		    outfitNode.attribute("premium").as_bool(), outfitNode.attribute("unlocked").as_bool(true));
+		const uint16_t lookType = pugi::cast<uint16_t>(lookTypeAttribute.value());
```
MillhioreBT, line 43: Since we no longer have two maps it is understandable to add a new value for each outfit to store the gender of said outfit.

<details>

<summary>Outfit</summary>

```hpp
struct Outfit
{
	Outfit(std::string_view name, uint16_t lookType, PlayerSex_t sex, bool premium, bool unlocked) :
	    name{name}, lookType{lookType}, sex{sex}, premium{premium}, unlocked{unlocked}
	{}

	bool operator==(const Outfit& otherOutfit) const
	{
		return name == otherOutfit.name && lookType == otherOutfit.lookType && sex == otherOutfit.sex &&
		       premium == otherOutfit.premium && unlocked == otherOutfit.unlocked;
	}

	std::string name;
	uint16_t lookType;
	PlayerSex_t sex;
	bool premium;
	bool unlocked;
};
```

</details>

And to emplace the values:
```cpp
const auto lookType = pugi::cast<uint16_t>(lookTypeAttribute.value());
outfits.emplace(std::piecewise_construct, std::forward_as_tuple(lookType),
		                std::forward_as_tuple(outfitNode.attribute("name").as_string(), lookType,
		                                      static_cast<PlayerSex_t>(type), outfitNode.attribute("premium").as_bool(),
		                                      outfitNode.attribute("unlocked").as_bool(true)));
```

Remember use cclang if u copy this code.

MillhioreBT (changes requested):
There I left several suggestions for changes, also remember that you must add the new sex field to the outfit in luascript.cpp

PR #4482, opened 2023-06-08
Fix double initialization of Database class

ArturKnopik:
Fix double initialization of `Database` class, `Database`  now become true singleton 
### Pull Request Prelude

### Changes Proposed

**Issues addressed:**

### DIFF
--- a/src/database.h
+++ b/src/database.h
@@ -12,7 +12,6 @@ using DBResult_ptr = std::shared_ptr<DBResult>;
 class Database
 {
 public:
-	Database() = default;
 	~Database();
 
 	// non-copyable
@@ -96,6 +95,7 @@ public:
 	uint64_t getMaxPacketSize() const { return maxPacketSize; }
 
 private:
+	Database() = default;
 	/**
 	 * Transaction related methods.
 	 *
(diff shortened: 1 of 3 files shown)


EPuncker (approved):

nekiro:
DatabaseTasks operates in its own thread, currently with this change you access the singleton from two different threads simultaneously. This code now is probably not thread safe, thus broken.

ArturKnopik:
Good point, need to be checked

kygov:
So database tasks are no longer asynchronously?

nekiro:
> So database tasks are no longer asynchronously?
> 
> https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L4184 https://github.com/otland/forgottenserver/blob/master/src/luascript.cpp#L4224

They are, using the same instance doesn't make it synchronous, but in special cases where simultaneous access could occur it will not cause undefined behaviour. If you meant it will not execute at the same time, it never did, mysql has to queue these actions anyway for obvious reasons

kygov:
AFAIK previous it had its own connection to mysql and you could execute two queries in same time if they were not to same table (then mysql would lock it).
I used to move queries to database tasks many times to offload main thread.
Servers sometimes have heavy sql queries that take up to 100ms, in this case I think they will just block the main thread.

Someone should check it.

PR #4483, opened 2023-06-08
Fix Memory Leak from Npc events

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes remove the global lua interface for all npc, and now each npc has its own lua interface, in turn the handler and interface become smart pointers so that their removal is handled automatically

asdasdsa
Here is a visual example of a graph showing how memory usage changes when creating and deleting 5000 Npcs, and then the garbage collector cleaning up everything.

**Issues addressed:** #4288

### DIFF
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -54,11 +54,6 @@ bool Npc::load()
 
 	reset();
 
-	if (!scriptInterface) {
-		scriptInterface.reset(new NpcScriptInterface());
-		scriptInterface->loadNpcLib("data/npc/lib/npc.lua");
-	}
-
 	loaded = loadFromXml();
 	return loaded;
 }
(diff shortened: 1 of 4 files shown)


nekiro:
I would advocate for making NpcEventsHandler per npc and not global for all npcs, for some odd reason its global for npcs and the memory usage changes are marginal

nekiro (dismissed):

src/npc.h
```diff
@@ -72,7 +72,7 @@ class NpcEventsHandler
 
 private:
 	Npc* npc;
-	NpcScriptInterface* scriptInterface;
+	std::shared_ptr<NpcScriptInterface> scriptInterface;
```
ranisalt, line 75: How about making it a `unique_ptr` to avoid the extra cost? Since the Npc class just needs to access it, you can store a reference to the unique_ptr there.

src/npc.cpp
```diff
@@ -60,7 +55,7 @@ bool Npc::load()
 	reset();
 
 	if (!scriptInterface) {
-		scriptInterface = new NpcScriptInterface();
+		scriptInterface.reset(new NpcScriptInterface());
```
ranisalt, line 58: `scriptInterface = std::make_unique/shared<NpcScriptInterface>()`

nekiro: is there any technical difference or just preference?

ranisalt: Mostly semantics, as `unique_ptr<T>::operator=(unique_ptr<T>&& rhs)` is equivalent to `lhs.reset(rhs.release())`. However, it is good practice that _every_ `new` has a matching `delete`, but this one will not have.

However, in the _extremely unlikely_ scenario where there's an exception after `new` but before `reset`, then you get another memory leak.

src/npc.cpp
```diff
@@ -208,12 +197,11 @@ bool Npc::loadFromXml()
...
 	if (scriptFile) {
-		npcEventHandler = new NpcEventsHandler(scriptFile.as_string(), this);
-		if (!npcEventHandler->isLoaded()) {
-			delete npcEventHandler;
-			npcEventHandler = nullptr;
+		auto handler = std::make_unique<NpcEventsHandler>(scriptFile.as_string(), this);
+		if (!handler->isLoaded()) {
 			return false;
 		}
+		npcEventHandler.reset(handler.release());
```
ranisalt, line 204:
```suggestion
		npcEventHandler = std::move(handler.release());
```

ranisalt (dismissed):
Code LGTM

ranisalt (approved):

PR #4495, opened 2023-06-27
Add storage values to creatures

ranisalt:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:**
Moves storage values up from players to creatures, meaning that monsters and NPCs also get storage values. 

Only player storage is saved to the database, since creatures are not stable between server restarts. It could be extended to save NPC storage values, though.

Implementation detail: I opted for an ordered map instead of unordered because `std::unordered_map` uses a lot of memory when empty, as it is optimized for high usage.

### DIFF
--- a/data/events/events.xml
+++ b/data/events/events.xml
@@ -5,6 +5,7 @@
 	<event class="Creature" method="onAreaCombat" enabled="0" />
 	<event class="Creature" method="onTargetCombat" enabled="0" />
 	<event class="Creature" method="onHear" enabled="0" />
+	<event class="Creature" method="onUpdateStorage" enabled="1" />
 
 	<!-- Party methods -->
 	<event class="Party" method="onJoin" enabled="0" />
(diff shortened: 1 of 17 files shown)


EPuncker (approved):

MillhioreBT:
I think the change is beautiful and necessary, however it must be taken into account that all otland scripts will break.
because now -1 is replaced by nil

omarcopires:
> I think the change is beautiful and necessary, however it must be taken into account that all otland scripts will break.
> 
> because now -1 is replaced by nil

maybe it's something very stupid, but just adding to compat wouldn't be enough?

MillhioreBT:
> > I think the change is beautiful and necessary, however it must be taken into account that all otland scripts will break.
> > because now -1 is replaced by nil
> 
> maybe it's something very stupid, but just adding to compat wouldn't be enough?

sure, this is probably the way to go.

EPuncker (changes requested):
why nil instead of -1? 🤔

ranisalt:
> why nil instead of -1? 🤔

Trying to sneakily fix this prehistoric fuck up

MillhioreBT:
> > why nil instead of -1? 🤔
> 
> Trying to sneakily fix this prehistoric fuck up

Add compatibility functions in compat.lua and you are good to ready 😊

EPuncker (approved):

ranisalt:
> Add compatibility functions in compat.lua and you are good to ready blush

I think the way I did it covers everything

EPuncker (approved):

EPuncker (approved):
neat

PR #4505, opened 2023-07-19
Separate onGainExperience logic into its own file

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes add a new eventcallback file for the default onGainExperience event, where the logic is worked on separately. Additionally, an event is added for the message, which is also separated and uses the positive infinite activation index to ensure that this event is activated at the end of all events. This guarantees that it is the last event and therefore any changes to the experience made in the previous events are taken into account.


data/scripts/eventcallbacks/player/default_onGainExperience.lua
```diff
@@ -0,0 +1,95 @@
...
+
+default:register()
+
+-- For this event, we use the trigger index math.huge so that this event is called last, thus ensuring that the
+-- experience message is sent to the client with the correct value.
+
+local message = EventCallback
+
+function message.onGainExperience(player, source, exp, rawExp, sendText)
+	if sendText then
```
omarcopires, line 80:
```lua
if sendText and exp > 0 then
```

MillhioreBT: This is why I did the PR and I forgot xd
thanks you <3

omarcopires (approved):

EPuncker (approved):

PR #3452, opened 2021-05-10
Fix depot crash on ubuntu

Znote:
Fix https://github.com/otland/forgottenserver/issues/3450

### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

**Issues addressed:** #3450

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4510, opened 2023-07-23
fix: depotchest memory leak

lyuz1n:
### Pull Request Prelude

### Changes Proposed

items on depot cannot destroy after player logout.
after some relogins with around 100k items on depot...

BEFORE FIX:
Captura de tela 2023-07-23 164438

AFTER FIX:
Captura de tela 2023-07-23 164420

### DIFF
--- a/src/player.h
+++ b/src/player.h
@@ -21,6 +21,8 @@ class Npc;
 class Party;
 class SchedulerTask;
 
+using DepotChest_ptr = std::shared_ptr<DepotChest>;
+
 enum skillsid_t
 {
 	SKILLVALUE_LEVEL = 0,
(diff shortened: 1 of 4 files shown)


nekiro:
https://github.com/otland/forgottenserver/pull/3423
https://github.com/otland/forgottenserver/pull/3452

src/player.h
```diff
@@ -21,6 +21,8 @@ class Npc;
 class Party;
 class SchedulerTask;
 
+using DepotChest_ptr = std::shared_ptr<DepotChest>;
```
ranisalt, line 24: Just use `std::shared_ptr<DepotChest>` directly

moviebr (approved):
Nice :)

nekiro (changes requested):
This will crash, because of double pointer deletion, one happens from container destructor, because depot is a regular item inside locker (you have to remove depot from locker on player destructor to avoid that) and the second will trigger from shared_ptr trying to delete assigned object

ranisalt:
@nekiro is right, if you're going to change to a shared ptr, you must change it _everywhere_ or you will get nasty double frees

ramon-bernardo:
I ran some tests, and it seems that the code below solves this issue 👇🏻 
```c++
Player::~Player()
{
	for (auto [_, depotChest] : depotChests) {
		delete depotChest;
	}
}

nekiro:
superseded by #4879

PR #4514, opened 2023-08-05
actions: build .deb & fixes

DSpeichert:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 
* Ubuntu builds were not properly uploading binary to artifacts
* a .deb package is now uploaded too
* test workflow does not run unnecessarily

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -31,6 +31,7 @@ jobs:
       security-events: write
 
     strategy:
+      fail-fast: false
       matrix:
         buildtype: [Debug, Release]
         luajit: [on, off]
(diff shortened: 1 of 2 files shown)


ranisalt:
I doubt enough people use raw TFS without changes to warrant maintaining this.

DSpeichert:
> I doubt enough people use raw TFS without changes to warrant maintaining this.

It's mostly easy of use for those who are sticking to a tagged branch, especially if I figure out a way to maintain an APT repo. Doesn't cost us anything to keep it here.

ranisalt:
People can always use Docker for that, which is way easier, and works in every distro.

> Doesn't cost us anything to keep it here.

I can see it does just by looking at the package list and seeing different packages from that Ubuntu and vcpkg install. It's a third list of dependencies (not counting the ones on the wiki) that we need to keep updated and that we will not keep updated

PR #4518, opened 2023-08-21
Load MySQL variables from environment

ranisalt:
### Pull Request Prelude

### Changes Proposed

Loads MySQL configuration from the environment first, keeping compatibility with Lua config.

This allows better compatibility with database containers, such as [Docker's mariadb](https://hub.docker.com/_/mariadb) and a single source of configuration can be used for the database. It's also easier for other software, such as AACs, that can load the credentials from the environment rather than poorly parsing the Lua config.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -25,6 +25,14 @@ extern Game g_game;
 
 namespace {
 
+const char* getEnv(const char* envVar, const char* defaultValue)
+{
+	if (auto value = std::getenv(envVar)) {
+		return value;
+	}
+	return defaultValue;
+}
+
 std::string getGlobalString(lua_State* L, const char* identifier, const char* defaultValue)
 {
 	lua_getglobal(L, identifier);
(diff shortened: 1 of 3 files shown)


src/configmanager.cpp
```diff
@@ -170,10 +178,10 @@ bool ConfigManager::load()
...
-		string[MYSQL_USER] = getGlobalString(L, "mysqlUser", "forgottenserver");
-		string[MYSQL_PASS] = getGlobalString(L, "mysqlPass", "");
-		string[MYSQL_DB] = getGlobalString(L, "mysqlDatabase", "forgottenserver");
+		string[MYSQL_HOST] = getGlobalString(L, "mysqlHost", getEnv("MYSQL_HOST", "127.0.0.1"));
+		string[MYSQL_USER] = getGlobalString(L, "mysqlUser", getEnv("MYSQL_USER", "forgottenserver"));
+		string[MYSQL_PASS] = getGlobalString(L, "mysqlPass", getEnv("MYSQL_PASSWORD", ""));
+		string[MYSQL_DB] = getGlobalString(L, "mysqlDatabase", getEnv("MYSQL_DATABASE", "forgottenserver"));
 		string[MYSQL_SOCK] = getGlobalString(L, "mysqlSock", "");
 
 		integer[SQL_PORT] = getGlobalNumber(L, "mysqlPort", 3306);
```
DSpeichert, line 187: Why not extend environment support to socket and port?

ranisalt: Unintentional, will fix

DSpeichert (approved):

moviebr (approved):
Nice

EPuncker (approved):

slawkens:
Would be nice to have an .env.example file, so there is something just to fill out and the user doesn't need to create it himself.

gesior:
> This allows better compatibility with database containers, such as [Docker's mariadb](https://hub.docker.com/_/mariadb) and a single source of configuration can be used for the database

In most (all?) other config/env loaders in PHP/Python/JS/TS it works different:
Env variables overwrite config file variables.
In current implementation, it loads env variables only, if there is no SQL config in `config.lua` - which results in removing SQL config from `config.lua`!

I think it should work different: it should load `config.lua` first, then check, if there are env variables defined (not empty?) and overwrite loaded `config.lua` values.

This PR changes TFS basic config and makes it pretty confusing for newbie users who use any tutorial 'how to run first OTS' (ex. https://otland.net/threads/cannot-open-config-lua-no-such-file-or-directory.290452/#post-2763504 ), as there is no MySQL config anymore in `config.lua`.

@ranisalt 
**What do you think about it?**
I can create PR that will make it work as I described above, if you think it's ok.

ranisalt:
@gesior you're correct, just need swapping the order to call `getEnv` with `getGlobalString` as a fallback and not the other way around. Feel free to open it or I'll to it soon

PR #4521, opened 2023-08-26
Improving Xiaolin Wu path finding w/ change factor

tornadia:
### Pull Request Prelude

### Changes Proposed

Improves upon the path finding algorithm used by Zbizu in [PR#3487](https://github.com/otland/forgottenserver/pull/3487), i.e. commit [7aa016fdf60bbd7916199d8674872e3a8ad05e18](https://github.com/otland/forgottenserver/commit/7aa016fdf60bbd7916199d8674872e3a8ad05e18).

**Issues addressed:** 
Initially mwalls positioning, but also throw items sight clear. For reference:

Broken:
https://github.com/otland/forgottenserver/assets/32970897/86ad27c6-d727-4284-9e85-58d65cbbc15e

Fix:
https://github.com/otland/forgottenserver/assets/32970897/029aa5d8-e090-407c-b7cc-51f10e1e17fe
https://github.com/otland/forgottenserver/assets/32970897/eb461fd9-c3fd-4ec5-ab56-011a852cceff

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -509,38 +509,44 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
 
 namespace {
 
-bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
+bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool reversed)
 {
 	float dx = x1 - x0;
-	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
-	float yi = y0 + slope;
+	float slope = (dx == 0) ? 1 : std::abs((y1 - y0) / dx);
+	if (y1 < y0) {
+		slope *= -1;
+	}
+    float yi = y0 + slope + (y0 == y1 ? 0 : (reversed && y0 < y1 || !reversed && y1 < y0 ? 0.85 : 0.05));
 
 	for (uint16_t x = x0 + 1; x < x1; ++x) {
-		// 0.1 is necessary to avoid loss of precision during calculation
+		// g_game.addMagicEffect(Position(std::floor(yi + 0.1),x,z), CONST_ME_POFF); // debugging
+		//0.1 is necessary to avoid loss of precision during calculation
 		if (!g_game.map.isTileClear(std::floor(yi + 0.1), x, z)) {
 			return false;
 		}
 		yi += slope;
 	}
-
-	return true;
 }
 
-bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z)
+bool checkSlightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z, bool reversed)
 {
-	float dx = x1 - x0;
-	float slope = (dx == 0) ? 1 : (y1 - y0) / dx;
-	float yi = y0 + slope;
+    float dx = x1 - x0;
+    float slope = (dx == 0) ? 1 : std::abs((y1 - y0) / dx);
+    if (y1 < y0) {
+        slope *= -1;
+    }
+    float yi = y0 + slope + (y0 == y1 ? 0 : (reversed && y0 < y1 || !reversed && y1 < y0 ? 0.05 : 0.85));
 
-	for (uint16_t x = x0 + 1; x < x1; ++x) {
-		// 0.1 is necessary to avoid loss of precision during calculation
-		if (!g_game.map.isTileClear(x, std::floor(yi + 0.1), z)) {
-			return false;
-		}
-		yi += slope;
-	}
+    for (uint16_t x = x0 + 1; x < x1; ++x) {
+        // g_game.addMagicEffect(Position(x,std::floor(yi + 0.1),z), CONST_ME_EXPLOSIONHIT); // debugging
+        //0.1 is necessary to avoid loss of precision during calculation
+        if (!g_game.map.isTileClear(x, std::floor(yi + 0.1), z)) {
+            return false;
+        }
+        yi += slope;
+    }
 
-	return true;
+    return true;
 }
 
 } // namespace
@@ -551,18 +557,20 @@ bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uin
 		return true;
 	}
 
+	// from top to bottom
 	if (std::abs(y1 - y0) > std::abs(x1 - x0)) {
 		if (y1 > y0) {
-			return checkSteepLine(y0, x0, y1, x1, z);
+			return checkSteepLine(y0, x0, y1, x1, z, true);
 		}
-		return checkSteepLine(y1, x1, y0, x0, z);
+		return checkSteepLine(y1, x1, y0, x0, z, false);
 	}
 
+	// from left to right
 	if (x0 > x1) {
-		return checkSlightLine(x1, y1, x0, y0, z);
+		return checkSlightLine(x1, y1, x0, y0, z, true);
 	}
 
-	return checkSlightLine(x0, y0, x1, y1, z);
+	return checkSlightLine(x0, y0, x1, y1, z, false);
 }
 
 bool Map::isSightClear(const Position& fromPos, const Position& toPos, bool sameFloor /*= false*/) const


tornadia:
credits to dc: `hide_zadymiarz` for finding this bug (Hide)

tornadia:
@ranisalt @nekiro @MillhioreBT review pl0x

Zbizu:
not a bug
obraz

tornadia:
> not a bug obraz

Did you try that in RL or your own server? It makes more sense that the range of throwable events matches a more cone-shaped one... for reference:
https://i.imgur.com/KJgCbub.mp4

tornadia:
Also see:
https://cdn.discordapp.com/attachments/965188004694065204/1145091269350138056/client_Re52B1CyHU.gif

ranisalt:
This is a known >bug< in Tibia for ages. I very recently wrote a [lengthy response on Reddit](https://www.reddit.com/r/TibiaMMO/comments/15zz3bh/comment/jxjrzur/) to someone who noticed the same in Cipsoft Tibia.

We tried to fix this in the past, but people grew so used to it that they do not want it to be fixed.

EPuncker:
> This is a known >bug< in Tibia for ages. I very recently wrote a [lengthy response on Reddit](https://www.reddit.com/r/TibiaMMO/comments/15zz3bh/comment/jxjrzur/) to someone [who noticed the same in Cipsoft Tibia](https://www.reddit.com/r/TibiaMMO/comments/15zz3bh/why_is_this_not_patched_yet_cant_believe_its_a/).
> 
> We tried to fix this in the past, but people grew so used to it that they do not want it to be fixed.

perhaps we could add it with a config.lua option with default value the "buggy" behavior?

Zbizu:
It's a matter of preference which behaviour of mw you want in your server. Wu algo is 1:1. Old tfs also is known from 8.6 servers and there are also 8.0 players who preferred old asymmetric sightline (same as old tfs but one ray only) for pvp purposes.

tornadia:
Interesting opinions in this thread.

> This is a known >bug< in Tibia for ages. I very recently wrote a [lengthy response on Reddit](https://www.reddit.com/r/TibiaMMO/comments/15zz3bh/comment/jxjrzur/) to someone [who noticed the same in Cipsoft Tibia](https://www.reddit.com/r/TibiaMMO/comments/15zz3bh/why_is_this_not_patched_yet_cant_believe_its_a/).
> 
> We tried to fix this in the past, but people grew so used to it that they do not want it to be fixed.

I agree with this post. I will add it to the PR, in case it should be reviewed and merged. I believe it's a good idea to keep cross compatibility with both preferences, and more appealing considering our userbase. If **they** (the unmentionable ones) do not want to fix broken functionality, good, good for them. But we should allow this software's users to decide if they want to make the same foolish choices.

EPuncker (dismissed):

ranisalt:
I would rather have a compile-time switch, to avoid paying the cost in runtime. It's not like people will change this a lot. Wdyt @EPuncker

EPuncker:
> I would rather have a compile-time switch, to avoid paying the cost in runtime. It's not like people will change this a lot. Wdyt @EPuncker

sounds better

DSpeichert:
> 



> I would rather have a compile-time switch, to avoid paying the cost in runtime. It's not like people will change this a lot. Wdyt @EPuncker

That's not very friendly for binary distributions :/

MillhioreBT:
> I would rather have a compile-time switch, to avoid paying the cost in runtime. It's not like people will change this a lot. Wdyt @EPuncker

I agree @ranisalt

those little checks will add an extra delay to everything related to line of sight, especially on expansive spells.

Plus it looks ugly.
I think it's time to get used to how Tibia RL is.

This seems more like a change for: Nekiro Downgrades

tornadia:
> > 
> 
> > I would rather have a compile-time switch, to avoid paying the cost in runtime. It's not like people will change this a lot. Wdyt @EPuncker
> 
> That's not very friendly for binary distributions :/

How so? Just need to provide 1 binary distro with the default, being zbizu's original.
I will add the changes soon. I do not agree with the comment above, although I do agree in it's present state the code is not optimal. Compile time variables will fix this.

src/map.cpp
```diff
@@ -509,13 +509,20 @@ bool Map::isTileClear(uint16_t x, uint16_t y, uint8_t z, bool blockFloor /*= fal
...
+	if (y1 < y0) {
+		slope *= -1;
+	}
 	float yi = y0 + slope;
+	if (modifier != 0) {
+		yi += modifier;
+	}
 
 	for (uint16_t x = x0 + 1; x < x1; ++x) {
+		// g_game.addMagicEffect(Position(std::floor(yi + 0.1),x,z), CONST_ME_POFF); // debugging
```
MillhioreBT, line 525: What do these codes commented in a PR do to merge?

src/map.cpp
```diff
@@ -526,13 +533,20 @@ bool checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t
...
+	if (y1 < y0) {
+		slope *= -1;
+	}
 	float yi = y0 + slope;
+	if (modifier != 0) {
+		yi += modifier;
+	}
 
 	for (uint16_t x = x0 + 1; x < x1; ++x) {
+		// g_game.addMagicEffect(Position(x,std::floor(yi + 0.1),z), CONST_ME_EXPLOSIONHIT); // debugging
```
MillhioreBT, line 549: What do these codes commented in a PR do to merge?

src/map.cpp
```diff
@@ -551,18 +565,33 @@ bool Map::checkSightLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uin
 		return true;
 	}
 
+	// from top to bottom
+	float modifier = 0;
 	if (std::abs(y1 - y0) > std::abs(x1 - x0)) {
 		if (y1 > y0) {
-			return checkSteepLine(y0, x0, y1, x1, z);
+			if (!g_config.getBoolean(ConfigManager::CLASSIC_ATTACK_SPEED) && x0 != x1) {
```
MillhioreBT, line 572: What does CLASSIC_ATTACK_SPEED have to do with this?

MillhioreBT (changes requested):
In general, I don't like this PR.
and not because of the behavior of the old tibia vs the new one.
if not rather those checks within the line of sight calculations.

tornadia:
> In general, I don't like this PR.
Hey, I guess if you don't like the PR then we can just close it and forget everything. Because that's how things work right?

> and not because of the behavior of the old tibia vs the new one.
Uh-huh. It's not old XXX vs new XXX, it's how TFS used to work and how it works now, more accurately depicting the gameplay sought after by the 1:1 experience playerbase. It was fixed by zbizu and now it's being re-added for people who will it or not.

> if not rather those checks within the line of sight calculations.
As I mentioned before, with compile time variables it will not be an issue. I will address your reviews now, thanks for finding those issues and bringing them forth.

src/map.cpp
tornadia: Initial confounding. This was meant to be `ORIGINAL_SIGHT_CALCULATION`, which will be removed regardless in favor of precompiler code substitution.

src/map.cpp
tornadia: Debug logic. It can be removed, at the small hindsight of TFS code pollution I propose that it shall have precompiler variables to debug certain scenarios, this being one of them.

src/map.cpp
tornadia: Please, see the other comment with identical response.

tornadia:
So, we have some decision making to do. Do you guys want to:
a) Include precompiler variables from some `*.cmake` file?
image

b) Add it as a CMake parameter and visual studio?
image
image
image

c) Hard-code it within the map file and let ppl decide in-code in true ad hoc fashion?
```cpp
// COMMENT THE LINE BELOW IF YOU WANT THE CLASSIC PATHFIND
#define _USE_TFS_PATHFINDING

#ifdef _USE_TFS_PATHFINDING
bool Map::checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const
{
// code #1
}
#else
bool Map::checkSteepLine(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1, uint8_t z) const
{
// code #2
}
#endif
```

Let me know which one you guys prefer. In any case, we can drop the PR. I don't really care.

ranisalt:
@tornadia definitely B

Zbizu:
@tornadia there is no need to adjust cmake settings, this can be put in definitions.h

tornadia:
@Zbizu ok I will see if I add that.
@MillhioreBT sorry for my previous post, that was uncalled for

felipepips:
AOE spells are passing through the walls with this changes..

original
modified

original2:
https://jmp.sh/s/RVzafFCxBNfzXGseBGB6
modified2:
https://jmp.sh/s/erBFTo6fVdJrheBdROEX

PR #4524, opened 2023-09-02
Support for hundredths after the decimal point

ArturKnopik:
### Pull Request Prelude

### Changes Proposed

image

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1800,7 +1800,7 @@ void Player::addExperience(Creature* source, uint64_t exp, bool sendText /* = fa
 	}
 
 	if (nextLevelExp > currLevelExp) {
-		levelPercent = Player::getPercentLevel(experience - currLevelExp, nextLevelExp - currLevelExp);
+		levelPercent = Player::getPercentLevel(experience - currLevelExp, nextLevelExp - currLevelExp) / 100;
 	} else {
 		levelPercent = 0;
 	}
@@ -1883,7 +1883,7 @@ void Player::removeExperience(uint64_t exp, bool sendText /* = false*/)
 
 	uint64_t nextLevelExp = Player::getExpForLevel(level + 1);
 	if (nextLevelExp > currLevelExp) {
-		levelPercent = Player::getPercentLevel(experience - currLevelExp, nextLevelExp - currLevelExp);
+		levelPercent = Player::getPercentLevel(experience - currLevelExp, nextLevelExp - currLevelExp) / 100;
 	} else {
 		levelPercent = 0;
 	}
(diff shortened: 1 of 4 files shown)


EPuncker (dismissed):

src/player.cpp
```diff
@@ -1892,14 +1892,14 @@ void Player::removeExperience(uint64_t exp, bool sendText /* = false*/)
...
-uint8_t Player::getPercentLevel(uint64_t count, uint64_t nextLevelCount)
+uint16_t Player::getPercentLevel(uint64_t count, uint64_t nextLevelCount)
 {
 	if (nextLevelCount == 0) {
 		return 0;
 	}
 
-	uint8_t result = (count * 100) / nextLevelCount;
-	if (result > 100) {
+	uint16_t result = ((count * 100.) / nextLevelCount * 100.);
```
ranisalt, line 1901:
```suggestion
	uint16_t result = ((count * 10000.) / nextLevelCount);
```

ArturKnopik: Are you sure about this?? Doesn't look right at first glance, of course I could be wrong

ArturKnopik:
TODO: 
Player::death
Player::removeExperience
 Player::addExperience

src/player.cpp
ranisalt: Yes, the parenthesis here are not necessary since you only use multiplications (division is multiplication) and they already execute left to right. So

```cpp
((count * 100.) / nextLevelCount * 100.)
```

is equivalent to

```cpp
count * 100. / nextLevelCount * 100.
```

So you can just group 100 * 100 into 10000

EPuncker (approved):

ranisalt:
Hey, before we merge, can you rename these functions from `get...Percent` to `get...BasisPoints`? Since it returns the 1/10000, it is incorrect to call it percent, [basis points](https://en.wikipedia.org/wiki/Basis_point) is the correct term

ArturKnopik:
> Hey, before we merge, can you rename these functions from `get...Percent` to `get...BasisPoints`? Since it returns the 1/10000, it is incorrect to call it percent, [basis points](https://en.wikipedia.org/wiki/Basis_point) is the correct term

Good idea :)(

EPuncker (dismissed):

src/player.h
```diff
@@ -1310,7 +1310,7 @@ class Player final : public Creature, public Cylinder
 
 	uint32_t getAttackSpeed() const;
 
-	static uint8_t getPercentLevel(uint64_t count, uint64_t nextLevelCount);
+	static uint16_t getBasicPointLevel(uint64_t count, uint64_t nextLevelCount);
```
DSpeichert, line 1313: I think it's supposed to be "basi**s** point", not "basi**c** point", it's a typo

DSpeichert (changes requested):

src/player.h
ArturKnopik: yeap, typo...

ranisalt (approved):

DSpeichert (approved):

Issue #4333, opened 2023-02-23
stacks with 200 or 255 instead of 100


Zbizu:
### Explanation of what you want to do that is currently impossible
QT client and OTC have full client-side support for items stacked as 255
Tried to implement it myself, but the way item stacking works has to be rewritten because evil things happen when you try to combine two stacks (lots of hardcoded 100s there also)

### Desired functionality
stacking items as 200 or 255 (preferable with config.lua option)

### Available workarounds
commenting a line in `item:transform()` source code and using it to change item subtype

### Prior art
![obraz](https://user-images.githubusercontent.com/5205079/221027220-6506327d-4dbc-4b8c-8de9-14383d46dd2a.png)

![obraz](https://user-images.githubusercontent.com/5205079/221027286-6e0c21df-1a58-4919-9237-dd0a87fb9567.png)

EPuncker:
that would be cool

conde2:
Agreed with this feature, very useful.

PR #4528, opened 2023-09-02
Configurable stack size

ArturKnopik:
Issue: https://github.com/otland/forgottenserver/issues/4333

### Pull Request Prelude

### Changes Proposed
Stack size can be set in game.h
static constexpr uint8_t ITEM_STACK_SIZE = 100;

image: stack size set to 250 (max 255)
image

### Note
It requires many tests to verify that each case is covered, a lot of magic 100 values

### DIFF
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -20,7 +20,7 @@ function Container.createLootItem(self, item)
 	end
 
 	while itemCount > 0 do
-		local count = math.min(Game.getMaxStackSize(), itemCount)
+		local count = math.min(ITEM_STACK_SIZE, itemCount)
 
 		local subType = count
 		if itemType:isFluidContainer() then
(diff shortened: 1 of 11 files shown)


ranisalt:
If it's meant to be configurable, what about adding to config.lua?

Otherwise, push the constant to Lua with `registerEnum` to avoid the function call

ArturKnopik:
> If it's meant to be configurable, what about adding to config.lua?
> 
> Otherwise, push the constant to Lua with `registerEnum` to avoid the function call
Configuration I meant you can change one value and everything works, usually the owner won't want to change it so the value is set at compile time for optimization

MillhioreBT:
> If it's meant to be configurable, what about adding to config.lua?
> 
> Otherwise, push the constant to Lua with `registerEnum` to avoid the function call

@ArturKnopik do what @ranisalt says, why make a call to lua when we can already have the constant in the global environment with registerEnum, the call `Game.getMaxStackSize` is unnecessary

ranisalt:
That, keep in mind that `registerEnum` does not do what it says, it registers any constant with the same name, enum or not.

EPuncker (approved):

ranisalt (approved):

PR #4531, opened 2023-09-07
Decay time random

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Added the possibility to items to have random decay time (`minvalue`, `maxvalue`)

**Issues addressed:** closes [#3148](https://github.com/otland/forgottenserver/issues/3148)


EPuncker (approved):

src/item.cpp
```diff
@@ -238,7 +238,12 @@ void Item::setID(uint16_t newid)
 	id = newid;
 
 	const ItemType& it = Item::items[newid];
-	uint32_t newDuration = it.decayTime * 1000;
+	uint32_t newDuration;
+	if (it.decayTimeMax != 0) {
+		newDuration = normal_random(it.decayTime, it.decayTimeMax) * 1000;
+	} else {
+		newDuration = it.decayTime * 1000;
+	}
```
ranisalt, line 246: You don't need this branch, if min and max are the same, `normal_random` will always return that value

MillhioreBT: > You don't need this branch, if min and max are the same, `normal_random` will always return that value

Do you mean IF this way?
```cpp
if (it.decayTimeMax != it.decayTime) { ... }
```

ranisalt: No if.

```
uint32_t newDuration = normal_random(it.decayTime, it.decayTimeMax) * 1000;
```

In the case that `it.decayTime == it.decayTimeMax` it will just return that value.

Speaking of which, it's probably wise to rename `decayTime` to `decayTimeMin`.

EPuncker (approved):

Issue #4536, opened 2023-09-19
Mailbox steal depot items bug, crash OTS bug [clone items]


gesior:
Tested on TFS 1.4

### Steps to reproduce
You will need:
- 2 players
- mailbox within 1 SQM from depot [as it's in Kazordoon on 13+], so player can stand next to depot and mailbox in same time
![image](https://github.com/otland/forgottenserver/assets/2642737/5ed0d15d-3c52-4e52-b5dd-11e0f75a3dc9)

Test 1 is random guy who visits depot.
Test 2 is attacker who wants to steal items/crash OTS.

**How to reproduce:**
1. Test 1 goes to depot locker and opens it. Then he walks away.
2. Test 2 goes to same depot locker, so he is within 1 SQL from depot locker and mailbox.
3. Test 2 put Parcel addressed to Test 1 in his BP.
4. Test 2 opens BP in one window in client and Parcel in second window - he sees Parcel from outside [can throw it] and also content of Parcel.
5. Test 2 puts Parcel on mailbox next to him. Parcel converts into 'Stamped parcel'.
6. Now Test 2 can click 'move up' in 'Stamped parcel' and he is inside Test 1 Inbox (depot)!

Now Test 2 can steal items of Test 1.
If Test 1 logouts and Test 2 try to move item in Test 1 depot, it will crash server.
I also heard about people who were able to clone items without crashing server on otservbr, but I cannot reproduce it on TFS.

**Problems:**
**Step 1** binds Test 1 item Inbox position to position on which player opened locker last time.
**Step 5** checks, if item moved is within 1 SQM range and does not close it. If mailbox and 'Test 1 Inbox' positions are within 1 SQM from player who sends parcel (Test 2), it does not close sent Parcel.

Looks like someone who added Inbox system in TFS forgot to copy code that closes containers, when they are moved into someone's depot. Inbox is not inside Depot Chest, so it does not protect parcels anymore.

### Expected behaviour
Parcel gets closed in client.

### Actual behaviour
Parcel stays open. You can click 'move up' in client and get into other player depot!

Reported by **Christianlb** on Discord. He asked me for help with crash analysis, then he found crash reason himself and told me how to reproduce it.
Already reproduced [abused] on top OTSes. Gunzodus solution: block possibility to send parcels, when you are within 1 SQM from Mailbox :smiley:

ArturKnopik:
> 1. Test 1 goes to depot locker and opens it. Then he walks away.
[...]

Issue #4535, opened 2023-09-14
Parcel Clone Bug


Mirkaanks:
All TFS based codes had this error, so this is aplicable to any Canary/Otbr version.
https://gifyu.com/image/S4NL6

by [marcosvf132](https://github.com/marcosvf132)

Possible fix: https://github.com/opentibiabr/canary/pull/1598

ArturKnopik:
I can't reproduce on top of master, can someone confirm this issue?

EPuncker:
when will people learn that canary/otbr base is OTX and not tfs?

unless someone provides real proof that this issue is present on tfs master, this issue is useless

PR #4537, opened 2023-09-19
#4536 close parcel send thru mail system to other player

gesior:
Auto close parcel, when it's send to other player thru mail system.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -3094,6 +3094,12 @@ void Player::postRemoveNotification(Thing* thing, const Cylinder* newParent, int
 					if (!isOwner) {
 						autoCloseContainers(container);
 					}
+				} else if (const Inbox* inboxContainer = dynamic_cast<const Inbox*>(topContainer)) {
+					if (inboxContainer == inbox) {
+						onSendContainer(container);
+					} else {
+						autoCloseContainers(container);
+					}
 				} else {
 					onSendContainer(container);
 				}


EPuncker (approved):

Berzekz (approved):

Zbizu (approved):

luanluciano93:
The solution referenced in issue #4535 is better, but you ignore it because it is a canary.

joseluis2g:
> The solution referenced in issue #4535 is better, but you ignore it because it is a canary.

so if you really think that solution is somehow "better", would you be able to elaborate how that PR is better?

gesior:
> The solution referenced in issue #4535 is better, but you ignore it because it is a canary.

I would call it hotfix. Like Gunzodus's "block possibility to send parcel, when you are standing 1 sqm from mailbox". It fixed crash issue, but not solved real problem.

Canary fix does not solve all possible scenarios [ex. moving container into player Inbox from Lua - without 'mail system'] and add new bug to engine: close parcel when you put it on mailbox, even if it's not delivered to player [ex. it's addressed to player that does not exist].

EPuncker:
just a note in case anyone is wondering on why this PR has not been merged _yet_, it was targeted specifically to branch 1.4, which currently doesn't have any active maintainer thus we still haven't decided how to handle this as of yet

if anyone is feeling like doing so, open a new PR targeting master branch and adding the due credits and we will merge it as soon as possible

Zbizu:
> The solution referenced in issue #4535 is better, but you ignore it because it is a canary.

Canary solution is fetching tiles and spectators again and doing at least two extra loops (getSpectators and "for" spectators) when all that had to be done was to add one "if" statement in place that already had all required variables and was already part of the loop that is going through the container spectators.

in short: canary solution has two unnecessary loops

PR #4544, opened 2023-09-30
Improve IO ban information

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Use optional when instance a ban info
- Define "none" when empty reason 

### DIFF
--- a/src/ban.cpp
+++ b/src/ban.cpp
@@ -44,7 +44,7 @@ bool Ban::acceptConnection(const Connection::Address& clientIP)
 	return true;
 }
 
-const std::shared_ptr<BanInfo> IOBan::isAccountBanned(uint32_t accountId)
+const std::optional<BanInfo> IOBan::isAccountBanned(uint32_t accountId)
 {
 	Database& db = Database::getInstance();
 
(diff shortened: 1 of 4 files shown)


EPuncker (approved):

src/ban.cpp
```diff
@@ -44,15 +44,15 @@ bool Ban::acceptConnection(const Connection::Address& clientIP)
 	return true;
 }
 
-bool IOBan::isAccountBanned(uint32_t accountId, BanInfo& banInfo)
+const std::shared_ptr<BanInfo> IOBan::isAccountBanned(uint32_t accountId)
```
ranisalt, line 47: `std::optional<BanInfo>` is a better option, since it will be stack allocated and use less memory.

src/ban.cpp
```diff
@@ -44,15 +44,15 @@ bool Ban::acceptConnection(const Connection::Address& clientIP)
 	return true;
 }
 
-bool IOBan::isAccountBanned(uint32_t accountId, BanInfo& banInfo)
+const std::optional<BanInfo> IOBan::isAccountBanned(uint32_t accountId)
```
MillhioreBT, line 47: The methods should be called: `getMethodName` instead of `isMethodName`.

MillhioreBT (changes requested):

src/ban.cpp
ramon-bernardo: _has_ may be better than _get_

EPuncker: both ugly 😄

Shawak: Why not call them `getAccountBanInfo` and `getIPBanInfo`?

EPuncker (approved):

EPuncker (approved):

MillhioreBT (approved):

PR #4551, opened 2023-10-07
Obtain the master of summons on death

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Possibility of obtaining the `Master` of the `Summons` on death.
[More info](https://github.com/otland/forgottenserver/pull/4551#issuecomment-1788173083)

**Issues addressed:** [It should happen the same way in this repository.](https://github.com/MillhioreBT/forgottenserver-downgrade/issues/7)

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -717,6 +717,10 @@ void Creature::onDeath()
 
 	if (master) {
 		setMaster(nullptr);
+
+		if (getMonster()) {
+			decrementReferenceCounter();
+		}
 	}
 
 	if (droppedCorpse) {
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -718,7 +718,7 @@ void Creature::onDeath()
 	if (master) {
 		setMaster(nullptr);
 
-		if (getMonster()) {
+		if (dynamic_cast<Monster*>(this) != nullptr) {
 			decrementReferenceCounter();
 		}
 	}
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1840,7 +1840,6 @@ void Monster::death(Creature*)
 
 	for (Creature* summon : summons) {
 		summon->changeHealth(-summon->getHealth());
-		summon->removeMaster();
 	}
 	summons.clear();
 


EPuncker (approved):

conde2:
Could you explain better your changes?

For what I see the 
removeMaster is called in the summons list not in the creature from onDeath, therefore it should go inside the
```
	if (master) {
		setMaster(nullptr);
	}
```

conde2:
After reading the issue I could understand the changes.

The Master dies, removes the summon then the summon die but now it is not considered summon anymore because removeMaster was called for the summons of the master.



There is this change that addTask in the onDeath for the changeHealth
https://github.com/otland/forgottenserver/pull/3030

Removing this addTask would fix the problem, but not sure what are the consequences in doing it.

MillhioreBT:
> Removing this addTask would fix the problem, but not sure what are the consequences in doing it.
Executing the method immediately is not good because this method is called too much on a production server.

There's nothing wrong with the change I made, we're just now removing the master in a place where it doesn't affect the data that should be expected in the onKill event.

conde2:
Although there is nothing wrong with it, is just that is hard to understand why the decrement reference is there. Thinking of it, if you are a programmer and don't see this issue, you probably will take a lot of time to figure it out.

Thats why I gave 2 suggestion to improve readability

MillhioreBT:
> Although there is nothing wrong with it, it is hard to understand why the decrement reference is there. Thinking of it, if any programmer don't see this issue (#4551), they will probably have a hard time to figure it out.
> 
> Thats why I gave 2 suggestion to improve readability

This is the death event, onDeath is called when the monster is already considered dead, that means it should be removed from memory

Why would decrementing a reference have no relationship?

conde2:
```
Why would decrementing a reference have no relationship?
```

And where I stated that?


Look, I'm saying that it is difficult do understand what is happening, for example why you are decrementing reference 2 times for example?

1 at setMaster(nullptr) and the other after in the if statement?

image

Imagine you never read through the code. Isn't that confusing?

MillhioreBT:
> 1 at setMaster(nullptr) and the other after in the if statement?

I was reading the code again to be sure, and Yes, is necessary delete one more reference, otherwise the object is not deleted.
I invite you to add some impressions of the references with the changes you propose and you will understand what I say.

Obviously I didn't make the changes, without having tested them first.

It seems strange, but what happens is that the dispatcher sometimes takes care of the task of deleting the master first, so when setMaster is called, the master actually no longer exists, so setMaster(nullptr) will no longer dereference it, because [this line](https://github.com/otland/forgottenserver/blob/16a985c33c80544af0ac0668bc35da11cfcaf9f9/src/creature.cpp#L1192) will be false

EPuncker (approved):

Issue #1431, opened 2015-06-25
[BUG] Too many items in the inbox.


luanlucian93:
Look:
http://i.imgur.com/MqKWFn2.png

I use command: 
select `player_id`, count(`player_id`) from `player_inboxitems` group by `player_id` having count(`player_id`) > 500

Look:
http://i.imgur.com/2DHE9ZG.png

Help me?

ranisalt:
It's not a bug and is probably a duplicate of #1403

PR #4561, opened 2023-11-08
[WIP] Asynchronous Inbox implementation

gunzino:

### Changes Proposed

I'm happy to share a contribution to the TFS which consists of async, zero-blocking Inbox implementation to prevent abusing the current sync system with large Inboxes and causing lags due to database IO like mentioned here #1431. The similar implementation is rotating between server owners for real life money but does not include complete zero-blocking IO for both saving and loading and parcel/item delivery to Inbox when player is offline. The implementation can be further used for more components like depots, store inbox etc. but those are currently either limited or controlled by server.

## The implementation highlights:
- The database IO and some of the serialization is done in separate thread (secondary Dispatcher g_dispatcherInbox) with dedicated database connection
- Data structures protected by mutex
- Async zero-blocking delivery of parcels and items from market/houses

Items delivery currently done for:
- Houses

TODO (will work on in next days, just have to implement usage of ```void IOInbox::pushDeliveryItems```)
- Market
- Parcels

I would be really happy for your hints/recommendations regarding so this contribution can be pushed to the master branch.

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -682,11 +682,20 @@ void Container::postRemoveNotification(Thing* thing, const Cylinder* newParent, 
 
 void Container::internalAddThing(Thing* thing) { internalAddThing(0, thing); }
 
-void Container::internalAddThing(uint32_t, Thing* thing)
+void Container::internalAddThing(uint32_t index, Thing* thing)
 {
 	Item* item = thing->getItem();
 	if (!item) {
 		return;
+	}
+
+	if (index == 0 || itemlist.empty()) {
+		itemlist.push_front(item);
+	} else {
+		size_t pos = std::min<size_t>(std::max<size_t>(0, itemlist.size() - 1), index);
+		auto it = itemlist.begin();
+		std::advance(it, pos);
+		itemlist.insert(it, item);
 	}
 
 	item->setParent(this);
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -213,14 +213,14 @@ bool DBResult::next()
 	return row;
 }
 
-DBInsert::DBInsert(std::string query) : query(std::move(query)) { this->length = this->query.length(); }
+DBInsert::DBInsert(std::string query, Database& db) : query(std::move(query)), db(db) { this->length = this->query.length(); }
 
 bool DBInsert::addRow(const std::string& row)
 {
 	// adds new row to buffer
 	const size_t rowLength = row.length();
 	length += rowLength;
-	if (length > Database::getInstance().getMaxPacketSize() && !execute()) {
+	if (length > db.getMaxPacketSize() && !execute()) {
 		return false;
 	}
 
@@ -253,7 +253,7 @@ bool DBInsert::execute()
 	}
 
 	// executes buffer
-	bool res = Database::getInstance().executeQuery(query + values);
+	bool res = db.executeQuery(query + values);
 	values.clear();
 	length = query.length();
 	return res;
--- a/src/database.h
+++ b/src/database.h
@@ -162,13 +162,14 @@ private:
 class DBInsert
 {
 public:
-	explicit DBInsert(std::string query);
+	explicit DBInsert(std::string query, Database& db = Database::getInstance());
 	bool addRow(const std::string& row);
 	bool addRow(std::ostringstream& row);
 	bool execute();
 
 private:
 	std::string query;
+	Database& db;
 	std::string values;
 	size_t length;
 };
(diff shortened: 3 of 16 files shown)


src/container.cpp
```diff
@@ -682,13 +682,22 @@ void Container::postRemoveNotification(Thing* thing, const Cylinder* newParent,
...
 	if (!item) {
 		return;
 	}
 
+	if (index == 0 || itemlist.empty()) {
+		itemlist.push_front(item);
+	} else {
+		size_t pos = std::min<size_t>(std::max<size_t>(0, itemlist.size() - 1), index);
+		auto it = itemlist.begin();
+		std::advance(it, pos);
```
ranisalt, line 697: You don't need to check the bounds when advancing from `begin()`, so

```suggestion
		auto it = itemlist.begin() + index;
```

Also works. You can also drop the `if` here, since `push_front(item)` and `insert(begin(), item)` are equivalent.

src/database.cpp
```diff
@@ -213,14 +213,17 @@ bool DBResult::next()
 	return row;
 }
 
-DBInsert::DBInsert(std::string query) : query(std::move(query)) { this->length = this->query.length(); }
+DBInsert::DBInsert(std::string query, Database& db) : query(std::move(query)), db(db)
+{
+	this->length = this->query.length();
```
ranisalt, line 218: `this->length` seems to always be `this->query.length()` so I think you can delete this attribute.

src/game.cpp
```diff
@@ -5629,11 +5694,23 @@ void Game::playerAcceptMarketOffer(uint32_t playerId, uint32_t timestamp, uint16
...
-					break;
+			if (!player->getInbox()) {
+				ItemBlockList inboxDelivery;
+				for (uint16_t i = 0; i < offer.amount; ++i) {
+					Item* item = Item::CreateItem(it.id, subType);
+					if (item) {
+						inboxDelivery.emplace_back(0, item);
+					}
+				}
+				IOInbox::getInstance().pushDeliveryItems(player->getGUID(), inboxDelivery);
```
ramon-bernardo, line 5705: It can be simplified by adding a Game function to add item for player inbox. See my example.

src/mailbox.cpp
```diff
@@ -81,24 +82,28 @@ bool Mailbox::sendItem(Item* item) const
...
-			IOLoginData::savePlayer(&tmpPlayer);
+		ItemBlockList inboxDelivery;
+		Item* cloneItem = item->clone();
+		if (g_game.internalRemoveItem(item) == RETURNVALUE_NOERROR) {
+			cloneItem->setID(cloneItem->getID() + 1);
+			inboxDelivery.emplace_back(0, cloneItem);
+			IOInbox::getInstance().pushDeliveryItems(player->getGUID(), inboxDelivery);
 			return true;
+		} else {
+			delete cloneItem;
```
ramon-bernardo, line 106: There has a problem, player may be null and is using its guid and not receiverGuid

ramon-bernardo (commented):
I liked the PR, it looks promising.

I made some improvements, removed duplicate code, see
https://github.com/ramon-bernardo/forgottserver/commit/0f34a2f209c3aeada80024ec809792717c4164a1

I'm still reviewing IOInbox!

src/container.cpp
gunzino: Fixed!

src/mailbox.cpp
gunzino: applied your improvements :)

gunzino:
@ramon-bernardo 
I've added async house saving today as well which uses similar approach. Can you review?

src/database.h
```diff
@@ -30,6 +30,12 @@ class Database
 		return instance;
 	}
 
+	static Database& getAsyncInstance()
+	{
+		static Database asyncInstance;
+		return asyncInstance;
+	}
```
ranisalt, line 37: IIRC `getInstance` already returns an "async" instance, used for other purposes but should be suitable. Why having another instance?

On that note, singleton sucks, but a "doubleton" is even worse 😆 let's just make it not a global and pass it around

gunzino: I agree that "doubleton" sux and also was actually about using databaseTasks but we need to run each save which consists of multiple queries in single transaction which is currently not supported by databaseTasks, it's only single query per task. So we need kind of QuerySet which can be pushed to databaseTasks and run in single transaction. What's the proposed approach?

src/ioinbox.h
```diff
@@ -0,0 +1,77 @@
...
+	void assignInbox(const uint32_t& guid, Inbox* inbox);
+	bool deliverItems(const uint32_t& guid, Inbox* inbox);
+	void savePlayerItemsAsync(const uint32_t& guid);
+
+	Database& db;
+	std::recursive_mutex lock;
+	std::map<uint32_t, PlayerDBEntryPtr> inboxCache;
+	std::map<uint32_t, std::list<ItemBlockList>> pendingItemsToSave;
+	std::set<uint32_t> pendingPlayerSet;
+};
```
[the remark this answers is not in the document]
gunzino: Is there similar approach somewhere else in the code? What about recursive mutex?

Codinablack:
This looks like a very promising PR, what is the status of this?

PR #4562, opened 2023-11-14
Set writer and date in setItemText method

rookgaard:
### Pull Request Prelude

### Changes Proposed

Since you can set the text of the item, it may also be useful for someone to add the author and date, e.g. to a parcel given by the server or to a quest where a letter is given to the player.

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -1066,13 +1066,13 @@ function doSetItemText(uid, text, writer, date)
 		item:removeAttribute(ITEM_ATTRIBUTE_TEXT)
 	end
 
-	if writer ~= "" then
+	if writer then
 		item:setAttribute(ITEM_ATTRIBUTE_WRITER, tostring(writer))
 	else
 		item:removeAttribute(ITEM_ATTRIBUTE_WRITER)
 	end
 
-	if date ~= "" then
+	if date then
 		item:setAttribute(ITEM_ATTRIBUTE_DATE, tonumber(date))
 	else
 		item:removeAttribute(ITEM_ATTRIBUTE_DATE)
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

data/lib/compat/compat.lua
```diff
@@ -1065,6 +1065,19 @@ function doSetItemText(uid, text)
 	else
 		item:removeAttribute(ITEM_ATTRIBUTE_TEXT)
 	end
+
+	if writer ~= "" then
```
MillhioreBT, line 1069: `if writer ~= nil then`

data/lib/compat/compat.lua
```diff
@@ -1065,6 +1065,19 @@ function doSetItemText(uid, text)
...
+		item:setAttribute(ITEM_ATTRIBUTE_WRITER, tostring(writer))
+	else
+		item:removeAttribute(ITEM_ATTRIBUTE_WRITER)
+	end
+
+	if date ~= "" then
```
MillhioreBT, line 1075: `if date ~= nil then`

MillhioreBT (changes requested):

data/lib/compat/compat.lua
EPuncker: we removed almost every nil from lua scripts, so I'd suggest:
```suggestion
	if writer then
```

same for date

EPuncker (changes requested):

data/lib/compat/compat.lua
luanluciano93:
```suggestion
	if writer and writer ~= "" then
```

data/lib/compat/compat.lua
luanluciano93:
```suggestion
	if date and date ~= "" then
```

MillhioreBT (approved):

EPuncker (approved):

PR #4563, opened 2023-11-16
Send 0/1 as value in quest completion check

rookgaard:
### Pull Request Prelude

### Changes Proposed

We should send 0 or 1 as value instead of true/false in addByte, just like here https://github.com/otland/forgottenserver/blob/a9049040279bba57ef987196fd548b6c730593ef/data/lib/core/player.lua#L380

### DIFF
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -417,7 +417,7 @@ function Player.sendQuestLog(self)
 	for _, quest in pairs(quests) do
 		msg:addU16(quest.id)
 		msg:addString(quest.name)
-		msg:addByte(quest:isCompleted(self))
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

ranisalt (approved):
I don't get why Lua had to reinvent every single behavior. It's the only language I know that can't coerce booleans to 0/1 lol

MillhioreBT:
> I don't get why Lua had to reinvent every single behavior. It's the only language I know that can't coerce booleans to 0/1 lol

In fact, it does, simply because the TFS API is like this
The getNumber method only converts the given value to a number if it is actually a number.
otherwise it will always return zero

PR #4578, opened 2023-12-01
refactor: replace nil checks with not in conditions

omarcopires:
### Pull Request Prelude

### Changes Proposed 
This commit introduces a refactor to replace nil checks with the use of not in logical conditions, aligning with the preferred style in the codebase. This change improves code readability and consistency throughout the affected sections.

### DIFF
--- a/data/creaturescripts/scripts/login.lua
+++ b/data/creaturescripts/scripts/login.lua
@@ -13,9 +13,12 @@ function onLogin(player)
 	-- Promotion
 	local vocation = player:getVocation()
 	local promotion = vocation:getPromotion()
-	if player:isPremium() and player:getStorageValue(PlayerStorageKeys.promotion) == 1 then
-		player:setVocation(promotion)
-	elseif not player:isPremium() and promotion then
+	if player:isPremium() then
+		local value = player:getStorageValue(PlayerStorageKeys.promotion)
+		if value and value == 1 then
+			player:setVocation(promotion)
+		end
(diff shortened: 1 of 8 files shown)


Zbizu:
this change introduces `nil` comparisons which tfs moved away from, except for situations when both nil and false can be returned by a function and need to be distinguished

if the statement does not need to distinguish between false and nil:
- `~= nil` is not necessary in logical statements
- `if x == nil then` should be replaced with `if not x then`

MillhioreBT:
What @Zbizu says is true, please make the change from `if not x then`
and for cases of setStorageValue(-1), change to removeStorageValue

data/creaturescripts/scripts/login.lua
```diff
@@ -13,20 +13,17 @@ function onLogin(player)
 	-- Promotion
 	local vocation = player:getVocation()
 	local promotion = vocation:getPromotion()
-	if player:isPremium() then
-		local value = player:getStorageValue(PlayerStorageKeys.promotion)
-		if value == 1 then
-			player:setVocation(promotion)
-		end
-	elseif not promotion then
+	if player:isPremium() and player:getStorageValue(PlayerStorageKeys.promotion) == 1 then
```
MillhioreBT, line 16: The value here can be null, leave it as it was, just add `value and value == 1`

MillhioreBT (changes requested):
Change what you check on line 16 in login.lua

PR #4580, opened 2023-12-04
removing obsolete methods: getStorageValue and setStorageValue

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Since enough time has passed that TFS users have learned that stores now return when empty, it is no longer necessary to maintain deprecated functions.


MillhioreBT (approved):
Since enough time has passed for TFS users to have learned that storages now return `nil` when empty, I approve of removing these two methods which are not really necessary beyond this simple warning

PR #4589, opened 2023-12-10
Database transaction & row insert in lua

ramon-bernardo:
### Pull Request Prelude
- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Possibility to start db transaction in lua
Create a db insert in lua

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4590, opened 2023-12-10
Move account storage to lua

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Move account storage to lua.

#4588 and #4589 need to be merged first


data/scripts/globalevents/account_storage.lua
```diff
@@ -0,0 +1,63 @@
+-- load accounts storages
+local load = GlobalEvent("LoadAccountStorage")
+
+function load.onStartup()
+	local resultId = db.asyncStoreQuery("SELECT `account_id`, `key`, `value` FROM `account_storage`")
```
ramon-bernardo, line 5: Any problems with being async? 🦆

data/scripts/globalevents/account_storage.lua
```diff
@@ -0,0 +1,63 @@
...
+load:register()
+
+-- save accounts storages
+local function trySaveAccountsStorage()
+	local transaction = DBTransaction()
+	if not transaction:begin() then
+		return false
+	end
+
+	local result = db.query("DELETE FROM `account_storage`")
```
ramon-bernardo, line 30: Some here

data/scripts/globalevents/account_storage.lua
MillhioreBT: you should leave it as it has always been

This change will possibly mean that when you want to verify a storage value in the onLogin event we will not be able to obtain the expected values ​​because it is likely that the asynchronous task of executing the query has not yet been executed.

Also this script is wrong, because this task doesn't really return a results ID, in fact it doesn't return anything...
An asynchronous task is supposed to have a callback that will be called when the query is executed, this callback is passed as the second parameter

```lua
db.asyncStoreQuery("SELECT `account_id`, `key`, `value` FROM `account_storage`", function(result) ... end)
```

ramon-bernardo: > you should leave it as it has always been
> 
> This change will possibly mean that when you want to verify a storage value in the onLogin event we will not be able to obtain the expected values ​​because it is likely that the asynchronous task of executing the query has not yet been executed.
> 
> Also this script is wrong, because this task doesn't really return a results ID, in fact it doesn't return anything... An asynchronous task is supposed to have a callback that will be called when the query is executed, this callback is passed as the second parameter
> 
> ```lua
> db.asyncStoreQuery("SELECT `account_id`, `key`, `value` FROM `account_storage`", function(result) ... end)
> ```

Right, I didn't notice that the callback would be a func. Added comments in luascript to help others

ramon-bernardo:
Ready to merge!

data/scripts/globalevents/account_storage.lua
```diff
@@ -0,0 +1,63 @@
...
+	if not resultId then
+		return true
+	end
+
+	repeat
+		local accountId = result.getNumber(resultId, "account_id")
+        local key = result.getNumber(resultId, "key")
+        local value = result.getNumber(resultId, "value")
+        Game.setAccountStorageValue(accountId, key, value)
+    until not result.next(resultId)
```
reyaleman, line 15: indentation

data/scripts/globalevents/account_storage.lua
```diff
@@ -0,0 +1,63 @@
...
+
+local save = GlobalEvent("SaveAccountStorage")
+
+function save.onSave()
+	local success = trySaveAccountsStorage()
+	if not success then
+		print("Failed to save account-level storage values.")
+	end
+	return true
+end
```
ramon-bernardo, line 61: Is it necessary to involve it in maintain game state?
image

data/scripts/globalevents/account_storage.lua
ramon-bernardo: Thanks

data/scripts/globalevents/account_storage.lua
ramon-bernardo: If this is necessary, wouldn't it be ideal to wrap saveEvent in a maintain state?

ramon-bernardo:
Ready to merge, right?

ranisalt (dismissed):

EPuncker (approved):

data/scripts/globalevents/account_storage.lua
Sajgon: @ramon-bernardo for me this onSave isn't triggered when running ``g_game.saveGameState();``, is that intended?

PR #4602, opened 2023-12-14
Add NetworkMessage deletion in Player.sendHotkeyPreset function

omarcopires:
### Pull Request Prelude

### Changes Proposed 
Adds network message deletion in the Player.sendHotkeyPreset function for optimization.

### DIFF
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -724,6 +724,7 @@ function Player.sendHotkeyPreset(self)
 	msg:addByte(0x9D)
 	msg:addU32(self:getVocation():getClientId())
 	msg:sendToPlayer(self)
+	msg:delete()
 	return true
 end
 


ranisalt:
This is not needed, it will be deleted when it is garbage collected

omarcopires:
> This is not needed, it will be deleted when it is garbage collected; if anything, this de-optimizes by deleting twice, the second being no-op

so it would be interesting to remove it from everyone else?

ranisalt:
> so it would be interesting to remove it from everyone else?

Logic-wise, it makes no difference. For consistency, then yes, it should always be called or never be called.

omarcopires:
> > so it would be interesting to remove it from everyone else?
> 
> 
> 
> Logic-wise, it makes no difference. For consistency, then yes, it should always be called or never be called.

I'll remove them all in a future pull request!

PR #4605, opened 2023-12-15
remove msg:delete() for consistency

omarcopires:
### Pull Request Prelude

### Changes Proposed 
This change removes all msg:delete() calls in Lua scripts to maintain consistency and avoid inconsistencies regarding memory management. The aim is to ensure uniform practice and improve code clarity and readability.
Discussed in: #4602

### DIFF
--- a/data/lib/core/player.lua
+++ b/data/lib/core/player.lua
@@ -348,7 +348,6 @@ function Player.updateKillTracker(self, monster, corpse)
 		msg:sendToPlayer(self)
 	end
 
-	msg:delete()
 	return true
 end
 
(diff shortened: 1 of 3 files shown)


MillhioreBT:
Lately many systems have moved to lua, and NetworkMessage objects have started to be used a lot.
These weigh ~24596 Bytes, over time these objects can accumulate until reaching ~100MB or even much more the more players are online using systems

The garbage collector starts a cycle every so often, but I think the time is too long. Don't you think it would be a problem if the garbage collector has to clean several GBs of information every so often?

Although `delete` cleans the memory it does not eliminate the reference of the object in lua, but this reference of the dead object does not consume anything, so the collector will do its job normally deleting the reference, but it will no longer be necessary to clean the object in C

Maybe the best option is to change the behavior of the garbagecollector

PR #4608, opened 2023-12-16
Move client debug asserts to lua + database

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Move client debug assert to lua.
- Use database instead files.
- Async store query.


data/scripts/network/debug_assert.lua
```diff
@@ -0,0 +1,16 @@
+local handler = PacketHandler(0xE8)
+
+function handler.onReceive(player, msg)
+	local playerGuid = player:getGuid()
+	if Game.hasDebugAssert(playerGuid) then
```
marmichalski, line 5: You are missing call to `Game.addDebugAssert` for this to ever have effect.

I see that this is reimplemented from C++, but I think it works differently here. As previously the state of `debugAssertSent` would be reset on reconnection (when character stayed in the game) in protocolgame.h. Here you are keeping that information until the character logs out, but character may stay in the game on crash and reconnect. Am I missing something?

ramon-bernardo: yeah! you are right regarding the protocol. I had no other idea than onJoin/Leave

ranisalt:
Does this still go through game protocol? I know the new client sends crash data to an HTTP endpoint

PR #4610, opened 2023-12-20
Move player idle to lua

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Move player idle to lua.

image


Codinablack (approved):
Looking good to me!

data/scripts/creaturescripts/player/idle_time.lua
```diff
@@ -0,0 +1,38 @@
...
+	local tile = Tile(player:getPosition())
+	if not tile or tile:hasFlag(TILESTATE_NOLOGOUT) then
+		return true
+	end
+
+	local idleTime = player:getIdleTime() + interval
+	player:setIdleTime(idleTime)
+
+	if idleTime == config.maxIdleTime then
+		player:sendTextMessage(MESSAGE_STATUS_WARNING, "There was no variation in your behaviour for " .. (config.maxIdleTime / (60 * 1000)) .. " minutes. You will be disconnected in one minute if there is no change in your actions until then.")
```
amatria, line 31:
```suggestion
		player:sendTextMessage(MESSAGE_STATUS_WARNING, "There was no variation in your behaviour for " .. (config.maxIdleTime / (60 * 1000)) .. " minutes. You will be disconnected in " .. (config.kickAfterWarn / (60 * 1000)) .. " minutes if there is no change in your actions until then.")
```

src/luascript.cpp
```diff
@@ -2725,6 +2724,7 @@ void LuaScriptInterface::registerFunctions()
 
 	registerMethod("Player", "isPzLocked", LuaScriptInterface::luaPlayerIsPzLocked);
 
+	registerMethod("Player", "hasClient", LuaScriptInterface::luaPlayerHasClient);
```
amatria, line 2727: Does this change belong in this PR? You register this method but are not using it anywhere

src/luascript.cpp
```diff
@@ -10644,6 +10645,18 @@ int LuaScriptInterface::luaPlayerIsPzLocked(lua_State* L)
...
+	// player:hasClient()
+	Player* player = getUserdata<Player>(L, 1);
+	if (player) {
+		lua_pushboolean(L, player->hasClient());
+	} else {
+		lua_pushnil(L);
+	}
+	return 1;
+}
+
```
amatria, line 10659: Again: I don't think this change belongs in this PR

src/luascript.cpp
```diff
@@ -17113,11 +17139,7 @@ int LuaScriptInterface::luaCreatureEventOnCallback(lua_State* L)
 	// creatureevent:onLogin / logout / etc. (callback)
 	CreatureEvent* creature = getUserdata<CreatureEvent>(L, 1);
 	if (creature) {
-		if (!creature->loadCallback()) {
-			pushBoolean(L, false);
-			return 1;
-		}
-		pushBoolean(L, true);
+		pushBoolean(L, creature->loadCallback());
```
amatria, line 17142: Good pick

src/luascript.h
```diff
@@ -1004,6 +1004,7 @@ class LuaScriptInterface
 
 	static int luaPlayerIsPzLocked(lua_State* L);
 
+	static int luaPlayerHasClient(lua_State* L);
```
amatria, line 1007: Again: I don't think this change belongs in this PR

src/player.h
```diff
@@ -211,6 +211,8 @@ class Player final : public Creature, public Cylinder
 		return client->getVersion();
 	}
 
+	bool hasClient() const { return client != nullptr; };
```
amatria, line 214: Again: I don't think this change belongs in this PR

data/scripts/creaturescripts/player/idle_time.lua
EPuncker: why is that? the kick happens exactly one minute after the message is sent, that is why it is "hard coded"

reyaleman: I think it's because he's making it "configurable" now, so if somebody sets to kick after 2 minutes it will still send the "one minute" message...

amatria: Exactly: he made it configurable through a table in the script
https://github.com/otland/forgottenserver/blob/5ccc6419e742623535aa938ec4673fc17ca37650/data/scripts/creaturescripts/player/idle_time.lua#L1-L4

src/luascript.cpp
MillhioreBT, line 17142: It's not bad, but why do you have to modify things that have nothing to do with the changes you mention?

data/scripts/creaturescripts/player/idle_time.lua
```diff
@@ -0,0 +1,38 @@
+local config = {
+	maxIdleTime = 15 * 60 * 1000, -- 15 minutes
+	kickAfterWarn = 1 * 60 * 1000 -- 1 minute
+}
+
+-- register player event
+local event = CreatureEvent("RegisterIdleTime")
+
+function event.onLogin(player)
+	if not player:getGroup():getAccess() then
```
MillhioreBT, line 10: The event should be added to all players, the group check is done in the onThink event, so that if you switch to a online player's group it will still work correctly.

MillhioreBT (changes requested):

PR #4619, opened 2024-01-12
New event Spell Check

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Did you ever think about blocking all spells, or maybe just one specific spell for a player? Or for a group of players? Or maybe something even crazier?

Well, it's actually possible to do that using global variables or player storage. However, in order for it to work properly, we really have to modify and add some extra conditionals to all the spells we want to control. This is totally horrible.

This new event comes to solve this problem. The `onSpellCheck` event has two arguments: one is the player and the other is a table with some fields. In these fields, you can find some properties of the spell in question, for example the name.

Here is an example code fragment using the new event:
```lua
local event = Event()

function event.onSpellCheck(player, spell)
	if BossEvent:hasPlayer(player) then
		player:sendCancelMessage("You can't use spells in this event.")
		return false
	end
	return true
end

event:register()
```

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -1342,7 +1342,7 @@ void Events::eventPlayerOnNetworkMessage(Player* player, uint8_t recvByte, Netwo
 
 bool Events::eventPlayerOnSpellCheck(Player* player, const Spell* spell)
 {
-	// Player:onCanCastSpell(spell)
+	// Player:onSpellCheck(spell)
 	if (info.playerOnSpellCheck == -1) {
 		return true;
 	}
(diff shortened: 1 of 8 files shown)


src/events.cpp
```diff
@@ -1338,6 +1340,32 @@ void Events::eventPlayerOnNetworkMessage(Player* player, uint8_t recvByte, Netwo
 	scriptInterface.callVoidFunction(3);
 }
 
+bool Events::eventPlayerOnSpellCheck(Player* player, const Spell* spell)
+{
+	// Player:onCanCastSpell(spell)
```
EPuncker, line 1345: Player:onSpellCheck?

EvilHero90 (approved):
Looks fine to me

Issue #4621, opened 2024-01-13
World Light


Sajgon:
DEFAULT_WORLD_LIGHT is missing from configKeys enums
`registerEnumIn("configKeys", ConfigManager::DEFAULT_WORLD_LIGHT)`

Also in game.h there are unused variables
- checkLight
- EVENT_LIGHTINTERVAL

EPuncker:
@ramon-bernardo can you take a look at this please?

ramon-bernardo:
Fix with https://github.com/otland/forgottenserver/pull/4790

PR #4623, opened 2024-01-14
Fix world light config

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Move world light config to game core

**Issues addressed:**
https://github.com/otland/forgottenserver/issues/4621


omarcopires:
we had everything centralized in config.lua, now we have several scripts with several different configurations, this is a great return.

Sajgon:
Why is the config of defaultWorldLight even a thing? there's nothing to handle it being false

Sajgon:
I think there's another issue with world light

Here it's returned as "worldLightLevel, worldLightColor"
```
	function Game.getWorldLight()
		return worldLightLevel, worldLightColor
	end
```

but on other places it's fetched as:
`local worldLightColor, worldLightLevel = Game.getWorldLight()`

data/lib/core/game.lua
```diff
@@ -189,6 +189,12 @@ function Game.getUnpromotedVocations()
 end
 
 do
+	-- World Light
+	-- NOTE: if defaultWorldLight is set to true the world light algorithm will
+	-- be handled in the sources. set it to false to avoid conflicts if you wish
+	-- to make use of the function Game.setWorldLight(color, level)
```
amatria, line 195: Is this comment still true? Is it handled in the sources or is it handled in Lua? `data/scripts/globalevents/world_light.lua` seems to be doing just that.

amatria (approved):
LGTM

amatria:
> we had everything centralized in config.lua, now we have several scripts with several different configurations, this is a great return.

You are absolutely right. It is a pity this parameter is now decentralized.

MillhioreBT:
If it will no longer be a configuration then make the variable local, since it will only be used in that specific place, if someone changes it to false they know they have to create their own lighting system and they do not have to check if it is false or true elsewhere

Honestly, it doesn't bother me that this boolean is in the Lua configuration, it can be accessed from anywhere and it also doesn't litter the global environment with global variables.

ramon-bernardo:
Similar https://github.com/otland/forgottenserver/pull/4790

PR #4627, opened 2024-02-14
Add an option to disable CMake Unity builds

amatria:
### Pull Request Prelude

### Changes Proposed

Add a CMake option to disable Unity builds. `clangd`, for instance, does not support unity builds, so we'd rather add this convenient switch to turn them off.

**Issues addressed:** None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,4 +111,10 @@ if(NOT SKIP_GIT)
 endif()
 ### END  Git Version ###
 
+# Option to disable unity builds
+option(ENABLE_UNITY_BUILD "Enable unity build" ON)
+if(ENABLE_UNITY_BUILD)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -111,4 +111,10 @@ if(NOT SKIP_GIT)
 endif()
 ### END  Git Version ###
 
+# Option to disable unity builds
+option(DISABLE_UNITY_BUILD "Disable unity build" OFF)
```
ranisalt, line 115: Negated names are weird, what do you think about

```suggestion
option(ENABLE_UNITY_BUILD "Disable unity build" ON)
```

instead?

amatria: Done. TYSM. Negated names are super hard to reason about indeed.

EPuncker (approved):

Issue #4560, opened 2023-11-06
Encoding Issue (VIP)


Diego-OT:
- [*] This is a bug in the software that resides in this repository, and not a
      support matter (use https://otland.net/forums/support.16/ for support)
- [*] This issue is reproducible without changes to the C++ code in this repository

### Steps to reproduce (include any configuration/script required to reproduce)
1. Add a player in VIP
2. Edit VIP
3. Write something like: filme de ação

### Expected behaviour
normal save

### Actual behaviour
[Error - executeQuery]

The code below seems to fix the error with saving correct in db, but not display correct after relogin (need to reverse the conversion):

// cip client use Latin-1 encoding (ISO-8859-1)
std::string iso_8859_1_to_utf8(std::string &str)
{
    std::string strOut;
    for (std::string::iterator it = str.begin(); it != str.end(); ++it)
    {
        uint8_t ch = *it;
        if (ch < 0x80) {
            strOut.push_back(ch);
        } else {
            strOut.push_back(0xc0 | ch >> 6);
            strOut.push_back(0x80 | (ch & 0x3f));
        }
    }
    return strOut;
}

I hope there is some better solution, thanks.

amatria:
I spent some time evaluating this issue, and I don't think it should be marked as a "good first issue." I believe the underlying problem is profound and requires careful evaluation.

The issue at hand is that since #2403 (which addressed issue #2175), all text in the database has been stored using UTF-8 encoding. However, the data that comes from the client is Latin-1 encoded. This isn't problematic as long as the characters from the client stay within the ASCII range, as both Latin-1 and UTF-8 encode these characters the same way. However, if we receive from the client a string of text containing characters beyond the ASCII range (e.g., "ñññççç"), this is where data loss occurs, resulting in the situation described by the original poster. We end up storing in the database a set of characters that are not UTF-8 encoded.

One possible solution is to use the `CONVERT()` function in SQL to convert text from Latin-1 to UTF-8 before storing it in the database:

```sql
INSERT INTO table (utf8_column) VALUES (CONVERT('latin1_text_ñññççç' USING utf8));
```

Similarly, we can use `CONVERT()` to convert the UTF-8 encoded text back to Latin-1 encoding:

```sql
SELECT CONVERT(utf8_column USING latin1) AS latin1_column FROM table;
```

[...]

PR #4628, opened 2024-02-14
Encode Latin-1 to UTF-8 and vice versa

amatria:
### Pull Request Prelude

### Changes Proposed

The issue at hand is that since #2403 (which addressed issue #2175), all text in the database has been stored using UTF-8 encoding. However, the data that comes from the client is Latin-1 encoded. This isn't problematic as long as the characters from the client stay within the ASCII range, as both Latin-1 and UTF-8 encode these characters the same way. However, if we receive from the client a string of text containing characters beyond the ASCII range (e.g., "ñññççç"), this is where data loss occurs, resulting in the situation described by the original poster in #4560. We end up storing in the database a set of characters that are not UTF-8 encoded.

This commit addresses this issue by implementing explicit conversion mechanisms to handle text encoding conversions seamlessly during data transmission between the client and server.

The solution at hand may incur too much overhead and should be evaluated.

### How to reproduce

1. From server to client (character name is not displayed correctly):

https://github.com/otland/forgottenserver/assets/34030065/1432cb35-dda6-4bb7-97d9-6d29ea8dc8c0

2. From client to server (can't login because the password hash is Latin-1 encoded):

https://github.com/otland/forgottenserver/assets/34030065/0587773c-de43-4f8b-a363-629b364b1ca3

**Issues addressed:** #4560

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -54,7 +54,7 @@ jobs:
         run: >
           sudo apt update -q &&
           sudo apt install -yq build-essential libluajit-5.1-dev libmysqlclient-dev
-          libboost-system-dev libboost-iostreams-dev
+          libboost-system-dev libboost-iostreams-dev libboost-locale-dev
           libpugixml-dev libcrypto++-dev libfmt-dev
 
       - name: Build with cmake
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
         run: >
           sudo apt update -q &&
           sudo apt install -yq git build-essential libluajit-5.1-dev libmysqlclient-dev
-          libboost-system-dev libboost-iostreams-dev libboost-test-dev
+          libboost-system-dev libboost-iostreams-dev libboost-locale-dev libboost-test-dev
           libpugixml-dev libcrypto++-dev libfmt-dev
 
       - name: Build with CMake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,10 @@ if (NOT FORCE_LUAJIT)
     find_package(Lua REQUIRED)
 endif ()
 
+if (APPLE)
+    find_package(Iconv REQUIRED)
+endif()
+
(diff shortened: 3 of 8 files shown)


ranisalt:
I'd rather use [Boost.Locale](https://www.boost.org/doc/libs/1_82_0/libs/locale/doc/html/index.html) to do that with less code

amatria:
> I'd rather use [Boost.Locale](https://www.boost.org/doc/libs/1_82_0/libs/locale/doc/html/index.html) to do that with less code
> 
> Edit: more specifically, [`std::string utf8_string = boost::locale::conv::to_utf<char>(latin1_string, "Latin1")`](https://www.boost.org/doc/libs/1_82_0/libs/locale/doc/html/charset_handling.html) does the trick

Done. TYSM for the suggestion

ranisalt (approved):

EPuncker (approved):

nekiro:
shouldn't this happen before query is pushed to db?

amatria:
> shouldn't this happen before query is pushed to db instead of message parse?

Quoting my comment in #4560:

> […]
>
> One possible solution is to use the `CONVERT()` function in SQL to convert text from Latin-1 to UTF-8 before storing it in the database:
> 
> ```sql
> INSERT INTO table (utf8_column) VALUES (CONVERT('latin1_text_ñññççç' USING utf8));
> ```
> 
> Similarly, we can use `CONVERT()` to convert the UTF-8 encoded text back to Latin-1 encoding:
> 
> ```sql
> SELECT CONVERT(utf8_column USING latin1) AS latin1_column FROM table;
> ```
> 
> However, I'm hesitant to consider this solution as it may introduce compatibility issues with existing code, which currently assumes that strings are encoded in UTF-8 format. Another approach could involve addressing the encoding mismatch at the game protocol level, implementing explicit conversion mechanisms to handle text encoding conversions seamlessly during data transmission between the client and server.
> 
> Thoughts?

ranisalt:
It is better to handle everything in UTF-8 in the server and convert from as soon (or to as late) as possible, since UTF-8 is the de facto standard. We don't want to handle Latin1 things everywhere.

DSpeichert (approved):

PR #4636, opened 2024-03-14
Added Loyalty System (10.74)

reyaleman:
### Pull Request Prelude

### Changes Proposed

- Implemented [LOYALTY SYSTEM](https://tibia.fandom.com/wiki/Loyalty_System)

**Issues addressed:** 
https://github.com/otland/forgottenserver/issues/3871 Version 10.74 (The Loyalty System was introduced.)


schema.sql
```diff
@@ -16,6 +17,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `name` varchar(255) NOT NULL,
   `group_id` int NOT NULL DEFAULT '1',
   `account_id` int NOT NULL DEFAULT '0',
+  `main_character` tinyint NOT NULL DEFAULT '0',
```
DSpeichert, line 20: You also need to update https://github.com/otland/forgottenserver/blob/master/schema.sql#L382

DSpeichert (changes requested):

data/migrations/36.lua
```diff
@@ -1,3 +1,7 @@
 function onUpdateDatabase()
-	return false
+	print("> Updating database to version 37 (loyalty system)")
+	db.query("ALTER TABLE `accounts` ADD COLUMN `loyalty_points` SMALLINT UNSIGNED NOT NULL DEFAULT 0 AFTER `premium_ends_at`")
+	db.query("ALTER TABLE `players` ADD COLUMN `main_character` TINYINT(1) NOT NULL DEFAULT 0 AFTER `account_id`")
```
ranisalt, line 4: I don't think this belongs in here, if anything it should be stored in the account. Even better if it uses account storage.

Another improvement would be storing premium time spans rather than a counter of premium days, so we can store _when_ a player was premium and not only _how long_

reyaleman: Sure, I can move `main_character` column to account storages, I'm just using it as a way to choose which character I'd show on loyalty highscores.

Issue #4617, opened 2024-01-03
Wrong interactions of monsters with pathfinding.


floki21uu:
![wrong interations of monsters with pathfinding](https://github.com/otland/forgottenserver/assets/96269029/857ecbd6-1a04-4a82-9dff-66d1187d89cb)
GIF:
As a result, you can kill almost every monster.
Hard to explain - pathfinding should have + 1 more sqaure for monsters visibility?

EPuncker:
if I recall correctly, that is how rl tibia works, can you check it?

floki21uu:
Sure! Just tested wolves near Thais and can not reproduce on Cipsofts. I cant make a video, windows+G and LiceCap records blackscreen instead of RL Tibia on my computer :(

Zbizu:
if I remember correctly, in rl monster takes two steps before path update

EPuncker:
> Sure! Just tested wolves near Thais and can not reproduce the bug on Cipsofts. I cant make a video, windows+G and LiceCap records blackscreen instead of RL Tibia on my computer :(

to record rl tibia now its only possible by using OBS software

btw is the speed of this vampire accurate? I know tfs pathfinding and monster behavior is broken compared to rl tibia, but this vampire seems painfully slow

Zbizu:
@EPuncker most vampires are located in areas that have tiles with 120-200 ground resistance points. The tile on this recording has 100 resistance points. Similar to Yalahar giant spiders being super fast compared to PoH ones.

floki21uu:
do u guys know which part of the code should be responsible for the solution? i've done lots of work in tfs, i might try to fix it

floki21uu:
> if I remember correctly, in rl monster takes two steps before path update

this is absolitely true and it geenrates another issue. for example, after changing floor, monsters wait that second before moving and they should move straight off after changing the floor.

PR #4637, opened 2024-03-23
Optimize pathfinding

NRH-AA:
### Pull Request Prelude

### Changes Proposed

This is a overhaul of when and how pathfinding is done. I did attempt this before but I closed the pull request due to congestion. I believe all problems from the original pull request have been fixed.

This will change the pathfinding to work as A* algorithm rather than djikstra's.
It will also add more pathfinding calls as needed fixing monster moving around slowly due to 1 second pathfinding interval.
This PR also includes separation of when creatures will update their path from creatures onThink.
Lastly it includes some fixes for bugs which were not detectable until pathfinding was able to be called more often. 

**Issues addressed:** 
Might fix: #4617


MillhioreBT:
I did a quick test but this PR doesn't work, the monsters walk randomly and move away from the target slowly until they disappear from sight, and the summons don't have any type of movement.

NRH-AA:
> I did a quick test but this PR doesn't work, the monsters walk randomly and move away from the target slowly until they disappear from sight, and the summons don't have any type of movement.

Should be fixed now. I just overlooked the method call in game.cpp.

NRH-AA:
I want to note one problem found so far. I will work on a fix soon. Can we confirm this is the behavior we want?

https://otland.net/threads/tfs-a-algorithm-d.282100/post-2748212

src/map.cpp
```diff
@@ -765,20 +863,26 @@
 
 				neighborNode->f = newf;
 				neighborNode->parent = n;
-				nodes.openNode(neighborNode);
+				nodes.push_back(neighborNode);
 			} else {
 				// Does not exist in the open/closed list, create a new node
-				neighborNode = nodes.createOpenNode(n, pos.x, pos.y, newf);
-				if (!neighborNode) {
+				AStarNode* newNode = new AStarNode;
```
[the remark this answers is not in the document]
NRH-AA: I think this is a mistake by gitbot

NRH-AA:
Alright, the code is in a good spot imo. I removed some of the functionality with checking if the path is blocked or not. Given the speed of the algorithm is far faster, actually works like A*, and I tried to mimic the patterns of the previous code. The next thing at this point I think is adding a vector or list for creatures that are attacking/following a target so whenever that target moves it updates all the paths of the creatures attacking/following it. Then we can slow down the rate paths are drawn by EVENT_CREATURE_PATH_INTERVAL because we will know we have the most current path to the target any time he moves.

With that being said here is a detailed list of where we see major improvements:
1) We are no longer assigning 512 nodes into memory without using them.
2) We are not looping through all nodes O(n) to find the best one. We are sorting them O(log n) and O(1) to select/remove the best node from the list
3) We are using half or less nodes to create paths because the old algorithm checked in a circle rather than using target position
4) Paths that are close to a straight line use D(3) nodes with D being the distance to the target instead of D(8)
5) With a fraction of the nodes being stored compared to before obviously sorting, and search happen far faster.
6) Paths can be drawn when ever needed instead of being linked to onThink.

This has been a passion project of mine for a while now. I want to thank everyone that has tested and help me improve the code by pointing out things I don't see.

NRH-AA:
Alright, I made the final edits. Here is what was added:

1) Create a list to store creatures that are following/attack so we can update their paths with onWalk.
2) Slow down forceUpdatePath to 2000ms. This is just to fix weird situations like opening a door that was blocking path.
3) Upgraded the algorithm to use Euclidean distance. Increased performance even more and made paths a little more realistic

I tested many situations and was not able to find any problems. Here is a print out of the speed of the algorithm.

https://pastebin.com/hgtgTGBf

Codinablack:
This is great! You definitely spent a lot of time working through all the problem! Thank you! @NRH-AA

Codinablack (approved):
I am still relatively new to C++, but other than a few questions I had (to help me with learning) I honestly don't notice anything that stands out to me as a reason to hold back this PR.

I ask all reviewers following me, to keep in mind this was being extensively tested by multiple people on otland for sure, who knows how many others... and to also bear in mind, this "issue" has been something that has needed resolved for almost a decade or longer.

So all in all. LGTM

src/creature.cpp
```diff
@@ -1017,19 +1009,57 @@ bool Creature::setFollowCreature(Creature* creature)
...
+// Pathfinding Events
+void Creature::updateFollowingCreaturesPath()
+{
+	if (followedByCreatures.empty()) {
+		return;
+	}
+
+	std::list<Creature*> newFollowedByList;
+	for (Creature* followedByCreature : followedByCreatures) {
+		if (followedByCreature == nullptr) {
```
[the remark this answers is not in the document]
NRH-AA: Yes, If the creature died, logs out, ect.

src/map.cpp
```diff
@@ -652,32 +652,39 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
...
 
 	AStarNode* found = nullptr;
-	while (fpp.maxSearchDist != 0 || nodes.getClosedNodes() < 100) {
-		AStarNode* n = nodes.getBestNode();
-		if (!n) {
-			if (found) {
-				break;
-			}
+	int32_t bestMatch = 0;
+	int32_t iterations = 0;
```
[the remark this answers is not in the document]
NRH-AA: It didn't use it in the original code so I just left it the same. I can change it if needed.

floki21uu:
One of the bests scripts! Thank You, man!

src/creature.cpp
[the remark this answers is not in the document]
marksamman: It shouldn't ever be nullptr in this context, in the case while the creature is dying/disappearing it might just be that isRemoved() returns true. When the object is released the pointer should no longer be in followedByCreatures, if it is then it won't be nullptr and we'll likely have a use-after-free bug.

NRH-AA:
If I could get some help with why I am getting creature is defined but never used from the compilers in the setFollowCreature and setAttackedCreature. This is a little outside my knowledge base as creature is both defined and used :D

src/map.cpp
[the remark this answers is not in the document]
ranisalt: Keep it as is, they're not beneficial in modern CPUs (it was in 16-bit to 32-bit transition)

Codinablack:
I can't see any logical error with the code, but I am thinking you may be able to fix it if you rename the parameter "creature" to "target" and then drop the attackedcreature = creature line, and change creature::setFollowCreature(this), to maybe target->addFollowedByCreature(this)

I am not totally sure this will work, but it seems the "label" can also be an error/warning when the method you are calling isn't used properly.... it also seems like with the Creature being the class, and creature being a parameter, perhaps it's too confusing? I am unsure, but I think that maybe those changes might resolve the warnings (which are turned into errors because of compiler flags)

floki21uu:
Found a bug after changing floors (oldpos.z != newpos.z). The mosters takes a double danceStep

floki21uu:
The path is calculated wrongly now:
[
wrong path
](url)

Codinablack:
> The path is calculated wrongly now: [ wrong path ](url)

I don't think that is wrong... I don't recall a snake deciding to walk at angles over and over again until it reaches someone because they are four squares away at an angle (this would take all day). I did quite a bit of kiting as I played a paladin, and I am certain that creatures never walk at angles unless they have no other option, and will also take the fastest path (depending on terrain), which will often times lead them to walk the exact way you are saying is the "wrong way"..

EvilHero90:
> The path is calculated wrongly now: [ wrong path ](url)

Monsters never move diagonal unless there is no other way to reach the target, atleast I have never seen that

src/creature.cpp
```diff
@@ -931,6 +922,7 @@ bool Creature::setAttackedCreature(Creature* creature)
 		}
 
 		attackedCreature = creature;
+		creature:addFollowedByCreature(this);
```
EvilHero90, line 925: `creature->addFollowedByCreature(this);`

src/creature.cpp
```diff
@@ -1017,19 +1009,53 @@ bool Creature::setFollowCreature(Creature* creature)
 			onWalkAborted();
 		}
 
-		hasFollowPath = false;
-		forceUpdateFollowPath = false;
 		followCreature = creature;
-		isUpdatingPath = true;
+		creature:addFollowedByCreature(this);
```
EvilHero90, line 1013: `creature->addFollowedByCreature(this);`

src/creature.cpp
EvilHero90, line 925:
```suggestion
		creature->addFollowedByCreature(this);
```

src/creature.cpp
EvilHero90, line 1013:
```suggestion
		creature->addFollowedByCreature(this);
```

Codinablack: I was correct then, the changing of : to -> is what the fix was, that is great to know! I still think a parameter name change might also be apt, but this works

floki21uu: can u also check why does monster take addictiona randomStep when hasfollowpath || followcreature, for example when changing floors?
double dancestep

NRH-AA: I am have already fixed that, but have not pushed to this repo yet. I am still working on one thing which is turning out to be a nuisance. Once I fix it I will push all the updates.

floki21uu: Also monsters in real Tibia moves instantly when they see player. In TFS they wait event interval.
Same when they die, in TFS they die one by one, in real Tibia all together in the same time. As a result you get weird looking experience points (like many "5exp" for killing rats)
image
Do you have any idea how to fix it?

NRH-AA: Its already fixed wait for the next push

NRH-AA:
> The path is calculated wrongly now: [ wrong path ](url)

This was correct. The algorithm was choosing a path that was longer than the best one which would be to move south and east to the target. Instead of moving in a straight line east then south. This is now fixed with the new algorithm. Should be gtg.

src/game.h
```diff
@@ -428,7 +428,8 @@
 	void updateCreatureWalk(uint32_t creatureId);
 	void checkCreatureAttack(uint32_t creatureId);
 	void checkCreatures(size_t index);
-	void checkLight();
+	void updateCreaturesPath(size_t index);
+	// void checkLight();
```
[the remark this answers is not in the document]
NRH-AA: Check light isn't used anymore. I commented it out. I can uncomment it before push if needed.

Codinablack (approved):
It appears you have everything covered pretty well! I say, let's do it!

src/game.h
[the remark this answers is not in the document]
Codinablack: I don't know if that's an issue or not tbh. I really don't see why it should hold this pr back, if it's unused, might as well be deleted, but it's unrelated to this pr, so I am unsure what others will say.... I would think the best option would be just to delete it entirely, and chalk it up as code cleanup of code that was in the way :)

NRH-AA:
[https://youtu.be/QjCR51DXTJ4](url)

src/game.h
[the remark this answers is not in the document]
EvilHero90: Revert it for now, should be done in another pr as this doesn't really belong to this one

EvilHero90 (approved):
Code wise this looks good now (havn't tested the changes myself)
If another maintainer approves I'll merge it

floki21uu:
Lock 2 monsters within a square of walls (bonelord and ancient scarab in my case). The monsters look for the path even if they are in the building - is it intended? It also crashes the client.

floki21uu:
Also another BUG, weird monster behaviour:
weird orcs XD

floki21uu:
Also it's possible again to do this trick:
guards

NRH-AA:
You didn't add something correctly it would seem. As for the monster not pathing around the bush that doesn't have to do with this PR and I might add something to fix it when it gets merged. You should take some time to make sure you have added the changes correctly before posting problems that do not exist on here.

NRH-AA:
> Also another BUG, weird monster behaviour: weird orcs XD [ weird orcs XD ](https://private-user-images.githubusercontent.com/96269029/322173638-ffe4f023-dcd1-4cd8-b964-8e9882227d2f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTMwMTIwOTIsIm5iZiI6MTcxMzAxMTc5MiwicGF0aCI6Ii85NjI2OTAyOS8zMjIxNzM2MzgtZmZlNGYwMjMtZGNkMS00Y2Q4LWI5NjQtOGU5ODgyMjI3ZDJmLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDEzVDEyMzYzMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTE2MzgwNDM1ZjhiNzZhMDA4NmZmZjNkYTdkNTk1NTE0YmU4ODUwNWJkNjE3YjI4MmEwMDg3NzliMGE0MTY5YmYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.WB79Rc8TONdExPKyoRLb78jqWBJA7rm2lo-uJ_ld9Ic) [ ](https://private-user-images.githubusercontent.com/96269029/322173638-ffe4f023-dcd1-4cd8-b964-8e9882227d2f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTMwMTIwOTIsIm5iZiI6MTcxMzAxMTc5MiwicGF0aCI6Ii85NjI2OTAyOS8zMjIxNzM2MzgtZmZlNGYwMjMtZGNkMS00Y2Q4LWI5NjQtOGU5ODgyMjI3ZDJmLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDEzVDEyMzYzMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTE2MzgwNDM1ZjhiNzZhMDA4NmZmZjNkYTdkNTk1NTE0YmU4ODUwNWJkNjE3YjI4MmEwMDg3NzliMGE0MTY5YmYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.WB79Rc8TONdExPKyoRLb78jqWBJA7rm2lo-uJ_ld9Ic)

Not sure how I missed adding the last change here. Good catch.



> Also it's possible again to do this trick: guards [ guards ](https://private-user-images.githubusercontent.com/96269029/322202001-86b8ed2c-2dc5-49e2-a6ea-ebd85922b51b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTMwMTIwOTIsIm5iZiI6MTcxMzAxMTc5MiwicGF0aCI6Ii85NjI2OTAyOS8zMjIyMDIwMDEtODZiOGVkMmMtMmRjNS00OWUyLWE2ZWEtZWJkODU5MjJiNTFiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDEzVDEyMzYzMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWRlYTE3Zjg0ZWVmYzUzMWE0OGRmMGViODZkZGIzMzJkMGYwOGZhM2Q0ZWFmODdhY2M0MDcyYmI0YTdjYmU1NWEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.tbrBpWeXSnUZKjaGIalOUwx1GUmu95QNi0Av1-kIQe0) [ ](https://private-user-images.githubusercontent.com/96269029/322202001-86b8ed2c-2dc5-49e2-a6ea-ebd85922b51b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTMwMTIwOTIsIm5iZiI6MTcxMzAxMTc5MiwicGF0aCI6Ii85NjI2OTAyOS8zMjIyMDIwMDEtODZiOGVkMmMtMmRjNS00OWUyLWE2ZWEtZWJkODU5MjJiNTFiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA0MTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNDEzVDEyMzYzMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWRlYTE3Zjg0ZWVmYzUzMWE0OGRmMGViODZkZGIzMzJkMGYwOGZhM2Q0ZWFmODdhY2M0MDcyYmI0YTdjYmU1NWEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.tbrBpWeXSnUZKjaGIalOUwx1GUmu95QNi0Av1-kIQe0)

This will have to be fixed in another PR as it is not really related to this one.

NRH-AA:
RFR

NRH-AA:
Alright, I believe it is ready now. I will have to look into monsters being able to get stuck around bushes in another PR. It will involve blocking pathfinding after a certain amount of failed attempts and allowing the creature to just finish the path before updating pathfinding again. With all the things that are fixed I believe that one trade off is worth it until I fix it.

src/creature.h
```diff
@@ -53,6 +53,7 @@ struct FindPathParams
 
 static constexpr int32_t EVENT_CREATURECOUNT = 10;
 static constexpr int32_t EVENT_CREATURE_THINK_INTERVAL = 1000;
+static constexpr int32_t EVENT_CREATURE_PATH_INTERVAL = 100;
```
Codinablack, line 56: IMO this should be configurable by the end user. I would suggest making it a config for config.lua, and using that variable where you have "200" hardcoded. I will also mark that location with this review.

src/creature.cpp
```diff
@@ -139,27 +139,27 @@ void Creature::onThink(uint32_t interval)
...
+{
+	if (!attackedCreature && !followCreature) {
+		return;
+	}
+
+	if (lastPathUpdate - OTSYS_TIME() > 0) {
+		return;
+	}
+
+	lastPathUpdate = OTSYS_TIME() + 200;
```
Codinablack, line 159: Here I would replace the "200" with EVENT_CREATURE_PATH_INTERVAL instead, or even possibly give it its own variable that would also be configurable, but I think one variable for both things should suffice.

src/creature.cpp
```diff
@@ -947,7 +949,7 @@ void Creature::getPathSearchParams(const Creature*, FindPathParams& fpp) const
 {
 	fpp.fullPathSearch = !hasFollowPath;
 	fpp.clearSight = true;
-	fpp.maxSearchDist = 12;
+	fpp.maxSearchDist = Map::maxViewportX + Map::maxViewportY;
```
Codinablack, line 952: I like this! It is good and doesn't need changed, but I think IMO it would be better if it was +1 over the viewport max, as this would allow monsters to start being aggro'd just one square off of screen, which is how RL cibia is, but it's not really that important. Others may also prefer it to only be the size of the viewport.

src/creature.cpp
```diff
@@ -1017,19 +1019,38 @@ bool Creature::setFollowCreature(Creature* creature)
...
 	}
 
 	onFollowCreature(creature);
 	return true;
 }
 
+// Pathfinding Functions
+void Creature::addFollowedByCreature(Creature* creature)
+{
+	if (creature) {
```
Codinablack, line 1037: I am not certain, but is it not possible to end up with a nullptr here? Maybe we can check` if (creature != nullptr)` or perhaps  `if (creature->isCreature())` , just to play it as safe as possible?

src/map.h
```diff
@@ -17,36 +17,32 @@ struct FindPathParams;
...
 	AStarNode* getBestNode();
-	void closeNode(AStarNode* node);
-	void openNode(AStarNode* node);
-	int_fast32_t getClosedNodes() const;
-	AStarNode* getNodeByPosition(uint32_t x, uint32_t y);
+	AStarNode* getNodeByPosition(uint16_t x, uint16_t y) { return nodeMap[x][y]; };
 
-	static int_fast32_t getMapWalkCost(AStarNode* node, const Position& neighborPos);
-	static int_fast32_t getTileWalkCost(const Creature& creature, const Tile* tile);
+	static double calculateHeuristic(const Position& p1, const Position& p2);
```
Codinablack, line 39: I would prefer a bit more descriptive name for this function, but it's not really a deal breaker tbh.

Codinablack (changes requested):
Just a few minor touch ups, after which I will adapt the commit to my server and do some testing myself before giving one more final approval, as it does seem to be more things that need addressed discovered daily on this one. 

All in all. Good *bleeping* work!

src/creature.h
NRH-AA: I was thinking the same thing but am trying to make sure I didn't add any extras that would prevent the PR from being able to be implemented.

NRH-AA:
> Just a few minor touch ups, after which I will adapt the commit to my server and do some testing myself before giving one more final approval, as it does seem to be more things that need addressed discovered daily on this one.
> 
> All in all. Good _bleeping_ work!

Its true that new things are discovered but the main problem is that the whole system that has to do with pathfinding needs to be recreated. I am trying to get everything working with the smallest amount of edits so that the touch ups can be done in later PR's. Then again maybe I should just revamp the whole system completely.

In the end the idea is to be able to remove things like getDistanceStep and utilize hasFollowPath, ect correctly. The code from the old system is making this kind of like putting two puzzle pieces together that do not fit.

Maybe if we can decide if the current state of the code is adequate I can focus on modifying all of the other stuff that can be changed in this PR before it gets pushed.

src/creature.cpp
NRH-AA: We don't really want to have them linked so having a secondary variable would be better. The idea here is just to stop paths from updating too fast, which having a 200ms delay is a pretty good amount of time that doesn't make the monsters seem like they are slow.

src/creature.cpp
NRH-AA: I was trying to decide here because in RL it is possible for monsters to path further than maxViewportX + 1. For instance if you are on the other side of a wall and right next to the monster on the other side. I am actually working on getting a character leveled up on RL to see exactly how they handle this exact situation. I do know monsters can path further, but if they do go to far away while trying to path to you they will stop pathing to you. Which is the effect this change has created.

src/creature.cpp
NRH-AA: I am pretty sure this is safe to remove the creature check all together. I will do so in the next commit.

src/map.h
NRH-AA: Maybe we can make up a name together. If you learn a bit about pathfinding algorithms that value is actually called the heuristic so I figured naming things as they are seen when learning about pathfinding would be the best way.

NRH-AA:
Okay, with the last commit I made it so we can allow paths to be drawn farther but also lock monsters to follow the viewport + 1sqm system.

src/creature.cpp
floki21uu: solution:
	fpp.maxSearchDist = (Map::maxViewportX + Map::maxViewportY) / 2;

floki21uu:
```
// Allow paths to be drawn a little further like rl tibia
fpp.maxSearchDist = (Map::maxViewportX * Map::maxViewportY) / 2;
```

src/map.h
floki21uu: chatGPT:
calculateDistance
computeDistance
estimateDistance
evaluateDistance
findDistance
computeHypotenuse
estimateHypotenuse
computeEuclideanDistance
evaluateEuclideanDistance
computeTriangleHypotenuse
estimateTriangleHypotenuse
calculateTriangleDistance
estimateTriangleDistance
computePositionDistance
estimatePositionDistance

NRH-AA: I think the name I chose beats these lol

src/creature.cpp
NRH-AA: This would evaluate to like 60 something sqm which is a bit too much :D. Right now its 22 which I think it pretty adequate. Keep in mind most of the time if the distance ever reaches past like 10 sqm's it will stop as soon as the monster leaves the screen so it really isn't that big of a deal. I really think X+Y is the most logical and allows for people to change their viewports with minimal change to the range. X + 1 is a little to low based on paths I have seen cibia calculate but again once the monster leaves the screen + 2sqm it just stops trying to path.

floki21uu: Sorry. It should have been divided by 4 and the reason why is (166/8)+1 = 22

floki21uu:
I've tested the code in multiple ways, it works very well.

floki21uu:
> Okay, with the last commit I made it so we can allow paths to be drawn farther but also lock m now.

floki21uu:
> Okay, with the last commit I made it so we can allow paths to be drawn farther but also lock monsters to follow the viewport + 1sqm system.


I've tested the code in multiple ways, it works very well.

src/map.h
Codinablack: findShortcut()
shortestPath()
fastRoute()
quickLine()

These aren't that great either, but not bad.

Codinablack:
I read on otland like three minutes ago ralke said this thing has a major memory leak somewhere?

NRH-AA:
I am on it. Its weird because its almost like adding the config variable is when it started.

NRH-AA:
Memory leak fixed

src/map.h
```diff
@@ -17,36 +17,32 @@ struct FindPathParams;
...
+	static double getTileWalkCost(const Creature& creature, const Tile* tile);
 
 private:
-	AStarNode nodes[MAX_NODES];
-	bool openNodes[MAX_NODES];
-	std::unordered_map<uint32_t, AStarNode*> nodeTable;
-	size_t curNode;
-	int_fast32_t closedNodes;
+	std::vector<AStarNode*> nodes;
+	std::map<uint16_t, std::map<uint16_t, AStarNode*>> nodeMap;
```
EvilHero90, line 45: You never call `clear()` on this `map` either, this could be a potential memory leak, unless it's intended to keep this data static

NRH-AA: Is it needed? Wont the garbage collector handle that?

src/map.h
NRH-AA: Maybe something like distanceToEnd() or getStraightLineToEnd()

src/map.h
EvilHero90: There is no garbage collector in c++
You could convert those `AStarNode*` into `unique_ptr` this way they would free themselves once they run out of lifetime (reference count) but raw pointers only get deleted when you explicit delete them
```c++
AStarNode* newNode = new AStarNode;
delete newNode // frees the pointer
```

Codinablack (approved):
LGTM!

floki21uu:
BUG: after changing the floor for the first time, the monsters have a slow walk reaction.
If the monsters had seen you before, they will have the correct walk reaction.
guards

src/map.h
ranisalt: This is usually just called `h` in most papers and implementations anyway. `estimateCost` is fine, it's not supposed to be too descriptive because we can change the function at our discretion

src/map.cpp
```diff
@@ -824,89 +798,62 @@ bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirL
...
 
-void AStarNodes::openNode(AStarNode* node)
-{
-	size_t index = node - nodes;
-	assert(index < MAX_NODES);
-	if (!openNodes[index]) {
-		openNodes[index] = true;
-		--closedNodes;
-	}
+	return std::sqrt((dx * dx) + (dy * dy));
```
ranisalt, line 847: Just use [`std::hypot`](https://en.cppreference.com/w/cpp/numeric/math/hypot)

NRH-AA:
The new constexpr update seems to have destroyed this lol

NRH-AA:
Can someone test the last commit. I think this should fix the pathfinding again. I also tried to fix the reaction speed when a monster first targets a player that has changed floors.

NRH-AA:
std::sqrt is running almost 3x as fast as std::hypot

ranisalt:
> edit: I think changing int_fast32_t, ect is the problem.

FP math is orders of magnitude slower than integer math. I don't see why you changed the `get*Cost` functions to use doubles without changing how they work.



> std::sqrt is running almost 3x faster than std::hypot

How did you measure it?

NRH-AA:
> > edit: I think changing int_fast32_t, ect is the problem.
> 
> FP math is orders of magnitude slower than integer math. I don't see why you changed the `get*Cost` functions to use doubles without changing how they work.
> 
> > std::sqrt is running almost 3x faster than std::hypot
> 
> How did you measure it?

I used time measurements strictly around the sqrt operation. I checked 1sqm-100sqm paths. I also created a small project just as a frame of reference using only the sqrt and hypot so there wasn't any additional bloatware. All instances were 2-3x faster from what I saw.

floki21uu:
> The new constexpr update seems to have destroyed this
> 
> edit: I think changing int_fast32_t, ect is the problem.

destroyed what?

NRH-AA:
> > The new constexpr update seems to have destroyed this
> > edit: I think changing int_fast32_t, ect is the problem.
> 
> destroyed what?

The path finding stopped working with @ranisalt's constexpr position change. I did revert the datatype changes in this repo for pathmatching so I think it should be working properly again.

NRH-AA:
Lets get this added peeps

NRH-AA:
> When you made this using the constexpr positions you had problems. It seems there were problems with that commit that have since been fixed, have you tried using the current version of the constexpr positions?

Yes, and the fixes did solve the problem. The system works as it is so I will push updates to get it ready to merge with tfs master when people have tested it and say it can be merged as it is. I can't keep doing micro updates.

EvilHero90:
Feel free to update this, once `1.6` is released it'll be merged directly into `1.7` dev cycle

ranisalt (changes requested):

NRH-AA:
Have some time. Lets get it finished :). Thanks for the review rain I was able to add everything you requested I believe.

src/creature.cpp
```diff
@@ -1025,19 +1026,36 @@ bool Creature::setFollowCreature(Creature* creature)
...
 
+// Pathfinding Events
+void Creature::updateFollowingCreaturesPath()
+{
+	if (followedByCreatures.empty()) {
+		return;
+	}
+
+	const Position& thisPosition = getPosition();
+	for (auto follower : followedByCreatures) {
```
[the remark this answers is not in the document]
NRH-AA: yessir

src/creature.h
```diff
@@ -195,6 +196,12 @@ class Creature : virtual public Thing
 	virtual void onFollowCreature(const Creature*) {}
 	virtual void onFollowCreatureComplete(const Creature*) {}
 
+	// Pathfinding functions
+	virtual void addFollowedByCreature(Creature* creature) { followedByCreatures.push_back(creature); };
```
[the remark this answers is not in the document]
NRH-AA: This is a great point and it is not overidden anywhere. I will look more into this. A for emplace back I would urge you to read this chatgpt message about it: https://pastebin.com/iRwHSKP6

src/creature.h
```diff
@@ -384,8 +392,10 @@ class Creature : virtual public Thing
 	Creature* attackedCreature = nullptr;
 	Creature* master = nullptr;
 	Creature* followCreature = nullptr;
+	std::list<Creature*> followedByCreatures;
```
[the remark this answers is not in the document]
NRH-AA: This needs to be changed for vector probably.

src/map.cpp
```diff
@@ -657,32 +657,47 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
...
+	if (startPos.getDistanceX(targetPos) > maxDistanceX || startPos.getDistanceY(targetPos) > maxDistanceY) {
+		return false;
+	}
 
-	static int_fast32_t dirNeighbors[8][5][2] = {
-	    {{-1, 0}, {0, 1}, {1, 0}, {1, 1}, {-1, 1}},    {{-1, 0}, {0, 1}, {0, -1}, {-1, -1}, {-1, 1}},
-	    {{-1, 0}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}}, {{0, 1}, {1, 0}, {0, -1}, {1, -1}, {1, 1}},
-	    {{1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}},  {{-1, 0}, {0, -1}, {-1, -1}, {1, -1}, {-1, 1}},
-	    {{0, 1}, {1, 0}, {1, -1}, {1, 1}, {-1, 1}},    {{-1, 0}, {0, 1}, {-1, -1}, {1, 1}, {-1, 1}}};
 	static int_fast32_t allNeighbors[8][2] = {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}};
```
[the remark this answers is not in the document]
NRH-AA: I didn't notice any difference, I did play around with this stuff. The code is so optimized at this point this kind of stuff is meaningless. However, if it is better practice to do so I will include it. So if someone else can give their opinion would be great.

src/map.cpp
```diff
@@ -829,89 +806,55 @@ bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirL
...
+	AStarNode* firstNode = new AStarNode;
+	firstNode->parent = nullptr;
+	firstNode->x = x;
+	firstNode->y = y;
+	firstNode->g = 0;
+	firstNode->f = 0;
+
+	// Add node to node vector and map
+	nodes.reserve(50);
+	nodes.push_back(firstNode);
```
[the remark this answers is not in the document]
Codinablack: Upon further consideration about push_back being more suited for objects already created, there is the opportunity to construct in place here, probably more beneficial and more in line with modern C++ standards of RAII if we created the object with an initializer list, initializing the variables during construction... I think you would have to write your own constructor for the AStarNode class though, unless it happens to already have a contructor that uses an initializer list. Managing the data this way would allow the in-place construction benefits from emplace_back (including its performance increase) as well as the RAII.

This comment is **not** meant to be any kind of restrictive comment for review. I still stand by my approving review on this PR (the latest one).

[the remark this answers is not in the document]
NRH-AA: I don't see any issue using emplace_back. It does seem to be better for performance so I have added it. I also tested and it didn't seem to break anything.

src/map.cpp
```diff
@@ -686,7 +686,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		return false;
 	}
 
-	static int_fast32_t allNeighbors[8][2] = {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}};
+	static constexpr int_fast32_t allNeighbors[8][2] = {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}};
```
ranisalt: Using `std::array` is preferred over C arrays

```suggestion
	static constexpr std::array allNeighbors = {std::pair{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}};
```

src/creature.h
```diff
@@ -392,7 +392,7 @@ class Creature : virtual public Thing
 	Creature* attackedCreature = nullptr;
 	Creature* master = nullptr;
 	Creature* followCreature = nullptr;
-	std::list<Creature*> followedByCreatures;
+	std::vector<Creature*> followedByCreatures;
```
ranisalt, line 395: Since you call the iteratee `follower`, you can call this `followers` for consistency

```suggestion
	std::vector<Creature*> followers;
```

NRH-AA:
I want to change the name of this method:
updateFollowingCreaturesPath

Maybe
updateFollowers
updateFollowersPaths

Codinablack (approved):

ramon-bernardo:
I am analyzing the pathfinding code flow to better understand its operation and direction. I realized that it would be more appropriate to force the inheriting classes to implement the `goToFollowCreature` method, instead of keeping this logic in the base class. 

This not only improves code readability but also avoids responsibility inversion. The `getMonster()` method should be encapsulated within the `Monster` class, where decisions about monster behavior should be made, rather than in the base Creature class. 

My suggestion is to ensure that `goToFollowCreature` is implemented in each derived class, respecting the principle that monster-specific logic should remain inside Monster, not in Creature. This eliminates the inversion of control flow and keeps responsibilities within the correct classes.

Check [my fully review](https://github.com/ramon-bernardo/forgottenserver/commit/1167d90f96d9b26a6609547b51338b2e5fca81c5#diff-5a026da090b6664f91c5f3bc4a9d0844cff966ef6fa5785d258b3aa47b370a2cL967) of goToFollowCreature, no more:
```c++
// Creature.h
Monster* monster = getMonster();
if (monster && !monster->getMaster() && (monster->isFleeing() || fpp.maxTargetDist > 1)) { ... }
```


Note: that I have also implemented this in Npc and Player.


Overview:
```c++
// Creature.h
virtual void goToFollowCreature();
// changed to
virtual void goToFollowCreature() = 0;

// removed, moved to goToFollowCreature in Monster.h
virtual void onFollowCreatureComplete(const Creature*) {}
```

```c++
void Npc::goToFollowCreature()
{
	if (followCreature) {
		FindPathParams fpp;
		getPathSearchParams(followCreature, fpp);
		listWalkDir.clear();
		if (getPathTo(followCreature->getPosition(), listWalkDir, fpp)) {
			hasFollowPath = true;
			startAutoWalk();
		} else {
			hasFollowPath = false;
		}
	}
}

void goToFollowCreature() override;
```

```c++
void Player::goToFollowCreature()
{
	// before
	Creature::goToFollowCreature();
	// after
	if (followCreature) {
		FindPathParams fpp;
		getPathSearchParams(followCreature, fpp);

		listWalkDir.clear();
		if (getPathTo(followCreature->getPosition(), listWalkDir, fpp)) {
			hasFollowPath = true;
			startAutoWalk();
		} else {
			hasFollowPath = false;
		}
	}
}
```

If the duplicated code is bothersome, feel free to move it to a new function in `Creature`, and call it in the Monster, Player and Npc class.

Codinablack:
> I am analyzing the pathfinding code flow to understand its operation and direction. I realized that it would be more appropriate to force the inheriting classes to implement the `goToFollowCreature` method, instead of keeping this logic in the base class.
> 
> This not only improves code readability but also avoids responsibility inversion. The `getMonster()` method should be encapsulated within the `Monster` class, where decisions about monster behavior should be made, rather than in the base Creature class.
> 
> My suggestion is to ensure that `goToFollowCreature` is implemented in each derived class, respecting the principle that monster-specific logic should remain inside Monster, not in Creature. This eliminates the inversion of control flow and keeps responsibilities within the correct classes.
> 
> Check [my fully review](https://github.com/ramon-bernardo/forgottenserver/commit/1167d90f96d9b26a6609547b51338b2e5fca81c5#diff-5a026da090b6664f91c5f3bc4a9d0844cff966ef6fa5785d258b3aa47b370a2cL967) of goToFollowCreature, no more:
> 
> ```c++
> // Creature.h
> Monster* monster = getMonster();
> if (monster && !monster->getMaster() && (monster->isFleeing() || fpp.maxTargetDist > 1)) { ... }
> ```
> 
> Note: that I have also implemented this in Npc and Player.
> 
> Overview:
> 
> ```c++
> // Creature.h
> virtual void goToFollowCreature();
> // changed to
> virtual void goToFollowCreature() = 0;
> 
> // removed, moved to goToFollowCreature in Monster.h
> virtual void onFollowCreatureComplete(const Creature*) {}
> ```
> 
> ```c++
> void Npc::goToFollowCreature()
> {
> 	if (followCreature) {
> 		FindPathParams fpp;
> 		getPathSearchParams(followCreature, fpp);
> 		listWalkDir.clear();
> 		if (getPathTo(followCreature->getPosition(), listWalkDir, fpp)) {
> 			hasFollowPath = true;
> 			startAutoWalk();
> 		} else {
> 			hasFollowPath = false;
> 		}
> 	}
> }
> 
> void goToFollowCreature() override;
> ```
> 
> ```c++
> void Player::goToFollowCreature()
> {
> 	// before
> 	Creature::goToFollowCreature();
> 	// after
> 	if (followCreature) {
> 		FindPathParams fpp;
> 		getPathSearchParams(followCreature, fpp);
> 
> 		listWalkDir.clear();
> 		if (getPathTo(followCreature->getPosition(), listWalkDir, fpp)) {
> 			hasFollowPath = true;
> 			startAutoWalk();
> 		} else {
> 			hasFollowPath = false;
> 		}
> 	}
> }
> ```
> 
> If the duplicated code is bothersome, feel free to move it to a new function in `Creature`, and call it in the Monster, Player and Npc class.

I mean, this PR has been sitting here for ages, undergone MANY code reviews. I believe the issue you are raising, is one that is **already currently present in the source code.** I understand that this PR deals with that code, but can't your preferences on the control flow be set aside for the sake of merging this PR? You could easily take 10 minutes or less and make those changes to the codebase after this is merged. I mean, how long does it have to sit, how many revisions does it have to undergo before its finally merged?

ramon-bernardo:
Rebase? @NRH-AA

Codinablack:
> Rebase? @NRH-AA

You are funny. You go and do things that you know will force the need for a rebase, which just further hold back this PR, and you do it knowing full well what you are doing, and that it will make "rebasing" impossible without ALOT of conflict's being created that need resolved... and then you have the nerve to come back here and say... Rebase?

These kinds of things are the reason why people don't want to spend time contributing PR's to this project. 

Even after I made it a point that what you were talking about would just hold this PR back, YOU STILL GO AND DO IT ANYWAY! 

Don't be surprised if this never gets merged now, because you are selfish and wanted to feed your own desires.

ramon-bernardo:
Why all this anger, @Codinablack, seek help.

If I had the permission, I would have done the rebase myself... in projects with multiple people working, rebasing is normal, my friend. I ran tests on this PR and LGTM anyway.

Codinablack:
> Why all this anger, @Codinablack, seek help.
> 
> If I had the permission, I would have done the rebase myself... in projects with multiple people working, rebasing is normal, my friend. I ran tests on this PR and LGTM anyway.

Just because I typed something in all caps to **accentuate its importance**, doesn't mean I need help.
I would appreciate it if you kept the conversation and responses either about the PR or the context of the message you are responding to, and leave your feelings about my personal mental health out of it.

It might be true that a project with multiple maintainers and contributors will need PR's to be rebased often... but it is also true that when a person knowingly throws roadblocks onto someone else's PR just for the sake of "moving some logic from one class to another" **because you prefer it that way,** is discouraging. To constantly fix everything in a review over and over again, just to have someone else say.. hmm I wanna move a big chunk of that code you were working on NOW rather than after you have gotten your PR merged, is quite discouraging indeed... and to be completely fair, it was selfish of you to proceed to do it anyway after I clearly already raised the issue of this PR...

But I guess my breath and energy is wasted in trying to get you to see it this way, instead this comment and the previous one will just be taken as "hostility" apparently, since that is the way weaker minds always twist things they can't argue against.

Sajgon:
I just tested this commit, and overtime with 100 players online the memory goes to the roof, takes about 30 minutes and I can see it steadily growing to max and crashes my host.

It could be I've done something else wrong, but reverting this PR on my branch doesnt have memory issues anymore.

ranisalt:
> I just tested this commit, and overtime with 100 players online the memory goes to the roof, takes about 30 minutes and I can see it steadily growing to max and crashes my host.

Smells like memory leak. If that happens within 30 minutes I wouldn't merge, but definitely we can look into that.

Codinablack:
> > I just tested this commit, and overtime with 100 players online the memory goes to the roof, takes about 30 minutes and I can see it steadily growing to max and crashes my host.
> 
> Smells like memory leak. If that happens within 30 minutes I wouldn't merge, but definitely we can look into that.

I didn't read over all the code, I just stared from top down looking for any instances that could cause a leak. What I found was the method "addFollowers" stores raw creature pointers in a vector but I never saw those raw pointers being cleaned... also its worth mentioning they weren't used along with "incrementReferenceCount()" and "decrementReferenceCount()" so that is likely the cause of the big leak... there might be other reasons, I just did a quick glance... and maybe I missed where he was cleaning that vector, but it doesn't change the fact that the manual ref counting system in place in TFS is not being used.

NRH-AA:
> I just tested this commit, and overtime with 100 players online the memory goes to the roof, takes about 30 minutes and I can see it steadily growing to max and crashes my host.
> 
> It could be I've done something else wrong, but reverting this PR on my branch doesnt have memory issues anymore.

I did have additional code to remove anything that could cause a memory leak but was told it wasn't needed. idk. I have tested this code not with 100 players but over 500 monsters following my character and didn't see any memory problems. So idk what is going on unfortunately.



> Rebase? @NRH-AA

I have rebased it 100x. So if we can try to look through the code and find any problems before I do that it would be better.

With that being said I have figured out how to remove the need for sqrt calculation in the pathfinding system and I can revert it back to normal int's. It actually increased the speed even more which was nice to see. If I had to put a number probably around 5-10% on most paths.

NRH-AA:
> > I just tested this commit, and overtime with 100 players online the memory goes to the roof, takes about 30 minutes and I can see it steadily growing to max and crashes my host.
> 
> Smells like memory leak. If that happens within 30 minutes I wouldn't merge, but definitely we can look into that.

You are my inspiration sir

NRH-AA:
> > Rebase? @NRH-AA
> 
> You are funny. You go and do things that you know will force the need for a rebase, which just further hold back this PR, and you do it knowing full well what you are doing, and that it will make "rebasing" impossible without ALOT of conflict's being created that need resolved... and then you have the nerve to come back here and say... Rebase?
> 
> These kinds of things are the reason why people don't want to spend time contributing PR's to this project.
> 
> Even after I made it a point that what you were talking about would just hold this PR back, YOU STILL GO AND DO IT ANYWAY!
> 
> Don't be surprised if this never gets merged now

Don't worry bud we aren't giving up that easily. This code may not be stable enough for a merge. We don't want to hold back any progress or better code changes just because it would conflict. This isn't something I am going to give up on. If I have time I will always come back to it. It is the one thing I want to for sure upgrade in TFS because its such a major problem right now.

NRH-AA:
I found where the memory leak is happening and I am working on the fix now.

NRH-AA:
> > > I just tested this commit, and overtime with 100 players online the memory goes to the roof, takes about 30 minutes and I can see it steadily growing to max and crashes my host.
> > 
> > 
> > Smells like memory leak. If that happens within 30 minutes I wouldn't merge, but definitely we can look into that.
> 
> I didn't read over all the code, I just started from top down looking for any instances that could cause a leak. What I found was the method "addFollowers" stores raw creature pointers in a vector but I never saw those raw pointers being cleaned... also its worth mentioning they weren't used along with "incrementReferenceCount()" and "decrementReferenceCount()" so that is likely the cause of the big leak... there might be other reasons, I just did a quick glance... and maybe I missed where he was cleaning that vector, but it doesn't change the fact that the manual ref counting system in place in TFS is not being used.

I believe the ref counter is specifically for new creations and deletions. I am not creating new instances just pointing to already referenced ones.

Codinablack:
> I believe the ref counter is specifically for new creations and deletions. I am not creating new instances just pointing to already referenced ones.

The purpose of the ref counting is to prevent dangling pointers, and use-after-free errors. 

Anytime you are creating a pointer to an object that already exists (that currently uses the manual ref counting system), and storing it in a container (C++ container, not the TFS class), it needs to have its reference count increased, and when you remove it, it needs to have the reference count decreased.

It's been awhile since I looked at this code, but I don't know if onThink is the best solution for keeping that list tidy. Perhaps the list could be cleaned of old followers when adding new ones? Maybe there is an even better place than either of these options that is more logical.. I don't know, like I said, been awhile since I looked at this code, but at least that leak was found and addressed.. 

Are we certain that was the only leak?

NRH-AA:
> The purpose of the ref counting is to prevent dangling pointers, and use-after-free errors.
> 
> Anytime you are creating a pointer to an object that already exists (that currently uses the manual ref counting system), and storing it in a container (C++ container, not the TFS class), it needs to have its reference count increased, and when you remove it, it needs to have the reference count decreased.
> 
> It's been awhile since I looked at this code, but I don't know if onThink is the best solution for keeping that list tidy. Perhaps the list could be cleaned of old followers when adding new ones? Maybe there is an even better place than either of these options that is more logical.. I don't know, like I said, been awhile since I looked at this code, but at least that leak was found and addressed..
> 
> Are we certain that was the only leak?

Based on what I am seeing for the reference counter I am pretty sure its safe to not utilize in my code. Again, I am not 100% certain what exactly it is doing. From what I see it just increases and decreases based on whenever it is used and if it decreases to 0 the creature is deleted. So I could include it but even if the creature doesn't exist while it is still referenced by the pathfinding system it will not create a crash as those situations are handled. I guess if everyone wants me to implement it in I can but I think it would add additional logic that isn't needed at this point. Another way to look at it is pathfinding shouldn't stop a creature from being deleted when it should be.

The idea of putting it in onThink is it will only be called once per second or so instead of being called far more often than is needed. I would suggest it gets included somewhere that isn't called very often as it is an 0(n) deletion for every follower that is being removed.

NRH-AA:
I always speak too soon. I think that last change is the last place there could of been a memory leak haha <(-.-)>

NRH-AA:
Well I read somewhere that when objects are created with new they needed to also be deleted but all my attempts to do so wont work. Not sure if this is the memory leak or not.

src/map.cpp
```diff
@@ -649,32 +649,55 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
 	return tile;
 }
 
-bool Map::getPathMatching(const Creature& creature, std::vector<Direction>& dirList,
+uint16_t calculateHeuristic(const Position& p1, const Position& p2)
```
Shawak, line 652: Is it a problem that this overflows for positions more than 256 tiles away from each other?

NRH-AA: How do you figure that?

NRH-AA: The code should only overflow on positions where: differenceX + differenceY > 65535

If you are trying to run large paths on this code you will need to change:

image

Shawak: Yeah no worries, just wanted to point that out. I didn't take a closer look into this, just wanted you to be aware of this.

NRH-AA:
The last commit fixed a problem that occurs when a creature walks. It force updates all the followers paths. Without the change it would draw paths extremely fast as the creature moved which caused lag on lower end servers. I also included a delay check in the actual Map::pathfind just to make sure there were no cases it would ignore the check.

I tested this on a very low end 2 core and before the change I would get up to 2s lag from walking around in a spot where monsters couldn't get to me. Making the paths pretty large. I am not seeing increasing memory usage (outside the norm) and now that the lag is fixed I think we have a pretty stable pathfinding system.

src/game.cpp
```diff
@@ -3884,6 +3886,21 @@ void Game::checkCreatures(size_t index)
...
+	                                         [=, this]() { updateCreaturesPath((index + 1) % EVENT_CREATURECOUNT); }));
+
+	auto& checkCreatureList = checkCreatureLists[index];
+	for (Creature* creature : checkCreatureList) {
+		if (!creature->isDead()) {
+			creature->forceUpdatePath();
+		}
+	}
+
+	cleanup();
```
MillhioreBT, line 3901: This call is questionable because it is calling `cleanup` much earlier, when the chain of calls is already doing it after the actual movement of the creature on the map.

`forceUpdatePath > updateCreatureWalk > goToFollowCreature > startAutoWalk > addEventWalk > checkCreatureWalk > cleanup`

This means it is releasing resources long before the creature has moved or been filtered by some event that sends the creature to the release vector.

MillhioreBT: I'm not sure if it will cause a problem, but it seems like an unnecessary call in this context.

src/creature.h
```diff
@@ -199,6 +200,13 @@ class Creature : virtual public Thing
 	virtual void onFollowCreature(const Creature*);
 	virtual void onUnfollowCreature();
 
+	// Pathfinding functions
+	void addFollower(Creature* creature) { followers.emplace_back(creature); };
```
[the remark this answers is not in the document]
NRH-AA: I am not sure what you mean it doesn't handle duplicate creatures. Each creature should be unique memory. There shouldn't be any duplicate creatures ever. Emplace back was recommended by rainsalt here. I did initially use push_back. I think the reason is because emplace_back pushes a copy rather than creating a new object. Something along those lines.

src/creature.cpp
```diff
@@ -804,13 +819,47 @@ void Creature::onFollowCreature(const Creature*)
...
 	}
+}
 
-	hasFollowPath = false;
-	forceUpdateFollowPath = false;
-	isUpdatingPath = true;
+void Creature::onUnfollowCreature() { hasFollowPath = false; }
+
+// Pathfinding Events
+void Creature::removeFollower(Creature* creature)
```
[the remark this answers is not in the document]
NRH-AA: I am not sure what you are saying here. Are you saying all rats are a single creature? Each rat should be an instance of a creature which is what is being stored. The creature instance not the whole creature class.

src/game.cpp
NRH-AA: Maybe someone who understands what the cleanup does can determine if this is needed. I check it out but I don't want to remove it if its needed.

src/creature.h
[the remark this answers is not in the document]
MillhioreBT: Semantically, it makes more sense to use `push_back` to pass a fundamental data type like a pointer, but it's a matter of preference. I suppose the performance difference is literally negligible.

> emplace_back pushes a copy rather than creating a new object

The pointers are not objects.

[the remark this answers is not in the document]
NRH-AA: > Semantically, it makes more sense to use `push_back` to pass a fundamental data type like a pointer, but it's a matter of preference. I suppose the performance difference is literally negligible.
> 
> > emplace_back pushes a copy rather than creating a new object
> 
> The pointers are not objects.

Yeah nvm I understand what you are saying. You are 100% right. Idk if I remembered it backwards or what but push_back is the correct one to use here as we are storing pointers to already created objects. We do not need to construct anything while adding it into the vector.

src/creature.cpp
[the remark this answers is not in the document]
NRH-AA: Oh I see. Because I wasn't handling duplicates I think I ended up with my "solution" because I was having issues getting it to remove all the followers. It was requiring more passes than I would expect anyway. Alright, I am glad I could made you laugh and thank you so much for pointing this stuff out!

> > I am not sure what you are saying here. Are you saying all rats are a single creature? Each rat should be an instance of a creature which is what is being stored. The creature instance not the whole creature class.
> 
> Ahaha, your comment made me laugh a little, but no, you definitely misunderstood the message. Let me rephrase it again:
> 
> If your `addFollower` method didn't allow pushing multiple pointers of the same creature, you wouldn't need to traverse the entire vector:
> 
> ```c++
> auto it = std::find(followers.begin(), followers.end(), creature);
> if (it != followers.end()) {
> 	followers.erase(it);
> }
> ```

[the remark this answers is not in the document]
MillhioreBT: Can you review the file: `spectators.h`, There is a good example there of how to handle duplicates and how to remove them efficiently.

src/creature.h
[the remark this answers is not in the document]
MillhioreBT: Yes, exactly. We are not constructing anything; pointers are not objects. `emplace_back` is for constructing directly in the vector, while `push_back` simply pushes the value, in this case, a pointer.

If they were objects, then `push_back` would be a bad option because it would create two copies. But this is not the case here.

src/game.cpp
NRH-AA: Nope you are right again lol. It is not needed here.

MillhioreBT: Don't worry, that's what we're here for—to help each other, discuss, and propose ideas. Sometimes I make mistakes too. On the contrary, it's very kind of you to dedicate your valuable time to helping TFS.

src/creature.cpp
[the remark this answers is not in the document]
NRH-AA: > Can you review the file: `spectators.h`, There is a good example there of how to handle duplicates and how to remove them efficiently.

So I did this but added the check for duplicates in its own method.  -_O_-

NRH-AA:
@ranisalt

Coldensjo:
What are the required commits for this PR to work?

ranisalt:
I lack the understanding to approve this, the only thing I dislike is using the `int_fast` variants but if @Codinablack and @MillhioreBT give it the ✅ we should merge it

src/creature.cpp
```diff
@@ -135,27 +135,32 @@ void Creature::onThink(uint32_t interval)
...
-		if (forceUpdateFollowPath || walkUpdateTicks >= 2000) {
-			walkUpdateTicks = 0;
-			forceUpdateFollowPath = false;
-			isUpdatingPath = true;
+	const Position& position = getPosition();
+	for (auto* creature : followers) {
+		const Position& followerPosition = creature->getPosition();
+		uint16_t distance = position.getDistanceX(followerPosition) + position.getDistanceY(followerPosition);
+		if (distance >= Map::maxViewportX + Map::maxViewportY || getPosition().z != creature->getPosition().z) {
+			removeFollower(creature);
```
Codinablack, line 143: It seems this particular piece of code's only purpose is to remove followers when out of viewport range. Since doing this in onThink would cause a delay in when the actual removal takes place, as well as add un-necessary load to onThink, I believe a more event driven approach best... this particular removal logic should instead be relocated to the onMove.

src/map.cpp
```diff
@@ -649,32 +649,57 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
...
 
-	const Position startPos = pos;
+	// Don't update path. The target is too far away.
+	int32_t maxDistanceX = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxClientViewportX + 1;
+	int32_t maxDistanceY = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxClientViewportY + 1;
+	if (startPos.getDistanceX(targetPos) > maxDistanceX || startPos.getDistanceY(targetPos) > maxDistanceY) {
+		return false;
+	}
+
+	static constexpr std::array<std::pair<int, int>, 8> allNeighbors = {
```
Codinablack, line 689: Spectacular use of static constexpr

src/creature.cpp
```diff
@@ -804,13 +815,60 @@ void Creature::onFollowCreature(const Creature*)
...
+}
+
+void Creature::updateFollowersPaths()
+{
+	if (followers.empty()) {
+		return;
+	}
+
+	const Position& thisPosition = getPosition();
+	for (const auto follower : followers) {
```
Codinablack, line 854: lets pass by reference in loops like this when we can please

const auto& follower

Codinablack (changes requested):
Overall I think it looks pretty well polished at this point. Just make the two changes I requested and you have my approval.

src/creature.cpp
```diff
@@ -804,13 +815,60 @@ void Creature::onFollowCreature(const Creature*)
...
+void Creature::onUnfollowCreature() { hasFollowPath = false; }
+
+// Pathfinding Events
+bool Creature::isFollower(Creature* creature)
+{
+	auto it = std::find(followers.begin(), followers.end(), creature);
+	if (it != followers.end()) {
+		return true;
+	}
+	return false;
```
ranisalt, line 829:
```suggestion
	return it != followers.end();
```

src/map.cpp
```diff
@@ -686,14 +686,14 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		return false;
 	}
 
-	static constexpr std::array<std::pair<int, int>, 8> allNeighbors = {
+	const std::array<std::pair<int, int>, 8> allNeighbors = {
```
Codinablack, line 689: When I said spectacular use of static constexpr, I wasn't being sarcastic, that was a really great usage of it! I'm sorry if there was a miscommunication there as these reviews typically only have comments that are meant for changes, but this was too good to not say something nice about it, that's all.

NRH-AA: Something changed in the main branch and it really messed up this build lol. Idk what I need to do to fix it.

Codinablack (approved):
LGTM

NRH-AA:
> LGTM

What happened to the build lmao?

Codinablack:
> > LGTM
> 
> What happened to the build lmao?

It's a WIP. There are a few issues due to updates from cmake, github runners, vcpkg... it's a whole toolchain thing. The project still builds fine when downloading and trying locally. I have more information if you want to talk about it you can message me on discord.

NRH-AA:
COMON

src/creature.cpp
```diff
@@ -828,16 +828,16 @@ void Creature::addFollower(Creature* creature)
...
+
+	followers.erase(std::remove_if(followers.begin(), followers.end(),
+	                               [&position](Creature* creature) {
+		                               const Position& followerPosition = creature->getPosition();
+		                               uint16_t distance = position.getDistanceX(followerPosition) +
+		                                                   position.getDistanceY(followerPosition);
+		                               return distance >= Map::maxViewportX + Map::maxViewportY ||
+		                                      position.z != followerPosition.z;
+	                               }),
+	                followers.end());
```
ranisalt, line 840: Nice change!

ranisalt (approved):
Why tf Github keeps saying I requested changes if I already approved this

NRH-AA:
Just going to throw in some final touches since this hasn't been merged yet.

ranisalt (approved):

ranisalt:
Let's get this over with!

PR #4638, opened 2024-03-28
Optimize loops

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Remove redundant function calls in loops

**Issues addressed:** 

### DIFF
--- a/src/iologindata.cpp
+++ b/src/iologindata.cpp
@@ -687,7 +687,7 @@ bool IOLoginData::saveItems(const Player* player, const ItemBlockList& itemList,
 		}
 	}
 
-	for (size_t i = 0; i < containers.size(); i++) {
+	for (size_t i = 0, n = containers.size(); i < n; i++) {
 		const ContainerBlock& cb = containers[i];
 		Container* container = cb.first;
 		int32_t parentId = cb.second;
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1141,7 +1141,7 @@ void Monster::pushCreatures(Tile* tile)
 		uint32_t removeCount = 0;
 		Monster* lastPushedMonster = nullptr;
 
-		for (size_t i = 0; i < creatures->size();) {
+		for (size_t i = 0, n = creatures->size(); i < n;) {
 			Monster* monster = creatures->at(i)->getMonster();
 			if (monster && monster->isPushable()) {
 				if (monster != lastPushedMonster && Monster::pushCreature(monster)) {
--- a/src/talkaction.cpp
+++ b/src/talkaction.cpp
@@ -39,8 +39,8 @@ bool TalkActions::registerEvent(Event_ptr event, const pugi::xml_node&)
 	TalkAction_ptr talkAction{static_cast<TalkAction*>(event.release())}; // event is guaranteed to be a TalkAction
 	std::vector<std::string> words = talkAction->getWordsMap();
 
-	for (size_t i = 0; i < words.size(); i++) {
-		if (i == words.size() - 1) {
+	for (size_t i = 0, n = words.size(); i < n; i++) {
+		if (i == n - 1) {
 			talkActions.emplace(words[i], std::move(*talkAction));
 		} else {
 			talkActions.emplace(words[i], *talkAction);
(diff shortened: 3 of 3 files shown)


MillhioreBT:
I'm not very sure but I think that these special container methods are optimized by the compiler, I have not seen anywhere that it is a good practice or optimization method, but I think @ranisalt might have a quick and correct answer for this.

NRH-AA:
At the very least its just good practice and there are many loops already in the source code that follow it. I figured its better to make it more consistent. As for if it matters for modern C++ I am in the same boat as you.  It could be that it is not redefining the size every iteration but if it isn't might as well follow best practice anyway.

amatria:
> I'm not very sure but I think that these special container methods are optimized by the compiler

They probably are. This is good practice nonetheless: https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

MillhioreBT:
Honestly you should leave that as is, you're just adding unnecessary extra logic. the compiler already does this job and much better

Here I leave you an example to close this PR:
https://quick-bench.com/q/1TeRsXRYPbRehoc7AQCFj1H0snk

Even adding temporary variables to save the precalculated size is more time-consuming, it is negligible but...

<details>
<summary>Test image</summary>

image

</details>

amatria:
> Even adding temporary variables to save the precalculated size is more time-consuming, it is negligible but...

I am sorry, but your benchmark is wrong. [Here](https://quick-bench.com/q/BY0l2olq81BEhZyvmEzvuGmrwws) are the correct results:

- `-O3`:

BY0l2olq81BEhZyvmEzvuGmrwws

- No optimizations:

TF4dOUi6aS20ivStSLwkoge-tNs

The version that stores the size of the vector is never slower than calling `size()` repeatedly.

NRH-AA:
> Honestly you should leave that as is, you're just adding unnecessary extra logic. the compiler already does this job and much better
> 
> Here I leave you an example to close this PR: https://quick-bench.com/q/1TeRsXRYPbRehoc7AQCFj1H0snk
> 
> Even adding temporary variables to save the precalculated size is more time-consuming, it is negligible but...
> 
> Test image
> And if that were not enough, here is the comparison without compiler optimizations: https://quick-bench.com/q/2ZQg5xSsugJ-KwCjbdZHQCCy2gA image

Had to make a couple edits to the code you ran. I also increased the size of the arrays being checked and you can see the difference a little better. Please try with this.

`static void ContainerPrecalculateSize(benchmark::State& state) {
  std::vector<int> numbers = {};
  for (int i = 1; i < 50000; ++i) {
    numbers.push_back(i);
  }

  for (auto _ : state) {
    for (size_t i = 0, n = numbers.size(); i < n; i++) {
    }
  }
}
BENCHMARK(ContainerPrecalculateSize);

static void ContainerSize(benchmark::State& state) {
  std::vector<int> numbers = {};
  for (int i = 1; i < 50000; ++i) {
    numbers.push_back(i);
  }

  for (auto _ : state) {
    for (size_t i = 0; i < numbers.size(); i++) {
    }
  }
}
BENCHMARK(ContainerSize);`

amatria:
@NRH-AA @MillhioreBT 

This benchmarking code is erroneous, resulting in the compiler optimizing away all meaningful computations. Please enable the "Show Noop bar" option. Refer to [my comment above](https://github.com/otland/forgottenserver/pull/4638#issuecomment-2027793844) for the correct benchmark.

src/iologindata.cpp
```diff
@@ -687,7 +687,7 @@ bool IOLoginData::saveItems(const Player* player, const ItemBlockList& itemList,
 		}
 	}
 
-	for (size_t i = 0; i < containers.size(); i++) {
+	for (size_t i = 0, n = containers.size(); i < n; i++) {
```
ranisalt, line 690: Note that this is _not_ equivalent; the size of `containers` is changed (in L700) so this change breaks the code.

ranisalt:
One of these changes is not equivalent and makes the code incorrect; the rest, I think we should always write the simplest code, in that case it is keeping unchanged.

NRH-AA:
I probably should of looked a little closer on the loops to make sure the sizes weren't changing during them. I do want to challenge the idea of simplest code and tfs though haha. Anyway, ill keep that in mind. Just trying to find places to help. Thanks for feedback.

Issue #4630, opened 2024-02-17
Create an .env.example file


EPuncker:
Would be nice to have an .env.example file, so there is something just to fill out and the user doesn't need to create it himself.

_Originally posted by @slawkens in https://github.com/otland/forgottenserver/issues/4518#issuecomment-1949876919_

PR #4642, opened 2024-04-05
Add .env.example file

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Add .env.example file

**Issues addressed:** 
#4630


EPuncker (approved):

MillhioreBT:
image
you forgot the line break at the end

Codinablack:
Fix that little line break and I will also give you an approved review

EPuncker (approved):

slawkens:
For what is the .env.sample file, if TFS is not even reading the .env?

I tested, it's reading only the environment variables set by export command on linux.

So you need to export each variable with:

```bash
export MYSQL_HOST=127.0.0.1
export MYSQL_USER=root
```

etc.

So I'm wondering, this PR was merged 1 year ago, didn't anyone noticed since then, that it is not working?

Or am I wrong with my thoughts?

ranisalt:
@slawkens it's not TFS' responsibility to apply the environment file, it should be added to the environment prior to starting the server

If using docker/podman, you can use `--env-file` to load them

slawkens:
Maybe you can use library like this one or similar to load the info from .env file?

https://github.com/adeharo9/cpp-dotenv

Or do you have other solution on clean system (Ubuntu 24.04) to apply the .env? (not the docker)

ranisalt:
> Or do you have other solution on clean system (Ubuntu 24.04) to apply the .env?

You can use [dotenvx](https://dotenvx.com/)

I'm not against adding a solution based on cpp-dotenv, but for the time being it would be easier to use dotenvx since it _just works_

PR #4644, opened 2024-04-05
Fix monster not updating their look direction

### LINKED ISSUE #4339: Wrong direction of monster with hasFollowPath
The Demon in the right bottom corner should be looking to the  east. I have seen many videos from nowadays annihilators done by single players. Can't find them yet.
image

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Monster do not update their look direction if they can't attack. They should still update even if they cant.
**Issues addressed:** 
closes #4339

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -171,6 +171,8 @@ void Creature::onAttacking(uint32_t interval)
 
 	if (g_game.isSightClear(getPosition(), attackedCreature->getPosition(), true)) {
 		doAttacking(interval);
+	} else if (getMonster()) {
+		getMonster()->updateLookDirection();
 	}
 }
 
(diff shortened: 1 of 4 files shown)


NRH-AA:
I could also change this so it uses the main path matching so their isn't a duplicate but I figured its better to keep path matching clean by adding an additional once for sightline. Note that the other code for checkSightLine does not work for this problem because we need to know if there is any way to get to them not only if we can see them in a straight line.

MillhioreBT:
This comment shows a current video of tibia and it looks like the current behavior is correct.
https://github.com/otland/forgottenserver/issues/4339#issuecomment-1460899713

NRH-AA:
Actually on the main repo the bottom right demon does not turn as it is supposed to.

PR #4645, opened 2024-04-05
Allow multiple different slot type items in single moveevent

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Iterate over all id's linked in onEquip and onDequip to allow multiple slot types to be registered with the same event.
Currently it only selects the first id and assumes all other ids will use the same slot type.

**Issues addressed:** 

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -17214,9 +17214,15 @@ int LuaScriptInterface::luaMoveEventRegister(lua_State* L)
 	if (moveevent) {
 		if ((moveevent->getEventType() == MOVE_EVENT_EQUIP || moveevent->getEventType() == MOVE_EVENT_DEEQUIP) &&
 		    moveevent->getSlot() == SLOTP_WHEREEVER) {
+			std::vector<uint16_t> slots;
 			for (const auto id : moveevent->getItemIdRange()) {
 				ItemType& it = Item::items.getItemType(id);
-				moveevent->setSlot(it.slotPosition);
+				uint16_t itemSlot = it.slotPosition;
+
+				if (std::find(slots.begin(), slots.end(), itemSlot) != slots.end()) {
+					moveevent->setSlot(itemSlot);
+					slots.push_back(itemSlot);
+				}
 			}
 		}
 		if (!moveevent->isScripted()) {
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -17214,9 +17214,10 @@ int LuaScriptInterface::luaMoveEventRegister(lua_State* L)
 	if (moveevent) {
 		if ((moveevent->getEventType() == MOVE_EVENT_EQUIP || moveevent->getEventType() == MOVE_EVENT_DEEQUIP) &&
 		    moveevent->getSlot() == SLOTP_WHEREEVER) {
-			uint32_t id = moveevent->getItemIdRange().at(0);
-			ItemType& it = Item::items.getItemType(id);
-			moveevent->setSlot(it.slotPosition);
+			for (const auto id : moveevent->getItemIdRange()) {
+				ItemType& it = Item::items.getItemType(id);
+				moveevent->setSlot(it.slotPosition);
+			}
 		}
 		if (!moveevent->isScripted()) {
 			pushBoolean(L, g_moveEvents->registerLuaFunction(moveevent));


EPuncker (approved):

EvilHero90 (changes requested):
This actualy does not give you the expected outcome, you only create one moveevent and you basicly keep changing the slot and in the end it'll just pick the last set one.
Rather change:
https://github.com/otland/forgottenserver/blob/master/src/movement.cpp#L193-L219
and
https://github.com/otland/forgottenserver/blob/master/src/movement.cpp#L243-L299
to apply the slot comming from the itemid (drawn from items.xml) and set the correct slot accordingly
This would kinda make the `:slot` method redundant, since we can skip that incase we change it that way

EvilHero90:
> I think it's fine to duplicate some lines to solve the issue. Is the problem deeper than I'm thinking?

There is already an existing workaround which can be fully done in lua, which consists of looping over the ids and then dynamicaly create the moveevents in every loop cycle where we can just set the slot accordingly
example:
```lua
local equipmentData = {
    [2161] = {slot = "necklace"},
    [2170] = {slot = "necklace"},
    [2172] = {slot = "necklace"},
    [2173] = {slot = "necklace"},
    [2197] = {slot = "necklace"},
    [2198] = {slot = "necklace"},
    [2199] = {slot = "necklace"},
    [2200] = {slot = "necklace"},
    [2201] = {slot = "necklace"},
    [7887] = {slot = "necklace", level = 60},
    [7888] = {slot = "necklace", level = 60},
    [7889] = {slot = "necklace", level = 60},
    [7890] = {slot = "necklace", level = 60},
    [8266] = {slot = "necklace"},
    [10218] = {slot = "necklace", level = 80},
    [10219] = {slot = "necklace", level = 80},
    [10220] = {slot = "necklace", level = 80},
    [10221] = {slot = "necklace", level = 80},
    [11374] = {slot = "necklace"},
    [15403] = {slot = "necklace", level = 120},
    [18402] = {slot = "necklace", level = 150},
    [18407] = {slot = "necklace", level = 150},
    [21691] = {slot = "necklace", level = 150},
    [23541] = {slot = "necklace", level = 75},
    [23554] = {slot = "necklace", level = 75, vocation = {{"Sorcerer", true}, {"Master Sorcerer"}, {"Druid", true}, {"Elder Druid"}}},
    [24717] = {slot = "necklace"},
    [24790] = {slot = "necklace"},
    [24851] = {slot = "necklace", level = 60},
    [25423] = {slot = "necklace", level = 100},
    [25424] = {slot = "necklace", level = 100},
    [26182] = {slot = "necklace", level = 100},
    [26183] = {slot = "necklace", level = 100},
    [26184] = {slot = "necklace", level = 100},
    [26198] = {slot = "necklace", level = 150, vocation = {{"Paladin", true}, {"Royal Paladin"}}},
    [26199] = {slot = "necklace", level = 150, vocation = {{"Sorcerer", true}, {"Master Sorcerer"}, {"Druid", true}, {"Elder Druid"}}},
    [26200] = {slot = "necklace", level = 150, vocation = {{"Knight", true}, {"Elite Knight"}}}
}

local deEquipData = {
    2161, 2170, 2172, 2173, 2197, 2198, 2199, 2200, 2201, 7887, 7888, 7889, 7890, 8266, 10218, 10219, 10220, 10221, 11374, 15403, 18402, 18407, 21691, 23541, 23554, 24790, 24851, 25423, 25424, 26182, 26183, 26184, 26186, 26188, 26190
}

for id, data in pairs(equipmentData) do
    local equip = MoveEvent()
    equip:type("equip")
    equip:id(id)
    equip:slot(data.slot)
    
    if data.level then
        equip:level(data.level)
    end
    
    if data.vocation then
        for _, vocation in ipairs(data.vocation) do
            equip:vocation(unpack(vocation))
        end
    end
    
    equip:register()
end

for id, data in pairs(deEquipData) do
    local deEquip = MoveEvent()
    deEquip:type("deequip")
    deEquip:id(id)
    deEquip:slot("necklace")
    deEquip:register()
end
```
This obviously just lists up necklaces in this case but it could obviously be a mixed use case aswell.

I had plans to refactor this entirely in #4629 that it's always dependant on the `items.xml` slot type, that way we wouldn't have to add it manualy anyway.

NRH-AA:
Are you sure it will only grab the last one? I am currently using this and it fixed the problem for me. All items are in the same table and I do not have to specify slots nor do I have to do the lua work around you showed. It seems to work just fine. Could you please explain why it is the case you believe it only grabs the last slot?

ranisalt:
> Are you sure it will only grab the last one?

Yes, it does. When you run `setSlot`, it overrides the slot value, it does not add to the existing slots.

src/luascript.cpp
```diff
@@ -17214,9 +17214,16 @@ int LuaScriptInterface::luaMoveEventRegister(lua_State* L)
...
 		    moveevent->getSlot() == SLOTP_WHEREEVER) {
-			uint32_t id = moveevent->getItemIdRange().at(0);
-			ItemType& it = Item::items.getItemType(id);
-			moveevent->setSlot(it.slotPosition);
+			std::vector<uint16_t> slots;
+			for (const auto id : moveevent->getItemIdRange()) {
+				ItemType& it = Item::items.getItemType(id);
+				uint16_t itemSlot = it.slotPosition;
+
+				if (std::find(slots.begin(), slots.end(), itemSlot) != slots.end()) {
```
ranisalt, line 17222: This is very inefficient. There's a very limited number of slots, you can track using a `bitset` instead

EvilHero90:
I've been researching this for a while now, actually this is quite interesting, overall we wouldn't need to add a `slot` function at all we would have to just change one line and it should work directly from `items.xml` getting rid of a lot non necessary functionality which actualy doesn't even work as it was expected.
https://github.com/otland/forgottenserver/blob/a4471edb3aebd3e1eaf41d9584f59deab391a414/src/movement.cpp#L365
```c++
if (item->getSlotPosition() == slotp) {
```
This should already do the trick, if someone cares to test.

EPuncker:
> I've been researching this for a while now, actually this is quite interesting, overall we wouldn't need to add a `slot` function at all we would have to just change one line and it should work directly from `items.xml` getting rid of a lot non necessary functionality which actualy doesn't even work as it was expected.
> 
> https://github.com/otland/forgottenserver/blob/a4471edb3aebd3e1eaf41d9584f59deab391a414/src/movement.cpp#L365
> 
> ```c++
> if (item->getSlotPosition() == slotp) {
> ```
> 
> This should already do the trick, if someone cares to test.

indeed it always bothered me that we had to declare the same stuff in two different places (items.xml and movements)

NRH-AA:
I will check this out and report back

NRH-AA:
Closing this for now. Going to focus on pathfinding until that hopefully gets merged.

PR #4647, opened 2024-04-09
Fix tobool function

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
the `tobool` function was not getting any characters from the str parameter, so it would always return `false`
In this case the correct `string.sub` method is being used and the correct index is `1`

### DIFF
--- a/data/scripts/lib/xml.lua
+++ b/data/scripts/lib/xml.lua
@@ -1,7 +1,7 @@
 local boolTrue = {"1", "y", "Y", "t", "T"}
 
 function tobool(str)
-	return str and table.contains(boolTrue, str[0])
+	return str and table.contains(boolTrue, str:sub(1, 1))
 end
 
 function XMLNode.children(self)


EPuncker (approved):

EvilHero90 (changes requested):
This needs rather an entire re work, it's not like bools can only consist of `true`
Also we should do strict comparison and not by just `y` because `str` could be equal to `year` and it would still return as `true` in this case

MillhioreBT:
> This needs rather an entire re work, it's not like bools can only consist of `true` Also we should do strict comparison and not by just `y` because `str` could be equal to `year` and it would still return as `true` in this case

What do you suggest, why is this better than nothing, the current code is broken, it cannot index a string with the `[]` operator

ranisalt:
> it's not like bools can only consist of `true`

This function accepts `1`, `true`, `yes`, or any word that starts with `1yYtT`, which is consistent with other languages, i.e. [Python](https://github.com/python/cpython/blob/v3.11.2/Lib/distutils/util.py#L308-L321) and [Go](https://pkg.go.dev/strconv#ParseBool)

It doesn't accept `on` though (like Python, unlike Go) which could be an option, and doesn't reject unknown values

ranisalt (approved):
For the time being, this works and fixes an otherwise broken function. We can move the discussion to what strings to accept to a separate issue.

EvilHero90:
> > it's not like bools can only consist of `true`
> 
> This function accepts any word that starts with `1yYtT`, which is consistent with other languages, i.e. [Python](https://github.com/python/cpython/blob/v3.11.2/Lib/distutils/util.py#L308-L321) and [Go](https://pkg.go.dev/strconv#ParseBool). I don't think we need to guard against stupidity, but it has been raised in the past
> 
> It doesn't accept `on` though (like Python, unlike Go) which could be an option, and doesn't reject unknown values

Well what's the purpose of fixing something for the sake of just fixing it halfway that we open up another pr to fully fix it?
The other concern I have is the placement of this function, it's currently placed in `xml.lua` which implies for me that it is meant for `xml` bools only, if we want to make it reactive to all different kind of forms (even support `Go` and `Python`) then we should consider moving that to a more descriptive file

ranisalt:
> The other concern I have is the placement of this function, it's currently placed in `xml.lua` which implies for me that it is meant for `xml` bools only

Good point, it was a support function for loading XML but if it's used elsewhere, it should indeed be moved. I agree with all your points, I would just do one step at a time, but I would approve a major fix that checks all boxes 😉

EvilHero90:
> 
> Good point, it was a support function for loading XML but if it's used elsewhere, it should indeed be moved. I agree with all your points, I would just do one step at a time, but I would approve a major fix that checks all boxes 😉

alright sounds fine to me

EvilHero90 (approved):

PR #4650, opened 2024-04-11
onStepOut FIXED

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
Currently, there is a scenario where an `onStepOut` event moves the creature. So, when the creature is removed, the event is triggered, and immediately the `onStepOut` event will try to `move the creature`. However, when the creature is removed from the `tile`, the `onStepOut` event will call the `Map::moveCreature` method, which will try to `remove the creature` from the `tile` again and also attempt to update the `new tile`. This results in unexpected behavior and can lead to any undefined behavior, typically a crash.

This fixes the problem by ignoring the movements... perhaps more checks are needed?

### Steps to reproduce the problem:

Connect any player.
Place the character on top of anything that triggers the onStepOut event.
Test onStepOut event:
```lua
local leave = MoveEvent()

function leave.onStepOut(creature, item, pos, fromPosition)
    local player = creature:getPlayer()
    if player then player:teleportTo(player:getTown():getTemplePosition()) end
    return true
end

leave:aid(3000)
leave:register()

local talkAction = TalkAction("!test")

function talkAction.onSay(player, words, param, type)
    player:remove()
    return false
end

--talkAction:accountType(ACCOUNT_TYPE_GOD)
--talkAction:access(true)
--talkAction:separator(" ")
talkAction:register()
```
- You can use the !test command to delete or simply logout.
- Next you could find a crash server or bugged client

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -252,6 +252,11 @@ void Map::moveCreature(Creature& creature, Tile& newTile, bool forceTeleport /* 
 {
 	Tile& oldTile = *creature.getTile();
 
+	// If the tile does not have the creature it means that the creature is ready for elimination, we skip the move.
+	if (!oldTile.hasCreature(&creature)) {
+		return;
+	}
+
 	Position oldPos = oldTile.getPosition();
 	Position newPos = newTile.getPosition();
 
--- a/src/tile.cpp
+++ b/src/tile.cpp
@@ -1143,6 +1143,14 @@ void Tile::removeThing(Thing* thing, uint32_t count)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


ranisalt:
So this will trigger only if the `onStepOut` event also moves the player to a different location?

MillhioreBT:
> So this will trigger only if the `onStepOut` event also moves the player to a different location?

That is correct, although it can also happen with creatures in general, not only with players.

EvilHero90 (approved):

EPuncker (approved):

PR #4651, opened 2024-04-14
onConnectionRestoration event

ArturKnopik:
### Pull Request Prelude

### Changes Proposed
new CreatureEvent `onConnectionRestoration`
Handle switching game client without logout which is currently impossible.
example: call login CreatureEvent that will set bless icon status, after switching the client the icon will not be updated

```lua
local blessStatusLogin = CreatureEvent("blessStatusLogin")
function blessStatusLogin.onLogin(player)
	local blessCount = 0
	for b = 1, 5 do
		if player:hasBlessing(b) then
			blessCount = blessCount + 1
		end
	end

	local msg = NetworkMessage()
	msg:addByte(0x9C);
	if blessCount == 0 then
		msg:addU16(0)
		msg:addByte(1)
		msg:sendToPlayer(player)
		msg:delete()
		return true
	end

	local bits = bit.bor(4, 8, 16, 32, 64)
	if blessCount == 4 then
		msg:addU16(2)
		msg:addByte(2)
	elseif blessCount == 5 then
		msg:addU16(255)
		msg:addByte(3)
	end
	msg:sendToPlayer(player)
	msg:delete()
	return true
end
blessStatusLogin:register()

-- solution

local ce = CreatureEvent("blessStatusConnectionRestoration")
function ce.onConnectionRestoration(player)
	print("blessStatusConnectionRestoration")
	local blessCount = 0
	for b = 1, 5 do
		if player:hasBlessing(b) then
			blessCount = blessCount + 1
		end
	end

	local msg = NetworkMessage()
	msg:addByte(0x9C);
	if blessCount == 0 then
		msg:addU16(0)
		msg:addByte(1)
		msg:sendToPlayer(player)
		msg:delete()
		return true
	end

	local bits = bit.bor(4, 8, 16, 32, 64)
	if blessCount == 4 then
		msg:addU16(2)
		msg:addByte(2)
	elseif blessCount == 5 then
		msg:addU16(255)
		msg:addByte(3)
	end
	msg:sendToPlayer(player)
	msg:delete()
	return true
end
ce:register()
```

### DIFF
--- a/src/creatureevent.cpp
+++ b/src/creatureevent.cpp
@@ -111,6 +111,18 @@ bool CreatureEvents::playerLogin(Player* player) const
 	return true;
 }
 
+bool CreatureEvents::playerConnectionRestoration(Player* player) const
+{
+	for (const auto& it : creatureEvents) {
+		if (it.second.getEventType() == CREATURE_EVENT_CONNECTIONRESTORATION) {
+			if (!it.second.executeOnConnectionRestoration(player)) {
+				return false;
+			}
+		}
+	}
+	return true;
+}
+
 bool CreatureEvents::playerLogout(Player* player) const
 {
 	// fire global event if is registered
@@ -164,6 +176,8 @@ bool CreatureEvent::configureEvent(const pugi::xml_node& node)
 		type = CREATURE_EVENT_LOGIN;
 	} else if (tmpStr == "logout") {
 		type = CREATURE_EVENT_LOGOUT;
+	} else if (tmpStr == "connectionrestoration") {
+		type = CREATURE_EVENT_CONNECTIONRESTORATION;
 	} else if (tmpStr == "think") {
 		type = CREATURE_EVENT_THINK;
 	} else if (tmpStr == "preparedeath") {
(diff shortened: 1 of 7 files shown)


src/protocolgame.cpp
```diff
@@ -248,6 +248,10 @@ void ProtocolGame::login(uint32_t characterId, uint32_t accountId, OperatingSyst
 			eventConnect = g_scheduler.addEvent(
 			    createSchedulerTask(1000, [=, thisPtr = getThis(), playerID = foundPlayer->getID()]() {
 				    thisPtr->connect(playerID, operatingSystem);
+				    Player* player = g_game.getPlayerByID(playerID);
```
ArturKnopik, line 251: probably to remove and change to use `if (thisPtr->player)` instead

src/creatureevent.cpp
```diff
@@ -275,6 +292,25 @@ bool CreatureEvent::executeOnLogin(Player* player) const
...
+
+	ScriptEnvironment* env = scriptInterface->getScriptEnv();
+	env->setScriptId(scriptId, scriptInterface);
+
+	lua_State* L = scriptInterface->getLuaState();
+
+	scriptInterface->pushFunction(scriptId);
+	LuaScriptInterface::pushUserdata(L, player);
+	LuaScriptInterface::setMetatable(L, -1, "Player");
+	return scriptInterface->callFunction(1);
```
ramon-bernardo, line 311: Use tfs::lua::pushUserdata and tfs::lua::setMetatable.

src/protocolgame.cpp
ramon-bernardo: It seems correct to me; the player may no longer exist after 1 second. The same valid. is present in `connect`.

ranisalt (commented):
Should be called `onReconnect`, it's in a good shape

PR #4661, opened 2024-04-26
Add random session tokens

ranisalt:
### Pull Request Prelude

### Changes Proposed

Currently we use a fixed session token that is insecure if sniffed, since it contains plaintext username and password, and vulnerable to replay attacks, although extremely unlikely.

Changing this to a random session token, without any account information, is both safer and integrates better with external tools, while also (slightly) improving game login performance, since there's no verification needed - 16 random bytes is enough that it can't be guessed.

This is not ready yet, I can't test it since I'm not on my PC and there are two disconnection messages that need fixing.

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -440,7 +440,6 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
 	}
 
 	if (ip != boost::asio::ip::make_address(result->getString("session_ip"))) {
-		// TODO: get the correct message
 		disconnectClient("Your game session is already locked to a different IP. Please log in again.");
 		return;
 	}
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -388,9 +388,8 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
 
 	msg.skipBytes(1); // Gamemaster flag
 
-	// acc name or email, password, token, timestamp divided by 30
-	auto sessionArgs = explodeString(msg.getString(), "\n", 4);
-	if (sessionArgs.size() < 2) {
+	auto sessionToken = msg.getString();
+	if (sessionToken.empty()) {
 		disconnectClient("Malformed session key.");
 		return;
 	}
@@ -398,38 +397,6 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
 	if (operatingSystem == CLIENTOS_QT_LINUX) {
 		msg.getString(); // OS name (?)
 		msg.getString(); // OS version (?)
-	}
-
-	auto accountName = sessionArgs[0];
-	auto password = sessionArgs[1];
-	if (accountName.empty()) {
-		disconnectClient("You must enter your account name.");
-		return;
-	}
-
-	std::string_view token;
-	uint32_t tokenTime = 0;
-
-	// two-factor auth
-	if (getBoolean(ConfigManager::TWO_FACTOR_AUTH)) {
-		if (sessionArgs.size() < 4) {
-			disconnectClient("Authentication failed. Incomplete session key.");
-			return;
-		}
-
-		token = sessionArgs[2];
-
-		try {
-			tokenTime = std::stoul(sessionArgs[3].data());
-		} catch (const std::invalid_argument&) {
-			disconnectClient("Malformed token packet.");
-			return;
-		} catch (const std::out_of_range&) {
-			disconnectClient("Token time is too long.");
-			return;
-		}
-	} else {
-		tokenTime = std::floor(challengeTimestamp / 30);
 	}
 
 	auto characterName = msg.getString();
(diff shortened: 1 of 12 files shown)


src/protocolgame.cpp
```diff
@@ -451,21 +418,36 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
...
+	uint32_t accountId = result->getNumber<uint32_t>("account_id");
+	if (accountId == 0) {
 		disconnectClient("Account name or password is not correct.");
 		return;
 	}
 
-	g_dispatcher.addTask(
-	    [=, thisPtr = getThis()]() { thisPtr->login(authIds.second, authIds.first, operatingSystem); });
+	if (ip != boost::asio::ip::make_address(result->getString("session_ip"))) {
+		// TODO: get the correct message
```
reyaleman, line 443: IIRC, the message is `Your game session is already locked to a different IP. Please log in again.`

ranisalt: Thanks! I usually trigger it with VPN server change but I'm not in my home PC 🤣

Zbizu:
will is be compatible with both legacy character list and weblogin? (both can be tested with otc)

ranisalt:
> will this be compatible with both legacy character list and weblogin? (both can be tested with otc)

I won't touch the current protocol login (it's time to let go) but yes it should be with no worries

ranisalt:
> what's the status on this one?

Still missing the login protocol part

EvilHero90:
I can understand from a logging pov that you would want to save those session tokens into database.
Wouldn't it make more sense tho instead of fetching them from database on each authentication to rather have them stored in a `std::unordered_list<ip, token>` and just check from there if it's valid? and if it's expired just remove it out of the list.
They're not really meant to be persistent data since they expire at some point, that's why I'm making that suggestion.

If you save them for the purpose that if the server would suffer from a crash, that they could just login back again without re authenticating then the same could be achieved by fetching all the valid tokens from database into the list again at server startup.

We could get rid of database transactions from authentication then, which I kinda consider the purpose of a token system.

ranisalt:
They are persisted to work with external authentication (i.e. separate login server). If we don't persist, then we just block all login servers (while ours doesn't exist)

EvilHero90:
> They are persisted to work with external authentication (i.e. separate login server). If we don't persist, then we forbid all login servers (while ours doesn't exist)

makes sense, completly forgot that a web service or any other software keeps track of them aswell
anything else you want to change or is it up for review?

ranisalt:
It's ready but I can't really test since there are no working clients for Linux :shrug:

src/protocollogin.cpp
```diff
@@ -53,7 +53,7 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
 	}
 
 	// Generate and add session key
-	static std::independent_bits_engine<std::default_random_engine, CHAR_BIT, unsigned char> rbe;
+	static std::independent_bits_engine<std::default_random_engine, CHAR_BIT, unsigned short> rbe;
```
ranisalt, line 56: Note: the standard says "the effect is undefined unless the parameter is cv-unqualified and is one of `unsigned short`, `unsigned int`, `unsigned long`, or `unsigned long long`."

GCC and Clang accept `unsigned char`, but MSVC refuses to compile with an error message.

src/protocolgame.cpp
```diff
@@ -450,21 +417,35 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
...
 
-	// TODO: use structured binding when C++20 is adopted
-	auto authIds = IOLoginData::gameworldAuthentication(accountName, password, characterName, token, tokenTime);
-	if (authIds.first == 0) {
+	Database& db = Database::getInstance();
+	auto result = db.storeQuery(fmt::format(
+	    "SELECT `a`.`id` AS `account_id`, INET6_NTOA(`s`.`ip`) AS `session_ip`, `p`.`id` AS `character_id` FROM `accounts` `a` JOIN `sessions` `s` ON `a`.`id` = `s`.`account_id` JOIN `players` `p` ON `a`.`id` = `p`.`account_id` WHERE `s`.`session_token` = FROM_BASE64({:s}) AND `s`.`expired_at` IS NULL AND `p`.`name` = {:s} AND `p`.`deletion` = 0",
+	    db.escapeString(sessionToken), db.escapeString(characterName)));
+	if (!result) {
+		disconnectClient("Account name or password is not correct.");
```
kygov, line 432: Missing return

src/protocollogin.cpp
```diff
@@ -52,9 +52,20 @@ void ProtocolLogin::getCharacterList(const std::string& accountName, const std::
...
+	// Generate and add session key
+	static std::independent_bits_engine<std::default_random_engine, CHAR_BIT, unsigned short> rbe;
+	std::array<char, 16> sessionKey;
+	std::generate(sessionKey.begin(), sessionKey.end(), std::ref(rbe));
+
 	output->addByte(0x28);
-	output->addString(accountName + "\n" + password + "\n" + token + "\n" + std::to_string(ticks));
+	output->addString({sessionKey.data(), sessionKey.size()});
+
+	if (Database& db = Database::getInstance(); !db.executeQuery(fmt::format(
```
nekiro, line 63: could be async query?

src/protocolgame.cpp
ranisalt: Fixed

ramon-bernardo (approved):

PR #4662, opened 2024-04-28
Game state changes

ramon-bernardo:
### Pull Request Prelude

### Objective
- Avoid `Game.GameState` redundancy

### Changelog
- added Game.inState(state)
- Game.getGameState() changed to Game.getState()
- Game.setGameState(state) changed to Game.setState(state)

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -1350,6 +1350,10 @@ function doSetGameState(state)
 	return Game.setGameState(state)
 end
 
+function Game.inGameState(state)
+	return Game.inState(state)
+end
+
 function Game.getGameState()
 	return Game.getState()
 end
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -1350,6 +1350,14 @@ function doSetGameState(state)
 	return Game.setGameState(state)
 end
 
+function Game.getGameState()
+	return Game.getState()
+end
+
+function Game.setGameState(state)
+	return Game.setState(state)
+end
+
 function doExecuteRaid(raidName)
 	debugPrint("Deprecated function, use Game.startEvent('" .. raidName .. "') instead.")
 	return Game.startEvent(raidName)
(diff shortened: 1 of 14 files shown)


src/game.h
```diff
@@ -419,16 +419,20 @@ class Game
 	void updateCreatureWalkthrough(const Creature* creature);
 	void updateKnownCreature(const Creature* creature);
 
-	GameState_t getGameState() const;
-	void setGameState(GameState_t newState);
-	void saveGameState();
+	template <GameState_t S>
```
MillhioreBT, line 422: there is no point in using a template in this case, we should avoid that if not necessary

MillhioreBT (changes requested):

src/game.h
MillhioreBT: just use a classic argument of type GameState_t and that's it, why add so much magic for a simple return int == int;

EPuncker (approved):

data/lib/compat/compat.lua
```diff
@@ -1350,6 +1350,18 @@ function doSetGameState(state)
 	return Game.setGameState(state)
 end
 
+function Game.inGameState(state)
+	return Game.inState(state)
```
ranisalt, line 1354:
```suggestion
	return Game.getState() == state
```

Drop that C++ thingy

src/game.h
```diff
@@ -419,16 +419,19 @@ class Game
 	void updateCreatureWalkthrough(const Creature* creature);
 	void updateKnownCreature(const Creature* creature);
 
-	GameState_t getGameState() const;
-	void setGameState(GameState_t newState);
-	void saveGameState();
+	bool inState(GameState_t state) const
+	{
+		return gameState == state;
+	}
```
ranisalt, line 425: Same, you can drop it and move the definition of `getState` to the header so it gets inlined (likely gets inlined anyway with LTO)

nekiro:
What's the benefit from this change? I fail to find one, looks purely cosmetical

ramon-bernardo:
> What's the benefit from this change? I fail to find one, looks purely cosmetical and irrelevant

Yeah, i like cosmetical code

PR #4664, opened 2024-04-29
onDeEquip properly handled at logout/death

EvilHero90:
### Pull Request Prelude

### Changes Proposed

We didn't deequip items on logout/death but we equip everytime the player login.
That way onDeEquip didn't work correctly as intended.
following example code shows what the problem is:
```lua
function onEquip(player, item, slot)
    player:setMaxHealth(player:getMaxHealth() + 100)
    return true
end

function onDeEquip(player, item, slot)
    player:setMaxHealth(player:getMaxHealth() - 100)
    return true
end
```
The above would have resulted on every logout/login that the HP would increase by 100 but not decrease on logout/death
which is now fixed

**Issues addressed:** 
none

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1250,6 +1250,13 @@ void Player::onRemoveCreature(Creature* creature, bool isLogout)
 	Creature::onRemoveCreature(creature, isLogout);
 
 	if (creature == this) {
+		for (int32_t slot = CONST_SLOT_FIRST; slot <= CONST_SLOT_LAST; ++slot) {
+			Item* item = inventory[slot];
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

nekiro (approved):

nekiro:
It honestly doesn't matter, as long as the player object is deleted anyway and none of the equip are persistent.

MillhioreBT (approved):

MillhioreBT:
Thanks for noticing, I added the notification at the beginning of the section and completely forgot about the logout/death

PR #4666, opened 2024-05-01
Use single shared lua interface for all npc

MillhioreBT:

### Changes Proposed
A while ago this PR was merged: #4483 and it fixes the memory leak problem with NPCs very well, however that PR includes the creation of a separate instance of NpcLuaScriptInterface, when in reality we can use a single one that is shared with everyone the NPCs, so this PR adds this possibility

- The `Npcs class` has been removed for an `Npcs namespace`, I decided to make this change because it is not possible to have a static shared pointer and this class does not have any singletons to reuse.

- Added a destructor to the `NpcEventsHandler` class to remove loaded event ids from the npc lua interface reference table. (inspired by #3553)

- Now when we do `/reload npc`, the libraries will also be reloaded. It was always annoying to have to restart the server to be able to test changes to the libraries when we are in development. I don't know if maybe we should move it to the `global reload type`?

- Tested with the following simple script:
```lua
local talkAction = TalkAction("!npcs")

function talkAction.onSay(player, words, param, type)
	local pos = player:getPosition()
	local npcs = {}
	for i = 1, 1000 do
		Game.createNpc("Deruno", Position(pos.x + math.random(-5, 5), pos.y + math.random(-5, 5), pos.z)):remove()
	end

	return false
end

talkAction:register()
```

and reloads...

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -424,6 +424,27 @@ int32_t LuaScriptInterface::getMetaEvent(const std::string& globalName, const st
 	return runningEventId++;
 }
 
+void LuaScriptInterface::removeEvent(int32_t scriptId)
+{
+	if (scriptId == -1) {
+		return;
+	}
+
+	// get our events table
+	lua_rawgeti(luaState, LUA_REGISTRYINDEX, eventTableRef);
+	if (!isTable(luaState, -1)) {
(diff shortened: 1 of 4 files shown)


EvilHero90:
I'll close this, since this is already included in #4671

PR #4671, opened 2024-05-07
[Revscriptsys] RevNpcSys version 1.0 (lua npcs / NpcType / new npc system)

EvilHero90:
### Pull Request Prelude

### Changes Proposed

Finaly introducing lua npcs with an oop based new npc system.
Both systems lua and xml are now based on NpcType.
This pr doesn't change anything on the already existing xml npcs and jiddos npc system, that's still valid and functional.
Fixed a memory leak regarding xml npcs.
I've included 2 example npcs in data/npc/lua/#test.lua

**Issues addressed:** 
none


EvilHero90:
This is a good base for version 1.0
Time to review and merge.

data/lib/compat/compat.lua
```diff
@@ -132,6 +132,42 @@ do
...
+			return
+		elseif key == "onPlayerEndTrade" then
+			self:eventType("endtrade")
+			self:onPlayerEndTrade(value)
+			return
+		elseif key == "onThink" then
+			self:eventType("think")
+			self:onThink(value)
+			return
+		end
```
ranisalt, line 165: How about we put these in a table rather than if/elseif chaining?

data/lib/compat/compat.lua
```diff
@@ -588,8 +624,6 @@ function isPremium(cid) local p = Player(cid) return p and p:isPremium() or fals
 
 STORAGEVALUE_EMPTY = -1
 function Player:getStorageValue(key)
-	print("[Warning - " .. debug.getinfo(2).source:match("@?(.*)") .. "] Invoking Creature:getStorageValue will return nil to indicate absence in the future. Please update your scripts accordingly.")
```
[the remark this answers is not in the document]
EvilHero90: Actually no, this is spamming for everytime we use getStorageValue somewhere (spams console)
I just forgot to move it outside the function for a just one time notice print

data/lib/compat/compat.lua
EvilHero90: I would do this in another pr and include all of the others ones in the similar style then, so we change all of them at once in one pr instead ok?

EvilHero90:
> Really cool to have Lua NPCs, it's one step further away from XML. I don't have enough knowledge to do a proper review here though.

Feel free to just review the .cpp/h changes, those are mostly the critical changes, we still have the old npc system if something shouldn't correctly work at some point I can just push fixes and increase the version

src/luascript.h
```diff
@@ -377,6 +378,12 @@ class LuaScriptInterface
 		lua_setfield(L, -2, index);
 	}
 
+	static void setField(lua_State* L, const std::string index, std::string value)
+	{
+		pushString(L, value);
+		lua_setfield(L, -2, index.c_str());
+	}
+
```
ranisalt, line 386: If you make `index` and `value` a `std::string_view` you can merge this declaration with the previous one `void setField(lua_State* L, const char* index, const std::string& value)`

src/npc.h
```diff
@@ -82,12 +75,106 @@ class NpcEventsHandler
...
+	bool loadFromXml();
+
+	std::string name;
+	std::string filename;
+	std::string scriptFilename;
+	bool fromLua = false;
+	std::string eventType;
+
+	NpcTypeInfo info;
+	std::unique_ptr<NpcEventsHandler> npcEventHandler;
```
ranisalt, line 159:
```suggestion
	std::unique_ptr<NpcEventsHandler> npcEventHandler = std::make_unique<NpcEventsHandler>();
```

Then drop the default constructor and destructor

src/npc.h
```diff
@@ -82,12 +75,106 @@ class NpcEventsHandler
...
+	void setIgnoreHeight(bool b) { info.ignoreHeight = b; }
+	bool getIgnoreHeight() const { return info.ignoreHeight; }
+	void setIsIdle(bool b) { info.isIdle = b; }
+	bool getIsIdle() const { return info.isIdle; }
+	void setPushable(bool b) { info.pushable = b; }
+	bool getPushable() const { return info.pushable; }
+	void setDefaultOutfit(Outfit_t outfit) { info.defaultOutfit = outfit; }
+	Outfit_t getDefaultOutfit() const { return info.defaultOutfit; }
+	void setParameter(std::string key, std::string value) { info.parameters[key] = value; }
+	std::map<std::string, std::string> getParameters() const { return info.parameters; }
```
ranisalt, line 144:
```suggestion
	const std::map<std::string, std::string>& getParameters() const { return info.parameters; }
```

src/npc.h
```diff
@@ -82,12 +75,106 @@ class NpcEventsHandler
...
+	void setSkull(Skulls_t s) { info.skull = s; }
+	Skulls_t getSkull() const { return info.skull; }
+	void setName(std::string n) { name = n; }
```
ranisalt, line 147:
```suggestion
	void setName(std::string_view n) { name = n; }
```

src/npc.h
```diff
@@ -82,12 +75,106 @@ class NpcEventsHandler
...
+	std::string getName() const { return name; }
```
ranisalt, line 148:
```suggestion
	const std::string& getName() const { return name; }
```

src/npc.h
```diff
@@ -82,12 +75,106 @@ class NpcEventsHandler
...
+	bool getAttackable() const { return info.attackable; }
+	void setIgnoreHeight(bool b) { info.ignoreHeight = b; }
+	bool getIgnoreHeight() const { return info.ignoreHeight; }
+	void setIsIdle(bool b) { info.isIdle = b; }
+	bool getIsIdle() const { return info.isIdle; }
+	void setPushable(bool b) { info.pushable = b; }
+	bool getPushable() const { return info.pushable; }
+	void setDefaultOutfit(Outfit_t outfit) { info.defaultOutfit = outfit; }
+	Outfit_t getDefaultOutfit() const { return info.defaultOutfit; }
+	void setParameter(std::string key, std::string value) { info.parameters[key] = value; }
```
ranisalt, line 143:
```suggestion
	void setParameter(const std::string& key, std::string_view value) { info.parameters[key] = value; }
```

(the key must be `std::string` but the value will be copied anyway)

ranisalt (changes requested):

data/lib/core/player.lua
```diff
@@ -158,6 +158,8 @@ function Player.transferMoneyTo(self, target, amount)
 	end
 
 	self:setBankBalance(self:getBankBalance() - amount)
+	-- incase server crashes that we do not duplicate money
+	db.query("UPDATE `players` SET `balance` = " .. self:getBankBalance() .. " WHERE `id` = '" .. self:getGuid() .. "'")
```
ramon-bernardo, line 162: However, if the target is online, it will be duplicated.

EvilHero90: The problem is at current stage we cannot fully avoid this unless we remove transfer to offline players or we introduce some kind of function which only saves the transfer money of the offline player once the player who send it was saved aswell, or this will always have the potential to go out of sync when the server crashes

EvilHero90: https://github.com/otland/forgottenserver/pull/4671/commits/53dbb165f5dce048fe072e4f4f8f58593de9d530
better approach but I don't like the saving player part, as it could introduce more out of sync stuff on crash

ranisalt (approved):
LGTM, let's see what @MillhioreBT and @ramon-bernardo have to say

ramon-bernardo (approved):
I don't like having to perform an update/save every time a player transfers gold. We should assume that the server will NEVER crash. However, if it's okay with this, it LGTM.

ranisalt (approved):

MillhioreBT (approved):
LGTM

data/cpplinter.lua
```diff
@@ -1,7 +1,11 @@
----@alias os.mtime fun(): number
+---@class os
+---@field mtime fun(): number
+os = {}
```
EPuncker, line 3: is this and line 8 correct?

ranisalt:
Let's go 🚀

PR #4673, opened 2024-05-07
Fix possible crash

kygov:
### Pull Request Prelude

### Changes Proposed

Fixed broken iteration.

**Issues addressed:**
Crash because removing element from map and iterating.

### DIFF
--- a/src/creatureevent.cpp
+++ b/src/creatureevent.cpp
@@ -23,9 +23,11 @@ void CreatureEvents::clear(bool fromLua)
 
 void CreatureEvents::removeInvalidEvents()
 {
-	for (auto it = creatureEvents.begin(); it != creatureEvents.end(); ++it) {
+	for (auto it = creatureEvents.begin(); it != creatureEvents.end(); ) {
 		if (it->second.getScriptId() == 0) {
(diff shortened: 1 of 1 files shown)


EvilHero90 (changes requested):
This doesn't work either as expected, I'd suggest to do it the safe way with 2 loops
```cpp
std::vector<std::string> invalid;
for (auto& event : creatureEvents) {
	if (event.second.getScriptId() == 0) {
		invalid.push_back(event.first);
	}
}
for (auto& toRemove : invalid) {
	creatureEvents.erase(toRemove);
}
invalid.clear();
```

kygov:
> This doesn't work either as expected, I'd suggest to do it the safe way with 2 loops
> 
> ```c++
> std::vector<std::string> invalid;
> for (auto& event : creatureEvents) {
> 	if (event.second.getScriptId() == 0) {
> 		invalid.push_back(event.first);
> 	}
> }
> for (auto& toRemove : invalid) {
> 	creatureEvents.erase(toRemove);
> }
> invalid.clear();
> ```
Why not? What I did works fine

EvilHero90 (approved):

nekiro:
could be std::erase_if

PR #4675, opened 2024-05-08
Move static LuaScriptInterface methods to namespace

ranisalt:
### Pull Request Prelude

### Changes Proposed

Part 1 of a major rewrite to our dearest file that hangs low-powered computers and triggers out-of-memory situations all the time.

At first, move static functions (except functions that will be bound to Lua methods) out of the `LuaScriptInterface` class into the `tfs::lua` namespace. This should help reducing the load to compile this file and enables cleaning up the header in the future.

A few unused functions were removed, and some that weren't used outside `luascript.cpp` were moved into it as static, within an anonymous namespace.


src/weapons.cpp
```diff
@@ -505,20 +505,20 @@ int32_t Weapon::getHealthCost(const Player* player) const
 bool Weapon::executeUseWeapon(Player* player, const LuaVariant& var) const
 {
 	// onUseWeapon(player, var)
-	if (!scriptInterface->reserveScriptEnv()) {
+	if (!tfs::lua::reserveScriptEnv()) {
```
ramon-bernardo, line 508: "Is there a particular reason for naming it `tfs::lua` instead of simply `lua`

ranisalt: @ramon-bernardo avoiding possible clashes with other libs. If at some point we use a dependency that defines a namespace `lua` (say a C++ wrapper for Lua) then it may clash with our functions.

This happened with OpenSSL already (it defines a global `RSA` struct). Ideally, we will wrap the entire code in `namespace tfs` so we can drop `tfs::` and it will be just `lua::`

ramon-bernardo: @ranisalt thanks for clarifying, that's what I imagined. We could do a lot like `tfs::x`, right?

I've been looking at game events, I thought something like...
```cpp
tfs::game::events::x
tfs::game::map::tows::x
```

DSpeichert (approved):

ramon-bernardo (approved):

EPuncker (approved):

Issue #4676, opened 2024-05-10
classicAttackSpeed makes server go 100% CPU


gesior:
### Steps to reproduce
1. Turn on `classicAttackSpeed` in `config.lua`
2. Login 2 players few SQM from each other.
3. Attack one player using melee weapon.
4. Watch how number of `doAttacking` (or `checkCreatureAttack`) executions per second grows over time (OTS Stats or add 'std::cout' in that function)

You can make it go high faster by running around. Each step of attacked/attacking player adds 1 extra event.

### Expected behaviour
Server checks players attacks every `getAttackSpeed()` - once, not multiple times.

### Actual behaviour
Server adds new `checkCreatureAttack` event to Scheduler every second (onThink) and every player step. Each event is recurring, if target cannot be hit by player.
Server generates infinite number of Scheduler events and CPU goes 100% after few minutes/hours.

### Environment
Tested on TFS 1.4, but master code looks the same.

### Fix
1. In `player.h` under:
```cpp
uint32_t walkTaskEvent = 0;
```
add:
```cpp
uint32_t classicAttackEvent = 0;
```

2. In `player.cpp` in function `void Player::doAttacking(uint32_t)` replace:
```cpp
g_scheduler.addEvent(task);
```
with:
```cpp
g_scheduler.stopEvent(classicAttackEvent);
classicAttackEvent = g_scheduler.addEvent(task);
```

It will limit number of `checkCreatureAttack` events to 1 per player.

PR #4677, opened 2024-05-10
Fix classicAttackSpeed makes server lag (#4676)

gesior:
`classicAttackSpeed` algorithm is bugged and may generate infinite number of scheduler (dispatcher) events.
1. TFS calls Player::doAttacking every second for every player.
2. It creates new event `g_scheduler.addEvent(task);`, which calls `Player::doAttacking` again with `getAttackSpeed()` delay. So there is repeating event every 0.2 sec (with attack speed 200) and TFS still calls `Player::doAttacking` every second creating new event.
That way server can create infinite number of events for 1 player. 1 new event every second.

Often it all stops on `if ((OTSYS_TIME() - lastAttack) >= getAttackSpeed()) {` - player cannot attack as fast as these events execute, so if there is more than 1 dispatcher event for given player, it won't re-create it after execution (`addEvent` is inside that IF).
But if player uses melee weapon and his target is not within 1 sqm, his attack will never execute and `lastAttack` will never update, so it will be able to re-create events, even if there is more than 1 running in same time.

Fixed code limits number of scheduled `doAttacking` events to 1 per player.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -3394,7 +3394,8 @@ void Player::doAttacking(uint32_t)
 		if (!classicSpeed) {
 			setNextActionTask(task, false);
 		} else {
-			g_scheduler.addEvent(task);
+			g_scheduler.stopEvent(classicAttackEvent);
+			classicAttackEvent = g_scheduler.addEvent(task);
 		}
 
 		if (result) {
--- a/src/player.h
+++ b/src/player.h
@@ -1226,6 +1226,7 @@ private:
 	uint32_t magLevel = 0;
 	uint32_t actionTaskEvent = 0;
 	uint32_t walkTaskEvent = 0;
+	uint32_t classicAttackEvent = 0;
 	uint32_t MessageBufferTicks = 0;
 	uint32_t accountNumber = 0;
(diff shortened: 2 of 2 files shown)


src/player.cpp
```diff
@@ -3394,7 +3394,8 @@ void Player::doAttacking(uint32_t)
 		if (!classicSpeed) {
 			setNextActionTask(task, false);
 		} else {
-			g_scheduler.addEvent(task);
+			g_scheduler.stopEvent(classicAttackEvent);
```
EvilHero90, line 3397: Don't you have to check if the `classicAttackEvent` is even valid before trying to stop it?

gesior: > Don't you have to check if the `classicAttackEvent` is even valid before trying to stop it?

No. If it's not set yet by script, it's set to `0` in `player.h` and `stopEvent` code checks, if it's not 0:
```cpp
void Scheduler::stopEvent(uint32_t eventId)
{
	if (eventId == 0) {
		return;
	}
...
```

nekiro (approved):

EPuncker (approved):

PR #4678, opened 2024-05-12
Replace boost::variant and tagged unions with std::variant

ranisalt:
### Pull Request Prelude

### Changes Proposed

An effort to reduce the amount of code in headers and remove a dependency, uses `std::variant` to hold values that can have varying types. This is used for item attributes.

Added a `tfs::visitors` helper class to avoid writing verbose classes when using `std::visit`


ranisalt:
Apple Clang is too handicapped for this

PR #4682, opened 2024-05-14
Use Guild shared_ptr instead of raw pointers

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Replace guild raw pointers with shared_ptr.
- Use `auto` and `const auto`, at all, when possible.
- Added `Guild::MEMBER_RANK_LEVEL_DEFAULT`.

### Discussion
Should we use `shared_ptr<T>` or `T_ptr` at all?
`GuildWars` instead of `GuildWarVector`?

### DIFF
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -76,11 +76,9 @@ bool ChatChannel::addUser(Player& player)
 
 	// TODO: Move to script when guild channels can be scripted
 	if (id == CHANNEL_GUILD) {
-		if (const auto& guild = player.getGuild()) {
-			if (!guild->getMotd().empty()) {
-				g_scheduler.addEvent(
-				    createSchedulerTask(150, [playerID = player.getID()]() { g_game.sendGuildMotd(playerID); }));
-			}
+		if (const auto& guild = player.getGuild(); !guild->getMotd().empty()) {
+			g_scheduler.addEvent(
+			    createSchedulerTask(150, [playerID = player.getID()]() { g_game.sendGuildMotd(playerID); }));
 		}
 	}
 
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -476,7 +476,7 @@ bool Chat::talkToChannel(const Player& player, SpeakClasses type, const std::str
 	}
 
 	if (channelId == CHANNEL_GUILD) {
-		const auto rank = player.getGuildRank();
+		const auto& rank = player.getGuildRank();
 		if (rank && rank->level > 1) {
 			type = TALKTYPE_CHANNEL_O;
 		} else if (type != TALKTYPE_CHANNEL_Y) {
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -76,7 +76,7 @@ bool ChatChannel::addUser(Player& player)
 
 	// TODO: Move to script when guild channels can be scripted
 	if (id == CHANNEL_GUILD) {
-		if (const auto guild = player.getGuild()) {
+		if (const auto& guild = player.getGuild()) {
 			if (!guild->getMotd().empty()) {
 				g_scheduler.addEvent(
 				    createSchedulerTask(150, [playerID = player.getID()]() { g_game.sendGuildMotd(playerID); }));
@@ -330,7 +330,7 @@ ChatChannel* Chat::createChannel(const Player& player, uint16_t channelId)
 
 	switch (channelId) {
 		case CHANNEL_GUILD: {
-			if (const auto guild = player.getGuild()) {
+			if (const auto& guild = player.getGuild()) {
 				auto ret =
 				    guildChannels.emplace(std::make_pair(guild->getId(), ChatChannel(channelId, guild->getName())));
 				return &ret.first->second;
@@ -553,7 +553,7 @@ ChatChannel* Chat::getChannel(const Player& player, uint16_t channelId)
 {
 	switch (channelId) {
 		case CHANNEL_GUILD: {
-			if (const auto guild = player.getGuild()) {
+			if (const auto& guild = player.getGuild()) {
 				auto it = guildChannels.find(guild->getId());
 				if (it != guildChannels.end()) {
 					return &it->second;
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -374,7 +374,7 @@ bool Chat::deleteChannel(const Player& player, uint16_t channelId)
 {
 	switch (channelId) {
 		case CHANNEL_GUILD: {
-			const auto guild = player.getGuild();
+			const auto& guild = player.getGuild();
 			if (!guild) {
 				return false;
 			}
--- a/src/chat.cpp
+++ b/src/chat.cpp
@@ -76,10 +76,11 @@ bool ChatChannel::addUser(Player& player)
 
 	// TODO: Move to script when guild channels can be scripted
 	if (id == CHANNEL_GUILD) {
-		Guild* guild = player.getGuild();
-		if (guild && !guild->getMotd().empty()) {
-			g_scheduler.addEvent(
-			    createSchedulerTask(150, [playerID = player.getID()]() { g_game.sendGuildMotd(playerID); }));
+		if (const auto guild = player.getGuild()) {
+			if (!guild->getMotd().empty()) {
+				g_scheduler.addEvent(
+				    createSchedulerTask(150, [playerID = player.getID()]() { g_game.sendGuildMotd(playerID); }));
+			}
 		}
 	}
 
@@ -329,8 +330,7 @@ ChatChannel* Chat::createChannel(const Player& player, uint16_t channelId)
 
 	switch (channelId) {
 		case CHANNEL_GUILD: {
-			Guild* guild = player.getGuild();
-			if (guild) {
+			if (const auto guild = player.getGuild()) {
 				auto ret =
 				    guildChannels.emplace(std::make_pair(guild->getId(), ChatChannel(channelId, guild->getName())));
 				return &ret.first->second;
@@ -376,7 +376,7 @@ bool Chat::deleteChannel(const Player& player, uint16_t channelId)
 {
 	switch (channelId) {
 		case CHANNEL_GUILD: {
-			Guild* guild = player.getGuild();
+			const auto guild = player.getGuild();
 			if (!guild) {
 				return false;
 			}
@@ -478,7 +478,7 @@ bool Chat::talkToChannel(const Player& player, SpeakClasses type, const std::str
 	}
 
 	if (channelId == CHANNEL_GUILD) {
-		GuildRank_ptr rank = player.getGuildRank();
+		const auto rank = player.getGuildRank();
 		if (rank && rank->level > 1) {
 			type = TALKTYPE_CHANNEL_O;
 		} else if (type != TALKTYPE_CHANNEL_Y) {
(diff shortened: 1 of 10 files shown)


MillhioreBT:
I definitely wouldn't remove the name `GuildRank_ptr`
It would be nice to add Guild_ptr instead of using `shared_ptr<T>()` everywhere
It's really horrible to see `shared_ptr<T>` a thousand times throughout the source code.

Codinablack:
> I definitely wouldn't remove the name `GuildRank_ptr` It would be nice to add Guild_ptr instead of using `shared_ptr<T>()` everywhere It's really horrible to see `shared_ptr<T>` a thousand times throughout the source code.

I'm the opposite on this topic. I feel that seeing std::shared_ptr<TypeName>() in the code, everywhere it's used, gives a very clear indication to whomever is reading it, what it is, without having to think about it, or take the time to get the compiler to show its underlying structure. 

I also am opposite on the use of "auto" everywhere possible though too. I feel it should only really be used in certain situations, but for the most part, one should specify the type, so that whomever is reading it, know's what it is without having to think about it. 

That's just my opinion though.

EvilHero90:
The problem is rather if you have `shared_ptr` shattered across everywhere, if you ever wanted to change it to another pointer type you'd have to change every single occurency instead of just changing the superior declaration.

I'm also not happy with the declaration of variables inside if statements, we never did that before and we shouldn't start doing it now, it kinda makes the code unreadable just for decreasing the line size by 1 each time, not worth the trade

MillhioreBT:
> The problem is rather if you have `shared_ptr` shattered across everywhere, if you ever wanted to change it to another pointer type you'd have to change every single occurency instead of just changing the superior declaration.
> 
> I'm also not happy with the declaration of variables inside if statements, we never did that before and we shouldn't start doing it now, it kinda makes the code unreadable just for decreasing the line size by 1 each time, not worth the trade

for this there is the `using` keyword which is quite similar to `typedef`
used to create alias

Although directly using std::shared_ptr<T> in some cases is useful/fast and feasible, there are others in which it is not, for example having to use this in 100 different places, for that there is a better way to do it and that is to create an alias, this way the code becomes more consistent/legibility/maintainable

and whether or not people can recognize the type being used, I'm not sure but in almost all projects when an alias starts with the `class name` and an abbreviation of `ptr`, it implies that we are dealing with a smart pointer, At least that's what I've noticed, and in fact I like it that way.

ramon-bernardo:
I prefer shared_ptr<T> over T_ptr; it's more idiomatic and friendlier to the language. I tend to avoid using aliases when concealing types. As for maintenance, who would ever need to change shared_ptr to something else? It's unlikely to happen.

The use of auto is debatable, although nowadays it's generally preferable.

Initializing variables within an if statement is already common in many parts of TFS. Not only does it restrict the variable's scope, but it also checks for nullptr right away.

MillhioreBT:
> I prefer `shared_ptr<T>` over `T_ptr`; it's more idiomatic and friendlier to the language. I tend to avoid using aliases when concealing types. As for maintenance, who would ever need to change shared_ptr to something else? It's unlikely to happen.
> 
> The use of `auto` is debatable, although nowadays it's generally preferable.
> 
> Initializing variables within an if statement is already common in many parts of TFS. Not only does it restrict the variable's scope, but it also checks for nullptr right away.

Nowadays, who uses the notes blog to work on a project? The visual studio or visual studio code linter immediately shows you the type, and saying that something is unlikely to happen as an excuse to do things one way is not a good justification.

I could well say, well, since no one will change X thing, then I will add Y things...

Also, whatever the case, in TFS the smart pointers each have their alias and this is how it should be done, respect the code style of the repository

At least it is my personal opinion, if anyone else agrees or disagrees, please leave your opinion. ty

ramon-bernardo:
I doubt anyone will test this pull request, and that's just how it goes. GitHub doesn't have a built-in linter, and this is where we review the code, looks good to me. At this stage, using `auto` is irrelevant, as you need to know what each `getT` returns. Aliasing `T_ptr` achieves the same effect as `auto`.

Regardless, I'll adhere to the established pattern.

ranisalt:
I'm 100% with @ramon-bernardo here. "if we ever have to change" OK but why would we ever? We probably have bigger fish to fry if we need to rename a class. `Guild` has that name for over 11 years now

About `auto`, everyone should probably always use it, as it helps avoiding lots of pitfalls related to conversion and casting, and I'll [let Herb Sutter explain](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/):

> **What does “write code against interfaces, not implementations” mean, and why is it generally beneficial?**
>
> It means we should care principally about “what,” not “how.” This separation of concerns applies at all levels in high-quality modern software—hiding code, hiding data, and hiding type. Each increases encapsulation and reduces coupling, which are essential for large-scale and robust software.
> _(...)_
> **Hiding type (run-time polymorphism).** Second, OO also gave us “separation of interfaces to hide type.” A base class or interface can delegate work to a concrete derived implementation via virtual functions. Now the interface the caller sees and the implementation are actually different types, and the caller knows the base type only—he doesn’t know or care about the concrete type, including even its size. The point, once again, is that the caller does not, and should not, commit to a single concrete type, which would make the caller’s code less general and less able to be reused with new types.

Also read answer 3 if it piques your interest.

src/chat.cpp
```diff
@@ -76,10 +76,11 @@ bool ChatChannel::addUser(Player& player)
 
 	// TODO: Move to script when guild channels can be scripted
 	if (id == CHANNEL_GUILD) {
-		Guild* guild = player.getGuild();
-		if (guild && !guild->getMotd().empty()) {
-			g_scheduler.addEvent(
-			    createSchedulerTask(150, [playerID = player.getID()]() { g_game.sendGuildMotd(playerID); }));
+		if (const auto& guild = player.getGuild()) {
+			if (!guild->getMotd().empty()) {
```
ranisalt, line 80: You can merge these ifs

src/game.cpp
```diff
@@ -61,12 +61,7 @@ Game::Game()
 	offlineTrainingWindow.priority = true;
 }
 
-Game::~Game()
-{
-	for (const auto& it : guilds) {
-		delete it.second;
-	}
-}
+Game::~Game() = default;
```
ranisalt, line 64: Move this to the header (or possibly omit it since it's `default`)

src/player.h
```diff
@@ -187,7 +187,7 @@ class Player final : public Creature, public Cylinder
 
 	uint32_t getClientIcons() const;
 
-	const GuildWarVector& getGuildWarVector() const { return guildWarVector; }
+	const std::vector<uint32_t>& getGuildWarVector() const { return guildWarVector; }
```
ranisalt, line 190: Just return `auto`. I would go further and change it to `getGuildWars` in a separate PR, we're not Hungarians

nekiro (approved):
Looks good, one step forward to smart ptr usage.
I like the Guild_ptr better than `std::shared_ptr<Guild>`, it's longer to write xD

ramon-bernardo:
@nekiro, can you see the latest changes? I always have doubts about converting a `const shared_ptr` to `shared_ptr`...

src/guild.cpp
```diff
@@ -51,27 +56,23 @@ GuildRank_ptr Guild::getRankByLevel(uint8_t level) const
...
 
-		if ((result = db.storeQuery(
-		         fmt::format("SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `guild_id` = {:d}", guildId)))) {
-			do {
-				guild->addRank(result->getNumber<uint32_t>("id"), result->getString("name"),
-				               result->getNumber<uint16_t>("level"));
-			} while (result->next());
-		}
-		return guild;
+	const auto guild = std::make_shared<Guild>(guildId, result->getString("name"));
```
nekiro, line 67: isnt this const redundant? I dont think it does anything

src/luascript.cpp
```diff
@@ -11795,9 +11798,8 @@ int LuaScriptInterface::luaGuildCreate(lua_State* L)
 	// Guild(id)
 	uint32_t id = getNumber<uint32_t>(L, 2);
 
-	Guild* guild = g_game.getGuild(id);
-	if (guild) {
-		pushUserdata<Guild>(L, guild);
+	if (const auto guild = g_game.getGuild(id)) {
```
nekiro, line 11801: this can be const auto&?

src/player.cpp
```diff
@@ -4114,7 +4114,7 @@ bool Player::isInWar(const Player* player) const
 		return false;
 	}
 
-	const Guild* playerGuild = player->getGuild();
+	const auto playerGuild = player->getGuild();
```
nekiro, line 4117: const auto&

src/player.cpp
```diff
@@ -4255,7 +4255,7 @@ GuildEmblems_t Player::getGuildEmblem(const Player* player) const
 		return GUILDEMBLEM_NONE;
 	}
 
-	const Guild* playerGuild = player->getGuild();
+	const auto playerGuild = player->getGuild();
```
nekiro, line 4258: const auto&

src/guild.cpp
ranisalt: But if it does, against our expectations, the compiler will catch :wink:

src/luascript.cpp
ranisalt: This returns a `shared_ptr` so using reference or a copy doesn't matter

nekiro: well probably, because we rely on the compiler doing the job for us

maattch:
Since this is now a shared pointer, i guess we dont need the `delete this` in `Guild::removeMember` anymore.

ramon-bernardo:
@maattch thanks!

src/guild.cpp
```diff
@@ -51,27 +55,23 @@ GuildRank_ptr Guild::getRankByLevel(uint8_t level) const
...
-		         fmt::format("SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `guild_id` = {:d}", guildId)))) {
-			do {
-				guild->addRank(result->getNumber<uint32_t>("id"), result->getString("name"),
-				               result->getNumber<uint16_t>("level"));
-			} while (result->next());
-		}
-		return guild;
+	const auto& guild = std::make_shared<Guild>(guildId, result->getString("name"));
+	if ((result = db.storeQuery(
+	         fmt::format("SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `guild_id` = {:d}", guildId)))) {
```
ranisalt, line 68: No need to reuse the variable

```suggestion
	if (auto result = db.storeQuery(
	         fmt::format("SELECT `id`, `name`, `level` FROM `guild_ranks` WHERE `guild_id` = {:d}", guildId))) {
```

(likely needs reformatting)

ranisalt (approved):

src/guild.cpp
Codinablack: Yeah it's just redundant. It does however make it clear to the reader that the variable is const.

Codinablack (approved):
LGTM too

EPuncker (approved):

ramon-bernardo:
After reviewing everything, it seems correct now. The last change for inline rank check

EPuncker (approved):

ranisalt:
:rocket:

Issue #4680, opened 2024-05-14
Bug transfering gold to player who doesn't exist in bank NPC


Coldensjo:
### Steps to reproduce (include any configuration/script required to reproduce)
1. Talk to the bank NPC
2. Say "hi" "transfer 500 playerwhodoesntexist"
3. Error in console

### Expected behaviour
NPC should say "This player does not exist."

### Actual behaviour
Nothing happens

### Environment
Basic tfs

### Reproduction environment
Basic tfs

EvilHero90:
Can't reproduce this
![client_nPcOlf6jTe](https://github.com/otland/forgottenserver/assets/3682541/058d5c59-aa98-460a-9567-9a0cd5f03a06)

Coldensjo:
Try saying "transfer 500 randomname" and not "transfer" "500" "randomname"

EvilHero90:
Yea this gives an error, I'll include this in #4671

PR #4683, opened 2024-05-15
Bank.lua fix on transfer if params are shifted #4680

EvilHero90:
### Pull Request Prelude

### Changes Proposed

Now it doesn't throw up an error if you say `transfer playername 1000` instead of `transfer 1000 playername`
The npc will just replay with `Hmm, my ledgers have no records of anyone with the name <1000>. Please ensure the name is correct.`

**Issues addressed:** 
#4680 

### DIFF
--- a/data/npc/scripts/bank.lua
+++ b/data/npc/scripts/bank.lua
@@ -199,6 +199,10 @@ local function creatureSayCallback(cid, type, msg)
 			receiver = receiver:trim()
 
 			-- Immediate topicList.TRANSFER_PLAYER_GOLD simulation
+			if tonumber(parts[3]) then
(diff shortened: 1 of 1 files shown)


data/npc/scripts/bank.lua
```diff
@@ -199,6 +199,10 @@ local function creatureSayCallback(cid, type, msg)
 			receiver = receiver:trim()
 
 			-- Immediate topicList.TRANSFER_PLAYER_GOLD simulation
+			if tonumber(parts[3]) then
```
marmichalski, line 202: why not try and swap them if parts[2] is not a number?

EvilHero90: sure makes sense actually

EPuncker (approved):

PR #4687, opened 2024-05-16
Use Party shared_ptr instead of raw pointers

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Replace party raw pointers with shared_ptr.
- Use `auto` and `const auto`, at all, when possible.


src/events.cpp
```diff
@@ -380,7 +380,7 @@ void Events::eventCreatureOnUpdateStorage(Creature* creature, uint32_t key, std:
 }
 
 // Party
-bool Events::eventPartyOnJoin(Party* party, Player* player)
+bool Events::eventPartyOnJoin(Party_ptr party, Player* player)
```
nekiro, line 383: const Party_ptr&? I think we are copying pointer here unnecessarily

src/game.cpp
```diff
@@ -4946,17 +4946,16 @@ void Game::playerInviteToParty(uint32_t playerId, uint32_t invitedId)
 		return;
 	}
 
-	Party* party = player->getParty();
+	auto party = player->getParty();
```
nekiro, line 4949: in this case I think Party_ptr would be better for readability

ranisalt: Are you sure you prefer `Party_ptr party = player->getParty()`? :thinking: fiesta

src/chat.cpp
```diff
@@ -564,8 +563,7 @@ ChatChannel* Chat::getChannel(const Player& player, uint16_t channelId)
 		}
 
 		case CHANNEL_PARTY: {
-			Party* party = player.getParty();
-			if (party) {
+			if (const auto& party = player.getParty()) {
```
Codinablack, line 566: It's been my experience that declaring a variable and checking it inside the if statement causes crashes. I'm sure this isn't true in all cases, but I'm just stating what has been my experience so far.

ranisalt: How so? It's equivalent other than the scoping (which would cause compilation errors, not crashes)

Would be interesting to see a form of reproduction though

src/party.cpp
```diff
@@ -12,11 +12,11 @@
 extern Game g_game;
 extern Events* g_events;
 
-Party::Party(Player* leader) : leader(leader) { leader->setParty(this); }
+Party::Party(Player* leader) : leader(leader) { leader->setParty(shared_from_this()); }
```
nekiro, line 15: shared_from_this() is not available in constructor, so this will throw exception

ranisalt: Also, arguably `setParty` is responsibility of the function that creates the party.

nekiro: There should be clear ownership which is now unknown, we could say its the leader who owns a party, but this may lead to crashes if leader is deleted ex on logout. Imo this should follow the pattern and store every created party in game class, party class should have static factory method such as createParty (similar to createNpc)

PR #4688, opened 2024-05-16
[Feature] Logger

ArturKnopik:
### Pull Request Prelude

### Changes Proposed
Logger that can handle LUA and C++ calls.

logger.conf - logger config file, not required

```python 
#log level (debug, info, warning, error, fatal), default warning
log_level=warning
#max log files, default 2
max_log_files=2
#max size for single log file in bytes, default 1048576(1MB)
log_file_size=1048576
#log file name without ".log", default "tfs"
log_file_name=tfs
#the maximum length of the file address that will be printed in log message, default 35 characters
file_adress_size=35
```

```LUA
local logger = GlobalEvent("loggerTest")
function logger.onStartup()
	logD("debug msg")
	logI("info msg")
	logW("warning msg")
	logE("error msg")
	logF("fatal msg")
end
logger:register()
```
image

**Issues addressed:** 
Lack of true logger.


vc17/theforgottenserver.vcxproj
```diff
@@ -338,4 +340,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
```
ArturKnopik, line 341: VS autoformater...

logger.conf.dist
```diff
@@ -0,0 +1,10 @@
+#log level (debug, info, warning, error, fatal), default debug
```
ArturKnopik, line 1: fix default to warning

ArturKnopik:
todo: missing month

src/luascript.cpp
```diff
@@ -1163,6 +1164,13 @@ void LuaScriptInterface::registerFunctions()
 	luaL_register(luaState, "result", LuaScriptInterface::luaResultTable);
 	lua_pop(luaState, 1);
 
+	// logger
+	lua_register(luaState, "logD", LuaScriptInterface::luaLogDebug);
+	lua_register(luaState, "logI", LuaScriptInterface::luaLogInfo);
+	lua_register(luaState, "logW", LuaScriptInterface::luaLogWarning);
+	lua_register(luaState, "logE", LuaScriptInterface::luaLogError);
+	lua_register(luaState, "logF", LuaScriptInterface::luaLogFatal);
```
ramon-bernardo, line 1172: It could be a `class` with some methods

```lua
Log.debug("debug")
Log.info("info")
Log.warn("warn")
Log.error("error")
Log.fatal("fatal")
```

src/logger.h
```diff
@@ -0,0 +1,68 @@
...
+#include <source_location>
+#include <string>
+#include <thread>
+
+enum LogSeverity
+{
+	LDEBUG = 0,
+	LINFO = 1,
+	LWARNING = 2,
+	LERROR = 3,
```
ArturKnopik, line 17: L prefix is added to shut up VS, it's complaining about ERROR(is treated as maro to something)

src/logger.cpp
```diff
@@ -0,0 +1,299 @@
+#include "otpch.h"
+
+#include "Logger.h"
```
ArturKnopik, line 3:
```suggestion
#include "logger.h"
```

Shawak:
I really dislike the fact that it doesn't even prints error messages on stderr. Also the output looks kinda ugly, especially with these `_E_`, _F_`, etc.

I would also probably replace fatal with trace.

ranisalt:
What's the purpose of making it a separate thread? I would rather keep it simple and just wrap around [Boost.Log](https://www.boost.org/doc/libs/1_85_0/libs/log/doc/html/index.html), which is already thread-safe.

ArturKnopik:
clarification:
why no boost.log/spdlog - I didn't want to add another dependency
1) message format - I use the format that is used in my work (the project is over 25 years old), it works quite well
2) why separated thread: the file open + write operation is expensive, in many cases the file will be opened every time we log, I don't want to block the main thread by writing logs etc.
3) why a separate configuration file for logger: config.lua is loaded relatively late and I want to have a logger from the very beginning

ranisalt:
> * why separated thread: the file open + write operation is expensive, in many cases the file will be opened every time we log, I don't want to block the main thread by writing logs etc.

It would be better than to write to stdout or stderr and pipe to a file if necessary. So it stays open, managed by the OS

> * why a separate configuration file for logger: config.lua is loaded relatively late and I want to have a logger from the very beginning

How about using environment variables for that? So you don't need to read any file

ArturKnopik:
> It would be better than to write to stdout or stderr and pipe to a file if necessary. So it stays open, managed by the OS

In my opinion, the log size limit is quite significant feature

PR #4692, opened 2024-05-19
Move Events methods to namespace

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Move Events methods to namespace. 


src/events.h
```diff
@@ -23,135 +23,75 @@ enum class EventInfoId
...
-		}
-	};
-
-private:
-	LuaScriptInterface scriptInterface;
-	EventsInfo info;
-};
+namespace tfs::events {
+
+bool load();
```
ramon-bernardo, line 28: I prefer to change this for each events::x, as it gives us better control over each type. However, this approach will result in repeated code and reading the same file multiple times. I suggest this because I plan to add `global events` and `move events` to the namespace.

src/events.cpp
```diff
@@ -8,18 +8,90 @@
...
+PlayerEvents playerEvents;
+MonsterEvents monsterEvents;
+
+} // namespace
+
+namespace tfs::events {
+
+bool load()
+{
+	scriptInterface.initState();
```
ramon-bernardo, line 82: Every time we call load, it initializes a new state. Do you have any suggestions to improve this?

src/combat.cpp
```diff
@@ -260,8 +259,7 @@ ReturnValue Combat::canDoCombat(Creature* caster, Tile* tile, bool aggressive)
 	if (aggressive && tile->hasFlag(TILESTATE_PROTECTIONZONE)) {
 		return RETURNVALUE_ACTIONNOTPERMITTEDINPROTECTIONZONE;
 	}
-
-	return g_events->eventCreatureOnAreaCombat(caster, tile, aggressive);
+	return tfs::events::creature::onAreaCombat(caster, tile, aggressive);
```
ramon-bernardo, line 262: I'm still unsure whether to use tfs::events or tfs::game::events. What do you think?

ramon-bernardo (commented):
I need assistance with these points. What decision should I make?

src/events.cpp
ranisalt: Isn't there a `reinitState` function?

src/combat.cpp
ranisalt: `tfs::events`

src/events.cpp
```diff
@@ -8,18 +8,90 @@
...
+	int32_t onSpellCheck = -1;
+};
+
+struct MonsterHandlers
 {
+	int32_t onDropLoot = -1;
+	int32_t onSpawn = -1;
+};
+
+CreatureHandlers creatureHandlers;
```
ranisalt, line 71: Use `struct { ... } creatureHandlers;` so you create everything in a single statement

ramon-bernardo: Thanks

MillhioreBT:
@ranisalt to what extent is it good to use namespace to handle all singletons?
I personally like namespaces, but is it in its main function? Anyway, I can't think of any other idea.

src/events.cpp
ramon-bernardo: Done

ranisalt:
> to what extent is it good to use namespace to handle all singletons?

Singletons are ~almost~ always a bad idea. Namespaces are an enhanced `private` section in a class, since you hide the implementation details and the compiler can assume it's not visible and perform more aggressive optimizations.

I like [this post](https://archive.is/UPBo8) on how singletons are an awful idea in general. Namespaces don't solve it, but they make it less bad.

PR #3522, opened 2021-07-30
Fix potential overflow crash in NPC system

Znote:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Restrict checks above 32bit to prevent overflow crash in-game from etc bank NPC's. 

**Issues addressed:** 

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4693, opened 2024-05-20
Fix bank NPC overflow protection being bypassed

glecko:
### Pull Request Prelude

### Changes Proposed

A bank NPC overflow protection was added back in https://github.com/otland/forgottenserver/pull/3522. However (at least with the current LuaJIT version, haven't tested others), `tonumber` is reserved for (and interpreted as) a standard library basic function.

This change simply renames the variable to avoid the conflict.

As a side note, when testing this (even when the check was being bypassed), I was never able to reproduce any crash, if I said "deposit [number larger than 2^32]" while having that amount of gold in my inventory, the NPC correctly allowed me to deposit & withdraw those amounts. Perhaps the check needs to be removed completly.

### DIFF
--- a/data/npc/lib/npc.lua
+++ b/data/npc/lib/npc.lua
@@ -103,11 +103,11 @@ end
 
 function getCount(string)
 	local b, e = string:find("%d+")
-	local tonumber = tonumber(string:sub(b, e))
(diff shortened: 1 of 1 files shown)


ranisalt (approved):
LGTM, just a variable name change to avoid clashing :+1:

PR #4707, opened 2024-05-26
Remove Ubuntu build workflow

ranisalt:
### Pull Request Prelude

### Changes Proposed

Ubuntu is outdated by definition, which prevents us from adopting updated versions of compilers and dependencies. There isn't much added value by testing with Ubuntu anyway, and the `vcpkg` workflow covers building and testing already.

`vcpkg` is well adopted in the community, and has been supported for years.

This does not mean Ubuntu is not supported: it is still officially supported, just like other common distros, as long as we have documentation about it in the wiki.


EvilHero90 (approved):

DSpeichert:
That workflow was providing CodeQL reporting, so I guess that needs to be moved to a different workflow now.

PR #4716, opened 2024-06-01
Http login missing VS Project Include files and login problem when on localhost

EvilHero90:
### Pull Request Prelude

### Changes Proposed
Including the files into the VS Project file.
Fixes a problem when server and client are both on localhost.
The localhost session ip always gets converted into a ipv4 while the client ip is ipv6 which gives a missmatch when connecting.

**Issues addressed:** 

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -441,8 +441,7 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
 	}
 
 	Connection::Address sessionIP = boost::asio::ip::make_address(result->getString("session_ip"));
-	Connection::Address localhost = boost::asio::ip::make_address("127.0.0.1");
-	if (sessionIP != localhost && ip != sessionIP) {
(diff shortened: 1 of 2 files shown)


src/protocolgame.cpp
```diff
@@ -440,9 +440,10 @@ void ProtocolGame::onRecvFirstMessage(NetworkMessage& msg)
 		return;
 	}
 
-	if (ip != boost::asio::ip::make_address(result->getString("session_ip"))) {
+	Connection::Address sessionIP = boost::asio::ip::make_address(result->getString("session_ip"));
+	Connection::Address localhost = boost::asio::ip::make_address("127.0.0.1");
+	if (sessionIP != localhost && ip != sessionIP) {
```
ranisalt, line 445:
```suggestion
	Connection::Address sessionIP = boost::asio::ip::make_address(result->getString("session_ip"));
	if (!sessionIP.is_loopback() && ip != sessionIP) {
```

Will work for both IPv4 and IPv6 localhost

ranisalt (approved):

PR #4723, opened 2024-06-03
Update build-vcpkg.yml

EvilHero90:
### Pull Request Prelude

### Changes Proposed
Fixes #4721 since it introduced syntax errors.

**Issues addressed:** 


EPuncker (approved):

ranisalt:
Weird af

> When you use expressions in an if conditional, you can, optionally, omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression.

[ref](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)

Or was it the quoting?

EvilHero90:
> Weird af
> 
> > When you use expressions in an if conditional, you can, optionally, omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression.
> 
> [ref](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)
> 
> Or was it the quoting?

Could probably be the case, I just sticked to what was the most common I've found

PR #4724, opened 2024-06-03
[Revnpcsys] Revisioned Npc System v1.1

EvilHero90:
### Pull Request Prelude

### Changes Proposed
**Revisioned Npc System v1.1**
- npc now correctly clears all npc id data when he is removed
- farewell and greet messages can be individually set per talk state if not they fall back to default
- fixed a bug if the npc disappeared while the player was talking to him, then the npc would not respond to the player again once he was spawned
- better warning/error messages which show which npc it occured and which file it originates from
- added shop callback function `callback(npc, player, handler, items, afterDiscount)` order:
`discount (if there is any) -> callback`
- added `NpcsHandler:player()` this is basicly the equivalent to `requirements` but instead of requireing you can add/set stuff to the player (modules.lua)
- implemented modules they're load after callbacks order:
`requirements (if returns true) -> callback (if returns true) -> modules`
- moved teleport/addItems/setStorageValue to modules
- added new requirements
- changed the requirements functions
I've included the error message directly inside the functions
- removed `:failureRespond` & `:getFailureRespond` as they're redundant with requirement re work
- changed modules `outfit` function it requires to insert 2 different outfits now because of 2 different player sexes
ex: `:outfit(id1, id2, addon)`
- a few linter/annotation fixes
- if the player changed the talk state but still had the shop open and then tried to purchase something it thrown errors, this is now fixed

**Issues addressed:** 

### DIFF
--- a/data/npc/lib/revnpcsys/events.lua
+++ b/data/npc/lib/revnpcsys/events.lua
@@ -235,7 +235,9 @@ if not NpcEvents then
             -- checking for requirements
             local ret, msg = handler:getTalkState(creature):requirements():init(creature)
             if not ret then
-                talkQueue:addToQueue(creature, msg, TALK.defaultDelay)
+                if msg then
+                    talkQueue:addToQueue(creature, msg, TALK.defaultDelay)
+                end
                 local _, start = next(handler.keywords)
                 handler:setTalkState(start, creature)
                 handler:getTalkState(creature):checkOnStorage(creature, handler)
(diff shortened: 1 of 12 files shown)


data/npc/lib/revnpcsys/events.lua
```diff
@@ -73,18 +73,24 @@ if not NpcEvents then
...
+        -- If the creature is a player and is focused on the NPC, the focus is removed and the talk state is reset
+        if creature:isPlayer() and focus:isFocused(creature) then
             local handler = NpcsHandler(npc)
+            focus:removeFocus(creature)
             talkQueue:clearQueue(creature)
             handler:setTalkState(handler, creature)
             handler:resetData(creature)
         end
+
+        -- Npc is beeing removed we clear all the data which is specific to it's npc id he holds to not leak memory
```
marmichalski, line 88: 🐝 
```suggestion
        -- Npc is being removed we clear all the data which is specific to it's npc id it holds to not leak memory
```

EvilHero90: good catch 👍

EPuncker (approved):

PR #4725, opened 2024-06-03
[bugfix] default duration on decaying items with no min or maxvalue set

Coldensjo:
This fixes an issue with items that have no min or max value set. For example.
Before:
`<attribute key="duration" value="300"/>`
Means duration 1-300 seconds.
After
`<attribute key="duration" value="300"/>`
Means duration 300-300 seconds.
Same as
`<attribute key="duration" minvalue="300" maxvalue="300"/>`

problem was introduced in #4531


EvilHero90 (approved):
works fine now

PR #4730, opened 2024-06-04
Remove Account class and account.h file

ranisalt:
### Pull Request Prelude

### Changes Proposed

The `Account` class serves just as a temporary storage for database queries and is not really necessary, since it's never used to pass data between functions. The `loadAccount` function is only used once and can be inlined and some columns can be omitted from it.


EvilHero90:
Need to remove the dependencies of the vs Project File

ranisalt:
> To be honest I'd much rather see Account becoming a metatable, there's been a couple times where that would have been nice

Agreed, but it can be done in Lua, doesn't need the C++ overhead :P

EPuncker (approved):

PR #4734, opened 2024-06-05
Added support for the Premake build system.

Codinablack:
### Pull Request Prelude

### Changes Proposed

No issues

Adds support for generating project files using the Premake build system.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -1,6 +1,6 @@
 workspace "The Forgotten Server"
    configurations { "Debug", "Release"}
-   platforms { "64" }
+   platforms { "64", "ARM64" }
    location ""
    editorintegration "On"
 
@@ -14,8 +14,8 @@ workspace "The Forgotten Server"
       files { "src/**.cpp", "src/**.h" }
       removefiles {"src/tests/*.cpp", "src/tests/*.h", "src/http/tests/*.cpp", "src/http/test/*.h"}
       flags {"LinkTimeOptimization", "MultiProcessorCompile"}
-      vectorextensions "AVX"
       enableunitybuild "On"
+      intrinsics   "On"
 
       filter "configurations:Debug"
          defines { "DEBUG" }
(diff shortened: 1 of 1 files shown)


ranisalt:
Why tho? Is this needed for some other project to embed TFS, or is it just another build system that will fade into darkness like vcxproj?

Codinablack:
Well to be honest, I would hope that everyone would get used to how much easier and how little maintenance work is needed, to use premake, and could come to realize that it also doesn't give people as many problems as cmake, and possibly come to replace it.

The thing is this, it's way easier to read and write the premake file, and it doesn't actually hurt anything nor affect CMake builds in any way at all. I understand that there isn't really a "need" for it, but it also wouldn't hurt anything, and its literally a single file. I can make the alteration's I recently made to the black tek premake file as well, so that it will work completely out of the box with vcpkg manifest mode in both linux and windows (I have no one who can test mac stuff for me). 

Another bonus I like about premake, with the way I have it configured, when building in release, all the code is built-in to the executable (with the single exception of a lua.dll for luajit builds), which means all the code is optimized by the compiler. I'm sure there are ways to alter the cmake configurations or to manually make changes in visual studio for this, but I don't know how to do it in cmake (and honestly don't care to learn), and changing it in visual studio solution doesn't enable this kind of feature for users building on linux/unix/mac. 

One more benefit of premake, is that it is HIGHLY configurable, as its written in lua, it allows all the logic and functionality that lua brings to the table. So instead of creating MANY files to make things work for cmake, or add additional "configurations", it can all be done very easily in the premake file, and is also familiar to our end users, seeing as how lua is the chosen interface for our end users. 

TL:DR
It doesn't hurt anything, it gives more options to the user, and is not really a burden to maintain.

Codinablack:
Update. I made the alterations to add ARM64 support that gesior made for black-tek. This includes alterations that will set the path for dependencies installed with vcpkg manifest mode. Currently, there is no "config" used to determine whether or not the user is using vcpkg manifest, instead, this build system presumes you are for linux systems, and enforces the use on windows systems. These settings ofc are easy enough to change, but as it is now, everything works out of box using vcpkg manifest mode for installation of the dependencies.

PR #4742, opened 2024-06-07
[Revscriptsys] fixing memory leak in Actions & Moveevents

EvilHero90:
### Pull Request Prelude

### Changes Proposed
Actions and Moveevents where leaking memory due to having the temporary vectors for registering ids in the pointer itself
Since we copy the values when we register the event we also copied the vectors which wheren't cleared already thus creating memory which was not intended.
I've moved them to their singletons (Actions/MoveEvents) and clear them every registration

**Issues addressed:** 
- #3694 
- #3692

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -191,9 +191,9 @@ bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node)
 bool Actions::registerLuaEvent(Action* event)
 {
 	Action_ptr action{event};
-	if (!getItemIdRange().empty()) {
-		const auto& range = getItemIdRange();
-		for (auto id : range) {
+	if (!getItemIdRange(event).empty()) {
+		const auto& range = getItemIdRange(event);
+		for (auto& id : range) {
 			auto result = useItemMap.emplace(id, *action);
 			if (!result.second) {
 				std::cout << "[Warning - Actions::registerLuaEvent] Duplicate registered item with id: " << id
@@ -201,9 +201,9 @@ bool Actions::registerLuaEvent(Action* event)
 			}
 		}
 		return true;
-	} else if (!getUniqueIdRange().empty()) {
-		const auto& range = getUniqueIdRange();
-		for (auto id : range) {
+	} else if (!getUniqueIdRange(event).empty()) {
+		const auto& range = getUniqueIdRange(event);
+		for (auto& id : range) {
 			auto result = uniqueItemMap.emplace(id, *action);
 			if (!result.second) {
 				std::cout << "[Warning - Actions::registerLuaEvent] Duplicate registered item with uid: " << id
@@ -211,9 +211,9 @@ bool Actions::registerLuaEvent(Action* event)
 			}
 		}
 		return true;
-	} else if (!getActionIdRange().empty()) {
-		const auto& range = getActionIdRange();
-		for (auto id : range) {
+	} else if (!getActionIdRange(event).empty()) {
+		const auto& range = getActionIdRange(event);
+		for (auto& id : range) {
 			auto result = actionItemMap.emplace(id, *action);
 			if (!result.second) {
 				std::cout << "[Warning - Actions::registerLuaEvent] Duplicate registered item with aid: " << id
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -191,8 +191,8 @@ bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node)
 bool Actions::registerLuaEvent(Action* event)
 {
 	Action_ptr action{event};
-	if (!action->getItemIdRange().empty()) {
-		const auto& range = action->getItemIdRange();
+	if (!getItemIdRange().empty()) {
+		const auto& range = getItemIdRange();
 		for (auto id : range) {
 			auto result = useItemMap.emplace(id, *action);
 			if (!result.second) {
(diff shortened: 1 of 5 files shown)


EPuncker:
related to #3694 and perhaps #3692 (linked on the mentioned PR)

MillhioreBT:
It looks good, although these changes make Revscript objects unsafe to store, since they don't contain their own content themselves.

In OTLand I have seen several scripts that create objects at the beginning of the file and register these objects at the end of the file, so each new object will overwrite the vectors and previous objects would be registered incorrectly.

ranisalt:
> It looks good, although these changes make Revscript objects unsafe to store, since they don't contain their own content themselves.

+1 I don't think this is a good solution

EvilHero90:
> It looks good, although these changes make Revscript objects unsafe to store, since they don't contain their own content themselves.
> 
> In OTLand I have seen several scripts that create objects at the beginning of the file and register these objects at the end of the file, so each new object will overwrite the vectors and previous objects would be registered incorrectly.

Good catch, however this was actually never intended to work that way anyway, we gave a clear structure on how to create revscripts.
I could pass the superior object along with it and if the vectors are not empty but the object is different then throw a warning
`You are trying to overwrite ids/aids/uids from another Action/Moveevent, please fix your script setup and first register the object before adding ids/aids/uids to the next object`

MillhioreBT:
> > It looks good, although these changes make Revscript objects unsafe to store, since they don't contain their own content themselves.
> > In OTLand I have seen several scripts that create objects at the beginning of the file and register these objects at the end of the file, so each new object will overwrite the vectors and previous objects would be registered incorrectly.
> 
> Good catch, however this was actually never intended to work that way anyway, we gave a clear structure on how to create revscripts. I could pass the superior object along with it and if the vectors are not empty but the object is different then throw a warning `You are trying to overwrite ids/aids/uids from another Action/Moveevent, please fix your script setup and first register the object before adding ids/aids/uids to the next object`

Lua is well known for its flexibility when it comes to doing what we want with Lua objects, forcing the programmer or user to follow rules and design patterns seems to me to be too much, I know that your `revscript 2.0` system solves this , as long as you use your assistant,
But if I don't want to use it I will be forced to follow these rules.

I prefer the solution of stealing the vector using swap or simply creating the temporary copy, then cleaning it and then moving it to the actions vector.

example:
image

or swap:
image
image

EvilHero90:
https://github.com/otland/forgottenserver/pull/4742/commits/6e0a6f3d4c51570c342dbd762abd46b728199301
I've reworked it and made a `std::map<Object*, std::vector<uint16_t>>` out of it, now it does behave like it was before but we avoid all this copy/borrow/steal stuff
@MillhioreBT

ranisalt (dismissed):
Seems quite solid now, solves the leak and also saves memory on actions after registering.

EPuncker (approved):

gesior:
@EvilHero90 
I get into similar problem on 1.4. Can you explain how can I reproduce memory leak?
Or if it was memory leak or memory management optimization to save some KBs of RAM?

> Actions and Moveevents where leaking memory due to having the temporary vectors for registering ids in the pointer itself

These pointers - created in `luascripts.cpp` - are deleted from memory with all their IDs structures by unique_pointer `Action_ptr action{event};` (add destructor to `Action` to test it or compile server with ASAN - it will fail with old code):
https://github.com/otland/forgottenserver/pull/4742/files#diff-1445d56ba3a14e3ecba7e7ad4f2b573d7798789e2547f44cbf092aa0ee69c668R193
This part of old code in `luascript.cpp`:
```cpp
		pushBoolean(L, g_actions->registerLuaEvent(action));
// 'action' is already deleted
		action->clearActionIdRange();
		action->clearItemIdRange();
		action->clearUniqueIdRange();
```
executed functions on already deleted object `action`. You can delete these 3 lines and there is no more ASAN errors (same for movements).

As I understand, thing you fixed is `Actions` singleton `useItemMap` etc. do not store item IDs inside `Action` copies anymore:
```cpp
	using ActionUseMap = std::map<uint16_t, Action>;
	ActionUseMap useItemMap;
```
**Right? So it's not 'memory leak' in C++ meaning of that word, but memory optimization?**

MillhioreBT:
> @EvilHero90 I get into similar problem on 1.4. Can you explain how can I reproduce memory leak? Or if it was memory leak or memory management optimization to save some KBs of RAM?
> 
> > Actions and Moveevents where leaking memory due to having the temporary vectors for registering ids in the pointer itself
> 
> These pointers - created in `luascripts.cpp` - are deleted from memory with all their IDs structures by unique_pointer `Action_ptr action{event};` (add destructor to `Action` to test it or compile server with ASAN - it will fail with old code): https://github.com/otland/forgottenserver/pull/4742/files#diff-1445d56ba3a14e3ecba7e7ad4f2b573d7798789e2547f44cbf092aa0ee69c668R193 This part of old code in `luascript.cpp`:
> 
> ```c++
> 		pushBoolean(L, g_actions->registerLuaEvent(action));
> // 'action' is already deleted
> 		action->clearActionIdRange();
> 		action->clearItemIdRange();
> 		action->clearUniqueIdRange();
> ```
> 
> executed functions on already deleted object `action`. You can delete these 3 lines and there is no more ASAN errors (same for movements).
> 
> As I understand, thing you fixed is `Actions` singleton `useItemMap` etc. do not store item IDs inside `Action` copies anymore:
> 
> ```c++
> 	using ActionUseMap = std::map<uint16_t, Action>;
> 	ActionUseMap useItemMap;
> ```
> 
> **Right? So it's not 'memory leak' in C++ meaning of that word, but memory optimization?**

In that specific case, yes, it is memory optimization, since the Actions maintained the vector with IDs that were not even being used; they were only there for their registration but nothing more.

PR #4745, opened 2024-06-09
Fix session table in schema

ArturKnopik:
### Pull Request Prelude

### Changes Proposed
Fix session table default value for expired_at field.

**Issues addressed:** 

https://github.com/otland/forgottenserver/blob/7011488166d1b6a398a51fc56fe854702a69a160/src/protocolgame.cpp#L429 
```CPP
	auto result = db.storeQuery(fmt::format(
	    "SELECT `a`.`id` AS `account_id`, INET6_NTOA(`s`.`ip`) AS `session_ip`, `p`.`id` AS `character_id` FROM `accounts` `a` JOIN `sessions` `s` ON `a`.`id` = `s`.`account_id` JOIN `players` `p` ON `a`.`id` = `p`.`account_id` WHERE `s`.`token` = {:s} AND `s`.`expired_at` IS NULL AND `p`.`name` = {:s} AND `p`.`deletion` = 0",
	    db.escapeString(sessionToken), db.escapeString(characterName)));
```

is checks if expired_at is null(must be null), Xampp for example set default value to 0000-00-00 00:00:00

### DIFF
--- a/data/migrations/36.lua
+++ b/data/migrations/36.lua
@@ -8,7 +8,7 @@ function onUpdateDatabase()
 			`account_id` int NOT NULL,
 			`ip` varbinary(16) NOT NULL,
 			`created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
-			`expired_at` timestamp,
+			`expired_at` timestamp NULL DEFAULT NULL,
 			PRIMARY KEY (`id`),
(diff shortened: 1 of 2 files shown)


ranisalt (approved):
`NULL` should be the default when omitted, but I guess it doesn't hurt to repeat it

EPuncker (approved):

PR #4747, opened 2024-06-09
[Revnpcsys] Revisioned Npc System v1.3

EvilHero90:
### Pull Request Prelude

### Changes Proposed
- Added `onSight` event, this has to be enabled with `NpcType:sight(x, y)` unless that is not set the event wont fire. This event triggers whenever a creature (monsters/npcs/players) steps into the sight range, will only re trigger once you walk out of idle range and back in again.
- fixed event callbacks, they're now moved into `NpcCallbacks` they can be set like: `NpcType:callbacks().onSay/onSight/etc.`
- fixed a crash which could occur if you had a npc disabled then enabled him and /reload npcs, this is now fixed
- fixed a crash that if a xml npc is invalid on server startup that it throws an error instead of crashing
- made all the functions in the lib overloadable, only the table data is protected, you can now change functions and they will be correctly changed on /reload npcs
- changed everything related to focus from seconds into ms
- changed focus so it works for all creatures not only players
- changed NpcTalkQueue it now works for creatures not only for players
- changed the onCreatureSay event, it now triggers for all creatures not only for players
- selfSay has a new parameter now called talkType which is optional and falls back to TALKTYPE_SAY
- added `onSpeechBubble` event, with this we can set the speech bubble for each player individual.
example:
Npc has a quest for the player he only shows for this player the quest speech bubble, once the quest is finished the npc wont have a quest speech bubble anymore

**Issues addressed:** 

### DIFF
--- a/data/npc/lib/revnpcsys/focus.lua
+++ b/data/npc/lib/revnpcsys/focus.lua
@@ -28,56 +28,56 @@
 if not NpcFocus then
     -- If NpcFocus doesn't exist, it's created as an empty table
     NpcFocus = {}
-    -- The metatable is set up to call the function when NpcFocus is called
-    setmetatable(NpcFocus, {
-        __call = function(self, npc)
-            -- If the NPC doesn't have a NpcFocus, one is created for it
-            if not self[npc:getId()] then
-                self[npc:getId()] = {
-                    focus = {},
-                    currentFocus = nil
-                }
-                setmetatable(self[npc:getId()], {__index = NpcFocus})
-            end
-            -- The NpcFocus is returned
-            return self[npc:getId()]
+end
+-- The metatable is set up to call the function when NpcFocus is called
+setmetatable(NpcFocus, {
+    __call = function(self, npc)
+        -- If the NPC doesn't have a NpcFocus, one is created for it
+        if not self[npc:getId()] then
+            self[npc:getId()] = {
+                focus = {},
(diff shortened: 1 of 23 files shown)


EPuncker (dismissed):

data/npc/lib/revnpcsys/focus.lua
```diff
@@ -2,27 +2,27 @@
     >> NpcFocus <<
 
     Description:
-        - The NpcFocus module provides a way to manage the focus of NPCs on players.
-        - It keeps track of the players that are currently being focused by an NPC, as well as the duration of the focus.
-        - The class provides methods to add, check, and remove focus on players, as well as retrieve the currently focused player.
+        - The NpcFocus module provides a way to manage the focus of NPCs on creatures.
+        - It keeps track of the creatures that are currently being focused by an NPC, as well as the duration of the focus.
+        - The class provides methods to add, check, and remove focus on creaturess, as well as retrieve the currently focused creature.
```
EPuncker, line 7:
```suggestion
        - The class provides methods to add, check, and remove focus on creatures, as well as retrieve the currently focused creature.
```

EvilHero90: nice catch :D

EPuncker (dismissed):

EPuncker (approved):

PR #4753, opened 2024-06-13
fix issues with house invites string cases

unclecellphone:
### Pull Request Prelude

### Changes Proposed
1. Do not use `toLower()` on player inputs for House, Subowner, and Door invite list, it will break for guilds and/or guild ranks with uppercase letters.
2. Fix an issue where `parseList(...)` did not properly parse strings for guild rank (off-by-1 error in substring math)
3. Additionally, to prevent issues like this in the future, compare guild names and ranks ignoring case.

**Issues addressed:** 

The code in TFS does not ignore case for guild invites, but it does force `toLower()` on inputs, so any guilds with uppercase letters in the name, or ranks with uppercase letters in the name, will not work via `@guildName` or `guildName@rank` invites, subowners, or for door control.  

Note that this is okay on some versions, including master, for guild names.  Guild names in fact *will* function on master here, but guild RANKS are still compared by string comparison across the board.  However, I think forcing both to be `toLower()` is much more readable in both scenarios, despite the fact that it is only fixing a bug for the *ranks* not the *guild names*.

### example
aleta sio -->
```
@guildName  
guildName@guildRank
guildName@guildrank
```
1. first line would work on most systems, although it would previously have done `toLower()` for no real reason
2. second line would be broken because of the capital R in guild Rank 
3. third line would also be broken, because of the bug in the substring math

### DIFF
--- a/src/guild.cpp
+++ b/src/guild.cpp
@@ -38,8 +38,8 @@ GuildRank_ptr Guild::getRankById(uint32_t rankId)
 
 GuildRank_ptr Guild::getRankByName(const std::string& name) const
 {
-	for (auto rank : ranks) { 
-		if (caseInsensitiveEqual(rank->name, name))  {
+	for (auto rank : ranks) {
+		if (caseInsensitiveEqual(rank->name, name)) {
 			return rank;
 		}
 	}
--- a/src/guild.cpp
+++ b/src/guild.cpp
@@ -37,8 +37,12 @@ GuildRank_ptr Guild::getRankById(uint32_t rankId)
 
 GuildRank_ptr Guild::getRankByName(const std::string& name) const
 {
+	auto nameToLower = name;
+	toLowerCaseString(nameToLower);
 	for (auto rank : ranks) {
-		if (rank->name == name) {
+		auto rankToLower = rank->name;
+		toLowerCaseString(rankToLower);
+		if (rankToLower == nameToLower) {
 			return rank;
 		}
 	}
--- a/src/guild.cpp
+++ b/src/guild.cpp
@@ -4,6 +4,7 @@
 #include "otpch.h"
 
 #include "guild.h"
+#include "tools.h"
 
 #include "game.h"
 
@@ -37,12 +38,8 @@ GuildRank_ptr Guild::getRankById(uint32_t rankId)
 
 GuildRank_ptr Guild::getRankByName(const std::string& name) const
 {
-	auto nameToLower = name;
-	toLowerCaseString(nameToLower);
-	for (auto rank : ranks) {
-		auto rankToLower = rank->name;
-		toLowerCaseString(rankToLower);
-		if (rankToLower == nameToLower) {
+	for (auto rank : ranks) { 
+		if (caseInsensitiveEqual(rank->name, name))  {
 			return rank;
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/guild.cpp
```diff
@@ -37,8 +37,12 @@ GuildRank_ptr Guild::getRankById(uint32_t rankId)
 
 GuildRank_ptr Guild::getRankByName(const std::string& name) const
 {
+	auto nameToLower = name;
+	toLowerCaseString(nameToLower);
 	for (auto rank : ranks) {
-		if (rank->name == name) {
+		auto rankToLower = rank->name;
+		toLowerCaseString(rankToLower);
+		if (rankToLower == nameToLower) {
```
nekiro, line 45: use caseInsensitiveEqual(rank->name, name)

https://github.com/otland/forgottenserver/blob/master/src/tools.cpp#L137

nekiro (changes requested):

src/guild.cpp
unclecellphone: thanks, this is way better, done

src/house.cpp
```diff
@@ -390,8 +390,6 @@ void AccessList::parseList(std::string_view list)
 			break;
 		}
 
-		boost::algorithm::trim(line);
```
nekiro, line 393: just a question about this, are you sure this should be moved to player only?
as far as I know trim trims all spaces from current line, which would be logical to do to avoid comparing whitespaces?

unclecellphone: wait you're right i misread this

unclecellphone: intent is to only use the toLower in the case where we might actually need it, which if anything might be nowhere actually, but definitely not in the first 3 cases here

unclecellphone (commented):
accidental review click

src/house.cpp
nekiro: I dont think we need toLower anywhere, and it is not used anywhere, cause we compare without case sensitibity, trim on the other hand should stay, because you'll want to trim trailling whitespaces from stuff players write into the window, like " nekiro" should be trimmed to "nekiro"

src/house.cpp
```diff
@@ -391,26 +391,24 @@ void AccessList::parseList(std::string_view list)
...
-				addGuildRank(line.substr(0, at_pos - 1), line.substr(at_pos + 1));
+				addGuildRank(line.substr(0, at_pos), line.substr(at_pos + 1));
 			}
 		} else if (line == "*") {
 			allowEveryone = true;
 		} else if (line.find("!") != std::string::npos || line.find("*") != std::string::npos ||
 		           line.find("?") != std::string::npos) {
 			continue; // regexp no longer supported
 		} else {
+			boost::algorithm::to_lower(line);
```
nekiro: sorry to say, but this is also not needed
to_lower is already used in getPlayerByName here: https://github.com/otland/forgottenserver/blob/master/src/game.cpp#L438

nekiro (approved):
Looks good

EPuncker (approved):

EPuncker (approved):

ramon-bernardo (approved):

ramon-bernardo:
Great, im not sure how this wasn’t noticed earlier :anguished:
image
image

PR #4761, opened 2024-06-30
Dont recreate stackables every move

nekiro:
### Pull Request Prelude

### Changes Proposed
Don't recreate stackables on every move, create new stackables only when splitting the stack.

**Issues addressed:** 
Stackable items are recreated even when they shouldn't, if you move a full stack of stackable item, why would you create a new item and delete the old one?
This seems pretty wasteful and unnecessary.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -1214,7 +1214,7 @@ ReturnValue Game::internalMoveItem(Cylinder* fromCylinder, Cylinder* toCylinder,
 		}
 
 		int32_t newCount = m - n;
-		if (newCount == item->getItemCount()) {
+		if (!updateItem && newCount == item->getItemCount()) {
 			// full item is moved (move count is the same as item count)
 		} else if (newCount > 0) {
 			moveItem = item->clone();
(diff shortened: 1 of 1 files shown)


src/game.cpp
```diff
@@ -1214,20 +1214,22 @@ ReturnValue Game::internalMoveItem(Cylinder* fromCylinder, Cylinder* toCylinder,
...
+		} else if (newCount > 0) {
 			moveItem = item->clone();
 			moveItem->setItemCount(newCount);
 		} else {
+			if (item->isRemoved()) {
+				ReleaseItem(item);
+			}
+			
 			moveItem = nullptr;
 		}
```
ranisalt, line 1228: Assuming the new count can never be negative, how about this?

```suggestion
		assert(newCount >= 0);
		if (newCount != item->getItemCount()) {
			moveItem = item->clone();
			moveItem->setItemCount(newCount);
		} else if (newCount == 0) {
			if (item->isRemoved()) {
				ReleaseItem(item);
			}
			
			moveItem = nullptr;
		}
```

ranisalt: Attention: the code above may cause duplication bugs that I couldn't trace. Please don't apply it without testing beforehand.

nekiro:
Good idea, but needs some extra work, closing this for now, cause I don't plan on working on this anymore.

PR #4764, opened 2024-07-03
Use Town as plain Lua table

ranisalt:
### Pull Request Prelude

### Changes Proposed

The `Town` metatable cannot be used to mutate towns, so it can be pushed as a table for a lighterweight implementation and to cleanup some functions in luascript


EPuncker (dismissed):

src/luascript.cpp
```diff
@@ -9765,7 +9766,7 @@ int LuaScriptInterface::luaPlayerGetTown(lua_State* L)
 	// player:getTown()
 	Player* player = tfs::lua::getUserdata<Player>(L, 1);
 	if (player) {
-		tfs::lua::pushUserdata(L, player->getTown());
+		pushTown(L, *player->getTown());
 		tfs::lua::setMetatable(L, -1, "Town");
```
MillhioreBT, line 9770: This is not necessary, since the pushTown method already adds the metatable

src/luascript.cpp
```diff
@@ -12876,9 +12824,9 @@ int LuaScriptInterface::luaHouseGetTown(lua_State* L)
 		return 1;
 	}
 
-	Town* town = g_game.map.towns.getTown(house->getTownId());
+	const Town* town = g_game.map.towns.getTown(house->getTownId());
 	if (town) {
-		tfs::lua::pushUserdata(L, town);
+		pushTown(L, *town);
 		tfs::lua::setMetatable(L, -1, "Town");
```
MillhioreBT, line 12830: Do not set the metatable, since it is already added in the pushTown method

data/lib/core/town.lua
```diff
@@ -0,0 +1,24 @@
+Town = {}
```
MillhioreBT, line 1: This file is never read, it must be added to the core.lua file

MillhioreBT: It is also necessary to add the `__index` metamethod that refers to the same table
```lua
Town.__index = Town
```

src/luascript.cpp
```diff
@@ -1177,6 +1177,16 @@ static void pushLoot(lua_State* L, const std::vector<LootBlock>& lootList)
...
 }
 
+static void pushTown(lua_State* L, const Town& town)
+{
+	lua_createtable(L, 0, 3);
+	setField(L, "id", town.id);
+	setField(L, "name", town.name);
+	tfs::lua::pushPosition(L, town.templePosition);
+	lua_setfield(L, -2, "templePosition");
+	tfs::lua::setMetatable(L, -1, "Town");
```
MillhioreBT, line 1187: adding the metatable is impossible, since the table `Town` does not actually exist in the registry `LUA_REGISTRYINDEX` this is because the TFS API registers the metatables with the `registerClass` method in this index

so it will be necessary to search for the table in the global environment and then apply it as a metatable manually

```cpp
	lua_getglobal(L, "Town");
	lua_setmetatable(L, -2);
```

data/lib/core/town.lua
ranisalt: Fixed

EPuncker (approved):

src/player.h
```diff
@@ -1219,7 +1219,7 @@ class Player final : public Creature, public Cylinder
 	Party* party = nullptr;
 	Player* tradePartner = nullptr;
 	SchedulerTask* walkTask = nullptr;
-	Town* town = nullptr;
+	const Town* town = nullptr;
```
ramon-bernardo, line 1222: could you explain why it became a const?

src/town.h
```diff
@@ -6,60 +6,27 @@
...
 	{
 		auto it = townMap.find(townId);
 		if (it == townMap.end()) {
 			return nullptr;
 		}
-		return it->second;
+		return it->second.get();
 	}
 
 	const TownMap& getTowns() const { return townMap; }
```
ramon-bernardo, line 32: We could use auto and keep the type inferred directly from the townMap variable ;)

ramon-bernardo (commented):
Here are a few considerations: as I always say, we can use auto whenever possible ;)

src/player.h
ranisalt: This is assigned with the return value of `getTown` (i.e. `player->town = towns->getTown(id)`) which became `const Town*`

src/town.h
ranisalt: I'd rather do it separately. Probably would change to `decltype(auto) getTowns() const` to avoid copying

ramon-bernardo (approved):
Some suggestions, such as moving it to a namespace like `tfs::game::towns` and using `auto` where possible. However, LTGM!

MillhioreBT (approved):

PR #4767, opened 2024-07-15
check if config file is already defined

rookgaard:
### Pull Request Prelude

### Changes Proposed

Config file may have been set in [HERE](https://github.com/otland/forgottenserver/blob/98ff8a81a4bfbf1bced304bcde356f797f870d0a/src/main.cpp#L27), so check that it is not empty.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -163,7 +163,10 @@ bool ConfigManager::load()
 
 	luaL_openlibs(L);
 
-	string[CONFIG_FILE] = "config.lua";
+	if (string[CONFIG_FILE].empty()) {
+		string[CONFIG_FILE] = "config.lua";
+	}
+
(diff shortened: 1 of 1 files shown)


MillhioreBT:
This is not configurable, so this change makes no sense.

rookgaard:
> This is not configurable, so this change makes no sense.

But I linked a place where it can be configured earlier, using `--config` parameter

MillhioreBT:
> > This is not configurable, so this change makes no sense.
> 
> But I linked a place where it can be configured earlier, using `--config` parameter

You're right, I apologize, so it does make sense for this one case.

If you want, reopen the PR and we will merge it.
thank you very much <3

MillhioreBT (approved):

EPuncker (approved):

PR #4773, opened 2024-08-14
Fix memory leaks on reloads

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
These changes will cause that when you do `/reload` the scripts will be cleaned from the reference table, this way there will be no memory leak.

The only case that is not covered is `/reload items`, for some reason this method produces an increase in memory on each reload: `bool Items::loadFromXml()`


nekiro:
Looks pretty good, I'll check this in-game soon

NRH-AA:
Finally.

PR #4785, opened 2024-09-15
Fix Monster Look Direction Behavior

### LINKED ISSUE #4339: Wrong direction of monster with hasFollowPath

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Allow monsters to update their look direction even if they cannot attack.
Stop fleeing monsters from updating their look direction.

As of now monsters will not update their look direction if they cannot attack (rl they do)
As of now fleeing monsters will look at the player randomly (rl they do not)

I can't figure out how to record rl so if someone can tell me, otherwise I can't get a video of it.

**Issues addressed:** 
closes #4339 


ranisalt:
Screencast from RL, fleeing sheep will always look in the direction it's fleeing, never looks back at player:

[Screencast from 2024-09-16 01-34-01.webm](https://github.com/user-attachments/assets/7b15c09a-5945-40bc-ad59-fbf87ca04d7d)

EPuncker:
> Screencast from RL, fleeing sheep will always look in the direction it's fleeing, never looks back at player:
> 
>  Screencast.from.2024-09-16.01-34-01.webm 
> However, if the creature is fleeing AND decides to attack player (i.e. Dragon running in low health) with a directioned attack (a wave), it **will** direct the attack towards the player. It turns to the player for a split second only to wave, and then will continue fleeing.

perhaps we should make it work only with creatures that have aggressive flag?

NRH-AA:
https://youtu.be/_cMRslA4FBw

NRH-AA:
It seems we need to make whatever spell is cast wait until the monster finishes its walk cycle. As of now the monsters behave correctly as far as updating their look direction. The only problem is the spell casts before the monster has time to finish his last walk and turn himself.

NRH-AA:
Alright I think this looks even better than rl tibia.

https://youtu.be/TLm8ExV40Hg

NRH-AA:
Alright, LGTM

src/monster.cpp
```diff
@@ -806,9 +810,11 @@ void Monster::doAttacking(uint32_t interval)
 
 		if (canUseSpell(myPos, targetPos, spellBlock, interval, inRange, resetTicks)) {
 			if (spellBlock.chance >= static_cast<uint32_t>(uniform_random(1, 100))) {
-				if (!lookUpdated) {
+				if ((isFleeing() ||
+				     mType->info.targetDistance > (myPos.getDistanceX(targetPos) + myPos.getDistanceY(targetPos))) &&
+				    inRange) {
 					updateLookDirection();
-					lookUpdated = true;
+					lastStep = OTSYS_TIME() - 200;
```
ramon-bernardo, line 817: This made the sheep extremely slow 🐢

NRH-AA: This commit needs to happen in lock step with #4637 The monsters move slowly because the 1 second delay between each walk without my optimization.

NRH-AA:
Note: This commit should be tested after #4637 has been merged.

PR #4788, opened 2024-09-19
Move map cache to monster

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed:
This update aims to enhance of the map cache usage within the codebase. The map cache is currently relevant only for the `Monster` class, which overrides the `useCacheMap` method. 

- **Implementation Details**:
  - The `useCacheMap` method specific to `Monster` is defined as follows:
    ```cpp
    bool useCacheMap() const override { return !randomStepping; }
    ```
    This method provides behavior unique to the `Monster` class by determining when the map cache should be utilized based on the `randomStepping` attribute.

  - In contrast, the base `Creature` class retains a default implementation for the `useCacheMap` method. This default implementation indicates that creatures, by default, do not use the map cache:
    ```cpp
    virtual bool useCacheMap() const { return false; }
    ```

By relocating the map cache functionality directly into the `Monster` class, we improve and reduces unnecessary complexity delineates the responsibilities of each class regarding map cache usage.

#### Additional Considerations
- This refactor does not alter any existing functionality for the `Creature` class, ensuring backward compatibility.

### DIFF
--- a/src/creature.h
+++ b/src/creature.h
@@ -289,7 +289,7 @@ public:
 	virtual void onRemoveTileItem(const Tile* tile, const Position& pos, const ItemType& iType, const Item* item) {}
 
 	virtual void onCreatureAppear(Creature* creature, bool isLogin);
-	virtual void onRemoveCreature(Creature* creature, bool isLogout);
+	virtual void onRemoveCreature(Creature* creature, bool isLogout) {}
 	virtual void onCreatureMove(Creature* creature, const Tile* newTile, const Position& newPos, const Tile* oldTile,
 	                            const Position& oldPos, bool teleport);
 
--- a/src/creature.h
+++ b/src/creature.h
@@ -282,11 +282,11 @@ public:
 	virtual void onWalk();
 	virtual bool getNextStep(Direction& dir, uint32_t& flags);
 
-	virtual void onAddTileItem(const Tile* tile, const Position& pos) {}
-	virtual void onUpdateTileItem(const Tile* tile, const Position& pos, const Item* oldItem, const ItemType& oldType,
-	                              const Item* newItem, const ItemType& newType)
+	virtual void onAddTileItem(const Tile*, const Position&) {}
+	virtual void onUpdateTileItem(const Tile*, const Position&, const Item*, const ItemType&, const Item*,
+	                              const ItemType&)
 	{}
-	virtual void onRemoveTileItem(const Tile* tile, const Position& pos, const ItemType& iType, const Item* item) {}
+	virtual void onRemoveTileItem(const Tile*, const Position&, const ItemType&, const Item*) {}
 
 	virtual void onCreatureAppear(Creature* creature, bool isLogin);
 	virtual void onRemoveCreature(Creature* creature, bool isLogout) {}
--- a/src/creature.h
+++ b/src/creature.h
@@ -282,10 +282,11 @@ public:
 	virtual void onWalk();
 	virtual bool getNextStep(Direction& dir, uint32_t& flags);
 
-	void onAddTileItem(const Tile* tile, const Position& pos);
+	virtual void onAddTileItem(const Tile* tile, const Position& pos) {}
 	virtual void onUpdateTileItem(const Tile* tile, const Position& pos, const Item* oldItem, const ItemType& oldType,
-	                              const Item* newItem, const ItemType& newType);
-	virtual void onRemoveTileItem(const Tile* tile, const Position& pos, const ItemType& iType, const Item* item);
+	                              const Item* newItem, const ItemType& newType)
+	{}
+	virtual void onRemoveTileItem(const Tile* tile, const Position& pos, const ItemType& iType, const Item* item) {}
 
 	virtual void onCreatureAppear(Creature* creature, bool isLogin);
 	virtual void onRemoveCreature(Creature* creature, bool isLogout);
@@ -328,8 +329,6 @@ public:
 	Tile* getTile() override final { return tile; }
 	const Tile* getTile() const override final { return tile; }
 
-	int32_t getWalkCache(const Position& pos) const;
-
 	const Position& getLastPosition() const { return lastPosition; }
 	void setLastPosition(Position newLastPos) { lastPosition = newLastPos; }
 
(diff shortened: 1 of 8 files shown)


src/monster.h
```diff
@@ -218,7 +231,17 @@ class Monster final : public Creature
 	uint32_t getDamageImmunities() const override { return mType->info.damageImmunities; }
 	uint32_t getConditionImmunities() const override { return mType->info.conditionImmunities; }
 	void getPathSearchParams(const Creature* creature, FindPathParams& fpp) const override;
-	bool useCacheMap() const override { return !randomStepping; }
```
ramon-bernardo, line 221: Override here.

src/creature.h
```diff
@@ -355,19 +354,12 @@ class Creature : virtual public Thing
 	decltype(auto) getStorageMap() const { return storageMap; }
 
 protected:
-	virtual bool useCacheMap() const { return false; }
```
ramon-bernardo, line 358: Virtual fn.

ramon-bernardo (commented):
See

EPuncker (approved):

src/monster.cpp
```diff
@@ -1813,6 +2027,29 @@ bool Monster::getDistanceStep(const Position& targetPos, Direction& direction, b
...
+			if (int32_t dy = pos.getOffsetY(myPos); std::abs(dy) <= maxWalkCacheHeight) {
+				if (walkabilityCache[maxWalkCacheHeight + dy][maxWalkCacheWidth + dx]) {
+					return 1;
+				}
+				return 0;
+			}
+		}
+	}
+
+	return 2;
```
ranisalt, line 2050: These magic numbers deserve names, let's change this later on

src/monster.h
```diff
@@ -218,7 +231,17 @@ class Monster final : public Creature
...
+
+	// map cache
+	bool mapLoaded = false;
+	bool walkabilityCache[walkCacheHeight][walkCacheWidth] = {{false}};
```
ranisalt, line 237: Note this will only set the first value of the first nested array to false, not every entry. You probably just want

```suggestion
	bool walkabilityCache[walkCacheHeight][walkCacheWidth] = {};
```

Or even a [`bitset`](https://en.cppreference.com/w/cpp/utility/bitset)

src/monster.h
```diff
@@ -129,13 +129,21 @@ class Monster final : public Creature
...
 	MonsterIconHashMap& getSpecialIcons() { return monsterIcons; }
 	const MonsterIconHashMap& getSpecialIcons() const { return monsterIcons; }
 
 	static uint32_t monsterAutoID;
 
 private:
+	static constexpr int32_t walkCacheWidth = Map::maxViewportX * 2 + 1;
+	static constexpr int32_t walkCacheHeight = Map::maxViewportY * 2 + 1;
+	static constexpr int32_t maxWalkCacheWidth = (walkCacheWidth - 1) / 2;
+	static constexpr int32_t maxWalkCacheHeight = (walkCacheHeight - 1) / 2;
```
ranisalt, line 145: Aren't these just

```suggestion
	static constexpr int32_t maxWalkCacheWidth = Map::maxViewportX;
	static constexpr int32_t maxWalkCacheHeight = Map::maxViewportY;
```
?

It's just reversing the calculation

src/monster.h
ramon-bernardo: Right, I'll try using bitset later

src/monster.h
ramon-bernardo: Used map constexpr direcly

ramon-bernardo:
Would this provide any performance gain if activated for players? I ran some tests, and it works for players as well as monsters.

@gesior  do you know anything about this?

gesior:
I will try to benchmark this code (player walk around vs 5-20 monsters) + somehow activate it for all creatures (players), but I've checked git and walkCache was added in first TFS 1.x commit (2013).
Someone made it for some reason 10-15 years ago. There is no documentation and I'm 99% sure nobody benchmarked this change on popular server.
It's even possible that after XX other optimizations added in TFS 1.x it does not reduce CPU usage at all.

It `queryAdd` every active (not idle) Monster to every Tile on it's screen 1 time a second (onThink), any other Creature step on it's screen resets cache (imagine there are 10 fast walking creatures, it will reset cache 10 times per 1 read).
It looks like this cache is more often written than read.
`getSpectators` map cache works other way. It generates cache, when there is query to 'get cache', but there is no cache. This walkCache algorithm may generate cache that is not used by anything.

ramon-bernardo:
@gesior that's right, I see the same thing. A cache is created, but it doesn't seem to give a significant gain here. What if we remove it? It would reduce code complexity at the cost of a few percentage of performance, or maybe even none at all... let me check who added it

gesior:
> let me check who added it

I will try to benchmark monsters with cache and without cache. Two edge scenarios are:
- 10 players vs 200 fast-walking monsters (20 per player, high exp server)
- 100 players vs 200 slow-walking monsters (2 per player, low exp server/trainers).

It's in second `forgottenserver` commit. Just after creating Git repository with README.md file.
Which means it was moved from TFS 0.2 into TFS 1.x, when Mark started developing TFS 1.x on Git (TFS 0.2 was on OTLand SVN, TFS 1.x was first version on Git).
I've checked tfs-old-svn converted into Git ( https://github.com/otland/tfs-old-svn/tree/0.2 ). WalkCache code was added in 2008 is FIRST commit of TFS 0.2 (2008). So there is no prior history. Someone made it before 2008 to fix some problem.
What problem? We don't know.
Is it faster with cache in 2024? We don't know.

Screens from TFS master and TFS-old-svn Git repos:
image
image

ramon-bernardo:
Yeah, @gesior, i was able to check the `git blame` as well. Well… unless we had TFS 0.1 👴🏻 , we won’t fully understand why it was added, but that’s fine. If the benchmark results are poor, I can remove it.

gesior:
> Yeah, @gesior, i was able to check the `git blame` as well. Well… unless we had TFS 0.1 👴🏻 , we won’t fully understand why it was added, but that’s fine. If the benchmark results are poor, I can remove it.

I always post what I've found, so others won't waste time searching for same thing.

I've tried https://github.com/opentibia/server which has history since 2002, but without branches. Only information there is:
```
Merging branch avesta
```
in 2008-11-02. Without SVN copy with all branches and changes, we won't find who/why added it.

**ANYWAY**, I will try to benchmark it in next 2 weeks and we will know, if we can remove it or we need to refactor/optimize it.

ramon-bernardo:
I like it; we'll be waiting. If you have a tutorial on how you create your tests, I could even try setting some up myself

Zbizu:
@gesior @ramon-bernardo mind that tfs was based on otserv at some point so if you are interested in digging deeper, opentibia/server repo, or sourceforge page is the way to go

ramon-bernardo:
Removed!

PR #4795, opened 2024-09-25
Improve OTB loader iteration

ranisalt:
### Pull Request Prelude

### Changes Proposed

Improves file loader performance by having less indirection and skipping checks aggressively. Loading the map is at least 20% faster on my machine using [EPuncker/orts2](https://github.com/EPuncker/orts2) map.

When running a release build, most of the sanity checks will be skipped (e.g. checking if the file does not underflow, if the nodes have expected types) since we expect map editors to generate correct map files. The map only needs to be loaded _once_ with a debug build if one wants to make sure it works, so there is no need to check again on every start.


src/item.h
ramon-bernardo, line 1: Remove Attr_ReadValue enum

ramon-bernardo (changes requested):

src/iomap.cpp
```diff
@@ -31,209 +31,114 @@
...
-		OTBM_Tile_coords tile_coord;
-		if (!propStream.read(tile_coord)) {
-			setLastErrorString("Could not read tile position.");
-			return false;
-		}
+		auto first = tileNode.propsBegin, last = tileNode.propsEnd;
 
-		uint16_t x = base_x + tile_coord.x;
-		uint16_t y = base_y + tile_coord.y;
+		uint16_t x = base_x + OTB::read<uint8_t>(first, last);
```
ramon-bernardo, line 140: We can use `auto` in most cases here.

src/iomap.cpp
```diff
@@ -384,81 +255,124 @@ bool IOMap::parseTileArea(OTB::Loader& loader, const OTB::Node& tileAreaNode, Ma
...
+	}
+
+	if (minorVersionItems > Item::items.minorVersion) {
+		fmt::print("[Warning - IOMap::loadMap] This map needs an updated items.otb.\n");
+	}
+
+	fmt::print("> Map size: {:d}x{:d}\n.", map->width, map->height);
+
+	const auto& rootNodes = loader.children();
+	if (rootNodes.size() != 1 or rootNodes.front().type != OTBM_MAP_DATA) {
```
ramon-bernardo, line 345: `or`?

src/fileloader.h
```diff
@@ -4,50 +4,103 @@
...
 
-public:
-	Loader(const std::string& fileName, const Identifier& acceptedIdentifier);
-	bool getProps(const Node& node, PropStream& props);
-	const Node& parseTree();
-};
+	(void)last;
+#ifndef NDEBUG
+	if (last - first < len) {
+		throw std::invalid_argument("Not enough bytes to read as string.");
```
ramon-bernardo, line 72: assert debug maybe?

src/fileloader.cpp
```diff
@@ -5,105 +5,71 @@
...
-using NodeStack = std::stack<Node*, std::vector<Node*>>;
-static Node& getCurrentNode(const NodeStack& nodeStack)
+auto checkIdentifier(iterator it, std::string_view acceptedIdentifier)
 {
-	if (nodeStack.empty()) {
-		throw InvalidOTBFormat{};
-	}
-	return *nodeStack.top();
+	auto identifier = std::string_view(it, 4);
+	return identifier == acceptedIdentifier or identifier == wildcard;
```
ramon-bernardo, line 17: `or`?

ramon-bernardo (commented):
rani, here are some considerations I noted while reviewing; do we have something like a debug assert? Is it the same thing?

src/items.cpp
```diff
@@ -328,51 +404,33 @@ bool Items::reload()
...
-			buildNumber = vi.dwBuildNumber;   // revision
+		if (auto length = OTB::read<uint16_t>(first, last); length != VERSION_INFO_SIZE) {
+			throw std::invalid_argument(
+			    fmt::format("Invalid data length for version info: expected 140, got {:d}", length));
 		}
+
+		majorVersion = OTB::read<uint32_t>(first, last); // items otb format file version
+		minorVersion = OTB::read<uint32_t>(first, last); // client version
+		buildNumber = OTB::read<uint32_t>(first, last);  // revision
+		OTB::skip(first, last, VERSION_INFO_SIZE - 3 * sizeof(uint32_t));
```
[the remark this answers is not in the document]
ranisalt: Yes, we don’t use that field

src/iologindata.cpp
```diff
@@ -960,15 +960,11 @@ void IOLoginData::loadItems(ItemMap& itemMap, DBResult_ptr result)
...
-		PropStream propStream;
-		propStream.init(attr.data(), attr.size());
 
 		Item* item = Item::CreateItem(type, count);
 		if (item) {
-			if (!item->unserializeAttr(propStream)) {
-				std::cout << "WARNING: Serialize error in IOLoginData::loadItems" << std::endl;
-			}
-
+			auto first = attr.begin();
```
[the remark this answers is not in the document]
ranisalt: The function takes a reference, so I can't pass a temporary (i.e. inline `attr.begin()`)

ranisalt:
Just learned about [`[[likely]]`/`[[unlikely]]`](https://en.cppreference.com/w/cpp/language/attributes/likely) in C++20 that allows me to keep ifs instead of asserts with virtually no penalty, but with much better error handling (catchable exceptions with meaningful messages)

src/iomap.cpp
```diff
@@ -384,81 +261,148 @@ bool IOMap::parseTileArea(OTB::Loader& loader, const OTB::Node& tileAreaNode, Ma
...
+		auto [x, y, z] = read_coords(first, last);
+		town->setTemplePos({x, y, z});
 	}
-	return true;
 }
 
-bool IOMap::parseWaypoints(OTB::Loader& loader, const OTB::Node& waypointsNode, Map& map)
+void parseWaypoints(const OTB::Node& waypointsNode, Map& map)
 {
 	PropStream propStream;
```
ramon-bernardo, line 292: unused?

src/iomap.cpp
```diff
@@ -31,209 +31,118 @@
...
-		if (!propStream.read(tile_coord)) {
-			setLastErrorString("Could not read tile position.");
-			return false;
-		}
+		auto first = tileNode.propsBegin, last = tileNode.propsEnd;
 
-		uint16_t x = base_x + tile_coord.x;
-		uint16_t y = base_y + tile_coord.y;
+		uint16_t x = base_x + OTB::read<uint8_t>(first, last);
+		uint16_t y = base_y + OTB::read<uint8_t>(first, last);
```
ramon-bernardo, line 145: static_cast + auto could be an alternative?

ramon-bernardo (commented):
A few more points to consider; I really liked the new approach!

Unfortunately, I won't be able to test these changes (VS is having some issues...)

src/iomap.cpp
ranisalt: Cast not needed, it is promoted automatically

src/iomap.cpp
ranisalt: Unused. I wonder how gcc didn't catch it

ranisalt:
I'm still working on it to make a single pass loader, it should save a significant amount of time

Edit: but not today

omarcopires:
I've been thinking, maybe it's time to look at @nekiro's protobuf and abandon the use of otb, like it was done in canary. I know it's not related to otb, but this pr reminded me of it!

ranisalt:
Why does Windows always have to get in the way

PR #4798, opened 2024-10-05
Rework quadtree and tests

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Rework quadtree.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -67,17 +67,7 @@ Tile* Map::getTile(uint16_t x, uint16_t y, uint8_t z) const
 	if (z >= MAP_MAX_LAYERS) {
 		return nullptr;
 	}
-
-	const QTreeLeafNode* leaf = QTreeNode::getLeafStatic<const QTreeLeafNode*, const QTreeNode*>(&root, x, y);
-	if (!leaf) {
-		return nullptr;
-	}
-
-	const Floor* floor = leaf->getFloor(z);
-	if (!floor) {
-		return nullptr;
-	}
-	return floor->tiles[x & FLOOR_MASK][y & FLOOR_MASK];
+	return tfs::quadtree::find_tile(x, y, z);
 }
 
 void Map::setTile(uint16_t x, uint16_t y, uint8_t z, Tile* newTile)
@@ -87,41 +77,7 @@ void Map::setTile(uint16_t x, uint16_t y, uint8_t z, Tile* newTile)
 		return;
 	}
 
-	QTreeLeafNode::newLeaf = false;
-	QTreeLeafNode* leaf = root.createLeaf(x, y, 15);
-
-	if (QTreeLeafNode::newLeaf) {
-		// update north
-		QTreeLeafNode* northLeaf = root.getLeaf(x, y - FLOOR_SIZE);
-		if (northLeaf) {
-			northLeaf->leafS = leaf;
-		}
-
-		// update west leaf
-		QTreeLeafNode* westLeaf = root.getLeaf(x - FLOOR_SIZE, y);
-		if (westLeaf) {
-			westLeaf->leafE = leaf;
-		}
-
-		// update south
-		QTreeLeafNode* southLeaf = root.getLeaf(x, y + FLOOR_SIZE);
-		if (southLeaf) {
-			leaf->leafS = southLeaf;
-		}
-
-		// update east
-		QTreeLeafNode* eastLeaf = root.getLeaf(x + FLOOR_SIZE, y);
-		if (eastLeaf) {
-			leaf->leafE = eastLeaf;
-		}
-	}
-
-	Floor* floor = leaf->createFloor(z);
-	uint32_t offsetX = x & FLOOR_MASK;
-	uint32_t offsetY = y & FLOOR_MASK;
-
-	Tile*& tile = floor->tiles[offsetX][offsetY];
-	if (tile) {
+	if (auto tile = tfs::quadtree::find_tile(x, y, z)) {
 		TileItemVector* items = newTile->getItemList();
 		if (items) {
 			for (auto it = items->rbegin(), end = items->rend(); it != end; ++it) {
@@ -137,7 +93,7 @@ void Map::setTile(uint16_t x, uint16_t y, uint8_t z, Tile* newTile)
 		}
 		delete newTile;
 	} else {
-		tile = newTile;
+		tfs::quadtree::create_tile(x, y, z, newTile);
 	}
 }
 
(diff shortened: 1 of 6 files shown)


src/quadtree.cpp
```diff
@@ -0,0 +1,199 @@
+#include "quadtree.h"
+
+namespace {
+std::unique_ptr<QuadTree> nodes[4] = {};
```
ranisalt, line 4:
```suggestion
std::array<std::unique_ptr<QuadTree>, 4> nodes = {};
```

src/quadtree.h
```diff
@@ -0,0 +1,65 @@
...
+	explicit QuadTree() = default;
+
+	virtual bool is_leaf() = 0;
+
+	std::unique_ptr<QuadTree> nodes[4] = {};
+};
+
+static constexpr int32_t FLOOR_BITS = 3;
+static constexpr int32_t FLOOR_SIZE = (1 << FLOOR_BITS);
+static constexpr int32_t FLOOR_MASK = (FLOOR_SIZE - 1);
```
ranisalt, line 29:
```suggestion
inline constexpr int32_t FLOOR_BITS = 3;
inline constexpr int32_t FLOOR_SIZE = (1 << FLOOR_BITS);
inline constexpr int32_t FLOOR_MASK = (FLOOR_SIZE - 1);
```

`static` was a mistake here

src/quadtree.cpp
```diff
@@ -0,0 +1,199 @@
...
+	int32_t start_x_aligned = start_x - (start_x % FLOOR_SIZE);
+	int32_t start_y_aligned = start_y - (start_y % FLOOR_SIZE);
+	int32_t end_x_aligned = end_x - (end_x % FLOOR_SIZE);
+	int32_t end_y_aligned = end_y - (end_y % FLOOR_SIZE);
+
+	if (auto start_leaf = find_leaf_in_root(start_x_aligned, start_y_aligned)) {
+		Leaf* south_leaf = start_leaf;
+		Leaf* east_leaf;
+
+		for (int_fast32_t ny = start_y_aligned; ny <= end_y_aligned; ny += FLOOR_SIZE) {
```
ranisalt, line 72:
```suggestion
		for (int32_t ny = start_y_aligned; ny <= end_y_aligned; ny += FLOOR_SIZE) {
```

src/quadtree.cpp
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	create_leaf(x * 2, y * 2, z - 1, node->nodes[index].get());
+}
+
+} // namespace
+
+void tfs::quadtree::find(uint16_t start_x, uint16_t start_y, uint16_t end_x, uint16_t end_y, SpectatorVec& spectators,
+                         std::function<bool(Creature*S)> comparasion)
```
ranisalt, line 61: ~~Comparación~~ `comparison`
```suggestion
                         std::function<bool(Creature*)> comparison)
```

Or just `cmp`

src/quadtree.cpp
```diff
@@ -0,0 +1,199 @@
...
+		players.insert(player);
+	}
+}
+
+void Leaf::remove_creature(Creature* creature)
+{
+	auto it = creatures.find(creature);
+	if (it != creatures.end()) {
+		creatures.erase(it);
+	}
```
ranisalt, line 181: No need to check for `erase`, if the key is not present it does nothing

src/quadtree.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace tfs::quadtree {
+void find(uint16_t start_x, uint16_t start_y, uint16_t end_x, uint16_t end_y, SpectatorVec& spectators,
+          std::function<bool(Creature*)> comparasion);
+Tile* find_tile(uint16_t x, uint16_t y, uint8_t z);
+void create_tile(uint16_t x, uint16_t y, uint8_t z, Tile* tile);
+void move_creature(uint16_t old_x, uint16_t old_y, uint8_t old_z, uint16_t x, uint16_t y, uint8_t z,
+                   Creature* creature);
+void insert_creature(uint16_t x, uint16_t y, uint8_t z, Creature* creature);
+void remove_creature(uint16_t x, uint16_t y, uint8_t z, Creature* creature);
```
ranisalt, line 14: Can we use `Position` rather than 3 ints for xyz?

src/quadtree.cpp
ramon-bernardo: Changed.

src/quadtree.h
ramon-bernardo: Thanks

src/quadtree.cpp
ramon-bernardo: Changed.

src/quadtree.cpp
ramon-bernardo: I was thinking of something like `std::function<bool(T)>` and `quadtree::find<T*>(...)`. I did some tests, but I will leave it for after we finish base quadtree.

src/quadtree.cpp
ramon-bernardo: No problem, I kept the search as it was a `vector`, but I changed it to a `set`. ;)

src/quadtree.h
ramon-bernardo: Initially, I used pos, but to maintain compatibility with the `Map` class, I used x, y, and z. I would prefer if `Map` were also changed to support only Position. See https://github.com/otland/forgottenserver/blob/244465537b57230aa6824b390d979d2fe1a8eec2/src/map.h#L177-L190

vc17/theforgottenserver.vcxproj
```diff
@@ -112,6 +112,7 @@
       <EnableEnhancedInstructionSet>AdvancedVectorExtensions</EnableEnhancedInstructionSet>
       <AdditionalIncludeDirectories>$(VcpkgRoot)include\luajit;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <LanguageStandard>stdcpp20</LanguageStandard>
+      <PrecompiledHeader>NotUsing</PrecompiledHeader>
```
ramon-bernardo, line 115: Remove this before merge ;)

ramon-bernardo:
We could rename the namespace to namespace `tfs::map::quadtree` and later change the Map class to namespace `tfs::map`. I think it's a good move, along with namespace `tfs::map::astar`.

Zbizu:
suggestion: add extra layer for Z coordinate in order to support up to 256 floors

this can still be regulated by MAP_MAX_LAYERS, but would allow otc users to break 16 floors limit more easily

ramon-bernardo:
@Zbizu I didn't understand your suggestion. Are you proposing to have a configuration for `MAP_LAYER`? If so, we can do that when we move the `Map` class to a namespace. ;)

ramon-bernardo:
Any suggestions for `std::generator`?

Zbizu:
@ramon-bernardo
as far as I know, the current quad tree implementation in tfs doesn't allow more than 16 floors, my suggestion is to make it possible to allocate either full u8 range (256 floors), or full u16 range (65535 floors) in Z coordinate, as in X and Y axes

you don't have to refactor entire Position class, supporting the floors in the new QTree (or even octree if you decide to use u16 for Z coordinate) will be good enough for now, full support for more floors (otc only) can be added in later commits

ramon-bernardo:
@Zbizu can we leave this for another PR? I'm considering using to `std::generator`, but we don't have support for it yet

Zbizu:
@ramon-bernardo of course, this one is quite big already

src/map.cpp
```diff
@@ -470,8 +352,37 @@ void Map::getSpectators(SpectatorVec& spectators, const Position& centerPos, boo
...
+			}
+
+			const auto offsetZ = centerPos.getOffsetZ(position);
+			if ((min_y + offsetZ) > position.y || (max_y + offsetZ) < position.y || (min_x + offsetZ) > position.x ||
+			    (max_x + offsetZ) < position.x) {
+				continue;
+			}
+
+			spectators.emplace_back(creature);
+		}
```
ramon-bernardo, line 385: @ranisalt  can you implement `C++24` in forgotten? This would avoid push twice into the `SpectatorVec` 🥳

ranisalt: We're still not there yet :/

PR #4806, opened 2024-10-07
Fix condition light crash

### LINKED ISSUE #4782: [Bug]: Condition light with light level < 1 and >=0 will crash server

### Does this bug crash tfs?

yes

### Server Version

1.7 (Master) 1.4.2 and others

### Operation System

Windows, Ubuntu 20.04, Ubuntu 22.04

### Bug description

Server crashes with condition light that has level set to 0

_No response_

### Steps to reproduce

1. Create Light Condition
2. Set CONDITION_PARAM_LIGHT_LEVEL param to value >= 0 and < 1
3. add condition to player
**Code to reproduce the issue:**
```lua
local condition = Condition(CONDITION_LIGHT)
condition:setParameter(CONDITION_PARAM_LIGHT_COLOR, 215)
condition:setParameter(CONDITION_PARAM_TICKS, 10000)
condition:setParameter(CONDITION_PARAM_LIGHT_LEVEL, 0.2)
player:addCondition(condition)
```
It will crash at: https://github.com/otland/forgottenserver/blob/416df1912872d406e2cef4ef2ff57f9373cac7bf/src/condition.cpp#L1744
because of division by 0

### Actual Behavior

Server crashes

### Expected Behavior

Server do not crash ^^
maybe warning on setting condition param?

### Backtrace

_No response_

nekiro:
### Pull Request Prelude

### Changes Proposed
Ensure light level is always 1, optionally warn user about incorrect value being set

**Issues addressed:** 
Fix #4782

### DIFF
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -1779,7 +1779,7 @@ void ConditionLight::addCondition(Creature* creature, const Condition* condition
 		const ConditionLight& conditionLight = static_cast<const ConditionLight&>(*condition);
 		lightInfo.level = conditionLight.lightInfo.level;
 		lightInfo.color = conditionLight.lightInfo.color;
-		lightChangeInterval = ticks / lightInfo.level;
+		lightChangeInterval = ticks / std::max<uint8_t>(1, lightInfo.level);
 		internalLightTicks = 0;
 		creature->setCreatureLight(lightInfo);
 		g_game.changeLight(creature);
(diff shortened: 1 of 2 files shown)


EPuncker (approved):

src/condition.cpp
```diff
@@ -1794,9 +1794,14 @@ bool ConditionLight::setParam(ConditionParam_t param, int32_t value)
...
 
 	switch (param) {
-		case CONDITION_PARAM_LIGHT_LEVEL:
-			lightInfo.level = value;
+		case CONDITION_PARAM_LIGHT_LEVEL: {
+			if (value < 1) {
+				std::cout << "[ConditionLight::setParam] trying to set invalid light value: " << value
+				          << " defaulting to 1" << std::endl;
+			}
+			lightInfo.level = std::max(1, value);
```
Codinablack, line 1802: since lightinfo.level is a uint8_t and value is an int32_t, maybe we should go ahead and add some safety here and use clamp instead?

Codinablack (changes requested):

ramon-bernardo:
Could there also be validation in the lua core? https://github.com/otland/forgottenserver/blob/624e8884497cbdf8c1a56ae55e3eddedd55120e9/data/lib/core/game/world_light.lua#L18-L21

nekiro:
> Could there also be validation in the lua core?
> 
> https://github.com/otland/forgottenserver/blob/624e8884497cbdf8c1a56ae55e3eddedd55120e9/data/lib/core/game/world_light.lua#L18-L21

nah, it will still crash for the special people that like to break stuff, if source crashes it should be fixed in source imo

PR #4812, opened 2024-10-18
Correct misuses of `__has_cpp_attribute`

CaseyCarter:
### Pull Request Prelude

### Changes Proposed

`__has_cpp_attribute` is for detecting support for attribute-tokens like `fallthrough` or `unlikely`. These queries with feature-test macro names are always going to fail since no implementation will ever support an attribute-token with the same name as a feature-test macro. Given the conditional use of library facilities, the intent here is clearly to test the feature-test macros to determine support for those facilities, so let's do that instead.

I've also fixed the using-declaration for `std::to_underlying`.

Drive-by: Removed extraneous `inline` in the definition of the `constexpr` fallback for `to_underlying`; `constexpr` functions are implicitly inline.

**Issues addressed:** None

### DIFF
--- a/src/tools.h
+++ b/src/tools.h
@@ -77,18 +77,17 @@ const std::vector<Direction>& getShuffleDirections();
 
 namespace tfs {
 
-#if __has_cpp_attribute(__cpp_lib_to_underlying)
+#ifdef __cpp_lib_to_underlying
 
-template <class T>
-using std::to_underlying<T>;
+using std::to_underlying;
 
 #else
 
-inline constexpr auto to_underlying(auto e) noexcept { return static_cast<std::underlying_type_t<decltype(e)>>(e); }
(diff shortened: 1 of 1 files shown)


src/tools.h
```diff
@@ -77,18 +77,17 @@ const std::vector<Direction>& getShuffleDirections();
...
 
 #else
 
-inline constexpr auto to_underlying(auto e) noexcept { return static_cast<std::underlying_type_t<decltype(e)>>(e); }
+constexpr auto to_underlying(auto e) noexcept { return static_cast<std::underlying_type_t<decltype(e)>>(e); }
 
 #endif
 
-#if __has_cpp_attribute(__cpp_lib_unreachable)
+#ifdef  __cpp_lib_unreachable
```
CaseyCarter, line 90: Typo:
```suggestion
#ifdef __cpp_lib_unreachable
```

ranisalt:
This part of the code has been removed, but thanks for the fix! Sorry for not seeing it earlier, it wasn't affecting existing builds

PR #4814, opened 2024-10-24
New methods to handle creatures with a master

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Introduced methods to enhance the `Creature` class, allowing for better handling of creature ownership and master relationships.
- Removed the static methods `Combat::isPlayerCombat` and `Player::lastHitIsPlayer`.
- New getter methods for retrieving the respective owners: `getPlayerOwned()`, `getNpcOwned()` and `getMonsterOwned()`. Methods to determine if the creature is owned, either directly or through its master. 
- Methods like `isPlayerSummon()`, `isNpcSummon()` and `isMonsterSummon()` to check the type of creature master.
 
#### Usage
c++
```c++
// Check if the creature is a summon.
if (creature->isSummon()) {
    std::cout << "This creature is a summon." << std::endl;
}

// Retrieve the player that owns this creature.
if (Player* owner = creature->getPlayerOwned()) {
    std::cout << "This creature is owned by player: " << owner->getName() << std::endl;
}
```

lua
```lua
if creature:isSummon() then
    print("This creature is a summon.")
end

local owner = creature:getPlayerOwned()
if owner then
    print("This creature is owned by player: " .. owner:getName())
end
```

### DIFF
--- a/data/creaturescripts/scripts/drop_loot.lua
+++ b/data/creaturescripts/scripts/drop_loot.lua
@@ -6,19 +6,7 @@ function onDeath(player, corpse, killer, mostDamageKiller, lastHitUnjustified, m
 	local amulet = player:getSlotItem(CONST_SLOT_NECKLACE)
 	local isRedOrBlack = table.contains({SKULL_RED, SKULL_BLACK}, player:getSkull())
 	if amulet and amulet.itemid == ITEM_AMULETOFLOSS and not isRedOrBlack then
-		local isPlayer = false
-		if killer then
-			if killer:isPlayer() then
-				isPlayer = true
-			else
-				local master = killer:getMaster()
-				if master and master:isPlayer() then
-					isPlayer = true
-				end
-			end
-		end
-
-		if not isPlayer or not player:hasBlessing(6) then
+		if not killer or not killer:hasPlayerOwned() or not player:hasBlessing(6) then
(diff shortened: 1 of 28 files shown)


MillhioreBT:
These changes are supposed to clean up the code and make it more readable, but it is quite the opposite, in fact it adds 200 more lines of code that are not necessary. (Especially in C++), in Lua it would be as easy as making a function that calls another and that's it and it would be acceptable

Basically what we would be avoiding with these changes is simply not checking null or nil of the master and then knowing if the master is one type or another.

I think everyone agrees that it is nicer to write it as you would normally do:

```lua
local master = creature:getMaster()

if not master then
	return
end

if master:isPlayer() then
	return
end
```

```lua
local master = creature:getMaster()
if master and master:isPlayer() then
	return
end
```

or in any case if we wish we can create our custom methods
```lua
function Creature.getPlayerOwned(self)
	local owner = self:getMaster()
	if owner and owner:isPlayer() then
		return owner
	end
end
```

ramon-bernardo:
@MillhioreBT 

Perhaps we could improve the function names; 'Owned' seems confusing at first. The validation appears to check if something is 'controlled by a player,' whether it's directly a player or a creature summoned by.

LOC here doesn't equate to quality, as it's `luascript` (15k LOL)

ramon-bernardo:
However, I agree that much of luascript could be impl in lua (Creature)

PR #4817, opened 2024-10-30
CreatureEvents improvement: avoid registering with similar names, enforce lowercase in all cases

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- CreatureEvents improvement: avoid registering with similar names, enforce lowercase in all cases. `PlayerLogin` and `playerlogin` are the same event, preventing issues with `creature:registerEvent("name")`.

### DIFF
--- a/src/creatureevent.cpp
+++ b/src/creatureevent.cpp
@@ -62,7 +62,7 @@ bool CreatureEvents::registerEvent(Event_ptr event, const pugi::xml_node&)
 	}
 
 	// if not, register it normally
-	creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
+	creatureEvents.emplace(boost::algorithm::to_lower_copy(creatureEvent->getName()), std::move(*creatureEvent));
 	return true;
 }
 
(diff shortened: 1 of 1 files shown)


EPuncker (approved):

src/creatureevent.cpp
```diff
@@ -62,7 +62,7 @@ bool CreatureEvents::registerEvent(Event_ptr event, const pugi::xml_node&)
 	}
 
 	// if not, register it normally
-	creatureEvents.emplace(creatureEvent->getName(), std::move(*creatureEvent));
```
ramon-bernardo, line 65: Maybe we don't need this, since queries will always be in lowercase 🤗

MillhioreBT (approved):
I really like it better this way, it makes sense.

src/creatureevent.cpp
MillhioreBT: Find will always be in lowercase, but if the event is logged in uppercase then you won't be able to find it.

Another option is to set the name to lowercase when creating the CreatureEvent, so getting the name with your getName method would always return the name in lowercase.

PR #4828, opened 2024-11-07
Fix Container::hasParent

ramon-bernardo:
### Pull Request Prelude
- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Fix https://github.com/otland/forgottenserver/pull/4810
Close https://github.com/otland/forgottenserver/pull/4826, https://github.com/otland/forgottenserver/pull/4835

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4810, opened 2024-10-15
Add a new method to check if Thing has a parent

ramon-bernardo:
### Pull Request Prelude
- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Introduces a new method to the `Thing` class that allows checking if a Thing instance has a parent. This providing a direct way to determine parent-child relationships, in addition to the existing `if (getParent()) {}`.

Before
```c++
Tile* tile = nullptr;
if (!parent->getParent()) {
    tile = parent->getTile();
}
```

After
```c++
Tile* tile = nullptr;
if (!parent->hasParent()) {
    tile = parent->getTile();
}
```

New Lua method
```lua
local container = Container()
if container:hasParent() then
    ...
end
```

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4835, opened 2024-11-16
Revert "Add a new method to check if Thing has a parent (#4810)"

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
This causes the containers or cylinders in general to stop working and causes a crash.
It would be better if @ramon-bernardo solves it, if not, it is simply better to revert it

https://github.com/user-attachments/assets/cd2d9396-e733-4b96-b7c4-1f27d7537057


ramon-bernardo:
Fixed https://github.com/otland/forgottenserver/pull/4828

PR #4840, opened 2024-11-17
Move players record to lua

ramon-bernardo:
### Pull Request Prelude

### Changes Proposed
- Move players record to lua.
- Saves the record only when the save.
- Remove globalevent onRecord.

### Test required
- I added a `ScriptInterface` to `Game`, but I am unsure of its limitations. It’s necessary to check whether what I’ve implemented is correct, including `Game::getPlayersRecord()`

### DIFF
--- a/data/cpplinter.lua
+++ b/data/cpplinter.lua
@@ -1048,7 +1048,6 @@ MoveEvent = {}
 ---@field interval fun(self:GlobalEvent, interval:integer):boolean
 ---@field onStartup fun():boolean
 ---@field onShutdown fun():boolean
----@field onRecord fun(current:integer, old:integer):boolean
 ---@field onThink fun(interval:integer):boolean
 ---@field onTime fun(interval:integer):boolean
 ---@operator call(string):GlobalEvent
--- a/data/globalevents/globalevents.xml
+++ b/data/globalevents/globalevents.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <globalevents>
 	<globalevent type="startup" name="ServerStartup" script="startup.lua" />
-	<globalevent type="record" name="PlayerRecord" script="record.lua" />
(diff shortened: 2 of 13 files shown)


MillhioreBT:
Horrible, having to instantiate a pointer to a lua interface in the Game class just to get a variable from lua.
More code, same functionality, less efficient, harder to maintain.

First of all, it would be easier to have a method that accepts a `luaState* L` and use it to interact with lua.
Secondly, you can obtain the number of players with the `Game.getPlayerCount` method.
Thirdly you are not freeing the results table from the database when using storeQuery

MillhioreBT (changes requested):
Horrible, having to instantiate a pointer to a lua interface in the Game class just to get a variable from lua.
More code, same functionality, less efficient, harder to maintain.

First of all, it would be easier to have a method that accepts a luaState* L and use it to interact with lua.
Secondly, you can obtain the number of players with the Game.getPlayerCount method.
Thirdly you are not freeing the results table from the database when using storeQuery

ramon-bernardo:
@MillhioreBT I agree—if only we had a status server in Lua or something like an Http Client, we wouldn’t need to call Lua in Game.cpp.

- However, I disagree regarding performance. Unlike before, where an UPDATE was performed every time a player joined, now it only occurs when save the game.

- I didn’t recall `Game.getPlayerCount()`. I'll make the adjustment!

- If I’m not mistaken, it’s no longer necessary to call `result.free`. This is because of the automatic cleanup we perform after every Lua call, using `tfs::lua::resetScriptEnv()`. However, I can add it explicitly if needed.

I’m not a fan of `Game::Game() : scriptInterface("Game Interface")`. If anyone has a better idea, im open to suggestions 🫤

ranisalt:
Just keep a running counter of players online that increases with a creaturescript of type `login` and decreases with a `logout`

NRH-AA:
Wasn't this already a thing back in like 1.2 1.0 and it got removed?

PR #4844, opened 2024-11-19
Fix reload XML files from lua

ramon-bernardo:
### Pull Request Prelude

Fix the problem with XML reloading via Lua; when a reload occurred, the Lua state was reset, but the XML file data was not reloaded correctly.

### Changes Proposed 
- It no longer loads during the start event but instead when the Lua files are loaded
- Move it to a separate folder in `data/scripts/xml`.

### On reload
image

### DIFF
--- a/data/scripts/xml/#quests.lua
+++ b/data/scripts/xml/#quests.lua
@@ -1,15 +1,13 @@
 -- If you don't intend to use quests.xml, you can delete this file.
-local xmlQuest = GlobalEvent("Load XML Quests")
 
-function xmlQuest.onStartup()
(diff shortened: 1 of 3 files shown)


ramon-bernardo:
The other alternative would be to create a [new global event](https://github.com/ramon-bernardo/forgottenserver/commit/f1a116289d9ce62664ea0f61c4a38cb91b590fb7#diff-839663fc63a5de75aa08d38518c2d625e87a139c1a42b027bcdd363e83f063a2) for handling server reloads. However, some points to consider are when to trigger it: only when reloading scripts? actions? ...

EPuncker (approved):

ranisalt (approved):

Issue #4841, opened 2024-11-18
[Bug]: using interval with onTime no longer work


Sajgon:
### By submitting this bug issue, you agree to the following.

- [X] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [X] This issue is reproducible without changes to the C++ code in this repository
- [X] This bug has not been resolved in master branch
- [X] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below)

### Bug description

using interval with onTime no longer work

### Possible Pull Requests which are to blame

- [ ] # https://github.com/otland/forgottenserver/commit/46e2b7a47fdf4a2d6e39ac4a76018398a903d307#diff-9cab8e66cea1067bf36772fb84b9bbd73f55b26137be07de21c9cf046d028870R240

### Steps to reproduce

Example:
1. In the script world_light.lua see the value of interval (10000), which equals to 10 seconds
2. The script should run every 10 seconds, but isn't
3. The reason is because onTime has a type in compat.lua, the interval isn't set. See the code below, if type is used interval isn't.

```lua
} else if ((attr = node.attribute("type"))) {
		const char* value = attr.value();
		if (caseInsensitiveEqual(value, "startup")) {
			eventType = GLOBALEVENT_STARTUP;
		} else if (caseInsensitiveEqual(value, "shutdown")) {
			eventType = GLOBALEVENT_SHUTDOWN;
		} else if (caseInsensitiveEqual(value, "record")) {
			eventType = GLOBALEVENT_RECORD;
		} else if (caseInsensitiveEqual(value, "save")) {
			eventType = GLOBALEVENT_SAVE;
		} else {
			std::cout << "[Error - GlobalEvent::configureEvent] No valid type \"" << attr.as_string()
			          << "\" for globalevent with name " << name << std::endl;
			return false;
		}
	} else if ((attr = node.attribute("interval"))) {
		interval = std::max<int32_t>(SCHEDULER_MINTICKS, pugi::cast<int32_t>(attr.value()));
		nextExecution = OTSYS_TIME() + interval;
	} 
```

**Code to reproduce the issue:**
```lua
function event.onTime(interval)
	console.log("🚀 ~ WorldLight onTime:", onTime)
	if not defaultWorldLight then
		return true
	end

	local worldLightColor, worldLightLevel = Game.getWorldLight()
[...]

PR #4846, opened 2024-11-20
fix onTime event

ArturKnopik:
### Pull Request Prelude

### Changes Proposed
fix onTime event

**Issues addressed:** 
https://github.com/otland/forgottenserver/issues/4841

### DIFF
--- a/src/globalevent.cpp
+++ b/src/globalevent.cpp
@@ -138,12 +138,8 @@ void GlobalEvents::timer()
 			continue;
 		}
 
-		nextExecutionTime = 86400000;
-		if (nextExecutionTime < nextScheduledTime) {
-			nextScheduledTime = nextExecutionTime;
-		}
-
-		globalEvent.setNextExecution(globalEvent.getNextExecution() + nextExecutionTime);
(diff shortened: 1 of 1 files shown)


src/globalevent.cpp
```diff
@@ -138,12 +138,8 @@ void GlobalEvents::timer()
 			continue;
 		}
 
-		nextExecutionTime = 86400000;
```
ArturKnopik, line 141: I don't know what it was for, maybe there's something I don't know about

ArturKnopik:
Additional note:
https://github.com/otland/forgottenserver/blob/14ba662ad6d6ad2d7dc0439fc2350f816f4c11e2/src/luascript.cpp#L18246
in lua we are using old style to retrieve time**(time(nullptr)** instead **OTSYS_TIME()**)

ramon-bernardo (approved):
Looks good, to test just use:

```
local event = GlobalEvent("4846")

function event.onTime(interval)
	Game.broadcastMessage("1 second...", MESSAGE_STATUS_WARNING)
	return true
end

event:interval(1000)
event:register()
```

EPuncker (approved):

PR #4854, opened 2024-11-24
Configurable stack size per ItemType

ArturKnopik:
### Pull Request Prelude

### Changes Proposed
Configurable stack size per ItemType.
Configuration is done in items.xml file by using attribue "stacksize"
Example:
```XML
<item id="2316" article="an" name="animate dead rune">
	<attribute key="type" value="rune" />
	<attribute key="runeSpellName" value="adana mort" />
	<attribute key="weight" value="210" />
	<attribute key="charges" value="1" />
	<attribute key="stacksize" value="2" /> <!--  <<< HERE-->
</item>
```

note: npc now can sell multiple bp's with stackable items.

**Issues addressed:** 
N/A

**How to test:** 
Set 'stacksize' in any stackable item
- Collect items from monsters
- Move items in BP
- Move items to Arrow/Weapon slot
- Buy items from NPC


src/luascript.cpp
```diff
@@ -13685,6 +13686,18 @@ int LuaScriptInterface::luaItemTypeGetWorth(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaItemTypeGetStackSize(lua_State* L)
+{
+	// itemType:getStackSize()
```
ramon-bernardo, line 13691: Suggestion
```
itemType:stackSize(value) // set
itemType:stackSize() // get
```

ArturKnopik: it's used only in shop system
setting stack size should be done only on startup, your change reguest will affect ItemType that is and should be static

ramon-bernardo: You're right, I hadn't seen that 😲

nekiro: not a fan of this "revscriptsys" way of getters and setters, make it obvious getAnything, setAnything
anything() is confusing and may result in undefined behaiour if you by any chance pass value or actually pass no value, which could happen, because lua is a dynamic language

src/game.cpp
```diff
@@ -1593,7 +1593,7 @@ void Game::addMoney(Cylinder* cylinder, uint64_t money, uint32_t flags /*= 0*/)
 
 		money -= currencyCoins * worth;
 		while (currencyCoins > 0) {
-			const uint16_t count = std::min<uint32_t>(ITEM_STACK_SIZE, currencyCoins);
+			const uint16_t count = std::min<uint32_t>(Item::items[it.second].stackSize, currencyCoins);
```
MillhioreBT, line 1596: I think it’s better to define a temporary variable to store the `stackSize`, then reuse it in the loop. Otherwise, what you’re doing is accessing `it` and getting a reference to the `itemtype`, then accessing the `itemtype` and obtaining a copy of the number in each iteration.

Issue #4839, opened 2024-11-17
[Bug]: Closing door delete items on the ground


Feeltz:
### By submitting this bug issue, you agree to the following.

- [X] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [X] This issue is reproducible without changes to the C++ code in this repository
- [X] This bug has not been resolved in master branch
- [X] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master), 1.4.x (Release)

### Operation System

all (listed below)

### Bug description

Auto closing door can delete item on the ground

### Steps to reproduce

Best place to test it is on the real map is on Darashia Mino Tower
![obraz](https://github.com/user-attachments/assets/7d9805be-c859-4686-a1bf-cb5af33d7614)

1. Stand on auto closing door, that points downwoards, and the mailbox has to be below you (y+1)
2. put items on the ground, under your character
3. step out north, to trigger auto-close door

**Code to reproduce the issue:**
This part of the closing_door.lua is responsible for deleting items
```lua
local i, tileItem, tileCount = 1, true, tile:getThingCount()
while tileItem and i < tileCount do
        tileItem = tile:getThing(i)
        if tileItem and tileItem:getUniqueId() ~= item.uid and tileItem:getType():isMovable() then
	        tileItem:remove()
        else
	        i = i + 1
        end
end
```
https://github.com/otland/forgottenserver/blob/173ab186e0d218d69acff746a2a795f6dcb7b323/data/scripts/movements/closing_door.lua#L11

It can be corelated with this issue: https://github.com/otland/forgottenserver/issues/4467 partialy fixed in here https://github.com/otland/forgottenserver/pull/4468

### Actual Behavior

Items disappear. If you die on door SQM, your corpse will disappear. The problem is if the mailbox (or other solid item) is south or east from door. It should not happen because normally you can throw items on the mailbox.

### Expected Behavior

Items are moved on the top of the mailbox or pushed oposite way if instead of mailbox is something like wall.

Diego-OT:
Instead of remove, just move it.

https://github.com/otland/forgottenserver/blob/14ba662ad6d6ad2d7dc0439fc2350f816f4c11e2/data/scripts/movements/closing_door.lua#L25

tileItem:moveTo(newPosition)

PR #4856, opened 2024-11-24
fix item destroy by doors

ArturKnopik:
### Pull Request Prelude

### Changes Proposed
Fix items damaged by doors

**Issues addressed:**
https://github.com/otland/forgottenserver/issues/4839

**How to test:** 
Use doors in position 121, 83, 10
Create item 1787 like on image
image

Open doors
Put items under player
Move back from doors

### DIFF
--- a/data/scripts/movements/closing_door.lua
+++ b/data/scripts/movements/closing_door.lua
@@ -22,7 +22,7 @@ function closingDoor.onStepOut(creature, item, position, fromPosition)
 	while tileItem and i < tileCount do
 		tileItem = tile:getThing(i)
 		if tileItem and tileItem:getUniqueId() ~= item.uid and tileItem:getType():isMovable() then
-			tileItem:remove()
+			doRelocate(position, creature:getPosition())
 		else
 			i = i + 1
 		end


data/scripts/movements/closing_door.lua
```diff
@@ -22,7 +22,7 @@ function closingDoor.onStepOut(creature, item, position, fromPosition)
 	while tileItem and i < tileCount do
 		tileItem = tile:getThing(i)
 		if tileItem and tileItem:getUniqueId() ~= item.uid and tileItem:getType():isMovable() then
```
ArturKnopik, line 24: can't use position or fromPosition from function parameters, they returns always the same positions, creature position is good becouse it's should be always valid

data/scripts/movements/closing_door.lua
```diff
@@ -22,7 +22,7 @@ function closingDoor.onStepOut(creature, item, position, fromPosition)
...
-			tileItem:remove()
```
ArturKnopik, line 25: I can't use position or fromPosition from function parameters, they returns always the same positions, creature position is good becouse it's should be always valid

data/scripts/movements/closing_door.lua
```diff
@@ -22,7 +22,7 @@ function closingDoor.onStepOut(creature, item, position, fromPosition)
...
+			doRelocate(position, creature:getPosition())
```
ArturKnopik, line 25: I can't use position or fromPosition from function parameters, they returns always the same positions, creature position is good becouse it's should be always valid

ramon-bernardo:
Could you create Position::relocate and avoid using doRelocate

```lua
-- data\lib\core\position.lua
function Position:relocate(toPos)
	if self == toPos then
		return false
	end

	local fromTile = Tile(self)
	if not fromTile then
		return false
	end

	if not Tile(toPos) then
		return false
	end

	for i = fromTile:getThingCount() - 1, 0, -1 do
		local thing = fromTile:getThing(i)
		if thing then
			if thing:isItem() then
				if ItemType(thing:getId()):isMovable() then
					thing:moveTo(toPos)
				end
			elseif thing:isCreature() then
				thing:teleportTo(toPos)
			end
		end
	end
	return true
end
```

```lua
position::relocate(creature:getPosition())
```

nekiro:
this needs major rewriting imo, items should be moved and if cannot be moved then deleted

PR #4860, opened 2024-12-01
OTCv8 proxy + HAProxy

gesior:
### Pull Request Prelude

Code to read real connection IP from OTCv8 proxy and HAProxy `send-proxy-v2` protocols.
Changes also allow owner to edit server IP showed in "server status protocol", so it's possible to add different VPS IP - with `haproxy` running on 7171 - as OTS IP to otservlist etc. = hide real server IP from DDoSers.

Goal is to hide IP on which is running OTS, so it can't be DDoSed.
With OTCv8 packets to/from server will go thru multiple proxy VPSes, so you can use multiple anti-DDoS services in same time and as long as any of them keep one proxy online, OTS will be online and does not lag.

### Changes Proposed

OTCv8 proxy is proxy system available in: OTCv8, OTCR (Mehah) and real Tibia Client (with extra .exe, more info: https://github.com/mehah/otclient/pull/978#issuecomment-2509871565 )

## How OTCv8 proxy works

OTC with Proxy System enabled will connect to multiple `haproxy` servers - each can be on other VPS behind other anti-DDoS system.
Each second it measures ping to OTS thru each `haproxy` - it measures real ping to OTS machine, not just ping to VPS hosting `haproxy`.
If there is any packet to send, OTC sends it thru 2 proxies with lowest ping (duplicates packet).

List of proxy server you must add in `init.lua` of OTC ex.:
```
g_proxy.addProxy('arm.skalski.pro', 6501, 0)
g_proxy.addProxy('arm.skalski.pro', 6502, 0)
g_proxy.addProxy('arm.skalski.pro', 6503, 0)
g_proxy.addProxy('arm.skalski.pro', 6504, 0)
```
Then every connection to `127.0.0.1` will be handled by Proxy System and send thru haproxy to OTS server machine.
On machine with OTS you must install "OTCv8 Proxy System server side app" (re-host: https://downloads-oracle.ots.me/?dir=data/otcv8-proxy-server).
This server side app will detect duplicated packets and send only first packet that comes with given ID to OTS.
Packets send by OTS will also go thru this app, it will duplicate them and send thru 2 `haproxy` servers to OTC.
Server side app is compatible with `send-proxy-v2` protocol and may send real IP of player to OTS - without that feature all players would have IP `127.0.0.1` on OTS, as connections to OTS come from server side proxy app.

How to configure server/OTCv8/haproxy by OTCv8 author:
http://web.archive.org/web/20201109120714/http://otclient.net/showthread.php?tid=438

More info about configuration and other apps required:
https://github.com/mehah/otclient/pull/978

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -168,3 +168,18 @@ ownerName = ""
 ownerEmail = ""
 url = "https://otland.net/"
 location = "Sweden"
+
+-- Proxy settings
+-- NOTE:
+-- if you allow any proxy connections, you should configure your firewalls to not allow direct connections to OTS
+-- only your own HAProxy server IP addresses should be allowed to connect to OTS ports / OTCv8 proxy server ports,
+-- otherwise anyone can set up their own modified HAProxy server and spoof their in-game IP
+
+-- allow OTCv8 proxy connections
+allowOtcProxy = false
+-- allow HAProxy connections - for players or for server status protocol ex. otservlist
+allowHaProxy = false
+-- if you allow HAProxy connections and want to add the HAProxy VPS IP address as OTS IP address to otservlist etc.,
+-- here you can set IP address that will be returned in OTS status as the IP address of the server
+-- empty string = return actual IP address of the server
+statusIp = ""
--- a/src/http/login.cpp
+++ b/src/http/login.cpp
@@ -125,13 +125,20 @@ std::pair<status, json::value> tfs::http::handle_login(const json::object& body,
 		} while (result->next());
 	}
 
+	std::string serverIp = getString(ConfigManager::IP);
+	if (getBoolean(ConfigManager::ALLOW_OTC_PROXY)) {
+		serverIp = std::string("127.0.0.1");
+	} else if (getBoolean(ConfigManager::ALLOW_HAPROXY)) {
+		serverIp = getString(ConfigManager::STATUS_IP);
+	}
+
 	json::array worlds{
 	    {
 	        {"id", 0}, // not implemented
 	        {"name", getString(ConfigManager::SERVER_NAME)},
-	        {"externaladdressprotected", getString(ConfigManager::IP)},
+	        {"externaladdressprotected", serverIp},
 	        {"externalportprotected", getNumber(ConfigManager::GAME_PORT)},
-	        {"externaladdressunprotected", getString(ConfigManager::IP)},
+	        {"externaladdressunprotected", serverIp},
 	        {"externalportunprotected", getNumber(ConfigManager::GAME_PORT)},
 	        {"previewstate", 0}, // not implemented
 	        {"location", getString(ConfigManager::LOCATION)},
--- a/data/cpplinter.lua
+++ b/data/cpplinter.lua
@@ -349,6 +349,8 @@ Creature = {}
 ---@field isPlayer fun(self: Player): boolean
 ---@field getGuid fun(self: Player): number
 ---@field getIp fun(self: Player): number
+---@field isOtcProxy fun(self: Player): boolean
+---@field isHaProxy fun(self: Player): boolean
 ---@field getAccountId fun(self: Player): number
 ---@field getLastLoginSaved fun(self: Player): number
 ---@field getLastLogout fun(self: Player): number
@@ -2202,6 +2204,8 @@ configKeys = {
 	MANASHIELD_BREAKABLE = 36,
 	CHECK_DUPLICATE_STORAGE_KEYS = 37,
 	MONSTER_OVERSPAWN = 38,
+	ALLOW_OTC_PROXY = 39,
+	ALLOW_HAPROXY = 40,
 
 	-- ConfigKeysString
 	MAP_NAME = 0,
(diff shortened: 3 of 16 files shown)


Zbizu:
how to determine which proxy to serve for a client that is using a webservice to login? (speaking of version 12+ vanilla client)

gesior:
> how to determine which proxy to serve for a client that is using a webservice to login? (speaking of version 12+ vanilla client)

Connection to webservice ex. `https://test.com/login.php` is not passed thru proxy. Only connections to `127.0.0.1` are passed thru OTC proxy. For webservice protection you can use CloudFlare or similar service.

For 12+ client you must use "proxy launcher app", which you can get ex. from giveria.com client.
`Giveria.exe` is launcher app made by OTCv8 author and `proxy.config` configures proxy servers and ports, it looks like this:
```
.\ClientFiles\bin\client.exe
7171 7172
proxy01.giveria.com proxy01.giveria.com 7162
proxy02.giveria.com proxy02.giveria.com 7162
proxyca.giveria.com proxyca.giveria.com 7162
proxyuk.giveria.com proxyuk.giveria.com 7162
proxyde.giveria.com proxyde.giveria.com 7162
proxy03.giveria.com proxy03.giveria.com 7162
proxy04.giveria.com proxy04.giveria.com 7162
(...)
```
This launcher will start server on localhost listening on ports 7171 and 7172 and then run `.\ClientFiles\bin\client.exe`, which is Tibia 13+ .exe.
All connections to these ports on localhost will be passed thru all proxy servers listed below.  "Launcher" is OTCv8 proxy client code running as standalone app. It will check ping to all servers every second, pick 2 servers with lowest ping and send packets thru them.

If you are asking about 'how to check which 2 proxy VPSes are used by OTC proxy' with Tibia vanilla client, it's probably impossible right now (except checking transfer per IP with WireShark or other network analysis tool).
There was an exe released by OTCv8 author that was a simple terminal app that showed "OTCv8 proxy launcher stats" similar to "Debug view" in OTCv8 module (how many packets/KB send/receive from each proxy IP), but it's exe and source code was lost.

Webservice has to return IP of OTS set to `127.0.0.1`. OTCv8 proxy takes control over all connections to `127.0.0.1` from OTC (or Launcher app). It sends their packets thru `haproxy` to OTCv8 proxy server app, which removes duplicated packets and "unpack" them to localhost network. So your connection to `127.0.0.1:7171` will be passed thru 2 proxies with lowest ping to OTS machine, "unpacked" and send to localhost to port `7171` (OTS must listen on 127.0.0.1 / 0.0.0.0[any IP] to allow "OTCv8 proxy server app" to connect to OTS).

gesior:
For anyone interested: TFS 1.4 version:
https://github.com/gesior/forgottenserver-gesior/commit/3de653a1c5096d64c6c66682bde8de332bd8ce9f
It's not clean 1.4, because it's based on my 1.4 branch that has some code updates from 1.6, to make it compile without errors on new Windows/Linuxes.

DSpeichert (commented):
Since "otcProxy" software/relay isn't widely available, I don't think it makes sense to merge it to this project. The PROXY protocol that HA Proxy uses, however, is an established standard and makes sense to add.

I would, however, change the boolean to a list of IPs/CIDRs that are allowed to send the PROXY header. This eliminates the potential for spoofing while testing proxies and still allowing direct connections.

src/http/login.cpp
```diff
@@ -125,13 +125,20 @@ std::pair<status, json::value> tfs::http::handle_login(const json::object& body,
 		} while (result->next());
 	}
 
+	std::string serverIp = getString(ConfigManager::IP);
+	if (getBoolean(ConfigManager::ALLOW_OTC_PROXY)) {
+		serverIp = std::string("127.0.0.1");
+	} else if (getBoolean(ConfigManager::ALLOW_HAPROXY)) {
+		serverIp = getString(ConfigManager::STATUS_IP);
+	}
```
DSpeichert, line 133: I think this makes it clear that only one of them should be set to true.
In addition, since the PROXY protocol is not limited to a trusted source, it should be noted that TFS should not be available publicly directly. Which is probably obvious in the context of using a proxy system but probably TFS should start rejecting connections without the PROXY header.

config.lua.dist
```diff
@@ -168,3 +168,18 @@ ownerName = ""
...
+-- Proxy settings
+-- NOTE:
+-- if you allow any proxy connections, you should configure your firewalls to not allow direct connections to OTS
+-- only your own HAProxy server IP addresses should be allowed to connect to OTS ports / OTCv8 proxy server ports,
+-- otherwise anyone can set up their own modified HAProxy server and spoof their in-game IP
+
+-- allow OTCv8 proxy connections
+allowOtcProxy = false
+-- allow HAProxy connections - for players or for server status protocol ex. otservlist
+allowHaProxy = false
```
DSpeichert, line 181: I'd squash these into one setting `enableProxy` and have it REQUIRE one of the two proxy protocols instead, making it obvious that direct connections shouldn't be used/allowed by firewall.

PR #4862, opened 2024-12-04
Fix use-after-free in Browse Field handler

diath:
Transforming an item placed on a `Tile` with Browse Field open on that tile will call `Tile::updateThing`, which will subsequently call `Tile::onUpdateTileItem`, which then ensures that the Browse Field's `Container` is in sync with the `Tile`, however, `Tile::updateThing` passes the same `Item` pointer as both the `oldItem` and `newItem` arguments to `Tile::onUpdateTileItem` which only coincidentally works for the majority of items whose flags do not happen to transition between being movable and not being movable. If you transform a movable item to a non-movable item, the second branch of the Browse Field handling in `Tile::onUpdateTileItem` will never execute, leaving the `Item` that has now been deleted in the Browse Field `Container`'s `itemlist` member, any further interaction with the item, or waiting for the Browse Field to expire will then result in use-after-free. This fixes the issue by passing a proper `oldItem` argument to detect such item state transitions. Unfortunately, this comes at the cost of `Item::clone` call, I'm not sure if there's a more efficient way to handle this.

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -26,8 +26,7 @@ Container::Container(Tile* tile) : Container(ITEM_BROWSEFIELD, 30, false, true)
 	TileItemVector* itemVector = tile->getItemList();
 	if (itemVector) {
 		for (Item* item : *itemVector) {
-			if ((item->getContainer() || item->hasProperty(CONST_PROP_MOVEABLE)) &&
-			    !item->hasAttribute(ITEM_ATTRIBUTE_UNIQUEID)) {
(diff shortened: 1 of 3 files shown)


nekiro:
Is this reproducible by any steps or does it happen randomly?

EPuncker (approved):

nekiro:
Can't reproduce, steps I tried:
1. Wrote a script that creates item on ground (moveable item - 2110) and after 500ms transformed it into moveable one (2784)
2. Item was removed from my browse field window and transformed on map
3. Nothing happens when I remove the item, no crashes

Also tried with decay system, made torch decay into non moveable item in 5 seconds, no crash either.
I noticed Tile::onUpdateTileItem nor Tile::updateThing is not even triggered for this action.

Issue #4865, opened 2024-12-04
[Bug]: Monster not following target


ArturKnopik:
### By submitting this bug issue, you agree to the following.

- [X] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [X] This issue is reproducible without changes to the C++ code in this repository
- [X] This bug has not been resolved in master branch
- [X] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

Windows

### Bug description

Monster not following target

### Possible Pull Requests which are to blame

Probably this one: https://github.com/otland/forgottenserver/pull/4811

### Steps to reproduce

1. Enter to game world using normal character
2. Find monster that attack(ex: cyclops)
3. Watch monster behavior

### Actual Behavior

Monster moving in random directions

### Expected Behavior

Monster follow player

### Backtrace

```cpp
n/a
```

PR #4866, opened 2024-12-04
fix monster attack behavior

ArturKnopik:
### Pull Request Prelude

### Changes Proposed

Fix monster attack behavior

**Issues addressed:** 
https://github.com/otland/forgottenserver/issues/4865

**How to test:** 
go to monsters using normal character
Troll - hit to red, let them fleeing
Cyclops
Hunter 
Rabbit 
IDK if there is monster with different walking mechanic

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -578,14 +578,13 @@ bool Monster::searchTarget(TargetSearchType_t searchType /*= TARGETSEARCH_DEFAUL
 void Monster::goToFollowCreature()
 {
 	if (!followCreature) {
-		onFollowCreatureComplete();
 		return;
 	}
 
 	FindPathParams fpp;
 	getPathSearchParams(followCreature, fpp);
 
-	if (getMaster() && (isFleeing() || fpp.maxTargetDist > 1)) {
+	if (!isSummon()) {
 		Direction dir = DIRECTION_NONE;
 
 		if (isFleeing()) {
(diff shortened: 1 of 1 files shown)


nekiro:
imo there is missing updateFollowCreaturePath(fpp); after this code so specifically:

https://github.com/otland/forgottenserver/blob/a78d2ed78fab73069d429e18d4929f77aaf9db5c/src/monster.cpp#L606
```
} else {
    updateFollowCreaturePath(fpp);
}
```
genuinely sad this code was merged, its broken at the first sight when **joining the game and checking**

ramon-bernardo:
@nekiro I don't see where this `update` might be missing, as redundancies were removed (**highlighted in red**), and the logic was simplified (**highlighted in yellow**)

ArturKnopik:
@nekiro @ramon-bernardo, can you test my fix?

src/monster.cpp
```diff
@@ -605,10 +612,14 @@ void Monster::goToFollowCreature()
 			startAutoWalk();
 		}
 	} else {
-		updateFollowCreaturePath(fpp);
```
nekiro, line 608: use updateFollowCreaturePath
its the exact duplicate of the code you pasted here, so there is no reason to duplicate the code again

src/monster.cpp
```diff
@@ -578,21 +578,28 @@ bool Monster::searchTarget(TargetSearchType_t searchType /*= TARGETSEARCH_DEFAUL
...
+	if (getMaster() && (isFleeing() || fpp.maxTargetDist > 1)) {
 		Direction dir = DIRECTION_NONE;
 
 		if (isFleeing()) {
 			getDistanceStep(followCreature->getPosition(), dir, true);
-		} else if (fpp.maxTargetDist > 1) {
+		} else { // maxTargetDist > 1
 			if (!getDistanceStep(followCreature->getPosition(), dir)) {
 				// if we can't get anything then let the A* calculate
-				updateFollowCreaturePath(fpp);
```
nekiro, line 595: same as previously, use this method as its the same code, so its okay

nekiro (changes requested):

ramon-bernardo:
Unfortunately, i cant test it because my cip client doesn't connect to login server. I could try using OTClient, but not sure if it support v13

nekiro (approved):

PR #4614, opened 2023-12-24
Charms

ArturKnopik:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Introduction of the charm system(offensive perks + dodge) 
image

https://github.com/otland/forgottenserver/issues/3871 issue

PR #4868, opened 2024-12-08
Features from protocol 12+ 

ArturKnopik:
# Feature from Tibia 12+
The content of this PR is not intended to be merged because there are voices in the community that the engine should be clean without unnecessary things such as things from the 12+ protocol.
Features are written in such a way as to modify the engine as little as possible.
`IMPORTANT - The features may not fully reflect how the features behaves on RL tibia, the code is intended to show how it can be implemented with all necessary network messages.`

# Features:
## Charms - to rewrite
Note: Only offensive charms and dodge
image
todo: 
1) Add missing charms

## Boostiary
Note: There may be a large portion of bosses missing
image
image
todo:
1) Fix drop items only for slot/boosted boss
2) Fix onDrop event when there is no corpse owner
3) Reroll cost per slot


ArturKnopik:
https://github.com/otland/forgottenserver/pull/4614 is closed and content is moved here.

PR #4882, opened 2024-12-24
Fix scheaduler leaks

ArturKnopik:
### Pull Request Prelude

### Changes Proposed

Fix scheduler memory leaks

**Issues addressed:** 

https://github.com/otland/forgottenserver/issues/4288

**How to test:** 

Requirements: Linux
1) Build project in debug mode
2) Run server using valgrind(valgrind --leak-check=full ./tfs)
3) Shutdown server
4) Looks for logs like (They shouldn't be there)
```
==44907== 56 bytes in 1 blocks are definitely lost in loss record 461 of 568
==44907==    at 0x4846FA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==44907==    by 0x19D5DA: createSchedulerTask(unsigned int, std::function<void ()>&&) (scheduler.cpp:65)
==44907==    by 0x2DC8B5: IOMarket::checkExpiredOffers() (iomarket.cpp:193)
==44907==    by 0x278E46: (anonymous namespace)::mainLoader(ServiceManager*) (otserv.cpp:248)
==44907==    by 0x2794D6: startServer()::{lambda()#1}::operator()() const (otserv.cpp:285)
==44907==    by 0x27B005: void std::__invoke_impl<void, startServer()::{lambda()#1}&>(std::__invoke_other, startServer()::{lambda()#1}&) (invoke.h:61)
==44907==    by 0x27AD34: std::enable_if<is_invocable_r_v<void, startServer()::{lambda()#1}&>, void>::type std::__invoke_r<void, startServer()::{lambda()#1}&>(startServer()::{lambda()#1}&) (invoke.h:111)
==44907==    by 0x27A772: std::_Function_handler<void (), startServer()::{lambda()#1}>::_M_invoke(std::_Any_data const&) (std_function.h:290)
==44907==    by 0x168943: std::function<void ()>::operator()() const (std_function.h:591)
==44907==    by 0x16097F: Task::operator()() (tasks.h:23)
==44907==    by 0x14DAD0: Dispatcher::threadMain() (tasks.cpp:37)
==44907==    by 0x2D7CA3: void std::__invoke_impl<void, void (Dispatcher::*)(), Dispatcher*>(std::__invoke_memfun_deref, void (Dispatcher::*&&)(), Dispatcher*&&) (invoke.h:74)
==44907== 
```

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -634,11 +634,11 @@ void Game::playerMoveThing(uint32_t playerId, const Position& fromPos, uint16_t 
 		}
 
 		if (movingCreature->getPosition().isInRange(player->getPosition(), 1, 1, 0)) {
-			SchedulerTask* task = createSchedulerTask(
+			SchedulerTask_ptr task = createSchedulerTask(
 			    MOVE_CREATURE_INTERVAL, [=, this, playerID = player->getID(), creatureID = movingCreature->getID()]() {
 				    playerMoveCreatureByID(playerID, creatureID, fromPos, toPos);
 			    });
-			player->setNextActionTask(task);
+			player->setNextActionTask(std::move(task));
 		} else {
 			playerMoveCreature(player, movingCreature, movingCreature->getPosition(), tile);
 		}
@@ -680,12 +680,12 @@ void Game::playerMoveCreature(Player* player, Creature* movingCreature, const Po
 {
 	if (!player->canDoAction()) {
 		uint32_t delay = player->getNextActionTime();
-		SchedulerTask* task =
+		SchedulerTask_ptr task =
 		    createSchedulerTask(delay, [=, this, playerID = player->getID(), movingCreatureID = movingCreature->getID(),
 		                                toPos = toTile->getPosition()]() {
 			    playerMoveCreatureByID(playerID, movingCreatureID, movingCreatureOrigPos, toPos);
 		    });
-		player->setNextActionTask(task);
+		player->setNextActionTask(std::move(task));
 		return;
 	}
 
@@ -703,13 +703,13 @@ void Game::playerMoveCreature(Player* player, Creature* movingCreature, const Po
 			g_dispatcher.addTask([this, playerID = player->getID(), listDir = std::move(listDir)]() {
 				playerAutoWalk(playerID, listDir);
 			});
-			SchedulerTask* task =
+			SchedulerTask_ptr task =
 			    createSchedulerTask(RANGE_MOVE_CREATURE_INTERVAL, [=, this, playerID = player->getID(),
 			                                                       movingCreatureID = movingCreature->getID(),
 			                                                       toPos = toTile->getPosition()] {
 				    playerMoveCreatureByID(playerID, movingCreatureID, movingCreatureOrigPos, toPos);
 			    });
-			player->setNextWalkActionTask(task);
+			player->setNextWalkActionTask(std::move(task));
 		} else {
 			player->sendCancelMessage(RETURNVALUE_THEREISNOWAY);
 		}
@@ -891,10 +891,10 @@ void Game::playerMoveItem(Player* player, const Position& fromPos, uint16_t spri
 {
 	if (!player->canDoAction()) {
 		uint32_t delay = player->getNextActionTime();
-		SchedulerTask* task = createSchedulerTask(delay, [=, this, playerID = player->getID()]() {
+		SchedulerTask_ptr task = createSchedulerTask(delay, [=, this, playerID = player->getID()]() {
 			playerMoveItemByPlayerID(playerID, fromPos, spriteId, fromStackPos, toPos, count);
 		});
-		player->setNextActionTask(task);
+		player->setNextActionTask(std::move(task));
 		return;
 	}
 
@@ -960,11 +960,11 @@ void Game::playerMoveItem(Player* player, const Position& fromPos, uint16_t spri
 			g_dispatcher.addTask([this, playerID = player->getID(), listDir = std::move(listDir)]() {
 				playerAutoWalk(playerID, listDir);
 			});
-			SchedulerTask* task =
+			SchedulerTask_ptr task =
 			    createSchedulerTask(RANGE_MOVE_ITEM_INTERVAL, [=, this, playerID = player->getID()]() {
 				    playerMoveItemByPlayerID(playerID, fromPos, spriteId, fromStackPos, toPos, count);
 			    });
-			player->setNextWalkActionTask(task);
+			player->setNextWalkActionTask(std::move(task));
 		} else {
 			player->sendCancelMessage(RETURNVALUE_THEREISNOWAY);
 		}
@@ -1022,12 +1022,12 @@ void Game::playerMoveItem(Player* player, const Position& fromPos, uint16_t spri
 				g_dispatcher.addTask([this, playerID = player->getID(), listDir = std::move(listDir)]() {
 					playerAutoWalk(playerID, listDir);
 				});
-				SchedulerTask* task = createSchedulerTask(
+				SchedulerTask_ptr task = createSchedulerTask(
 				    RANGE_MOVE_ITEM_INTERVAL,
 				    [this, playerID = player->getID(), itemPos, spriteId, itemStackPos, toPos, count]() {
 					    playerMoveItemByPlayerID(playerID, itemPos, spriteId, itemStackPos, toPos, count);
 				    });
-				player->setNextWalkActionTask(task);
+				player->setNextWalkActionTask(std::move(task));
 			} else {
 				player->sendCancelMessage(RETURNVALUE_THEREISNOWAY);
 			}
@@ -2137,10 +2137,10 @@ void Game::playerUseItemEx(uint32_t playerId, const Position& fromPos, uint8_t f
 				g_dispatcher.addTask([this, playerID = player->getID(), listDir = std::move(listDir)]() {
 					playerAutoWalk(playerID, listDir);
 				});
-				SchedulerTask* task = createSchedulerTask(RANGE_USE_ITEM_EX_INTERVAL, [=, this]() {
+				SchedulerTask_ptr task = createSchedulerTask(RANGE_USE_ITEM_EX_INTERVAL, [=, this]() {
 					playerUseItemEx(playerId, itemPos, itemStackPos, fromSpriteId, toPos, toStackPos, toSpriteId);
 				});
-				player->setNextWalkActionTask(task);
+				player->setNextWalkActionTask(std::move(task));
 			} else {
 				player->sendCancelMessage(RETURNVALUE_THEREISNOWAY);
 			}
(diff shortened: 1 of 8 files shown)


src/tasks.h
```diff
@@ -41,17 +45,17 @@ class Task
...
+Task_ptr createTask(uint32_t expiration, TaskFunc&& f);
 
 class Dispatcher : public ThreadHolder<Dispatcher>
 {
 public:
-	void addTask(Task* task);
+	void addTask(Task_ptr task);
 
-	void addTask(TaskFunc&& f) { addTask(new Task(std::move(f))); }
+	void addTask(TaskFunc&& f) { addTask(std::make_unique<Task>(std::move(f))); }
```
ranisalt, line 56:
```suggestion
	void addTask(TaskFunc&& f) { addTask(std::make_unique<Task>(std::forward(f))); }
```

src/tasks.h
```diff
@@ -41,17 +45,17 @@ class Task
...
 
-	void addTask(uint32_t expiration, TaskFunc&& f) { addTask(new Task(expiration, std::move(f))); }
+	void addTask(uint32_t expiration, TaskFunc&& f) { addTask(make_unique<Task>(expiration, std::move(f))); }
```
ranisalt, line 58:
```suggestion
	void addTask(uint32_t expiration, TaskFunc&& f) { addTask(std::make_unique<Task>(expiration, std::forward(f))); }
```

src/game.cpp
```diff
@@ -634,11 +634,11 @@ void Game::playerMoveThing(uint32_t playerId, const Position& fromPos, uint16_t
 		}
 
 		if (movingCreature->getPosition().isInRange(player->getPosition(), 1, 1, 0)) {
-			SchedulerTask* task = createSchedulerTask(
```
ranisalt, line 637: I believe in these situations we should use `auto`, because it doesn't even matter what the type is here, we just pass it along. Wdyt?

src/tasks.cpp
```diff
@@ -30,29 +30,26 @@ void Dispatcher::threadMain()
 		tmpTaskList.swap(taskList);
 		taskLockUnique.unlock();
 
-		for (Task* task : tmpTaskList) {
+		for (auto& task : tmpTaskList) {
 			if (!task->hasExpired()) {
 				++dispatcherCycle;
 				// execute it
 				(*task)();
```
ranisalt, line 37: While at it, what do you think of writing
```suggestion
				std::invoke(*task);
```
here? I'm not a fan of the `(*ptr)()` syntax because of the extra parenthesis needed 😆

ranisalt:
This is really neat, we should do more of that.

ArturKnopik:
> This is really neat, we should do more of that.
> 
> I'd love to help more but valgrind borked on my machine and won't run :(

Thanks, you can help by doing a review :)

src/game.cpp
nekiro: or just move the call directly to method call

src/tasks.h
ArturKnopik: Can you explain why std::forward?

Codinablack: > Can you explain why std::forward?

To maintain the current value type (L or R) for the meta side of things

src/scheduler.cpp
```diff
@@ -62,4 +63,7 @@ void Scheduler::shutdown()
 	});
 }
 
-SchedulerTask* createSchedulerTask(uint32_t delay, TaskFunc&& f) { return new SchedulerTask(delay, std::move(f)); }
+SchedulerTask_ptr createSchedulerTask(uint32_t delay, TaskFunc&& f)
+{
+	return SchedulerTask_ptr(new SchedulerTask(delay, std::move(f)));
```
MillhioreBT, line 68: The correct way would be to use `std::make_unique` instead of calling `new`
There are many advantages to using make_unique

less allocations, safe in case of exceptions, and it's the modern way of doing things

src/game.cpp
ArturKnopik: > or just move the creation directly to method call, then we can let the compiler do its best thing

it's looks like some abomination
Zrzut ekranu z 2024-12-27 09-49-31

src/scheduler.h
```diff
@@ -9,29 +9,33 @@
...
 static constexpr int32_t SCHEDULER_MINTICKS = 50;
 
+class SchedulerTask;
+
+using SchedulerTask_ptr = std::unique_ptr<SchedulerTask>;
+
 class SchedulerTask : public Task
 {
 public:
+	SchedulerTask(uint32_t delay, TaskFunc&& f) : Task(std::forward<TaskFunc>(f)), delay(delay) {}
```
ArturKnopik, line 19: must become public to give abllity to call c-tor by make_unique from ```SchedulerTask_ptr createSchedulerTask(uint32_t delay, TaskFunc&& f)```

src/game.cpp
nekiro: std::move shouldnt be needed here, but yeah, might be true it doesnt look the best

src/scheduler.h
nekiro: you could use a friend class

ArturKnopik: Friend is here but somehow it's not compile :<

ArturKnopik: the problem is make_unique becouse it has no access to private c-tor of SchedulerTask class
there is no way to use something like as friend:
```template <typename... Args> Friend std::unique_ptr<SchedulerTask> std::make_unique(Args&&... args);```

**Comment by MillhioreBT** [comment]
By the way, wouldn't it be dangerous to use `walkTask` after being moved?
for example in the `goToFollowCreature` method it is used to validate in a boolean context
Using an object that was moved is undefined behavior.

File: src/scheduler.h (hunk 6 above)
Comment (ranisalt) [reply src/scheduler.h:19]: `friend` sucks, just make it public

File: src/player.cpp [hunk 7]
```diff
@@ -1508,26 +1508,25 @@ void Player::checkTradeState(const Item* item)
...
+void Player::setNextWalkActionTask(SchedulerTask_ptr task)
 {
 	if (walkTaskEvent != 0) {
 		g_scheduler.stopEvent(walkTaskEvent);
 		walkTaskEvent = 0;
 	}
 
-	delete walkTask;
-	walkTask = task;
+	walkTask = std::move(task);
```
ranisalt, line 1518: I believe if you use

```suggestion
	std::swap(walkTask, task);
```

You may skip one deletion. This is absolutely minor, change it only if you touch this code again 😆

src/player.cpp
```diff
@@ -3469,8 +3468,7 @@ void Player::onWalkAborted()
 void Player::onWalkComplete()
 {
 	if (walkTask) {
-		walkTaskEvent = g_scheduler.addEvent(walkTask);
-		walkTask = nullptr;
+		walkTaskEvent = g_scheduler.addEvent(std::move(walkTask));
```
ranisalt, line 3471: Prone to use-after-move, you still need `walkTask = nullptr`

src/scheduler.h
```diff
@@ -9,29 +9,33 @@
...
 	uint32_t getDelay() const { return delay; }
 
 private:
-	SchedulerTask(uint32_t delay, TaskFunc&& f) : Task(std::move(f)), delay(delay) {}
-
 	uint32_t eventId = 0;
 	uint32_t delay = 0;
 
-	friend SchedulerTask* createSchedulerTask(uint32_t, TaskFunc&&);
+	friend SchedulerTask_ptr createSchedulerTask(uint32_t delay, TaskFunc&& f);
```
ranisalt, line 30: Not needed anymore I think, since the constructor is public

(boo `friend` is bad)

src/player.cpp
MillhioreBT: I think your suggestion is good, good capture, I was thinking about this:

```cpp
walkTaskEvent = g_scheduler.addEvent(SchedulerTask_ptr(walkTask.release()));
```

However, I am not convinced about having to build a new unique_ptr, although it should be free, I don't know how profitable it is.

The option to assign nullptr is correct and I like it

ArturKnopik:
after std::move source ptr is nullified.
image

nekiro:
> after std::move source ptr is nullified. ![image](https://private-user-images.githubusercontent.com/17506599/399293282-

I recall reading that this behaviour isnt explicitly predictable to be always nullified. There is no guarantee, so it's not safe to assume.

MillhioreBT:
> https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf
> 
> section 20.7.1 point 4 Additionally, u can, upon request, transfer ownership to another unique pointer u2. Upon completion of such a transfer, the following postconditions hold: — u2.p is equal to the pre-transfer u.p, — **u.p is equal to nullptr**, and — if the pre-transfer u.d maintained state, such state has been transferred to u2.d.
> 
> what would be the point if after transferring ownership the pointer address remained unchanged... that was at odds with the idea of ​​unique ptr

It is better not to assume anything and do things as safely as possible.
By the way the linter actually warns you not to use objects that were moved, no matter if it is a unique_ptr or something else

ArturKnopik:
[n3337 (1).pdf](https://github.com/user-attachments/files/18276492/n3337.1.pdf)
This is official standard document for c++11, 20.7.1 point 4 confirm assamtions about nullify

ArturKnopik:
https://stackoverflow.com/questions/36071220/what-happens-to-unique-ptr-after-stdmove

src/player.cpp
ArturKnopik: why? deletion will be moved to end of the scope

ArturKnopik:
> Nice, so this is one of the few use-after-move situations that are defined 😆 good to go

yes, this is one of the few cases where it is defined in the standard

ArturKnopik:
Can someone do a review?

src/game.cpp
```diff
@@ -680,12 +680,12 @@ void Game::playerMoveCreature(Player* player, Creature* movingCreature, const Po
 {
 	if (!player->canDoAction()) {
 		uint32_t delay = player->getNextActionTime();
-		SchedulerTask* task =
+		SchedulerTask_ptr task =
```
ranisalt, line 683:
```suggestion
		auto task =
```

src/player.cpp
ranisalt: If you `move`, it will destroy `walkTask`, move, and then destroy what's left in `task`

If you `swap`, it will not destroy `walkTask`, but swap the contents with `task` which will be destroyed afterwards

I don't think it matters though so use what you prefer 😉

src/game.cpp
ArturKnopik: I think the current code is better, the function name does not indicate that we are creating a pointer

src/player.cpp
ArturKnopik: The swap looks strange.
Looking at it, I have no idea what author meant by doing swap... for me, it just obscures the code

src/game.cpp
ranisalt: All other function calls are using auto, it's either one or the other code style. I'm OK with the explicit name instead

PR #4883, opened 2024-12-27
Fix MySQL memory leaks

ArturKnopik:
### Pull Request Prelude

### Changes Proposed

Fix MySQL memory leaks

**Issues addressed:** 

https://github.com/otland/forgottenserver/issues/4288

**How to test:** 

Requirements: Linux
1) Build project in debug mode
2) Check if Mysql server is not working
3) Run server using valgrind(valgrind --leak-check=full ./tfs)
4) Looks for logs like (They shouldn't be there)
```
==7227== 56 bytes in 1 blocks are possibly lost in loss record 3 of 3
==7227==    at 0x4846828: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==7227==    by 0x4FD4977: ??? (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.21.2.40)
==7227==    by 0x4F73EC4: mysql_server_init (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.21.2.40)
==7227==    by 0x4F83AF8: mysql_init (in /usr/lib/x86_64-linux-gnu/libmysqlclient.so.21.2.40)
==7227==    by 0x48B7CC: connectToDatabase(bool) (database.cpp:22)
==7227==    by 0x48BC34: Database::connect() (database.cpp:66)
==7227==    by 0x278F1E: (anonymous namespace)::mainLoader(ServiceManager*) (otserv.cpp:116)
==7227==    by 0x27A320: startServer()::{lambda()#1}::operator()() const (otserv.cpp:285)
==7227==    by 0x27BE4F: void std::__invoke_impl<void, startServer()::{lambda()#1}&>(std::__invoke_other, startServer()::{lambda()#1}&) (invoke.h:61)
==7227==    by 0x27BB7E: std::enable_if<is_invocable_r_v<void, startServer()::{lambda()#1}&>, void>::type std::__invoke_r<void, startServer()::{lambda()#1}&>(startServer()::{lambda()#1}&) (invoke.h:111)
==7227==    by 0x27B5BC: std::_Function_handler<void (), startServer()::{lambda()#1}>::_M_invoke(std::_Any_data const&) (std_function.h:290)
==7227==    by 0x168943: std::function<void ()>::operator()() const (std_function.h:591)
==7227== 
```

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -61,7 +61,7 @@ static bool executeQuery(tfs::detail::Mysql_ptr& handle, std::string_view query,
 	return true;
 }
 
-Database::~Database() { mysql_server_end(); }
+Database::~Database() { mysql_library_end(); }
 
 bool Database::connect()
 {
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -61,6 +61,8 @@ static bool executeQuery(tfs::detail::Mysql_ptr& handle, std::string_view query,
 	return true;
 }
 
+Database::~Database() { mysql_server_end(); }
+
 bool Database::connect()
 {
 	auto newHandle = connectToDatabase(false);
--- a/src/database.h
+++ b/src/database.h
@@ -37,6 +37,11 @@ public:
 	}
 
 	/**
+	 * Close MySQL client
+	 */
+	~Database();
+
+	/**
 	 * Connects to the database
 	 *
 	 * @return true on successful connection, false on error


nekiro (approved):

nekiro:
looks like mysql_server_end is deprecated, mysql_library_end should be used
https://dev.mysql.com/doc/c-api/5.7/en/mysql-server-end.html

ArturKnopik:
unique_ptr is calling mysql_close as custom deleter
https://github.com/otland/forgottenserver/blob/8d79ba472b29d81a9c00dbe87c17af450d53181f/src/database.h#L16

https://github.com/otland/forgottenserver/blob/8d79ba472b29d81a9c00dbe87c17af450d53181f/src/database.h#L20

ArturKnopik:
from MySQL header 

#define mysql_library_init mysql_server_init
#define mysql_library_end mysql_server_end

image

Codinablack (approved):
LGTM

nekiro (approved):

MillhioreBT (approved):

gesior:
TFS has 2 connections (2 `Database` instances). One for dispatcher/network threads and one for databasetasks.

`mysql_library_end();` does not accept `handle` parameter to cleanup given connection.
Doesn't it remove some variables required by active MySQL connection in second thread?
We may remove not existing memory leak (`Database` class is destroyed on server shutdown, so OS will cleanup all app RAM anyway) and add possible crash on server shutdown (DatabaseTasks thread can go down before Dispatcher thread finish saving all players/houses).

ArturKnopik:
Game class is in global scope, thats means it will be destroyed as last thing and most likely after 
```cpp
g_databaseTasks.join();
g_dispatcher.join();
```

MillhioreBT:
> TFS has 2 connections (2 `Database` instances). One for dispatcher/network threads and one for databasetasks.
> 
> `mysql_library_end();` does not accept `handle` parameter to cleanup given connection. Doesn't it remove some variables required by active MySQL connection in second thread? We may remove not existing memory leak (`Database` class is destroyed on server shutdown, so OS will cleanup all app RAM anyway) and add possible crash on server shutdown (DatabaseTasks thread can go down before Dispatcher thread finish saving all players/houses).



> mysql_library_end();

It is possible that he is referring to the fact that `mysql_library_end();` can be called twice, since when `g_databaseTasks` and then the singleton are destroyed, there will be a double call

I don't know if it's safe or if it causes any problems.

ArturKnopik:
only Game class is calling mysql_library_end(), there is no way to call it second time

MillhioreBT:
> only Game class is calling mysql_library_end(), there is no way to call it second time

I don't know, it would be good if you made sure of that.

since `g_databaseTasks` is global
the singleton is global

At some point these objects must be destroyed and the destructor called

According to you, what is it that forces the destructor to be called only once?

If there are two Database instances, don't you think that at some point the two objects must call their destructor method?

I haven't tried anything, I'm just saying that it would be good if you took a good look at it.

ArturKnopik:
no d-tor for database, the same thing for databasetasks

MillhioreBT:
> no d-tor for database, the same thing for databasetasks

I don't know if you're playing a joke on me or you just don't know that objects in C++ have a destructor by default, right?

I know that Database does not have a destructor, but you are writing one in this PR so that it does now have one.

ArturKnopik:
Sorry for the confusion, I'm writing from my phone, I don't know why but I was convinced that this fix was for the game class and not the database... maybe this fix should be moved to Game class :p

ArturKnopik:
ok, it's looks like it's invalid fix, Database is not true singleton, destructor is called 2times, no information about behavior of mysql_library_end and multiple calls

PR #4887, opened 2025-01-24
Fix onNetworkMessage event memory leak

maattch:
### Pull Request Prelude

**Issues addressed:**
Fix memory leak

**How to test:**
Disable **Player:onNetworkMessage** event and send packets with unknown opcode to server, such as from store.

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -1403,7 +1403,7 @@ void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage_ptr msg)
 
 	lua_pushnumber(L, recvByte);
 
-	tfs::lua::pushUserdata(L, msg.release()); 
+	tfs::lua::pushUserdata(L, msg.release());
 	tfs::lua::setMetatable(L, -1, "NetworkMessage");
 
 	scriptInterface.callVoidFunction(3);
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -1380,7 +1380,7 @@ void onInventoryUpdate(Player* player, Item* item, slots_t slot, bool equip)
 	scriptInterface.callVoidFunction(4);
 }
 
-void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage_ptr msg)
+void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage_ptr& msg)
 {
 	// Player:onNetworkMessage(recvByte, msg)
 	if (playerHandlers.onNetworkMessage == -1) {
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -1380,7 +1380,7 @@ void onInventoryUpdate(Player* player, Item* item, slots_t slot, bool equip)
 	scriptInterface.callVoidFunction(4);
 }
 
-void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage* msg)
+void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage_ptr msg)
 {
 	// Player:onNetworkMessage(recvByte, msg)
 	if (playerHandlers.onNetworkMessage == -1) {
@@ -1403,7 +1403,7 @@ void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage* msg)
 
 	lua_pushnumber(L, recvByte);
 
-	tfs::lua::pushUserdata(L, msg);
+	tfs::lua::pushUserdata(L, msg.release()); 
 	tfs::lua::setMetatable(L, -1, "NetworkMessage");
 
 	scriptInterface.callVoidFunction(3);
--- a/src/events.h
+++ b/src/events.h
@@ -7,9 +7,9 @@
 #include "const.h"
 #include "creature.h"
 #include "luascript.h"
+#include "networkmessage.h"
 
 class ItemType;
-class NetworkMessage;
 class Party;
 class Spell;
 class Tile;
@@ -83,7 +83,7 @@ void onLoseExperience(Player* player, uint64_t& exp);
 void onGainSkillTries(Player* player, skills_t skill, uint64_t& tries);
 void onWrapItem(Player* player, Item* item);
 void onInventoryUpdate(Player* player, Item* item, slots_t slot, bool equip);
-void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage* msg);
+void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage_ptr msg);
 bool onSpellCheck(Player* player, const Spell* spell);
 
 } // namespace tfs::events::player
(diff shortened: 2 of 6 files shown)


MillhioreBT:
In the best-case scenario, Lua will clean up the object; in the worst-case scenario, there will be a memory leak. That is correct—thank you for catching this issue.

However, I would prefer if you made some adjustments to the code.

1) Construct the `unique_ptr` directly in the lambda's scope by calling `std::make_unique`.  
2) Make the `msg` parameter passed by reference in the methods: `parsePlayerNetworkMessage` and `playerOnNetworkMessage`.

src/events.cpp
```diff
@@ -1380,7 +1380,7 @@ void onInventoryUpdate(Player* player, Item* item, slots_t slot, bool equip)
 	scriptInterface.callVoidFunction(4);
 }
 
-void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage* msg)
+void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage_ptr msg)
```
MillhioreBT, line 1383:
```cpp
void onNetworkMessage(Player* player, uint8_t recvByte, const NetworkMessage_ptr& msg)
```

src/events.h
```diff
@@ -83,7 +83,7 @@ void onLoseExperience(Player* player, uint64_t& exp);
 void onGainSkillTries(Player* player, skills_t skill, uint64_t& tries);
 void onWrapItem(Player* player, Item* item);
 void onInventoryUpdate(Player* player, Item* item, slots_t slot, bool equip);
-void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage* msg);
+void onNetworkMessage(Player* player, uint8_t recvByte, NetworkMessage_ptr msg);
```
MillhioreBT, line 86:
```cpp
void onNetworkMessage(Player* player, uint8_t recvByte, const NetworkMessage_ptr& msg);
```

src/game.cpp
```diff
@@ -5548,14 +5548,14 @@ void Game::parsePlayerExtendedOpcode(uint32_t playerId, uint8_t opcode, const st
 	}
 }
 
-void Game::parsePlayerNetworkMessage(uint32_t playerId, uint8_t recvByte, NetworkMessage* msg)
+void Game::parsePlayerNetworkMessage(uint32_t playerId, uint8_t recvByte, NetworkMessage_ptr msg)
```
MillhioreBT, line 5551:
```cpp
void Game::parsePlayerNetworkMessage(uint32_t playerId, uint8_t recvByte, const NetworkMessage_ptr& msg)
```

src/game.cpp
```diff
@@ -5548,14 +5548,14 @@ void Game::parsePlayerExtendedOpcode(uint32_t playerId, uint8_t opcode, const st
...
 {
 	Player* player = getPlayerByID(playerId);
 	if (!player) {
 		return;
 	}
 
-	tfs::events::player::onNetworkMessage(player, recvByte, msg);
+	tfs::events::player::onNetworkMessage(player, recvByte, std::move(msg));
```
MillhioreBT, line 5558:
```cpp
tfs::events::player::onNetworkMessage(player, recvByte, msg);
```

src/game.h
```diff
@@ -392,7 +392,7 @@ class Game
 	void playerAcceptMarketOffer(uint32_t playerId, uint32_t timestamp, uint16_t counter, uint16_t amount);
 
 	void parsePlayerExtendedOpcode(uint32_t playerId, uint8_t opcode, const std::string& buffer);
-	void parsePlayerNetworkMessage(uint32_t playerId, uint8_t recvByte, NetworkMessage* msg);
+	void parsePlayerNetworkMessage(uint32_t playerId, uint8_t recvByte, NetworkMessage_ptr msg);
```
MillhioreBT, line 395:
```cpp
void parsePlayerNetworkMessage(uint32_t playerId, uint8_t recvByte, const NetworkMessage_ptr& msg);
```

src/protocolgame.cpp
```diff
@@ -793,8 +793,10 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
 			// case 0xFE: break; // store window history 2
 
 		default:
+			// we cannot pass an unique_ptr as capture here because
+			// std::function requires the callable object to be *copyable*
 			g_dispatcher.addTask([=, playerID = player->getID(), msg = new NetworkMessage(msg)]() {
```
MillhioreBT, line 798:
```cpp
g_dispatcher.addTask([=, playerID = player->getID(), msg = std::make_unique<NetworkMessage>(msg)]() {
```

src/protocolgame.cpp
```diff
@@ -793,8 +793,10 @@ void ProtocolGame::parsePacket(NetworkMessage& msg)
...
-				g_game.parsePlayerNetworkMessage(playerID, recvbyte, msg);
+				g_game.parsePlayerNetworkMessage(playerID, recvbyte, NetworkMessage_ptr(msg));
```
MillhioreBT, line 799:
```cpp
g_game.parsePlayerNetworkMessage(playerID, recvbyte, msg);
```

MillhioreBT (changes requested):

src/protocolgame.cpp
maattch: You can't do that, `std::function` **requires** the object to be **copyable**, and `std::unique_ptr` is not copyable.

src/events.cpp
maattch: Cannot be passed as const reference because we release the pointer in the end.

MillhioreBT:
You are right, anyway you can not pass a mutable lambda to the `addTask` method, so its solution seems to be the only option for now, it only fixes the cclang format.

src/events.cpp
MillhioreBT: Yes, I just noticed, using const was not the right option, although you can still use non -constant references in this method

src/events.h
```diff
@@ -7,9 +7,9 @@
 #include "const.h"
 #include "creature.h"
 #include "luascript.h"
+#include "networkmessage.h"
```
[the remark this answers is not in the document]
maattch: it doesnt, the only way would be if we also define `NetworkMessage_ptr` inside `events.h`, but then we would have two definitions of the same thing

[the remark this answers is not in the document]
nekiro: true, didnt notice there is no global definition file for pointer types

nekiro (approved):

MillhioreBT:
@maattch If you fix the format of the PR, it will be merged.

maattch:
The code is already formatted, the check-format action is broken (you can click it to see what is causing it).

ArturKnopik:
> The code is already formatted, the check-format action is broken (you can click it to see what is causing it).

you can reopen this PR to trigger jobs again.

PR #4891, opened 2025-02-02
remove seek from NetworkMessage

### LINKED ISSUE #4877: [Bug]: NetworkMessage always increase size even when it's not needed

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below)

### OS Description

_No response_

### Bug description

Adding data to network message always increase it size even if it's not required

N/A

### Steps to reproduce

put this code to lua script and execute
```LUA
local msg = NetworkMessage()
msg:addByte(1)
msg:addByte(2)
msg:addByte(0)
msg:addByte(4)
print(msg:len()) -- will return 4
msg:seek(1) -- set write position
msg:addByte(2) -- already used field
print(msg:len()) -- will return 5
```
@EDIT updated returned len() values to valid ones

### Actual Behavior

Message size increases when not required

### Expected Behavior

Message size increases only when required

### Backtrace

```cpp
N/A
```

xyron98:
### Pull Request Prelude

### Changes Proposed

Remove seek functionality from NetworkMessage since it is not used anywhere in the codebase and can behave unexpectedly when used with addByte.

**Issues addressed:** 

#4877 

**How to test:** 

Compile and run the server.

### DIFF
--- a/data/cpplinter.lua
+++ b/data/cpplinter.lua
@@ -167,7 +167,6 @@ Tile = {}
 ---@field addItem fun(self: NetworkMessage, item: Item)
 ---@field addItemId fun(self: NetworkMessage, itemId: number)
 ---@field reset fun(self: NetworkMessage)
----@field seek fun(self: NetworkMessage, position: number)
 ---@field tell fun(self: NetworkMessage): number
 ---@field len fun(self: NetworkMessage): number
 ---@field skipBytes fun(self: NetworkMessage, count: number)
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2501,7 +2501,6 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod(L, "NetworkMessage", "addItemId", LuaScriptInterface::luaNetworkMessageAddItemId);
 
 	registerMethod(L, "NetworkMessage", "reset", LuaScriptInterface::luaNetworkMessageReset);
-	registerMethod(L, "NetworkMessage", "seek", LuaScriptInterface::luaNetworkMessageSeek);
 	registerMethod(L, "NetworkMessage", "tell", LuaScriptInterface::luaNetworkMessageTell);
 	registerMethod(L, "NetworkMessage", "len", LuaScriptInterface::luaNetworkMessageLength);
 	registerMethod(L, "NetworkMessage", "skipBytes", LuaScriptInterface::luaNetworkMessageSkipBytes);
@@ -6213,18 +6212,6 @@ int LuaScriptInterface::luaNetworkMessageReset(lua_State* L)
 	if (message) {
 		message->reset();
 		tfs::lua::pushBoolean(L, true);
-	} else {
-		lua_pushnil(L);
-	}
-	return 1;
-}
-
-int LuaScriptInterface::luaNetworkMessageSeek(lua_State* L)
-{
-	// networkMessage:seek(position)
-	NetworkMessage* message = tfs::lua::getUserdata<NetworkMessage>(L, 1);
-	if (message && isNumber(L, 2)) {
-		tfs::lua::pushBoolean(L, message->setBufferPosition(tfs::lua::getNumber<uint16_t>(L, 2)));
 	} else {
 		lua_pushnil(L);
 	}
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -390,7 +390,6 @@ private:
 	static int luaNetworkMessageAddItemId(lua_State* L);
 
 	static int luaNetworkMessageReset(lua_State* L);
-	static int luaNetworkMessageSeek(lua_State* L);
 	static int luaNetworkMessageTell(lua_State* L);
 	static int luaNetworkMessageLength(lua_State* L);
 	static int luaNetworkMessageSkipBytes(lua_State* L);


ArturKnopik:
probably https://github.com/otland/forgottenserver/blob/b227678ab3584b8349913abfe4529292678a3fdf/src/networkmessage.h#L125 also can be removed, needs to be checked

MillhioreBT:
While it is true that the base does not use `seek`, you cannot be certain that someone else might use it. It is unsafe if used incorrectly, just like countless other things.

ranisalt:
If seek is not used, tell is likely not used too

Codinablack:
So the point is to just remove lua API from the end users is its not used in the datapack?

MillhioreBT:
> So the point is to just remove lua API from the end users is its not used in the datapack?

I thought that, in that case we must eliminate 50% of the LUA functions/methods that are not used in the Datapack
modalwindow, helper, event_helper, compat.lua, ect...

ArturKnopik:
> So the point is to just remove lua API from the end users because its not used in the datapack?


> > So the point is to just remove lua API from the end users is its not used in the datapack?
> 
> I thought that, in that case we must eliminate 50% of the LUA functions/methods that are not used in the Datapack modalwindow, helper, event_helper, compat.lua, ect...

please read issue linked in this PR, i hope you will understand why this PR exist

MillhioreBT:
> > So the point is to just remove lua API from the end users because its not used in the datapack?
> 
> > > So the point is to just remove lua API from the end users is its not used in the datapack?
> > 
> > 
> > I thought that, in that case we must eliminate 50% of the LUA functions/methods that are not used in the Datapack modalwindow, helper, event_helper, compat.lua, ect...
> 
> please read issue linked in this PR, i hope you will understand why this PR exist

Wouldn't it be better to add a check? Only increase if necessary.  
I've been running some tests, and the time difference is negligible in Release mode.

Figure_1

<details>

<summary>Test Code</summary>

```cpp
#include <benchmark/benchmark.h>
#include <cstring>
#include <type_traits>
#include <vector>

struct BufferInfo
{
	std::vector<char> buffer;
	size_t position = 0;
	size_t length = 0;
};

template <typename T>
void add_with_check(BufferInfo& info, T value)
{
	static_assert(std::is_trivially_copyable_v<T>, "T must be trivially copyable");

	if (info.position + sizeof(T) > info.buffer.size()) {
		return;
	}

	if (info.position + sizeof(T) > info.length) {
		info.length = info.position + sizeof(T);
	}

	std::memcpy(info.buffer.data() + info.position, &value, sizeof(T));
	info.position += sizeof(T);
}

template <typename T>
void add_without_check(BufferInfo& info, T value)
{
	static_assert(std::is_trivially_copyable_v<T>, "T must be trivially copyable");

	if (info.position + sizeof(T) > info.buffer.size()) {
		return;
	}

	std::memcpy(info.buffer.data() + info.position, &value, sizeof(T));
	info.position += sizeof(T);
	info.length += sizeof(T);
}

static void BM_AddWithCheck(benchmark::State& state)
{
	BufferInfo info;
	info.buffer.resize(1024 * 1024);
	int value = 42;

	for (auto _ : state) {
		add_with_check(info, value);
		info.position = 0;
	}
}
BENCHMARK(BM_AddWithCheck);

static void BM_AddWithoutCheck(benchmark::State& state)
{
	BufferInfo info;
	info.buffer.resize(1024 * 1024);
	int value = 42;

	for (auto _ : state) {
		add_without_check(info, value);
		info.position = 0;
	}
}
BENCHMARK(BM_AddWithoutCheck);

static void BM_AddWithCheckOverwrite(benchmark::State& state)
{
	BufferInfo info;
	info.buffer.resize(1024 * 1024);
	info.length = 512 * 1024;
	int value = 42;

	for (auto _ : state) {
		add_with_check(info, value);
		info.position = 0;
	}
}
BENCHMARK(BM_AddWithCheckOverwrite);

static void BM_AddWithoutCheckOverwrite(benchmark::State& state)
{
	BufferInfo info;
	info.buffer.resize(1024 * 1024);
	info.length = 512 * 1024;
	int value = 42;

	for (auto _ : state) {
		add_without_check(info, value);
		info.position = 0;
	}
}
BENCHMARK(BM_AddWithoutCheckOverwrite);

BENCHMARK_MAIN();
```

</details>

Honestly, scenarios where `add<T>` is called hundreds of thousands of times consecutively seem like something that never happens. And if they did occur, the overhead would be a couple of nanoseconds.

ArturKnopik:
the increase is at 30-35% and this code is commonly used almost everywhere, I prefer not to add checks, it's too big increase in execution time...
everything that happens almost always triggers sending some networkmessages: 
* movement (player rotation, sending of map + possibly a monster if present), 
* magic effect, 
* distance effect, 
* messages to player/channels
* market
* from latest protocol preys/bestiary/boostiary (timeout is updated by server)
* exp/mana spend/skills
* moving items
* etc...

ranisalt:
This is a very hot code path and I'd rather not add the check. If not even the engine needs to arbitrarily navigate a network message, I don't think the end user needs to.

`seek` is not needed, `tell` is not needed either (it will always equal `len`), and `NetworkMessage::setBufferPosition` in C++ will be unused too.

NRH-AA:
I vote we add: byteAppend(seekPos) stringAppend(seekPos) and they handle the seek, check and add the data, and go back to end. Features above all.

alysonjacomin:
@NRH-AA  I completely agree, the more resources the better, and your idea is great

DSpeichert (approved):

DSpeichert:
This is good but possibly not enough. There's another entrypoint:
msg:skipBytes(-3) (rewind) or msg:skipBytes(10) (jump forward)

They're possibly worse because with msg:seek(), the header was protected but with msg:skipBytes(-8) on a fresh msg, you're wiping the header.

Seek still should be discouraged.

PR #4892, opened 2025-02-11
Fix possible condition nulls

kaleohanopahala:
- nullptr in executeCondition in ConditionRegeneration.
- nullptr in executeCondition in ConditionSoul.
- nullptr in doDamage in ConditionDamage.
- nullptr in startCondition in ConditionManaShield.
- nullptr in executeCondition in ConditionLight.

### DIFF
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -880,6 +880,10 @@ void ConditionRegeneration::serialize(PropWriteStream& propWriteStream)
 
 bool ConditionRegeneration::executeCondition(Creature* creature, int32_t interval)
 {
+	if (!creature) {
+		return false;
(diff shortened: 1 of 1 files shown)


Codinablack (approved):
Solid! LGTM

MillhioreBT:
Hello again, the same thing happens here—meaningless check, `this` will never be `nullptr`.

The only context where `executeCondition` is called is when `Creature::executeConditions` is invoked, which sends `this` as the first parameter.

If you insist that it is a problem, it wouldn't hurt to add steps to reproduce the issue.

ArturKnopik:
LGTM

PR #4893, opened 2025-02-11
Fix container nullpoints

kaleohanopahala:
Prevent few nullptr's

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -56,6 +56,10 @@ Container::~Container()
 Item* Container::clone() const
 {
 	Container* clone = static_cast<Container*>(Item::clone());
+	if (!clone) {
+		return nullptr;
+	}
+
 	for (Item* item : itemlist) {
 		clone->addItem(item->clone());
 	}
@@ -94,6 +98,10 @@ bool Container::hasContainerParent() const
 
 void Container::addItem(Item* item)
 {
+	if (!item) {
+		return;
+	}
+
 	itemlist.push_back(item);
 	item->setParent(this);
 }
@@ -544,6 +552,10 @@ void Container::updateThing(Thing* thing, uint16_t itemId, uint32_t count)
 
 void Container::replaceThing(uint32_t index, Thing* thing)
 {
+	if (!thing) {
+		return;
+	}
+
 	Item* item = thing->getItem();
 	if (!item) {
 		return /*RETURNVALUE_NOTPOSSIBLE*/;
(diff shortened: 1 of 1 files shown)


Codinablack (approved):
Another Solid PR! LGTM

MillhioreBT:
It is true that `Item::CreateItem` can return a `nullptr`. However, in the context of the `clone` method, it is impossible for it to return `nullptr`, since if you are cloning an item that already exists, it means the item has its properties already defined and they exist. It would be impossible for the cloning not to happen.

If the `ItemType` were to be deleted in the middle of the game's execution, I believe that is no longer the engine's problem, but rather the programmer's, who is doing things that could break the game anyway, with or without that check.

MillhioreBT:
> > It is true that `Item::CreateItem` can return a `nullptr`. However, in the context of the `clone` method, it is impossible for it to return `nullptr`, since if you are cloning an item that already exists, it means the item has its properties already defined and they exist. It would be impossible for the cloning not to happen.
> 
> i thing is good to check nullptr, clone (Item::CreateItem) can fail and return nullptr in some cases.

Show me in what case? Are you going to clone an item that doesn't exist? And if the item already exists, should we consider that said item doesn't have an associated ID or ItemType? It sounds strange.

ArturKnopik:
https://github.com/otland/forgottenserver/blob/23911351a33102be60d0b088397d5d64c95b430c/src/item.cpp#L43 can return nullptr in case of some memory allocation problem, probaby check for nullptr should be added in this fiction, incrementimg reff count will crash server if mem allocation failed

MillhioreBT:
> can return nullptr in case of some memory allocation problem, probaby check for nullptr should be added in this fiction, incrementimg reff count will crash server if mem allocation failed

In that case, if you're sure about it, you should make a PR to revert this: https://github.com/otland/forgottenserver/pull/4532

ArturKnopik:
i think it should be reverted, for me every memory allocation should be checked before use

MillhioreBT:
> i think it should be reverted, for me every memory allocation should be checked before use @updage @MillhioreBT after getting knowledge about new operator failure cases, new in case of failure will throw std::bad_alloc exception by deafult, in tfs it will crash server, check for null probaby is not needed
> 
> new(std::nothrow) will return nullptr instead throwing exception

Exactly, it's not necessary currently, unless additional changes are made, such as what you mentioned: using `new(std::nothrow)`. Otherwise, it doesn't make sense.

ranisalt:
> can return nullptr in case of some memory allocation problem, probaby check for nullptr should be added in this fiction, incrementimg reff count will crash server if mem allocation failed

Failure to allocate memory will crash the program, not return `nullptr`

ArturKnopik (approved):

ArturKnopik:
LGTM, i will leave this PR for a while(few days) in case someone objects to the clone changes

PR #4894, opened 2025-02-12
fix: null dereferencing on onAttackedCreatureDrainHealth

kaleohanopahala:
This prevents potential crashes due to null dereferencing and avoids processing invalid damage values.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -917,6 +917,10 @@ void Creature::onAttacked()
 
 void Creature::onAttackedCreatureDrainHealth(Creature* target, int32_t points)
 {
+	if (!target || points <= 0) {
+		return;
+	}
+
 	target->addDamagePoints(this, points);
 }
 


ArturKnopik:
https://github.com/otland/forgottenserver/blob/23911351a33102be60d0b088397d5d64c95b430c/src/creature.cpp#L840

invalid value is checked by addDamagePoints method.

Codinablack:
> https://github.com/otland/forgottenserver/blob/23911351a33102be60d0b088397d5d64c95b430c/src/creature.cpp#L840
> 
> invalid value is checked by addDamagePoints method.

Yes but it doesn't validate the target, and early return is always preferable.

Codinablack (approved):
LGTM!

MillhioreBT:
In a normal situation, the pointer `this` will never be invalid, nor will it be `nullptr`, so the check simply doesn't make sense.

Although `changeHealth` may cause the creature's health to reach 0, death is postponed to the next task dispatcher cycle, while `onAttackedCreatureDrainHealth` will continue its execution in the current cycle. In the worst-case scenario, the creature will remain pending to be removed during resource cleanup in `checkCreatures`.

ArturKnopik (approved):

PR #4870, opened 2024-12-09
Add packet compression

nekiro:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Add missing packet compression for big packets

**Issues addressed:** 
None

**How to test:** 
Login to game, walk around, no crashes? works

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4897, opened 2025-02-22
Fix zlib context memory leak

DanielChabrowski:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:** 
Leak of memory allocated by deflateInit2 for zstream

```
Direct leak of 166656 byte(s) in 28 object(s) allocated from:                                                                                                                                                                                                  
    #0 0x7e72d3163c77 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69                                                                                                                                                                     
    #1 0x7e72d298418a in deflateInit2_ (/lib/x86_64-linux-gnu/libz.so.1+0x918a) (BuildId: b6e5d9c258e911b2e478f8dedf22400e53225509)
```

**How to test:** 

build with asan and log in

@nekiro

### DIFF
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -18,8 +18,7 @@ public:
 			std::cout << "ZLIB initialization error: " << (zstream.msg ? zstream.msg : "unknown") << std::endl;
 		}
 	}
-
-	virtual ~Protocol() { deflateEnd(&zstream); };
+	virtual ~Protocol();
 
 	// non-copyable
 	Protocol(const Protocol&) = delete;
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -18,7 +18,8 @@ public:
 			std::cout << "ZLIB initialization error: " << (zstream.msg ? zstream.msg : "unknown") << std::endl;
 		}
 	}
-	virtual ~Protocol() = default;
+
+	virtual ~Protocol() { deflateEnd(&zstream); };
 
 	// non-copyable
 	Protocol(const Protocol&) = delete;
--- a/src/protocol.cpp
+++ b/src/protocol.cpp
@@ -43,6 +43,16 @@ bool XTEA_decrypt(NetworkMessage& msg, const xtea::round_keys& key)
 
 } // namespace
 
+Protocol::~Protocol()
+{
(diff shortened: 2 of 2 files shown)


ranisalt:
What operating system and version of zlib do you have? We don't use zlib directly, it's a dependency of MySQL client libraries, so I suppose keeping those up to date avoids this leak.

DanielChabrowski:
> We don't use zlib directly, it's a dependency of MySQL client libraries, so I suppose keeping those up to date avoids this leak.

@ranisalt  Except we do, since #4870 in last December. Take a look at Protocol's constructor.

nekiro (approved):
Sure, if protocol is created for each player then it's a leak

ranisalt:
> @ranisalt Except we do, since #4870 in last December. Take a look at Protocol's constructor.

Ah for sure, I looked at a stale clone that wasn't updated 😓

src/protocol.h
```diff
@@ -18,7 +18,8 @@ class Protocol : public std::enable_shared_from_this<Protocol>
 			std::cout << "ZLIB initialization error: " << (zstream.msg ? zstream.msg : "unknown") << std::endl;
 		}
 	}
-	virtual ~Protocol() = default;
+
+	virtual ~Protocol() { deflateEnd(&zstream); };
```
ranisalt, line 22: How will this behave if `deflateInit2` fails to initialize? [Documentation indicates that it returns an error code](https://refspecs.linuxbase.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic/zlib-deflateend-1.html) so maybe wrap it with an if to output when it fails to end, too

MillhioreBT: I agree with including a verification to report errors in case they occur. `Z_DATA_ERROR` and `Z_STREAM_ERROR`

DanielChabrowski: Added couts for those two errors and moved the dtor to a .cpp file while at it.

nekiro (approved):

ArturKnopik (approved):

Zbizu (approved):
all return enums handled correctly now
function source code for context: https://github.com/madler/zlib/blob/master/deflate.c#L1258

PR #4908, opened 2025-04-04
prevent crash on addGuild.

kaleohanopahala:
its prevent crash in game.addGuild.


ArturKnopik:
LGTM

ArturKnopik (approved):

Codinablack:
> I really don't know why the tests are failing. It's works fine here.

The tests are failing because a vcpkg update recently that broke the builds. I already explained the cause and the solution to the Devs

PR #4910, opened 2025-04-06
Feat: cpplinter.lua improvements, commenting most library functions

AvajiOT:
### Pull Request Prelude

### Changes Proposed

Working off of EvilHero's initial work on the cpplinter, I did the following work;
* I went through every single function in luascript.cpp and ensuring all values to pass or returned matched properly. 
* Added aliases for constants and enums that are passed to lua to ensure tighter checks when passing variables to functions.
* Added a simple comment block to help new users of TFS understand some simple examples of how to use different classes
* Added a "source" line to nearly every line inside of cpplinter which will go directly to the function in luascript.cpp or where a constant/enum is defined. 

Understanding that luascript.cpp is quite big, and adding source lines to every single function starts making it near unmaintainable, I created a simple python script found here; https://github.com/AvajiOT/forgottenserver-cpplinter-updater which will take luascript.cpp and cpplinter.lua and update all ---@source lines for any luascript.cpp specific lines. To support this, I did add the function with a special prefix "--*" infront of the function. 

To maintain now, the function and source line needs to get added, and whenever something changes in luascript.cpp, the python script gets run and every line will be updated as required.

For updated enum/consts, I did not add that specific support, however these should be done extremely less often and can be handled with find/replace lines

Overall, I did this as I wanted to learn more about the lua language server by sumneko, https://github.com/LuaLS/lua-language-server which this required for full usage. I've tested it with VSCode and CLion, and both work extremely well. No more hovering over functions and getting zero information. 

**Issues addressed:** nil

**How to test:** Hover over lua functions with a properly typed variable, ctrl+click enums or functions, etc..

### DIFF
--- a/data/cpplinter.lua
+++ b/data/cpplinter.lua
@@ -16,8 +16,6 @@
 --- print("Client OS: " .. clientTable.os) -- Example: 2 (CLIENTOS_WINDOWS)
 --- ```
 --- <br>
----*int LuaScriptInterface::luaPlayerGetClient(lua_State* L)
----@source ../src/luascript.cpp:10844
 ---@class UserClientVersion
 ---@field version uint16 The version number of the client
 ---@field os OperatingSystem The operating system of the user 
@@ -40,8 +38,6 @@
 ---end -- As per this example, this for loop will only run once 
 --- ```
 --- <br>
----*int LuaScriptInterface::luaPlayerGetInstantSpells(lua_State* L)
----@source ../src/luascript.cpp:11032
 ---@class SpellsTable
 ---@field name string The name of the spell
 ---@field words string The words used to cast the spell
@@ -67,8 +63,6 @@
 ---end
 ---```
 ---<br>
----*int LuaScriptInterface::luaGameGetMounts(lua_State* L)
----@source ../src/luascript.cpp:4805
 ---@class MountInfoTable
 ---@field name string The name of the mount
 ---@field speed int32 The speed of the mount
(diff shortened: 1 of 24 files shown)


ArturKnopik:
i don't like idea of
`--@source ../src/luascript.cpp:XYZ`
small change in any existing function can affect most of functions location in src... a lot of work to fix source location in cpplinter.lua file...

AvajiOT:
> i don't like idea of `--@source ../src/luascript.cpp:XYZ` small change in any existing function can affect most of functions location in src... a lot of work to fix source location in cpplinter.lua file...

Yup! It would be.

Which is why a quick python script is provided to update every single line in cpplinter.lua. (https://github.com/AvajiOT/forgottenserver-cpplinter-updater) However, for this pull request it can be easily removed if running a python script is too much or seen as cumbersome. 
 
luascript.cpp is not touched very often, https://github.com/otland/forgottenserver/commits/master/src/luascript.cpp. In the past 5 months, the only change to lines was reverting the revised NPC system. The last change to luascript.cpp lines was 8b9e0b0, which added a couple lines. Again, updating this would take less then a second with the python script provided.

While using it for my own uses, it has been quite helpful to quickly go to the specific function instead of having to Ctrl+F through the file.

Overall, I did this for myself and my own use-case, and figured at least one person may benefit from the work as well. This pull request does not need to be merged, I won't be torn up about it. Wanted this to be available somewhere for people to take advantage of.

PR #4919, opened 2025-04-12
Fix Monster Visibility Behavior

kaleohanopahala:
Fix monster's behavior based on original Tibia:

**Monsters can see:**
Z 6 (one floor below — still part of the ground)
Z 7 (same floor)
Z 8 (one floor above — start of the underground)

**Monsters can’t see:**
Z 5 or lower (too far down)
Z 9 or higher (too far up)

I removed the artificial **offsetz** compensation, which only exists on the client, because I noticed it doesn’t apply to monsters.


Codinablack:
Changing it the way you have limits the player's ability for seeing as well, which affects everything else that the player can do, such as throwing items, casting spells, ect... This is not the way you should resolve the issue if you feel the monsters vision in particular is incorrect.

Codinablack:
It seems I was thinking of isSightClear or something else, my mistake. This method does not affect those other things I thought, just monsters and npcs. 

My apologies.

PR #4923, opened 2025-05-21
Add onConnect Event

CodingALS:
### Pull Request Prelude

### Changes Proposed

With this event, we can, for example:

- Send prior confirmation bytes to enable/disable features in Otclient, avoiding an unnecessary update on the client.

- Create new login verification forms with a personalized message.


Shawak:
Besides the code looking kinda sus, I would probably call this `onConnect` or something.

CodingALS:
> Besides the code looking kinda sus, I would probably call this `onConnect` or something.

It makes sense, I changed the name from "onPrevLogin" to "onConnect".

Regarding the code, there is nothing different from the standard already followed...

CodingALS:
> Besides the code looking kinda sus, I would probably call this `onConnect` or something.

It makes sense, I changed the name from "onPrevLogin" to "onConnect".

Regarding the code, there is nothing different from the standard already followed...

ArturKnopik:
i don't know is we need this feature, probably onLogin can handle most of the cases like disable/enabled otc features

CodingALS:
> i don't know is we need this feature, probably onLogin can handle most of the cases like disable/enabled otc features

Hi Artur, I understand your point of view, but the changes related to the map and the protocol must be performed before "onLogin", such as:

- GamePacketSizeU32
- GamePacketCompression
- GameChangeMapAwareRange

And that's not all, as I said before, new custom verification methods can also be created to allow Login, even applying a custom message!

CodingALS:
> i don't know is we need this feature, probably onLogin can handle most of the cases like disable/enabled otc features

Hi Artur, I understand your point of view, but the changes related to the map and the protocol must be performed before "onLogin", such as:

- GamePacketSizeU32
- GamePacketCompression
- GameChangeMapAwareRange

And that's not all, as I said before, new custom verification methods can also be created to allow Login, even applying a custom message!

ArturKnopik:
Instead of changing client settings during (pre)login, it's enough to setup client correctly and release for players, authentication can also be handled in onLogin
simple example: https://otland.net/threads/tfs-1-3-block-unofficial-clients.270246/

ArturKnopik:
You can reopen this PR. Let's wait for the opinions of the rest of the dev :)

CodingALS:
The main reason for adding this event is actually to enable/disable features through the server

Let's see what others think about this

Zbizu:
@ArturKnopik 
> i don't know is we need this feature, probably onLogin can handle most of the cases like disable/enabled otc features

metadata packets mainly, for example:
- wheel of destiny metadata (including gift of life cooldown)
- forge metadata
- bosstiary metadata
- reward shrine metadata

Zbizu (approved):
I didn't do any testing, but the implementation looks alright

the protocolgame::login is called on the game thread so the event is safe to call in that function

for the reference:
https://github.com/otland/forgottenserver/blob/8fe2872ca06bbccedaa83b7b76275a4b617dfcb5/src/protocolgame.cpp#L454

src/events.cpp
```diff
@@ -1435,6 +1438,39 @@ bool onSpellCheck(Player* player, const Spell* spell)
...
+{
+	// Player:onConnect(msg)
+	if (playerHandlers.onConnect == -1) {
+		return;
+	}
+
+	if (!tfs::lua::reserveScriptEnv()) {
+		std::cout << "[Error - events::player::onConnect] Call stack overflow" << std::endl;
+		return;
+	}
```
Zbizu, line 1451: unrelated to this PR, but these four lines appear every time a Lua event is bound in the sources
at this point it deserves a macro

gesior:
> i don't know is we need this feature, probably onLogin can handle most of the cases like disable/enabled otc features

It's not only about 'when it's executed', but also 'if it's executed'.
`onLogin` does not execute, when you login into character that is already online in game ex. you lost internet connection for 30 seconds (client shows characters list) and character has PZ lock for 15 minutes, so it won't be kicked. Game client changes to new, but player in game did not 'login', so it won't execute `onLogin` (and it should not, as player did not really login [did not execute 'logout' before]).

ArturKnopik:
> > i don't know is we need this feature, probably onLogin can handle most of the cases like disable/enabled otc features
> 
> It's not only about 'when it's executed', but also 'if it's executed'. `onLogin` does not execute, when you login into character that is already online in game ex. you lost internet connection for 30 seconds (client shows characters list) and character has PZ lock for 15 minutes, so it won't be kicked. Game client changes to new, but player in game did not 'login', so it won't execute `onLogin` (and it should not, as player did not really login [did not execute 'logout' before]).
for this scenario you can use "onReconnect" creature event

Zbizu:
@gesior 
> @CodingALS Can this event replace `onReconnect` event?

They should be separate.
onConnect triggers early and might be too early to send metadata packets. It can also deny the connection when the Lua script returns some message.

MillhioreBT (commented):
In general, I don't dislike the new onConnect event; it has advantages when using OTC as a client.  

However, coupling the event to the player doesn't seem to be the most consistent approach for this method.  

I think it would be better to push the `Operating System` and `Character ID` to the event.  

Then, we can use the `Player` constructor to find the player if they exist and determine whether it's an `onReconnect` or an `onLogin`.

It is true that we can also push the `player` or `nil` in case it is a new object, and thus also determine if it is `onReconnect` or `onLogin`, but the most common case will be `onLogin`, so it is better and cheaper to push a `number` to Lua, and if the user wants to use player information, they can do it manually in the script, in the end it will be a safer method.

CodingALS:
@MillhioreBT 

They should be separate things, onConnect, onReconnect and onLogin, they are three different situations

Do you think the onConnect function should be a creatureEvent similar to the onReconnect function?

CodingALS:
Guys, one thing you may be confusing is that these are three completely different situations:

onConnect: the player has not logged in yet, so there is no way to analyze either the Login or a possible Reconnect (if the player has not logged in, how could he "Reconnect"?)

onLogin: when the player passes the onConnect check, he finally logs in and then the "onLogin" event can be used

onReconnect: the player has already logged in, and then the "onReconnect" event is used

CodingALS:
Remembering that both "onConnect" and "onReconnect" must send their bytes to the client, so that better performance is possible in OTClient

@Zbizu @MillhioreBT See my modification, I moved the "onReconnect" call before "sendAddCreature", so that we can obtain this "better performance" related to OTClient.

PR #4928, opened 2025-06-06
Fix potential crashes

kaleohanopahala:


src/player.cpp
```diff
@@ -3148,6 +3148,10 @@ void Player::postAddNotification(Thing* thing, const Cylinder* oldParent, int32_
 
 			for (const auto& it : openContainers) {
 				Container* container = it.second.container;
+				if (container == nullptr) {
```
MillhioreBT, line 3151: This check is unnecessary; pointers are not useful for validating whether a pointer points to valid memory. There is no instance where an openContainer is assigned a Container that is nullptr.

src/game.cpp
```diff
@@ -1393,6 +1393,10 @@ ReturnValue Game::internalAddItem(Cylinder* toCylinder, Item* item, int32_t inde
 
 ReturnValue Game::internalRemoveItem(Item* item, int32_t count /*= -1*/, bool test /*= false*/, uint32_t flags /*= 0*/)
 {
+	if (!item) {
```
MillhioreBT, line 1396: This check is not necessary, since internalRemoveItem never receives nullptr, it only receives pointers (or shoulded) and pointers do not work to validate if it is pointing to valid memory.

in any case an assert is much more appropriate.

PR #4929, opened 2025-06-08
Safe Handling of Null Function Pointer in fireAddRemItem

kaleohanopahala:
it's ensure they are valid before calling them. If moveFunction is not set (i.e., is nullptr), directly invoking it will cause undefined behavior, most likely resulting in a crash.

### DIFF
--- a/src/movement.cpp
+++ b/src/movement.cpp
@@ -993,7 +993,12 @@ uint32_t MoveEvent::fireAddRemItem(Item* item, Item* tileItem, const Position& p
 	if (scripted) {
 		return executeAddRemItem(item, tileItem, pos);
 	}
-	return moveFunction(item, tileItem, pos);
+
+	if (moveFunction) {
+		return moveFunction(item, tileItem, pos);
+	}
+
+	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


MillhioreBT:
This scenario should never occur, because in Lua the event is already added by default when you [assign a function to the event](https://github.com/otland/forgottenserver/blob/master/data/lib/compat/compat.lua#L206-L236), and in the case of XML, an event that doesn't have a [configured type](https://github.com/otland/forgottenserver/blob/master/src/movement.cpp#L890-L917) won't be accepted.

So this will never happen according to the established API.

ghost:
> This scenario should never occur, because in Lua the event is already added by default when you [assign a function to the event](https://github.com/otland/forgottenserver/blob/master/data/lib/compat/compat.lua#L206-L236), and in the case of XML, an event that doesn't have a [configured type](https://github.com/otland/forgottenserver/blob/master/src/movement.cpp#L890-L917) won't be accepted.
> 
> So this will never happen according to the established API.

I see that it has zero real cost, but it helps prevent potential crashes or undefined behavior in case something unexpected happens especially if the code is modified in custom servers.

PR #4931, opened 2025-06-11
Add Missing MySQL Variables

GooseWithAKnife:
### Pull Request Prelude

### Changes Proposed

Basic MySQL Variables are missing from the config.lua.dist

**Issues addressed:** 

**How to test:** 

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -37,6 +37,16 @@ replaceKickOnLogin = true
 maxPacketsPerSecond = 25
 enableTwoFactorAuth = true
 
+-- MySQL
+mysqlHost = "hostname"
+mysqlUser = "mysql_user"
+mysqlPass = "mysql_password"
+mysqlDatabase = "database_name"
+mysqlDatabaseBackup = false
+mysqlPort = 3306
+mysqlSock = ""
+passwordType = "sha1"
+
 -- Pathfinding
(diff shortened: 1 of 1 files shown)


GooseWithAKnife:
> they are not missing, they were removed intentionally: [4bc7f61](https://github.com/otland/forgottenserver/commit/4bc7f619d5ee8add9ce8e7362b12d418d1b348c4)
> 
> but yeah there are improvements to be made as of @gesior comments here: [#4518 (comment)](https://github.com/otland/forgottenserver/pull/4518#issuecomment-2486773326)

Having them removed prevents new users from setting a server up without googling and guessing based on other config.lua's
No release should be marked as "stable" in an incomplete, not ready-to-run state.

The starting of the server expects a config.lua to be present, and most users will copy the .dist file and adjust the file extension to be able to run the server.

ranisalt:
These values are described in `.env.example` that is shipped and shouldn't be ignored. How about you add these entries back but instead of defaulting to a possibly invalid value, load the environment variables using [`os.getenv`](https://www.lua.org/pil/22.2.html) and add a note that this is the wrong place to set them?

ArturKnopik:
there is no "passwordType" field
the rest was handled by https://github.com/otland/forgottenserver/commit/c519cc6e6f9864e933c8a5a94d20afee538e70de

PR #4933, opened 2025-06-18
Weapons Nullchecks

hack3dahck:
missing nullcheks ?

@MillhioreBT

### DIFF
--- a/src/weapons.cpp
+++ b/src/weapons.cpp
@@ -808,7 +808,7 @@ bool WeaponDistance::useWeapon(Player* player, Item* item, Creature* target) con
 
 int32_t WeaponDistance::getElementDamage(const Player* player, const Creature* target, const Item* item) const
 {
-	if (elementType == COMBAT_NONE) {
+	if (!player || !item || elementType == COMBAT_NONE) {
 		return 0;
 	}
 
@@ -839,6 +839,10 @@ int32_t WeaponDistance::getElementDamage(const Player* player, const Creature* t
 int32_t WeaponDistance::getWeaponDamage(const Player* player, const Creature* target, const Item* item,
                                         bool maxDamage /*= false*/) const
 {
+	if (!player || !item) {
+		return 0;
+	}
+		
 	int32_t attackValue = item->getAttack();
 
 	if (item->getWeaponType() == WEAPON_AMMO) {


ArturKnopik:
we can leave it as is, if the compiler determines that the call always has a valid pointer, it may remove those checks at compile time

gesior:
> missing nullcheks ?
> 
> @MillhioreBT

It's OTS. Highly efficient C++ application for 1000+ online players in a single C++ thread.
Probably missing 1k or 10k “does the object exist” checks, but it's by design. Don't report random AI problems.

@hack3dahck 
If you really want to report some problem, report what we have to do to make server crash (how to replicate this). If we cannot crash server with default config, it's fine.
AI bug reports are useless, please ignore them and don't create Issues/PRs.

I'm pretty sure we can close thais PR. It looks like random AI problem.

hack3dahck:
> > missing nullcheks ?
> > @MillhioreBT
> 
> It's OTS. Highly efficient C++ application for 1000+ online players in a single C++ thread. Probably missing 1k or 10k “does the object exist” checks, but it's by design. Don't report random AI problems.
> 
> @hack3dahck If you really want to report some problem, report what we have to do to make server crash (how to replicate this). If we cannot crash server with default config, it's fine. AI bug reports are useless, please ignore them and don't create Issues/PRs.
> 
> I'm pretty sure we can close thais PR. It looks like random AI problem.

Nice try, but nah.
Need tests, myb tomorrow ill do.

gesior:
> Nice try, but nah.
> Need tests, myb tomorrow ill do.

Ok. Post me scenario in which I can crash/memory leak official TFS (modified config or .otbm). Step by step.
If you cannot, close PR.
If you cannot, close your account and stop posting AI crap.

hack3dahck:
> > Nice try, but nah.
> > Need tests, myb tomorrow ill do.
> 
> Ok. Post me scenario in which I can crash/memory leak official TFS (modified config or .otbm). Step by step. If you cannot, close PR. If you cannot, close your account and stop posting AI crap.

Dnt cry and be better 🤫

Mirkaanks:
This checks are unnecessary here because the engine guarantees valid input. 
These functions are only called from controlled where all already validated, ensuring no null pointers or invalid.

@hack3dahck It might not be super useful, but it’s always nice when someone tries to help, after all, not everyone’s a supreme expert like @gesior

Don’t fight the AI you won’t win. 😄

ArturKnopik:
We should introduce some performance tests - we are talking about something without knowing whether it has any impact on performance. This check likely has no impact(up to 20 calls/s per player), and if correct input is guaranteed, the compiler will eliminate the check anyway.

MillhioreBT:
Closed because it's just an unnecessary check
While adding the check doesn't hurt anyone, what we want is to avoid writing **redundant code** and maintain the **format and style of TFS**.

In case this is a **real error**, please **reopen the PR** and **explain the steps to reproduce the null check here**.
If possible, also expose the **call stack of the crash** in case it occurs.

Thank you.

PR #4934, opened 2025-06-20
Pathfinding Changes

### LINKED ISSUE #4884: [Bug]: Fleeing creatures aren't properly taking distance steps

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below)

### OS Description

_No response_

### Bug description

Fleeing monsters arent properly taking distance steps

_No response_

### Steps to reproduce

1. Create rabbit and check how it runs from the target

### Actual Behavior

Monster takes a step every 1 second instead of the delay of step duration (it doesn't run away smoothly)
This works fine for creatures that are running on low health, like rat or dragon

https://github.com/user-attachments/assets/1882a4e6-d038-41b8-b403-2d2a8196fc6e

### Expected Behavior

Monster should seamlessly run from fleeing target as long as target is not near (1 sqm)

### Backtrace

_No response_

NRH-AA:
Performance enhancements to pathfinding algorithm. Remove new from node construction to help with memory problems.

1) Nodes are stored in each data structure using a hash of their x,y coordinates for quick look up.
2) Added a visited set so we do not repeat nodes.
3) Moved open nodes to a priority queue that stores them based on their f value for faster f value look up. (No more harsh sorting)
4) Added a sightline check to optimize clear paths performance.
5) Optimized heuristic calculation which increased performance on all paths.
6) Fixes #4884

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -653,7 +653,7 @@ uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
 	uint16_t dx = std::abs(p1.getX() - p2.getX());
 	uint16_t dy = std::abs(p1.getY() - p2.getY());
-	return 10 * (dx + dy) - 6 * std::min(dx, dy); // Octile distance
+	return 10 * (dx + dy);
 }
 
 bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::vector<Direction>& dirList,
@@ -690,7 +690,6 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	    {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}}};
 
 	Position endPos;
-	bool sightClear = false;
 	AStarNodes nodes(pos.x , pos.y);
 
 	AStarNode* found = nullptr;
@@ -728,43 +727,6 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 
 			if (fpp.keepDistance && !pathCondition.isInRange(startPos, pos, fpp)) {
 				continue;
-			}
-
-			// If sight is clear we can ignore a lot of nodes.
-			if (sightClear) {
-				int32_t startX = startPos.getX();
-				int32_t startY = startPos.getY();
-				int32_t targetX = targetPos.getX();
-				int32_t targetY = targetPos.getY();
-
-				// We don't need nodes behind us.
-				// We also do not need nodes on a different x or y if target and start is same x/y.
-				if (startX > targetX && pos.x > startX) {
-					continue;
-				} else if (startX == targetX && pos.x != startX) {
-					continue;
-				} else if (startX < targetX && pos.x < startX) {
-					continue;
-				}
-				if (startY > targetY && pos.y > startY) {
-					continue;
-				} else if (startY == targetY && pos.y != startY) {
-					continue;
-				} else if (startY < targetY && pos.y < startY) {
-					continue;
-				}
-
-				// We don't need nodes past the targetPos
-				if (startX > targetX && pos.x < targetX) {
-					continue;
-				} else if (startX < targetX && pos.x > targetX) {
-					continue;
-				}
-				if (startY > targetY && pos.y < targetY) {
-					continue;
-				} else if (startY < targetY && pos.y > targetY) {
-					continue;
-				}
 			}
 
 			const Tile* tile;
@@ -849,7 +811,10 @@ AStarNodes::AStarNodes(uint16_t x, uint16_t y) : nodes(), nodeMap()
 {
 	// Needs to be large enough to never resize 250 should be plenty
 	// If you want paths larger than 20-30 sqm this must be increased.
-	nodes.reserve(250);
+	uint8_t reserveSize = 250;
+	nodes.reserve(reserveSize);
+	nodeMap.reserve(reserveSize);
+	visited.reserve(reserveSize);
 	createNode(nullptr, x, y, 0, 0);
 }
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -691,6 +691,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	    {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}}};
 
 	Position endPos;
+	bool sightClear = isSightClear(startPos, targetPos, true);
 	AStarNodes nodes(pos.x , pos.y);
 
 	AStarNode* found = nullptr;
@@ -728,6 +729,43 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 
 			if (fpp.keepDistance && !pathCondition.isInRange(startPos, pos, fpp)) {
 				continue;
+			}
+
+			// If sight is clear we can ignore a lot of nodes.
+			if (sightClear) {
+				int32_t startX = startPos.getX();
+				int32_t startY = startPos.getY();
+				int32_t targetX = targetPos.getX();
+				int32_t targetY = targetPos.getY();
+
+				// We don't need nodes behind us.
+				// We also do not need nodes on a different x or y if target and start is same x/y.
+				if (startX > targetX && pos.x > startX) {
+					continue;
+				} else if (startX == targetX && pos.x != startX) {
+					continue;
+				} else if (startX < targetX && pos.x < startX) {
+					continue;
+				}
+				if (startY > targetY && pos.y > startY) {
+					continue;
+				} else if (startY == targetY && pos.y != startY) {
+					continue;
+				} else if (startY < targetY && pos.y < startY) {
+					continue;
+				}
+
+				// We don't need nodes past the targetPos
+				if (startX > targetX && pos.x < targetX) {
+					continue;
+				} else if (startX < targetX && pos.x > targetX) {
+					continue;
+				}
+				if (startY > targetY && pos.y < targetY) {
+					continue;
+				} else if (startY < targetY && pos.y > targetY) {
+					continue;
+				}
 			}
 
 			const Tile* tile;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -668,7 +668,7 @@ static uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
 	uint16_t dx = std::abs(p1.getX() - p2.getX());
 	uint16_t dy = std::abs(p1.getY() - p2.getY());
-	return 10 * (dx + dy);
+	return MAP_NORMALWALKCOST * (dx + dy);
 }
 
 bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::vector<Direction>& dirList,
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -664,11 +664,11 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
 	return tile;
 }
 
-uint16_t calculateHeuristic(const Position& p1, const Position& p2)
+static uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
 	uint16_t dx = std::abs(p1.getX() - p2.getX());
 	uint16_t dy = std::abs(p1.getY() - p2.getY());
-	return 10 * (dx + dy);
+	return MAP_NORMALWALKCOST * (dx + dy) + (MAP_DIAGONALWALKCOST - 2 * 10) * std::min(dx, dy);
 }
 
 bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::vector<Direction>& dirList,
@@ -865,7 +865,7 @@ AStarNodes::AStarNodes(uint16_t x, uint16_t y) : nodes(), nodeMap()
 {
 	// Needs to be large enough to never resize 250 should be plenty
 	// If you want paths larger than 20-30 sqm this must be increased.
-	uint8_t reserveSize = 250;
+	uint8_t reserveSize = 150;
 	nodes.reserve(reserveSize);
 	nodeMap.reserve(reserveSize);
 	visited.reserve(reserveSize);
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -707,7 +707,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	bool sightClear = isSightClear(startPos, targetPos, true, true);
 
 	Position endPos;
-	AStarNodes nodes(pos.x , pos.y);
+	AStarNodes nodes(pos.x, pos.y);
 
 	AStarNode* found = nullptr;
 	int32_t bestMatch = 0;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -690,7 +690,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	    {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}}};
 
 	Position endPos;
-	bool sightClear = isSightClear(startPos, targetPos, true);
+	bool sightClear = false;
 	AStarNodes nodes(pos.x , pos.y);
 
 	AStarNode* found = nullptr;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -661,7 +661,6 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
                           const FrozenPathingConditionCall& pathCondition, const FindPathParams& fpp) const
 {
 	Position pos = creature.getPosition();
-	Position endPos;
 	const Position startPos = pos;
 
 	// We can't walk, no need to create path.
@@ -674,24 +673,25 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		return false;
 	}
 
+	uint16_t distanceX = startPos.getDistanceX(targetPos);
+	uint16_t distanceY = startPos.getDistanceY(targetPos);
 	// We are next to our target. Let dance step decide.
-	if (fpp.maxTargetDist <= 1 && startPos.getDistanceX(targetPos) <= 1 && startPos.getDistanceY(targetPos) <= 1) {
+	if (fpp.maxTargetDist <= 1 && distanceX <= 1 && distanceY <= 1) {
 		return true;
 	}
 
 	// Don't update path. The target is too far away.
 	int32_t maxDistanceX = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxClientViewportX + 1;
 	int32_t maxDistanceY = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxClientViewportY + 1;
-	if (startPos.getDistanceX(targetPos) > maxDistanceX || startPos.getDistanceY(targetPos) > maxDistanceY) {
+	if (distanceX > maxDistanceX || distanceY > maxDistanceY) {
 		return false;
 	}
 
 	static constexpr std::array<std::pair<int, int>, 8> allNeighbors = {
 	    {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}}};
 
-	uint16_t width = startPos.getDistanceX(targetPos);
-	uint16_t height = startPos.getDistanceY(targetPos);
-	AStarNodes nodes(width, height);
+	Position endPos;
+	AStarNodes nodes(pos.x , pos.y);
 
 	AStarNode* found = nullptr;
 	int32_t bestMatch = 0;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -651,10 +651,9 @@ const Tile* Map::canWalkTo(const Creature& creature, const Position& pos) const
 
 uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
-	uint16_t dx = p1.getX() - p2.getX();
-	uint16_t dy = p1.getY() - p2.getY();
-
-	return dx * dx + dy * dy;
+	uint16_t dx = std::abs(p1.getX() - p2.getX());
+	uint16_t dy = std::abs(p1.getY() - p2.getY());
+	return 10 * (dx + dy) - 6 * std::min(dx, dy); // Octile distance
 }
 
 bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::vector<Direction>& dirList,
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -687,16 +687,16 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		return false;
 	}
 
-	uint16_t distanceX = startPos.getDistanceX(targetPos);
-	uint16_t distanceY = startPos.getDistanceY(targetPos);
+	int32_t distanceX = startPos.getDistanceX(targetPos);
+	int32_t distanceY = startPos.getDistanceY(targetPos);
 	// We are next to our target. Let dance step decide.
 	if (fpp.maxTargetDist <= 1 && distanceX <= 1 && distanceY <= 1) {
 		return true;
 	}
 
 	// Don't update path. The target is too far away.
-	int32_t maxDistanceX = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxClientViewportX + 1;
-	int32_t maxDistanceY = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxClientViewportY + 1;
+	int32_t maxDistanceX = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxViewportX + 1;
+	int32_t maxDistanceY = fpp.maxSearchDist ? fpp.maxSearchDist : Map::maxViewportY + 1;
 	if (distanceX > maxDistanceX || distanceY > maxDistanceY) {
 		return false;
 	}
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -810,7 +810,9 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 // AStarNodes
 AStarNodes::AStarNodes(uint16_t x, uint16_t y) : nodes(), nodeMap()
 {
-	nodes.reserve(static_cast<size_t>(x * y));
+	// Needs to be large enough to never resize 250 should be plenty
+	// If you want paths larger than 20-30 sqm this must be increased.
+	nodes.reserve(250);
 	createNode(nullptr, x, y, 0, 0);
 }
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -711,12 +711,12 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 
 	AStarNode* found = nullptr;
 	int32_t bestMatch = 0;
-	uint8_t iterations = 0;
+	uint16_t iterations = 0;
 	AStarNode* n = nodes.getBestNode();
 	while (n) {
 		iterations++;
 
-		if (iterations >= Map::nodeReserveSize) {
+		if (iterations >= Map::maxViewportX * Map::maxViewportY) {
 			return false;
 		}
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -717,7 +717,6 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		iterations++;
 
 		if (iterations >= Map::nodeReserveSize) {
-			nodes.clear();
 			return false;
 		}
 
@@ -812,7 +811,10 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 				neighborNode->parent = n;
 			} else {
 				// Does not exist in the open/closed list, create a new node
-				nodes.createNode(n, pos.x, pos.y, g, newf);
+				if (!nodes.createNode(n, pos.x, pos.y, g, newf)) {
+					// Limit of nodes reached
+					return false;
+				}
 			}
 		}
 
@@ -820,7 +822,6 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	}
 
 	if (!found) {
-		nodes.clear();
 		return false;
 	}
 
@@ -859,7 +860,6 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		found = found->parent;
 	}
 
-	nodes.clear();
 	return true;
 }
 
@@ -872,18 +872,12 @@ AStarNodes::AStarNodes(uint16_t x, uint16_t y) : nodes(), nodeMap()
 	createNode(nullptr, x, y, 0, 0);
 }
 
-void AStarNodes::clear()
-{
-	nodes.clear();
-	visited.clear();
-	nodeMap.clear();
-	while (!openSet.empty()) {
-		openSet.pop();
-	}
-}
-
 AStarNode* AStarNodes::createNode(AStarNode* parent, uint16_t x, uint16_t y, uint16_t g, uint16_t f)
 {
+	if (nodes.size() == Map::nodeReserveSize) {
+		return nullptr;
+	}
+
 	uint32_t key = hashCoord(x, y);
 	nodes.emplace_back(AStarNode{parent, x, y, g, f});
 	AStarNode* node = &nodes.back();
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -689,7 +689,9 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	static constexpr std::array<std::pair<int, int>, 8> allNeighbors = {
 	    {{-1, 0}, {0, 1}, {1, 0}, {0, -1}, {-1, -1}, {1, -1}, {1, 1}, {-1, 1}}};
 
-	AStarNodes nodes(maxDistanceX, maxDistanceY);
+	uint16_t width = startPos.getDistanceX(targetPos);
+	uint16_t height = startPos.getDistanceY(targetPos);
+	AStarNodes nodes(width, height);
 
 	AStarNode* found = nullptr;
 	int32_t bestMatch = 0;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -717,6 +717,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		iterations++;
 
 		if (iterations >= Map::nodeReserveSize) {
+			nodes.clear();
 			return false;
 		}
 
@@ -819,6 +820,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	}
 
 	if (!found) {
+		nodes.clear();
 		return false;
 	}
 
@@ -856,6 +858,8 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 
 		found = found->parent;
 	}
+
+	nodes.clear();
 	return true;
 }
 
@@ -868,18 +872,21 @@ AStarNodes::AStarNodes(uint16_t x, uint16_t y) : nodes(), nodeMap()
 	createNode(nullptr, x, y, 0, 0);
 }
 
-AStarNodes::~AStarNodes()
+void AStarNodes::clear()
 {
-	for (auto* node : nodes) {
-		delete node;
+	nodes.clear();
+	visited.clear();
+	nodeMap.clear();
+	while (!openSet.empty()) {
+		openSet.pop();
 	}
 }
 
 AStarNode* AStarNodes::createNode(AStarNode* parent, uint16_t x, uint16_t y, uint16_t g, uint16_t f)
 {
 	uint32_t key = hashCoord(x, y);
-	nodes.emplace_back(new AStarNode(parent, x, y, g, f));
-	AStarNode* node = nodes.back();
+	nodes.emplace_back(AStarNode{parent, x, y, g, f});
+	AStarNode* node = &nodes.back();
 	nodeMap[key] = node;
 	openSet.push(node);
 	return node;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -866,12 +866,9 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 // AStarNodes
 AStarNodes::AStarNodes(uint16_t x, uint16_t y) : nodes(), nodeMap()
 {
-	// Needs to be large enough to never resize 250 should be plenty
-	// If you want paths larger than 20-30 sqm this must be increased.
-	uint8_t reserveSize = 150;
-	nodes.reserve(reserveSize);
-	nodeMap.reserve(reserveSize);
-	visited.reserve(reserveSize);
+	nodes.reserve(Map::nodeReserveSize);
+	nodeMap.reserve(Map::nodeReserveSize);
+	visited.reserve(Map::nodeReserveSize);
 	createNode(nullptr, x, y, 0, 0);
 }
 
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -716,7 +716,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 	while (n) {
 		iterations++;
 
-		if (iterations >= 120) {
+		if (iterations >= Map::nodeReserveSize) {
 			nodes.clear();
 			return false;
 		}
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -737,12 +737,11 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 			pos.x = x + allNeighbors[i].first;
 			pos.y = y + allNeighbors[i].second;
 
-			if (fpp.maxSearchDist != 0 &&
-			    (startPos.getDistanceX(pos) > fpp.maxSearchDist || startPos.getDistanceY(pos) > fpp.maxSearchDist)) {
+			int32_t startDistanceToNode = startPos.getDistanceX(pos) + startPos.getDistanceY(pos);
+			if (fpp.maxSearchDist != 0 && startDistanceToNode > fpp.maxSearchDist) {
 				continue;
-			} else if ((startPos.getDistanceX(pos) + startPos.getDistanceY(pos) >
-			            Map::maxViewportX + Map::maxViewportY)) {
-				break;
+			} else if (fpp.maxSearchDist == 0 && (startDistanceToNode > (Map::maxViewportX + Map::maxViewportY))) {
+				continue;
 			}
 
 			if (fpp.keepDistance && !pathCondition.isInRange(startPos, pos, fpp)) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -668,7 +668,7 @@ static uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
 	uint16_t dx = std::abs(p1.getX() - p2.getX());
 	uint16_t dy = std::abs(p1.getY() - p2.getY());
-	return MAP_NORMALWALKCOST * (dx + dy) + (MAP_DIAGONALWALKCOST - 2 * 10) * std::min(dx, dy);
+	return 10 * (dx + dy);
 }
 
 bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::vector<Direction>& dirList,
@@ -740,6 +740,9 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 			if (fpp.maxSearchDist != 0 &&
 			    (startPos.getDistanceX(pos) > fpp.maxSearchDist || startPos.getDistanceY(pos) > fpp.maxSearchDist)) {
 				continue;
+			} else if ((startPos.getDistanceX(pos) + startPos.getDistanceY(pos) >
+			            Map::maxViewportX + Map::maxViewportY)) {
+				break;
 			}
 
 			if (fpp.keepDistance && !pathCondition.isInRange(startPos, pos, fpp)) {
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -699,6 +699,7 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 		iterations++;
 
 		if (iterations >= 120) {
+			nodes.clear();
 			return false;
 		}
 
@@ -752,15 +753,16 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
 				neighborNode->g = g;
 				neighborNode->f = newf;
 				neighborNode->parent = n;
-				nodes.addNode(neighborNode);
 			} else {
 				// Does not exist in the open/closed list, create a new node
-				nodes.createNewNode(n, pos.x, pos.y, g, newf);
+				nodes.createNode(n, pos.x, pos.y, g, newf);
 			}
 		}
 
 		n = nodes.getBestNode();
 	}
+
+	nodes.clear();
 
 	if (!found) {
 		return false;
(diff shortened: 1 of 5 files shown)


NRH-AA:
Here are some speed tests based on the new algorithm. When I ran the current one I would see up to ~80k nanoseconds to execute. This also DOES NOT USE: NEW to create nodes so memory leaks shouldn't be a huge issue anymore.

[DISTANCE]: 6 [Calculated In]: 14 Iterations and 9600 nanoseconds
[DISTANCE]: 10 [Calculated In]: 37 Iterations and 20800 nanoseconds
[DISTANCE]: 9 [Calculated In]: 39 Iterations and 21000 nanoseconds
[DISTANCE]: 6 [Calculated In]: 21 Iterations and 13900 nanoseconds
[DISTANCE]: 4 [Calculated In]: 8 Iterations and 6200 nanoseconds
[DISTANCE]: 4 [Calculated In]: 5 Iterations and 5200 nanoseconds
[DISTANCE]: 2 [Calculated In]: 2 Iterations and 1500 nanoseconds
[DISTANCE]: 4 [Calculated In]: 7 Iterations and 5900 nanoseconds
[DISTANCE]: 4 [Calculated In]: 4 Iterations and 3300 nanoseconds
[DISTANCE]: 6 [Calculated In]: 18 Iterations and 9900 nanoseconds
[DISTANCE]: 7 [Calculated In]: 19 Iterations and 9700 nanoseconds
[DISTANCE]: 6 [Calculated In]: 12 Iterations and 10200 nanoseconds
[DISTANCE]: 4 [Calculated In]: 8 Iterations and 8400 nanoseconds
[DISTANCE]: 5 [Calculated In]: 8 Iterations and 6900 nanoseconds
[DISTANCE]: 6 [Calculated In]: 18 Iterations and 13800 nanoseconds
[DISTANCE]: 8 [Calculated In]: 28 Iterations and 19500 nanoseconds
[DISTANCE]: 4 [Calculated In]: 6 Iterations and 5400 nanoseconds
[DISTANCE]: 9 [Calculated In]: 34 Iterations and 33000 nanoseconds
[DISTANCE]: 7 [Calculated In]: 25 Iterations and 19900 nanoseconds
[DISTANCE]: 4 [Calculated In]: 3 Iterations and 3800 nanoseconds
[DISTANCE]: 4 [Calculated In]: 8 Iterations and 7600 nanoseconds
[DISTANCE]: 7 [Calculated In]: 16 Iterations and 19000 nanoseconds
[DISTANCE]: 8 [Calculated In]: 20 Iterations and 13100 nanoseconds
[DISTANCE]: 7 [Calculated In]: 26 Iterations and 14200 nanoseconds
[DISTANCE]: 9 [Calculated In]: 32 Iterations and 18500 nanoseconds
[DISTANCE]: 8 [Calculated In]: 25 Iterations and 12900 nanoseconds
[DISTANCE]: 7 [Calculated In]: 22 Iterations and 11700 nanoseconds
[DISTANCE]: 6 [Calculated In]: 21 Iterations and 12400 nanoseconds
[DISTANCE]: 5 [Calculated In]: 13 Iterations and 12000 nanoseconds
[DISTANCE]: 3 [Calculated In]: 2 Iterations and 1600 nanoseconds
[DISTANCE]: 4 [Calculated In]: 5 Iterations and 4600 nanoseconds
[DISTANCE]: 3 [Calculated In]: 2 Iterations and 1700 nanoseconds
[DISTANCE]: 6 [Calculated In]: 6 Iterations and 3700 nanoseconds
[DISTANCE]: 11 [Calculated In]: 35 Iterations and 19000 nanoseconds
[DISTANCE]: 5 [Calculated In]: 12 Iterations and 11100 nanoseconds
[DISTANCE]: 6 [Calculated In]: 15 Iterations and 11400 nanoseconds
[DISTANCE]: 9 [Calculated In]: 26 Iterations and 15300 nanoseconds
[DISTANCE]: 8 [Calculated In]: 38 Iterations and 19000 nanoseconds
[DISTANCE]: 8 [Calculated In]: 26 Iterations and 14400 nanoseconds
[DISTANCE]: 6 [Calculated In]: 16 Iterations and 13000 nanoseconds
[DISTANCE]: 8 [Calculated In]: 17 Iterations and 12900 nanoseconds
[DISTANCE]: 4 [Calculated In]: 8 Iterations and 6500 nanoseconds
[DISTANCE]: 4 [Calculated In]: 6 Iterations and 3500 nanoseconds
[DISTANCE]: 4 [Calculated In]: 5 Iterations and 3600 nanoseconds
[DISTANCE]: 6 [Calculated In]: 18 Iterations and 10900 nanoseconds
[DISTANCE]: 6 [Calculated In]: 9 Iterations and 6700 nanoseconds
[DISTANCE]: 6 [Calculated In]: 13 Iterations and 12900 nanoseconds
[DISTANCE]: 3 [Calculated In]: 5 Iterations and 11800 nanoseconds
[DISTANCE]: 6 [Calculated In]: 15 Iterations and 13000 nanoseconds
[DISTANCE]: 8 [Calculated In]: 19 Iterations and 11200 nanoseconds
[DISTANCE]: 10 [Calculated In]: 61 Iterations and 34500 nanoseconds
[DISTANCE]: 9 [Calculated In]: 39 Iterations and 21300 nanoseconds
[DISTANCE]: 11 [Calculated In]: 34 Iterations and 16900 nanoseconds
[DISTANCE]: 10 [Calculated In]: 38 Iterations and 18600 nanoseconds
[DISTANCE]: 8 [Calculated In]: 11 Iterations and 6300 nanoseconds
[DISTANCE]: 7 [Calculated In]: 15 Iterations and 8500 nanoseconds
[DISTANCE]: 7 [Calculated In]: 19 Iterations and 9200 nanoseconds
[DISTANCE]: 5 [Calculated In]: 6 Iterations and 5700 nanoseconds
[DISTANCE]: 3 [Calculated In]: 2 Iterations and 2200 nanoseconds
[DISTANCE]: 3 [Calculated In]: 5 Iterations and 5300 nanoseconds
[DISTANCE]: 5 [Calculated In]: 13 Iterations and 9900 nanoseconds
[DISTANCE]: 8 [Calculated In]: 27 Iterations and 15100 nanoseconds
[DISTANCE]: 8 [Calculated In]: 33 Iterations and 16000 nanoseconds
[DISTANCE]: 12 [Calculated In]: 51 Iterations and 23000 nanoseconds
[DISTANCE]: 11 [Calculated In]: 34 Iterations and 30600 nanoseconds
[DISTANCE]: 6 [Calculated In]: 14 Iterations and 27200 nanoseconds
[DISTANCE]: 8 [Calculated In]: 24 Iterations and 21300 nanoseconds
[DISTANCE]: 11 [Calculated In]: 34 Iterations and 25300 nanoseconds
[DISTANCE]: 10 [Calculated In]: 43 Iterations and 25700 nanoseconds
[DISTANCE]: 10 [Calculated In]: 30 Iterations and 17800 nanoseconds
[DISTANCE]: 8 [Calculated In]: 17 Iterations and 14200 nanoseconds
[DISTANCE]: 10 [Calculated In]: 16 Iterations and 10800 nanoseconds
[DISTANCE]: 5 [Calculated In]: 7 Iterations and 6600 nanoseconds
[DISTANCE]: 6 [Calculated In]: 9 Iterations and 7200 nanoseconds
[DISTANCE]: 4 [Calculated In]: 4 Iterations and 4200 nanoseconds
[DISTANCE]: 6 [Calculated In]: 11 Iterations and 6500 nanoseconds
[DISTANCE]: 3 [Calculated In]: 2 Iterations and 2100 nanoseconds
[DISTANCE]: 8 [Calculated In]: 17 Iterations and 10800 nanoseconds
[DISTANCE]: 8 [Calculated In]: 21 Iterations and 11400 nanoseconds
[DISTANCE]: 7 [Calculated In]: 12 Iterations and 16500 nanoseconds
[DISTANCE]: 4 [Calculated In]: 5 Iterations and 5600 nanoseconds
[DISTANCE]: 7 [Calculated In]: 22 Iterations and 15400 nanoseconds
[DISTANCE]: 9 [Calculated In]: 21 Iterations and 12300 nanoseconds
[DISTANCE]: 10 [Calculated In]: 52 Iterations and 31700 nanoseconds
[DISTANCE]: 10 [Calculated In]: 41 Iterations and 20100 nanoseconds
[DISTANCE]: 12 [Calculated In]: 38 Iterations and 19100 nanoseconds
[DISTANCE]: 11 [Calculated In]: 44 Iterations and 19500 nanoseconds
[DISTANCE]: 9 [Calculated In]: 22 Iterations and 10500 nanoseconds
[DISTANCE]: 8 [Calculated In]: 19 Iterations and 9600 nanoseconds
[DISTANCE]: 8 [Calculated In]: 11 Iterations and 6800 nanoseconds
[DISTANCE]: 6 [Calculated In]: 8 Iterations and 4200 nanoseconds
[DISTANCE]: 4 [Calculated In]: 5 Iterations and 3900 nanoseconds
[DISTANCE]: 4 [Calculated In]: 6 Iterations and 5200 nanoseconds
[DISTANCE]: 5 [Calculated In]: 9 Iterations and 5400 nanoseconds
[DISTANCE]: 10 [Calculated In]: 17 Iterations and 8200 nanoseconds
[DISTANCE]: 9 [Calculated In]: 23 Iterations and 11700 nanoseconds
[DISTANCE]: 13 [Calculated In]: 35 Iterations and 20800 nanoseconds
[DISTANCE]: 5 [Calculated In]: 7 Iterations and 5400 nanoseconds
[DISTANCE]: 4 [Calculated In]: 4 Iterations and 11400 nanoseconds
[DISTANCE]: 10 [Calculated In]: 25 Iterations and 22700 nanoseconds
[DISTANCE]: 10 [Calculated In]: 13 Iterations and 17600 nanoseconds
[DISTANCE]: 9 [Calculated In]: 18 Iterations and 21000 nanoseconds
[DISTANCE]: 6 [Calculated In]: 19 Iterations and 19100 nanoseconds
[DISTANCE]: 8 [Calculated In]: 43 Iterations and 38700 nanoseconds
[DISTANCE]: 9 [Calculated In]: 28 Iterations and 21300 nanoseconds
[DISTANCE]: 5 [Calculated In]: 10 Iterations and 9600 nanoseconds
[DISTANCE]: 5 [Calculated In]: 13 Iterations and 12500 nanoseconds

NRH-AA:
After more testing I realized the issue with how large the reserve needed to be was because I was casting the to size_t. Now we can reserve far less memory and it works as intended.

NRH-AA:
Alright, the final optimization is including a path check to see if the sightline is clear. If it is we can save a lot of performance ignoring nodes that do not need to be checked. This increases the performance of clear paths anywhere from 2-10x.

NRH-AA:
Added some optimization to the heuristic function. Here are the results:

[DISTANCE]: 3 [Calculated In]: 2 Iterations and 2300 nanoseconds
[DISTANCE]: 3 [Calculated In]: 2 Iterations and 3100 nanoseconds
[DISTANCE]: 3 [Calculated In]: 2 Iterations and 3400 nanoseconds
[DISTANCE]: 4 [Calculated In]: 3 Iterations and 6800 nanoseconds
[DISTANCE]: 4 [Calculated In]: 4 Iterations and 5500 nanoseconds
[DISTANCE]: 4 [Calculated In]: 4 Iterations and 5700 nanoseconds
[DISTANCE]: 4 [Calculated In]: 4 Iterations and 6200 nanoseconds
[DISTANCE]: 4 [Calculated In]: 4 Iterations and 4900 nanoseconds
[DISTANCE]: 5 [Calculated In]: 6 Iterations and 9000 nanoseconds
[DISTANCE]: 5 [Calculated In]: 6 Iterations and 9800 nanoseconds
[DISTANCE]: 6 [Calculated In]: 6 Iterations and 3700 nanoseconds
[DISTANCE]: 5 [Calculated In]: 6 Iterations and 11500 nanoseconds
[DISTANCE]: 6 [Calculated In]: 7 Iterations and 13500 nanoseconds
[DISTANCE]: 7 [Calculated In]: 7 Iterations and 3900 nanoseconds
[DISTANCE]: 7 [Calculated In]: 8 Iterations and 12500 nanoseconds
[DISTANCE]: 8 [Calculated In]: 7 Iterations and 9300 nanoseconds
[DISTANCE]: 9 [Calculated In]: 11 Iterations and 13300 nanoseconds
[DISTANCE]: 10 [Calculated In]: 9 Iterations and 9800 nanoseconds
[DISTANCE]: 12 [Calculated In]: 27 Iterations and 28500 nanoseconds
[DISTANCE]: 13 [Calculated In]: 37 Iterations and 33000 nanoseconds
[DISTANCE]: 14 [Calculated In]: 45 Iterations and 39300 nanoseconds
[DISTANCE]: 15 [Calculated In]: 56 Iterations and 50300 nanoseconds
[DISTANCE]: 15 [Calculated In]: 55 Iterations and 61600 nanoseconds

NRH-AA:
Video of the difference between tfs original algorithm working vs mine and the speed of the new algorithm. Keep in mind the speeds are slower than they would be because of the magic effects showing the paths. In both cases you can see the paths are created properly, but my algorithm doesn't check nodes in a circle. The h value makes it favor nodes closer to the target where the original doesn't favor any. It just makes a circle bigger and bigger until it reaches its target. The performance and memory difference is clear when visualizing it this way. CONST_ME_TELEPORT shows the path it used. CONST_ME_POFF shows all the nodes checked to find it.

[YouTube Video](https://youtu.be/bM9Mq51RHyg)

NRH-AA:
I think this is ready if anyone is looking. I tested it all day. I will create a way for me to test millions of paths sometime tonight so I can test for memory leaks. The only other thing I might try to do is remove x and y from the actual node to save a little more memory. I saw someone say that and it may be possible. Though it won't be super easy. We do use those values and a lot of the code is tied to it. So don't hold this back waiting for it haha.


I ran the following test.
```
local talkaction = TalkAction("/paths")

function talkaction.onSay(player, words, param)
	if player:getAccountType() < ACCOUNT_TYPE_GOD or player:getGroup():getId() < 7 then
		return false
	end
	
	local startPos = player:getPosition()
	
	local before = os.clock()
	
	local successful = 0
	local iterations = 10000
	
	for i = 1, iterations do
		local randomX = startPos.x + math.random(-9, 9)
		local randomY = startPos.y + math.random(-9, 9)
		
		local targetPos = Position(randomX, randomY, startPos.z)
		if (player:getPathTo(targetPos)) then
			successful = successful + 1
		end
	end
	
	local after = os.clock()
	
	print(string.format("Pathfinding took %0.8f seconds to run %d times. %d paths were successful", after - before, iterations, successful))
	return false
end

talkaction:separator(" ")
talkaction:register()
```


which resulted in this output. Memory did not increase indicating a memory leak:
`Pathfinding took 0.790000 seconds to run 100000 times. 99427 paths were successful`

My CPU usage increased about 0.5-1% to calculate the paths and memory remained where it started.

After running 1m paths through this code (all at the same time) the result is this:
`Pathfinding took 7.558000 seconds to run 1000000 times. 1000000 paths were successful`

The memory for the server actually decreased after the test assuming it freed memory on something else. I really doubt there is a memory leak at this point.

NRH-AA:
Results of changing the heuristic a little:

```
Pathfinding took 6.77500000 seconds to run 1000000 times. 1000000 paths were successful
Pathfinding took 6.81500000 seconds to run 1000000 times. 1000000 paths were successful
```


ALSO... I did a clean install of the old TFS algorithm. As if we reverted my last pull request. Here is the result:
```
Pathfinding took 18.52000000 seconds to run 1000000 times. 459925 paths were successful
Pathfinding took 19.84900000 seconds to run 1000000 times. 438003 paths were successful
```

A pretty clear sign that this version is faster, uses less memory, and no more memory leaks! NICE!

The low success rate is due to the old version of maxSearchDist. It was only like 15 or so before we increased it to clientViewX+Y and I am checking 9x9 around the player so a lot of paths end up being too far away. With that being said, the old version didn't handle paths outside its range well and still tried to calculate the path which means we did 250 iterations to find out we couldn't make the path. Between a slower implementation and all of the non existent pre-path conditions that are in my new code it takes almost 4x as long to run the paths.

Last check on the old tfs algorithm I reduce the path sizes to 5x5 to try and get more successful paths. It was still slower than this new algorithm on half the size of paths.
`Pathfinding took 9.88900000 seconds to run 1000000 times. 966979 paths were successful`

Codinablack:
I am just giving an opinion here.. take it or leave it... but if you are going to continue to optimize and fine hone this pathfinding algo, might I suggest you research JPS, as it's supposed to be an enhancement to A* which supposedly dramatically increases performance... at least for sure on big grids/nodemaps... it may or may not be worth implementing for our much smaller use cases, I have no idea... another possible optimization I thought of, but never really checked to see if it could be applied, was to limit the distance the node search continues away from the creature based on max viewport.. Anyways, good luck and hope you keep killing it on these work you are doing!

src/map.cpp
```diff
@@ -730,6 +731,43 @@ bool Map::getPathMatching(const Creature& creature, const Position& targetPos, s
...
+				}
+				if (startY > targetY && pos.y > startY) {
+					continue;
+				} else if (startY == targetY && pos.y != startY) {
+					continue;
+				} else if (startY < targetY && pos.y < startY) {
+					continue;
+				}
+
+				// We don't need nodes past the targetPos
```
Shawak, line 773: Wait this isn't true right? Because this stops us from  going around corners to a position closer to us than the corner

gesior:
> Results of changing the heuristic a little:

It looks like optimized algorithm. I will try to benchmark it later.

Problem with path finding CPU usage is not players walking, but monsters walking.
You should not benchmark player walking to random position. Benchmark every monster spawned walking to random position on screen (maybe better not random, but to each tile on screen to keep tests more consistent), so there will be many positions to which they cannot walk and a lot of map to load from RAM (not CPU cache).
Your benchmark reports 100% success rate, but there are many cases, when there is no way to target on OTS (ex. player and monster are in separate tunnels) and these are the worst - you must calculate path until algorithm hits limit ex. 250 iterations, to be 'sure' that there is no way.
Problem with benchmarking 'valid path' 100% of time is that CPU branch and cache predictions will heavily optimize that branch/cache and it may increase speed few (!) times, but in real world with multiple fails it will work slower.

It would be also good to print total number of 'steps to target' calculated, because it does not matter, if algorithm is faster, if it calculates longer (not optimal) paths more often.

NRH-AA:
> Problem with path finding CPU usage is not players walking, but monsters walking. You should not benchmark player walking to random position. Benchmark every monster spawned walking to random position on screen (maybe better not random, but to each tile on screen to keep tests more consistent), so there will be many positions to which they cannot walk and a lot of map to load from RAM (not CPU cache). 

This is something that would make either algorithm slower. I understand the idea you are presenting, but if an algorithm runs faster even if it is because of CPU cache it would almost always run faster in a more "real world" environment. I will mention the original pathfinding system in TFS wasn't "slow" it just needed optimizations to how it worked. With just a few if statements in the original code it would of made it perform a lot better. No matter what though, it will always be slower because it checks 7 extra nodes for each 1 node in A* because it doesn't utilize the target position.

> Your benchmark reports 100% success rate, but there are many cases, when there is no way to target on OTS (ex. player and monster are in separate tunnels) and these are the worst - you must calculate path until algorithm hits limit ex. 250 iterations, to be 'sure' that there is no way. Problem with benchmarking 'valid path' 100% of time is that CPU branch and cache predictions will heavily optimize that branch/cache and it may increase speed few (!) times, but in real world with multiple fails it will work slower.

In these tests, which I did the exact same test on both algorithms it is clear the new implementation is a lot faster. The high success rate is actually a testament to how good the pathfinder is. Making the two slower by having fails will not change that, but I can definitely do some tests to show that. The reason why is this: In my implementation I can figure out if a path is not possible in 60-80 iterations which is why I have the 120 iteration limit, it is actually more than it needs to be just because I wanted to give it some wiggle room. In the old version 250 was the limit of closed nodes (not iterations) it could check which limited it to somewhere between 15-20 sqms. The 120 iteration limit allows us to check paths, like in caves that do not link, up to around 30 tiles away. With would make it SEEM slower but in reality it is checking many more nodes. It is actually a place where I could optimize the code even more. I purposefully left out this optimization because I need to see tibias behavior in this area. I know their monsters will only check so far for a path but I am not sure exactly how far that is.

So, yes in the specific case of paths not being found (they are too long to find) but the target is close enough to not make my code ignore the path they will look as though they are the same speed. In reality, again, mine checked 2-3x as many nodes in the same time.

To fix this, all I would have to do is add a check to how far the node we are checking is from our startPos. If the nodes are 20sqm away for example we could just terminate because the path is too long.

I know the algorithm takes the appropriate (not longer) path from testing it on monsters in real world scenarios, and just by knowing how A* finds its paths. It will never choose a path that is 14sqm over a path that is 13sqm.

In most cases you will have two paths that can be correct, depending on how you have implemented the algorithm is how it will decide, but it doesn't matter which one it takes as long as it picks one of the two.

Here is an example:
Djikstras
image

A*
image

Both paths are the same distance to cover. The ONLY argument about A* is that it looks more like a computer made the path, which it did. It doesn't look very monster like to walk into a wall, but the performance cost of not using A* is way too much of a draw back.

> It would be also good to print total number of 'steps to target' calculated, because it does not matter, if algorithm is faster, if it calculates longer (not optimal) paths more often.

Check my original pull and the pull made for the memory leak problem. I test distance on paths extensively. The path should always be the same distance as the distance between startPos and targetPos. (maybe - 1) because we don't walk on top of the target. Trust me, I have worked on this for years. I know what is going on here very well.

Lastly, if you want to get a test with more accurate results in all cases add this:
Replace:
```
if (fpp.maxSearchDist != 0 &&
    (startPos.getDistanceX(pos) > fpp.maxSearchDist || startPos.getDistanceY(pos) > fpp.maxSearchDist)) {
	continue;
}
```

With:
```
if (fpp.maxSearchDist != 0 &&
    (startPos.getDistanceX(pos) > fpp.maxSearchDist || startPos.getDistanceY(pos) > fpp.maxSearchDist)) {
	continue;
} else if ((startPos.getDistanceX(pos) + startPos.getDistanceY(pos) >
            Map::maxViewportX + Map::maxViewportY)) {
	break;
}
```

I added the check when paths have maxSearchDist but not for ones that do not. We need to end those sooner too just not sure when.

> I am just giving an opinion here.. take it or leave it... but if you are going to continue to optimize and fine hone this pathfinding algo, might I suggest you research JPS, as it's supposed to be an enhancement to A* which supposedly dramatically increases performance... at least for sure on big grids/nodemaps... it may or may not be worth implementing for our much smaller use cases, I have no idea... another possible optimization I thought of, but never really checked to see if it could be applied, was to limit the distance the node search continues away from the creature based on max viewport.. Anyways, good luck and hope you keep killing it on these work you are doing!

I plan on looking at some of the "better" algorithms for sure. A* is kind of the go to for tibia because we don't calculate z position and the paths are pretty small. Any other algorithm is going to be slower than what you can get from a perfect A* implementation for tibia. (Mine is not perfect. We could use a heap for openSet, and utilize vector indices instead of parent node in AStarNode just as an example to make it a little better even now). However, if we wanted to make a more verbose pathfinder so we could delete a lot of the other game logic like if we can throw an item somewhere that uses z position, going for one of the other algos might be better again maybe not though because the small distance of paths.

I will check out JPS for sure.

You are exactly right about the node thing. I didn't see that before I recommended it to gesior for his tests lol. Nice eye. I purposefully left it out for now because idk tibias actual limit and 120 will only check around 30sqm anyway.

src/map.cpp
```diff
@@ -668,7 +668,7 @@ static uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
 	uint16_t dx = std::abs(p1.getX() - p2.getX());
 	uint16_t dy = std::abs(p1.getY() - p2.getY());
-	return MAP_NORMALWALKCOST * (dx + dy) + (MAP_DIAGONALWALKCOST - 2 * 10) * std::min(dx, dy);
+	return 10 * (dx + dy);
```
ArturKnopik: What this 10 mean, pleasue use some const variable instead magic number

gesior:
@NRH-AA 
I ran test as in previous optimized pathfinding memory leak fix (each monster on TFS map tries to walk to every position in 10 SQM range: add at start of `Game::checkDecay` this https://paste.ots.me/564228/text ) and with this PR `map.cpp` and `map.h` it **CRASHED** server!
Tested on Ubuntu 22.04 without vcpkg (`http` disabled to make it compile). `gdb` report:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000556c1268753a in Map::getPathMatching (this=0x556c1297cfd0 <g_game+80>, creature=..., targetPos=..., dirList=..., pathCondition=..., fpp=...) at /home/jskalski/p/prv/ots/forgottenserver-pf/src/map.cpp:834
834			pos.x = found->x;
[Current thread is 1 (Thread 0x7ff7de03c640 (LWP 68084))]
(gdb) bt full
#0  0x0000556c1268753a in Map::getPathMatching (this=0x556c1297cfd0 <g_game+80>, creature=..., targetPos=..., dirList=std::vector of length 25, capacity 32 = {...}, pathCondition=..., fpp=...)
    at /home/jskalski/p/prv/ots/forgottenserver-pf/src/map.cpp:834
        dx = 18255
        dy = 0
        pos = {x = 20736, y = 0, z = 13 '\r'}
        startPos = {x = 139, y = 485, z = 13 '\r'}
        distanceX = 10
        distanceY = 6
        maxDistanceX = 22
        maxDistanceY = 22
        allNeighbors = {_M_elems = {{first = -1, second = 0}, {first = 0, second = 1}, {first = 1, second = 0}, {first = 0, second = -1}, {first = -1, second = -1}, {first = 1, second = -1}, {first = 1, second = 1}, {first = -1, 
              second = 1}}}
        sightClear = false
        endPos = {x = 128, y = 480, z = 13 '\r'}
        nodes = {nodes = std::vector of length 0, capacity 308, nodeMap = std::unordered_map with 0 elements, visited = std::unordered_set with 0 elements, openSet = std::priority_queue wrapping: std::vector of length 0, capacity 0}
        found = 0x7ff7d81c949000
        bestMatch = 0
        iterations = 120 'x'
        n = 0x7ff7d865ba40
        prevx = 20736
        prevy = 0
#1  0x0000556c1279ec4b in Creature::getPathTo (this=0x7ff7d93f4f10, targetPos=..., dirList=std::vector of length 25, capacity 32 = {...}, fpp=...) at /home/jskalski/p/prv/ots/forgottenserver-pf/src/creature.cpp:1498
No locals.
```

**EDIT:**
I ran all monsters with default Creature parameters passed to fpp (min/max distance set to 1 = go to target position):
```
	FindPathParams fpp;
	fpp.fullPathSearch = true;
	fpp.clearSight = true;
	fpp.maxSearchDist = Map::maxViewportX + Map::maxViewportY;
	fpp.minTargetDist = 1;
	fpp.maxTargetDist = 1;
```
before merging this PR, someone should also test how monsters that keep distance work after changes ex. Warlock.

NRH-AA:
pos = {x = 20736, y = 0, z = 13 '\r'}
dx = 18255
prevx = 20736

startPos = {x = 139, y = 485, z = 13 '\r'}
endPos = {x = 128, y = 480, z = 13 '\r'}

What is this?

It looks like you made a creature that is 20k tiles away from the path it generated.

Shawak:
> It looks like you made a creature that is 20k tiles away from the path it generated.

Well even if so, that should not crash the server imo.

NRH-AA:
> Well even if so, that should not crash the server imo.

No it should not but that is the only thing that sticks out based on what he showed. I need more information because the code shouldn't of let that path execute at all. So I need to know if that is actually what happened based on that output.

NRH-AA:
So I ran the follow code to test monsters outside the players screen, on a bigger map, ect.

```
local talkaction = TalkAction("/paths")

local area = {min = {x = 26058, y = 25921}, max = {x = 26968, y = 26508}}
local creatures = {}

function talkaction.onSay(player, words, param)
	if player:getAccountType() < ACCOUNT_TYPE_GOD or player:getGroup():getId() < 6 then
		return false
	end
	
	local startPos = player:getPosition()
	local successful = 0
	local iterations = 10000
	
	
	if #creatures == 0 then
		for x = area.min.x, area.max.x do
			for y = area.min.y, area.max.y do
				local tile = Tile(Position(x, y, startPos.z))
				if tile then
					local creature = Creature(tile:getTopCreature())
					if creature then
						creatures[#creatures + 1] = creature
					end
				end
			end
		end
	end
	
	local before = os.clock()
	
	for i = 1, iterations do
		for _, monster in pairs(creatures) do
			local monsterPosition = monster:getPosition()
			local randomX = monsterPosition.x + math.random(-7, 7)
			local randomY = monsterPosition.y + math.random(-7, 7)
			if (monster:getPathTo(Position(randomX, randomY, startPos.z))) then
				successful = successful + 1
			end
		end
	end
	
	local after = os.clock()
	
	print(string.format("Pathfinding took %0.8f seconds to run %d times. %d paths were successful", after - before, iterations, successful))
	return false
end

talkaction:separator(" ")
talkaction:register()
```

Here is the results.

`Pathfinding took 1.52200000 seconds to run 10000 times. 230000 paths were successful`

Then I did a 25x25 area test so many paths would be outside the possible range.

```
Pathfinding took 0.42500000 seconds to run 10000 times. 25200 paths were successful
Pathfinding took 0.43700000 seconds to run 10000 times. 24973 paths were successful
Pathfinding took 0.40600000 seconds to run 10000 times. 25172 paths were successful
Pathfinding took 0.41800000 seconds to run 10000 times. 24977 paths were successful
Pathfinding took 0.40700000 seconds to run 10000 times. 25177 paths were successful
Pathfinding took 0.43600000 seconds to run 10000 times. 25127 paths were successful
Pathfinding took 0.40600000 seconds to run 10000 times. 25101 paths were successful
Pathfinding took 0.43800000 seconds to run 10000 times. 25338 paths were successful
```

I executed around 2m paths in total. I couldn't produce a crash.

I even added

```
if (monster:getPathTo(startPos)) then -- startPos is my player position which is at 1000,1000,7
	successful = successful + 1
end
```

Making them look for extremely long paths through black zones and still didn't produce a crash.
It does however seem to increase memory. I am not sure why that is happening when it wasn't on my server so I will look at that when I am back from my vacation.

Edit: it seems the memory increasing had to do with my lua code not the paths themselves.

NRH-AA:
Okay I am on my phone but @gesior  did you test the code before my most recent changes? The previous codes definitely could have caused a crash. The heuristic had to be changed again. Sorry, I know it’s probably annoying to test something that gets changed like that. I am positive that was the issue though.

I will use the test you posted to test when I am back home

gesior:
> Okay I am on my phone but @gesior did you test the code before my most recent changes? The previous codes definitely could have caused a crash. The heuristic had to be changed again. Sorry, I know it’s probably annoying to test something that gets changed like that. I am positive that was the issue though.
> 
> I will use the test you posted to test when I am back home

I tested on Windows and it does not crash. I will test again on Ubuntu tomorrow.

gesior:
I tested it yesterday on cloud with Ubuntu 24.04 and it did not crash.
I tested it today on cloud with Ubuntu 22.04 and my home PC with Ubuntu 22.04, and it crashed in both cases.

I compiled it with `-fsanitize=address` and it crashed with this report:
https://paste.ots.me/564230/text
with official TFS files it does not report anything.

To make sure this is not related to Ubuntu 22.04. I compiled it on Ubuntu 24.04 with `-fsanitize=address` and it crashed again with a similar report:
https://paste.ots.me/564231/text
so there is still some kind of memory management bug.

NRH-AA:
`std::priority_queue<AStarNode*, std::vector<AStarNode*>, NodeCompare>().swap(openSet);`

this is probably the culprit. It can be deleted from AStarNodes::clear()

Replace it with:
```
While (!openSet.empty()) {
     openSet.pop()
}
```

gesior:
> `std::priority_queue<AStarNode*, std::vector<AStarNode*>, NodeCompare>().swap(openSet);`
> 
> this is probably the culprit. It can be deleted from AStarNodes::clear()
> 
> Replace it with:
> 
> ```
> While (!openSet.empty()) {
>      openSet.pop()
> }
> ```

Changes in `AStarNodes::clear()` did not fix anything. Even after I removed all code from that function.

Report says that problem is in:
```
AStarNode* AStarNodes::createNode(AStarNode* parent, uint16_t x, uint16_t y, uint16_t g, uint16_t f)
{
	uint32_t key = hashCoord(x, y);
	nodes.emplace_back(AStarNode{parent, x, y, g, f});
	AStarNode* node = &nodes.back(); 
	nodeMap[key] = node;
	openSet.push(node);
	return node;
}
```
in line `nodes.emplace_back(AStarNode{parent, x, y, g, f});`. It says that this line frees (deletes) AStarNode, which is probably/partially true, because we overwrite `nodes` vector element on heap.

After all changes from this PR, `AStarNodes` follows RAII.
We can remove `clear` function, change `AStarNode` to `AStarNode*`, allocate `AStarNode` on stack and use `AStarNodes` destructor to remove all allocated `AStarNode`s. Other variables (`nodes`, `nodeMap`, `visited` and `openSet`) are allocated with `AStarNodes`  and just points to `nodes` elements, so they will be auto deleted by destructor.
I did these changes in:
https://github.com/NRH-AA/forgottenserver/commit/aff3db1f0d71b5279ae25ca0898426c02a39ca4e
there is no memory leak, no crashes, no reports from address sanitizer and execution time is pretty low:
```
Time: 4207.188155 Calls: 285120 Found: 138833 Not found: 146287 Total length: 1160385
Time: 4323.751988 Calls: 285120 Found: 138833 Not found: 146287 Total length: 1160385
Time: 4431.873102 Calls: 285120 Found: 138833 Not found: 146287 Total length: 1160385
Time: 4439.786814 Calls: 285120 Found: 138833 Not found: 146287 Total length: 1160385
```

**EDIT:**
From programming point of view, everything in this PR looks good.
**I cannot login into game with client on Linux, so someone has to check how monsters behave in game**:
- do they run to player correctly (1 sqm range)
- do they run from player on red hp correctly
- do they keep distance from player correctly (3-4 sqm range: Warlock, Amazon etc.)

After these 3 checks, we can merge this PR.

**EDIT 2:**
There is also some build failure with `AppleClang 15.0.0.15000309` compiler:
https://github.com/otland/forgottenserver/actions/runs/15946617531/job/44981468634?pr=4934
IDK, if it's real Clang problem or GitHub problem. After I pushed my commit, it showed old `map.h` code in `Files changed` for few minutes.

NRH-AA:
I think using new is a mistake. I tested monster behavior before the changes you made and it was proper. The changes you made shouldn’t have affected that at all but from what I gather new shouldn’t be used when working with pathfinders. What is the time being measured there?

Anyway, I can confirm monster behavior is correct. I also added a fix for slow fleeing monsters when the player isn’t moving. I am not home atm, if you want gesior you could modify the Monster::onWalk

and just add a check to see if the monsters lastStep has happened more recently then the pathfindingdelay. Block it unless it hasn’t. That way it doesn’t execute if it doesn’t need it.

gesior:
> from what I gather new shouldn’t be used when working with pathfinders

It should not affect performance much or at all. I can benchmark it tomorrow on Windows, where both versions of code work.
You can try to allocate it on heap, but I tried few solutions and all failed, when I tried to compile with address sanitizer.

> What is the time being measured there?

The same as before: https://github.com/otland/forgottenserver/pull/4924#issuecomment-2946887911

Results:
- **old algorithm**: 7808 ms (100%)
- new algorithm with memory leak: 8289 ms (106%)
- new algorithm without memory leak: 7416 ms (95%)
- **new algorithm from this PR**: 4350 ms (56%)

NRH-AA:
Alright, sounds like it is good enough then. If it stops the crash then let’s roll with it.

gesior:
> It should not affect performance much or at all.

It affects performance. It's 25% slower with 'new', but IDK how to make it work without 'new' and compile with address sanitizer/on Ubuntu 22.04.
Version with 'new' is still almost 2 times faster than old algorithm, so we should merge this PR as crash/memory leak fix and we can work on removing 'new' and other optimizations later.

NRH-AA:
> > It should not affect performance much or at all.
> 
> It affects performance. It's 25% slower with 'new', but IDK how to make it work without 'new' and compile with address sanitizer/on Ubuntu 22.04. Version with 'new' is still almost 2 times faster than old algorithm, so we should merge this PR as crash/memory leak fix and we can work on removing 'new' and other optimizations later.

I think we can add a null check inside NodeCompare to fix the crash before your change to fix it.

If you have time could you try this? In the meantime we can push this.

it seems the issue is we are removing nodes; somehow at the same time they are being compared resulting in a null reference.

```
struct NodeCompare
	{
		bool operator()(AStarNode* a, AStarNode* b) const
		{
                         if (!a) return false;
                         if (!b) return true;
			return a->f > b->f; // Min-heap based on f score
		}
	};

```

NRH-AA:
I had it calling nodes.clear() too early which was probably causing the invalid memory access.

src/map.cpp
```diff
@@ -868,18 +872,21 @@ AStarNodes::AStarNodes(uint16_t x, uint16_t y) : nodes(), nodeMap()
...
 	}
 }
 
 AStarNode* AStarNodes::createNode(AStarNode* parent, uint16_t x, uint16_t y, uint16_t g, uint16_t f)
 {
 	uint32_t key = hashCoord(x, y);
-	nodes.emplace_back(new AStarNode(parent, x, y, g, f));
-	AStarNode* node = nodes.back();
+	nodes.emplace_back(AStarNode{parent, x, y, g, f});
+	AStarNode* node = &nodes.back();
```
Shawak, line 883: When `nodes` is reallocated the pointer becomes invalid, thats probably also a problem here (e.g. exceeding the capacity)

gesior: Yes. This is a real problem. @NRH-AA can you stop working on branch right now? I would like to push my fix.

NRH-AA:
@gesior We still need to increase the allocated size * 7 wasn't enough. I picked that number "randomly". I figured it was close to enough to handle any paths we were trying to get. That is actually the only problem that was happening.

gesior:
@Shawak 
Yes. It's problem with reallocating vector after reaching reserved size.

_What I did to find out real problem is described at end of this comment._

I tested, if increasing `nodeReserveSize` size by 10 times fixes problem. It worked, but it's not a real solution.
Real solution would be to stop adding new `nodes` after reaching `nodes.size()` and that's what I did.
Only problem is that we set limit of nodes to constant value. Most complicated paths can't be calculated, because algorithm reaches limit of nodes and cannot increase it.
In tests code with `new` found `138.833` paths, code with heap allocation and limit of nodes found `138.238` paths (0.5% less), but code with 'new' is 24% slower, so I don't think that extra 0.5% found paths is worth +25% CPU usage.

> We still need to increase the allocated size * 7 wasn't enough

`7` is pretty good number.
Number of paths found and time to calculate all paths in my TFS map benchmark with distance up to 12 SQM x/y (`7` as base for +/- %):
- 5: `164872` paths (-3.5%), `4895` ms (-16% CPU)
- 6: `168892` paths (-1.2%), `5308` ms (-9% CPU)
- 7: `170923` paths, `5833` ms
- 12: `173231` paths (+1.3%), `7556` ms (+29% CPU)
- 50: `173687` paths (+1.6%), `8947` ms (+53% CPU)
- 500: `173687` paths (+1.6%), `10372` ms (+78% CPU)

Before this PR (optimized with memory leak):
- `150.617` paths (-11.9%), `13058` ms (+123% CPU)

Before optimized path finding:
- `169935` paths (-0.5%), `13322` ms (+228% CPU)

So with `* 7` and this PR code, we get 0.5% more paths found than before 'optimized path finding' and it uses 56 % less CPU.
**Nice optimization** @NRH-AA !

---

What I did to find out real problem:

I wrote simplified version of this problem:
https://paste.ots.me/564233/text
Most important part is:
```cpp
int main(int argc, const char** argv)
{
	AStarNodes nodes(1, 1);
	AStarNode* node1 = &nodes.nodes.back();
	std::cout << "n1:" << node1->x << "," << node1->y << std::endl;
	AStarNode* node2 = nodes.createNode(node1, 1, 2);
	std::cout << "n1:" << node1->x << "," << node1->y << std::endl;
	return 0;
}
```
You can run it with `g++ -fsanitize=address test.cpp && ./a.out` or here:
https://cpp.sh/
Result is (node1 values are modified by emplacing second element to `nodes`, so it really frees that address):
```
n1:1,1
n1:511,0
```
but if you uncomment:
```
//	nodes.reserve(2);
```
it will return valid result:
```
n1:1,1
n1:511,0
```
but if you add 3rd element in `main`, it will return invalid results again. Problem appears only when you add more elements to `nodes` than is reserved.

Code that returns invalid results is always detected with `-fsanitize=address`.

NRH-AA:
> @Shawak Yes. It's problem with reallocating vector after reaching reserved size.
> 
> _What I did to find out real problem is described at end of this comment._
> 
> I tested, if increasing `nodeReserveSize` size by 10 times fixes problem. It worked, but it's not a real solution. Real solution would be to stop adding new `nodes` after reaching `nodes.size()` and that's what I did. Only problem is that we set limit of nodes to constant value. Most complicated paths can't be calculated, because algorithm reaches limit of nodes and cannot increase it. In tests code with `new` found `138.833` paths, code with heap allocation and limit of nodes found `138.238` paths (0.5% less), but code with 'new' is 24% slower, so I don't think that extra 0.5% found paths is worth +25% CPU usage.
> 
> > We still need to increase the allocated size * 7 wasn't enough
> 
> `7` is pretty good number. Number of paths found and time to calculate all paths in my TFS map benchmark with distance up to 12 SQM x/y (`7` as base for +/- %):
> 
> * * 5: `164872` paths (-3.5%), `4895` ms (-16% CPU)
> * * 6: `168892` paths (-1.2%), `5308` ms (-9% CPU)
> * * 7: `170923` paths, `5833` ms
> * * 12: `173231` paths (+1.3%), `7556` ms (+29% CPU)
> * * 50: `173687` paths (+1.6%), `8947` ms (+53% CPU)
> * * 500: `173687` paths (+1.6%), `10372` ms (+78% CPU)
> 
> Before this PR (optimized with memory leak):
> 
> * `150.617` paths (-11.9%), `13058` ms (+123% CPU)
> 
> Before optimized path finding:
> 
> * `169935` paths (-0.5%), `13322` ms (+228% CPU)
> 
> So with `* 7` and this PR code, we get 0.5% more paths found than before 'optimized path finding' and it uses 56 % less CPU. Nice optimization @NRH-AA !
> 
> What I did to find out real problem:
> 
> I wrote simplified version of this problem: https://paste.ots.me/564233/text Most important part is:
> 
> ```c++
> int main(int argc, const char** argv)
> {
> 	AStarNodes nodes(1, 1);
> 	AStarNode* node1 = &nodes.nodes.back();
> 	std::cout << "n1:" << node1->x << "," << node1->y << std::endl;
> 	AStarNode* node2 = nodes.createNode(node1, 1, 2);
> 	std::cout << "n1:" << node1->x << "," << node1->y << std::endl;
> 	return 0;
> }
> ```
> 
> You can run it with `g++ -fsanitize=address test.cpp && ./a.out` or here: https://cpp.sh/ Result is (node1 values are modified by emplacing second element to `nodes`, so it really frees that address):
> 
> ```
> n1:1,1
> n1:511,0
> ```
> 
> but if you uncomment:
> 
> ```
> //	nodes.reserve(2);
> ```
> 
> it will return valid result:
> 
> ```
> n1:1,1
> n1:511,0
> ```
> 
> but if you add 3rd element in `main`, it will return invalid results again. Problem appears only when you add more elements to `nodes` than is reserved.
> 
> Code that returns invalid results is always detected with `-fsanitize=address`.

There is 2 things that need to be fine tuned.

image

```
Nodes: 227 Iterations: 161
Nodes: 191 Iterations: 141
Nodes: 183 Iterations: 135
```

1) The allocated size has to be correct for how many iterations are allowed in the algorithm.
```
if (iterations >= Map::nodeReserveSize) {
	return false;
}
```

I had this which is part of why it broke. I wanted to make it easy to increase what size paths can be drawn by just changing:
```
static constexpr int32_t maxViewportX = 11; // min value: maxClientViewportX + 1
static constexpr int32_t maxViewportY = 11; // min value: maxClientViewportY + 1
```

The issue is the amount of iterations can't be the same size as the reserveSize otherwise it will always overflow. So it should be changed to:

```
if (iterations >= (Map::maxViewportX + Map::maxViewportY) * 7) { // 7 so it can check all directions for a path
	return false;
}
```

2) The allocated size has to handle that many nodes. The worst case scenario is iterations * 7 (if we have to check every node on the worst possible path) but that should end way before based on the distance check on the node to the end position that we added. So to be safe we should do:
`(Map::maxViewportX + Map::maxViewportY) * 10)`

Yes, your solution will definitely stop any crashes which is perfect, but this is where I left off before I took my vacation. I didn't think about the fact the allowing the iterations to be constant with the reserve size would cause this problem. It still needs to be addressed as it is unstable. The iterations should be used as a way to stop paths that are too large.

gesior:
@NRH-AA 
I've done my fix. Now it does not crash and works super fast.
You can try to fine tune it, but results with `7` are already better than before path finding optimizations (+0.5% more paths found).
Maybe make 2 variables in `map.h`. One to limit iterations, second to limit nodes reserve.
Also remember to change `uint8_t iterations = 0;` to `uint32_t`, if you want to test it with higher number of iterations.

IDK what is wrong on that screenshot. Algorithm is not supposed to find a way out from any size of labyrinth. It must limit CPU usage by single monster. Also, walking too far, could make monster walk out of player screen (no players on screen) and make it go idle.

Old algorithm (TFS 0.x) had number of iterations - in our case also nodes reserved - set by variable passed to `getPathMatching`, so it was able to run different algorithm complexity in different parts of engine. Ex.
- when player use distance item(server must walk player to that item pos): run 500 iterations (we want good results)
- when monster run to player: run 100 iterations (save CPU)
  - if all 'run to player' of given monster failed in last second, run 500 iterations (monster is in some labyrinth, find way to player using more CPU)

That way normal path finding of monster uses less CPU, but if it fails to find path, it can run more complicated path finding.

gesior:
@NRH-AA 
With commit `Adjust iterations and reserve size` code is 8% faster than with previous commit `* 7`. It failed in 2 extra path calculations out of 404k paths (paths found before commit `170923`, after `170921`), but it still returns paths in +0.5% cases (old algorithm `169935`).

Right now **optimized algorithm uses 60%** (!) less CPU than old algorithm (5334 ms vs 13322 ms).
There is no memory leak, no crashes and no problem with compilation.

Can we merge this PR? @ranisalt  @ArturKnopik

ArturKnopik:
Tomorrow I will review this changes and test it, if everything is ok I will merge

src/monster.cpp
```diff
@@ -1085,7 +1085,17 @@ bool Monster::walkToSpawn()
 	return true;
 }
 
-void Monster::onWalk() { Creature::onWalk(); }
+void Monster::onWalk()
+{
+	Creature::onWalk();
+
+	if ((attackedCreature || followCreature) && isFleeing()) {
+		if (lastPathUpdate - OTSYS_TIME() > 0) {
```
ranisalt:
```suggestion
		if (lastPathUpdate > OTSYS_TIME()) {
```

ArturKnopik:
Tested for a while, looks like it works, code also looks ok, in case of any problem we can revert this changes

ArturKnopik (approved):

Codinablack:
@NRH-AA For the record JPS is not a different algo than A*, it's an enhancement for A*, one that keeps it from iterating over already known bad paths/nodes... which, if I read this PR correctly (the comments not the code), you already figured it out that such an alteration would indeed be a good optimization for this algo.

NRH-AA:
> @NRH-AA For the record JPS is not a different algo than A*, it's an enhancement for A*, one that keeps it from iterating over already known bad paths/nodes... which, if I read this PR correctly (the comments not the code), you already figured it out that such an alteration would indeed be a good optimization for this algo.

I made sure to ignore nodes we don't need to check again but JPS would work a little differently. I think JPS works best if you already have the path data. (like google maps) We know all routes that can be taken which can be used to ignore routes we know will end up in a bad place. We can also jump to certain routes if we know that a node at this x,y that is found for a path going north or whatever will always come out at a specific spot. So maybe 100 paths all use that same route. That would be a hot spot and we could automatically grab the nodes required and just jump to that point and start the path finding again. Doing this multiple times would for sure increase performance in that case, but outside knowing the paths before hand I don't think its plausible.

gesior:
> I made sure to ignore nodes we don't need to check again but JPS would work a little differently

I did not analyse JPS, but wikipedia says `In computer science, jump point search (JPS) is an optimization to the A* search algorithm for uniform-cost grids`. It looks like it tries to 'walk' X tiles on straight lines to reduce CPU usage, but it won't work on TFS.
It means that JPS works only for `uniform-cost grids`, but TFS has different costs on each step, as there are:
- tile ground speed (ground item)
- direction (diagonal is more expensive)
- magic fields (fire/energy/poison fields etc.) for monsters that are not attacked
- remove creature (ex. Demon destroys his own summon of Fire Elemental)

so JPS won't work.

PR #4935, opened 2025-06-23
Use Boost functions for case-insensitive string comparison

ranisalt:
### Pull Request Prelude

### Changes Proposed

To reduce the amount of code we own and maintain, using Boost functions to perform insensitive string comparison and prune the functions from tools

**Issues addressed:** 

**How to test:** 


Tofame:
Boost is inefficient.

473089891-31bce831-e54d-4066-b32c-c5187a0339ac

ranisalt:
(copying from the linked issue)

It is slower, but is it relevant? These function calls are not in any critical path. Do you measure any slowness caused by it in frequently executing code paths?

Furthermore, `boost::algorithm::to_lower_copy` is not equivalent to `std::transform` since it also allocates space for a copy (but it's not relevant here)

> Btw, why are we leaning into boost so much? I think, where possible, we should stick to std. 

No need for external boost libs. The only thing where boost is truly crucial is boost::asio, but I dont see a point in over-using it anywhere else. These functions do not exist in the STL.

We use Boost because it is rock solid and we don't need to write the functions that are not crucial. It reduces the surface for bugs and also prevents people from trying to prematurely optimize stuff without benchmarking (though you did benchmark this one).

For me, it's all about picking what to spend time on and what not to.

Tofame:
Nope, I just did a simple benchmark, not taking into consideration how it would do in real case scenario, though with that difference, it still imo says enough about boost just being impractical.

The best code is already currently in TFS, there is no point in changing it for worse. It does not require any upkeep as it's simple enough short function,

It is used 81 times
obraz
... and although most of it's uses is for loading XML, it generally is used a lot for ``get...ByName`` methods, which I think should be as fast as possible. If you are getting e.g. npc by name, and you gotta iterate all npcs there is, which might be thousands... it should make sense.

ranisalt:
> If you are getting e.g. npc by name, and you gotta iterate all npcs there is, which might be thousands... it should make sense.

Thousands will be unnoticeable. You will feel these extra nanoseconds when you have a few million items to iterate. 10 thousand iterations taking an extra 100ns (dividing the time you showed by the number of iterations) will still be less than a millisecond - loading the data from RAM will take _a lot more_ time here.

This is not a hill I want to die on though. Whatever floats the boat of everyone else :)

PR #4936, opened 2025-06-23
Replace fmt with std::format/print

ranisalt:
### Pull Request Prelude

### Changes Proposed

Updates C++ standard to C++23. Replaces a few usages of `fmt` with now built-in functions, since fmt is causing build issues when updating vcpkg baseline

**Issues addressed:** 

**How to test:** 

### DIFF
--- a/src/http/http.cpp
+++ b/src/http/http.cpp
@@ -27,7 +27,7 @@ void tfs::http::start(bool bindOnlyOtsIP, std::string_view otsIP, unsigned short
 	if (bindOnlyOtsIP) {
 		address = asio::ip::make_address(otsIP);
 	}
-	std::println(">> Starting HTTP server on {:s}:{:d} with {:d} threads.", address, port, threads);
+	std::println(">> Starting HTTP server on {:s}:{:d} with {:d} threads.", address.to_string(), port, threads);
 
 	auto listener = make_listener(ioc, {address, port});
 	listener->run();


ArturKnopik:
What about visual studio solution and cpp standard?

ranisalt:
> What about visual studio solution and cpp standard?

No idea how to do it but here's my best effort

ArturKnopik:
image
Iooks bad for VS, I would prefer to avoid using C++ 23 **preview** standard

Codinablack:
Yeah exactly what Arthur said.. I have been waiting to do the update to C++23 for once MSVC supports it.... I thought I would have the first public repo running on C++23 and you are over here making the switch before windows even has the language standard completed XD

I think that C++23 is going to be a complete game changer to be completely honest! Glad you are working towards that goal,

I can't wait to get my hands on that change myself, going to completely rework the includes with the std being a module now. 

Side question... did you not run into any problems with protobufs not being C++23 compatible yet??

ranisalt:
Dammit Microsoft DO SOMETHING

> Side question... did you not run into any problems with protobufs not being C++23 compatible yet??

No, could be because I don't use anything with protobuf 😅

vc17/theforgottenserver.vcxproj
```diff
@@ -101,7 +101,7 @@
       <EnableParallelCodeGeneration>true</EnableParallelCodeGeneration>
       <EnableEnhancedInstructionSet>AdvancedVectorExtensions</EnableEnhancedInstructionSet>
       <AdditionalIncludeDirectories>$(VcpkgRoot)include\luajit;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <LanguageStandard>stdcpp20</LanguageStandard>
+      <LanguageStandard>stdcpp23</LanguageStandard>
```
Codinablack, line 104: for the record it's "std:c++23preview" for visual studio right now, or technically, you can use "std:latest", but it would include some features already implemented for C++26.... we are all still looking super hard at Microsoft about them dragging their feet on this after having been the first one to have it ready for the two previous language versions (C++17/20)

MillhioreBT:
> Dammit Microsoft DO SOMETHING
> 
> > Side question... did you not run into any problems with protobufs not being C++23 compatible yet??
> 
> No, could be because I don't use anything with protobuf 😅

It’s been a while since I last tried using C++23 with protobuf, but back then it simply wasn’t supported according to the protobuf repository’s own documentation, though I’m not sure about the current state.

Although since TFS doesn’t use this, I agree with upgrading to C++23 as soon as C++23 is standardized in Visual Studio.

MillhioreBT:
@ranisalt I think it's a good time to **fix the PR and move TFS to C++23**, even on Linux with **GCC++14 most of the important features already work**.

ranisalt:
Did VS move C++23 to non-preview?

MillhioreBT:
> Did VS move C++23 to non-preview?

Oh no, sorry, I made a mistake, it's just that I use **VS2022 Insiders** and it is **much more advanced than the release version**.
But you are right, until it is released, it is better not to move it yet.

ranisalt:
After #5024 I rebased this PR to only contain the migration away from fmt to std since bumping the C++ version is not needed.

ranisalt:
#5032 is needed for this to work successfully

PR #4940, opened 2025-06-25
fix: load SQL config from config.lua (#4518)

gesior:
### Changes Proposed

After changes, it loads `config.lua` config first, then loads env variables and if they are not empty, it overwrites `config.lua` values.

Right now TFS loads env first, so to do not overwrite it, we had to remove SQL config from `config.lua` in #4518 
There is no reason to remove SQL config from `config.lua`.
Env variables should be loaded last and overwrite any previous config (like it does in any other PHP/JS/Python/Java framework).

**Issues addressed:** #4518 #4931

**How to test:**
Scenario 1 - load config from `config.lua` and overwrite it with wrong env config value:
- set valid SQL config in `config.lua`
- before server start type `export SQL_PORT=12345` in terminal
- start TFS from terminal
- it will show error, that it cannot connect to database (because of wrong MySQL port)
- type `export SQL_PORT=3306` or `export SQL_PORT=` (empty values are skipped)
- start TFS from terminal
- it will start without error

Scenario 2 - load config from env, overwrite all `config.lua` variables:
- leave default SQL config in `config.lua`
- set valid config with env variables (export https://github.com/otland/forgottenserver/blob/e51b5b72b216501aafbb2e17245d6a8234a1a0a4/.env.example values), type in terminal:
```
export MYSQL_HOST=127.0.0.1
export MYSQL_USER=youruser
export MYSQL_PASSWORD=yourpassword
export MYSQL_DATABASE=dbname
export MYSQL_SOCK=
export MYSQL_PORT=3306
```
- run TFS from terminal
- it will start without error using env variables

@JoaozinhoBrasil

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -75,6 +75,14 @@ premiumToCreateMarketOffer = true
 checkExpiredMarketOffersEachMinutes = 60
 maxMarketOffersAtATimePerPlayer = 100
 
+-- MySQL
+mysqlHost = "127.0.0.1"
+mysqlUser = "forgottenserver"
+mysqlPass = ""
+mysqlDatabase = "forgottenserver"
+mysqlPort = 3306
+mysqlSock = ""
+
 -- Misc.
 -- NOTE: classicAttackSpeed set to true makes players constantly attack at regular
 -- intervals regardless of other actions such as item (potion) use. This setting
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -34,11 +34,22 @@ ExperienceStages expStages;
 
 bool loaded = false;
 
-template <typename T>
-auto getEnv(const char* envVar, T&& defaultValue)
+std::string getEnv(const char* envVar, std::string defaultValue)
 {
 	if (auto value = std::getenv(envVar)) {
-		return pugi::cast<std::decay_t<T>>(value);
+		if (strlen(value)) {
+			return {value};
+		}
+	}
+	return defaultValue;
+}
+
+uint16_t getEnv(const char* envVar, uint16_t defaultValue)
+{
+	if (auto value = std::getenv(envVar)) {
+		if (strlen(value)) {
+			return pugi::cast<uint16_t>(value);
+		}
 	}
 	return defaultValue;
 }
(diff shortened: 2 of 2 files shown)


config.lua.dist
```diff
@@ -75,6 +75,14 @@ premiumToCreateMarketOffer = true
 checkExpiredMarketOffersEachMinutes = 60
 maxMarketOffersAtATimePerPlayer = 100
 
+-- MySQL
+mysqlHost = "127.0.0.1"
+mysqlUser = "forgottenserver"
+mysqlPass = ""
+mysqlDatabase = "forgottenserver"
+mysqlPort = 3306
+mysqlSock = ""
```
ranisalt, line 84: How about something like this?

```suggestion
mysqlHost = os.getenv("MYSQL_HOST")
mysqlUser = os.getenv("MYSQL_USER")
mysqlPass = os.getenv("MYSQL_PASS")
mysqlDatabase = os.getenv("MYSQL_DATABASE")
mysqlPort = tonumber(os.getenv("MYSQL_PORT"))
mysqlSock = os.getenv("MYSQL_SOCK")
```

Then we can skip handling the environment in C++, which is inherently ugly

HopeItBuilds: this breaks compatibility with old AAC systems that still read from config.lua. Most are unmaintained and won’t support both old and new methods

ranisalt: Are people using both unmaintained AAC AND bleeding edge TFS at the same time?

With that said, this is just the default config, nothing prevents the user to just hardcode values as always

HopeItBuilds: good point lol although I would not doubt ot users

NRH-AA: > Are people using both unmaintained AAC AND bleeding edge TFS at the same time?
> 
> With that said, this is just the default config, nothing prevents the user to just hardcode values as always

Lmfao

Zbizu:
wouldn't a solution like "uncomment these lines in config.lua to use instead of env variables" be better?
it would be more intuitive for less advanced users (and also would allow running multiple binaries on different databases)

example (LLM-generated):
```cpp
string[MYSQL_HOST] = getGlobalStringOrEnv(L, "mysqlHost", "MYSQL_HOST", "127.0.0.1");
````
```cpp
std::string getGlobalStringOrEnv(lua_State* L, const char* luaIdentifier, const char* envVariable, const char* defaultValue = "")
{
	// Step 1: Check if Lua global exists and is a string
	lua_getglobal(L, luaIdentifier);
	if (lua_isstring(L, -1)) {
		size_t len = 0;
		const char* luaStr = lua_tolstring(L, -1, &len);
		std::string result(luaStr, len);
		lua_pop(L, 1);
		return result;
	}
	lua_pop(L, 1); // Clean up Lua stack

	// Step 2: Try environment variable
	const char* envValue = std::getenv(envVariable);
	if (envValue) {
		return std::string(envValue);
	}

	// Step 3: Use default
	return std::string(defaultValue);
}
```

config.lua.dist
gesior: > Then we can skip handling the environment in C++, which is inherently ugly

That's a good idea. Remove all 'env' code from C++ and Lua, and let advanced users use `mysqlHost = os.getenv("MYSQL_HOST")` - if they want, not force it on everyone.

> With that said, this is just the default config

I think that default config should be something that is simple and works. Why would we require someone to understand what are env variables to run OTS? I'm pretty sure that 99% of users that only use Windows never heard about env variables.
I did not find any OTS tutorial for beginners that has step:
```
and now to run TFS, open PowerShell and run:
$env:MYSQL_USER='username'
$env:MYSQL_PASSWORD='password'
$env:MYSQL_DATABASE='databasename'
.\theforgottenserver-x64.exe
```

gesior:
> wouldn't a solution like "uncomment these lines in config.lua to use instead of env variables" be better?

Most intuitive solution would be to remove whole env variables code and keep TFS config like in every other OTS made in last 20 years.
There is also other way to configure some variables inside Docker with env variables. You can pass IP and ports as command line arguments to TFS:
https://github.com/otland/forgottenserver/blob/e51b5b72b216501aafbb2e17245d6a8234a1a0a4/src/main.cpp#L26-L33

ranisalt:
> Most intuitive solution would be to remove whole env variables code and keep TFS config like in every other OTS made in last 20 years.

"Do it this way because we've been doing it this way" is the bane of our existence. A lot of things didn't exist 20 years ago. Not sure you were here to be against migrating from XML files to SQL servers because we were doing XML for X years already.

People don't containers but they absolutely should, spinning up a server, database and website with a few lines of configuration is just way better than the hour-long tutorials they have to follow today. But you're absolutely right, we still want it to be possible, I like the idea of passing program arguments instead

I know you will be directly affected because you write the major account manager out there and is consuming the Lua configuration. Instead, let's improve the HTTP server and add a few routes to return the same information, which should be even more familiar and easy to use, what do you think? It's not like you can even validly use the config.lua file, experience stages for example would require you to _execute_ Lua.

I'll open a discussion so we can gather more info about it and start working on a real information protocol

gesior:
Why do we have to wait a week to merge PR that fixes 8 months old bug and is backward compatible?!

> People don't containers but they absolutely should, spinning up a server, database and website with a few lines of configuration is just way better than the hour-long tutorials they have to follow today

It's a good idea. You should create docker-compose that can be run on Windows: compiles and starts TFS with database inside container and allows user to connect from Windows Tibia client.

> I know you will be directly affected because you write the major account manager out there and is consuming the Lua configuration.

IDC. My acc. maker had last commit 3 years ago. It's not compatible with newest TFS and new versions of PHP. Everyone should use MyAAC.

gesior:
@ranisalt @ArturKnopik 
Can we merge this PR?
I cannot work on Windows in VS and in Docker in same time, because `config.lua` SQL config overwrites env variables inside Docker.

ArturKnopik:
Code style is already fixed on master.

PR #4942, opened 2025-06-26
Avoid searching player inventory twice when accepting market offer

Zbizu:
getMoney loops through all containers held by the player, you really don't want do to this twice on a busy server

if someone is interested in reducing execution time of this even further (it won't be part of this PR, just giving you an idea):
- getMoney could stop searching the moment it finds predefined amount
- getMoney could stop searching earlier, or be skipped entirely if you subtract player's bank balance from the total amount of money to find in player's inventory


ranisalt:
If you rebase it, the format check will pass and it can be merged

ranisalt (approved):

Issue #4943, opened 2025-06-28
[Bug]: Compilation on Ubuntu 22.04 does not work


gesior:
### By submitting this bug issue, you agree to the following.

- [x] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository
- [x] This bug has not been resolved in master branch
- [x] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below), Ubuntu 22.04

### Bug description

tfs::http makes it impossible to compile on Ubuntu 22.04

### Steps to reproduce

1. Try to compile TFS with `cmake --preset default -DHTTP=OFF -DUSE_LUAJIT=ON && cmake --build --config Release --preset default`
2. It fails with errors:
```
/usr/bin/ld: src/Release/libtfslib.a(unity_5_cxx.cxx.o): in function `(anonymous namespace)::mainLoader(ServiceManager*)':
unity_5_cxx.cxx:(.text+0x212b1): undefined reference to `tfs::http::start(std::basic_string_view<char, std::char_traits<char> >, unsigned short, int)'
/usr/bin/ld: src/Release/libtfslib.a(unity_2_cxx.cxx.o): in function `Game::setGameState(GameState_t)':
unity_2_cxx.cxx:(.text+0x16c7a): undefined reference to `tfs::http::stop()'
```

### Actual Behavior

It fails to compile.
It's impossible to compile without `-DHTTP=OFF`, because default boost version on Ubuntu 22.04 is 1.74.
TFS requires 1.75 with `HTTP` and 1.71 without it. With disabled `HTTP`, there is still code that tries to start/stop `tfs::http`, which makes it fail to compile.

### Expected Behavior

It should compile or we should drop Ubuntu 22.04 support.

There are 2 possible fixes:
1. Add `add_definitions(-DHTTP_ENABLED)` in `CMakeLists.txt` and `#ifdef HTTP_ENABLED` in 2 places around `tfs::http` code to make it compile on Ubuntu 22.04
2. Make boost 1.75 default version for TFS. Drop Ubuntu 22.04 support - it will still compile with vcpkg on Ubuntu 22.04, but not with system libraries.

### Backtrace

```cpp

```

gesior:
Fixed by https://github.com/otland/forgottenserver/pull/4876

PR #4876, opened 2024-12-18
Fix build without HTTP module

ArturKnopik:
### Pull Request Prelude

Fix build without HTTP module

- [X] I have followed [proper The Forgotten Server code styling][code].
- [X] I have read and understood the [contribution guidelines][cont] before making this PR.
- [X] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
  
Adding missing #ifdef

**Issues addressed:** 

Cannot build project with HTTP flag disabled

**How to test:** 

2) Build with flag HTTP=OFF or add in src/CMakeLists.txt just before `if (HTTP)` add `SET(HTTP OFF)`

PR #4944, opened 2025-06-28
fix: Ubuntu 22.04 compilation (#4943)

gesior:
### Changes Proposed

Fix Ubuntu 22.04 compilation with system libraries.

offtop: It's impossible to compile TFS without HTTP server. You can disable it by setting port to `0` in `config.lua`, but it's impossible to compile without HTTP.

**Issues addressed:**
https://github.com/otland/forgottenserver/issues/4943

**How to test:**
1. Install required libraries
2. Run `cmake --preset default -DHTTP=OFF -DUSE_LUAJIT=ON && cmake --build --config Release --preset default`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ option(HTTP "Enable HTTP support" ON)
 if (HTTP)
     list(APPEND VCPKG_MANIFEST_FEATURES "http")
     set(BOOST_REQUIRED_VERSION 1.75.0)
+    add_definitions(-DHTTP_ENABLED)
 endif()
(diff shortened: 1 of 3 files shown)


ArturKnopik:
https://github.com/otland/forgottenserver/pull/4876 almost the same fix

gesior:
Duplicated PR.
https://github.com/otland/forgottenserver/pull/4876 is better.
I looked for 'ubuntu' in open PRs and found nothing. I did not think to search for 'HTTP' PR as for me on/off HTTP wasn't important.

PR #4948, opened 2025-07-02
Fix slow reaction speed on first pathmatch

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Allow paths to be drawn as soon as a target/follower is set or removed.

**Issues addressed:** 
N/A

**How to test:** 
Check how long it takes for a monster to start moving when:
1) It is first spawned
2) When you go down stairs and the monster first targets you
3) When you set a target while you have summons
4) When you remove a target while you have summons

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1089,11 +1089,8 @@ void Monster::onWalk()
 {
 	Creature::onWalk();
 
-	if ((attackedCreature || followCreature) && isFleeing()) {
-		if (lastPathUpdate > OTSYS_TIME()) {
-			g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
-			lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
-		}
+	if (isFleeing() && lastPathUpdate > OTSYS_TIME()) {
+		forceUpdatePath();
 	}
 }
 
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -740,6 +740,7 @@ void Creature::setAttackedCreature(Creature* creature)
 	creature->addFollower(this);
 	onAttackedCreature(attackedCreature);
 	attackedCreature->onAttacked();
+	forceUpdatePath();
 
 	for (Creature* summon : summons) {
 		summon->setAttackedCreature(creature);
@@ -788,6 +789,7 @@ void Creature::setFollowCreature(Creature* creature)
 	creature->addFollower(this);
 	hasFollowPath = false;
 	onFollowCreature(creature);
+	forceUpdatePath();
 }
 
 void Creature::removeFollowCreature()


maattch:
There's another issue with summons that's probably not related to this, but when you get closer to the summon it should go away from you keeping 2 sqm distance, which is not happening right now.

NRH-AA:
> There's another issue with summons that's probably not related to this, but when you get closer to the summon it should go away from you keeping 2 sqm distance, which is not happening right now.

It is related to pathfinding but I will open a new PR for that one too. It is a simple fix that I forgot about x.x.

omarosso97:
I have an issue when master has already selected target and player summons creature, summon is lagged for about 2 seconds before it follows the target and start attack. Sometimes it happens when summon is already summoned as well.

NRH-AA:
> I have an issue when master has already selected target and player summons creature, summon is lagged for about 2 seconds before it follows the target and start attack. Sometimes it happens when summon is already summoned as well.

I can't recreate the problem. Did you test with these changes? When I test the monster walks within a second which is normal.

omarosso97:
> > I have an issue when master has already selected target and player summons creature, summon is lagged for about 2 seconds before it follows the target and start attack. Sometimes it happens when summon is already summoned as well.
> 
> I can't recreate the problem. Did you test with these changes? When I test the monster walks within a second which is normal.

Yup, after applying a commit it was still the same though. I double checked everything. I will record it and send a video.

NRH-AA:
> Yup, after applying a commit it was still the same though. I double checked everything. I will record it and send a video.

Try with this commit.

omarosso97:
> > Yup, after applying a commit it was still the same though. I double checked everything. I will record it and send a video.
> 
> Try with this commit.

So I've tested it, basically nothing changed after this commit but I've found the reason of lagged summons and I don't think it's connected to pathfinding at all. In Monster::selectTarget, targetList check seems to be the culprit of summons lag at least in my case. The question is what to do with it? Is it safe to make a check to ommit summons or is there any other solution for this problem?

NRH-AA:
How did you determine it is that part of the code? This is the code from my server (it is upgraded from 1.4.1)

```
bool Monster::selectTarget(Creature* creature)
{
	if (!isTarget(creature)) {
	         return false;
        }
        
        auto it = std::find(targetList.begin(), targetList.end(), creature);
        if (it == targetList.end()) {
	        // Target not found in our target list.
	        return false;
        }
        
        if (isHostile()) {
	        if (!isSummon() && setAttackedCreature(creature)) {
		        g_dispatcher.addTask(createTask([id = getID()]() { g_game.checkCreatureAttack(id); }));
	        }
        }
        
        return setFollowCreature(creature);
}
```

You have to fix the returns because it looks like in latest setFollowCreature has been made void instead of bool.

omarosso97:
Well in onThink (getMaster() && getMaster()->getAttackedCreature()) this part is responsible for a case where players have already selected target and are summoning creature while in combat and it calls directly selectTarget. I checked this function and after adding !isSummon() summons are much more responsive and faster. It works in my case but I don't know whether the code can stay like this.

	auto it = std::find(targetList.begin(), targetList.end(), creature);
    if (!isSummon()) {
        auto it = std::find(targetList.begin(), targetList.end(), creature);
        if (it == targetList.end()) {
            // Target not found in our target list.
            return false;
        }
    }

gesior:
@NRH-AA 
I got a question about `lastPathUpdate`. I think idea is to block next path update for 300 ms for given creature?
It's used like:
```
lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
(...)
if (lastPathUpdate > OTSYS_TIME()) {
```
shouldn't it be called `nextPathUpdate` as we store date of next path update?
Also shouldn't that `>` be `<` [in few places]? So it gets into that `if`, only if it's more than `PATHFINDING_DELAY` since last update.

Right now we set some time value in future (ex. +300 ms) and then check, if it's bigger than current time, so for fast fleeing creature, it will update path every step, but for slow fleeing creature (over 0.3 sec step time) will be blocked by that if?

With few wrong IFs, it may work like old Mark fix 'force update every step' and it's impossible to process with 200+ players online (CPU goes 100%), even with 60% faster path-finding.

NRH-AA:
> @NRH-AA I got a question about `lastPathUpdate`. I think idea is to block next path update for 300 ms for given creature? It's used like:
> 
> ```
> lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
> (...)
> if (lastPathUpdate > OTSYS_TIME()) {
> ```
> 
> shouldn't it be called `nextPathUpdate` as we store date of next path update? Also shouldn't that `>` be `<` [in few places]? So it gets into that `if`, only if it's more than `PATHFINDING_DELAY` since last update.
> 
> Right now we set some time value in future (ex. +300 ms) and then check, if it's bigger than current time, so for fast fleeing creature, it will update path every step, but for slow fleeing creature (over 0.3 sec step time) will be blocked by that if?
> 
> With few wrong IFs, it may work like old Mark fix 'force update every step' and it's impossible to process with 200+ players online (CPU goes 100%), even with 60% faster path-finding.

Yes, it should set +300ms and new path should be allowed if lastPath < os.time() so thats a mistake. Also, 300ms is fast enough that even creatures with 1800 speed should run at a normal speed while monsters with slow movement will just walk again as soon as they can.

The only way it would work like the "fix" mark had which destroyed CPU's is if we call paths constantly even when they are not needed. Marks original algorithm used a lot more CPU than you would think (not the algorithm itself per se but how the fix worked, though the old code used a ton of memory comparatively). There was no checks for monsters so if a monster could see a player it would constantly run 250 nodes almost instantly concurrent. That included monsters that couldn't move, ect. So, the only way to make it as bad is if we made all monsters constantly draw paths with like 10ms max in between.

I will add the correct checks in this pull. Just give me 20 mins or so.

NRH-AA:
Working.

NRH-AA:
Okay all the current pulls are good now. There is an issue with fleeing monsters which I will look into. It has to do with danceStep not with pathfinding. It was throwing me off which is why it took me a second to get these nailed down plus the incorrect os_time() checks.

gesior:
> lastPath < os.time()

Can you rename `lastPathUpdate` to `nextPathUpdate`?
`lastPathUpdate ` suggests that it stores time of last path calculation (`OTSYS_TIME()`), so we should check:
`if (lastPathUpdate + getNumber(ConfigManager::PATHFINDING_DELAY) < OTSYS_TIME()) {`.
If we don't want to calculate `+` in every `if`, we should keep current logic and rename variable to `nextPathUpdate`.

NRH-AA:
When I do the fpp changes ill add it to that PR.

NRH-AA:
This fix is GTM

NRH-AA:
@ArturKnopik When you get time please merge this. Very important for first path draws to happen right away instead of taking 300ms.

I will test #4969 tonight and get it working properly. After that I think monster AI and pathfinding will pretty much be solid.

src/monster.cpp
```diff
@@ -1085,17 +1085,7 @@ bool Monster::walkToSpawn()
...
-{
-	Creature::onWalk();
-
-	if ((attackedCreature || followCreature) && isFleeing()) {
-		if (lastPathUpdate < OTSYS_TIME()) {
-			g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
-			lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
-		}
-	}
-}
```
ranisalt, line 1098: For comparison, the code for `forceUpdatePath()`:

```cpp
void Creature::forceUpdatePath()
{
	if (!attackedCreature && !followCreature) {
		return;
	}

	lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
	g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
}
```

https://github.com/ranisalt/forgottenserver/blob/62946948afc6fc751a3dcce6c70fff793477da70/src/creature.cpp#L145-L153

Basically the same, except for the check for `isFleeing()`, and it doesn't check if the path needs to be updated (hence `force`)

Issue #4941, opened 2025-06-25
Lua reloads


NRH-AA:
It seems when you create tables globally in lua and do /reset and the table repopulates. The memory is not freed from the previous tables. Does anyone know something about this?

ArturKnopik:
Probably Lua is still holding references to global variables, and the reload just shadows them. This prevents the reference count from dropping to zero. As long as any reference exists, the garbage collector won't remove this data

Zbizu:
aside from event registration (eg. adding onUse script to a lever), everything else is not cleared at the time of reload
so if you don't design your scripts to be reload-proof, they will execute second time at the time of reloading

example of script that runs every load/reload:
```lua
function test()
    print("loaded!")
end
test()
```

example of script that ignores reload:
```lua
function test()
    print("loaded!")
end

if not YOUR_CUSTOM_SYSTEM_INIT then
    test()
    YOUR_CUSTOM_SYSTEM_INIT = true
end
```

NRH-AA:
> aside from event registration (eg. adding onUse script to a lever), everything else is not cleared at the time of reload so if you don't design your scripts to be reload-proof, they will execute second time at the time of reloading
> 
> example of script that runs every load/reload:
> 
> function test()
>     print("loaded!")
> end
> test()
> example of script that ignores reload:
> 
> function test()
>     print("loaded!")
> end
> 
> if not YOUR_CUSTOM_SYSTEM_INIT then
>     test()
>     YOUR_CUSTOM_SYSTEM_INIT = true
> end

Interesting I will play around with that. Thanks.

NRH-AA:
So I am tinkering around with this problem. I have something that is working to dereference the memory. It also doesn't seem to be breaking anything but I am sure there is something I am missing. Here is what I did:

script.h
`void clear();`

script.cpp
```
void Scripts::clear()
{
	scriptInterface.reInitState();
	scriptInterface.initState(); 
}
```

game.cpp
```
g_scripts->clear();
g_scripts->loadScripts("scripts", false, true);
```

baseevents.cpp
void BaseEvents::reInitState(bool fromLua)
```
{
	getScriptInterface().reInitState();
	getScriptInterface().initState();
}
```

Pretty much if we do a reload it will reInitState and initState. It dereferences all memory and all of the scripts are working after a reload. Am I missing something?

PR #4949, opened 2025-07-02
Fix lua memory leak on reload

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Dereference all lua memory on reload.

**Issues addressed:** 
#4941

**How to test:** 
Create a global table in a lua script file, fill it with data. /reload scripts
Without this change the memory is not released.

### DIFF
--- a/src/baseevents.cpp
+++ b/src/baseevents.cpp
@@ -77,6 +77,7 @@ void BaseEvents::reInitState(bool fromLua)
 {
 	if (!fromLua) {
 		getScriptInterface().reInitState();
+		getScriptInterface().initState();
 	}
 }
 
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -5902,6 +5902,7 @@ bool Game::reload(ReloadTypes_t reloadType)
 			g_weapons->clear(true);
 			g_weapons->loadDefaults();
 			g_spells->clear(true);
+			g_scripts->clear(true);
 			g_scripts->loadScripts("scripts", false, true);
 			g_creatureEvents->removeInvalidEvents();
 			/*
--- a/src/script.cpp
+++ b/src/script.cpp
@@ -13,6 +13,12 @@ Scripts::Scripts() : scriptInterface("Scripts Interface") { scriptInterface.init
 
 Scripts::~Scripts() { scriptInterface.reInitState(); }
 
+void Scripts::clear(bool fromLua)
+{
+	scriptInterface.reInitState();
+	scriptInterface.initState();
+}
+
 bool Scripts::loadScripts(std::string folderName, bool isLib, bool reload)
 {
 	namespace fs = std::filesystem;
--- a/src/script.h
+++ b/src/script.h
@@ -12,6 +12,7 @@ public:
 	Scripts();
 	~Scripts();
 
+	void clear(bool fromLua);
 	bool loadScripts(std::string folderName, bool isLib, bool reload);
 	LuaScriptInterface& getScriptInterface() { return scriptInterface; }
 


NRH-AA:
Okay, as of now this is working. I had to add reloads because when we init it would dereference everything so we needed to reload the scripts to avoid nils.

Right now everything seems to reload fine. I tested most scripts: spells, weapons, actions, talkactions, ect. And it all works after the reloads. All memory references are removed so the memory when looking in task manager for the server decreases properly. I think this takes care of the lua memory leak. I believe @MillhioreBT is the one that created this for rev's. Is there a problem I am not seeing?

@Shawak @gesior @Codinablack @nekiro @ranisalt

gesior:
IDK how `/reload` can be made memory safe (no crashes) and not leaking.
`/reload` was always for dev env, not for production server and it was better to leak, than to crash, but it crashed server in some scenarios anyway.
There are many scenarios to test before we merge this:
- mixing scripts and data/lib code
- run other `/reload X` functions, how they affect each other
- removing events that are bind to some players (ex. onKill)
- what about `addEvent` functions created before reload?

We got to compare crashes before and after this PR changes.

ranisalt:
> IDK how `/reload` can be made memory safe (no crashes) and not leaking.

It can't unless we use smart pointers for _everything_ and even then, it may lead to weird behavior - functions that schedule themselves such as periodic events will linger forever

You're right, it should be blocked in production servers

NRH-AA:
> IDK how `/reload` can be made memory safe (no crashes) and not leaking. `/reload` was always for dev env, not for production server and it was better to leak, than to crash, but it crashed server in some scenarios anyway. There are many scenarios to test before we merge this:
> 
> * mixing scripts and data/lib code
> * run other `/reload X` functions, how they affect each other
> * removing events that are bind to some players (ex. onKill)
> * what about `addEvent` functions created before reload?
> 
> We got to compare crashes before and after this PR changes.



> > IDK how `/reload` can be made memory safe (no crashes) and not leaking.
> 
> It can't unless we use smart pointers for _everything_ and even then, it may lead to weird behavior - functions that schedule themselves such as periodic events will linger forever
> 
> You're right, it should be blocked in production servers

This all makes sense. I do wonder though, I have tested different scripts quite a bit without crashing. I am not saying it wont crash but the big thing that I noticed is after a /reload over 30mb of memory is released. So why is there so much memory removed without the memory being scripts that aren't loaded?

This probably is a lost cause and /reload shouldn't every be used as you said but there seems to be something here. Perhaps scripts are being loaded multiple times for some reason, idk. I don't have the time atm to really dig into testing/hunting.

ranisalt:
It's just weird to add these extra function calls, like `initState` after `reinitState`. `reinitState` should probably be modified to do more of what `initState` does instead, or just dropped and let's use `initState` only.

Same for `reload` after `clear`, `reload` should probably incorporate `clear` anyway

NRH-AA:
@ranisalt @gesior So after doing some more testing what I found is that g_scripts was responsible for the memory leak. We constantly reloaded scripts without ever clearing them out so all the old scripts held onto their old data. (this includes addEvents). I won't be able to do a bunch of testing for now but the code now works as stated before. I would imagine this will take care of any dangling addEvents that caused crashes before.

This is only effecting /reload scripts

Zbizu:
@NRH-AA
does this mean that we can stop reload-proofing our scripts once this gets fixed?

gesior:
> @NRH-AA does this mean that we can stop reload-proofing our scripts once this gets fixed?

I don't think so, this commit was related to memory leak, not problems that we get, when we reload Lua scripts.
Always when you reload Lua scripts all their states (variables) are destroyed, so script must assume that some Lua tables/variables may be empty/nil.

We still cannot keep list of player IDs, monsters, positions etc. in Lua. They should be kept in Player Storage or Global Storage - after you make Global Storage keep data in database and it's state in C++ like it does on TFS 0.x, but it does not on 1.x ( ex. https://github.com/gesior/forgottenserver-gesior/commit/d03f2e88dbf361b4caa9bd2e0d26eb08302a5314 ) - or make script retrive state from game (ex. using config with positions to check, if there are players/monsters/items).

PR #4952, opened 2025-07-03
Revert to old heuristic

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Revert to old calculation. The new one was messing up a few paths. Mainly small paths, it picked the correct distance path but it wasn't the right one.

**Issues addressed:** 
N/A

**How to test:** 
Check the difference after fix summon PR.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -667,7 +667,7 @@ static uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
 	uint16_t dx = std::abs(p1.getX() - p2.getX());
 	uint16_t dy = std::abs(p1.getY() - p2.getY());
-	return MAP_NORMALWALKCOST * (dx + dy);
+	return dx * dx + dy * dy;
 }
 
 bool Map::getPathMatching(const Creature& creature, const Position& targetPos, std::vector<Direction>& dirList,


NRH-AA:
This needs to be added. Other PR's wont work properly without it.

src/map.cpp
```diff
@@ -667,7 +667,7 @@ static uint16_t calculateHeuristic(const Position& p1, const Position& p2)
 {
 	uint16_t dx = std::abs(p1.getX() - p2.getX());
 	uint16_t dy = std::abs(p1.getY() - p2.getY());
-	return MAP_NORMALWALKCOST * (dx + dy);
```
ArturKnopik, line 670: remove MAP_NORMALWALKCOST definition if is not used.

NRH-AA: It is used in getMapWalkCost and getTileWalkCost

NRH-AA:
```
uint16_t AStarNodes::getMapWalkCost(AStarNode* node, const Position& neighborPos)
{
	if (std::abs(node->x - neighborPos.x) == std::abs(node->y - neighborPos.y)) {
		// diagonal movement extra cost
		return MAP_DIAGONALWALKCOST;
	}
	return MAP_NORMALWALKCOST;
}
uint16_t AStarNodes::getTileWalkCost(const Creature& creature, const Tile* tile)
{
	uint16_t cost = 0;
	if (tile->getTopVisibleCreature(&creature)) {
		// destroy creature cost
		cost += MAP_NORMALWALKCOST * 3;
	}
	if (const MagicField* field = tile->getFieldItem()) {
		CombatType_t combatType = field->getCombatType();
		const Monster* monster = creature.getMonster();
		if (!creature.isImmune(combatType) && !creature.hasCondition(Combat::DamageToConditionType(combatType)) &&
		    (monster && !monster->canWalkOnFieldType(combatType))) {
			cost += MAP_NORMALWALKCOST * 18;
		}
	}
	return cost;
}
```

ArturKnopik (approved):

Issue #4956, opened 2025-07-04
[Bug]: Visual Studio builds other than Release x64 does not work


gesior:
### By submitting this bug issue, you agree to the following.

- [x] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository
- [x] This bug has not been resolved in master branch
- [x] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below)

### Bug description

Visual Studio builds other than Release x64 does not work

### Possible Pull Requests which are to blame

- #4793 
- #4932 

### Steps to reproduce

1. Select build other than `Release` `x64`
2. Try to compile in Visual Studio

### Actual Behavior

Only Release x64 build compiles in Visual Studio.
Other builds return error `static_assert failed: 'Unicode support requires compiling with /utf-8'`.

### Expected Behavior

All Visual Studio builds should work.

### Backtrace

```cpp

```

Issue #4955, opened 2025-07-04
[Bug]: TFS always builds with build-in HTTP server disabled


gesior:
### By submitting this bug issue, you agree to the following.

- [x] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository
- [x] This bug has not been resolved in master branch
- [x] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below)

### Bug description

TFS always builds with build-in HTTP server disabled

### Possible Pull Requests which are to blame

- #4876 

### Steps to reproduce

1. Compile server using any compiler.
2. `HTTP` is not defined in C++.
3. Start server.
4. It does not show: `Starting HTTP server on 127.0.0.1:8080 with 1 threads.` and server does not listen on `8080`.

### Actual Behavior

HTTP server is always disabled.

### Expected Behavior

HTTP server should be enabled with `HTTP` parameter passed to `cmake` or `HTTP` compiler option in Visual Studio.

### Backtrace

```cpp

```

ArturKnopik:
Confirmed!
Fun fact: you can use an old engine that uses HTTP to get the character list, then shutdown server, run another one without http, and still connect(form characters list) — as long as the same database is used.

PR #4957, opened 2025-07-04
fix: build problems (#4955 #4956)

gesior:
This PR fixes build problems:
- TFS always builds with HTTP server disabled in C++ ( https://github.com/otland/forgottenserver/issues/4955 ), `-DHTTP=ON` and `-DHTTP=OFF` does not work, HTTP is always disabled
- builds other than Release x64 does not work in Visual Studio ( https://github.com/otland/forgottenserver/issues/4956 )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ option(HTTP "Enable HTTP support" ON)
 if (HTTP)
     list(APPEND VCPKG_MANIFEST_FEATURES "http")
     set(BOOST_REQUIRED_VERSION 1.75.0)
+    add_definitions(-DHTTP)
 endif()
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ option(HTTP "Enable HTTP support" ON)
 if (HTTP)
     list(APPEND VCPKG_MANIFEST_FEATURES "http")
     set(BOOST_REQUIRED_VERSION 1.75.0)
+    add_definitions(-DHTTP)
```
ranisalt, line 20: This should have been `add_compile_definitions(HTTP)`

gesior: It can be changed in some 'cleanup' PR with fixes for other compilation warnings (they are some in Release build in Visual Studio).

PR #4958, opened 2025-07-04
feat: make HTTP listen on the same IP as OTS

gesior:
This PR makes it possible to host OTS inside docker with working HTTP server.
After changes from this commit (and #4957 ), you can run local server inside docker with server IP set to `127.0.0.1` in `config.lua`.
It also fixes problem described in #4945 

Example `docker-compose.yaml`:
```
services:
  server:
    build:
      context: .
    restart: on-failure
    deploy:
      restart_policy:
        condition: any
        delay: 1s
    volumes:
      - .:/srv
    environment:
      MYSQL_HOST: mariadb
      MYSQL_USER: root
      MYSQL_PASSWORD: root
      MYSQL_DATABASE: tfs
    ports:
      - "7171:7171"
      - "7172:7172"
      - "8080:8080"

  mariadb:
    image: mariadb:latest
    container_name: mariadb-compose
    environment:
      MYSQL_ROOT_PASSWORD: root
      MYSQL_DATABASE: tfs
    volumes:
      - ./schema.sql:/docker-entrypoint-initdb.d/01-schema.sql
      - ./docker/data.sql:/docker-entrypoint-initdb.d/02-data.sql
    ports:
      - "3306:3306"
```
`docker/data.sql` file to create account `tfs@tfs` with password `tfs` and character `GOD`:
```
-- e-mail: tfs@tfs password: tfs
INSERT INTO `accounts` (`id`, `name`, `password`, `secret`, `type`, `premium_ends_at`, `email`, `creation`) VALUES ('1', 'tfs', 'e65018eee3ab1e3ecdfc67a91a2aa327434ab5b5', NULL, '6', '0', 'tfs@tfs', '0');
INSERT INTO `players` (`id`, `name`, `group_id`, `account_id`, `level`, `vocation`, `health`, `healthmax`, `experience`, `lookbody`, `lookfeet`, `lookhead`, `looklegs`, `looktype`, `lookaddons`, `lookmount`, `lookmounthead`, `lookmountbody`, `lookmountlegs`, `lookmountfeet`, `currentmount`, `randomizemount`, `direction`, `maglevel`, `mana`, `manamax`, `manaspent`, `soul`, `town_id`, `posx`, `posy`, `posz`, `conditions`, `cap`, `sex`, `lastlogin`, `lastip`, `save`, `skull`, `skulltime`, `lastlogout`, `blessings`, `onlinetime`, `deletion`, `balance`, `offlinetraining_time`, `offlinetraining_skill`, `stamina`, `skill_fist`, `skill_fist_tries`, `skill_club`, `skill_club_tries`, `skill_sword`, `skill_sword_tries`, `skill_axe`, `skill_axe_tries`, `skill_dist`, `skill_dist_tries`, `skill_shielding`, `skill_shielding_tries`, `skill_fishing`, `skill_fishing_tries`) VALUES ('1', 'GOD', '6', '1', '1', '0', '150', '150', '0', '0', '0', '0', '0', '136', '0', '0', '0', '0', '0', '0', '0', '0', '2', '0', '0', '0', '0', '0', '1', '0', '0', '0', NULL, '400', '0', '0', 0x30, '1', '0', '0', '0', '0', '0', '0', '0', '43200', '-1', '2520', '10', '0', '10', '0', '10', '0', '10', '0', '10', '0', '10', '0', '10', '0');
```

### DIFF
--- a/src/http/http.h
+++ b/src/http/http.h
@@ -4,7 +4,7 @@
 
 namespace tfs::http {
 
-void start(bool onlyOtsIP, std::string_view otsIP, unsigned short port = 8080, int threads = 1);
+void start(bool bindOnlyOtsIP, std::string_view otsIP, unsigned short port = 8080, int threads = 1);
 void stop();
 
 } // namespace tfs::http


src/http/http.cpp
```diff
@@ -17,15 +17,19 @@ std::vector<std::thread> workers = {};
...
 	if (port == 0 || threads < 1) {
 		return;
 	}
 
-	fmt::print(">> Starting HTTP server on {:s}:{:d} with {:d} threads.\n", address, port, threads);
+	asio::ip::address address = asio::ip::address_v6::any();
+	if (bindOnlyOtsIP) {
+		address = asio::ip::make_address(otsIP);
+	}
+	fmt::print(">> Starting HTTP server on {:s}:{:d} with {:d} threads.\n", address.to_string(), port, threads);
```
ranisalt, line 30: Setting `bindOnlyGlobalAddress` is exactly the same as setting `ip` to `0.0.0.0` (you can see the code inside `asio::ip::address_v6::any`, the default constructor zeroes the IP fields). Just changing the comment in `config.lua` to use `0.0.0.0` when using Docker really should suffice

gesior: > Setting `bindOnlyGlobalAddress` (OFF) is exactly the same as setting `ip` to `0.0.0.0`

Yes, it is, but we cannot change IP in config to `0.0.0.0`, because it's used in login server.
In build-in HTTP:
https://github.com/otland/forgottenserver/blob/c519cc6e6f9864e933c8a5a94d20afee538e70de/src/http/login.cpp#L132
In MyAAC (does not matter, it's not compatible with TFS new sessions code):
https://github.com/slawkens/myaac/blob/45d6047031c9c3a0e7e512dc5d15c75629aec5a2/login.php#L117

So we got to split IP used for listening and IP used as 'server IP' in login server, like other OTS ports do, except HTTP.
That's what this PR does. Makes HTTP server listens on other IP, than it returns in login server as server IP.

That's how Tibia works. After login to account, server returns characters list with list of worlds and their IPs/ports, so we got to get server IP from somewhere.

As I described in #4945 there are 2 problems with getting it "somehow" from operating system:
- There are servers (ex. Oracle Cloud, probably many coulds), that does not know own public IP. They know some cloud network local IP, but packets are routed to them from public IP.
- There are OTSes that want to listen only on IP defined in config.lua, not all IPs of server. They buy few IPs for single dedic and run few OTSes on 7171 using different IPs.

**EDIT:**
If you want to run public OTS inside docker, you would setup `ip` in `config.lua` to `1.2.3.4` (public IP of dedic/VPS), to let players login into OTS.

src/http/http.cpp
```diff
@@ -17,15 +17,19 @@ std::vector<std::thread> workers = {};
 
 } // namespace
 
-void tfs::http::start(std::string_view address, unsigned short port /*= 8080*/, int threads /*= 1*/)
+void tfs::http::start(bool bindOnlyOtsIP, std::string_view otsIP, unsigned short port /*= 8080*/, int threads /*= 1*/)
```
ranisalt, line 20: Let's use `ConfigManager` here then, instead of growing the list of arguments. In retrospect, it should have been like that since always.

gesior: This PR is to make TFS work without Docker and with Docker the same way.
This probably also solves problem of running public server on Windows with IP behind DMZ/port forward.

We can change `ConfigManager` in the next PR, but this should change the code in all places that use/not use `ConfigManager`, not just on the HTTP server. To make it consistent across TFS code.
It may be a part of https://github.com/otland/forgottenserver/issues/4963 solution.

ranisalt:
@Shawak can you confirm this solves the initial issue with Docker?

Shawak:
I didnt test it but it should since the game server now binds on ip address any (not sure if it only works for ~~ipv6~~ ipv4 tho)

gesior:
> (not sure if it only works for ipv6 tho)

Tested on Windows (.exe) and using Docker with Linux container, in both cases bind to `ip::address_v6::any()` worked with login to `127.0.0.1`. It's the same code all other ports use (7171/7172).

Issue #4960, opened 2025-07-05
[Bug]: Duplicate sessions tokens


maattch:
### By submitting this bug issue, you agree to the following.

- [x] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository
- [x] This bug has not been resolved in master branch
- [x] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below), Windows

### Bug description

Sessions are being duplicated

### Steps to reproduce

1. Start tfs and login a few times
2. Close tfs and open again
3. Try to login and get session error messages
4. Retry until you can finally login

### Actual Behavior

<img width="1077" height="167" alt="Image" src="https://github.com/user-attachments/assets/0f9495ef-821e-4f3a-b0b5-4ad12244e595" />

<img width="403" height="284" alt="Image" src="https://github.com/user-attachments/assets/bf469e65-54fe-4932-8396-c72ce04fdb80" />

Every time you restart TFS it's like the session token start all over again, like a sequence and not random/unique, which cause duplicate tokens.
I'm testing in Windows and OTC Mehah.

### Expected Behavior

Login with not problems.

### Backtrace

```cpp

```

ArturKnopik:
Some time ago Nekiro had the same problem, we didn't find a solution. The token is generated using random from the standard library

https://github.com/otland/forgottenserver/blob/c519cc6e6f9864e933c8a5a94d20afee538e70de/src/http/login.cpp#L87

https://github.com/otland/forgottenserver/blob/c519cc6e6f9864e933c8a5a94d20afee538e70de/src/tools.cpp#L210

probably std::random_device is a problem and it platform implementation

maattch:
I guess i found the problem, the random algorithm in protocollogin is not being seeded
https://github.com/otland/forgottenserver/blob/c519cc6e6f9864e933c8a5a94d20afee538e70de/src/protocollogin.cpp#L116-L118

`rbe` should receive a seed just like the one in `getRandomGenerator`

I will do some test and if successfully i will open a PR.

PR #4962, opened 2025-07-05
Fix duplicate session tokens

maattch:
### Pull Request Prelude

### Changes Proposed
Fix duplicate session tokens in protocollogin by using the same method that we are already using in http logins.

**Issues addressed:** 
#4960

The bug was cause by the  random algorithm not being seeded, using the default constructor means the same value every time you restart tfs.


ArturKnopik:
@nekiro can you check this fix?

Issue #4963, opened 2025-07-05
Draft: Session Token Format


ranisalt:
Historically, session tokens were derived directly from user credentials (email, password, optionally 2FA token), but TFS has recently (#4706) transitioned to randomly-generated base64-encoded token stored in the database. However, this change has introduced compatibility issues with existing AACs (such as [MyAAC](https://my-aac.org/)), which rely on the previous credential-based scheme.

## Background

The Forgotten Server has evolved significantly in its handling of account authentication. Previously, the session token provided to the client was simply a concatenation of credentials (see below).

This method was simple and easily interoperable with AACs and third-party tools, allowing for stateless verification on the server side.

In newer versions, TFS now issues a randomly-generated token, which is:

- Base64-encoded
- Stored in the database upon generation
- Used as a true bearer token for session continuity

This change was motivated by security concerns with the previous approach, but it has broken compatibility with existing account management websites and tooling.

See #4945 for initial discussion.

### Plaintext credential tokens (legacy)

Format
```
email\npassword[\n2FA]
```

#### Pros

- Stateless: requires no server-side session persistence
- Easy to implement: AACs and client tools can generate this format with minimal logic
- Portable: Can be easily reused across tools that know the user credentials

#### Cons

- Vulnerable to replay attacks
- Encourages storing plaintext credentials on clients
- No expiration or revocation mechanism: the token is valid as long as the credentials remain unchanged and 2FA token remains valid (probably lenghtened from 1 minute for session purposes)

### Random session tokens (current)
Format

base64(random_bytes(32))

Stored in a sessions table and tied to the account.

#### Pros

- Can be rotated and invalidated at any time
- Resistant to replay attacks and session hijacking
- Flexible: can store metadata like expiration and IP

#### Cons

- Requires state: server must store and validate tokens
- Breaks compatibility with current AACs expecting a credentials-based token
- More complex implementation, especially for third-party tools, requiring access to the database

## Problem

While the new session token mechanism theoretically improves security, its incompatibility with existing AACs causes usability issues for a large portion of the community. Without official support or clear documentation on the token format and usage, users are forced to modify AACs or downgrade to insecure practices.

There is a need for a well-documented, interoperable session token format that balances security with ease of integration.

### Option A: Hybrid Token Scheme

Introduce a transitional token format where AACs can optionally request a session token by submitting credentials (via HTTPS) to a /token endpoint, which returns a proper random token.

- AACs don’t generate the token manually
- Server maintains secure session tokens
[...]

PR #4967, opened 2025-07-05
JWT tokens for HTTP login

ranisalt:
### Pull Request Prelude

### Changes Proposed

See #4963 

Current issues:
- Tibia client sends a malformed packet for login (does not decrypt), could be my client misbehaving.
- OTClient crashes hard because the token is too long for it's expectations. Probably just a bug.


src/rsa.cpp
```diff
@@ -33,7 +33,11 @@ void decrypt(uint8_t* msg, size_t len)
 	EVP_PKEY_decrypt_init(pctx.get());
 	EVP_PKEY_CTX_set_rsa_padding(pctx.get(), RSA_NO_PADDING);
 
-	EVP_PKEY_decrypt(pctx.get(), msg, &len, msg, len);
+	auto ret = EVP_PKEY_decrypt(pctx.get(), msg, &len, msg, len);
+	if (ret <= 0) {
+		throw std::runtime_error(
```
Shawak: Could this be abused to crash the server with invalid packets?

ranisalt: Yes, I need to add a try/catch around it

gesior:
> Current issues:
> 
> * Tibia client sends a malformed packet for login (does not decrypt), could be my client misbehaving.
> * OTClient crashes hard because the token is too long for it's expectations. Probably just a bug.

I think I know why it happens. Session key is send encrypted using RSA. Tibia uses 1024-bit RSA key, which is 128 bytes, from which ~~117~~ 127 bytes are for data (~~11~~ 1 for padding).
(**EDIT:** I tested how many bytes has RSA packet after decrypting, it has 127, as Tibia protocol does not use padding.)
I saw the same problem on some 8.x server, when some OTS acc. maker allowed very long logins and passwords.

All bytes from `msg` in this part of code:
https://github.com/otland/forgottenserver/blob/c519cc6e6f9864e933c8a5a94d20afee538e70de/src/protocolgame.cpp#L367-L411
must fit together into 117 bytes. IDK how many bytes are there for session token.

JWT won't work in this scenario. We need much shorter format ex. first format proposed here:
https://github.com/otland/forgottenserver/issues/4963#issuecomment-3040349084

4 lines (no JSON, as it wastes too many characters for formatting):
1. account identifier (1-10 bytes)
2. account password hash truncated to 20 first characters (20 bytes)
3. 'created at' date as unix timestamp (10 bytes) - OPTIONAL, if we want sessions time limit
4. SHA1 sign: checksum of first 3 lines (concat using ; separator) with key.pem SHA1 (SHA1 of RSA key of OTS) added at end of SHA1 sign  truncated to 20 first characters (20 bytes)

I don't see any chance to put full client IP there. It would be possible, if we allow only IPv4, but with IPv6 it's too long.
If we want IP verification, we would need to add it as part of 4th line using IP as part of sign. Or add some weak IP verification by putting ex. 5 bytes of IP SHA1 hash and 5 bytes of sign of that IP.

**EDIT:**
Actually we don't need account ID and password as separate fields. We will know what account ID to load, because after session token, client sends character name, so all we need is some small checksum, to detect, if account ID/password changed ex. first 10 letters of SHA1.
4th line 'SHA1 sign' can be also shorter ex. 10 letters, it must be only secure enough to prevent someone from brute forcing it - we are still talking about brute forcing it by someone who knows password to account.

ranisalt:
I see no value in adding a checksum at the end. However, sending the account id + timestamp and signing those with HMAC should be safe enough to assume it isn't fake and still be secure. It's a thinner version of JWT without the fixed header + JSON boilerplate

gesior:
> I see no value in adding a checksum at the end.

By checksum, I meant "4. SHA1 sign:", which is like HMAC-SHA1, but as I described, we must truncate it to 10-20 ASCII characters (from 40 in hexadecimal format). I didn't want to call it HMAC, because HMAC is known format and we will use something custom.
`login.php`/`login.cpp` returns this value as part of JSON, so it must be in format compatible with JSON string, without extra encoding (base64/hex). Otherwise it will add more bytes.

> However, sending the account id + timestamp and signing those with HMAC should be safe enough to assume it isn't fake and still be secure.

Don't forget about password. We need some hash of password to make session token fail to login into game, after password to account is changed.

I checked how many bytes are used right now by Tibia 13.10, there are 50-70 free bytes (of 127 RSA-1024 bytes), even, if we allow character name with up to 30 letters.

Let's go back with discussion about format to https://github.com/otland/forgottenserver/issues/4963

Issue #4961, opened 2025-07-05
[Bug]: Potential followers crash


maattch:
### By submitting this bug issue, you agree to the following.

- [x] This is a bug in the software that resides in this repository, and not a support matter (use https://otland.net/forums/support.16/ for support)
- [x] This issue is reproducible without changes to the C++ code in this repository
- [x] This bug has not been resolved in master branch
- [x] There is no existing issue for this bug already

### Does this bug crash tfs?

not sure

### Server Version

1.7 (Master)

### Operation System

all (listed below)

### Bug description

Following creature can cause crash

### Possible Pull Requests which are to blame

- [x] #4637 

### Steps to reproduce

I could not reproduce a crash but just removing or killing the creature following you and walking away could potentialy crash.

### Actual Behavior

When a creature follows you, it is added to your `followers` list, which is updated every time you walk removing creatures that are too far from you. However, when the creature in the list is removed or killed, it is not removed from the list, causing a dangling pointer.
https://github.com/otland/forgottenserver/blob/c519cc6e6f9864e933c8a5a94d20afee538e70de/src/creature.cpp#L838-L841

I couldn't reproduce a crash, but the debugger reported corrupted positions on that line (e.g x=42678, y=39234, z=219), confirming that it's indeed a dangling pointer.

### Expected Behavior

Not crash?

### Backtrace

```cpp

```

ArturKnopik:
It's looks like possible nullptr dereference

NRH-AA:
Probably just need to replace:

creature.cpp

```
void Creature::removeFollowers()
{
	const Position& position = getPosition();

	followers.erase(std::remove_if(followers.begin(), followers.end(),
	                               [&position](Creature* creature) {
		                               const Position& followerPosition = creature->getPosition();
		                               uint16_t distance = position.getDistanceX(followerPosition) +
		                                                   position.getDistanceY(followerPosition);
		                               return distance >= Map::maxViewportX + Map::maxViewportY ||
		                                      position.z != followerPosition.z;
	                               }),
	                followers.end());
}
```

with

```
void Creature::removeFollowers()
{
	const Position& position = getPosition();

	followers.erase(std::remove_if(followers.begin(), followers.end(),
	                               [&position](Creature* creature) {
		                               if (!creature) {
			                               return true;
		                               }
		                               const Position& followerPosition = creature->getPosition();
		                               uint16_t distance = position.getDistanceX(followerPosition) +
[...]

PR #4968, opened 2025-07-06
Add reference counter

NRH-AA:
### Pull Request Prelude

### Changes Proposed

Add reference counter to followers.

**Issues addressed:** 
#4961

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -826,6 +826,7 @@ void Creature::addFollower(Creature* creature)
 {
 	if (!isFollower(creature)) {
 		followers.push_back(creature);
+		creature->incrementReferenceCounter();
 	}
 }
 
@@ -838,8 +839,12 @@ void Creature::removeFollowers()
 		                               const Position& followerPosition = creature->getPosition();
 		                               uint16_t distance = position.getDistanceX(followerPosition) +
 		                                                   position.getDistanceY(followerPosition);
-		                               return distance >= Map::maxViewportX + Map::maxViewportY ||
-		                                      position.z != followerPosition.z;
+		                               bool isInRemoveRange = distance >= Map::maxViewportX + Map::maxViewportY ||
+		                                                      position.z != followerPosition.z;
+		                               if (isInRemoveRange) {
+			                               creature->decrementReferenceCounter();
+		                               }
+		                               return isInRemoveRange;
 	                               }),
 	                followers.end());
 }


maattch:
The followers reference counter must be released if the creature is removed too, i guess a good approach is to create a function to clear the followers and call it in `Game::removeCreature` and in the creature destructor.

NRH-AA:
Okay, I created a way to release the follow creature in both places. Now I wonder if we need to use reference counter at all again?

src/creature.cpp
```diff
@@ -826,6 +828,18 @@ void Creature::addFollower(Creature* creature)
 {
 	if (!isFollower(creature)) {
 		followers.push_back(creature);
+		creature->incrementReferenceCounter();
+	}
+}
+
+void Creature::removeFollower(Creature* creature)
+{
+	if (isFollower(creature)) {
```
maattch: Since you are already calling `std::find` here, i don't think calling `isFollower` is needed

src/creature.cpp
```diff
@@ -838,12 +852,26 @@ void Creature::removeFollowers()
...
+		                                                      position.z != followerPosition.z;
+		                               if (isInRemoveRange) {
+			                               creature->decrementReferenceCounter();
+		                               }
+		                               return isInRemoveRange;
 	                               }),
 	                followers.end());
 }
 
+void Creature::releaseFollowCreature()
```
maattch: I didn't understand this one, since you are only calling it when the creature is removed/deleted would not be better to decrement reference counter for all followers and call `followers.clear()`?

gesior:
> Now I wonder if we need to use reference counter at all again?

I just though about same thing, while reviewing this code.
`followCreature` and `attackedCreature` pointers are kept in `Creature` object, but code does not increment/decrement references count. It just informs other creature whenever it's deleted/walks too far/change visibility (to invisible).

As long as you want to use:
https://github.com/otland/forgottenserver/blob/f169c6c3dc0d3f1f20ce09d3fdc32c81b05794ca/src/creature.cpp#L832-L845
you must keep increment/decrement references code to detect that creature was removed between that function calls.
To get rid of that function, you would need to iterate over `followers` and remove them in every place, where TFS removes `followCreature` and `attackedCreature`, but then it could use more CPU, than it did before all changes related to followers.

What problem that `followers` list fix at all? What would work wrong, if we remove it (stop calling `updateFollowersPaths`)? Maybe there is a better solution to that problem.

src/creature.cpp
NRH-AA: Uhhh, let me look at it again. I got pulled onto a job so I have been busy. I may have done the logic backwards x.x

src/creature.cpp
NRH-AA: You are exactly right. I have been on time constraints so my bad for putting up stuff without taking more time on it. I will fix in next commit.

src/creature.cpp
NRH-AA: So what this is doing is if the creature is following something (attacking or following) it will go into that other creatures following vector and remove itself. We don't want to clear the other creatures following vector and we are not concerned with the creature that is being deleted following vector because it will be destroyed.

This will make sure the deleted creature's reference counter (added when he started following/attacking) is deleted before the creature is deleted. Does that make sense?

NRH-AA: I may however, need to go through its own following vector and remove the reference counter of each creature in it. So yeah, I think you are right that dereference for all in following needs to be added. We don't need to clear it though.

NRH-AA:
> > Now I wonder if we need to use reference counter at all again?
> 
> I just though about same thing, while reviewing this code. `followCreature` and `attackedCreature` pointers are kept in `Creature` object, but code does not increment/decrement references count. It just informs other creature whenever it's deleted/walks too far/change visibility (to invisible).
> 
> As long as you want to use:
> 
> https://github.com/otland/forgottenserver/blob/f169c6c3dc0d3f1f20ce09d3fdc32c81b05794ca/src/creature.cpp#L832-L845
> 
> 
> you must keep increment/decrement references code to detect that creature was removed between that function calls.
> To get rid of that function, you would need to iterate over `followers` and remove them in every place, where TFS removes `followCreature` and `attackedCreature`, but then it could use more CPU, than it did before all changes related to followers.
> What problem that `followers` list fix at all? What would work wrong, if we remove it (stop calling `updateFollowersPaths`)? Maybe there is a better solution to that problem.

Yeah it's much easier to just use it for sure. 

As for updateFollowersPaths:
The current solution is pretty much just update the path whenever it changes (target moves). It is delayed by 300ms so it doesn't spam paths too fast but we also force a path update every second which handles anytime the following creatures somehow gets a follow target and doesn't update its path correctly. It's pretty much just a catch all to always make sure we start getting paths. There may be a better way to do it, but I can't think of anything better.

ArturKnopik (approved):
LGTM

PR #4969, opened 2025-07-06
Modify fpp & allow fleeing creatures to use pathMatching

NRH-AA:
### Pull Request Prelude

### Changes Proposed
Remove unused fpp vars, add some vars to handle different pathMatching senarios, change how we check for correct paths so it is easier to make different pathMatching senarios. Move fleeing monsters to pathMatching instead of distanceStep.

**Issues addressed:** 
N/A

**How to test:** 
https://youtu.be/VFpEPI7FSvU

Make sure when player walks to items on ground at a distance when moved (tested)
Make sure player walks to creatures when moving them
Make sure all paths are still being made.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -148,7 +148,7 @@ void Creature::forceUpdatePath()
 		return;
 	}
 
-	lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
+	nextPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
 	g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
 }
 
@@ -211,9 +211,9 @@ void Creature::onWalk()
 	}
 
 	if (attackedCreature || followCreature) {
-		if (lastPathUpdate < OTSYS_TIME()) {
+		if (nextPathUpdate < OTSYS_TIME()) {
 			g_dispatcher.addTask(createTask([id = getID()]() { g_game.updateCreatureWalk(id); }));
-			lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
+			nextPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
 		}
 	}
 }
@@ -852,7 +852,7 @@ void Creature::updateFollowersPaths()
 		if (follower != nullptr) {
 			const Position& followerPosition = follower->getPosition();
 
-			if (follower->lastPathUpdate < OTSYS_TIME()) {
+			if (follower->nextPathUpdate < OTSYS_TIME()) {
 				continue;
 			}
 
@@ -862,7 +862,7 @@ void Creature::updateFollowersPaths()
 			}
 
 			g_dispatcher.addTask(createTask([id = follower->getID()]() { g_game.updateCreatureWalk(id); }));
-			follower->lastPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
+			follower->nextPathUpdate = OTSYS_TIME() + getNumber(ConfigManager::PATHFINDING_DELAY);
 		}
 	}
 }
--- a/src/creature.h
+++ b/src/creature.h
@@ -382,7 +382,7 @@ protected:
 	std::vector<Creature*> followers;
 
 	uint64_t lastStep = 0;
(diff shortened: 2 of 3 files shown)


gesior:
Looks like new flee algorithm works better, it does not stuck in random places ex. current master algorithm stuck here forever:
image
New algorithm after few steps decided to run down from position 2.

New algorithm has some problem with keeping distance:
image
Master algorithm: walks positions 1 and 4 all time.
New algorithm: walks positions 1, 2 and sometimes (rarely) 3, which is wrong, as is reduces number of steps to player.

NRH-AA:
Yes, the code for fleeing monsters isn't perfect yet but the old code wasn't either. I will try to get it working better but I think having fleeing monsters that are only stuck until you get close is better than ones that are stuck even when you are close. (I know my other fix helped that a lot but there was still plenty of ways to get them stuck when the player was close to them). Now if the player is close to them they at least know how to get away.

I will try to get it working better if it will stop this PR from being merged but this is good bones for adding different "AI" monster behavior. It's also simplified so hopefully more people can understand what is happening.

We could also add the changes for this other than moving the fleeing monsters and just use my other fix. If that is preferred I can modify this PR and reopen my other one again.

Also as far as it moving to 3 which it shouldn't because its closer to player, there are many cases where this will allow the monster to find a way out of corners.

I think I have an idea to get it working better. I will test it out here in a bit.

gesior:
> Also as far as it moving to 3 which it shouldn't because its closer to player, there are many cases where this will allow the monster to find a way out of corners.

IDK, if they changed something on RL Tibia, but monsters getting stuck was a feature on RL Tibia and it was added in some version of TFS.
Before it, TFS always used full path finding with distance 11 SQM from player when fleeing, that way monsters never went into small rooms with no way out.
After changes, it runs ~straight line from player, so player can control fleeing monsters by his steps and ex. push it to small room where it will stuck, and player will finish it with spells/distance attacks.

To replicate RL Tibia 'dumb monsters', I had to add a lot of IFs in monsters logic and write second path finding algorithm (check tiles in 2-3 SQM range), that does not use A* to find path, when player is 1-2 SQM from monster. Depending on situation monster switched between path finding algorithms.

NRH-AA:
When I was doing tests I never saw pathfinding being called on fleeing creatures. I send a magiceffect out to show paths when they are drawn.

NRH-AA:
Alright, this should be a pretty good implementation now. Needs some more testing probably but from what I have tested everything is looking really good.

ArturKnopik:
image
looks like it's still broken

I had a case where monster, while fleeing, chose a square that was closer to the player then the rest
image

NRH-AA:
Alright, I will do some more testing and see if I can figure out what the issue is. That behavior would happen in the new getDistanceStep code I created. Give me some time and ill get it.

@ArturKnopik You may need to add the revert heuristic PR for this to work correctly.

ArturKnopik:
https://youtu.be/v7s-jcYlbws
looks like it's still broken :<

src/player.cpp
```diff
@@ -3373,7 +3373,6 @@ void Player::goToFollowCreature()
 void Player::getPathSearchParams(const Creature* creature, FindPathParams& fpp) const
 {
 	Creature::getPathSearchParams(creature, fpp);
-	fpp.fullPathSearch = true;
 }
```
ranisalt, line 3376: Remove this function from the class since it does nothing but call the base class function

src/npc.cpp
```diff
@@ -626,9 +626,8 @@ int NpcScriptInterface::luaActionMove(lua_State* L)
 
 int NpcScriptInterface::luaActionMoveTo(lua_State* L)
 {
-	// selfMoveTo(x, y, z[, minTargetDist = 1[, maxTargetDist = 1[, fullPathSearch = true[, clearSight = true[,
-	// maxSearchDist = 0]]]]]) selfMoveTo(position[, minTargetDist = 1[, maxTargetDist = 1[, fullPathSearch = true[,
-	// clearSight = true[, maxSearchDist = 0]]]]])
+	// selfMoveTo(x, y, z[, targetDist = 1[, clearSight = true[, maxSearchDist = 0]]])
+	// selfMoveTo(position[, targetDist = 1[, clearSight = true[, maxSearchDist = 0]]])
```
ranisalt, line 630: This is a breaking change, can the previous signature be emulated using the new signature? If yes, then add a compat layer

NRH-AA:
> https://youtu.be/v7s-jcYlbws looks like it's still broken :<

I am back to work on it.

NRH-AA:
This will have to be completely scratched. I will work on a different version.

PR #4970, opened 2025-07-08
Secure session tokens format (#4963)

gesior:
This PR is based on https://github.com/otland/forgottenserver/pull/4958 , because without it, it was very hard to test build-in HTTP server.

### TODO

- decide, if it's fine, that class `SessionToken` is in `game.cpp` / `game.h` - should it be moved to separate files `sessiontoken.cpp` and `sessiontoken.h`?
- add migration that drops `sessions` table in database

### Changes Proposed

Replace current TFS session tokens algorithm with new version that:
- does not use database
- detects when password to account changes
- has optional IP validation (default `false`, in `config.lua`: `sessionTokenIpVerification = false`)
- has configurable expiration time (default `30 days`, in `config.lua`: `sessionTokenExpirationTime = 30 * 24 * 3600`)
- uses SHA-256 sign to detect token modification with configurable secret sign key (default `""` [empty string], in `config.lua`: `sessionTokenSecretKey = ""`)
- works with PHP acc. makers (MyAAC, `login.php` that works with MyAAC: https://gist.github.com/gesior/6fb7e194a1aedb7980e98399bcca628d ) - I will make PR to MyAAC after this PR is merged into TFS


data/cpplinter.lua
```diff
@@ -2203,6 +2205,7 @@ configKeys = {
 	QUEST_TRACKER_PREMIUM_LIMIT = 42,
 	STAMINA_REGEN_MINUTE = 43,
 	STAMINA_REGEN_PREMIUM = 44,
+    SESSION_TOKEN_EXPIRATION_TIME = 47,
```
HopeItBuilds: shouldn't it be 45 btw?
```suggestion
	SESSION_TOKEN_EXPIRATION_TIME = 47,
```

gesior: No, but it's good you noticed that variable. It has 4 spaces offset, it should be Tab. I will fix it in next commit.

"Pathfinding" ( @NRH-AA ) commit missed that part of TFS code, added 2 `integer_config_t` keys that enum automatically numbered to `45` and `46`: 
https://github.com/otland/forgottenserver/blob/c519cc6e6f9864e933c8a5a94d20afee538e70de/src/configmanager.h#L124-L125
My new variable `SESSION_TOKEN_EXPIRATION_TIME` has value `47` in enum and that's what I put in `cpplinter.lua`.

Missing variables `PATHFINDING_INTERVAL` and `PATHFINDING_DELAY`, and their IDs must be added to that file in some future commit. There are some other config variables put in wrong places in C++ code of `luascripts.cpp`, that has to be fixed, but not as a part of my PR.

PR #4938, opened 2025-06-25
Add option to select libmariadb/libmysql

ranisalt:
### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Adds a CMake option to toggle between libmariadb (default) and libmysql. Using `default-features` is broken because it's impossible to _remove_ from the array, so selecting `libmysql` would cause _both_ to be installed - which is impossible because they conflict in install paths.

**Issues addressed:** failing builds in #4937 

**How to test:** 

[cont]: https://github.com/otland/forgottenserver/wiki/Contributing
[code]: https://github.com/otland/forgottenserver/wiki/TFS-Coding-Style-Guide

PR #4971, opened 2025-07-13
fix mariadb connection issue

### LINKED ISSUE #4954: [Bug]: MySQL Error Message: TLS/SSL error: SSL is required, but the server does not support it

- [x] This bug has not been resolved in master branch
- [x] There is no existing issue for this bug already

### Does this bug crash tfs?

no

### Server Version

1.7 (Master)

### Operation System

all (listed below)

### OS Description

_No response_

### Bug description

Server cannot connect to MySQL on server start

- #4932 

### Steps to reproduce

1. Compile server with vcpkg
2. Try to run server, it fails on connection to MySQL

It was first time reported here: https://otland.net/threads/server-does-not-open-tls-ssl-problem.290112/
In canary it was fixed in 2024.

I have this problem on Ubuntu and Windows when I compile with vcpkg.

### Actual Behavior

TFS compiled with vcpkg cannot connect to MySQL/MariaDB without SSL

### Expected Behavior

TFS compiled with vcpkg can connect to MySQL/MariaDB without SSL.

### Backtrace

```cpp

```

Zbizu:
tl;dr: This PR fixes "SSL is required, but the server does not support it"

### Pull Request Prelude

#### test 1 - libmysql

```json
"default-features": [
  "lua",
  "http",
  "libmysql"
],
"builtin-baseline": "0cf34c184ce990471435b5b9c92edcf7424930b1"
```

**RESULT:** ISSUE DID NOT OCCUR

#### test 2 - mariadb before changes

```json
"default-features": [
  "lua",
  "http",
  "libmariadb"
],
"builtin-baseline": "0cf34c184ce990471435b5b9c92edcf7424930b1"
```

**RESULT:** ISSUE REPRODUCED

#### test 3 - mariadb after changes

```json
"default-features": [
  "lua",
  "http",
  "libmariadb"
],
"builtin-baseline": "0cf34c184ce990471435b5b9c92edcf7424930b1"
```

**RESULT:** ISSUE NO LONGER OCCURS

#### test 4 - libmysql after changes

```json
"default-features": [
  "lua",
  "http",
  "libmysql"
],
"builtin-baseline": "0cf34c184ce990471435b5b9c92edcf7424930b1"
```

**RESULT:** build successful, connection ok

### Changes Proposed

Explicit SSL settings for MariaDB

**Issues addressed:** closes #4954

**How to test:** compile and run using visual studio

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -24,8 +24,8 @@ retry:
 // For more details see issue #4954 ( https://github.com/otland/forgottenserver/issues/4954 )
 #ifdef MARIADB_VERSION_ID
 	// this needs to be above "goto" otherwise it won't build
-	bool ssl_enforce = 0;
-	bool ssl_verify = 0;
+	bool ssl_enforce = false;
+	bool ssl_verify = false;
 #endif
 
 	tfs::detail::Mysql_ptr handle{mysql_init(nullptr)};


Zbizu:
apparently renaming a branch while the PR is open closes the PR without a way to reopen it 🤦

alysonjacomin:
@Zbizu Very good !

What do you think about merging this with PR #4938 ?

I will test this later

Zbizu:
@alysoncode I think that they should be handled separately. You suggest making a change to cmake settings, this one is about changes to database.cpp

alysonjacomin:
It makes sense to treat them separately, but this other PR should also be merged as it is a great complement to this topic

src/database.cpp
```diff
@@ -19,11 +19,28 @@ static tfs::detail::Mysql_ptr connectToDatabase(const bool retryIfError)
 	}
 	isFirstAttemptToConnect = false;
 
+// MariaDB requires explicit SSL settings to avoid the following error:
+// "SSL is required, but the server does not support it"
+// For more details see issue #4954 ( https://github.com/otland/forgottenserver/issues/4954 )
+#ifdef MARIADB_VERSION_ID
```
ranisalt, line 25: Do we need to guard it? What happens if you let this code build against libmysql?

Zbizu: top block will cause unused variable warning (if bottom one remain under that ifdef)
bottom block would be breaking libmysql builds (they would fail to compile) because these exact "MYSQL_OPT" globals are mariadb exclusives, this is why I placed ifdef there

alysonjacomin:
Lets go !

Tofame:
As I said in issue discussion, since some actually do want to use SSL like for example @Codinablack , that `false` values could be instead taken from config.

Zbizu:
@ranisalt any changes needed or it's ready to merge?

ranisalt:
I can't build without #4938 so I can't really say 🤷

Tofame:
configmanager.h, configmanager.cpp, config.lua.dist - true/false value.

ranisalt (approved):

ranisalt:
Let's get this merged to fix the issue and those that want to use SSL can submit a patch to add the config option then

PR #4973, opened 2025-07-22
Feat: add getInternalThing to LUA

alysoncode:
### Pull Request Prelude

### Changes Proposed

It is a resource that can be widely used in revscripts, especially when building OTClient modules.


src/luascript.cpp
```diff
@@ -4880,6 +4881,39 @@ int LuaScriptInterface::luaGameGetReturnMessage(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaGameGetInternalThing(lua_State* L)
+{
+	// Game.getInternalThing(player, pos, index, spriteId, type)
```
ArturKnopik: Providing spriteId is inconsistent with providing serverId. I don't like this approach.

alysonjacomin: https://github.com/otland/forgottenserver/blob/b577e5452b86ea5d6f42198e7219daccdedced4c/src/game.cpp#L290

Yeah, but this is used to get the real ID of the item within the function itself, I didn't create anything new, I just followed the pattern that is already used throughout game.cpp

alysonjacomin: Remembering that this function is commonly used when receiving an action from the client so that only then can the server perform the recognition and the necessary action

Tofame:
Maybe add comment below
`// Game.getInternalThing(player, pos, index, spriteId, type)`
do
`// spriteId equals clientId`
Not everyone may realise clientId === spriteId basically. 🤔

alysonjacomin:
@Tofame  Yeah, I think we can change it to spriteID/clientID, both make sense

ArturKnopik:
@alysonjacomin Unfortunately, I will not accept this PR, in my opinion it is an unnecessary feature, apart from the fact that it uses clientId where everywhere lua is using serverId

alysonjacomin:
@ArturKnopik  How can you say that a feature is unnecessary when this function is the main function used for all item/creature verification reception that are received by the client and we still don't use it in LUA ?

And as for the argument of using spriteID, it's the same principle as the "Game.getItemTypeByClientId(clientId)" function

This feature would be of enormous value for creating modules that need to check items/creatures

src/luascript.cpp
```diff
@@ -4880,6 +4881,39 @@ int LuaScriptInterface::luaGameGetReturnMessage(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaGameGetInternalThing(lua_State* L)
+{
+	// Game.getInternalThing(player, position, index, spriteId, type)
```
Tofame: then call it 'clientId' and not 'spriteId'. This is the confusing thing in TFS, as spriteId is in reality clientId. Or include the comment, just as I said.

alysonjacomin: @Tofame  True, to avoid any doubts it is better to use "clientId"

Codinablack:
> @alysonjacomin Unfortunately, I will not accept this PR, in my opinion it is an unnecessary feature, apart from the fact that it uses clientId where everywhere lua is using serverId

I have to somewhat agree with Arthur here... this is an "internal" function as it clearly states... these types of functions really shouldn't be just forwarded to lua without any concern... I personally feel like some certain functions should always be kept internal and never exposed to the lua api... I also don't really see a purpose of this function... 

OP you keep saying it's needed for OTC and modules, but in reality, modules have been being built just fine for a very long time without this function, and also, one of the core things they like to mention here, is that if there is something you need for OTC then you need to request it on OTC not TFS... I'm not so cut and dry on this stance, but it is the stance that TFS has had for a very long time now...

All that being said... what's wrong with just building a lua function to "getClientId" or "getRealID" instead of porting an internal function? I think this would be more appropriate, and would also allow for a much clearer understanding of what is actually going on for the lua side of things, not to mention it would be less costly to execute as well... Just my own personal opinion... 

Otherwise, the PR it's self doesn't look bad.

alysonjacomin:
@Codinablack Yeah, I understand your point perfectly, but this feature poses no risk and would be a great resource for creating modules that need to check an item that's far from the main backpack.

I also agree that modules have been made without this feature for a long time, but they're not as optimized because of the process involved in verifying that the player actually has "the right item in the right slot."

Well, this was just a PR to try to optimize the way modules are made and item checks in their exact slot provided, but if you can't see that and think it's unnecessary, I'll understand and close the PR.

alysonjacomin:
@ArturKnopik @Codinablack I'll give a very clear example of how this would be a great feature. If after this example you still find it unnecessary, we can safely close the PR.

Example: Trade Module - Imagine you've created a module where the player can freely drag items from their inventory to the Trade window. But let's assume they're in some backpacks outside the main backpack. How would you check for this item?

In my opinion, the best way to check for it is through this function

Codinablack:
> @Codinablack Yeah, I understand your point perfectly, but this feature poses no risk and would be a great resource for creating modules that need to check an item that's far from the main backpack.
> 
> I also agree that modules have been made without this feature for a long time, but they're not as optimized because of the process involved in verifying that the player actually has "the right item in the right slot."
> 
> Well, this was just a PR to try to optimize the way modules are made and item checks in their exact slot provided, but if you can't see that and think it's unnecessary, I'll understand and close the PR.

To be perfectly clear, I'm not a TFS Dev and so my opinion shouldn't be a reason to give up on something you want to see in this codebase... I just like to give my opinion :)

alysonjacomin:
@Codinablack  No problem bro, for me this is a healthy discussion on the subject, soon I think I will create an Issue just for a better discussion about adding basic features for a good connection with otclient but without straying from the TFS standard

alysonjacomin:
@ArturKnopik  Even after this last example explaining in detail why it would be something very useful, do you think it is still not feasible to add this function in LUA? Since it could be used in several other cases, in my opinion it would be something very useful

gesior:
@alysonjacomin
IDK why it was closed. It's function required to create any OTClient module that uses item/creature - except functions that are processed in C++ in OTS, which use `internalGetThing`.

@ArturKnopik 
You cannot use server ID, because client does not know it and sends only client ID.
This function is to process 'item' or 'creature' send from OTC, when player clicks on it and calls some custom module (ex. some 'upgrade item module', where you have to pick what item you want to upgrade, it can be in BP or in EQ slot, or on ground).

Items in BP (including pagination in depot), items on map and creatures on map are all passed as `x, y, z, stackpos` from Tibia client. For items it also sends item sprite ID (client ID), to detect that player clicked on some item, but it was moved by someone else before server received packet, so you cannot click on ladder to go up and use GFB rune someone throw on ladder position).
**That's Tibia protocol, not some OTC/OTS/TFS idea.**

Maybe `Game::internalGetThing` name is wrong. It should be `getClientThing` as it processes 'what tibia client of given player sees on given position + stackpos'.

alysonjacomin:
Hi @gesior 


It was automatically closed because I had deleted some branches and accidentally deleted this branch as well

My idea was exactly what you said. If I consider merging this into the code, I can open the PR again

PR #4979, opened 2025-08-04
feat: Allow only reporting first X players per IP in server status

DSpeichert:
### Pull Request Prelude

### Changes Proposed

**Issues addressed:**

Allows counting only first X players per IP. This is often mandated by some server lists. Defaults to disabled.

Originally discussed in [this post](https://otland.net/threads/very-important-rules-change-on-otservlist-org.247531/page-3#post-2407794).

**How to test:**
* set `statusCountMaxPlayersPerIp = 4`
* log in 5 characters (multi-client)
* watch server status report only 4 of them online


MillhioreBT (approved):

src/protocolstatus.cpp
```diff
@@ -104,7 +104,30 @@ void ProtocolStatus::sendStatusString()
...
+					}
+				} else {
+					playersPerIp[it.second->getIP()] = 1;
+					reportableOnlinePlayerCount++;
+				}
+			}
+		}
+	} else {
+		reportableOnlinePlayerCount = g_game.getPlayersOnline();
+	}
```
ranisalt: Easier to split counting + limiting phases. `std::map` default allocates if you access a key that hasn't been inserted before

```suggestion
	std::map<Connection::Address, uint32_t> playersPerIp;
	for (const auto& it : g_game.getPlayers()) {
		if (!it.second->getIP().is_unspecified()) {
			++playersPerIp[it.second->getIP()];
		}
	}
	
	for (auto&& [ip, players] : playersPerIp) {
		reportableOnlinePlayerCount += std::min(players, maxPlayersPerIp);
	}
```

ranisalt:
Not the status protocol getting updates in 2025 😭 any reason why we would allow more than 1 per IP?

ranisalt:
> the culture of spoofing that has been around for decades lol I blame it entirely on otservlist . org

Yeah it's as certain as death and taxes. But I mean, will someone ever set it to like 2, or 4? Otherwise I'd add a separate entry `unique` instead, listing how many IPs are connected, and keep `players` as is. Or, given how arbitrary the rule is, I'd even name it `number_for_otservlist`.

DSpeichert:
It does say `players` online, not `characters`, so I guess counting 1 per IP sounds reasonable. I also know server owners won't be happy about it if they are "allowed" up to 4/IP by a leading server list today.

src/protocolstatus.cpp
```diff
@@ -104,7 +106,25 @@ void ProtocolStatus::sendStatusString()
...
+	uint32_t maxPlayersPerIp = getNumber(ConfigManager::STATUS_COUNT_MAX_PLAYERS_PER_IP);
+	if (maxPlayersPerIp > 0) {
+		std::map<Connection::Address, uint32_t> playersPerIp;
+		for (const auto& it : g_game.getPlayers()) {
+			if (!it.second->getIP().is_unspecified()) {
+				++playersPerIp[it.second->getIP()];
+			}
+		}
+
+		for (auto &p: playersPerIp | std::views::values) {
```
ranisalt: This is **gorgeous**

(but needs formatting)

config.lua.dist
```diff
@@ -20,6 +20,7 @@ expFromPlayersLevelRange = 75
 -- NOTE: maxPlayers set to 0 means no limit
 -- NOTE: allowWalkthrough is only applicable to players
 -- NOTE: two-factor auth requires token and timestamp in session key
+-- NOTE: statusCountMaxPlayersPerIp allows you to only count first X players per IP in status response (0 = disabled)
```
ranisalt: Might be me being overly pedantic but the wording here seems of. Doesn't really matter if it's the first X or last X or any X:


```suggestion
-- NOTE: statusCountMaxPlayersPerIp allows you to only count up to X players per IP in status response (0 = disabled)
```

Feel free to mark as resolved if you'd rather ignore 😆

DSpeichert:
Is "disabled" the right default? Or do we want to default to 4? Or 1?

hugoj-goncalves:
in Brazil there are several ISPs (Copel/Ligga for example) that implements CGNAT, so basically multiple clients have the same public IPV4
other than that, it's not uncommon to have multiple computer / people playing from the same internet
there are even condominiums that have just a single contract with multiple repeaters (like 10+ buildings)

dunno about IPV6

I always find a hard sell ip restrictions other than geolocation / region based...

ranisalt:
> in Brazil there are several ISPs (Copel/Ligga for example) that implements CGNAT, so basically multiple clients have the same public IPV4

This is common everywhere now (I also am behind CGNAT in Sweden). Just shows how stupid all of this is. It's just layers of workarounds because lists don't want to implement a popularity rank better than number of online players reported. By the way, one can still spoof numbers with a decent VPN and a bunch of containers/VMs.

> dunno about IPV6

There are no issues with IPv6. In fact, you're usually allocated a few million addresses per network in IPv6, so you can have a million logged in characters with a different IP each.

PR #4980, opened 2025-08-08
Fead: add LUA function "getRuneSpells"

alysonjacomin:
### Pull Request Prelude

- [x] I have followed proper The Forgotten Server code styling.
- [x] I have read and understood the contribution guidelines before making this PR.

### Changes Proposed

It follows the same logic as "getInstantSpells" but applied to runes

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2896,6 +2896,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod(L, "Player", "getContainerById", LuaScriptInterface::luaPlayerGetContainerById);
 	registerMethod(L, "Player", "getContainerIndex", LuaScriptInterface::luaPlayerGetContainerIndex);
 
+	registerMethod(L, "Player", "getRuneSpells", LuaScriptInterface::luaPlayerGetRuneSpells);
 	registerMethod(L, "Player", "getInstantSpells", LuaScriptInterface::luaPlayerGetInstantSpells);
 	registerMethod(L, "Player", "canCast", LuaScriptInterface::luaPlayerCanCast);
 
@@ -11025,6 +11026,36 @@ int LuaScriptInterface::luaPlayerGetContainerIndex(lua_State* L)
 		lua_pushnumber(L, player->getContainerIndex(tfs::lua::getNumber<uint8_t>(L, 2)));
 	} else {
 		lua_pushnil(L);
+	}
+	return 1;
+}
+
+int LuaScriptInterface::luaPlayerGetRuneSpells(lua_State* L) {
+	// player:getRuneSpells()
+	Player* player = tfs::lua::getUserdata<Player>(L, 1);
+	if (!player) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	std::vector<const RuneSpell*> spells;
+	for (auto& spell : g_spells->getRuneSpells()) {
+		spells.push_back(&spell.second);
+	}
+
+	lua_createtable(L, spells.size(), 0);
+
+	int index = 0;
+	for (auto spell : spells) {
+		lua_createtable(L, 0, 4);
+
+		setField(L, "id", spell->getId());
+		setField(L, "name", spell->getName());
+		setField(L, "level", spell->getLevel());
+		setField(L, "mlevel", spell->getMagicLevel());
+
+		tfs::lua::setMetatable(L, -1, "Spell");
+		lua_rawseti(L, -2, ++index);
 	}
 	return 1;
 }
--- a/src/luascript.h
+++ b/src/luascript.h
@@ -776,6 +776,7 @@ private:
 	static int luaPlayerGetContainerById(lua_State* L);
 	static int luaPlayerGetContainerIndex(lua_State* L);
 
+	static int luaPlayerGetRuneSpells(lua_State* L);
 	static int luaPlayerGetInstantSpells(lua_State* L);
 	static int luaPlayerCanCast(lua_State* L);
 
--- a/src/spells.h
+++ b/src/spells.h
@@ -42,6 +42,7 @@ public:
 	static Position getCasterPosition(Creature* creature, Direction dir);
 	std::string_view getScriptBaseName() const override { return "spells"; }
 
+	const std::map<uint16_t, RuneSpell>& getRuneSpells() const { return runes; };
 	const std::map<std::string, InstantSpell>& getInstantSpells() const { return instants; };
 
 	void clearMaps(bool fromLua);


src/luascript.cpp
```diff
@@ -11029,6 +11030,36 @@ int LuaScriptInterface::luaPlayerGetContainerIndex(lua_State* L)
...
+	lua_createtable(L, spells.size(), 0);
+
+	int index = 0;
+	for (auto spell : spells) {
+		lua_createtable(L, 0, 4);
+
+		setField(L, "id", spell->getId());
+		setField(L, "name", spell->getName());
+		setField(L, "level", spell->getLevel());
+		setField(L, "mlevel", spell->getMagicLevel());
```
ranisalt: I think it should also push the rune item ID and charges

src/luascript.cpp
```diff
@@ -11029,6 +11030,36 @@ int LuaScriptInterface::luaPlayerGetContainerIndex(lua_State* L)
...
+	Player* player = tfs::lua::getUserdata<Player>(L, 1);
+	if (!player) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	std::vector<const RuneSpell*> spells;
+	for (auto& spell : g_spells->getRuneSpells()) {
+		spells.push_back(&spell.second);
+	}
```
ranisalt: Instead of duplicating data, pipe the map into [`std::views::values`](https://en.cppreference.com/w/cpp/ranges/values_view.html) i.e.

```
	for (auto spell : g_spells->getRuneSpells() | std::views::values) {
```

src/luascript.cpp
```diff
@@ -11029,6 +11030,36 @@ int LuaScriptInterface::luaPlayerGetContainerIndex(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaPlayerGetRuneSpells(lua_State* L) {
+	// player:getRuneSpells()
+	Player* player = tfs::lua::getUserdata<Player>(L, 1);
```
ranisalt, line 11036: This does not belong to `Player` (note how you don't use `player` anywhere). It should be in `Game`.

ranisalt:
> It follows the same logic as "getInstantSpells" but applied to runes

It does not, `getInstantSpells` will return spells the player can cast _including spells to create runes_. `getRuneSpells` is the effect when _using_ a rune instead.

src/luascript.cpp
alysonjacomin: True

src/luascript.cpp
alysonjacomin: You had a good observation, but it is possible to add vocation limitations for rune spells as well as instant spells, so wouldn't the correct thing be to add a function similar to "canCast" that exists for instant spells?

alysonjacomin: Therefore, it would be necessary to maintain this function for the player, to obtain only the rune spells of the player's correct vocation

src/luascript.cpp
alysonjacomin: Very good, I followed the model already used in Instant Spell. Should we apply this same fix to Instant Spell?

Or, in the near future, we can apply this optimization to the entire source code

ranisalt: Yes, but do that in a separate PR. No need to rewrite everything now 😝

src/luascript.cpp
ranisalt: > You had a good observation, but it is possible to add vocation limitations for rune spells as well as instant spells, so wouldn't the correct thing be to add a function similar to "canCast" that exists for instant spells?

Maybe, but that would be a weird interaction. `Player:getInstantSpells()` should really be `Player:getLearntSpells()` instead, and then have `Game.getInstantSpells()` and `Game.getRuneSpells()`. What do you think?

alysonjacomin: This is perfect as these are functions that should be used differently

I'll finish this PR then and then I can make another PR to fix the Instant Spells you mentioned above, and another PR to create "Game.getInstantSpells()" and "Game.getRuneSpells()"

alysonjacomin:
@ranisalt  In this PR I created a function "canUse" for get only rune spells per player vocation

I also created PR #4988  to get all Rune Spells and all Instant Spells using "std::views::values" as you mentioned

src/spells.cpp
```diff
@@ -1220,3 +1220,20 @@ bool RuneSpell::executeCastSpell(Creature* creature, const LuaVariant& var, bool
...
+	if (player->hasFlag(PlayerFlag_IgnoreSpellCheck)) {
+		return true;
+	}
+
+	if (vocationSpellMap.empty() || hasVocationSpellMap(player->getVocationId())) {
+		return true;
+	}
+
+	return false;
+}
```
ranisalt: Files need to have a trailing newline (i.e. not this red circle)

src/spells.cpp
```diff
@@ -1220,3 +1220,20 @@ bool RuneSpell::executeCastSpell(Creature* creature, const LuaVariant& var, bool
...
+
+	if (player->hasFlag(PlayerFlag_IgnoreSpellCheck)) {
+		return true;
+	}
+
+	if (vocationSpellMap.empty() || hasVocationSpellMap(player->getVocationId())) {
+		return true;
+	}
+
+	return false;
```
ranisalt: Let's simplify these statements, also `hasVocationSpellMap` already checks emptiness

```suggestion
	return hasVocationSpellMap(player->getVocationId());
```

src/luascript.cpp
ranisalt: I noticed on the other PR but this too should use `pushSpell`

By doing that the rest of the values will be available through the `Spell` metadata e.g. `runeId()` and `charges()`

MillhioreBT: In fact, the pushSpell method doesn't work, and assigning a metatable to a simple table is also an error, since the Spell metatable expects user data, not simple tables.

Attempting to use these tables by calling methods like name() would cause an error/crash

There is already some [current code](https://github.com/otland/forgottenserver/blob/1cfda85bc1f97b99d621010a62ac00567d70ce3b/src/events.cpp#L1433) that has this problem.

alysonjacomin: @MillhioreBT  See how I did it in PR #4988, I made it return the entire spell, this way we can get all the parameters of a spell

MillhioreBT: > @MillhioreBT See how I did it in PR #4988, I made it return the entire spell, this way we can get all the parameters of a spell

Yes, I think the pushSpell method needs to be removed and the spellbook.lua script updated accordingly.
But yes, pushing a userdata to the spell is the correct way, since we can use the Spell metatable to access it.

alysonjacomin: > Yes, I think the pushSpell method needs to be removed and the spellbook.lua script updated accordingly. But yes, pushing a userdata to the spell is the correct way, since we can use the Spell metatable to access it.

I'll follow this path then, and I can now open another PR to fix the getInstantSpell function and another PR to remove the pushSpell function that is not usable

MillhioreBT:
@alysonjacomin Fix this PR, just delete this line.

`tfs::lua::setMetatable(L, -1, "Spell");`

But it would be better if you **pushed the `usardata` of the spell and left the `set metatable`**.
Then, in another PR, we can **change this to `getInstantSpells`** and **adjust the Lua script `spellbook.lua`**.

alysonjacomin:
I was thinking about using the same methodology applied in PR #4988 

@MillhioreBT What do you think about this?

MillhioreBT:
> I was thinking about using the same methodology applied in PR #4988
> 
> @MillhioreBT What do you think about this?

do it, it's the right choice.

MillhioreBT (approved):

ranisalt:
I think this will crash. You're copying the spell maps, pushing the copy memory addresses to Lua space via `pushUserdata`, then deleting the copy, making the Lua userdata point to released memory.

MillhioreBT:
> I think this will crash. You're copying the spell maps, pushing the copy memory addresses to Lua space via `pushUserdata`, then deleting the copy, making the Lua userdata point to released memory.

Pointers are being copied, that is correct. Something like this could happen, possibly with a script or spell reload, especially if these **userdatas** are saved in **Lua** for later use.

---

Something similar to what happened with the **Combat objects** in the past.

---

I suppose until we move certain things to **smart pointers**.

This is something that will also happen with the other changes related to player:getInstantSpells and getRuneSpells, etc...
the good news is that migrating spells to smart pointers is not that difficult since they do not derive from the Thing class.

We just have to convert the base class Event to shared_ptr, and all events will then be smart pointers

alysonjacomin:
Currently, it's necessary to obtain getRuneSpells and getInstantSpells by reference; I'm waiting for PR #5017 to be approved to implement this change

PR #4982, opened 2025-08-11
fix: clean up customAttrMap when empty after erase

Tofame:
After removing custom attribute, if the map of custom attributes is empty, we should remove it and not keep it in the memory, as it prevents item::equals method from properly comparing items.

### Pull Request Prelude

- [ X] I have followed [proper The Forgotten Server code styling][code].
- [ X] I have read and understood the [contribution guidelines][cont] before making this PR.
- [X ] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Cleanup the custom attribute map and the ITEM_ATTRIBUTE_CUSTOM, if while removing the attribute, it turned out to be the only one in the map (making it empty).

**How to test:** 
Have 23 arrows.
Put custom attribute on 3 arrows
Pick 1 out of 3 arrows and remove the custom attribute from it.
Now that 1 arrow should stack with the 20 arrows, however that is not the case until relog.

https://github.com/user-attachments/assets/26dbf298-52df-43ea-ae60-08c43decc347

Tbh it's not the TFS 1.7 I am using (and can't use 1.7 either), so would be nice if another person confirmed. I just looked at the code and since it's basically 1:1 I guess the bug is also in here :)

### DIFF
--- a/src/item.h
+++ b/src/item.h
@@ -452,6 +452,12 @@ private:
 			auto lowercaseKey = boost::algorithm::to_lower_copy(std::string{key});
 			if (auto it = customAttrMap->find(lowercaseKey); it != customAttrMap->end()) {
 				customAttrMap->erase(it);
+
+				if (customAttrMap->empty()) {
+					delete customAttrMap;
+					getAttr(ITEM_ATTRIBUTE_CUSTOM).value.custom = nullptr;
+					removeAttribute(ITEM_ATTRIBUTE_CUSTOM);
+				}
 				return true;
 			}
 		}


alysonjacomin:
I think today or tomorrow I can test with the tfs-main to be sure about this

Codinablack:
LGTM 👍

alysonjacomin:
https://github.com/user-attachments/assets/884dd9d9-e804-4717-80ef-e2b222e1faeb

@Tofame Yeah, in tfs-main this also happens

alysonjacomin:
NOTE: with this modification it works correctly

Tofame:
Needs to be thought out why it would crash. I was checking if destructor works, and it does, so I need to delve deeper. Im away though, so need a few days.

ArturKnopik:
@Tofame can you describe when crash occurred?

Codinablack:
> @Tofame can you describe when crash occurred?

crash occurs when applying "setCustomAttribute". For me, I tried on a necklace that had charges to add to it a custom attribute, for another user, I believe they were trying to set the custom attribute to something it already had, not entirely sure.... I didn't save my call stack... I was ofc testing on my own project, but it's worth knowing that I have not at all changed anything about the class "ItemAttributes" or CustomAttributes either... 

Best thing I could say to reproduce it, is to try to add a custom attribute to a necklace with charges...

I did a few things to try to eliminate this specific change as the cause, but ultimately, removing this change stopped the crashes.... I didn't have the time to dig deep into this specific matter yet so I just left it commented in my sources, maybe this week I will have time to see what extra changes are needed to go along with this one to get it to work properly

alysonjacomin:
@Codinablack  I tried to replicate this and couldn't

Codinablack:
> @Codinablack I tried to replicate this and couldn't

So I went back and caused the crash again to see the callstack after I didn't notice any changes to the item attributes and you said you couldn't replicate it... turns out, I did actually make one single change to the item's attributes back at the beginning of the year'sh... I replaced the hashmaps with GTL (aka phmap, aka abseil) containers, and they are the last X amount of lines in the callstack... so, likely, this is something to do with their internals and TFS wouldn't have same crash... but keep in mind, these containers were originally designed by google years ago and have been ported to C++ in the GTL/PHMAP library to be completely compatible replacements for standard library containers... so if we are somehow breaking them with this commit, there may be another underlying problem that is just not apparent right now (or, even more likely I could have been using the wrong replacement container in this case)

TLDR; Probably the crash is exclusive to my codebase, not TFS.

Tofame:
Yea, but even when you are having this difference, I got also reports of having a crash from Nekiro too, when he also was fixing it months ago.
I havent experienced it, but I still want to take another look at the code and try to crash myself to see if its also applying to 1.7, but that's gonna be when I'm back from my trip.

alysonjacomin:
Hi guys, I may make a few more extra attempts throughout the week as well

MillhioreBT:
Good detail, this PR fixes it, although maybe it would be better to fix the Item::equals method.

MillhioreBT (approved):

Tofame:
obraz

I think that's the crash case scenario that might happen.
@nekiro @MillhioreBT @alysonjacomin @ArturKnopik

So, if we call ``setCustomAttribute`` **my change actually makes the destructor work properly**, and so after the ``if`` is entered, it removes custom attribute -> properly destructs now -> makes that value nullptr.

And so we will get crash in the line below when we try to use that value.

I will just need to do instead of ``else`` part, turn it into:
```
	if (hasAttribute(ITEM_ATTRIBUTE_CUSTOM)) {
		removeCustomAttribute(key);
	}

	if (!getAttr(ITEM_ATTRIBUTE_CUSTOM).value.custom) {
		getAttr(ITEM_ATTRIBUTE_CUSTOM).value.custom = new CustomAttributeMap();
	}
```
Any confirmation that I'm actually right about that happening?

			
Btw, MillhioreBT, regarding your comment on Item::equals, there is nothing to fix in there, atleast in relation to the issue we talk about. 
What we are fixing here right now, is clearing up stuff - especially the custom attribute bit - and this is the role of the ``removeCustomAttribute`` method, so items will no longer have a _CUSTOM bit on them while not even having custom attributes in the map.

PR #4988, opened 2025-08-17
Create LUA functions for get ALL spells

alysonjacomin:
### Pull Request Prelude

### Changes Proposed

* Game.getRuneSpells() for get all Rune Spells

* Game.getInstantSpells() for get all Instant Spells

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -4845,13 +4845,13 @@ int LuaScriptInterface::luaGameGetVocations(lua_State* L)
 int LuaScriptInterface::luaGameGetRuneSpells(lua_State* L)
 {
 	// Game.getRuneSpells()
-	const auto& runeSpells = g_spells->getRuneSpells();
+	auto runeSpells = g_spells->getRuneSpells();
 
 	lua_createtable(L, runeSpells.size(), 0);
 
 	int index = 0;
-	for (const auto& spell : runeSpells | std::views::values) {
-		tfs::lua::pushUserdata<Spell>(L, const_cast<RuneSpell*>(&spell));
+	for (auto& spell : runeSpells | std::views::values) {
+		tfs::lua::pushUserdata<Spell>(L, &spell);
 		tfs::lua::setMetatable(L, -1, "Spell");
 		lua_rawseti(L, -2, ++index);
 	}
@@ -4862,13 +4862,13 @@ int LuaScriptInterface::luaGameGetRuneSpells(lua_State* L)
 int LuaScriptInterface::luaGameGetInstantSpells(lua_State* L)
 {
 	// Game.getInstantSpells()
-	const auto& instantSpells = g_spells->getInstantSpells();
+	auto instantSpells = g_spells->getInstantSpells();
 
 	lua_createtable(L, instantSpells.size(), 0);
 
 	int index = 0;
-	for (const auto& spell : instantSpells | std::views::values) {
-		tfs::lua::pushUserdata<Spell>(L, const_cast<InstantSpell*>(&spell));
+	for (auto& spell : instantSpells | std::views::values) {
+		tfs::lua::pushUserdata<Spell>(L, &spell);
 		tfs::lua::setMetatable(L, -1, "Spell");
 		lua_rawseti(L, -2, ++index);
 	}


alysonjacomin:
@ranisalt  I used "std::views::values" as you mentioned in PR #4980.

Instead of creating a subtable with (id, name, and others), I used const_cast to get the spell itself. This way, it's possible to get any information about the spell itself. What did you think about this ?

src/luascript.cpp
```diff
@@ -4840,6 +4842,40 @@ int LuaScriptInterface::luaGameGetVocations(lua_State* L)
...
+
+int LuaScriptInterface::luaGameGetInstantSpells(lua_State* L)
+{
+	// Game.getInstantSpells()
+	const auto& instantSpells = g_spells->getInstantSpells();
+
+	lua_createtable(L, instantSpells.size(), 0);
+
+	int index = 0;
+	for (const auto& spell : instantSpells | std::views::values) {
```
Shawak: I think in another PR `std::views::values` was removed because of being a cpp 20 feature

alysonjacomin: Oh, I only used because @ranisalt  requested it in other PR

ranisalt: We require C++20

https://github.com/otland/forgottenserver/blob/1cfda85bc1f97b99d621010a62ac00567d70ce3b/CMakeLists.txt#L96-L97

src/luascript.cpp
```diff
@@ -4840,6 +4842,40 @@ int LuaScriptInterface::luaGameGetVocations(lua_State* L)
...
+	// Game.getRuneSpells()
+	const auto& runeSpells = g_spells->getRuneSpells();
+
+	lua_createtable(L, runeSpells.size(), 0);
+
+	int index = 0;
+	for (const auto& spell : runeSpells | std::views::values) {
+		tfs::lua::pushUserdata<Spell>(L, const_cast<RuneSpell*>(&spell));
+		tfs::lua::setMetatable(L, -1, "Spell");
+		lua_rawseti(L, -2, ++index);
```
ranisalt: Use the `pushSpell` function instead

alysonjacomin: But the "pushSpell" function returns a table with name, level, mlevel, mana, and manapercent

My idea was to return the spell freely, so the we could then freely obtain the information in LUA, for example: spell:name(), spell:vocation(), and others

Don't you think this is a good idea?

alysonjacomin: Or we can return all, absolutely all the parameters of an Instant Spell and all the parameters of a Rune Spell

ranisalt: The `Spell` metatable will already give it methods for all relevant attributes:

https://github.com/otland/forgottenserver/blob/1cfda85bc1f97b99d621010a62ac00567d70ce3b/src/luascript.cpp#L3371-L3414

alysonjacomin: Yeah, that's what I did, but you said to use the pushSpell function since it returns a table with just a few attributes (I'm not sure if I understood what you said correctly)

alysonjacomin:
```
void tfs::lua::pushSpell(lua_State* L, const Spell& spell)
{
	lua_createtable(L, 0, 5);
	setField(L, "name", spell.getName());
	setField(L, "level", spell.getLevel());
	setField(L, "mlevel", spell.getMagicLevel());
	setField(L, "mana", spell.getMana());
	setField(L, "manapercent", spell.getManaPercent());
	setMetatable(L, -1, "Spell");
}
```

src/luascript.cpp
ArturKnopik: Personally i preffer style from 
https://en.cppreference.com/w/cpp/ranges/keys_view.html

```cpp
for (std::string const& name : std::views::keys(quark_mass_charge))

for (const double mass : std::views::values(quark_mass_charge))
```

instead

```cpp
for (const auto& spell : instantSpells | std::views::values)
```

src/luascript.cpp
```diff
@@ -4840,6 +4842,40 @@ int LuaScriptInterface::luaGameGetVocations(lua_State* L)
...
+int LuaScriptInterface::luaGameGetRuneSpells(lua_State* L)
+{
+	// Game.getRuneSpells()
+	const auto& runeSpells = g_spells->getRuneSpells();
+
+	lua_createtable(L, runeSpells.size(), 0);
+
+	int index = 0;
+	for (const auto& spell : runeSpells | std::views::values) {
+		tfs::lua::pushUserdata<Spell>(L, const_cast<RuneSpell*>(&spell));
```
ranisalt: Here

```suggestion
		tfs::lua::pushSpell<Spell>(L, const_cast<RuneSpell*>(&spell));
```

src/luascript.cpp
ranisalt: I added a comment to the correct line 😆

src/luascript.cpp
alysonjacomin: @ranisalt  pushSpell dont use template, and if we just change it to "tfs::lua::push Spell(E, spell;" it will return just "name, level, mlevel, mana, manapercent" is that what you mean?

src/luascript.cpp
alysonjacomin: @ArturKnopik  Honestly, I've always used the "old way." I recently researched these new methods in more depth with this recommendation from Ranisalt.

The three styles are similar, but I liked the style recommended by Ranisalt the most. It avoids copying and also makes what we're referencing more informative

src/luascript.cpp
ranisalt: I lack the knowledge to properly help here. Let's go with what @MillhioreBT says.

alysonjacomin: @MillhioreBT  What we talked about in the other PR, so the way discussed here is the correct way? Should we use "tfs::lua::pushUserdata<Spell>(L, const_cast<RuneSpell*>(&spell));" ?

MillhioreBT: Yes, the correct way should be to push a userdata to the spell, because the table we’re currently using causes problems if used with the Spell metatable. I think there’s only 1 Lua script that makes use of that table, so it should also be updated accordingly.

alysonjacomin: Nice bro, I'm sick at the moment, as soon as I get better I'll update the PRs

alysonjacomin:
@MillhioreBT  Check if everything is correct now, if so I can apply for the other PRs too

PR #4989, opened 2025-08-24
Remove redundant despawnRange check

Codinablack:
### Pull Request Prelude

- [x ] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed

Removal of redundant evaluation

Currently this bit of code is inside this function twice, for no apparent reason, we are just removing one instance of it. You can find the second instance of this bit of code a couple lines beneath the next evaluation.

**Issues addressed:** Redundant Code

**How to test:** Compile and Run

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1910,10 +1910,6 @@ bool Monster::isInSpawnRange(const Position& pos) const
 		return true;
 	}
 
-	if (Monster::despawnRadius == 0) {
-		return true;
-	}
-
 	if (!Spawns::isInZone(masterPos, Monster::despawnRadius, pos)) {
 		return false;
 	}


Codinablack:
> one is despawn**Range** and the other is despawn**Radius**
> 
> I would do something like this thought:
> 
> ```
> bool Monster::isInSpawnRange(const Position& pos) const
> {
>     if (!spawn || Monster::despawnRadius == 0) {
>         return true;
>     }
> 
>     return Spawns::isInZone(masterPos, Monster::despawnRadius, pos) &&
>            (Monster::despawnRange == 0 || pos.getDistanceZ(masterPos) <= Monster::despawnRange);
> }
> ```

You are right... I must have been blind! This reading the computer screen in the dark doesn't help...

I checked their default values, which are indeed different... but they are both only ever even used right here in this function... so I'm not really sure what is their "intended" difference from one another??? I mean, it seems the radius is checked if it's in the "zone" but the radius is only checked for z distance... pretty confusing... 

Anyways, good catch... I'll close the PR

PR #4990, opened 2025-08-25
Eliminate double randomizing directions in Monster::getDanceStep

Codinablack:
### Pull Request Prelude

- [ x] I have followed [proper The Forgotten Server code styling][code].

### Changes Proposed & Issues Addressed

Before this change, the code shuffles the directions around, and then randomly picks a direction; Since we are already picking one of the directions at random, it serves no purpose to randomize their order.

The change is to eliminate the shuffle.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1352,7 +1352,6 @@ bool Monster::getDanceStep(const Position& creaturePos, Direction& direction, bo
 	}
 
 	if (!dirList.empty()) {
-		std::shuffle(dirList.begin(), dirList.end(), getRandomGenerator());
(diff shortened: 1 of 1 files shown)


MillhioreBT (approved):
LGTM

nekiro:
> you are correct
> 
> I would use something like this thought:
> 
> ```
> if (!dirList.empty()) {
>      std::vector<Direction> sampled;
>      std::sample(dirList.begin(), dirList.end(), std::back_inserter(sampled), 1, getRandomGenerator());
>      direction = sampled[0];
>      return true;
> }
> ```

additional penalty for instantiating vector for no reason? no thanks

PR #5000, opened 2025-09-07
fix onTickCondition missing nullcheck

kaleohanopahala:
Check to ensure the creature has a valid tile before querying for field items, preventing potential null pointer dereferences.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -931,8 +931,8 @@ void Creature::onEndCondition(ConditionType_t)
 
 void Creature::onTickCondition(ConditionType_t type, bool& bRemove)
 {
-	const auto &tile = getTile();
-	const auto &field = tile ? tile->getFieldItem() : nullptr;
+	const auto& tile = getTile();
+	const auto& field = tile ? tile->getFieldItem() : nullptr;
 	if (!field) {
 		return;
 	}
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -931,8 +931,12 @@ void Creature::onEndCondition(ConditionType_t)
 
 void Creature::onTickCondition(ConditionType_t type, bool& bRemove)
 {
-	const auto& tile = getTile();
-	const auto& field = tile ? tile->getFieldItem() : nullptr;
+	const Tile* tile = getTile();
+	if (!tile) {
+		return;
+	}
+
+	const MagicField* field = tile->getFieldItem();
 	if (!field) {
 		return;
 	}
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -931,7 +931,8 @@ void Creature::onEndCondition(ConditionType_t)
 
 void Creature::onTickCondition(ConditionType_t type, bool& bRemove)
 {
-	const MagicField* field = getTile()->getFieldItem();
+	const auto &tile = getTile();
+	const auto &field = tile ? tile->getFieldItem() : nullptr;
 	if (!field) {
 		return;
 	}


src/creature.cpp
```diff
@@ -931,7 +931,8 @@ void Creature::onEndCondition(ConditionType_t)
 
 void Creature::onTickCondition(ConditionType_t type, bool& bRemove)
 {
-	const MagicField* field = getTile()->getFieldItem();
+	const auto& tile = getTile();
```
ArturKnopik: maybe 
```cpp
const auto* tile = getTile();
const auto* field = tile ? tile->getFieldItem() : nullptr;
```

ghost: better now

ramon-bernardo:
Every creature in the game is assigned to a tile by default. If no tile exists, is destroyed.

All creatures go through [Map::placeCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L186) and then proceed to [Cylinder::internalAddThing](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L243-L244).

Let’s look at a real case. Imagine a monster is created via `Lua`, where an admin uses `/monster demon`. The following code execution flow occurs:
- [ LuaScriptInterface::luaGameCreateMonster ](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/luascript.cpp#L4977-L5003)
- [Game::placeCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/game.cpp#L518-L542) 
- [Game::internalPlaceCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/game.cpp#L501-L516) 
 - [Map::placeCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L186-L249) 
- [Cylinder::internalAddThing](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L243-L244) 
- [Tile::internalAddThing](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/tile.cpp#L1425-L1480) 
- [thing->setParent(this);](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/tile.cpp#L1429)

These hidden rules can sometimes lead to confusion 😕

ghost:
> Every creature in the game is assigned to a tile by default. If no tile exists, is destroyed.
> 
> All creatures go through [Map::placeCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L186) and then proceed to [Cylinder::internalAddThing](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L243-L244).
> 
> Let’s look at a real case. Imagine a monster is created via `Lua`, where an admin uses `/monster demon`. The following code execution flow occurs:
> 
> * [ LuaScriptInterface::luaGameCreateMonster ](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/luascript.cpp#L4977-L5003)
> * [Game::placeCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/game.cpp#L518-L542)
> * [Game::internalPlaceCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/game.cpp#L501-L516)
> * [Map::placeCreature](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L186-L249)
> * [Cylinder::internalAddThing](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/map.cpp#L243-L244)
> * [Tile::internalAddThing](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/tile.cpp#L1425-L1480)
> * [thing->setParent(this);](https://github.com/otland/forgottenserver/blob/d498855e773b6d9e77d310ae36f75c4c95ca3155/src/tile.cpp#L1429)
> 
> These hidden rules can sometimes lead to confusion 😕

Nah, getTile() can be nullptr in a few cases:

When a creature is created but not yet placed on the map. (Map::placeCreature() + Tile::addThing())
After being removed from the map but still alive in memory. (Tile::removeCreature())
During isRemoved() states where the parent/tile is already cleared.
In async events or Lua scripts referencing creatures out of phase.
important to know about the presence of Tile::nullptr_tile also indicates the engine expects “no tile” situations...

Those checks make sense, but in the end it’s up to you guys, no problem either way. :)

ramon-bernardo:
@kaleohanopahala absolutely right here, I always forget about this removed. remember seeing the same situation in other places, Il take a look 😆

MillhioreBT:
It's impossible for onTickCondition to be called if the creature is not on the checkCreatureList. If you have a valid trick to reproduce the issue, I'll be happy to know.

ArturKnopik:
In the current state it might not be possible to have a nullptr, but that doesn’t change the fact that TFS requires customization to be usable. It’s better not to provide code that could blow up in someone’s face if they use it differently than the current implementation.
It's also true for https://github.com/otland/forgottenserver/pull/4999

PR #5003, opened 2025-10-04
fix possible null at player trade

kaleohanopahala:
Abort the trade to keep the core stable when the backing item no longer exists.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -2679,7 +2679,7 @@ void Game::playerRequestTrade(uint32_t playerId, const Position& pos, uint8_t st
 	if (tradeItem->isRemoved() || !tradeItem->getParent()) {
 		player->sendCancelMessage(RETURNVALUE_NOTPOSSIBLE);
 		return;
-	}	
+	}
 
 	if (getBoolean(ConfigManager::ONLY_INVITED_CAN_MOVE_HOUSE_ITEMS)) {
 		if (const HouseTile* const houseTile = dynamic_cast<const HouseTile*>(tradeItem->getTile())) {


ArturKnopik:
did you check all of this events(c++ site)? 
```XML
<event class="Player" method="onTradeRequest" enabled="0" />
<event class="Player" method="onTradeAccept" enabled="0" />
<event class="Player" method="onTradeCompleted" enabled="0" />
```
item on accept and complet also can be "removed"

MillhioreBT:
Honestly, I don't know in what scenario this could happen. The description is very poor and doesn't exactly make anything clear.

It's an extra, unnecessary check. If `internalGetThing` returns something, that something exists, and no other asynchronous call can occur between [that line](https://github.com/otland/forgottenserver/blob/54baa4366ae3cf0a151d5505e4f5e6943a4d47a3/src/game.cpp#L2666) and where you use the [IF](https://github.com/otland/forgottenserver/blob/54baa4366ae3cf0a151d5505e4f5e6943a4d47a3/src/game.cpp#L2679).

I might be wrong, right? Well, I'd love to know how to reproduce this faithfully if you have anything in mind. Or was it just a random error you saw in the debugger one day?

Are you sure you encountered this error using **TFS master** without any modifications to validate the issue?
You're also validating the same thing twice; `isRemoved` already validates that a parent doesn't exist or that the parent is removed.

ArturKnopik:
@MillhioreBT  please note that modern compilers are quite intelligent — if the compiler determines that this check is unnecessary, it will remove it. I’ll skip the fact that this code is relatively rarely used, so the cost of one additional if doesn’t make any difference in performance anyway...

MillhioreBT:
> @MillhioreBT please note that modern compilers are quite intelligent — if the compiler determines that this check is unnecessary, it will remove it. I’ll skip the fact that this code is relatively rarely used, so the cost of one additional if doesn’t make any difference in performance anyway...

That's not true, compilers are not yet smart enough to know that, remember that compilers don't execute all the game logic ahead of time to determine things that might occur at runtime.
But if you want to teach me something new, I accept the explanation, tell me how the magic compiler can know if the IF is unnecessary or if it's necessary? How can it know if parent or if parent->isRemoved or if internalGetThing will return a nullptr or not a nullptr at that moment?

We also don't have to become so paranoid about this topic, if everything is well written we don't have to be placing ifs of this style everywhere.

In summary, I have no problem if this code fragment is merged, but at least change this:
```cpp
if (tradeItem->isRemoved() || !tradeItem->getParent()) {
	player->sendCancelMessage(RETURNVALUE_NOTPOSSIBLE);
	return;
}
```
it's the same as:
```cpp
if (tradeItem->isRemoved()) {
	player->sendCancelMessage(RETURNVALUE_NOTPOSSIBLE);
	return;
}
```

It doesn't make sense either, how can internalGetThing return something that doesn't have a parent if we need a parent to be able to get the item lol. All items that exist and aren't removed have a parent and the parent exists.

It simply can't happen unless there's some race condition corrupting memory, which could lead to any crash anywhere without any predictable order.

ranisalt:
I too am curious how to reproduce this issue. Like @MillhioreBT pointed out this could reveal some other race conditions that we need to think about, and possibly test against.

ghost:
> Honestly, I don't know in what scenario this could happen. The description is very poor and doesn't exactly make anything clear.
> 
> It's an extra, unnecessary check. If `internalGetThing` returns something, that something exists, and no other asynchronous call can occur between [that line](https://github.com/otland/forgottenserver/blob/54baa4366ae3cf0a151d5505e4f5e6943a4d47a3/src/game.cpp#L2666) and where you use the [IF](https://github.com/otland/forgottenserver/blob/54baa4366ae3cf0a151d5505e4f5e6943a4d47a3/src/game.cpp#L2679).
> 
> I might be wrong, right? Well, I'd love to know how to reproduce this faithfully if you have anything in mind. Or was it just a random error you saw in the debugger one day?
> 
> Are you sure you encountered this error using **TFS master** without any modifications to validate the issue? You're also validating the same thing twice; `isRemoved` already validates that a parent doesn't exist or that the parent is removed.

its true, just a false positive.

PR #5008, opened 2025-10-07
fix(houses): clear ownerName after house sale

schunski:
### Pull Request Prelude

### Changes Proposed

This pull request fixes an issue where the previous owner's name remained visible on a house after it was sold.

The problem occurred because ownerName was not being cleared when resetting ownership data.
To resolve this, a call to ownerName.clear() was added right after resetting owner and ownerAccountId.

### Issues addressed:

Fixes incorrect display of previous owner name after house sale.
Prevents confusion when houses are transferred or sold to a new player.

### DIFF
--- a/src/house.cpp
+++ b/src/house.cpp
@@ -64,7 +64,7 @@ void House::setOwner(uint32_t guid, bool updateDatabase /* = true*/, Player* pla
 		// clean access lists
 		owner = 0;
 		ownerAccountId = 0;
-		ownerName.clear()
+		ownerName.clear();
 		setAccessList(SUBOWNER_LIST, "");
 		setAccessList(GUEST_LIST, "");
 


schunski:
Two videos have been included:

1. Shows the issue without using **ownerName.clear()** — the previous owner's name remains visible until the server is restarted.

https://github.com/user-attachments/assets/b1c5c079-5f0d-4813-9479-d3b9ce87f288

2. Shows the behavior with **ownerName.clear()** — the previous owner's name is immediately cleared, even while the server is still running.

https://github.com/user-attachments/assets/0ac85bd2-d354-45b2-a748-51e0e9503592


Previously, the owner's name was reset only if the server was restarted. Without **ownerName.clear()**, the name remained visible during continuous server uptime.

With this fix implemented, there is no need to restart the server for the owner's name to be correctly cleared.

schunski:
Added missing semicolon.

PR #5010, opened 2025-10-08
Prevent crash on onAttackedCreatureChangeZone

kaleohanopahala:
Avoids crashes when getAttackedCreature() ends up null — for example, if there’s no target or it gets removed while changing zones...

_**onCreatureDisappear ~ attackedCreature ~ removeAttackedCreature ~ setAttackedCreature ~ Creature::internalCreatureDisappear**_

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1217,7 +1217,7 @@ void Player::onChangeZone(ZoneType_t zone)
 
 void Player::onAttackedCreatureChangeZone(ZoneType_t zone)
 {
-	const auto& attackedCreature = getAttackedCreature();
+	const auto attackedCreature = getAttackedCreature();
 	if (!attackedCreature) {
 		return;
 	}


ArturKnopik:
Why you assigning pointer as reference??

```cpp
Creature* attackedCreature = getAttackedCreature();
if (!attackedCreature) {
    return;
}
```
#edit
https://chatgpt.com/share/68e5de56-03b0-8010-af97-ed73298d8d9f

MillhioreBT:
This makes no sense.

`onAttackedCreatureChangeZone` is only called here: [https://github.com/otland/forgottenserver/blob/f4f8edbc07fe10e3852b2518e6809fa5f470e6ec/src/creature.cpp#L439](https://github.com/otland/forgottenserver/blob/f4f8edbc07fe10e3852b2518e6809fa5f470e6ec/src/creature.cpp#L439)

And as you can see, the pointer: `attackedCreature` is used to get the zone.
This is because at this point `attackedCreature` will always be a valid pointer.

It makes no sense to check inside the function if `attackedCreature` is `nullptr`, there is no way it can change.

Please don't assume that the problems that occur in Canary also occur here, TFS has its own problems, and Canary has its own as well.

It is much appreciated that you want to help TFS, but please try to verify the problem here first and describe the steps to follow or what you were doing when this error occurred and if possible also show the call stack or the crash LOG.

**Otherwise, refrain from making these PRs, because most of them are just copy-pastes of your PRs in the Canary repository. I went to check to confirm what I'm saying.**

---------------------------------------
If you think I'm wrong, please let me know and explain, because I want to learn too.

Kawaipunahele:
> This makes no sense.
> 
> `onAttackedCreatureChangeZone` is only called here:
> 
> https://github.com/otland/forgottenserver/blob/f4f8edbc07fe10e3852b2518e6809fa5f470e6ec/src/creature.cpp#L439
> 
> And as you can see, the pointer: `attackedCreature` is used to get the zone. This is because at this point `attackedCreature` will always be a valid pointer.
> 
> It makes no sense to check inside the function if `attackedCreature` is `nullptr`, there is no way it can change.
> 
> Please don't assume that the problems that occur in Canary also occur here, TFS has its own problems, and Canary has its own as well.
> 
> It is much appreciated that you want to help TFS, but please try to verify the problem here first and describe the steps to follow or what you were doing when this error occurred and if possible also show the call stack or the crash LOG.
> 
> **Otherwise, refrain from making these PRs, because most of them are just copy-pastes of your PRs in the Canary repository. I went to check to confirm what I'm saying.**
> 
> If you think I'm wrong, please let me know and explain, because I want to learn too.

We have to copy what works, right? Not this horrible thing full of bugs here lol.

kasajames:
> This makes no sense.
> 
> `onAttackedCreatureChangeZone` is only called here:
> 
> https://github.com/otland/forgottenserver/blob/f4f8edbc07fe10e3852b2518e6809fa5f470e6ec/src/creature.cpp#L439
> 
> And as you can see, the pointer: `attackedCreature` is used to get the zone. This is because at this point `attackedCreature` will always be a valid pointer.
> 
> It makes no sense to check inside the function if `attackedCreature` is `nullptr`, there is no way it can change.
> 
> Please don't assume that the problems that occur in Canary also occur here, TFS has its own problems, and Canary has its own as well.
> 
> It is much appreciated that you want to help TFS, but please try to verify the problem here first and describe the steps to follow or what you were doing when this error occurred and if possible also show the call stack or the crash LOG.
> 
> **Otherwise, refrain from making these PRs, because most of them are just copy-pastes of your PRs in the Canary repository. I went to check to confirm what I'm saying.**
> 
> If you think I'm wrong, please let me know and explain, because I want to learn too.

You have to copy what works, right? Not this horrible thing full of bugs here lol.


I've looked at your work, and there's nothing special about you. _Calm down a bit_, be less arrogant, because it's not even the **least bit.**

ArturKnopik:
@kaleohanopahala I suggest we just drop the whole PR thing.
The current approach makes no sense... performance is being prioritized over safety — sure, it works now (without any thought about potential future refactors).
The cost of those checks in all the methods you tried to “fix” was negligible, but fine... let it be as is.
I’ll skip the fact that setting Lua 5.4 as the default broke half of the project — that was apparently fine...
This is my last message in this repo. I’m cutting myself off from this mess because it’s pointless to fight over stupid if checks — I don’t even want to know what will happen with bigger changes.
I had planned — replacing raw pointers with smart pointers — it would probably never pass review.
Personally, I’ve already left the TFS dev team, I have no power anymore to merga any code.

lgrossi:
> This makes no sense.
> 
> `onAttackedCreatureChangeZone` is only called here: [https://github.com/otland/forgottenserver/blob/f4f8edbc07fe10e3852b2518e6809fa5f470e6ec/src/creature.cpp#L439](https://github.com/otland/forgottenserver/blob/f4f8edbc07fe10e3852b2518e6809fa5f470e6ec/src/creature.cpp#L439)
> 
> And as you can see, the pointer: `attackedCreature` is used to get the zone.
> This is because at this point `attackedCreature` will always be a valid pointer.
> 
> It makes no sense to check inside the function if `attackedCreature` is `nullptr`, there is no way it can change.
> 
> Please don't assume that the problems that occur in Canary also occur here, TFS has its own problems, and Canary has its own as well.
> 
> It is much appreciated that you want to help TFS, but please try to verify the problem here first and describe the steps to follow or what you were doing when this error occurred and if possible also show the call stack or the crash LOG.
> 
> **Otherwise, refrain from making these PRs, because most of them are just copy-pastes of your PRs in the Canary repository. I went to check to confirm what I'm saying.**
> 
> ---------------------------------------
> If you think I'm wrong, please let me know and explain, because I want to learn too.

Last time I checked, TFS was still an open-source project. And since this function isn’t marked as private or documented as internal, it naturally looks like part of the public contract of the codebase. 

From an outside perspective, that means callers are free to pass nullptr, and the responsibility to handle that safely stays with the callee.

So when I see the argument “we don’t use it that way internally,” I can’t help but wonder: is that really a reliable assumption for an open-source API? 😊

(Just to be clear, I’m talking about the code API, meaning the set of public C++ methods exposed by the *open-source* engine itself)

If the real intent is that `onAttackedCreatureChangeZone` should never receive a null target, maybe the contract itself should make that clear. For example:

```cpp
void Player::onAttackedCreatureChangeZone(const Creature& target, Zone newZone); 
```

That would send a clear signal to anyone using the API: “we always expect a valid target, you check it before calling.”

But as long as the current signature accepts a raw pointer, the expectation is the opposite, that the function can safely handle null. In that light, the null-check in this PR doesn’t feel like a workaround; it feels like fulfilling the contract that’s currently advertised.

ghost:
> This makes no sense.
> 
(...)

You’re right, I ran the tests again and couldn’t reproduce the crash that was reported to me and that led to my fix.

@kasajames   Canary is indeed ahead because the community is more active, but the base is the same? Maybe.

PR #5011, opened 2025-10-09
Revert "Add Followers Reference Counter (#4968)"

MillhioreBT:
This reverts commit 30da1554b46b6b91694a3d0cc2b308408a36edd9.

### Pull Request Prelude

- [x] I have followed [proper The Forgotten Server code styling][code].
- [x] I have read and understood the [contribution guidelines][cont] before making this PR.
- [x] I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
Currently, **TFS** is unusable; anyone who downloads it won't be able to use it because it crashes very quickly in almost all situations involving monsters.

It all started here: [30da1554b46b6b91694a3d0cc2b308408a36edd9] was a response to -> [https://github.com/otland/forgottenserver/issues/4961](https://github.com/otland/forgottenserver/issues/4961)

however: #4961 was never verified, it was a supposition of a random failure that is not known if it was tested on **TFS master** without changes.

I have already spent 1 hour juggling scripts and even deliberately manipulating the course of execution in debugging mode and I cannot reproduce said problem, at least not in the time I have been testing.

PR #5012, opened 2025-10-14
fix(creature): remove releaseFollowers call from destructor

MillhioreBT:
### Pull Request Prelude

### Changes Proposed
I was testing and simply deleting this call here, everything works normally
all references are correctly removed

If enemies are killed while they still remain in the followers vector, they'll simply be removed when the player logs out/changes floor or anything that calls Game:removeCreature

**Issues addressed:** #5011

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -37,8 +37,6 @@ Creature::~Creature()
 	for (auto condition : conditions) {
 		delete condition;
 	}
-
-	releaseFollowers();
 }
 
 bool Creature::canSee(const Position& myPos, const Position& pos, int32_t viewRangeX, int32_t viewRangeY)


alysonjacomin:
@NRH-AA  @gesior  It would be good to perform new benchmarks now with this PR to ensure security and check if performance remains good

Codinablack:
LGTM

gesior:
@MillhioreBT @ranisalt 
I did not test newest TFS crashes, but it would be better to debug why it crashes and fix crash reason. Probably `releaseFollowers` was called twice in some case.
It crashes, because it decrements refCount, but does not `clean()` `followers` list in Creature, so next call to `releaseFollowers` calls again 'decrement refCount' on all followers, so followers refCount is decremented more than once.

`followers.clear();` at end of `void Creature::releaseFollowers()` should fix it in this case and any other case, when `releaseFollowers` is called more than once for any reason.

@alysonjacomin This change should not affect ~Creature execution time (measurable time). When creature is killed/player is killed/player logouts, there are called many other functions that take much more CPU (drop loot, save player etc.).

ranisalt:
@gesior with #5022 this becomes a non-issue as references will always be correctly counted

PR #5014, opened 2025-10-19
fix: door removal reference counting

kaleohanopahala:
The set iterator was being dereferenced after erase, which seems risky.
Now I store the pointer before erasing the iterator from the set.

Is that the right  ?

### DIFF
--- a/src/house.cpp
+++ b/src/house.cpp
@@ -267,8 +267,9 @@ void House::removeDoor(Door* door)
 {
 	auto it = doorSet.find(door);
 	if (it != doorSet.end()) {
-		door->decrementReferenceCounter();
+		Door* doorToRemove = *it;
 		doorSet.erase(it);
+		doorToRemove->decrementReferenceCounter();
 	}
 }
 


MillhioreBT:
> @MillhioreBT why invalid? make sense. @ranisalt someone who knows about can check? i'll drop all code on gemini**PRO** and ill back. @kaleohanopahala

I marked it as invalid because what you're doing is simply copying the container pointer into the iterator, it points to the same memory address as the door variable, so they are just 2 lines of code that do exactly the same thing.

Codinablack:
> @MillhioreBT why invalid? make sense. @ranisalt someone who knows about can check? i'll drop all code on gemini**PRO** and ill back. @kaleohanopahala

Sure the ai will tell you there is a problem being fixed if you suggest there is a problem being fixed... ai is simply an application which generates based on input.

There is absolutely no point in creating a copy right there... like zero. This "fix" makes zero sense, and doesn't do anything... you are switching a reference for a pointer and they are pointing to the exact same thing... not even... I just looked at it again, you are switching a pointer for a pointer not even a ref for a pointer XDDDDD

kasajames:
> The set iterator was being dereferenced after erase, which seems risky. Now I store the pointer before erasing the iterator from the set.
> 
> Is that the right ?



i'll drop all code on gemini**ULTRA** and **gptPRO** ill back. @kaleohanopahala.

image
image

@kaleohanopahala let’s stay on our private repo. It’s active, and we don’t have ego battles or people insisting on insecure practices.

Codinablack:
What did I tell you

> > The set iterator was being dereferenced after erase, which seems risky. Now I store the pointer before erasing the iterator from the set.
> > Is that the right ?
> 
> i'll drop all code on gemini**ULTRA** and **gptPRO** ill back. @kaleohanopahala.
> image image
> 
> @kaleohanopahala let’s stay on our private repo. It’s active, and we don’t have ego battles or people insisting on insecure practices.

image


If you suggest it, ai will accept it
ai is twice as biased as you are

kasajames:
@kaleohanopahala The Brazilian version is way ahead of this; probably best to abandon it.

Codinablack:
> @kaleohanopahala The Brazilian version is way ahead of this; probably best to abandon it.

Here, I will show you how much your ai knows

image

So you keep citing AI as a source like it means something and ignore the ones who know what they are talking about when they try to tell you something... see how far that takes you.

kasajames:
You’ve got an anime profile pic, and everything you’ve done is irrelevant same as the other person up there. If it’s not your clique, you go after it. I know the reason. I’m heading out...
I need to adjust something on the largest active Brazilian OT.

ranisalt:
You don't have to act like children. This patch does not change anything because even if the pointed-to object (the `Door`) is destroyed and freed, `doorSet.erase` will get the _address_, never dereference it, and just remove from a set. For all it knows, it's just another number in the set.

If you want to be serious about use-after-free fixes, [try the address sanitizer](https://www.osc.edu/resources/getting_started/howto/howto_use_address_sanitizer). I suggest you go and check the compiler output and you will see it emits the exact same code - much better than trying to get an LLM to agree with you.

PR #5016, opened 2025-10-22
refactor(general): replace conditions raw pointers with smart pointers

ramon-bernardo:
### Pull Request Prelude
- [x I am aware that this PR may be closed if the above-mentioned criteria are not fulfilled.

### Changes Proposed
- This refactor replaces all usages of raw pointers for Condition instances with std::shared_ptr.
- Improve memory safety, simplify ownership semantics and prevent potential memory leaks or dangling pointers when handling condition objects.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -720,7 +720,7 @@ void Monster::updateIdleStatus()
 	bool idle = false;
 	if (!isSummon() && targetList.empty()) {
 		// check if there are aggressive conditions
-		idle = std::any_of(conditions.begin(), conditions.end(),
+		idle = std::none_of(conditions.begin(), conditions.end(),
 		                   [](const std::shared_ptr<Condition>& condition) { return condition->isAggressive(); });
 	}
 


alysonjacomin:
Very good. I'm thinking of doing this for all Thing subclasses (Creature, Cylinder and Item).

MillhioreBT:
I don't want to be a spoilsport, but this PR is a duplicate; it's a copy-paste of a previous PR I made.

Why should yours be merged? Especially since mine hasn't even been reviewed by anyone, and if I recall correctly, you were complaining about the way I wrote it XD. Seriously, you have no respect for others.

If you had at least mentioned something like: 'So-and-so didn't want to continue the work, so I'll take it over,' etc., maybe then I would gladly approve this PR just for the humility. But you lack humility.

However, I will leave this as is in case another DEV prefers your PR. Good luck, and I hope you continue wanting to dedicate your valuable time to TFS. 
image

ramon-bernardo:
@MillhioreBT Hey, no worries. its not the same PR, even though the idea might look similar at first glance. The main difference is that this one is part of the ongoing rewrite from `raw pointers` to `shared_ptr`, not just a copy of the previous impl.

Also, what I commented before was about the use of `aliases` and the `cognitive overhead` from type overloading, not the idea itself.

No need to overreact. Its all part of improve codebase and making it more consistent moving forward. Shared for all!

ramon-bernardo:
I need to review a few things, so please dont merge it yet ;)

src/combat.h
```diff
@@ -41,7 +41,7 @@ class TargetCallback final : public CallBack
 
 struct CombatParams
 {
-	std::forward_list<std::unique_ptr<const Condition>> conditionList = {};
+	std::forward_list<std::shared_ptr<const Condition>> conditions = {};
```
ranisalt: Let's change this one to a vector (preferred) or a deque, forward list is wasteful

src/creature.cpp
```diff
@@ -1516,7 +1506,7 @@ bool FrozenPathingConditionCall::operator()(const Position& startPos, const Posi
 
 bool Creature::isInvisible() const
 {
-	return std::find_if(conditions.begin(), conditions.end(), [](const Condition* condition) {
+	return std::find_if(conditions.begin(), conditions.end(), [](const std::shared_ptr<Condition>& condition) {
 		       return condition->getType() == CONDITION_INVISIBLE;
 	       }) != conditions.end();
```
ranisalt:
```suggestion
	return std::any_of(conditions.begin(), conditions.end(), [](const std::shared_ptr<Condition>& condition) {
		return condition->getType() == CONDITION_INVISIBLE;
	});
```

or


```suggestion
	return std::ranges::any_of(conditions, [](const std::shared_ptr<Condition>& condition) {
		return condition->getType() == CONDITION_INVISIBLE;
	});
```

if you're feeling adventurous

src/luascript.cpp
```diff
@@ -14131,18 +14130,17 @@ int LuaScriptInterface::luaConditionCreate(lua_State* L)
 int LuaScriptInterface::luaConditionDelete(lua_State* L)
 {
 	// condition:delete()
-	Condition** conditionPtr = tfs::lua::getRawUserdata<Condition>(L, 1);
-	if (conditionPtr && *conditionPtr) {
-		delete *conditionPtr;
-		*conditionPtr = nullptr;
+	auto& condition = getSharedPtr<Condition>(L, 1);
+	if (condition) {
+		condition.reset();
```
ranisalt, line 14135: This is gonna leak, it's resetting a copy. You probably need to pop from the Lua stack.

src/monster.cpp
```diff
@@ -720,8 +720,9 @@ void Monster::updateIdleStatus()
 	bool idle = false;
 	if (!isSummon() && targetList.empty()) {
 		// check if there are aggressive conditions
-		idle = std::find_if(conditions.begin(), conditions.end(),
-		                    [](Condition* condition) { return condition->isAggressive(); }) == conditions.end();
+		idle = std::find_if(conditions.begin(), conditions.end(), [](const std::shared_ptr<Condition>& condition) {
+			       return condition->isAggressive();
+		       }) == conditions.end();
```
ranisalt: Same as before, but using `none_of`

src/luascript.cpp
MillhioreBT: That's correct; it's **resetting a copy**, but the copy is being obtained **by reference, not by value**.
Therefore, the object that is being reset is **actually the one living in Lua**.

The **`getSharedPtr`** method already takes care of that.

src/combat.h
ramon-bernardo: Right, we also have places like `Player::getMuteConditions()` and `Player {storedConditionList}`. Should we change those as well?

src/creature.cpp
ramon-bernardo: Ready!

src/luascript.cpp
ramon-bernardo: @MillhioreBT is right, the pointer is owned by `Lua`, just like in `Combat` ;)

src/monster.cpp
ramon-bernardo: Ready! A moment, i had used `std::any_of`, but i reverted it

PR #5017, opened 2025-10-24
Repair "player:getInstantSpells()" to get "userdata spell" instead "comum table"

alysonjacomin:
### Pull Request Prelude

### Changes Proposed

Rework LUA function "player:getInstantSpells()"

### DIFF
--- a/data/lib/compat/compat.lua
+++ b/data/lib/compat/compat.lua
@@ -1636,6 +1636,33 @@ function getFormattedWorldTime()
 	return Game.getFormattedWorldTime()
 end
 
+function getPlayerInstantSpells(cid)
+	local result = {}
+
+	if not player then
+		return result
+	end
+
+	local spells = player:getInstantSpells()
+	if not spells then
+		return result
+	end
+
+	for i, spell in ipairs(spells) do
+		result[i] = {
+			name = spell:name(),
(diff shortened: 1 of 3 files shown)


MillhioreBT:
Now it only remains for you to fix this in this same PR.
https://github.com/otland/forgottenserver/blob/master/data/scripts/actions/others/spellbook.lua

Since we no longer have the fields available in a table, we must use the methods of the **Spell metatable**.

Or, if you prefer, you can add a **custom `__index` metamethod to the metatable** to simulate access to these fields. This would make it feel more **retrocompatible** with possible scripts that use these fields.

However, you must take into account that the **`spellbook.lua`** script also makes use of the **`__newindex` metamethod**, but **userdata cannot have fields associated directly**.

My complete recommendation would be to have an **`__index` metamethod for retrocompatibility** and also **update the `spellbook.lua` script**.

alysonjacomin:
> Now it only remains for you to fix this in this same PR. https://github.com/otland/forgottenserver/blob/master/data/scripts/actions/others/spellbook.lua
> 

I didn't understand what you meant by this PR ^

I added the compatibility function and reworked spellbook.lua

alysonjacomin:
@MillhioreBT  Is there still something missing to change in the code?

PR #5020, opened 2025-10-28
Add spdlog

javiertringol:
> The images are old, I’m too lazy to make them again.

-----------------------------

image
## terminal
image
## file    .log
image
image

test in windows , vs solution

I have a code in python that modifies all `std::cout << xxxx << std::endl;`  to `g_logger().info(xxxx)`

but because it would make the review more difficult, I will do it later.

### DIFF
--- a/src/configmanager.cpp
+++ b/src/configmanager.cpp
@@ -180,7 +180,7 @@ bool ConfigManager::load()
 	}
 
 	if (luaL_dofile(L, string[CONFIG_FILE].data())) {
-		g_logger().error("{}", lua_tostring(L, -1));
+		getLogger().error("{}", lua_tostring(L, -1));
 		lua_close(L);
 		return false;
 	}
--- a/src/databasemanager.cpp
+++ b/src/databasemanager.cpp
@@ -93,7 +93,7 @@ void DatabaseManager::updateDatabase()
 	int32_t version = getDatabaseVersion();
 	do {
 		if (luaL_dofile(L, fmt::format("data/migrations/{:d}.lua", version).c_str()) != 0) {
-			g_logger().error("[Error - DatabaseManager::updateDatabase - Version: {} {}", version, lua_tostring(L, -1));
+			getLogger().info("[Error - DatabaseManager::updateDatabase - Version: {} {}", version, lua_tostring(L, -1));
 
 			break;
 		}
@@ -105,7 +105,7 @@ void DatabaseManager::updateDatabase()
 		lua_getglobal(L, "onUpdateDatabase");
 		if (lua_pcall(L, 0, 1, 0) != 0) {
 			tfs::lua::resetScriptEnv();
-			g_logger().error("[Error - DatabaseManager::updateDatabase - Version: {} {}", version, lua_tostring(L, -1));
+			getLogger().info("[Error - DatabaseManager::updateDatabase - Version: {} {}", version, lua_tostring(L, -1));
 
 			break;
 		}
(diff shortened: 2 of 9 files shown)


src/logger.h
```diff
@@ -0,0 +1,217 @@
...
+	virtual LogLevel getLevel() const = 0;
+	virtual bool isEnabled(LogLevel level) const = 0;
+
+	void trace([[maybe_unused]] std::string_view msg)
+	{
+#if !defined(NDEBUG) || defined(DEBUG_LOG)
+		if (isEnabled(LogLevel::TRACE)) {
+			log(LogLevel::TRACE, msg);
+		}
+#endif
```
ranisalt: Nah, if there's a configuration entry for this, it should not be controlled by compile flags

src/logger.h
```diff
@@ -0,0 +1,217 @@
...
+			      std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count());
+			throw;
+		}
+	}
+
+protected:
+	virtual void log(LogLevel level, std::string_view message) = 0;
+};
+
+Logger& g_logger();
```
ranisalt: If it's a function, call it `getLogger`. `g_` is reserved for global _variables_ (🤮)

src/logger.cpp
```diff
@@ -0,0 +1,311 @@
...
+		case spdlog::level::warn:
+			return LogLevel::WARNING;
+		case spdlog::level::err:
+			return LogLevel::ERRORR;
+		case spdlog::level::critical:
+			return LogLevel::CRITICAL;
+		default:
+			return LogLevel::INFO;
+		}
+	}
```
ranisalt, line 60: Why not directly use the `spdlog` enum?

src/logger.cpp
```diff
@@ -0,0 +1,311 @@
...
+		auto now = std::chrono::system_clock::now();
+		auto time_t = std::chrono::system_clock::to_time_t(now);
+
+		std::stringstream ss;
+		ss << std::put_time(std::localtime(&time_t), "%Y%m%d_%H%M%S");
+
+		std::filesystem::path path(basePath);
+		std::string directory = path.parent_path().string();
+		std::string baseName = path.stem().string();
+		std::string extension = path.extension().string();
```
ranisalt: None of these need to be converted to `string`

src/logger.cpp
```diff
@@ -0,0 +1,311 @@
...
+					spdlog::register_logger(logger_);
+
+					logger_->info("=== TFS Logger Initialized ===");
+					logger_->info("Log file: {}", timestampedPath_);
+					logger_->flush();
+
+				}
+				catch (const std::exception& e) {
+					fprintf(stderr, "Error creating logger: %s\n", e.what());
+					throw;
```
ranisalt: Rethrow if you won't handle, this will swallow the stack

src/logger.cpp
```diff
@@ -0,0 +1,311 @@
...
+			return space.available > minSpaceBytes;
+		}
+		catch (const std::filesystem::filesystem_error&) {
+			return true;
+		}
+	}
+
+	class LogWithSpdLog final : public Logger {
+		public:
+			LogWithSpdLog(std::string_view filePath, size_t rotateSize, size_t rotateFiles) {
```
ranisalt: This belongs to a free function `createLogger` that does all the preparation and then creates a new `LogWithSpdlog` when everything is checked

src/logger.h
ranisalt, line 1: Could we use [`std::format`](https://en.cppreference.com/w/cpp/utility/format/format.html) facilities over `fmt::format`?

ranisalt:
Logging is a great addition, but do we need to wrap it? We could use spdlog directly, the API seems quite clean and easy

PR #5046, opened 2025-11-15
feat(talkactions): add Lua-based loader for XML talkactions, convert existing scripts

ranisalt:
### Pull Request Prelude

### Changes Proposed

Migrates talkactions to use the `TalkAction` Lua interface

Converted talkactions with this Python script and a few manual adjustments:
<details>

```python
import xml.etree.ElementTree as ET
from os import makedirs, path, unlink

tree = ET.parse("talkactions.xml")
root = tree.getroot()
assert root.tag == "talkactions"

for child in root:
    assert child.tag in [
        "talkaction",
    ], f"Invalid tag '{child.tag}', must be 'talkaction'"

    if "script" not in child.attrib:
        continue

    words = []
    lines = []
    script = ""

    for attrib in child.attrib:
        if attrib.lower() == "script":
            script = child.attrib[attrib]

        elif attrib.lower() == "words":
            words = child.attrib[attrib].split(";")

        elif attrib.lower() == "separator":
            lines.append(f'talkaction:separator("{child.attrib[attrib]}")')

        else:
            print(f"Unknown attribute: {attrib}")

    with open(f"scripts/{script}", "r") as f:
        talkaction = f.read().split("\n")

    for idx, line in enumerate(talkaction):
        if line.startswith("function onSay"):
            talkaction[idx] = talkaction[idx].replace("onSay", "talkaction.onSay")
            break
    else:
        print(f"talkaction {child.attrib['words']} does not have onSay function")
        continue

    talkaction.insert(0, f"local talkaction = TalkAction({', '.join(f'\"{word}\"' for word in words)})\n")
    talkaction.extend(lines)
    talkaction.append("talkaction:register()")

    print(f"Converting talkaction {child.attrib['words']}")

    makedirs(f"../scripts/talkactions/{path.dirname(script)}", exist_ok=True)
    with open(f"../scripts/talkactions/{script}", "w") as f:
        f.write("\n".join(talkaction))

    unlink(f"scripts/{script}")
```

</details>

### DIFF
--- a/src/talkaction.h
+++ b/src/talkaction.h
@@ -23,7 +23,7 @@ class TalkAction : public Event
 public:
 	explicit TalkAction(LuaScriptInterface* interface) : Event(interface) {}
 
-	bool configureEvent(const pugi::xml_node& node) override { return false; }
+	bool configureEvent(const pugi::xml_node&) override { return false; }
 
 	const std::string& getWords() const { return words; }
 	const std::vector<std::string>& getWordsMap() const { return wordsMap; }


kygov:
Last time I checked, commit https://github.com/otland/forgottenserver/commit/79138e837b0dcee71d3ca89a5d93ac9902f3c207 completely broke ability to reload actions because `bool configureEvent(const pugi::xml_node&) override { return false; }`

so it fails here https://github.com/otland/forgottenserver/blob/master/src/baseevents.cpp#L45

Looks like its not case here but you can check this.

ranisalt:
@kygov correct, I see a few calls to `g_actions->loadFromXml()` are still present which will print the error message. With the new code, reloading actions is done through reloading scripts instead, and `g_actions->loadFromXml()` should never be called.

There is no way of reloading a subset of scripts (so for example only reloading `data/scripts/actions`)

PR #5058, opened 2025-11-17
refactor(thing): remove Cylinder class and merge receiver logic into Thing

ramon-bernardo:
### Pull Request Prelude

### Changes proposed
Removes the obsolete Cylinder abstraction and merges all receiver and container-related logic directly into the base class Thing, effectively collapsing a multi-layer inheritance pyramid into a simpler and unified object hierarchy. This significantly reduces architectural complexity and eliminates redundant indirection that had accumulated over time.

The previous structure conceptually looked like this:
```
Thing
 └── Cylinder
      ├── Tile
      ├── Container + Item
      ├── Player + Creature
      └── Others...
```
This design introduced several issues:
- Thing provided only partial logic, while Cylinder held the actual receiver behaviors.
- Many systems expected a Thing but actually required a Cylinder, breaking abstraction boundaries.
- Different game systems (tile logic, container logic, player inventory logic) relied on different layers of the hierarchy.
- Core operations often required dynamic_cast or assumptions about whether an object was a Cylinder.

The additional layer made the hierarchy deeper than necessary, contributing to a “diamond-shaped” pattern when external systems interacted with both Thing and Cylinder.

#### Type confusion and forced casting
Many operations had to repeatedly check:
- “Is this a Thing?”
- “Is it also a Cylinder?”
- “If so, does it support queryAdd, queryRemove, etc.?”

This resulted in fragile code paths, excessive `RTTI` usage and higher maintenance overhead.

#### Logic fragmentation
Receiver and container logic was split between Thing and Cylinder, forcing developers to understand which class implemented which behavior. This fragmentation made extending or modifying interaction logic significantly more difficult.

#### Flattened hierarchy
After the refactor, the class hierarchy becomes:
```
Thing
 ├── Tile
 ├── Item
 │    └── Container
 ├── Creature
 │    └── Player
 └── Others...
```
The Cylinder layer has been removed entirely. All objects now share a consistent base class with a guaranteed receiver interface. 

All receiver-related operations (e.g., queryAdd, queryRemove, queryDestination, postAddNotification, postRemoveNotification, and indexing helpers) are now defined directly in Thing, ensuring they are always available without depending on intermediary classes.

More predictable class contracts, every subclass of Thing automatically inherits the same receiver interface. This gives fewer special cases, fewer casts and predictable behavior for all things.

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -313,21 +313,21 @@ ReturnValue Container::queryAdd(int32_t index, const Thing& thing, uint32_t coun
 		}
 	}
 
-	if (const auto topParent = getTopParent()) {
-		if (actor && getBoolean(ConfigManager::ONLY_INVITED_CAN_MOVE_HOUSE_ITEMS)) {
-			if (const auto tile = topParent->getTile()) {
-				if (const auto houseTile = tile->getHouseTile()) {
-					if (!topParent->getCreature() && !houseTile->getHouse()->isInvited(actor->getPlayer())) {
-						return RETURNVALUE_PLAYERISNOTINVITED;
-					}
+	if (actor && getBoolean(ConfigManager::ONLY_INVITED_CAN_MOVE_HOUSE_ITEMS)) {
+		const auto topParent = getTopParent();
+		if (const auto tile = topParent->getTile()) {
+			if (const auto houseTile = tile->getHouseTile()) {
+				if (!topParent->getCreature() && !houseTile->getHouse()->isInvited(actor->getPlayer())) {
+					return RETURNVALUE_PLAYERISNOTINVITED;
 				}
 			}
+		}
 
-			if (topParent != this) {
-				return topParent->queryAdd(INDEX_WHEREEVER, *item, count, flags | FLAG_CHILDISOWNER, actor);
-			}
+		if (topParent != this) {
+			return topParent->queryAdd(INDEX_WHEREEVER, *item, count, flags | FLAG_CHILDISOWNER, actor);
 		}
 	}
+
 	return RETURNVALUE_NOERROR;
 }
 
@@ -405,12 +405,11 @@ ReturnValue Container::queryRemove(const Thing& thing, uint32_t count, uint32_t 
 	}
 
 	if (actor && getBoolean(ConfigManager::ONLY_INVITED_CAN_MOVE_HOUSE_ITEMS)) {
-		if (const auto topParent = getTopParent()) {
-			if (const auto tile = topParent->getTile()) {
-				if (const auto houseTile = tile->getHouseTile()) {
-					if (!topParent->getCreature() && !houseTile->getHouse()->isInvited(actor->getPlayer())) {
-						return RETURNVALUE_PLAYERISNOTINVITED;
-					}
+		const auto topParent = getTopParent();
+		if (const auto tile = topParent->getTile()) {
+			if (const auto houseTile = tile->getHouseTile()) {
+				if (!topParent->getCreature() && !houseTile->getHouse()->isInvited(actor->getPlayer())) {
+					return RETURNVALUE_PLAYERISNOTINVITED;
 				}
 			}
 		}


src/thing.h
```diff
@@ -42,6 +65,178 @@ class Thing
 	virtual const Creature* getCreature() const { return nullptr; }
 
 	virtual bool isRemoved() const { return true; }
+
+	virtual Thing* getReceiver() { return nullptr; }
+	virtual const Thing* getReceiver() const { return nullptr; }
```
ranisalt, line 66: Note: besides getting rid of the class, this change is replacing the name "Cylinder" with the name "Receiver" as it clears the intent (it is an object that can have other objects added to it, or "receive" them)

This is an industry standard name, the other options being "sink" (not so obvious) or "container" (which we already use for something else)

src/player.cpp
```diff
@@ -2867,12 +2868,14 @@ Cylinder* Player::queryDestination(int32_t& index, const Thing& thing, Item** de
 		*destItem = destThing->getItem();
 	}
 
-	Cylinder* subCylinder = dynamic_cast<Cylinder*>(destThing);
-	if (subCylinder) {
-		index = INDEX_WHEREEVER;
-		*destItem = nullptr;
-		return subCylinder;
+	if (*destItem) {
```
ranisalt: This also solves a bug, no? I don't remember 😆

src/item.h
```diff
@@ -925,16 +924,16 @@
 	}
 
 	bool hasParent() const override { return getParent(); }
-	Cylinder* getParent() const override { return parent; }
-	void setParent(Cylinder* cylinder) override { parent = cylinder; }
-	Cylinder* getTopParent();
-	const Cylinder* getTopParent() const;
+	Thing* getParent() const override { return parent; }
+	void setParent(Thing* thing) override { parent = thing; }
```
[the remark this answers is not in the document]
ranisalt: Not relevant

src/item.cpp
```diff
@@ -253,62 +253,68 @@ void Item::setID(uint16_t newid)
...
+	auto parent = getTopParent();
+	if (!parent) {
+		return nullptr;
+	}
+
+	if (parent->hasParent()) {
+		parent = parent->getParent();
 	}
-	return dynamic_cast<const Tile*>(cylinder);
+	return parent->getTile();
```
ranisalt:
```suggestion
	auto topParent = getTopParent();
	if (!topParent) {
		return nullptr;
	}

	if (auto parent = parent->getParent()) {
		topParent = parent;
	}
	return topParent->getTile();
```

src/item.cpp
```diff
@@ -253,62 +253,60 @@
...
 
 Tile* Item::getTile()
 {
-	Cylinder* cylinder = getTopParent();
-	// get root cylinder
-	if (cylinder && cylinder->hasParent()) {
-		cylinder = cylinder->getParent();
+	auto parent = getTopParent();
+	if (parent->hasParent()) {
+		parent = parent->getParent();
```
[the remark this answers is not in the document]
ranisalt: Same as above/below

src/creature.h
```diff
@@ -344,11 +344,11 @@ class Creature : virtual public Thing
...
 	bool hasParent() const override { return getParent(); }
-	Cylinder* getParent() const override final { return tile; }
-	void setParent(Cylinder* cylinder) override final
+	Thing* getParent() const override final { return tile; }
+	void setParent(Thing* thing) override final
 	{
-		tile = static_cast<Tile*>(cylinder);
-		position = tile->getPosition();
+		tile = thing->getTile();
+		position = thing->getTile()->getPosition();
```
ranisalt:
```suggestion
		position = tile->getPosition();
```

src/game.cpp
```diff
@@ -980,13 +981,13 @@
 		return;
 	}
 
-	const Tile* toCylinderTile = toCylinder->getTile();
-	const Position& mapToPos = toCylinderTile->getPosition();
+	const Tile* toThingTile = toThing->getTile();
```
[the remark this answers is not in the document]
ranisalt: Can be ignored, we were not checking before either

src/container.cpp
```diff
@@ -284,54 +284,55 @@ ReturnValue Container::queryAdd(int32_t index, const Thing& thing, uint32_t coun
...
+				}
 			}
 		}
-	}
 
-	if (topParent != this) {
-		return topParent->queryAdd(INDEX_WHEREEVER, *item, count, flags | FLAG_CHILDISOWNER, actor);
+		if (topParent != this) {
+			return topParent->queryAdd(INDEX_WHEREEVER, *item, count, flags | FLAG_CHILDISOWNER, actor);
+		}
```
ranisalt, line 328: This changes behavior. Previously, it would check `topParent != this` even if `topParent == nullptr`, now it will not run.

`topParent` was used unchecked before

ramon-bernardo: Topparent will never be nullptr here, despite having added it in an `if state`
See `topParent->queryAdd`, it would throw a nullptr exception

src/item.cpp
ramon-bernardo: Ready.

src/player.cpp
ramon-bernardo: It's a bit confusin. I also questioned you about this once, but it seems correct.

Let's go through it.

When we look up the destination by index in the player, it searches within their inventory. So, for example, the item could be a backpack. If the destination item exists, we then check if it has a receiver, which would be a `Container`. It can't be anything other than a Container because, in a inventory, there can't be a Tile, TashHolder, Teleport or Mailbox.

This indicates that the location is anywhere within that container and also, the destination item would be nullptr if the container is present in the inventory.

src/container.cpp
```diff
@@ -284,54 +284,55 @@ ReturnValue Container::queryAdd(int32_t index, const Thing& thing, uint32_t coun
...
 		}
-		return ret;
-	}
 
-	if (!hasBitSet(FLAG_NOLIMIT, flags)) {
-		while (cylinder) {
-			if (cylinder == &thing) {
-				return RETURNVALUE_THISISIMPOSSIBLE;
-			}
+		if (hasBitSet(FLAG_NOLIMIT, flags)) {
```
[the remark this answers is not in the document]
ramon-bernardo: The logic was simply reversed

src/player.cpp
ramon-bernardo: I refactored it, much easier to read 😄

src/creature.h
```diff
@@ -344,10 +344,10 @@
 	bool unregisterCreatureEvent(const std::string& name);
 
 	bool hasParent() const override { return getParent(); }
-	Cylinder* getParent() const override final { return tile; }
-	void setParent(Cylinder* cylinder) override final
+	Thing* getParent() const override final { return tile; }
+	void setParent(Thing* thing) override final
 	{
-		tile = static_cast<Tile*>(cylinder);
+		tile = thing->getTile();
```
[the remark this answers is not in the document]
ramon-bernardo: A creature will NEVER lose its parent.

Compare `Tile::addThing` with `Tile::removeThing`

src/luascript.cpp
```diff
@@ -7159,59 +7140,57 @@ int LuaScriptInterface::luaItemRemoveCustomAttribute(lua_State* L)
...
+	// item:moveTo(position or thing[, flags])
 	Item** itemPtr = tfs::lua::getRawUserdata<Item>(L, 1);
 	if (!itemPtr) {
 		lua_pushnil(L);
 		return 1;
 	}
 
 	Item* item = *itemPtr;
-	if (!item || item->isRemoved()) {
+	if (!item) {
```
ramon-bernardo, line 7151: Necessary because `VirtualCylinder` no longer exists; everything now has a default Receiver behavior.

src/container.cpp
ranisalt: In that case, `const auto topParent = getTopParent()` should not be checked with an `if`

src/container.cpp
```diff
@@ -404,11 +405,12 @@ ReturnValue Container::queryRemove(const Thing& thing, uint32_t count, uint32_t
 	}
 
 	if (actor && getBoolean(ConfigManager::ONLY_INVITED_CAN_MOVE_HOUSE_ITEMS)) {
-		const Cylinder* const topParent = getTopParent();
-		if (const auto tile = topParent->getTile()) {
-			if (const auto houseTile = tile->getHouseTile()) {
-				if (!topParent->getCreature() && !houseTile->getHouse()->isInvited(actor->getPlayer())) {
-					return RETURNVALUE_PLAYERISNOTINVITED;
+		if (const auto topParent = getTopParent()) {
```
ranisalt: Same as before?

ramon-bernardo: Reverted 😄

src/player.cpp
```diff
@@ -2750,11 +2751,11 @@ ReturnValue Player::queryRemove(const Thing& thing, uint32_t count, uint32_t fla
 	return RETURNVALUE_NOERROR;
 }
 
-Cylinder* Player::queryDestination(int32_t& index, const Thing& thing, Item** destItem, uint32_t& flags)
+Thing* Player::queryDestination(int32_t& index, const Thing& thing, Item** destItem, uint32_t& flags)
 {
-	if (index == 0 /*drop to capacity window*/ || index == INDEX_WHEREEVER) {
-		*destItem = nullptr;
+	*destItem = nullptr;
```
ranisalt, line 2756: I think this can be dropped if we assume input will always be null.

This signature is 💩 and the `queryDestination` function should instead return a pair, but let's save it for later

ramon-bernardo: That's what i see too, but i intend to change on another PR

PR #5072, opened 2025-12-04
refactor: move static Combat methods out of header

ranisalt:
### Pull Request Prelude

### Changes Proposed

Moves static methods that are only used once out of the header file to avoid bloating the class namespace.


ranisalt:
> wouldn't it be possible to move to local namespace?

No, they are used elsewhere. Could be moved to a namespace but I would do it separately

PR #5083, opened 2026-04-09
Replace static constexpr with inline constexpr in headers (Fixes #4063)

### LINKED ISSUE #4063: Top-level constexprs in header files should be inline
When we declare a `static constexpr`, it means that the value will be present in each compilation unit i.e. each object generated with a cpp file. That means, if a `static constexpr` is declared in a header file, that every cpp file that includes that header (even indirectly) will have it's own copy of that value.

It can be seen that, if `static` is removed, the compiler will detect the name clash (which is avoided by `static`).

These top-level `constexpr`s present in header files should be `inline`, such that the declarations are merged when the executable is linked and only 1 copy will be present.

Tip: search using regex for `^static constexpr` in `src/*.h`

mouragst:
This PR replaces all 53 occurrences of `static constexpr` with `inline constexpr` in header files.

Fixes #4063

As described in #4063, when `static constexpr` is declared in a header file, every compilation unit (cpp file) that includes that header gets its own copy of that value. Using `inline constexpr` ensures the declarations are merged during linking and only one copy exists in the final executable, reducing binary size and avoiding potential ODR violations.

Replaced `static constexpr` with `inline constexpr` in the following files:
- connection.h (2 changes)
- const.h (5 changes)
- definitions.h (8 changes)
- creature.h (5 changes)
- game.h (14 changes)
- map.h (6 changes)
- monsters.h (2 changes)
- player.h (4 changes)
- party.h (2 changes)
- scheduler.h (1 change)

**Total:** 53 occurrences across 10 header files


DSpeichert:
## What the change actually does

At namespace scope, `constexpr` implies `const`, which implies **internal linkage**. So `static constexpr int32_t MAX_STACKPOS = 10;` in a header gives every translation unit its *own distinct object*. `inline constexpr` gives it **external linkage** with vague/COMDAT semantics — one entity, merged at link time. The `static` was redundant (it was already internal); the change flips linkage.

The values, types and names are untouched, so there's no behavioral change by construction.

## Practical difference — measured

I A/B'd it: built master, applied the PR, rebuilt, reverted.

**Release** (`build/Release/tfs`, -O2):

| | text | data | file bytes | defined symbols |
|---|---|---|---|---|
| master | 12,342,851 | 689,904 | 23,869,800 | 36,641 |
| PR 5083 | 12,342,851 | 689,904 | 23,869,800 | 36,641 |

Byte-identical binaries, identical symbol tables. **Zero difference in the shipped build.** These are integer/pointer constants used by value, so the optimizer folds them and never emits storage — there was nothing to deduplicate.

**Debug** (-O0) is where the duplication is real:

```
NETWORKMESSAGE_MAXSIZE   11 copies -> 0
STATUS_SERVER_NAME       12 copies -> 0
MAX_STACKPOS             10 copies -> 0
MINIMUM_SKILL_LEVEL      10 copies -> 0
SCHEDULER_MINTICKS        8 copies -> 1
...
```

Total: 76,907 -> 76,378 symbols; 71,240,288 -> 71,194,976 bytes (**-45 KB, 0.06%**). Real but negligible, and only in unoptimized builds.

So the "reducing binary size" motivation is effectively false for Release and rounding error for Debug.

## The part that *is* legitimate

The ODR argument holds, and it's the actual reason to take this. Several of these constants are odr-used from **inline functions and default initializers in headers**:

- `src/player.h:1340` — `getStepSpeed()` uses `PLAYER_MIN_SPEED`/`PLAYER_MAX_SPEED`
- `src/player.h:80` — `Skill::level = MINIMUM_SKILL_LEVEL` (default member init, runs in every TU's implicit ctor)
- `src/networkmessage.h:130,169` — inline members using `NETWORKMESSAGE_MAXSIZE`
- `src/tools.h:15` — default argument `= AUTHENTICATOR_DIGITS`

An inline function must have identical definitions across TUs; with internal-linkage constants each copy refers to a *different entity*, which is a formal [basic.def.odr] violation (IFNDR — no diagnostic required). It's benign here because the values are identical, but it's genuinely UB-on-paper and exactly the class of thing that bites under LTO or a future refactor. That's the correctness case for this PR, not size.

## Risks

Low, but not literally zero:

1. **Global-namespace symbols are now externally visible and unmangled.** Per the Itanium ABI a global-scope variable keeps its plain name — the Debug build emits `MAX_STACKPOS`, `FLOOR_SIZE`, `CHANNEL_GUILD` etc. as actual linker symbols. Names this generic in the global namespace could in principle collide with a linked library. I grepped for in-tree redefinitions and macro shadows (none) and both configs linked clean, so nothing is broken today; it's a slightly larger collision surface going forward. Putting these in a namespace would be the real fix, but that's a much bigger diff.
2. **GCC emits them as `STB_GNU_UNIQUE` (`u` in `nm`).** That makes a DSO containing them non-`dlclose`-able. Irrelevant for a standalone executable like `tfs`.
3. **No CI has run on this PR yet** (`statusCheckRollup` is empty), so MSVC/clang are unverified. `inline` variables are C++17 and the project is on C++23 (`CMakeLists.txt:138`), so I'd expect it to be fine, but it's unproven on the other toolchains.

## Verdict

Correct and safe, but mis-sold. It's a **standards-hygiene cleanup that fixes a latent ODR violation**, not a size or performance optimization — the Release binary is bit-for-bit unchanged. Worth merging on the ODR grounds; the description should be corrected so nobody cites it as a size win later.

Two follow-ups worth considering:

- **Inconsistency left behind.** The issue's premise applies equally to non-`constexpr` header constants this PR skipped: `src/monsters.h:13` (`const uint32_t MAX_LOOTCHANCE`) sits two lines from a changed line, plus `src/tasks.h:10-11` and the three fluid-map arrays in `src/const.h:354,359,380`. (Those arrays are the only ones with real per-TU storage cost — though in practice only 3 TUs use them, so one copy each in Release.)
- **`inline` is redundant on `src/tools.h:29`'s `constexpr` function** — functions are implicitly inline when `constexpr`; only variables needed the fix. This PR correctly leaves it alone.

<sub>Measurements: GCC, Ninja Multi-Config, `size`/`nm -C` on `build/{Release,Debug}/tfs`, same tree before and after applying the patch.</sub>

DSpeichert (approved):

PR #5089, opened 2026-08-14
ci: update 1.4 workflows, build with standalone Asio and C++23

DSpeichert:
Every workflow on this branch pins actions and runner images that have since been retired, so none of them can run as-is. This brings them up to currently supported versions.

Companion to #5088, which does the same for branch 1.6.

### Action version bumps

| Action | Before | After |
| --- | --- | --- |
| `actions/checkout` | v2 | v7 |
| `actions/upload-artifact` | v2 | v7 |
| `leafo/gh-actions-lua` | v8.0.0 | v13 |
| `lukka/run-vcpkg` | v7 | v11 |
| `ashutoshvarma/action-cmake-build` | `@master` | removed |
| `lukka/run-cmake` | v3 | removed |

`lukka/get-cmake@latest`, `thedoctor0/zip-release@0.7.6` and `ncipollo/release-action@v1` were already current and are unchanged.

`ashutoshvarma/action-cmake-build` has had no commits since 2024 and runs on a retired Node runtime; it only shelled out to `cmake`, so the workflows now call `cmake` directly. `lukka/run-cmake@v10` only drives `CMakePresets.json`, which this branch does not have, so those steps became direct `cmake` invocations too.

### Runner images

`ubuntu-20.04` was removed from GitHub-hosted runners, so `build-ubuntu.yml` and `coverity.yml` move to `ubuntu-24.04`. (An earlier revision of this PR used `ubuntu-22.04`; the C++23 changes below need GCC 13's `<format>` and `std::to_underlying`, which 22.04's GCC 11 does not have.)

The vcpkg jobs previously used `${{ matrix.os }}-latest`. They now pin `ubuntu-24.04` and `windows-2022` explicitly so the toolchain does not drift under this branch, and macOS moves from the `x64-osx` triplet to `arm64-osx`, since `macos-latest` is Apple silicon and the old triplet no longer matches the runner.

### vcpkg

`run-vcpkg@v11` dropped the classic-mode `vcpkgArguments` / `vcpkgTriplet` / `appendedCacheKey` inputs and is manifest driven, so the package list moves out of the workflow files and into a new `vcpkg.json`. This is the one change outside `.github/` and it is what makes the migration possible — worth calling out for review.

The dependency set follows the source changes below: standalone `asio` replaces the Boost libraries, and `fmt` is gone. Only the header-only `boost-lexical-cast` and `boost-variant` remain.

LuaJIT keeps its existing per-platform handling — apt on Ubuntu, brew on macOS, vcpkg on Windows. The Windows `luajit=off` combinations are excluded rather than silently building with LuaJIT on, which is what the split unix/windows jobs did before.

### Artifact uploads

The vcpkg jobs uploaded the binary, the Windows exe/dlls and the datapack as separate steps that all shared one artifact name. That relied on `upload-artifact@v3` merging uploads into a single artifact; v4 removed that and returns a conflict on the second upload. The binary and datapack are now collected into the workspace and uploaded by a single step, preserving the original intent of one downloadable bundle. A guard fails the job when no binary was produced, rather than silently publishing a datapack-only artifact.

### Docker

`docker.pkg.github.com` has been shut down. `docker-image.yml` now publishes to `ghcr.io` via `docker/login-action` and `docker/build-push-action`, mirroring what branch 1.6 already does, and replaces the hand-rolled `docker login`/`tag`/`push` block.

### `build-ubuntu.yml` trigger

The job carried `if: github.event_name == 'push'`, which disabled it on exactly the pull requests the workflow declares a `pull_request` trigger for. Removed, so the workflow now does something on PRs.

### Source changes: standalone Asio, no fmt, C++23

Bringing the CI back does not help if the branch no longer compiles against current dependencies, so this PR also carries the source-level fixes:

- **#5086 (@Zbizu)**, cherry-picked as-is: replaces Boost.Asio with standalone Asio, `fmt` with `std::format`, Boost.Filesystem with `std::filesystem`, Boost.Iostreams' `mapped_file` and Boost.Lockfree with small self-contained implementations, and moves the vc17 project to C++23. The follow-up "project name" commit from that PR is folded into the `vcpkg.json` conflict resolution.
- **`Fix build with Boost 1.90, fmt 10 and GCC 15`** (from `DSpeichert:v1.4-fix`), rebased on top of #5086: the Asio API modernisation (`io_context`, `expires_after`, `asio::post`, `make_address_v4`, `executor_work_guard`, `to_uint`) applies just the same to standalone Asio, so those hunks are kept in `asio::` form. The `lockfree.h` rebind and the enum-formatting fixes were already covered by #5086 and dropped from this commit.
- **`build: switch CMake, CI and Docker to standalone Asio and C++23`**: #5086 only touched the vc17 project and the vcpkg manifest, so `CMakeLists.txt` still required Boost.Asio and fmt. CMake now targets C++23, finds standalone Asio (vcpkg's `asio::asio` config or a plain `asio.hpp` on the include path), and links no Boost components. The apt lists in `build-ubuntu.yml`/`coverity.yml` and the `Dockerfile` (alpine 3.15 → 3.20 for GCC 13 and `asio-dev`) are updated to match.

- **`build: fix vcpkg builds on all platforms`**: the first CI run of the vcpkg matrix failed everywhere. Ubuntu/macOS linked vcpkg target names that the current ports no longer export (`libmariadb` → `unofficial::libmariadb`, `cryptopp-static` → `cryptopp::cryptopp`, the latter only ever "found" on macOS via case-insensitive matching); Windows lost the backslashes of the unquoted toolchain path in a bash step and now uses the `VCPKG_ROOT` exported by `run-vcpkg`.

Verified locally: Release + LuaJIT CMake build with GCC 15.2, Boost 1.90 headers and asio 1.30 completes with zero warnings under `-Wall -Werror`, and the binary starts. The same configuration also builds in manifest-mode vcpkg (x64-linux) against this manifest's baseline. The `Dockerfile` build stage was also built locally on alpine 3.20.


Znote:
### Testing to see how slop my AI review process is
copilot instructions: [Ponytail](https://github.com/DietrichGebert/ponytail)
git project environment setup locally to let the agent cook. 
Prompt: (Claude Fable 5, High 1M)
```md
You are on a github copilot chat harness in vscode. 
This project has pulled in a git repo for the forgottenserver code, (remote origin, renamed to otland)
I have also added another remote DSpeichert. This is a fork who occasionally submits pull requests to otland. 

He has submitted this PR:
https://github.com/otland/forgottenserver/pull/5089

Which in this project, I have pulled and checked out: DSpeichert/workflows/modernize-actions-1.4

Please review the pull request. Since github can be unstable, feel free to use local git commands to explore. 

Provide feedback as a markdown file/code so I can copy it back to the pull request. Feedback should be clear, concise, not nitpicky. If there are more than 3 things to say, prioritize and pick the most important 3 things. 
```

Updated todo list

Ran terminal command:  git log --oneline -10 && git branch -a | head -20 && git remote -v

Ran terminal command:  git diff --stat otland/1.4...HEAD

Ran terminal command:  git log otland/1.4..HEAD --format='--- %h %s ---' --stat | head -120



Fetched https://github.com/otland/forgottenserver/pull/5089

Ran terminal command:  git diff otland/1.4...HEAD -- src/lockfree.h src/fileloader.h src/fileloader.cpp

Now let me review the workflow files and check how the matrix issues flagged earlier were resolved, plus a few source spots.  

Read 

Read 

Ran terminal command:  ls .clang-format 2>/dev/null; grep -rn "LockfreePoolingAllocator" src/ | grep -v lockfree.h; git diff otland/1.4...HEAD -- src/connection.cpp src/connection.h src/server.cpp src/server.h src/scheduler.cpp src/otserv.cpp | head -250

Read , lines 1 to 280

Read , lines 280 to 304

Ran terminal command:  git diff otland/1.4...HEAD -- src/scheduler.cpp src/scheduler.h src/server.cpp src/server.h src/luascript.cpp src/luascript.h src/rsa.cpp src/otpch.h | head -220

Read 

Ran terminal command:  grep -n "catch (std::error_code" src/connection.cpp; grep -n -A2 "startupErrorMessage" src/otserv.cpp | head -8; grep -n "expires_after\|death_timer" src/server.h | head

All three findings confirmed. Writing the review file. 

Created 

Review written to `PR-5089-review.md`. How I got there:

- Diffed `otland/1.4...HEAD` (5 commits, 48 files) and read the PR page for context.
- The workflow changes are solid — the earlier bot complaint about the matrix `include` not merging was wrong (include entries keyed on `name` merge fine, and 11/11 checks passed on the final commit).

The three points I kept, in priority order:

1. **`lockfree.h` is now a mutex-guarded pool** on the `OutputMessage` allocation hot path — the name lies, and a locked pool likely loses to plain `operator new`, so simplest fix is deleting the allocator entirely.
2. **`MappedFile` reads via `istreambuf_iterator`** (byte-by-byte) — a real startup regression for large `.otbm` maps; one-line fix with a sized `read()`.
3. **Two conversion slips**: `catch (std::error_code&)` in `ConnectionManager::closeAll` (never thrown; should be `std::system_error`), and `startupErrorMessage` moved the newline before the message.

Things I noticed but left out as too nitpicky: `VCPKG_BUILD_TYPE: release` as an env var is a no-op (it's a triplet-file variable), and the new code in `lockfree.h`/`fileloader.h` uses 4-space indent while the rest of the codebase uses tabs.

Created 2 todos

Made changes.
(58.9K CT, 152.3 credits). 

# PR-5089-review.md
Thanks for this — the workflow modernization is well researched and well documented, the Asio/`std::format` migration is mechanical and clean, and having 11/11 checks green speaks for itself. Three things worth addressing, roughly in order of importance:

### 1. `lockfree.h` is no longer lock-free — and probably shouldn't exist anymore

The Boost.Lockfree stack was replaced with a `std::mutex`-guarded `std::vector`. That's a fine simplification, but the file/class names now lie, and more importantly this allocator sits on the `OutputMessage` hot path (`outputmessage.cpp`), where every network message allocation/deallocation now takes a global mutex shared across all threads touching that size class.

The original pool only existed to make allocation cheap and contention-free. A mutex-guarded pool is unlikely to beat plain `operator new` (modern allocators are already fast and thread-caching), so consider just dropping `LockfreePoolingAllocator` and using `std::make_shared<OutputMessage>()` — less code and no misleading name. If you keep the pool, a comment noting the deliberate tradeoff (and a rename) would help future readers.

### 2. `MappedFile` reads the map byte-by-byte

```cpp
data_ = { std::istreambuf_iterator<char>(file), std::istreambuf_iterator<char>() };
```

`istreambuf_iterator` reads one char at a time and grows the vector incrementally — for `.otbm` files that can be tens or hundreds of MB, this is a noticeable startup regression vs. the old `mmap`. A single sized read keeps the same simplicity:

```cpp
auto size = std::filesystem::file_size(path);
data_.resize(size);
file.read(data_.data(), size);
```

### 3. Two small slips in the Boost → std conversion

- `connection.cpp` (`ConnectionManager::closeAll`): `catch (std::error_code&)` — `std::error_code` is never thrown, so this catches nothing. The equivalent of the old `boost::system::system_error` is `std::system_error` (used correctly everywhere else in the file).
- `otserv.cpp` (`startupErrorMessage`): the fmt version printed `> ERROR: <msg>\n`; the new version prints `> ERROR: \n<msg>`, moving the newline before the message.

